PR #17, opened 2016-04-29
Add Dockerfile

pedrohml:
Hello,
I started using a Dockerfile to run the application and its dependencies as well as solving CORS issue with a **nginx** proxy.

### Readme

The **Dockerfile** is helpful in development mode to run the application and its dependencies.

##### Fast way (using **docker-compose**)

Commands:
```sh
cd docker/
docker-compose up
```

##### Flexible way (using **docker**)

Commands:
```sh
cd docker/
docker build -t ohif-viewers .
docker run --rm -p 8080:8080 -p 3000:3000 ohif-viewers
```

##### Usage

- OHIF Viewers: http://*DOCKER_MACHINE_IP*:3000
- Orthanc port: http://*DOCKER_MACHINE_IP*:8080

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,41 @@
+FROM jodogne/orthanc-plugins
+
+RUN apt-get update && apt-get install -y nginx curl unzip
+
+# ENVIRONMENT VARIABLES
+ENV METEOR_VERSION 1.2.1
+ENV OHIF_PATH /opt/OHIF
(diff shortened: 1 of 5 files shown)


swederik:
Hi Pedro,

Thanks for this pull request. This is probably not very necessary anymore since we now have the CORS proxy built into the Meteor server. For those that want to use Orthanc (or DCM4CHEE) on Docker, we can just point Meteor towards that.

Cheers,

Erik

PR #18, opened 2016-05-03
Add script to upload dicom images

pedrohml:
Hi there,
Since I needed to upload multiple dicom images to orthanc server I think this script would be useful. I propose to include this script in the original project to use mainly in development environment.
This script uses curl to upload a single dicom file or a directory recursively filtering by `*.dcm` and `*.dicom` extensions . 

Usage:

``` sh
DICOMWEB_HOST=<HOST> DICOMWEB_PORT=<PORT> scripts/dicom_upload.sh <DIRECTORY>|<FILE>
```

### DIFF
--- a/scripts/dicom_upload.sh
+++ b/scripts/dicom_upload.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+HOST=$DICOMWEB_HOST
+PORT=$DICOMWEB_PORT
+
+if [[ -f "$1" ]]; then
+	DICOM_FILE=$1
+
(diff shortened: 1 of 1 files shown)


swederik:
Hi Pedro,

Thanks for the contribution! We would rather not include utility functions in this repository. It's also probably just as easy for users to use DCMTK's storescu (http://support.dcmtk.org/docs/storescu.html).

If you put it in a Github gist we can link to it from the documentation, though!

Cheers,

Erik

PR #48, opened 2016-11-30
Add default user/pass to readme

quasarj:
It took a lot of time, googling and luck to figure out the default username/password for the docker image. Adding it to the README would probably be a good idea.


swederik:
Thanks! I will add this to the Wiki as well.

PR #112, opened 2017-09-27
OHIF-110: context-menu for measurements in the viewport

andrebot:
# Feature
Image measurements shall be removable through a dialog toggled by 
1 - Right-clicking on the measurement;
2 - Long-pressing on the measurement on a touch-capable device;

# Code
There were a few edge cases to take care of. First of all is how cornerstone name their tools and how it names its manager, there are mismatches for `simpleAngle`, which uses the toolManager `angle`, and `arrowAnnotate` which uses `annotate`.

The second catch was the `arrowAnnotate` which already had a solution for removing the annotation for mobile. When you do a long touch you can edit the annotation and there is a button to remove it. So I decided to keep that behavior and to not add the context-menu for that.

# Demos
Example for desktop:
context-menu

Example for mobile:
context-menu-mobile

### DIFF
--- a/Packages/ohif-core/client/components/base/mixins/dropdown.js
+++ b/Packages/ohif-core/client/components/base/mixins/dropdown.js
@@ -105,10 +105,19 @@ OHIF.mixins.dropdown = new OHIF.Mixin({
 
                 // Change the dropdown position if mouse event was given
                 if (event) {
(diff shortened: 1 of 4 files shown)


swederik:
Nice! Thanks!

PR #130, opened 2017-11-14
OHIF-184: download viewport

andrebot:
# Feature
We Added a new feature that allows us do take a snapshot of the active viewport and download it to our desktop. We can choose the name and one of two extensions available (PNG and JPEG).

The tools that are being able to reproduce in the image are: 

```
instance.availableTools = ['length', 'probe', 'simpleAngle', 'arrowAnnotate', 'ellipticalRoi', 'rectangleRoi'];
```

downlaodfeature

# Solution
The approach coded was to create a dynamic cornerstone enabled element to mimic all the transformation done into the image and download from there. We have a preview canvas in the dialog which previews the image selected and the annotations that will be shown. The size will not affect the preview, just the downloaded image to avoid a bad UX with the dialog being resized.

We are using a cornerstone enabled dynamic element to take advantage of all build-in image resize in the library. That way we write less code and something more readable.

# Improvements
We can make the available tools to be downloaded as configuration in the config files

# Depends on

- [ ] [Image Quality option](https://github.com/chafey/cornerstoneTools/pull/287)


Packages/ohif-viewerbase/client/components/viewer/downloadDialog/downloadDialog.js
```diff
@@ -0,0 +1,111 @@
...
+            cornerstone.fitToWindow(instance.$downloadElement);
+          });
+        }
+    })
+});
+
+Template.downloadDialog.onRendered(() => {
+    const instance = Template.instance();
+    const $dialog = instance.$('#downloadDialog');
+    const singleRowLayout = OHIF.uiSettings.displayEchoUltrasoundWorkflow;
```
swederik: What is this for?

PR #169, opened 2018-03-13
Feature/test suite

andrebot:
# Test Environment
This PR propose a test environment for OHIF repository. This environment was built with:
 - [Mocha](https://github.com/cult-of-coders/meteor-mocha)
 - [Chai](https://github.com/practicalmeteor/meteor-chai)
 - [Sinon](https://github.com/practicalmeteor/meteor-sinon)
 - [meteor-coverage](https://github.com/serut/meteor-coverage)

There is a test script inside of `/bin/testPackages.sh` folder so it could run all tests from all packages.

The way this was built it is expected that each packages implements its tests. We are able to test client and server side code.

It was added a small doc to first guide people to how to run the test. One thing we can add is more examples and how to make tests in it. In other words, just to improve the doc itself.

After you run the `/bin/testPackages.sh` a server at `http://localhost:3000` will be deployed with all tests results:
screen shot 2018-03-13 at 11 49 57

Thanks for @maltempi for coding this up


maltempi:
Note about coverage report:
1. To access the coverage report, just try `localhost:3000/coverage`
2. The coverage report does not work for template testings. It works only for direct methods calls, like a class or function.

``` 
// coverage does not work
describe(function() {
    const percentage = Template.gridLayout.__helpers[' height']();
});

// coverage works well -- using from a pure js class/function
describe(function() {
    const percentage = new GridLayout().getHeight();
});
```

swederik:
Thanks!

PR #190, opened 2018-05-08
Fix #172 - loading web images in StandaloneViewer (jpg, png...)

### LINKED ISSUE #172: Standalone client build i.promise.then is not a function
Hi, I tried making a standalone client build but I get this error

image

I tried using the client build included in the repository and it does work, but not with a new build

Zaid-Safadi:
This PR includes the cornerstone-web-image-loader as a package dependency to the  StandaloneViewer only and delete the old cornerstone-web-image-loader js file.


swederik:
Thanks!

Issue #145, opened 2017-12-10
Why does DIMSE still use DicomWeb?


EmilOJ:
Hi all,
First I want to say thank you for this amazing open source software. It's helped me tremendously in my work. 

I have a quick question. I want to use the OHIF Viewer in my hospital. We have a PACS server by GE which only supports DIMSE (no DicomWeb). For testing I tried disabling the DicomWeb plugin in Orthanc to simulate our PACS server and running the OHIF Viewer with the `orthancDIMSE.json` config file. 
In the studylist it fires a `C_FIND` request to Orthanc and successfully receives and displays the studies. 
But when I double click it and the viewer loads I get the error:
`An error has occurred while loading image: dicomweb:/__wado_proxy?url=http....`

I traced it to here: 
`https://github.com/OHIF/Viewers/blob/00e4949ab592c99cb36b7fecdb89ab73027502a4/Packages/ohif-studies/imports/server/services/dimse/retrieveMetadata.js#L136`
where the comment says 
`Retrieve the actual data over WADO-URI`
So although the metadata is retrieved via DIMSE, the pixel data is still retrieved via DicomWeb it seems.
Am I right in concluding that OHIF Viewer does not support a pure DIMSE PACS server? I see that the package `Ohif-dicom-services` does support `C_GET` but it's not used anywhere. 

Kind regards,
Emil

radicalrob:
Hi Emil,

You are correct, currently the viewer can only get pixel data using WADO, so if the GE PACS doesn't support that, then you'd need to first move the images into something (like Orthanc) that does.  We're working on improving support for this use case, but don't have anything working yet.  

- Rob

EmilOJ:
Good to know. Thanks for the swift reply. 

Emil

EmilOJ:
Hi,
Just a quick update. 
I manged to implement Rob's idea. The OHIF Viewer actually already implements C-MOVE and so the pipeline is this:
1. When searching in the studylist, the OHIF viewer sends a C-Find to the actual PACS server and displays the results.
2. When double clicking a study, the OHIF VIewer sends a C-Move request to the PACS server with the destination set to an Orthanc server. Once the entire study is moved to Orthanc, the OHIF Viewer continues and performs its usual magic of loading the instances via WADO from the Orthanc server.

Setting a maximum storage size on the Orthanc (which makes sure to delete the oldest instances once its full) makes it act as the perfect temporary storage and thus for all practical purposes as a DicomWeb proxy. 
[...]

PR #191, opened 2018-05-09
Feature: DIMSE proxy

EmilOJ:
As discussed in #145, this shows how to use the OHIFViewer and a dicomWeb proxy to emulate true DIMSE functionality. OHIF is fully capable of communicating via DIMSE but cannot receive image data over this protocol. 
The process is shown in the following flow chart:

DIMSE proxy flowchart 

The main changes to the OHIF code are:
`OHIFViewer/client/routes.js`: Perform a serverside move request of the selected studies to the PACS server before redirecting to the viewer.

`Packages/ohif-dicom-services/server/DIMSE/DIMSE.js`:
1. Force synchronous execution (with `future.wait()`), so the viewer is not rendered before the C-MOVE is done
2. Initiate a C-MOVE request on the STUDY level with the DicomWeb proxy as destination (the destination is hardcoded as of now). 

I have included a new server config file `config/orthancDIMSEproxy.json` for this setup and a `docker-compose.yml` for setting up a dummy PACS and a DicomWeb proxy using Orthanc for testing. 

For testing, follow this workflow:
1. Start OHIFviewer the old way using the included start script in `OHIFViewer/package.json`
2. Start the two orthanc servers with `OHIFViewer/docker-compose.yml`
3. Upload a DICOM series to the "PACS" server at localhost:8042
4. Open OHIFViewer at localhost:3000, this study should now be in the study list (fetched via DIMSE)
5. Double click the study. The study should now have been moved to the DicomWeb proxy server at localhost:8043 and loaded into OHIF via WADO.

### DIFF
--- a/OHIFViewer/client/routes.js
+++ b/OHIFViewer/client/routes.js
@@ -16,10 +16,19 @@ Router.route('/studylist', function() {
     this.render('ohifViewer', { data: { template: 'studylist' } });
 }, { name: 'studylist' });
 
-Router.route('/viewer/:studyInstanceUids', function() {
-    const studyInstanceUids = this.params.studyInstanceUids.split(';');
-    OHIF.viewerbase.renderViewer(this, { studyInstanceUids }, 'ohifViewer');
-}, { name: 'viewerStudies' });
+Router.route('/viewer/:studyInstanceUids', {
+    name: 'viewerStudies',
+    action: function() {
+      const studyInstanceUids = this.params.studyInstanceUids.split(';');
+  
+      Meteor.call('MoveRequest', {
+        studyInstanceUid: studyInstanceUids[0]
+      }, function(error, result){
+        console.log("DONE", error, result);
+      });
+  
+      OHIF.viewerbase.renderViewer(this, { studyInstanceUids }, 'ohifViewer');
+  } });
 
 // OHIF #98 Show specific series of study
 Router.route('/study/:studyInstanceUid/series/:seriesInstanceUids', function () {
--- a/OHIFViewer/docker-compose.yml
+++ b/OHIFViewer/docker-compose.yml
@@ -1,0 +1,30 @@
+version: '2'
+
+services:
+  pacs:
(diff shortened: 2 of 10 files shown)


swederik:
This is super awesome!

I might want to put the docker-compose setup in https://github.com/OHIF/integration-examples instead, but in general it's really cool that this works with so few changes.

EmilOJ:
Absolutely, I agree. I just wanted to include it so you can easily test it out. 

Of course this also breaks the normal DicomWeb-only setup that most people use, so there should be some way of enabling/disabling this functionality.

EmilOJ:
I'm happy to implement any suggestions you have to integrate this.

dannyrb:
@EmilOJ, thank you for such a high quality PR. Many, many users don't take the time to contribute back, so I want you to know that myself and the other maintainers truly appreciate seeing PRs like this.

Unfortunately, I'm going to close this PR. Since this PR was drafted, we've made a shift to a new major version that does not support DIMSE. I've added a `bookmark` tag, as I feel `version 1` users and others may find the content here useful.

EmilOJ:
Just leaving this here for reference for someone less experienced with the tools I mentioned in the PR description.

For step two you need have have docker-compose installed on your system.
https://docs.docker.com/compose/

Then from your terminal go the directory that the `docker-compose.yml` file is in and type the command `docker-compose up` .
This will start two Orthanc PACS servers on your laptop. 
Now you should be able to open your browser and go to localhost:8042
And login to the Orthanc server with credentials:
Username: orthanc
Password: orthanc
(I think this is correct)

Then follow the rest of the steps.
You have to setup the two Orthanc servers in the OHIF viewer (There’s an ‘add server’ or something like that option) just like you did with your DIMSE connection.
I don’t remember the exact configurations Im afraid. But I think that it’s something like below:

image

pablosbravo:
Hi everyone, where is the docker-compose.yml file? I can not  find it

sedghi:
@wayfarer3130

PR #204, opened 2018-06-11
fix: first time uploadMessage called with undefined as parameter

casaper:
- During uploading DICOMs the first time the function is called with
  no parameter -> this leads to the whole upload to be crashing


swederik:
Thanks!

swederik:
Thanks!

Issue #203, opened 2018-06-10
Series Details not working in Dropdown View Study Details-> Study LIst


nagodoy:
When I select the View Series Details option, the Dialog box loads only the demographic information from the Exam.
Detailed information does not load.
I tested the demo version of the OHIF site and error happens too
![captura de tela 2018-06-10 10 37 08](https://user-images.githubusercontent.com/3387863/41202192-d7b16812-6c9b-11e8-916d-997b537c79dd.png)

IZuaL05:
Change the code in Package/ohif-study-list/client/components/seriesDetailModal/seriesDetailsTable/seriesDetailsTable.js 

Template.seriesDetailsTable.onRendered(() => {
    const instance = Template.instance();
    const studies = instance.selectedStudies.get('studies');

    if (!studies) {
        return;
    }
    // Get series list for the study
    _.map(studies, (selectedStudy, index) => {
        studies[index].displaySeriesLoadingText = true;
        studies[index].seriesList = [];
        var study=OHIF.studies.retrieveStudyMetadata( selectedStudy.studyInstanceUid);
        study.then(function (result) {
            studies[index].seriesList = result.seriesList;
            studies[index].displaySeriesLoadingText = false;
            instance.selectedStudies.set('studies', studies);
        })
    });
});



it works for me

PR #214, opened 2018-07-03
Fix Series details modal info table

galelis:


swederik:
This should have closed https://github.com/OHIF/Viewers/issues/203

PR #222, opened 2018-07-13
feat(WADO) Allow WADO / QIDO requests to originate from client

swederik:
This moves the WADO / QIDO DICOMWeb functions to a separate package. This will likely be moved to an NPM package in the near future.

It updates the services to refer to the moved code, and adds conditional statements to execute queries from the client rather than the server, if the ** requestFromBrowser** Boolean is true.

# Notes:
- For users using HTTP basic authentication (e.g. with Orthanc user:password), making requests from the client will expose your credentials to any user which has access to the Servers collection. Without access controls, this is everyone.
- For requests from the client to be made securely, the system should allow users to log in, provide them with an access token, and then requests to the PACS / backend should be authenticated using this token. Tokens should expire, be revokable, etc..


pieper:
@swederik and @hackermd we should figure out how to coordinate our efforts on this front.  

https://github.com/dcmjs-org/dicomweb-client

It would be great if we could have OHIF use the general purpose npm package.

swederik:
100% agreed. I think we can deprecate ours and use that one instead. We'll just need to figure out what is supported already / what is missing.

pieper:
One thing missing is automated testing.  I just enabled travis but need to figure out how to set up a server.

PR #223, opened 2018-07-13
Fix importStudies dialog when import fails

frmdstryr:
If an import happens to fail the dialog promise never resolves. 

This is because `numberOfStudiesFailed` is incremented on error but `numberOfStudiesImported` is not, so the branch  to resolve or reject is never taken.


swederik:
Thanks!

PR #225, opened 2018-07-16
Display support for measurements from DICOM Structured Reports

galelis:
This pull request adds:
- Limited DICOM-SR read support for [TID 1500 Measurement Report](http://dicom.nema.org/medical/dicom/current/output/html/part16.html#sect_TID_1500)
- Sidebar UI - Measurement table to display length measurement read in from the structured report
- Measurement sidebar and limited SR read support for the Standalone Viewer

It changes:
- Sidebar UI - Removes protocol editor from the UI. The components remain in the protocol engine package, but are not used by default

**Note:** This is the first step in the process of supporting DICOM Structured Reports for all measurement types. At this stage we only support Length measurements.

Left for future PRs:
- Storage of created / edited measurements (C-STORE and STOW-RS)
- Support for other measurement types

### DIFF
--- a/LesionTracker/client/components/viewerSection/viewerSection.styl
+++ b/LesionTracker/client/components/viewerSection/viewerSection.styl
@@ -57,19 +57,19 @@
 
   .sidebar-right
     flex: 1
-    margin-right: - $lesionsSidebarMenuWidth
-    max-width: $lesionsSidebarMenuWidth
+    margin-right: - $rightSidebarMenuWidth
+    max-width: $rightSidebarMenuWidth
     order: 3
     position: relative
     visibility: hidden
 
     &[data-timepoints="3"]
-      margin-right: - ($lesionsSidebarMenuWidth + 135.5px)
-      max-width: $lesionsSidebarMenuWidth + 135.5px
+      margin-right: - ($rightSidebarMenuWidth + 135.5px)
+      max-width: $rightSidebarMenuWidth + 135.5px
 
     &[data-timepoints="4"]
-      margin-right: - ($lesionsSidebarMenuWidth + 270px)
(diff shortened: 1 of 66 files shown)


pieper:
Hi @galelis is this ready for review and testing?  Could you put it on a heroku site for testing?

Thanks,
Steve

galelis:
@pieper Hi Steve,
Yeah, we are reviewing and testing, I have fixed a few callouts and introduce the table into Standaloneviewer as well. 
Also create new test environments if you want to test: 

http://ohifviewer-tablev1.herokuapp.com/
https://standaloneviewer.herokuapp.com/

For standalone I had to create a json ile for test, so we can use this link:
https://standaloneviewer.herokuapp.com/?url=https://raw.githubusercontent.com/galelis/testData/master/studyWithMeasurements.json

Let me know if you see any issue.

pieper:
Hi @galelis great, this makes it much easier to test 👍 

A few quick things:
* I notice that the measurements doesn't auto-open anymore.  Was this on purpose?
* @swederik told me he asked you to remove the store but maybe we should still have it on the ohifviewer and not on standalone?  I like being able to test the roundtrip and show it to collaborators.  CSV and PDF export are also impressive to show people.
* I can now edit the description (nice!) but the input field isn't focused - it would be good to focus it when the dialog is popped up.

Otherwise it seems to be working as expected.

swederik:
Can we keep store for the next PR? This one is huge enough already, it's hard enough to review it as it is.

pieper:
No problem holding off stow for another PR.  Just want to be sure we don't lose it - maybe once this is in master switch over the heroku demo to include the stow options.

PR #236, opened 2018-07-23
STOW Structured Reports from Viewer's measurement table

galelis:
## Added
- Limited DICOM-SR STOW support for TID 1500 Measurement Report
- For Viewers that have server support the measurement table will be able to stow measurements (limited to Length measurements)
- Export CSV functionality included into measurement-table

## Left for future PRs:
Support to C-STORE
Support for other measurement types


Packages/ohif-measurements/client/components/measurementLightTable/measurementLightTableView/measurementLightTableView.js
```diff
@@ -8,9 +8,67 @@ Template.measurementLightTableView.onCreated(() => {
...
+
+    hasAnyMeasurement() {
+        const instance = Template.instance();
+        const groups = instance.data.measurementGroups.get();
+
+        if (!groups) {
+            return false;
+        }
+
+        const group = _.find(groups, item => item.measurementRows.length > 0);
```
swederik: Could you use [Array.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead?

Packages/ohif-measurement-table/client/utils/srUtils.js
```diff
@@ -0,0 +1,81 @@
...
+    
+    multipartArray.set(headerArray, 0);
+    multipartArray.set(contentArray, headerArray.length);
+    multipartArray.set(footerArray, headerArray.length + contentArray.length);
+    
+    return(multipartArray.buffer);
+};
+
+const getWADOProxyUrl = () => {
+    const server = OHIF.servers.getCurrentServer();
```
swederik: This function should probably be moved to OHIF.servers. It should use the value of proxy: enabled (true / false) to build the STOW-RS URL itself.

The _wado_proxy stuff should not be hardcoded. It should be created by the WADOProxy.convertURL as it is here: https://github.com/OHIF/Viewers/blob/ebe0a40f3ef5fa7d13c66e7076ddeb1f42e755b5/Packages/ohif-studies/imports/both/services/wado/retrieveMetadata.js#L306

Packages/ohif-measurement-table/client/configuration/dataExchange.js
```diff
@@ -18,10 +19,11 @@ export const retrieveMeasurements = (patientId, timepointIds) => {
 export const storeMeasurements = (measurementData, timepointIds) => {
     OHIF.log.info('storeMeasurements');
 
-    // Here is where we should do any required data transformation and API calls
+    const studyInstanceUid = measurementData[Object.keys(measurementData)[0]][0].studyInstanceUid
 
-    // TODO: Write SR, STOW back to PACS
-    return Promise.resolve();
+    return stowSRFromMeasurements(measurementData).then( () => {
```
swederik: Here we should check what type of server we are connected to and act accordingly.

A similar example is used for retrieving metadata here:

https://github.com/OHIF/Viewers/blob/ebe0a40f3ef5fa7d13c66e7076ddeb1f42e755b5/Packages/ohif-studies/imports/client/lib/retrieveStudyMetadata.js#L49

There are three potential scenarios:
- The server is DICOMWeb and we can make requests from the client (e.g. we have a Bearer token or are connecting to an insecure server)
- The server is DIMSE, so we have to make requests from the server
- The server is DICOMWeb, but we have to make requests from the server (e.g. if we are using HTTP basic auth and don't want to expose the user/pass to the client)

In both of the DICOMWeb cases, we can use STOW-RS. In the DIMSE case, we will need to write something to support C-STORE.

For now, we can hide the 'Save to PACS' button when we are using a DIMSE server, and we can get that functionality in another PR.

Packages/ohif-measurement-table/client/utils/request.js
```diff
@@ -0,0 +1,35 @@
+/**
```
swederik: We already have a function that does this. Please switch this to use  

https://github.com/OHIF/Viewers/blob/885dd9da2d0295290a61b9b591a4238387e88725/Packages/ohif-dicomweb-client/src/makeRequest.js

Packages/ohif-measurement-table/client/utils/retrieveDataFromMeasurements.js
```diff
@@ -0,0 +1,273 @@
+import { dcmjs } from 'meteor/ohif:cornerstone';
+
+const sopClassUid = dcmjs.data.DicomMetaDictionary.sopClassUIDsByName["CTImage"];
+
+const measurementGroupContentItem = (measurement) => {
```
swederik: We should probably make it clear that this only handles length measurements.

Packages/ohif-measurements/client/components/measurementLightTable/measurementLightTableView/measurementLightTableView.html
```diff
@@ -9,4 +9,21 @@
             {{/each}}
         {{/each}}
     {{/let}}
+    {{#if hasAnyMeasurement}}
+    <div class="report-area">
+        {{#if saveEnabled}}
+            {{#unless noUnsavedChanges}}
```
swederik: Why do we need the double negative (unless no changes)? Can you change this to "{{# if hasUnsavedChanges }} ?

Packages/ohif-measurements/client/components/measurementLightTable/measurementLightTableView/measurementLightTableView.js
```diff
@@ -8,9 +8,67 @@ Template.measurementLightTableView.onCreated(() => {
...
+
     instance.autorun(() => {
         measurementApi.changeObserver.depend();
         const data = OHIF.measurements.getMeasurementsGroupedByNumber(measurementApi, timepointApi);
         instance.data.measurementGroups.set(data);
     });
 });
+
+Template.measurementLightTableView.helpers({
+    noUnsavedChanges() {
```
swederik: Switch to hasUnsavedChanges to avoid the double negative above.

Packages/ohif-measurements/client/components/measurementLightTable/measurementLightTableView/measurementLightTableView.js
```diff
@@ -8,9 +8,67 @@ Template.measurementLightTableView.onCreated(() => {
...
+
+        if (!groups) {
+            return false;
+        }
+
+        const group = _.find(groups, item => item.measurementRows.length > 0);
+        return group;
+    },
+
+    saveEnabled() {
```
swederik: For now, you can make this true if and only if the server uses DICOMWeb.

Packages/ohif-studies/imports/client/lib/retrieveStudyMetadata.js
```diff
@@ -6,6 +6,19 @@ import 'meteor/ohif:viewerbase';
 // promises and prevent unnecessary subsequent calls to the server
 const StudyMetaDataPromises = new Map();
 
+
+/**
+ * Delete the study promise cached to ensure next retrieve will get it retrieved by server call
```
swederik: This isn't super clear to me. Could you change it to:

'Delete the cached study metadata retrieval promise to ensure that the browser will re-retrieve the study metadata when it is next requested.'

Packages/ohif-studies/imports/client/lib/retrieveStudyMetadata.js
```diff
@@ -6,6 +6,19 @@ import 'meteor/ohif:viewerbase';
...
+ * 
+ */
+OHIF.studies.deleteStudyMetadataPromise = (studyInstanceUid) => {
+    if (StudyMetaDataPromises.has(studyInstanceUid)) {
+        StudyMetaDataPromises.delete(studyInstanceUid);
+    }
+};
+
 /**
  * Delete the study promise cached to ensure next retrieve will get it retrieved by server call
```
swederik: How come there is a duplicate of this function several lines below?

Packages/ohif-measurements/client/components/measurementLightTable/measurementLightTableView/measurementLightTableView.html
```diff
@@ -9,4 +9,21 @@
...
+            <div class="unsaved-changes-alert">
+                You have unsaved changes
+            </div>
+            {{/unless}}
+        {{/if}}
+        {{#form (extend api=instance.api)}}
+            {{#button class='btn p-x-2' action='exportCSV'}}Export CSV{{/button}}
+            {{#if saveEnabled}}
+                {{#button class='btn p-x-2' action='save' disabled=(noUnsavedChanges)}}Save to SR{{/button}}
```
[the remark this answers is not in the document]
pieper: I think Save to Archive might be clearest

Packages/ohif-measurement-table/client/utils/retrieveDataFromMeasurements.js
pieper: The CTImage shouldn't be used - instead the sopClassUID should be extracted from the dataset that is being measured.

Packages/ohif-measurement-table/client/utils/request.js
galelis: I checked out the function but it does not feed all our need in this case. 
Missing things: 
- responseType as arraybuffer on GET method. 
- able to perform POST sending body as well. 

Question: Is that safe for us to be using fetch() as its not supported by IE11 ? I faced that when I was adding my request.js file.

Packages/ohif-measurement-table/client/utils/srUtils.js
galelis: I have moved the function to OHIF.servers but I was unable to use WADOProxy.convertURL because it only works on server. 
I could make the WADOProxy (currentURL, namespace and initialize) available on client as well.
What do you think?

Packages/ohif-measurement-table/client/utils/request.js
swederik: Yep we are polyfilling fetch with the 'isomorphic-fetch' package (https://github.com/OHIF/Viewers/blob/885dd9da2d0295290a61b9b591a4238387e88725/Packages/ohif-dicomweb-client/src/makeRequest.js#L3)

Should be easy to add POST and response.arraybuffer() to the existing function.

PR #238, opened 2018-07-24
Fixed images' paths in README

zivsha:
Fixed broken relative paths to documentation's images


swederik:
Hi there,

This actually isn't a bug. The paths in the README inside 'docs' are not meant to be visible on Github. They are meant to be correct when deployed on http://docs.ohif.org/

Thanks for the contribution though!

zivsha:
Oh wow, I actually missed all possible links to the documentation webpage... 🤦‍♂️

PR #242, opened 2018-08-07
Add initial version of ViewportPlugin class

swederik:
This provides an extension point for plugins which can render data into viewports using custom methods.

Two examples (both under heavy development, nowhere near stable):

**VTK Plugin**: https://github.com/OHIF/VTKPlugin

This builds on Kitware's [vtk.js](https://github.com/Kitware/vtk-js) to add basic volume rendering support.

**OpenLayers Plugin**: https://github.com/OHIF/OpenLayersPlugin

This uses @hackermd's excellent work on using DICOMWeb to bundle and include a whole-slide microscopy viewer.

**Other changes:**
- Cleans up use of jQuery to set / change the 'active' class on a viewport. This is now a template helper ("activeClass").
- Switched a number of cases where active viewport was retrieved using the 'imageViewerViewport' div, and instead we are now looking for its parent ('viewportContainer'), because that will exist across all plugins.
- Removed duplicate 'viewportActivated' Session variable, since it's useless
- Stopped triggering LayoutManagerUpdated changes when active viewport was updated (I don't see why this was ever added)

# Usage
- Make a directory Viewers/OHIFViewer/public/plugins
- Clone the repos above into Viewers/OHIFViewer/public/plugins, use their 'plugins-1' branches
- Make a directory Viewers/OHIFViewer/client/plugins
- Add the two files in the [attached ZIP file](https://github.com/OHIF/Viewers/files/2267892/startup.zip) into the client/plugins folder.
- Note: You may have to point to a remote version of VTK.js, or put a local one into public/plugins/vtk.js

https://deploy-preview-242--ohif.netlify.com/viewer/?url=https://deploy-preview-242--ohif.netlify.com/viewer/DICOMCloud_public.json&studyInstanceUids=1.2.826.0.1.3680043.8.1055.1.20111103111148288.98361414.79379639

## TODO
- [x] Drag/drop seems broken
- [x] Active viewport can't be changed after a plugin has been used
- [x] Ensure we are cleaning objects up properly after destroying a viewport
- [x] Ensure 'LayoutManagerUpdated' Session variable is not being triggered unnecessarily
- [x] Travis CI seems to fail because Python is not available?


Packages/ohif-viewerbase/client/lib/classes/plugins/OHIFPlugin.js
```diff
@@ -50,7 +50,5 @@ class OHIFPlugin {
 // TODO: Move to OHIF.plugins.entryPoints?
 OHIFPlugin.entryPoints = {};
 
-export { OHIFPlugin };
-
 // TODO: Should we remove this? Authors should be able to use new 'OHIF.viewerbase.OHIFPlugin()'
 window.OHIFPlugin = OHIFPlugin;
```
pieper: I'd think OHIF.plugins.OHIFPlugin  actually makes sense to me - not as much global namespace pollution but not so obscure as calling a method to access a variable.

Then we can also have OHIF.plugins.entryPoints.

Packages/ohif-viewerbase/client/lib/classes/plugins/ViewportPlugin.js
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Stop listening for changes to the viewport layout in order to
+     * automatically rerender viewports setup for use by this plugin.
+     */
+    stopListeners() {
+        this.destroyed = true;
+    }
+}
+
+window.ViewportPlugin = ViewportPlugin;
```
pieper: Here again, I'd expect something like OHIF.plugins.ViewportPlugin

Packages/ohif-viewerbase/client/lib/classes/plugins/OHIFPlugin.js
pieper: Makes me think we are almost to the point of writing down the API and class diagram for all the things we want to support with plugins.

PR #270, opened 2018-09-16
Update package.json

cperandahl:
Fix the 404 bcrypt issue when building the OHIFViewer


swederik:
Looks like this is not necessary anymore. Thanks anyway!

PR #274, opened 2018-09-19
feat(plugins): add callbacks array to ViewportPlugin

DominionSoftware:
This will allow all plugins to add their callback for rendering on progressive load.


swederik:
I think this is only necessary for the VTK plugin, so you can include this.callbacks in the VTK plugin here instead: https://github.com/OHIF/VTKPlugin/blob/master/multiplanarReformatting/main.js#L7

DominionSoftware:
It’s so needed for the volume rendering plugin so I think it does make sense to be in their parent class?

Issue #275, opened 2018-09-21
Issue with servers when built for production


swederik:
This is my fault. Introduced with OIDC changes.

````
b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:199 For some reason we could not list the studies.
d.studies.searchStudies.then.t @ b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:199
b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:199 TypeError: Cannot read property 'type' of undefined
    at c (b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:197)
    at new Promise (<anonymous>)
    at Object.i.studies.searchStudies.t [as searchStudies] (b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:197)
    at v (b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:199)
    at Object.change #studyDate (b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:199)
    at b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:97
    at Function.e._withTemplateInstanceFunc (b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:97)
    at f.View.<anonymous> (b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:97)
    at b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:97
    at Object.f._withCurrentView (b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:97)
d.studies.searchStudies.then.t @ b5f31924bf2903bebdae0c745a01480e38405fa8.js?meteor_js_resource=true:199
````

dannyrb:
I believe this has been resolved? @swederik, please feel free to open if the issue persists. More info would be helpful if it still needs resolved.

PR #276, opened 2018-09-21
fix(client-only): Fix client-only Server & CurrentServer usage

swederik:
Addresses #275

- Add startup.js to copy public.servers into the client Minimongo collection if we are using public.clientOnly = true
- Set CurrentServer and Servers collection to client-only if clientOnly = true
- Call Meteor.disconnect() if we are using clientOnly = true
- Guard If statements requiring Meteor.settings.public, since it was not defined on the Standalone Viewer

### DIFF
--- a/OHIFViewer/client/config.js
+++ b/OHIFViewer/client/config.js
@@ -30,3 +30,9 @@ Meteor.startup(function() {
         }
     });
 });
+
+if (Meteor.settings &&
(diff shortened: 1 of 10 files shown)


jarrettmivu:
I have the most recent build and I have an identical "Cannot read property 'type' of undefined" when running on a linux server, build works fine on my windows server.

PR #277, opened 2018-09-21
I18n

juandelsolar:
First code changes to get internationalization (i18n) in StandaloneViewer using tap:I18N.


swederik:
Thanks! I'll take a look ASAP.

juandelsolar:
> This looks nice, and definitely a feature we want to help with. My one concern is that it ties us onto yet another Meteor package, just as we are making strides reduce those dependencies. [TAPi18n](https://github.com/TAPevents/tap-i18n) hasn't been updated in two years.
> 
> Is there any way we can do this without a Meteor-specific package. Can we use some i18n package from NPM instead of Atmosphere? I don't want to have to migrate this off of Meteor in the future.
> 
> Thanks!

I'll try to change TAPi18n to npm i18n. Can you create a i18n branch to use for this purpose?

Thanks!

swederik:
Hi @juandelsolar,

I'll close this issue because we are not supporting v1.x anymore, but I would encourage you to contribute through the links on this issue: https://github.com/OHIF/Viewers/issues/618

We actually used your translations for the Spanish version linked to in that post, so thanks a lot for that!

PR #297, opened 2018-10-16
WIP: Consolidate packages

swederik:
**Work in progress**

We want to consolidate the logic in the OHIF packages into smaller set of NPM modules (and push whatever we can out of OHIF and into things like [dcmjs](https://github.com/dcmjs-org/dcmjs), [dicomweb-client](https://github.com/dcmjs-org/dicomweb-client), [dicom-dimse](https://github.com/OHIF/dicom-dimse), and [Cornerstone](https://github.com/cornerstonejs), and push the UI into separate packages. The UI can then be rewritten in something other than Blaze so it's more portable.

We can maybe replace some of the OHIF components with the React components we're building in https://github.com/crowds-cure/cancer

Created the PR now even though this isn't ready because (1) I want my [Hacktoberfest](https://hacktoberfest.digitalocean.com) t-shirt and (2) I don't want to forget about this branch.


swederik:
Closing because I'm doing this in the react branch.

PR #311, opened 2018-10-31
Add orthanc in docker compose

chsasank:
app.json of docker-compose anyway is orthanc specific. Might as well include orthanc in docker compose. Also added instructions to run the viewer using docker compose in readme.


dannyrb:
https://docs.ohif.org/deployment/recipes/nginx--image-archive.html

Our latest, version 2.x, has docker support for use with `Orthanc` locally (with appropriate CORS headers), or alongside a container that has our WebApp in it.

PR #322, opened 2018-12-11
Standalone Viewer Authorization Header

### LINKED ISSUE #140: How to include an authorization header in the standalone viewer parameters?
Currently you can load a study in the standalone viewer by providing a URL query parameter. 

How can an authorization header be included in the request for:
1. The JSON study in the query url parameter
2. The DICOM images described in the JSON study

nicolasvandooren:
A token fragment parameter can be specified. If present this value will be used for http bearer authorization when making requests.

A fragment parameter is used instead of a query parameter so that the token is never sent over the network.

Example :
 http://localhost:3000/?url=https://some.website.io/sampleDICOM.json#token=1a2b3c4d

Fixes #140


swederik:
Awesome, thanks!

PR #325, opened 2018-12-24
Quantumsoft pull request 

pavertomato:
- google dicom storage 
- optional google auth for demo site
- SR view //seems like optional and may be removed for this moment?


swederik:
Thanks Egor! I'll take a look after the holidays!

swederik:
Now that this is merged: For anyone that has signed up for early access to the Google Cloud Healthcare API (https://cloud.google.com/healthcare/), you can now use OHIF against it with your own data! You can also access all of The Cancer Imaging Archive datasets and some RSNA Chest X-Ray data. You just need to put a clientId in here: https://github.com/OHIF/Viewers/blob/master/config/oidc-googleCloud.json#L17

PR #378, opened 2019-04-05
Restrict Scope for Cloud Healthcare API Example Config

dzelemba:
The integration only needs read access to projects not all cloud
platform resources.


swederik:
Ah whoops, I fixed that somewhere else but I guess I missed it in the config. Thanks!

swederik:
I just pushed an update to ohif/viewer-google-cloud:latest that includes this fix. It would be good to get that building / pushing on every commit.

PR #384, opened 2019-04-15
Strict versioning for `react-viewerbase`

dannyrb:
I couldn't determine if the most recent lock-files have an effective dependency on a more recent version. This may need to be bumped `v0.2.17` if you encounter issues.


dannyrb:
Already addressed here: https://github.com/OHIF/Viewers/pull/383
Closing

PR #391, opened 2019-04-18
Added missing dependency.

alvarosan:


dannyrb:
@alvarosan, is this a peerDependency of an extension? Is there a reproducible bug that adding this dependency fixes?

I'm mostly asking, because I'm confused why we don't see a build error in the CI if this is a missing dependency.

alvarosan:
Good point @dannyrb, I didn't actually check the CI earlier and assumed it was a missing dependency here (since the issue showed up after I pulled the latest react changes). It could indeed be a peerDep of my extension, let's close this pr for now.

PR #408, opened 2019-04-28
Docs/deploying static site

dannyrb:
- Update build scripts to include production PWA output (not just rollup)
- Update netlify to use output at `/demo`
- Update docs to note `stand-alone` asset generation
- Steps to deploy to common static asset hosts


dannyrb:
Alright, this is starting to get a little large. I have `react-scripts` using `.env` configs building "production" output for the `/demo` portion of our netlify deploy. It helped me clarify how/where we can set config at build time, as well as how to override those values for `script` includes by establishing config on `window.config` (may need a better name)

Peppered in a couple of tests and some light cleanup

PR #419, opened 2019-05-04
Cleaning up dependencies

biharck:
### What
Cleaning up the unused dependencies 

### Why
Some libraries are specified as dependencies instead of peerDependencies, which may cause unnecessary duplication at the application level. We should check this and see if there is some way to avoid it.


biharck:
WIP -> Summary Check up to now

cornerstone-tools                         😎  NEW VERSION!
dcmjs                                     😎  NEW VERSION!
husky                                     😎  MAJOR UP
ohif-core                                 😎  NEW VERSION
react-redux                               😎  MAJOR UP
react-resize-detector                     😎  MAJOR UP
react-router                              😎  MAJOR UP
react-router-dom                          😎  MAJOR UP
@babel/core                               😍  UPDATE!   
@babel/plugin-proposal-class-properties   😍  UPDATE!   
@babel/plugin-transform-runtime           😍  UPDATE!   
@babel/preset-env                         😍  UPDATE!   
babel-eslint                              😎  MAJOR UP  
eslint                                    😎  MINOR UP  
eslint-plugin-import                      😍  UPDATE!   
eslint-plugin-node                        😎  MAJOR UP  
eslint-plugin-react                       😍  UPDATE!   
jest-junit                                😎  MINOR UP  
lint-staged                               😍  UPDATE!   
prettier                                  😎  MINOR UP  
react-scripts                             😎  MAJOR UP  
rollup                                    😍  UPDATE!   
rollup-plugin-json                        😎  MAJOR UP  
semantic-release                          😕  NOTUSED?  
stylelint                                 😎  MAJOR UP  
eslint-config-react-app                   😟  PKG ERR!  Not in the package.json. Found in: /.eslintrc
react-transition-group                    😟  PKG ERR!  Not in the package.json. Found in: /src/components/Labelling/LabellingTransition.js
loglevel                                  😟  PKG ERR!  Not in the package.json. Found in: /src/log.js

package.json
```diff
@@ -60,18 +59,13 @@
     "react-dom": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
-    "@babel/polyfill": "^7.2.5",
-    "@babel/runtime": "^7.2.0",
```
dannyrb: Removing these probably wouldn't create a build error, but are we sure they weren't modifying the code during build to help hit a build target or pollyfill behavior not available in Safari/IE11?

package.json
```diff
@@ -60,18 +59,13 @@
...
-    "classnames": "^2.2.6",
     "cornerstone-core": "^2.2.8",
-    "cornerstone-math": "^0.1.8",
```
dannyrb: I think this is okay. I think we mostly/only use this in the ohif-cornerstone-extension.

package.json
dannyrb: For example, `@babel/polyfill` is required for `redux-oidc` to work:
https://github.com/maxmantz/redux-oidc#babel-polyfill

biharck: Good to know. Thank you for sharing that @dannyrb

package.json
```diff
@@ -60,46 +59,40 @@
...
-    "lodash.isequal": "4.5.0",
+    "husky": "1.3.x",
     "moment": "^2.24.0",
-    "ohif-core": "0.4.2",
+    "ohif-core": "0.5.0",
     "ohif-cornerstone-extension": "^0.0.28",
     "ohif-dicom-html-extension": "^0.0.2",
     "ohif-dicom-microscopy-extension": "^0.0.5",
     "ohif-dicom-pdf-extension": "^0.0.6",
     "ohif-vtk-extension": "^0.0.1",
```
dannyrb: Can we lock all pre-v1.0.0 dependencies to an exact version?
And all post-v1.0.0 dependencies to `Major.Minor.x` (where we'll only update to latest patch)?

dannyrb:
@biharck, assuming everything continues to work as normal, then this looks good to me. Anything jumping out as a cause of concern, @evren217 ?

package.json
```diff
@@ -60,46 +59,40 @@
...
     "cornerstone-wado-image-loader": "^2.2.3",
-    "dcmjs": "^0.3.8",
+    "dcmjs": "^0.4.7",
     "dicom-parser": "^1.8.3",
-    "dicomweb-client": "^0.4.2",
+    "eslint-config-react-app": "^4.0.0",
     "hammerjs": "^2.0.8",
-    "lodash.isequal": "4.5.0",
+    "husky": "1.3.x",
     "moment": "^2.24.0",
```
dannyrb: We can probably use a lighter weight dependency here, but probably best for a separate PR

package.json
```diff
@@ -60,46 +59,40 @@
...
     "ohif-cornerstone-extension": "^0.0.28",
     "ohif-dicom-html-extension": "^0.0.2",
     "ohif-dicom-microscopy-extension": "^0.0.5",
     "ohif-dicom-pdf-extension": "^0.0.6",
     "ohif-vtk-extension": "^0.0.1",
     "oidc-client": "^1.6.1",
     "prop-types": "^15.6.2",
-    "react-bootstrap-modal": "^4.2.0",
-    "react-dnd": "^7.0.2",
-    "react-dnd-html5-backend": "^7.0.2",
```
dannyrb: Nice. Kill them with fire 🔥

src/components/Labelling/LabellingFlow.js
```diff
@@ -192,7 +191,7 @@ export default class LabellingFlow extends Component {
   }
 
   descriptionCancel = () => {
-    const { description = '' } = cloneDeep(this.state)
+    const { description = '' } = JSON.parse(JSON.stringify(this.state))
```
evren217: I do not think this is a good solution for deep clone since it converts some data types like Date to string, so it is really not the exact clone.

dannyrb:
@biharck, can you address/discuss comments that were made so we can move forward?

src/components/Labelling/LabellingFlow.js
biharck: You're right, Moving back to cloneDeep

package.json
biharck: Agreed

swederik:
Closing because I made most of these changes in another PR.

PR #484, opened 2019-05-24
fix(labelling)

galelis:
- Update naming of configuration to props as cornerstoneTools has also changed


galelis:
I messed up my tests with local yalc packages. 
Now its all fixed. 
@swederik once you have time you can re-check :)

PR #495, opened 2019-05-28
fixing date format when sorting

biharck:
### What
When sorting for date at the home page, the second time was transforming the formatted date to invalid date.

### Why
It was failing the sorting for date because of invalid date


biharck:
It looks like that in order to use sort with map we need to Convert Map to an array using Array.from, sort array, convert back to Map. I decided to split the map from the sort and it is working properly now.

evren217:
Sorting works only for the current page since it is sorted in client. We will need to sort with qido request.

PR #499, opened 2019-05-28
Roadmap and icon updates using latest react-viewerbase

dannyrb:
Uses new `react-viewerbase` `<Icon />` component. Docs here: https://react.ohif.org/elements/icon

Removes external dependencies for `icons.svg` and `font-awesome`
Icons are inlined and can by async loaded. Also possible to animate and other goodies.

Icon support will be expanded further as part of upcoming extension changes/needs. There may be some outstanding issues, but the bulk of them should be resolved, and now addressable in one central location.

Note: `ohif-vtk-extension` was pulling in it's own (outdated) copy of `react-viewerbase`. I updated it to reference it as a peerDependency; copying the strategy used in `ohif-cornerstone-extension`. I released to a new org/scoped package: https://www.npmjs.com/package/@ohif/extension-vtk

Other Notes:

- Sizing may be slightly off where styles accounted for `svg`, but not an icon font.
- I found many different versions of `icons.svg`. If an icon does not look how you would expect, replace it's SVG code in `react-viewerbase`
- `font-awesome` helper classes are dead. We'll likely need to add copies of the basic ones for:
    - Animating our `notch-circle` icon w/ `Loading...`
    - Rotating our 3D cube for axial, sag, cor


extensions/ohif-cornerstone-extension/package.json
```diff
@@ -32,14 +32,14 @@
...
-    "react-viewerbase": "^0.2.10",
+    "react-viewerbase": "^0.6.0",
     "redux": "^4.0.1"
   },
   "dependencies": {
     "@babel/runtime": "^7.2.0",
     "classnames": "^2.2.6",
     "lodash.throttle": "^4.1.1",
-    "react-cornerstone-viewport": "0.1.29"
+    "react-cornerstone-viewport": "0.1.28"
```
swederik: Downgrading?

dannyrb: Not on purpose. My local version say this was the value that was last set:

image

Possible I blew it out when merging in changes from master. I'll re-bump

dannyrb: Marking this resolved. Bumped versions and cut new releases.

Issue #494, opened 2019-05-28
(empty) should show up in the Study List when Patient Name is missing


swederik:
This is how it appears in the Meteor version. It should do the same in the latest version.

![Screenshot 2019-05-28 14 33 18](https://user-images.githubusercontent.com/607793/58479338-600bc500-8158-11e9-9582-8507fe9114d1.png)

biharck:
Fixed at https://github.com/OHIF/Viewers/pull/503

biharck:
The changes were moved to react-viewwerbase project and it was fixed at https://github.com/OHIF/react-viewerbase/pull/86

PR #503, opened 2019-05-29
Default empty name

biharck:
### What
This PR is to fix the opened issue https://github.com/OHIF/Viewers/issues/494 where (empty) should be shown when patients name is not defined.

### Why
It is a more intuitive show to the user (empty) rather than a piece of blank information.


src/studylist/StudyListWithData.js
```diff
@@ -55,77 +55,82 @@ class StudyListWithData extends Component {
...
-            study.studyDate = moment(study.studyDate, 'YYYYMMDD').format(
-              'MMM DD, YYYY'
-            );
-            return study;
-          });
+            if (!moment(study.studyDate, 'MMM DD, YYYY', true).isValid()) {
+              study.studyDate = moment(study.studyDate, 'YYYYMMDD').format(
+                'MMM DD, YYYY'
+              )
+              study.patientName = !study.patientName
```
swederik: Why would this be inside the if statement?

biharck:
Since I changed the emptyName at react-viewerbase project, I am closing this PR and deleting the branch.

PR #520, opened 2019-06-05
WIP: Use react proxy instead of nginx with cors headers for local dev

diego0020:
Use the proxy server [included in react scripts](https://facebook.github.io/create-react-app/docs/proxying-api-requests-in-development) to avoid CORS issues in local development environment. The approach used before was setting up a nginx container which set CORS headers as proxy.


dannyrb:
@all-contributors, please add @diego0020 for some helpful code

Issue #489, opened 2019-05-27
Study Search Broken for Google Healthcare API Integration


dzelemba:
OHIF is appending * to all QIDO request parameters, but this isn't supported yet by the Google Cloud Healthcare API (https://cloud.google.com/healthcare/docs/dicom#search_transaction).

dannyrb:
@dzelemba, my guess would be that this is something we won't have the bandwidth to fix for `v1.x` (but would be willing to review/merge a pull request for). It is good to be aware of this as we add support in `v2.x`

dannyrb:
I believe this has been resolved with the latest healthcare adapter merge. There is also a new configuration property to enable fuzzy searching if the target DICOMWeb implementation supports it.

PR #523, opened 2019-06-05
WIP: Optionally add Google Cloud Healthcare Adapter

swederik:
Things are sort of working but this PR is pretty messy.

You can run this with `REACT_APP_CONFIG=config/google-cloud.js npm run dev` but you should change `google-cloud.js` to your own CLIENT_ID. Right now it's using mine.

### TODO
- [ ] Fix horrible UI issues which prevent this from actually being usable.
- [ ] Find a cleaner way to load the [GCP Healthcare API Adapter](https://github.com/quantumsoftgroup/healthcare-api-adapter). Storing them in public and using loadScript is probably not the best way to do this. Hoping for advice from @dannyrb on this...
- [ ] Move the WADO-URI to WADO-RS changes for ohif-cornerstone-extension out of this PR and do it separately in another PR. GCP does not support WADO-URI.
- [ ] Figure out how we want to use this to build a docker container for usage with GCP. Should we have a separate container like we did before with ohif/viewer-google-cloud? Or should we make the main container flexible enough to accept GCP CLIENT_ID and _just work_?


dannyrb:
May be something to keep in mind: https://github.com/OHIF/Viewers/issues/489

swederik:
Closing in favour of https://github.com/OHIF/Viewers/pull/584

Issue #505, opened 2019-05-30
Use nprogress when navigating routes


dannyrb:
Today, when you navigate between the study list and the viewer, we see a flash of `Loading` text until the route is ready to be rendered. You'll notice SPAs employing different techniques to smooth out this experience.

Our meteor version transitions the header to it's new height, throws up an overlay, and has some nicer looking text. I think this works well for that specific route transition, but something more generic may be useful if we intend to add more routes, or for navigation between studies (as they use the same view).

GitHub users will notice that most route navigation shows a "progress bar" at the top of the screen in blue. There are a few different modules that pair well with React Router to achieve a similar effect.

Package: https://www.npmjs.com/package/@tanem/react-nprogress
Demo: https://codesandbox.io/s/github/tanem/react-nprogress/tree/master/examples/react-router

PR #535, opened 2019-06-06
feat(OHIFStandaloneViewer.js): adding NProgress loading bar

biharck:
### Changes:
- Adds a thin progress bar to the top of the page when changing routes (react-nprogress)
- Cleans up the 404 Not Found page
- Adds a 'fade' transition class while loading the next route (react-transition-group)

### Notes:
- The Switch / Route logic is weird now, but we have no choice if we want to use react-transition-group with react-router. Apparently you have to do it this way (see http://reactcommunity.org/react-transition-group/with-react-router)/

Feel free to squash into a single commit with a reasonable commit message when merging.

Addresses #505


src/Container.js
```diff
@@ -0,0 +1,15 @@
...
+      opacity: isFinished ? 0 : 1,
+      pointerEvents: 'none',
+      transition: `opacity ${animationDuration}ms linear`,
+    }}
+  >
+    {children}
+  </div>
+);
+
+export default Container;
```
dannyrb: Can we move this file out of root and into the components folder? A more descriptive name may also be helpful.

src/NotFound.js
```diff
@@ -0,0 +1,16 @@
...
+    <div className={'not-found'}>
+      <div>
+        <h4>Sorry, this page does not exist.</h4>
+        <h5>
+          <Link to={'/'}>Go back to the Study List</Link>
+        </h5>
+      </div>
+    </div>
+  );
+}
```
dannyrb: I'm happy that this is being added, but maybe we co-locate components functioning as route targets to a directory other than source root?

src/OHIFStandaloneViewer.css
```diff
@@ -1,17 +1,33 @@
 html,
 body,
 #root {
```
dannyrb: Because we target UMD, and this may be embedded outside of an iFrame, we need to be careful about setting root, body, and html styles the could impact an app other than our own. Generally, we should try to use our app container as much as possible, with styles like this potentially being set in our index.html or an external top level css file.

I hope that makes sense. This is probably fine for this PR, but worth noting as it will be an issue as we clean up our embedded use case.

src/OHIFStandaloneViewer.js
```diff
@@ -79,24 +87,84 @@ class OHIFStandaloneViewer extends Component {
...
+        <NProgress isAnimating={this.state.isLoading}>
+          {({ isFinished, progress, animationDuration }) => (
+            <Container
+              isFinished={isFinished}
+              animationDuration={animationDuration}
+            >
+              <Bar progress={progress} animationDuration={animationDuration} />
+            </Container>
+          )}
+        </NProgress>
```
dannyrb: If we don't intend to reuse bar or container, it may make more sense to create a single component out of this and kill those components.

src/OHIFStandaloneViewer.js
```diff
@@ -79,24 +87,84 @@ class OHIFStandaloneViewer extends Component {
       );
     }
 
+    /**
+     * Note: this approach for routing is caused by the conflict between
+     * react-transition-group and react-router's <Switch> component.
```
dannyrb: Hmm... What exactly was the issue?

“Routing transition effect for React — with react-router v4 and react-transition-group” by Hyeokwoo Alex Kwon https://link.medium.com/2IU3mc8jRX

I honestly don't know enough to weigh in on the removal of `<switch>`, but it does look like we can keep it.

swederik: Yeah I did it with Switch like that initially and everything appeared to work, but the window resize handlers would start being called in infinite loops with short delays for no clear reason. 

The Cornerstone element started off super small and would grow and grow forever. I tried to debug it for a while and eventually gave up on that approach. This is what react-transition-group recommends in their docs (https://reactcommunity.org/react-transition-group/with-react-router) so that's good enough for me.

src/NotFound.js
swederik: Good point

PR #541, opened 2019-06-11
fix(IHEInvokeImageDisplay): Fix the props to access query parameters

dlwire:
I encountered this bug trying to use the IHEInvokeImageDisplay API in order to render multiple DICOM
studies in the viewer. Unfortunately, the component was no longer working and required some change in order to access the query parameters used in the API.

These changes update the component to use react router to access the query parameters and pass them to ConnectedViewerRetrieveStudyData as appropriate. The move from ViewerRetrieveStudyData to ConnectedViewerRetrieveStudyData was necessary in order to access the active server from the redux store. I also made a change from a series of if clauses to a switch statement as it seemed a little more appropriate given the nature of the code paths. Happy to change that back if needed. The PATIENT and default paths were not fixed as a part of this work because we'll need a way to pass the filter in to the StudyList component.


swederik:
Live example: https://deploy-preview-541--ohif.netlify.com/demo/IHEInvokeImageDisplay?requestType=STUDY&studyUID=1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5

dannyrb:
@all-contributors, please add @dlwire for some rad code

PR #543, opened 2019-06-11
feat(StudyList): Add QIDO style query parameters

dlwire:
This pull request adds the ability to use query parameters to pass filters into the StudyList from
the url. These parameters are of the same form used by the QIDO requests to apply the filters.

This solution seemed to me to be the most obvious way to manage them.

There's one open question:
What should it do with the query parameter filters after the initial load. For now, the URL parameters are overridden the moment the user tries to add another filter. Depending on review feedback it seems like it may make more sense for those query parameter filters to be "permanent" and so any further filtering would augment those parameters rather than replace them.


src/studylist/StudyListWithData.js
```diff
@@ -28,23 +28,25 @@ class StudyListWithData extends Component {
...
 
   componentDidMount() {
     // TODO: Avoid using timepoints here
     //const params = { studyInstanceUids, seriesInstanceUids, timepointId, timepointsFilter={} };
 
-    this.searchForStudies();
+    this.searchForStudies({
+      ...StudyListWithData.defaultSearchData,
+      ...(this.props.filters || {}),
+    });
```
dlwire: I mentioned this in the description but figured inline would make sure it gets noticed. I wonder if we should apply these filters that are passed in on every subsequent search as well. This behavior might feel a little strange.

Thoughts?

swederik: I think that makes sense. Let's go with that for now and see if users like it? I suppose there should be some way to clear the filters though.

swederik:
So this works:

https://deploy-preview-543--ohif.netlify.com/demo/?patientId=0000003
https://deploy-preview-543--ohif.netlify.com/demo/?PatientId=0000003
https://deploy-preview-543--ohif.netlify.com/demo/?accessionNumber=0000155*
https://deploy-preview-543--ohif.netlify.com/demo/?AccessionNumber=0000155*

but this doesn't

https://deploy-preview-543--ohif.netlify.com/demo/?00100021=0000003

How does the study date filter work? I I think it would be best if the formats matched what a QIDO search expects, but I am also happy to merge this and then continue to improve it in that direction.

Thanks!

src/studylist/StudyListWithData.js
dlwire: Yeah, I guess there are two main ways to work on that:
* Pre-populate the column filters with those in the query parameters
This solves the problem where we want to search the result set, but would allow them to get out of sync with the url.
* Always filter based on what's in the query parameters but provide a new button to clear them
This is probably best long term solution since it keeps the results in sync with the URL but also allows us to clear the query parameters and start fresh.

swederik:
Let's keep the URL filtering it the way it is for now, since it's good enough for a first iteration. They can get out of sync, but that's not a big deal.

The only change I would like to see is adding the QIDO-RS study date filtering format: 

`http://dicomrs/studies​?PatientID=11235813​&StudyDate=20130509-20130510`

That way we can just point at that documentation. That would be better than ?studyDateTo=awdd&studyDateFrom=$#342.

Thanks!

swederik:
I'm just going to merge this and we can improve it later.

Thanks!

PR #549, opened 2019-06-12
Feature/extensions panels and docs

dannyrb:
## WIP

- Adds barebones example extension
- When in development mode, `redux-logger` is enabled by default:

image

### @ohif/extension-cornerstone

- Cornerstone commands in `Viewers` were moved to a local `CommandsModule`
- Removes `wrappedComponent` that would merge/pass in props and children. Idea was to pass in children that would render on top of the base components. I think we can accomplish this without the wrapper.

_`ConnectedCornerstoneViewport`_

There's nothing wrong with connecting, _per se_, but it would be nice if we could just expose a consistent API here, and allow the wiring to take place in `Viewers`. It forces us to be a bit more intentional/consistent, and it allows us to change our store/actions without blowing up every extension. Even if we choose to not use redux _at all_.

Props:
```
    layout: state.viewports.layout,
    isActive,
    activeTool: activeButton && activeButton.command,
    ...dataFromStore,
    enableStackPrefetch: isActive,
    //stack: viewportSpecificData.stack,
    cineToolData: viewportSpecificData.cine,
    viewport: viewportSpecificData.viewport
```

Actions:
```
setViewportActive,
setViewportSpecificData,
clearViewportSpecificData,
onElementEnabled, // calls setViewportSpecificData to register `dom`
onMeasurementsChanged
```

- Need to look closer at `ConnectedCineDialog`, `ConnectedCornerstoneViewport` and `ConnectedToolbarSection`

### TODO

- Toolbar to a combination of config + custom components
- Delete unused connected components from extensions
    - More consistent API across modules that are components

### Notes

- Consistent API for setting active tool for a given context
- Viewport::SupportsCinePlayer??


dannyrb:
https://github.com/OHIF/ohif-core/commit/9d7a997

PR #557, opened 2019-06-15
feat(@ohif/extension-vtk): Update app to include basic MPR with VTK

swederik:


jacklau88:
nice job.

pieper:
Looking great!

swederik:
There are about a million issues remaining with it. I'll be creating a big thread for them so we can slowly chip away.

pieper:
Yeah, I noticed an issue or two (or a maybe a million : )  But still it's a big step in the right direction.

PR #568, opened 2019-06-17
@ohif/extension-vtk: Enable crosshairs display on viewports (0.0.6)

swederik:


swederik:
Special thanks to @jourdain for the SVG handling with VTK.js. This builds on an open PR: https://github.com/Kitware/vtk-js/pull/884

Hopefully we can push it back to VTK soon.

Issue #550, opened 2019-06-13
Microscopy extension viewport resizing incorrectly


swederik:
It looks like the @ohif/extension-microscopy Viewport is resizing incorrectly when the sidebars are changed. It appears to be resizing in CSS but not calling a redraw of the rendering on the Canvas. Note how the actual cell aspect ratios are being changed. This is not what should be happening.

![MicroscopyStretch](https://user-images.githubusercontent.com/607793/59427889-0c49de80-8ddc-11e9-863e-c24f2ce92371.gif)

biharck:
There are two PRs to fix that.
https://github.com/dcmjs-org/dicom-microscopy-viewer/pull/30
https://github.com/OHIF/Viewers/pull/571

biharck:
waiting for a new version from Microscopy in order to close this issue.

Malik-Adil:
> It looks like the @ohif/extension-microscopy Viewport is resizing incorrectly when the sidebars are changed. It appears to be resizing in CSS but not calling a redraw of the rendering on the Canvas. Note how the actual cell aspect ratios are being changed. This is not what should be happening.
> 
> ![MicroscopyStretch](https://user-images.githubusercontent.com/607793/59427889-0c49de80-8ddc-11e9-863e-c24f2ce92371.gif)

Can you send me dicom file? I want to check the metadata of this file.

swederik:
Hi @Malik-Adil,

The data we are using comes from here: ftp://medical.nema.org/medical/dicom/Datasets/WG26/

It was part of the IHE Connectathon: http://dicom.nema.org/Dicom/DICOMWSI/

Hope that helps!

PR #571, opened 2019-06-19
Fixing issue #550

biharck:
fix(DicomMicroscopyViewport): adding updateSize call when the viewport gets resized

### DIFF
--- a/extensions/ohif-dicom-microscopy-extension/src/DicomMicroscopyViewport.js
+++ b/extensions/ohif-dicom-microscopy-extension/src/DicomMicroscopyViewport.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
-
+import ReactResizeDetector from 'react-resize-detector';
 import { api } from 'dicom-microscopy-viewer';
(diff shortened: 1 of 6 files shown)


biharck:
**Please, don't merge it until we get this PR https://github.com/dcmjs-org/dicom-microscopy-viewer/pull/30 approved and the new version updated at the microscopy extension.**

extensions/ohif-dicom-microscopy-extension/src/DicomMicroscopyViewport.js
```diff
@@ -8,10 +10,17 @@ class DicomMicroscopyViewport extends Component {
...
 
+  viewer = null;
+
   constructor(props) {
     super(props);
 
     this.container = React.createRef();
+
+    this.debouncedResize = debounce(() => {
+      debugger;
```
swederik: Missed a debugger

biharck: done

Issue #528, opened 2019-06-06
Add E2E Cypress smoke tests as a final check before deploying viewer.ohif.org


dannyrb:
E2E tests are slow, but a great way to make sure that major features still work as expected. Cypress makes adding them pretty painless, supports testing different browser vendors, can record testing, take screenshots of error states, and is free for open source.

Adding one or two very small tests as a last line of defense for our, now automated, deploy to viewer.ohif.org would be a good place to start:

- [ ] StudyList route loads, w/ expected data in first 2 rows
- [ ] Viewer route loads w/ expected data in first 2 thumbnails

Resources:

- https://github.com/bahmutov/cypress-react-unit-test
- https://blog.bitsrc.io/testing-react-apps-with-cypress-658bc482678

PR #573, opened 2019-06-19
Add e2e tests with cypress

jfmedeiros1820:
Issue [here](https://github.com/OHIF/Viewers/issues/528)


swederik:
Can you add some documentation somewhere about how this works, what it does, why we want it, etc?

dannyrb:
This can be cleaned up a little more, but I'm going to merge to test a few things.

dannyrb:
@all-contributors please add @jfmedeiros1820 for tests

PR #575, opened 2019-06-19
Replace use of context to import locales

### LINKED ISSUE #565: i18n breaks embedded viewer
Release 0.2.0 added i18n support, but this broke the unpkg script tag use of ohif.

You can replicate this in the code sandox linked from the docs:

https://codesandbox.io/s/ohif-viewer-script-tag-usage-svmm2

change line 98 from:
   ` src="https://unpkg.com/ohif-viewer@0.0.9/dist/index.umd.js"`
to
   `src="https://unpkg.com/ohif-viewer@0.2.0/dist/index.umd.js"`

and the study list no longer loads (any version between 0.0.9 and 0.1.19 works).

code sandbox doesn't report anything specific, but when I test locally I get this message:

```
index.es.js:104 Uncaught ReferenceError: require is not defined
    at getLocales (index.es.js:104)
    at initI18n (index.es.js:123)
    at index.es.js:142
    at index.umd.js:4
    at index.umd.js:5
```

rodrigolabs:
Fixes #565

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,10 @@
     "not ie <= 11",
     "not op_mini all"
   ],
+  "resolutions": {
+    "browserslist": "4.6.2",
+    "caniuse-lite": "1.0.30000974"
+  },
   "files": [
     "dist"
   ],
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -96,7 +98,7 @@
     "react-resize-detector": "^4.2.0",
     "react-router": "^5.0.1",
     "react-router-dom": "^5.0.1",
-    "react-viewerbase": "0.10.0",
+    "react-viewerbase": "file:.yalc/react-viewerbase",
```
swederik: Leftover yalc..

rodrigolabs: @swederik it's already  updated, check the newer version:

https://github.com/OHIF/Viewers/blob/a25f5bfa6381c836f9d4cca5c2a4738248fcb81d/package.json#L101

PR #588, opened 2019-06-24
feat(tools): Add additional tools

evren217:
Regarding to [this feature request](https://ohif.canny.io/feature-requests/p/update-toolbar-to-match-v1s-default-setup), here is the tools to add for cornerstone viewport:

- [x] ROI Window
- [x] Probe
- [x] Ellipse
- [x] Rectangle
- [x] Invert
- [x] Rotate Right
- [x] Flip H
- [x] Flip V
- [x] Clear

## Moved until the next PR:
- [ ] Annotate
- [ ] Crosshairs
- [ ] Previous Display Set
- [ ] Next Display Set
- [ ] Play / Stop
- [ ] [Download](https://ohif.canny.io/feature-requests/p/download-screenshot-menu) (might be a separate issue/PR)

### DIFF
--- a/extensions/ohif-cornerstone-extension/src/commandsModule.js
+++ b/extensions/ohif-cornerstone-extension/src/commandsModule.js
@@ -1,5 +1,6 @@
 import cornerstone from 'cornerstone-core';
 import cornerstoneTools from 'cornerstone-tools';
+import OHIF from 'ohif-core';
 
 const actions = {
   rotateViewport: ({ viewports, rotation }) => {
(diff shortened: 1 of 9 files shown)


extensions/ohif-cornerstone-extension/src/commandsModule.js
```diff
@@ -89,15 +90,60 @@ const actions = {
...
+          toolType,
+          _id,
+          lesionNamingNumber,
+          measurementNumber,
+        });
+      });
+    });
+
+    measurementsToRemove.forEach(measurementData => {
+      OHIF.measurements.MeasurementHandlers.onRemoved({
```
dannyrb: Does the `MeasurementHandlers` not already subscribe to cornerstone events? Why do we need to call this manually?

evren217: It somehow does not handle the removed event when a measurement is removed using cornerstone tool state directly, so measurements are not removed from the measurement table otherwise. Either, cornerstone does not trigger removed event or that does not handle the event, not sure. Feel free to investigate it and just update this code to remove measurements using cornerstone tool state here.

swederik: Yeah It does not trigger the event if you remove it from the toolState programmatically. You don't need an event if you just called the function. The events are triggered by the tool changes.

PR #594, opened 2019-06-26
ci(rollbar): Add Rollbar script during CI deploy

swederik:


dannyrb:
This is a weird hoop to jump threw. If we find ourselves slowly accruing more and more workarounds like this, we may want to consider a custom WebPack config (we already have our rollup one). CRA uses `html-webpack-plugin` under the hood. It's a pretty powerful option to accomplish what you're doing here in a cleaner way:

https://www.npmjs.com/package/html-webpack-plugin

There are also the common "rewired" variants of `react-scripts` if we prefer something that's maintained, but that we'd like a little more control over extending/configuring.

swederik:
Yeah this is a short term solution. I just wanted to make [Rollbar](https://rollbar.com/) work on `viewer.ohif.org` as easily as possible without impacting the rest of the application. It's only in `build:web:ci` for testing purposes, I'll remove those npm scripts before merging, so all that's left is the injection into the index.html prior to deployment to the google cloud bucket.

Long term we would want some configurable error handler setup so end users could use e.g. [Sentry ](https://sentry.io/welcome/) instead.

PR #597, opened 2019-06-27
feat(StudyList): Disable "study list" link using a config object

### LINKED ISSUE #574: Disable "study list" link using a config object
In many applications users only want to open the viewer with a particular study. For this cases it would be useful to disable the `study list` link.  This could be specially useful in the embedded view.

jfmedeiros1820:
Issue [here](https://github.com/OHIF/Viewers/issues/574)

Closes #574


src/config/index.js
```diff
@@ -0,0 +1,5 @@
+const CONFIG = {
+  SHOW_STUDY_LIST: process.env.SHOW_STUDY_LIST || true,
```
dannyrb: Won't this always be true?

`undefined || true` --> true
`false || true` --> true

dannyrb:
I was originally thinking this would be runtime, not build time. We have our configuration that's set on `window.config` by the config files in our `public/config` directory. You could set it and then simply hide the `study list` button in the `viewer`  route.

This would primarily be used for existing apps that have an existing concept of a patient/study list, and want to link or iFrame directly to the Viewer w/ the study loaded.

Your implementation does add some value though. Only shipping the necessary code decreases the bundle size and simplifies some things. However, I see a few issues with this PR as is, and I think it might require more effort than it's worth at this time to provide a build flag that accomplishes this well.

src/config/index.js
jfmedeiros1820: You are right, this doesn't work with booleans. I will fix using the runtime approach

README.md
```diff
@@ -119,10 +119,13 @@ Commands to run the tests:
 
 ```js
 // Open Cypress Dashboard that provides insight into what happened when your tests ran
-yarn run cypress
+yarn run cy
```
swederik: Why are these changes in this PR?

jfmedeiros1820: While testing the impacts of this change, I saw that we changed that commands in `package.json` but didn't update the README file. Since I was already changing the code here, I thought it would be ok to fix the documentation

swederik: It is a totally unrelated change, so it should be made in a separate commit and separate PR, with a specific commit message.

jfmedeiros1820: Okay, I will remove from here and create another PR for this

jfmedeiros1820: Done! Created the PR https://github.com/OHIF/Viewers/pull/610 to handle this changes

src/connectedComponents/ViewerMain.js
```diff
@@ -139,9 +139,9 @@ class ViewerMain extends Component {
       <div className="ViewerMain">
         {this.state.displaySets.length && (
           <ConnectedLayoutManager
-          studies={this.props.studies}
-          viewportData={this.getViewportData()}
-          setViewportData={this.setViewportData}
+            studies={this.props.studies}
```
swederik: Why are these whitespace changes in this PR?

jfmedeiros1820: Sorry, the prettier plugin was enabled in my IDE and fixing style problems when saving the files. I already disabled but I forgot to reset the changes of this specific file. I will reset the code here.

swederik:
One issue left to fix before we can merge it. Show study list should be true by default. Right now even the deploy preview doesn't work. See https://deploy-preview-597--ohif.netlify.com/demo/

jfmedeiros1820:
> One issue left to fix before we can merge it. Show study list should be true by default. Right now even the deploy preview doesn't work. See https://deploy-preview-597--ohif.netlify.com/demo/

PR #598, opened 2019-06-27
fix(@ohif/i18n): Tag script implementation

rodrigolabs:
This following export in `@ohif/i18n` is not working when using the Viewer tag script implementation: 

https://github.com/OHIF/Viewers/blob/cb565427b3d193b85072751a92b61a2bb9171a4b/extensions/ohif-i18n/src/index.js#L56

It's breaking and displaying the error `Object(...) is not a function`.

## Before merging this PR, we need to publish:

1 - `react-viewerbase`. — to fix the require issue in `react-dnd`. -- https://github.com/OHIF/react-viewerbase/pull/117

2 - `@ohif/extension-vtk` to update `react-viewerbase` version published above.

3 - `OHIF/Viewers` to update `react-viewerbase`, `@ohif/extension-vtk` and `@ohif/i18n`.

### DIFF
--- a/docs/latest/essentials/translating.md
+++ b/docs/latest/essentials/translating.md
@@ -115,9 +115,9 @@ When needed, you can also use available translations _without React_.
 E.g.
 
 ```js
-import { t } from '@ohif/i18n';
-console.log(t('my translated text'));
-console.log(t('$t(Common:Play) my translated text'));
+import { T } from '@ohif/i18n';
+console.log(T('my translated text'));
+console.log(T('$t(Common:Play) my translated text'));
 ```
 
 ---
--- a/extensions/ohif-i18n/src/index.js
+++ b/extensions/ohif-i18n/src/index.js
@@ -12,10 +12,14 @@ function addLocales(newLocales) {
   let resourceBundle = [];
 
   for (const key in newLocales) {
-    for (const namespace in newLocales[key]) {
-      const locale = newLocales[key][namespace];
-      resourceBundle.push({ key, namespace, locale });
-      i18n.addResourceBundle(key, namespace, locale, true, true);
+    if (newLocales.hasOwnProperty(key)) {
+      for (const namespace in newLocales[key]) {
+        if (newLocales[key].hasOwnProperty(namespace)) {
+          const locale = newLocales[key][namespace];
+          resourceBundle.push({ key, namespace, locale });
+          i18n.addResourceBundle(key, namespace, locale, true, true);
+        }
+      }
     }
   }
 
--- a/extensions/ohif-i18n/src/index.js
+++ b/extensions/ohif-i18n/src/index.js
@@ -6,20 +6,18 @@ import pkg from '../package.json';
 import { debugMode, detectionOptions } from './config';
 import locales from './locales';
 
-let translate;
-
 function addLocales(newLocales) {
   customDebug(`Adding locales ${newLocales}`, 'info');
 
   let resourceBundle = [];
 
-  Object.keys(newLocales).map(key => {
-    Object.keys(newLocales[key]).map(namespace => {
+  for (const key in newLocales) {
+    for (const namespace in newLocales[key]) {
       const locale = newLocales[key][namespace];
       resourceBundle.push({ key, namespace, locale });
       i18n.addResourceBundle(key, namespace, locale, true, true);
-    });
-  });
+    }
+  }
 
   customDebug(`Locales added successfully`, 'info');
   customDebug(resourceBundle, 'info');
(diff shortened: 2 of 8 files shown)


extensions/ohif-vtk-extension/package.json
```diff
@@ -39,7 +39,10 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.4.5",
+    "css-loader": "^3.0.0",
```
rodrigolabs: I had to add these three modules to run the project locally.
Is it really necessary to be added here?

swederik: Those should be devDependencies, not dependencies. You can add them if they are necessary, but I don't know why this error would not have shown up earlier.

extensions/ohif-vtk-extension/package.json
```diff
@@ -39,7 +39,10 @@
...
+    "postcss-loader": "^3.0.0",
     "react-vtkjs-viewport": "0.0.9",
+    "style-loader": "^0.23.1",
```
swederik: Do we really need three different style related loaders??

docs/latest/essentials/translating.md
```diff
@@ -115,9 +115,9 @@ When needed, you can also use available translations _without React_.
 E.g.
 
 ```js
-import { t } from '@ohif/i18n';
-console.log(t('my translated text'));
-console.log(t('$t(Common:Play) my translated text'));
+import { T } from '@ohif/i18n';
```
swederik: Why?

extensions/ohif-i18n/src/index.js
```diff
@@ -6,20 +6,22 @@ import pkg from '../package.json';
...
   let resourceBundle = [];
 
-  Object.keys(newLocales).map(key => {
-    Object.keys(newLocales[key]).map(namespace => {
-      const locale = newLocales[key][namespace];
-      resourceBundle.push({ key, namespace, locale });
-      i18n.addResourceBundle(key, namespace, locale, true, true);
-    });
-  });
+  for (const key in newLocales) {
```
swederik: I don't understand what is different between the Object.keys() loop and the `for in` loop. Is there a reason you did this?

yarn.lock
```diff
@@ -1253,7 +1258,16 @@
...
+"@ohif/i18n@0.1.1":
+  version "0.1.1"
+  resolved "https://registry.yarnpkg.com/@ohif/i18n/-/i18n-0.1.1.tgz#e4eee433a0a5fb15a41279910ad061e98a371c4c"
+  integrity sha512-ON1UzWDyi2RD79Z+632vvOh82tl6jglzLHZ1M2rQEm5FLP1PQyrwMEBjETA6jMx7mE1XIWQNiGpbln+7AaPSgg==
+  dependencies:
+    "@babel/runtime" "^7.2.0"
+    classnames "^2.2.6"
+    rollup-plugin-json "^4.0.0"
+
+"@ohif/i18n@^0.1.0":
```
swederik: What is still pulling in `@ohif/i18n@^0.1.0`? `react-viewerbase`?

docs/latest/essentials/translating.md
rodrigolabs: When I had put `t` in the `i18n` object to be exported, I got `Maximum call stack size exceeded.` -- maybe some conflict with the standard `t`.

I just changed to `T` and the error has gone. Just to be clear, this `T` is related to the use of the lib without react: https://docs.ohif.org/essentials/translating.html#without-react
It’s not the same `t` we are getting with HOCs.

extensions/ohif-i18n/src/index.js
rodrigolabs: I was just trying to test changing to `for ... in` to see if the error was here - it was not - and I ended up leaving it like this. I'm reverting this.

extensions/ohif-vtk-extension/package.json
rodrigolabs: I've updated my `yarn` and suddenly it worked without these dependencies! :)
I'll remove them.

yarn.lock
rodrigolabs: It's not that clear, but I guess it's coming from `react-viewerbase`, since it was not published yet using `@ohif/i18n` `0.1.1`.

extensions/ohif-vtk-extension/package.json
rodrigolabs: I was taking a look and they are being used in this webpack implementation here:

https://github.com/OHIF/Viewers/blob/cb565427b3d193b85072751a92b61a2bb9171a4b/extensions/ohif-vtk-extension/webpack.config.js#L6-L20

What do you think?

Issue #621, opened 2019-07-02
End-to-end tests for Script Tag usage with Cypress


swederik:
We need to add an end-to-end test to verify that the ohif-viewer can be used as a script tag. The test should clone the same stuff that is in the CodeSandbox (https://codesandbox.io/s/ohif-viewer-script-tag-usage-b3st9) and should verify that things load properly.

PR #633, opened 2019-07-08
feat(E2E): End-to-end tests for script tag

jfmedeiros1820:
Include example application that uses Viewer as script tag and includes configuration to run the rollup bundle after package.
fix [621](https://github.com/OHIF/Viewers/issues/621)


cypress/support/script-tag/index.html
```diff
@@ -0,0 +1,82 @@
...
+<body>
+  <noscript> You need to enable JavaScript to run this app. </noscript>
+
+  <div id="root"></div>
+
+  <!-- Load React. -->
+  <script crossorigin src="https://unpkg.com/react@16/umd/react.production.min.js"></script>
+  <script crossorigin src="https://unpkg.com/react-dom@16/umd/react-dom.production.min.js"></script>
+  <script>
+    // Workaround to deal with react-router-dom complaining
```
swederik: Could you double check if this section is still necessary?

cypress/support/script-tag/index.html
```diff
@@ -0,0 +1,82 @@
...
+        dicomWeb: [
+          {
+            name: 'DCM4CHEE',
+            wadoUriRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/wado',
+            qidoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
+            wadoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
+            qidoSupportsIncludeField: true,
+            imageRendering: 'wadors',
+            thumbnailRendering: 'wadors',
+            requestOptions: {
```
swederik: This can be removed, `requestOptions.requestFromBrowser` is no longer used.

cypress/support/script-tag/index.html
```diff
@@ -0,0 +1,82 @@
...
+  <meta name="MobileOptimized" content="320" />
+  <meta name="HandheldFriendly" content="True" />
+  <meta name="apple-mobile-web-app-capable" content="yes" />
+
+  <!-- Latest compiled and minified CSS -->
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css"
+    integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous" />
+
+  <!-- WEB FONTS -->
+  <link href="https://fonts.googleapis.com/css?family=Sanchez" rel="stylesheet" />
```
swederik: @dannyrb Do we still need Sanchez and font-awesome?

dannyrb: We may still use Sanchez for logo font, but it should he replaced with an inlined SVG React Component. FA can be nuked.

cypress/support/script-tag/index.html
```diff
@@ -0,0 +1,82 @@
...
+  <meta name="viewport"
+    content="width=device-width,initial-scale=1.0,minimum-scale=1,maximum-scale=1,user-scalable=no" />
+  <meta name="theme-color" content="#000000" />
+  <meta http-equiv="cleartype" content="on" />
+  <meta name="MobileOptimized" content="320" />
+  <meta name="HandheldFriendly" content="True" />
+  <meta name="apple-mobile-web-app-capable" content="yes" />
+
+  <!-- Latest compiled and minified CSS -->
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css"
```
swederik: I think we removed the dependency on this, but we should check.

swederik:
Can we pull in the WADO image loader stuff another way? Right now they will never be updated. We currently copy them from node_modules. (we should really move them to use worker-loader)

cypress/support/script-tag/index.html
jfmedeiros1820: It looks like, it's still necessary. When I removed this workaround, the application started to break. `Uncaught ReferenceError: process is not defined`

cypress/support/script-tag/index.html
jfmedeiros1820: Yeah, the logo was the only place I notice it changed after I remove the font link. Since we are not testing and considering the font for E2E, I will remove it.

cypress/support/script-tag/index.html
swederik: OK thanks

cypress/support/script-tag/index.html
jfmedeiros1820: We removed this dependency in `react-viewerbase` so far. The `Viewers` still uses some bootstrap classes and when I remove the import link, I notice some changes in the layout but no breaks. Should I remove then?

dannyrb: Ah shoot. It would be nice if we can resolve this in a separate PR. I would like our bootstrap dependency to be dead.

cypress/support/script-tag/index.html
jfmedeiros1820: Done

cypress/support/script-tag/index.html
jfmedeiros1820: Done

PR #634, opened 2019-07-09
I18n complement

biharck:
Adding missing i18n properties. This PR depends on react-viewerbase https://github.com/OHIF/react-viewerbase/pull/128


biharck:
Changes done @swederik

extensions/ohif-i18n/src/locales/en-US/index.js
```diff
@@ -3,7 +3,6 @@ import CineDialog from './CineDialog.json';
 import Common from './Common.json';
 import Header from './Header.json';
 import MeasurementTable from './MeasurementTable.json';
-import UserPreferencesModal from './UserPreferencesModal.json';
```
swederik: There shouldn't be any changes to the JSON or index.js files.

src/connectedComponents/ToolbarRow.js
```diff
@@ -96,6 +97,32 @@ class ToolbarRow extends Component {
...
+        icon: 'list',
+        bottomLabel: t('Measurements'),
+      },
+    ];
+
+    const leftSidebarValue = this.props.leftSidebarOpen
+      ? leftSidebarToggle[0].value
+      : null;
+
+    const rightSidebarValue = this.props.rightSidebarOpen
```
swederik: I don't see where these variables are being used at all?

extensions/ohif-i18n/src/locales/en-US/index.js
biharck: done

src/connectedComponents/ToolbarRow.js
biharck: It was after I merge with the latest master changes. I reverted already

PR #642, opened 2019-07-10
docs(google-cloud): Update the documentation

swederik:
docs(google-cloud): Update the documentation to explain how to use the Google Cloud Healthcare API

### Notes:
- Using docker cp to update the config is not as user-friendly as I would like. I tried using bind mounts but it didn't seem to replace the default.js file in nginx properly. If anyone has an easier suggestion, feel free to comment / make a PR.


swederik:
After discussing with @pieper we will switch this back to an environment variable which changes the config, it's just more user friendly that way.

PR #666, opened 2019-07-15
Add the Eraser to the available tools

dlwire:
Adding the Eraser tool to the list of supported tools.

We could also add this to the toolbar easily enough but here we go for now.

https://github.com/trustsitka/sitka/issues/3403

### DIFF
--- a/src/setupTools.js
+++ b/src/setupTools.js
@@ -217,6 +217,15 @@ export default function setupTools(store) {
     { name: 'ZoomTouchPinch' },
     { name: 'StackScrollMouseWheel' },
     { name: 'StackScrollMultiTouch' },
+    {
+      name: 'Eraser',
+      mouseButtonMasks: [1],
+      props: {
+        configuration: {
+          getMeasurementLocationCallback: toolLabellingFlowCallback,
+        },
+      },
+    },
   ];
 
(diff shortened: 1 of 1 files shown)


src/setupTools.js
```diff
@@ -217,6 +217,15 @@ export default function setupTools(store) {
     { name: 'ZoomTouchPinch' },
     { name: 'StackScrollMouseWheel' },
     { name: 'StackScrollMultiTouch' },
+    {
+      name: 'Eraser',
+      mouseButtonMasks: [1],
+      props: {
```
swederik: I don't think the Eraser tool should have these options

dlwire: You're absolutely right. Looks like I added this while I was trying to resolve a different issue that turned out to be caused by something in toolbarModule.js, which isn't included in this review. I'll get a fix.

Thanks Erik!

dlwire: I also removed the mouseButtonMasks this morning. Turns out that wasn't need either.

ranasrule:
its still not visible in the toolbar. am I missing something?

swederik:
Yes it hasn't been added to the toolbarModule: https://github.com/OHIF/Viewers/blob/master/extensions/ohif-cornerstone-extension/src/toolbarModule.js#L28

PR #667, opened 2019-07-15
add ArrowAnnotate as an available tool

dlwire:
Adding the ArrowAnnotate tool to the list of supported tools.

We could also add this to the toolbar easily enough but here we go for now.

https://github.com/trustsitka/sitka/issues/3403

### DIFF
--- a/src/setupTools.js
+++ b/src/setupTools.js
@@ -219,7 +219,6 @@ export default function setupTools(store) {
     { name: 'StackScrollMultiTouch' },
     {
       name: 'ArrowAnnotate',
-      mouseButtonMasks: [1],
       props: {
         configuration: {
           getMeasurementLocationCallback: toolLabellingFlowCallback,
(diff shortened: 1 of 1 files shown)


src/setupTools.js
```diff
@@ -217,6 +217,15 @@ export default function setupTools(store) {
     { name: 'ZoomTouchPinch' },
     { name: 'StackScrollMouseWheel' },
     { name: 'StackScrollMultiTouch' },
+    {
+      name: 'ArrowAnnotate',
+      mouseButtonMasks: [1],
```
dannyrb: If you remove this line, I'll merge this PR.
Specifying `mouseButtonMasks` is supposed to be used to set the initial active/bound tools for the viewport.
I get that the other config doesn't follow this pattern, just calling it out here.

dlwire: Yep, there was a lot of cargo culting here. Thanks Danny!

PR #668, opened 2019-07-15
fix(ToolbarRow): Update ToolbarRow to show active submenu buttons

dlwire:
Use the activeButtons list to see if there are any active buttons in submenus. If there are, pass one in as the activeCommand so it gets highlighted.

Related to sitka https://github.com/trustsitka/sitka/issues/3403


swederik:
Looks good to me! Thanks!

Issue #585, opened 2019-06-24
Google Healthcare API docker image update and MPR?


Stormert:
Hi OHIF team

Will the **ohif/viewer-google-cloud** docker image receive any more updates? 
Alternatively is there a way to use the **ohif/viewer** image with Google Healthcare API and are there any guides for how? 

I am particularly interested in getting MPR working so it this a possibility with Healthcare API as the backend?

swederik:
Not yet, but we are actively working on this. I'll post an update when it's possible.

Stormert:
Will this include MPR when the Google Healthcare API support is baked in with **ohif-viewer** image?

swederik:
Yep!

There are two ongoing PRs actually (https://github.com/OHIF/Viewers/pull/584 and https://github.com/OHIF/Viewers/pull/523). #523 is now superseded by #584 and we will work with @pavertomato to get it working ASAP.

swederik:
This should be working now, please give it a try. Docs are here: https://docs.ohif.org/connecting-to-image-archives/google-cloud-healthcare.html

Make sure you pull the latest docker image `docker pull ohif/viewer:latest`

Then you can run it with e.g.:

`docker run --env CLIENT_ID=723928408739-k9k9r3i44j32rhu69vlnibipmmk9i57p.apps.googleusercontent.com --publish 5000:80 ohif/viewer:latest`

Don't hesitate to reopen if you have any issues.

Stormert:
Hi, i have been trying to get this up and running in a container optimized GCP Compute engine, similarly to how i tested the ohif/viewer-google-cloud but i get this error message.

Error: redirect_uri_mismatch

The redirect URI in the request, http://localhost:5000/callback, does not match the ones authorized for the OAuth client. To update the authorized redirect URIs, visit: https://console.developers.google.com/apis/credentials/oauthclient/**MY_CLIENT_ID**.apps.googleusercontent.com?project=**MY_GCP_PROJECT**

It seems this docker image is redirecting to localhost rather than what is specified in the CLIENT_ID. The only thing i have changed is the container image from **ohif/viewer-google-cloud:latest** to **ohif/viewer:latest** and the redirrect worked perfectly with the google cloud image. Is there other changes that i need to make in order to get this working? I could not spot any in the documentation.

swederik:
[...]

PR #675, opened 2019-07-17
fix of hardcoded url

pavertomato:
[#585]

### DIFF
--- a/public/config/google.js
+++ b/public/config/google.js
@@ -13,7 +13,7 @@ window.config = {
       // Authorization Server URL
       authority: 'https://accounts.google.com',
       client_id: 'YOURCLIENTID.apps.googleusercontent.com',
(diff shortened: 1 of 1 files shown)


dannyrb:
This seems pretty straightforward, and I can see the value. Thanks for the PR -- I'll merge after these tests run.

pavertomato:
> This seems pretty straightforward, and I can see the value. Thanks for the PR -- I'll merge after these tests run.

@swederik had some doubts about it, i think it's better to wait for his opinion before merge

swederik:
Yeah I was just worried that this would cause issues if you arrived at e.g. viewer.ohif.org/studylist or something. This is better than what we have now though so I'll merge it and we can go from there. Thanks!

swederik:
TODO: switch to `${window.location.protocol}//${window.location.host}`

dannyrb:
@all-contributors, please add @pavertomato for this code

PR #676, opened 2019-07-17
straw man for cancelling the active draw

dlwire:
Just wanted to start some discussion here. This is an action and command that cancel the active draw.  I've done some testing and noticed a couple of things:
* Deleting all the active data doesn't stop the draw as mentioned in slack - this doesn't seem to cause any exceptions, just seems to swallow the next mouse click and then start drawing again
* This also deletes the data from the tool under the mouse on hover - apparently those data points are marked as active as well

That said, it seems like both `clearAnnotations` and `cancelActiveDraw` deal heavily in the details of `cornerstoneTools`. I'd like to suggest that we push both of those in to `cornerstoneTools`.

It'll be really helpful for us to have support for cancelling the active draw so it is probable something we'll be able to do and contribute back if there's interest.

Perhaps something like the `clearToolState` here: https://github.com/cornerstonejs/cornerstoneTools/blob/7d67e2a944c8fcd29eae58fdb6935eeaff8516f2/src/stateManagement/toolState.js#L116

https://github.com/trustsitka/sitka/issues/3417

### DIFF
--- a/extensions/ohif-cornerstone-extension/src/commandsModule.js
+++ b/extensions/ohif-cornerstone-extension/src/commandsModule.js
@@ -91,59 +91,30 @@ const actions = {
     console.warn('updateDisplaySet: ', direction);
   },
   clearAnnotations: ({ viewports }) => {
-    const element = _getActiveViewportEnabledElement(
-      viewports.viewportSpecificData,
-      viewports.activeViewportIndex
-    );
-    if (!element) {
-      return;
-    }
+    const enabledElement = _getEnabledElement(viewports);
+    const enabledElementToolState = _getElementToolState(enabledElement);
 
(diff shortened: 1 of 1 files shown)


dlwire:
Turns out this works alright for *most* of the tools. The freehand drawing tool really shits the bed when we cancel this way. I'm going to look into adding something to do this to cornerstoneTools today.

dlwire:
I figured out how to cleanly cancel with FreehandMouseTool - call its `cancelDraw` function after the changes in this PR. You'd also have to go through and make many of its event handlers check for the presence of data before using it. 

Going to close this - even with those changes this is a little messy and the FreehandMouseTool has a few "interesting" behaviors like marking all the measurement data on the image active when a draw activity starts.

We may come at this from another angle later.

PR #680, opened 2019-07-18
fix(package.json): Bump @ohif/vtk-extension usage to fix picking-rela…

swederik:
…ted performance issue

This disables picking on the widget manager before attaching it to the renderer. That way, if painting is disabled the picking code is avoided. It removes a lot of unnecessary cycles during the VTK component usage.


swederik:
Whoops, don't merge this. Looks like rotation is broken because I removed the export of vtkInteractorStyleMPRSlice from react-vtkjs-viewport.

PR #684, opened 2019-07-18
 fix(config): Fix hardcoded default Google Cloud OAuth redirect_uri 

pavertomato:
(#585)
https://github.com/OHIF/Viewers/pull/675#issuecomment-512328237

### DIFF
--- a/public/config/google.js
+++ b/public/config/google.js
@@ -13,7 +13,7 @@ window.config = {
       // Authorization Server URL
       authority: 'https://accounts.google.com',
       client_id: 'YOURCLIENTID.apps.googleusercontent.com',
-      redirect_uri: `${window.location}callback`, // `OHIFStandaloneViewer.js`
+      redirect_uri: `${window.location.protocol}//${window.location.host}/callback`, // `OHIFStandaloneViewer.js`
       response_type: 'id_token token',
       scope: 'email profile openid https://www.googleapis.com/auth/cloudplatformprojects.readonly https://www.googleapis.com/auth/cloud-healthcare', // email profile openid
       // ~ OPTIONAL


swederik:
Yeah that might not actually work if the viewer is hosted at `domain.com/viewer` or something, so I'll hold off for now.

dannyrb:
We're using this value in `App.js`

```js
    if (this.props.oidc.length) {
      const firstOpenIdClient = this.props.oidc[0];

      this.userManager = getUserManagerForOpenIdConnectClient(
        store,
        firstOpenIdClient
      );
    }
```

We can remove this value as something that can be specified in the config, and then compute it in `App.js`?

```js
const openIdConnectConfiguration = Object.merge({}, firstOpenIdClient, {
    // Make sure `routerBasename` slashes play nice
    redirect_uri: `${window.location.protocol}//${window.location.host}${this.state.props.routerBasename}callback`,
});

this.userManager = getUserManagerForOpenIdConnectClient(
    store,
    openIdConnectConfiguration 
);
```

Or at least something similar

pavertomato:
> We're using this value in `App.js`
> 
> ```js
>     if (this.props.oidc.length) {
>       const firstOpenIdClient = this.props.oidc[0];
> 
>       this.userManager = getUserManagerForOpenIdConnectClient(
>         store,
>         firstOpenIdClient
>       );
>     }
> ```
> 
> We can remove this value as something that can be specified in the config, and then compute it in `App.js`?
> 
> ```js
> const openIdConnectConfiguration = Object.merge({}, firstOpenIdClient, {
>     // Make sure `routerBasename` slashes play nice
>     redirect_uri: `${window.location.protocol}//${window.location.host}${this.state.props.routerBasename}callback`,
> });
> 
> this.userManager = getUserManagerForOpenIdConnectClient(
>     store,
>     openIdConnectConfiguration 
> );
> ```
> 
> Or at least something similar

not sure if this is a good idea, isn't it better to allow user to decide how redirection should work - dynamically or by static url from config?

dannyrb:
> not sure if this is a good idea, isn't it better to allow user to decide how redirection should work - dynamically or by static url from config?

I only gave this a quick glance, so you're probably right. I mostly wanted to comment on a potential solution to the:

> hosted at `domain.com/viewer`

comment that @swederik made. We'll likely need something that allows you to specify an absolute value, or adapts to handle the above situation when your redirect_uri should be relative to the URL the application is hosted at.

swederik:
Addressed this in https://github.com/OHIF/Viewers/pull/698

PR #689, opened 2019-07-19
fix(package.json): Update @ohif/extension-cornerstone to fix loading …

swederik:
…issue


swederik:
Looks like CornerstoneTools 3.15.2 is still pulled in via another package, so we will need to fix that before this fix actually takes effect.

PR #698, opened 2019-07-21
fix(OpenIDConnect): Switch to InMemoryWebStorage for token storage

swederik:
### Changes:
- OpenID Connect routes `redirect_uri`, `silent_redirect_uri`, and `post_logout_ redirect_uri` will have `${protocol}//${host}${routerBasename}` added as a prefix if they are provided as relative paths
- oidc-client-js's UserManager is provided to the Header component via a React Context Provider/Consumer. The UserManager is used to log out.
- Switched from using sessionStorage to using memory to store the current user. The reason is that localStorage/sessionStorage are vulnerable to XSS attacks (https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/HTML5_Security_Cheat_Sheet.md#local-storage).
- Since we are not using localStorage anymore for user credentials, the Google Cloud dataset picker has been switched to take the access token directly rather than the Oidc localStorage key. 
- The logout_redirect callback page that was previously served from `public` has been removed in favour of a React component and route.
- Updated oidc-client-js to 1.8.x.


dannyrb:
@swederik can you add a summary of this as a comment above the appropriate config fields in the oidc example config?

Issue #521, opened 2019-06-05
Improve UX of Script Tag Deployment/Usage


dannyrb:
One of our more popular use cases is dropping the viewer into an existing application, Google Colab, Python Notebook, or Codesandbox to provide a quick demonstration of _something else_, or a quick view of interesting imaging data.

Existing documentation for this approach lives here: https://docs.ohif.org/deployment/recipes/embedded-viewer.html

While it is currently easier than customizing and building the OHIF Viewer, it still has too many hurdles to provide the ease of use one would expect. These should be addressed so it truly feels "drop-in". We need to remember, and optimize for, the use cases and users that this approach makes sense for.

### Existing Issues

- [ ] Co-hosting, and configuring, the location of Cornerstone WebWorker scripts and codecs
    - We can improve this w/ the use of `worker-loader` in `CornerstoneWADOImageLoader`
- [ ] Inclusion of global external dependencies
    - React, ReactDom, Stylesheets, etc.
- [ ] Requiring React specific calls to attach our viewer to a DOM element
    - [ ] Provide a reasonable default `id` to attach to. Allow override w/ config when app bootstraps
- [ ] Use hash routing for embedded scenarios to avoid history push state security issues (seems to occur more frequently when located in an iFrame)

@pieper has a good `Google Colab` playground for testing these requirements.

pieper:
Here's an example that should be viewable by anyone:

https://colab.research.google.com/drive/1AVD_U1nO-vm5zZqXelUP1ZURJBpJXN-M

PR #721, opened 2019-07-27
fix(script-tag): Include React and ReactDOM as dependencies, add installViewer export

swederik:
Addresses #2 and #3 from this Issue: https://github.com/OHIF/Viewers/issues/521

Open to changing the name, not sure if `installViewer` is what users would expect.

Script tag usage is a bit simpler now:

```html
<div id="root"></div>

<script>
    const config = {
      // default: '/'
      routerBasename: '/',
      // default: ''
      relativeWebWorkerScriptsPath: '',
      servers: {
        dicomWeb: [
          {
            name: 'DCM4CHEE',
            imageRendering: 'wadors',
            thumbnailRendering: 'wadors',
          },
        ],
      },
      i18n: {
        LOCIZE_PROJECTID: 'a8da3f9a-e467-4dd6-af33-474d582a0294',
        LOCIZE_API_KEY: null, // Developers can use this to do in-context editing. DO NOT COMMIT THIS KEY!
        USE_LOCIZE: true
      }
    }

    // TODO: This is still necessary
    window.config = config;
</script>
<script src="index.umd.js" crossorigin></script>

<script>
    // Note: Uses the DOM element with the ID "root" by default
    // One could also run installViewer(config, containerId)
    window.OHIFStandaloneViewer.installViewer(config)
</script>
```

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -9,12 +9,8 @@ import svgr from '@svgr/rollup'
 import pkg from './package.json'
 // Deal with https://github.com/rollup/rollup-plugin-commonjs/issues/297
 import builtins from 'rollup-plugin-node-builtins'
+import replace from 'rollup-plugin-replace';
 import serve from 'rollup-plugin-serve'
-
-const globals = {
-  react: 'React',
-  'react-dom': 'ReactDOM',
-}
 
 const startServer = process.env.START_SERVER === 'true';
 
@@ -28,18 +24,19 @@ export default {
       exports: 'named',
       name: 'OHIFStandaloneViewer',
       esModule: false,
-      globals,
     },
     {
       file: pkg.module,
       format: 'es',
       exports: 'named',
       sourcemap: true,
-      globals,
     },
   ],
   plugins: [
     external(),
+    replace({
+      'process.env.NODE_ENV': JSON.stringify('production')
+    }),
     postcss({
       modules: false,
     }),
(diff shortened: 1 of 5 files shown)


rollup.config.js
```diff
@@ -54,6 +51,32 @@ export default {
...
+          'memo',
+          'useContext',
+          'useState',
+          'useEffect',
+          'useLayoutEffect',
+          'Component',
+          'PureComponent',
+          'useRef',
+          'useReducer',
+          'useMemo',
```
dannyrb: It's weird that this is a thing

src/config.js
```diff
@@ -4,9 +4,15 @@ import cornerstoneWADOImageLoader from 'cornerstone-wado-image-loader';
 import dicomParser from 'dicom-parser';
 import version from './version.js';
 
+let homepage;
+const { process } = window;
+if (process && process.env && process.env.PUBLIC_URL) {
+  homepage = process.env.PUBLIC_URL;
+}
+
 window.info = {
```
dannyrb: If we're attaching anything to the window with script-tag, you may want to vendor prefix:
`$ohif.info`

If only to reduce chances of conflict

src/index_publish.js
```diff
@@ -2,5 +2,19 @@
...
+
+function installViewer(props, containerId = "root", callback) {
+	const container = document.getElementById(containerId);
+
+	if (!container) {
+		throw new Error("No root element found to install viewer. Please add a <div> with the id 'root', or pass a DOM element into the installViewer function.");
+	}
+
+	return ReactDOM.render(<App {...props}/>, container, callback);
+}
```
dannyrb: I think users will appreciate this. Thank you 🙌

rollup.config.js
swederik: Yeah it's a pain, I don't know why this is necessary.

src/config.js
swederik: This doesn't change anything we are attaching, it just changes how `process` is checked, since it is undefined sometimes.

We need to get away from using `window.config` entirely and use Context or Redux to pass that data around. This can be another PR though.

Issue #707, opened 2019-07-22
add fuzzymatching


pavertomato:
fuzzymatching  is part of qido standard https://www.dicomstandard.org/dicomweb/query-qido-rs/ but it's not present in ohif-core https://github.com/OHIF/ohif-core/blob/master/src/studies/services/qido/studies.js#L37 
but may be not all DICOMweb server support it
my idea is to add this to ohif-core and parameter to config file:
- if parameter is enabled UI will show checkbox to pass true or false as value of fuzzymatching to request
- if parameter is disabled UI won't contain checkbox and and won't pass fuzzymatching parameter to request.

pavertomato:
any doubts about this plan?

dannyrb:
@pavertomato, my only hold-up is:

> UI will show checkbox to pass true or false as value of fuzzymatching

In my opinion, if fuzzymatching is supported, it's almost always preferred unless we're encountering performance issues from very large datasets. Heck, I even wish we had a _single_ search field that fuzzymatched across all searchable fields, with an "advanced" search option that expanded to show more boxes if you needed fine grained filtering.

tl;dr

I like adding `supportsFuzzyMatching` to the server settings for the application. If true, I think we should use fuzzy matching by default. I don't think we need to add a checkbox to the UI. @swederik may wish to weigh in.

PR #730, opened 2019-07-29
Feature/fuzzymatching

pavertomato:
#707
no UI changes in this PR, if supportsFuzzyMatching is enabled request get additional parameter  with hardcoded value  https://github.com/OHIF/Viewers/compare/master...pavertomato:feature/fuzzymatching?expand=1#diff-514cd21bc59d0a0eb85d700c6872977eR96
 if supportsFuzzyMatching is disabled request don't have fuzzymatching parameter at all

### DIFF
--- a/src/studylist/StudyListWithData.js
+++ b/src/studylist/StudyListWithData.js
@@ -27,11 +27,11 @@ class StudyListWithData extends Component {
     server: PropTypes.object,
     user: PropTypes.object,
     history: PropTypes.object,
-    studyListFunctionsEnabled: PropTypes.bool
+    studyListFunctionsEnabled: PropTypes.bool,
   };
 
   static defaultProps = {
-    studyListFunctionsEnabled: true
+    studyListFunctionsEnabled: true,
   };
 
   static rowsPerPage = 25;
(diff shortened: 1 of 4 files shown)


public/config/google.js
```diff
@@ -2,6 +2,7 @@ window.config = {
   routerBasename: '/',
   relativeWebWorkerScriptsPath: '',
   enableGoogleCloudAdapter: true,
+  supportsFuzzyMatching: true,
```
dannyrb: This setting would need to be per `server` config?

src/studylist/StudyListWithData.js
```diff
@@ -92,6 +92,10 @@ class StudyListWithData extends Component {
       offset: searchData.currentPage * searchData.rowsPerPage,
     };
 
+    if (window.config.supportsFuzzyMatching) {
```
dannyrb: You'll want to pull this from the app's `server` config data, not off the window.
The only place `window.config` _should_ be referenced is `App.js` if no config is provided via props.

public/config/google.js
pavertomato: yeah, good point

src/studylist/StudyListWithData.js
pavertomato: done

PR #740, opened 2019-07-30
fix(ConnectedCornerstoneViewport): Fix definition of function map so it happens less frequently

dlwire:
`onMeasurementsChanged` currently defines a function map every time it is called. This function map includes a throttle on the `modified` function which is not being used because the function is
called at most once per time it is instantiated. These changes elevate the map definition to the
file level, since what it is defining does not change based on the other executed code.

This cleans up the code a little bit, but more importantly because the `throttle` is now working as intended, the performance of the "Measurements" pane is improved.


dannyrb:
CC: @evren217, can you take a peak at this? If it looks good, feel free to merge.

Issue #723, opened 2019-07-29
Switch OHIF Logo from text + font to SVG


swederik:
We are currently loading the entire Sanchez font only to use it for the OHIF Logo. We should remove the font and use the Logo as an SVG instead.

References to Sanchez:
https://github.com/OHIF/Viewers/search?utf8=%E2%9C%93&q=sanchez&type=

Here's a link to the logo in Sanchez (it's in White so may not be visible): https://www.dropbox.com/s/j4ugchl97w74p13/ohiflogo.svg?dl=0

This will simplify script-tag usage, since this font will no longer need to be pulled from Google Fonts.

biharck:
Fixed at https://github.com/OHIF/Viewers/pull/744

dannyrb:
**Acceptance Criteria:**

- [ ] Remove Sanchez Font from template/import
- [ ] Add OHIF "text logo" only in `platform/viewer`
- [ ] Update Header usage to leverage SVGs for icon and text

dannyrb:
Closed by: https://github.com/OHIF/Viewers/pull/1021

PR #744, opened 2019-07-31
fix(index.html OHIFLogo.js variables.css): removing sanchez font

biharck:
#723

This PR depends on https://github.com/OHIF/react-viewerbase/pull/136


biharck:
> For this PR you can keep the Roboto references and only remove the Sanchez ones, since we don't have a solution for the Roboto question yet. If you remove Roboto the site will look weird.

Done

dannyrb:
Deploy Preview: 
image

biharck:
it is up to you. I was trying to do not reinvent the wheel using the existing ones from react-viewerbase such as the logo. If we decided to move it to the Viewers level we may have to move them all.

PR #770, opened 2019-08-14
chore: Cleanup legacy code in core

swederik:
I don't believe we are using any of this stuff anywhere anymore. @evren217 @igoroctaviano @dannyrb ?


platform/core/src/compatibility/dialogPolyfill.js
```diff
@@ -1,458 +0,0 @@
-var dialogPolyfill = (function() {
```
dannyrb: This would polyfill dialog on IE 11 -- I'm not sure if we're using it anywhere or not.
Quickest way to verify would be trying to use CINE dialog in IE11 with this removed?

platform/core/src/compatibility/styleProperty.js
```diff
@@ -1,66 +0,0 @@
-/*
- * https://github.com/swederik/dragula/blob/ccc15d75186f5168e7abadbe3077cf12dab09f8b/styleProperty.js
- */
-(function() {
```
dannyrb: Looks like this was for a previous DnD library. Should be safe to rip out

platform/core/src/schema.js
```diff
@@ -1,34 +0,0 @@
-//import SimpleSchema from 'simpl-schema'
```
dannyrb: We don't use this anymore, but having something similar as our config continues to grow might be nice. Curious if there's a good FOSS package that handles simple schema validation like this. I guess, in a way, react `prop-types` already do.

platform/core/src/classes/index.js
```diff
@@ -18,16 +18,14 @@ import { StudyMetadataSource } from './StudyMetadataSource';
...
-//import { StudySummary } from './metadata/StudySummary';
-
 export {
   OHIFStudyMetadataSource,
   MetadataProvider,
   CommandsManager,
   HotkeysManager,
   ImageSet,
   StudyPrefetcher,
-  //StudyLoadingListener,
```
dannyrb: I'm not familiar with what this was for or did/does

platform/core/src/compatibility/dialogPolyfill.js
swederik: I looked for the <dialog> tag in the code but didn't see it anywhere anymore

platform/core/src/classes/index.js
swederik: Monitoring the progress of image loading across the study. We have to figure out how this works with the change to an extension-based architecture.

evren217: I agree that it would be good to have it back.

PR #775, opened 2019-08-14
fix(Studies): qidoSupportsIncludeField should be true by default

### LINKED ISSUE #772: Catch repo up to latest ohif-core source
This issue got caught up in our migration and needs ported: https://github.com/OHIF/ohif-core/issues/63#event-2529931140

jfmedeiros1820:
fix #772


platform/core/src/studies/services/qido/studies.js
```diff
@@ -118,9 +118,13 @@ export default function Studies(server, filter) {
   };
 
   const dicomWeb = new api.DICOMwebClient(config);
+  let serverSupportsQIDOIncludeField = server.qidoSupportsIncludeField;
```
dannyrb:
```suggestion
 server.qidoSupportsIncludeField = server.qidoSupportsIncludeField === undefined ? true : server.qidoSupportsIncludeField;
```

If possible, I think we can avoid creating a new var, and just update value if it's undefined.

dannyrb: Looks like GitHub doesn't support multi-line suggestions. Replacing all changes with something like:

```js
server.qidoSupportsIncludeField = server.qidoSupportsIncludeField === undefined 
    ? true
    : server.qidoSupportsIncludeField;
```

omartoutounji: Added your suggestions to https://github.com/OHIF/Viewers/pull/801

dannyrb:
Fixed here: https://github.com/OHIF/Viewers/pull/801

Issue #62, opened 2017-03-27
Some problems with loading images with different compressions


christianvargasforero:
Good morning,

I have been testing the viewer with different modalities and different compressions and I have encountered the following problems:

In this example the image does not render well:
![captura de pantalla 2017-03-27 a las 5 01 48 p m](https://cloud.githubusercontent.com/assets/11984387/24380823/e39782b2-1312-11e7-8ad3-d9a9acaa6e7b.png)
file:
[lossy.dcm.zip](https://github.com/OHIF/Viewers/files/874002/lossy.dcm.zip)

Of equal way in this other parts of the image are lost:
![captura de pantalla 2017-03-27 a las 5 05 03 p m](https://cloud.githubusercontent.com/assets/11984387/24380835/e8ee953e-1312-11e7-8ffe-8a225c1b889b.png)
file:
[lossless.dcm.zip](https://github.com/OHIF/Viewers/files/874005/lossless.dcm.zip)

And lastly in this case does not work at levels tool:
![captura de pantalla 2017-03-27 a las 5 08 48 p m](https://cloud.githubusercontent.com/assets/11984387/24380841/ee44ed8a-1312-11e7-886b-629e71a98747.png)
file:
[levels no works.dcm.zip](https://github.com/OHIF/Viewers/files/874004/levels.no.works.dcm.zip)

I send the files so they can try.

thank you very much.

radicalrob:
Thank you for this report and for the details, sample images, etc.  Very helpful.  We will look into it.

swederik:
The first two images are now working with the latest Cornerstone WADO Image Loader.
The third image has a VOI LUT applied automatically. This is something we will have to address.

PR #781, opened 2019-08-15
feat: 🎸 Changes rendering image/thumbnail to use wadors

jfmedeiros1820:
Issue [here](https://github.com/OHIF/ohif-core/issues/62)
Closes: #62

### DIFF
--- a/platform/core/src/utils/getImageId.js
+++ b/platform/core/src/utils/getImageId.js
@@ -38,18 +38,14 @@ export default function getImageId(instance, frame, thumbnail = false) {
 
   const renderingAttr = thumbnail ? 'thumbnailRendering' : 'imageRendering';
 
-  if (
-    !instance[renderingAttr] ||
-    instance[renderingAttr] === 'wadouri' ||
-    !instance.wadorsuri
-  ) {
-    let imageId = 'dicomweb:' + instance.wadouri;
+  let imageId = getWADORSImageId(instance, frame, thumbnail); // WADO-RS Retrieve Frame
+
+  if (!imageId &&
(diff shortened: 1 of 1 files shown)


platform/core/src/utils/getImageId.js
```diff
@@ -38,18 +38,14 @@ export default function getImageId(instance, frame, thumbnail = false) {
 
   const renderingAttr = thumbnail ? 'thumbnailRendering' : 'imageRendering';
 
-  if (
-    !instance[renderingAttr] ||
-    instance[renderingAttr] === 'wadouri' ||
-    !instance.wadorsuri
-  ) {
-    let imageId = 'dicomweb:' + instance.wadouri;
+  let imageId = getWADORSImageId(instance, frame, thumbnail); // WADO-RS Retrieve Frame
```
swederik: I'm not sure I understand the purpose of this change. From what I can tell, this would mean that even if the user has specified that we should be using WADO-URI, if the server and instances support WADO-RS this would use that instead. That doesn't seem like the behaviour we want. What is this trying to fix? I think your other PR where you just fixed the default value for rendering to wadors was sufficient, wasn't it?

jfmedeiros1820: I was trying to use WADO-RS always by default and just using the WADO-URI if it was not possible to get the imageId from WADO-RS way. The previous else condition was getting WADO-RS by default already, that's why I changed to get everytime. But I agree with you, we can use the default value true for rendering that I fixed in the other PR.

swederik:
Closing because the change is unnecessary. Thanks!

Issue #753, opened 2019-08-06
Configure Google Healthcare extension?


bcli4d:
Our webapp presents the user with a list of DICOM studies related to cases in a cohort that the user has created based on various criteria. The user can then select a one more studies to view in separate tabs. In our current implementation, using OHIF 1.x, I just pass the study ID in the URL. BTW, we don't expose the studylist button.

I'm interested in moving to Google Healthcare as our PACS, but don't see a similar ability to command OHIF 2.0 to open a specific study in a chc datastore via the URL. Ideally, I think I'd like to be able to set the default project/location/dataset/dicomstore in a config file, but then override any of those in the URL. E.G., having configured the default project/location/dataset/dicomstore as chc-tcia/us-central1/foo/bar, I could then override the dataset/dicomstore in the URL. Something like:

`https://viewer.ohif.org/dataset/tcga-brca/dicomstore/tcga-brca/StudyInstanceUID/1.3.6...`
 
Is there any such functionality now or planned?

Thanks.

pieper:
+1 I'd like that functionality too but I don't believe it's available yet.  It would be great if someone could add an example.

dannyrb:
This sounds like a nice value add. A couple of ways to approach this:

_Option 1_

1. Configure a default Server
2. Provide an alternative server config with query parameters

_Option 2_

1. Configure multiple servers, identify one as the default
2. Specify a configured server's alias or index in the URL to use instead of the default

If I'm missing the mark, or you think there are alternative implementations worth considering, don't hesitate to chime in ^_^

pieper:
Hi @dannyrb  - It would be ideal if one could visit the url @bcli4d provided and then get redirected to a google oauth login / permissions page.  Then if you have permissions you would get redirected back to the ohif viewer with the correct tokens to access the study in the original url.  There shouldn't need to be any custom servers involved.

bcli4d:
Hi @dannyrb . We currently only expose TCGA images, but there are 21 TCGA cancer types with DICOM images, and Google Healthcare has dedicated a separate dataset/datastore to each in its TCIA public data set. If I understand your proposal, you would have us create and configure a server for each? (Not to mention the 56 other Google Healthcare dataset/datastores.)

dannyrb:
@bcli4d, I didn't have a clear understanding of the problem. @pieper's comment is a much more appropriate solution.

swederik:
This would definitely be a nice addition! We'd be happy to review a PR.

[...]

PR #790, opened 2019-08-21
feat: 🎸 adding server field in google.js

omartoutounji:
Regarding #753, I propose adding an optional `server` field that takes in an object which contains info about your preferred google cloud dataset. If `server`  is provided, the Google DICOM Store Picker modal is hidden/disabled and the studies are retrieved. Otherwise, the modal shows up normally.


platform/viewer/src/studylist/StudyListWithData.js
```diff
@@ -209,6 +209,23 @@ class StudyListWithData extends Component {
...
+          //     qidoRoot:
+          //       'https://healthcare.googleapis.com/v1beta1/projects/dicomserver-249915/locations/northamerica-northeast1/datasets/medicalImages/dicomStores/prostateImages/dicomWeb',
+          //     wadoRoot:
+          //       'https://healthcare.googleapis.com/v1beta1/projects/dicomserver-249915/locations/northamerica-northeast1/datasets/medicalImages/dicomStores/prostateImages/dicomWeb',
+          //     wadoUriRoot:
+          //       'https://healthcare.googleapis.com/v1beta1/projects/dicomserver-249915/locations/northamerica-northeast1/datasets/medicalImages/dicomStores/prostateImages/dicomWeb',
+          //     active: true,
+          //     supportsFuzzyMatching: false,
+          //   },
+          // ]}
```
dannyrb: Do we need this as a comment here?

platform/viewer/src/googleCloud/DicomStorePickerModal.js
```diff
@@ -33,35 +38,62 @@ class DicomStorePickerModal extends Component {
...
-          />
-        </Modal.Body>
-      </Modal>
-    );
+    if (this.props.server) {
+      return (
+        <Modal
+          show={this.props.isOpen}
+          onHide={this.props.onClose}
+          onEnter={this.setServer}
```
dannyrb: We do we need the if block to add this prop?

platform/viewer/src/googleCloud/DicomStorePickerModal.js
```diff
@@ -12,12 +12,17 @@ class DicomStorePickerModal extends Component {
...
+    server: PropTypes.object,
   };
 
   static defaultProps = {
     isOpen: false,
   };
 
+  setServer = () => {
+    if (this.props.server) this.props.setServers(this.props.server);
+  };
```
dannyrb: We can probably continue to call `setServers`, but also call `this.props.setServers` in `mounted` hook if `this.props.server` has a value?

platform/viewer/src/studylist/StudyListWithData.js
omartoutounji: I put it there just as a `server` prop example for the PR. But yes, this shouldn't be in the codebase if we'd like to merge, removing it.

platform/viewer/src/googleCloud/DicomStorePickerModal.js
omartoutounji: Agree. Removed if block with commit https://github.com/OHIF/Viewers/pull/790/commits/004e571168df85bfcb2bf635642da284fb27f812

platform/viewer/src/googleCloud/DicomStorePickerModal.js
omartoutounji: Yes, that's true. Added a mounted hook with commit https://github.com/OHIF/Viewers/pull/790/commits/0d752399dd608cc891bf26b48b8c4398f9fa596a

omartoutounji:
> Hi @omartoutounji, let me know if my feedback makes sense, or if you have pushback.

@dannyrb Your feedback made a lot of sense, thank you for taking the time to review the PR. I added all of your suggestions. Please let me know if there's more to do to get this PR merge-ready!

Also, I propose disabling the 'Change DICOM Store' button if the `server` prop is provided? I believe it makes sense that if a user provides a server prop then they probably don't want to ever change the DICOM store. What do you think?

omartoutounji:
@dannyrb Please let me know if this is not of satisfactory state so I can make necessary changes or close the PR, if needed.

platform/viewer/src/studylist/StudyListWithData.js
```diff
@@ -32,6 +32,10 @@ class StudyListWithData extends Component {
 
   static defaultProps = {
     studyListFunctionsEnabled: true,
+    server:
+      window.config.enableGoogleCloudAdapter && window.config.oidc[0].server
```
dannyrb: It's a bit dangerous to pull from `window.config` here, as it may not be set (null).

These values should be available in the new `appContext` created in: https://github.com/OHIF/Viewers/pull/852

Also, I'm not sure this is where I would set a default value. Maybe in `mounted` if it's not defined here and is set in `appContext`? That would also allow us to set this route prop by modifying the URL (later on). Which I believe is the heart of what this related issue is requesting: https://github.com/OHIF/Viewers/issues/753

dannyrb:
Sorry for the delay. One last comment that should help align what you want with something that's safe to merge :+1:

omartoutounji:
@dannyrb could not load gcloud version of OHIF to manually test my changes. https://github.com/OHIF/Viewers/pull/919 is addressing this

dannyrb:
@omartoutounji merged :+1:
Please let me know if you encounter issues.

omartoutounji:
@dannyrb as I was applying your suggestions, I realized something funny. The functionality we're trying to achieve with this PR, is already available by setting `dicomWeb` in `google.js`. Setting it also disables the modal without any of my changes being present.

dannyrb:
@omartoutounji, ha. That may be the case. I believe I initially thought you wanted a way to change the default/preferred programmatically by setting a prop; but yes, if you're keeping the same dataset, simply using the appConfig should suffice.

omartoutounji:
@dannyrb as far as my use case goes, `dicomWeb` suffices. I just didn't know about it before I made this PR.

I, however, understand better what this PR is for now and will apply your suggestions with that in mind.

Re-reading our previous comments, I believe there was a mismatch in each of our goals for this PR but now it is more aligned!

swederik:
Thanks for all your help!

PR #801, opened 2019-08-25
Qidosupportsincludefield should be true by default

omartoutounji:
Avoided creating new var and updates value if undefined


dannyrb:
Thanks, @omartoutounji!

Issue #700, opened 2019-07-22
Annotations and download image


bhavesh-pandey:
Hi Team,

Firstly great work with the new OHIF react viewer.

I was wondering when the annotate feature and download image feature will come as it's very important for sharing findings in a dicom image. When can we expect to see that?

Thanks

dannyrb:
@bhavesh-pandey, a large portion of the development of the OHIF viewer comes from volunteers like you. While these features will likely eventually be implemented, I can't make any promises around when someone will decide to champion them. If you or your team would be interested in contributing, I can provide some guidance on where to get started and what we would be looking for in a PR? Alternatively, I can direct you toward individuals/teams that have historically done OHIF related contract work.

bhavesh-pandey:
Yeah we would love to do that. If you can help us we can develop those features.

swederik:
The Image Download UI from the previous version was here:

https://github.com/OHIF/Viewers/tree/v1.x/Packages/ohif-viewerbase/client/components/viewer/downloadDialog

The first step would be to port this UI to React and make it work with the Cornerstone extension. Later we would need to decide how this should be handled for e.g. the VTK plugin.

As for the Annotation tool, it should be easy enough to add it to the list of available tools, but we need to add some UI to enter the actual annotation text. This stuff was previously here: https://github.com/OHIF/Viewers/tree/v1.x/Packages/ohif-viewerbase/client/components/viewer/annotationDialogs

swederik:
BTW, this is partly a duplicate of https://github.com/OHIF/Viewers/issues/674 so please search before creating new issues.

PR #840, opened 2019-09-02
feat: Snapshot Download Tool

romulobordezani:
Draft PR for Snapshot Download Tool.
Solves: https://github.com/OHIF/Viewers/issues/700


platform/core/src/redux/reducers/viewports.js
```diff
@@ -88,6 +90,10 @@ const viewports = (state = defaultState, action) => {
       } else {
         return defaultState;
       }
+    case FORCE_VIEWPORT_UPDATE:
+      return Object.assign({}, state, {
+        updateHash: new Date().getUTCMilliseconds().valueOf(),
+      });
```
dannyrb: This is an interesting approach. What are you changing that you need to force a viewport update in this way?

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
+import './DownloadDialog.styl';
+
+import React, { PureComponent } from 'react';
+import { withTranslation } from '../../utils/LanguageProvider';
+import Modal from 'react-bootstrap-modal';
```
dannyrb: I think this is fine for now, but I'm going to create a different issue. We shouldn't be creating _many_ modals that use this dependency. We should have one, and then each other modal is simply content that can be the _child_ for the modal component, and has the ability to supply _actions_ to influence buttons (or something like that)

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
...
+import PropTypes from 'prop-types';
+
+class DownloadDialog extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isOpen: this.props.isOpen,
+      width: 3000,
+      height: 3000,
```
dannyrb: The default values should be the DICOM image's native resolution. This probably needs to be set via props.

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
...
+
+    const formData = {
+      width: this.state.width,
+      height: this.state.height,
+      fileName: this.state.fileName,
+      fileType: this.state.fileType,
+      showAnnotations: this.state.showAnnotations,
+    };
+
+    const thereIsAnyEmpty = Object.values(formData).some(field => field === '');
```
dannyrb: I'm not sure we need validation on this form. It should be "safe enough" to fallback to defaults if no value is provided. I'm not sure what value validation would provide, and it could actually hamper the experience.

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
...
+
+    if (thereIsAnyEmpty) {
+      alert('Please full fill all fields.');
+      return false;
+    }
+
+    this.props.takeAndDownloadSnapShot(formData);
+
+    return false;
```
dannyrb: What does `return false;` do here?

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
...
+
+class DownloadDialog extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isOpen: this.props.isOpen,
+      width: 3000,
+      height: 3000,
+      fileName: '',
```
dannyrb: There should be a default fileName so one doesn't need to be provided if they click save without specifying.

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
...
+        {
+          key: 'jpg',
+          value: 'jpg'
+        },
+        {
+          key: 'png',
+          value: 'png'
+        }
+      ],
+      previewElementRef: null
```
dannyrb: Out of curiosity, why do we track this?

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
...
+   * static method getDerivedStateFromProps, what will replace componentWillReceiveProps
+   * See https://fb.me/react-async-component-lifecycle-hooks for details.
+   * @param references dom element ref to cache on state
+   */
+  assignRef = references => {
+    // Control recalls, allowing this instance on state only once
+    if (this.props.updateHash === null && references !== null) {
+      this.setState({ previewElementRef: references,  });
+
+      this.props.setCacheReferences(
```
dannyrb: What is `setCacheReferences`?

platform/viewer/src/lib/DownloadViewportEngine.js
```diff
@@ -0,0 +1,137 @@
...
+    // Adds the hidden spare viewer on body, away from React Updates
+    const parentNode = document.querySelector('body');
+    parentNode.appendChild(this.$downloadElement);
+  }
+
+  mountPreview() {
+    this.cloneDomElement();
+    this.enableCornerstoneTools();
+    this.showPreview();
+  }
```
dannyrb: Use the blob from the method I've provided so set either an image source or a background-image. We shouldn't need to to do anything like this.

platform/viewer/src/lib/DownloadViewportEngine.js
```diff
@@ -0,0 +1,137 @@
...
+    reRenderAction = this.reRenderActionCache,
+    showAnnotations = this.showAnnotations,
+  ){
+    this.$previewElement = previewElemReference;
+    this.$activeViewport = activeViewport;
+    this.reRenderActionCache = reRenderAction;
+    this.showAnnotations = showAnnotations;
+  }
+
+  enableCornerstoneTools(element = this.$downloadElement) {
```
dannyrb: UI components should never touch `cornerstone` or `cornerstone-tools` API

platform/viewer/src/lib/DownloadViewportEngine.js
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  mountPreview() {
+    this.cloneDomElement();
+    this.enableCornerstoneTools();
+    this.showPreview();
+  }
+
+  save(formData) {
+    cornerstoneTools.SaveAs(
```
dannyrb: This component should have no knowledge of `cornerstone` or `cornerstone-tools`. You'll likely want to provide a `save` or `submit` callback that receives the form data, or a POJO

platform/viewer/src/lib/DownloadViewportEngine.js
```diff
@@ -0,0 +1,137 @@
...
+  toggleAnnotations() {
+    const element = this.$downloadElement;
+    this.availableTools.forEach(tool => {
+      if (this.showAnnotations) {
+        cornerstoneTools.setToolEnabledForElement(element, tool.replace('Tool', ''));
+      } else {
+        cornerstoneTools.setToolDisabledForElement(element, tool.replace('Tool', ''));
+      }
+    });
+  }
```
dannyrb: @swederik, why would someone want an image without annotations? They already have that? If they really want an image without annotations, they could use this tool before adding annotations?

@romulobordezani, that's a dangerous forEach loop to make. We could probably cache that image's `imageIdSpecificToolState`, delete it, take the image, restore it.

platform/viewer/src/lib/DownloadViewportEngine.js
```diff
@@ -0,0 +1,137 @@
+const DOWNLOAD_ELEMENT_ID = 'download-element';
+const PREVIEW_ELEMENT_ID = 'preview-element';
+
+class DownloadViewportEngine {
+  constructor() {
+    this.$previewElement = null;
+    this.$downloadElement = document.createElement('div');
+    this.$activeViewport = null;
```
dannyrb: You should not need DOM references for anything in this component

platform/viewer/src/lib/DownloadViewportEngine.js
```diff
@@ -0,0 +1,137 @@
...
+    previewElemReference = this.$previewElement,
+    activeViewport =  this.$activeViewport,
+    reRenderAction = this.reRenderActionCache,
+    showAnnotations = this.showAnnotations,
+  ){
+    this.$previewElement = previewElemReference;
+    this.$activeViewport = activeViewport;
+    this.reRenderActionCache = reRenderAction;
+    this.showAnnotations = showAnnotations;
+  }
```
dannyrb: What cache are you updating?

platform/viewer/src/lib/DownloadViewportEngine.js
```diff
@@ -0,0 +1,137 @@
+const DOWNLOAD_ELEMENT_ID = 'download-element';
+const PREVIEW_ELEMENT_ID = 'preview-element';
+
+class DownloadViewportEngine {
```
dannyrb: What is meant by "Engine"?

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -239,12 +258,34 @@ function _getVisibleToolbarButtons() {
...
 }
 
+/**
+ * Toggles the Download Dialog Modal
+ */
+function forceRenderUpdate() {
+  this.setState({
+    forceRenderUpdate: !this.state.isDownloadScreenShotDialogOpen,
+  });
+}
```
dannyrb: I'm a little confused. Why would `forceRenderUpdate` toggle the DownloadDialogModal?

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
...
+  static propTypes = {
+    isOpen: PropTypes.bool.isRequired,
+    toggleDownloadDialog: PropTypes.func.isRequired,
+    takeAndDownloadSnapShot: PropTypes.func.isRequired,
+    mountPreview: PropTypes.func.isRequired,
+    onResize: PropTypes.func.isRequired,
+    toggleAnnotations: PropTypes.func.isRequired,
+    forceRenderUpdate: PropTypes.func.isRequired,
+    setCacheReferences: PropTypes.func.isRequired,
+    cleanViewPortClone: PropTypes.func.isRequired,
```
dannyrb: This is going into a "generic" component library. Some of these props feel like they're very specific to our implementation.

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
...
+
+    return false;
+  };
+
+  static propTypes = {
+    isOpen: PropTypes.bool.isRequired,
+    toggleDownloadDialog: PropTypes.func.isRequired,
+    takeAndDownloadSnapShot: PropTypes.func.isRequired,
+    mountPreview: PropTypes.func.isRequired,
+    onResize: PropTypes.func.isRequired,
```
dannyrb: What do we use this for?

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,235 @@
...
+                    placeholder="Width"
+                    value={this.state.width}
+                    onChange={e => {
+                      this.setState({ width: e.currentTarget.value });
+                      this.props.onResize(null, 'width', this.state.width);
+                    }}
+                    className="input-ohif"
+                  />
+                </div>
+                <div className="sizes_height col">
```
dannyrb: What naming pattern are you following? I don't recognize `sizes_height`

platform/ui/src/elements/form/TextInput.js
```diff
@@ -22,6 +22,7 @@ class TextInput extends React.Component {
         value={this.state.value}
         onChange={this.handleChange}
         id={this.props.id}
+        {...this.props}
```
dannyrb: Is there any issue with setting `id` twice?

platform/viewer/src/connectedComponents/ConnectedDownloadDialog.js
```diff
@@ -0,0 +1,43 @@
+import { connect } from 'react-redux';
```
dannyrb: This shouldn't need to be a connected component. All of it's state should be local.

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -110,6 +115,14 @@ class ToolbarRow extends Component {
       zIndex: 999,
     };
 
+    const downloadScreenShotContainerStyle = {
+      display: this.state.isDownloadScreenShotDialogOpen ? 'block' : 'none',
+      position: 'absolute',
+      top: '82px',
+      left: '600px',
```
dannyrb: This may work on your monitor, but what about on a table or phone?

platform/viewer/src/lib/DownloadViewportEngine.js
swederik: They might want to zoom in to a specific part of an image. It's definitely less common but still a requirement.

dannyrb:
Demo: https://5d769b73c82dfb7eb44e9dfb--ohif.netlify.com/pwa/

### Issues:

**Maintain aspect ratio when updating dimensions**

- Navigate to: https://5d769b73c82dfb7eb44e9dfb--ohif.netlify.com/pwa/viewer/1.2.826.0.13854362241694438965858641723883466450351448
- Open `Download` modal
- Notice difference between width/height is `24px`
- Change width to `0`; height is now `24`
- Change width to `3000`; height is now `3024`

The "aspect ratio" needs to be maintained, not the _difference_ between height and width. If the initial values are `232`x`256`.

So if we change `256` to `3000`, `232` becomes `2719`

2719/3000 approximately equals 232/256

**Toggling annotations off and closing modal disables tool**

- Navigate to: https://5d769b73c82dfb7eb44e9dfb--ohif.netlify.com/pwa/viewer/1.2.826.0.13854362241694438965858641723883466450351448
- Draw `Length` annotation
- Open `Download` modal
- Close `Download` modal
- Attempt to draw a new `Length` annotation

Unable to draw new length annotation

**Generated image does not match image preview:**

image

- **Download button should be in "more" submenu**
- **Layout is not responseive**
- **Overlay background should be transparent**
- **Input component styles have different selected state styling**
- **Select component styles differ from previous**
- **Numerous small color, font size, element spacing differences (not a blocker)**
- **Escape should close modal (not a blocker)**

We're matching the implementation at https://viewer-legacy.ohif.org/ . See screen comparisons below.

## Updated

### Toolbar: 

image

### Desktop:

image

### iPad:

image

### < 500px:

image


## Previous

### Desktop:

image

### < 500px

image

dannyrb:
Included quick development testing feedback. After these items are addressed, I can do a more formal code review.

platform/core/src/redux/reducers/viewports.js
romulobordezani: @deprecated

platform/ui/src/components/downloadDialog/DownloadDialog.js
romulobordezani: Indeed, it deserves a new issue.

platform/ui/src/components/downloadDialog/DownloadDialog.js
romulobordezani: Done.

platform/ui/src/components/downloadDialog/DownloadDialog.js
romulobordezani: Validation updated to avoid empties and bad scaled sizes.

platform/ui/src/components/downloadDialog/DownloadDialog.js
romulobordezani: @deprecated

platform/ui/src/components/downloadDialog/DownloadDialog.js
romulobordezani: This 'eccentric' way to pass trough a ref to state is needed now that
   * 'static' method getDerivedStateFromProps doesnt have access to the instance(this),
   * different of the old way with componentWillReceiveProps
   * See https://fb.me/react-async-component-lifecycle-hooks for details.

platform/ui/src/components/downloadDialog/DownloadDialog.js
romulobordezani: Every instance of the Download Engine caches the dom stuff to prevent new query selections and gets from Cornerstone. Improves performance and code readability, once inside the Engine Class we can use them as `this.$activeElement` instead of

`
const originalElement = cornerstone.getEnabledElement(this.$activeViewport).element;
const originalCanvas = originalElement.querySelector('canvas');
`

platform/viewer/src/lib/DownloadViewportEngine.js
romulobordezani: Done.

platform/viewer/src/lib/DownloadViewportEngine.js
romulobordezani: platform/viewer/src/lib/DownloadViewportEngine.js  is an Engine, not the component itself. Maybe its a misunderstood? Can we  keep these communications with cornerstone or cornerstone-tools here?

platform/viewer/src/lib/DownloadViewportEngine.js
romulobordezani: platform/viewer/src/lib/DownloadViewportEngine.js is an Engine, not the component itself. Maybe its a misunderstood? Can we keep these communications with cornerstone or cornerstone-tools here?

platform/viewer/src/lib/DownloadViewportEngine.js
romulobordezani: Proccess to cache that image's imageIdSpecificToolState, delete it, take the image, restore it -> Done.

platform/viewer/src/lib/DownloadViewportEngine.js
romulobordezani: I needed it to know wich is the active ViewPort dom elem and where, inside the React Modal, to put the blob.

platform/viewer/src/lib/DownloadViewportEngine.js
romulobordezani: @Deprecated.

platform/viewer/src/lib/DownloadViewportEngine.js
romulobordezani: Engine is the name I gave to where I connect to cornerstone, cornerstone tools and others libs like the one to generate blobs.

platform/viewer/src/connectedComponents/ToolbarRow.js
romulobordezani: @deprecated.

platform/ui/src/components/downloadDialog/DownloadDialog.js
romulobordezani: Please review. Props updated.

platform/ui/src/components/downloadDialog/DownloadDialog.js
romulobordezani: When the user changes the sizes (resizes) on  modal.

platform/ui/src/components/downloadDialog/DownloadDialog.js
romulobordezani: It´s BEM (http://getbem.com/). its is a block called `sizes` within an element called `height`, I forgot one underscore, sorry.  :)


It should be sizes__height

platform/viewer/src/connectedComponents/ConnectedDownloadDialog.js
romulobordezani: Please, could you send me an example of how to connect to the ToolbarRow without connect it? 

I followed what you advice with Cine  Dialog, if there is a better way to do this, would be nice to update cine dialog connection too.

platform/viewer/src/connectedComponents/ToolbarRow.js
romulobordezani: @deprecated - We are using Modals now.

dannyrb:
shoo stalebot

igoroctaviano:
@dannyrb I pushed the implementation with the cornerstone logic isolated in the connected component (viewer). It's testable right now and like you said, ideally we can extract it to extension commands.

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -0,0 +1,279 @@
...
+
+    const {
+      dataUrl,
+      width: viewportElementWidth,
+      height: viewportElementHeight
+    } = await updateViewportPreview(viewportElement, downloadCanvas, fileType);
+
+    setViewportPreviewSrc(dataUrl);
+    setViewportPreviewHeight(viewportElementHeight);
+    setViewportPreviewWidth(viewportElementWidth);
```
dannyrb: If you're always using these three together, you can do:

```
const [viewportPreview, setViewportPreview] = useState({
  src, // undef or null
  height,
  width
});

...

setViewportPreview({
  src: dataUrl,
  height: viewportElementHeight,
  width: viewportElementWidth,
});

File: platform/ui/src/components/downloadDialog/DownloadDialog.js [hunk 24]
```diff
@@ -0,0 +1,279 @@
...
+  const [viewportElementWidth, setViewportElementWidth] = useState(minimumSize);
+
+  const [downloadCanvas, setDownloadCanvas] = useState();
+  const [downloadCanvasHeight, setDownloadCanvasHeight] = useState(minimumSize);
+  const [downloadCanvasWidth, setDownloadCanvasWidth] = useState(minimumSize);
+
+  const [viewportPreview, setViewportPreview] = useState();
+  const [viewportPreviewSrc, setViewportPreviewSrc] = useState();
+  const [viewportPreviewHeight, setViewportPreviewHeight] = useState(minimumSize);
+  const [viewportPreviewWidth, setViewportPreviewWidth] = useState(minimumSize);
```
dannyrb: Love hooks. My only comment is that any state that you often update together can be grouped as a single `useState` with an object + properties. I think it's _slightly_ more performant and less code.

platform/ui/src/components/downloadDialog/DownloadDialog.styl
```diff
@@ -0,0 +1,102 @@
...
+
+.DownloadDialog
+  color: var(--text-secondary-color);
+  filter: drop-shadow(0 0 3px var(--ui-gray-darkest));
+  border: none;
+  border-radius: 8px;
+  width: inherit;
+  padding: 15px;
+  background: transparent;
+  z-index: 1080 !important;
```
dannyrb: Why `!important`?

platform/ui/src/components/downloadDialog/DownloadDialog.styl
```diff
@@ -0,0 +1,102 @@
...
+
+    @media screen and (max-width: 1023px)
+      flex-direction: column;
+
+    .col
+      flex-grow: 1;
+
+      .input-ohif
+        margin-left: 15px;
+        @media screen and (max-width: 1023px)
```
dannyrb: Oh, interesting. I had no idea you could do this with `stylus` (nested)

platform/ui/src/components/downloadDialog/DownloadDialog.styl
```diff
@@ -0,0 +1,102 @@
...
+          margin-left: 0;
+          margin-top: 5px;
+          width: 100% !important;
+
+      .file-type
+        .select-ohif
+          margin-left: 17px;
+          @media screen and (max-width: 1023px)
+            margin-left: 0;
+            width: 100%!important;
```
dannyrb: Missing space. Also, not sure of `!important` usage throughout file.

platform/ui/src/components/downloadDialog/DownloadDialog.styl
```diff
@@ -0,0 +1,102 @@
...
+    background-color: #16202b;
+
+    .form-control.input-ohif
+      padding: 6px 12px;
+
+    @media screen and (max-width: 1023px)
+      flex-direction: column;
+
+    .col
+      flex-grow: 1;
```
dannyrb: :+1: for using flexbox

platform/ui/src/elements/form/Select.css
```diff
@@ -1,42 +1,37 @@
...
-  font-family: Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue',
-    sans-serif;
-  font-size: 0.8em;
-  font-weight: 600;
-  color: #444;
-  line-height: 1.3;
-  padding: 0.6em 1.4em 0.5em 0.8em;
-  width: 100%;
-  max-width: 100%;
+.select-ohif-container {
```
dannyrb: Empty style?

platform/ui/src/elements/form/TextInput.css
```diff
@@ -1,7 +1,37 @@
...
+  height: 40px;
+  margin: 0 10px 20px;
+  padding: 0 20px;
+  cursor: pointer;
+  border: none;
+  background-color: var(--input-background-color);
+  color: var(--input-placeholder-color);
+  font-size: 10pt;
+  font-weight: normal;
+  width: auto !important;
```
dannyrb: Why important?

platform/viewer/src/connectedComponents/ConnectedDownloadDialog.js
```diff
@@ -0,0 +1,106 @@
...
+
+  return {
+    minimumSize: MINIMUM_SIZE,
+    defaultSize: DEFAULT_SIZE,
+    canvasClass: "cornerstone-canvas",
+    onClose: ownProps.toggleDownloadDialog,
+    activeViewport: activeEnabledElement,
+    enableViewport: viewportElement => {
+      if (viewportElement) {
+        cornerstone.enable(viewportElement);
```
dannyrb: Is this getting `cornerstone` from some global declaration? Can we import it instead of relying on it being set globally?

platform/viewer/src/connectedComponents/ConnectedDownloadDialog.js
```diff
@@ -0,0 +1,106 @@
...
+              const newHeight = Math.min(height || image.height, MAX_TEXTURE_SIZE);
+
+              resolve({ image, width: newWidth, height: newHeight });
+            });
+        }
+      }),
+    toggleAnnotations: (toggle, viewportElement) => {
+      cornerstoneTools.store.state.tools.forEach(({ name }) => {
+        if (toggle) {
+          cornerstoneTools.setToolEnabledForElement(viewportElement, name);
```
dannyrb: Same for tools

platform/viewer/src/lib/utils/b64toBlob.js
```diff
@@ -0,0 +1,22 @@
+/* Enabled JPEG images downloading on IE11. */
```
dannyrb: This might be worth shifting to core.

platform/ui/src/components/downloadDialog/DownloadDialog.js
igoroctaviano: Yeah, a lot better! setting big objects sometimes scares me a little but in this case its good. Done, thx.

platform/ui/src/components/downloadDialog/DownloadDialog.js
igoroctaviano: Done! thx.

platform/ui/src/components/downloadDialog/DownloadDialog.styl
igoroctaviano: In this specific case its because in Modal, the styles were applied in styles attribute directly, it's tricky to override using selectors.

platform/ui/src/elements/form/Select.css
igoroctaviano: Good catch, I just removed, thank you.

platform/ui/src/elements/form/Select.js
```diff
@@ -18,32 +28,27 @@ class Select extends Component {
...
-      >
-        {this.props.options.map(({ key, value }) => {
-          return (
-            <option key={key} value={value}>
-              {key}
-            </option>
-          );
-        })}
-      </select>
+      <div className="select-ohif-container">
```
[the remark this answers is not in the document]
igoroctaviano: > This is okay for now. Might be worth asking the product team the best way forward for class naming. @rodrigolabs brought this up in a recent code review.
> 
> My gut says vendor names should prefix styles, if we're going to use them at all, ie:
> 
> * `ohif-select-container`
> 
> But I think we can mitigate scope issues, and have simpler classes/styling by leveraging CSS Modules or something similar (scoped css).

I believe that vendor names should prefix styles as well, this was the old code but I think that css modules (scoped css) is the way to go.

platform/ui/src/elements/form/TextInput.css
igoroctaviano: Yeah, old stuff again, I just removed and tested, seems that it's not that important anymore.

platform/viewer/src/connectedComponents/ConnectedDownloadDialog.js
igoroctaviano: You are totally right, done. thanks.

platform/viewer/src/connectedComponents/ConnectedDownloadDialog.js
igoroctaviano: Done! thx.

platform/viewer/src/lib/utils/b64toBlob.js
igoroctaviano: Done! thx.

platform/ui/src/components/downloadDialog/DownloadDialog.styl
igoroctaviano: > See comments for code specific suggestions.
> 
> Here is some additional feedback after playing with the deploy preview:
> 
> * If I erase the name for the image, it downloads as `.png` or `.jpeg`. Can we use a default name if the value is an empty string?
> * For wider screens, the image preview is not centered; can we center it?
> * For smaller screens, the image preview can leak out of it's box.
>   
>   * Maybe a `max-width: 100%` or something? Could use
> 
> @igoroctaviano, everything is looking _really good_. I think we're very close to having this merged. Thank you for your hard work 🙏

So, I have added a default filename like you said, good catch! I also change the way i was aligning the image preview, now it seems to be centered. The only one I could not replicate was the small screen leak, can you share some steps? maybe im missing one.

igoroctaviano: > Missing space. Also, not sure of `!important` usage throughout file.

done! thx.

platform/ui/src/components/downloadDialog/DownloadDialog.styl
dannyrb: Ah, okay. Fine for now. Hopefully modal context and new component will make it easier to clean this up later

PR #841, opened 2019-09-02
fix: 🐛 Activating Pan and Zoom on right and middle click by def

galelis:
Issue:
- Pan and Zoom tools should be activated by default on right and middle click respectively. 

Cause: 
- `react-cornerstone-tools` will only activate tools that are left button (1) tools and also have another button mask. 

Fix: 
- Changing `mouseButtonMask` adding button 1 into mask array.

### DIFF
--- a/platform/viewer/src/setupTools.js
+++ b/platform/viewer/src/setupTools.js
@@ -142,8 +142,8 @@ function getResetLabellingAndContextMenu(store) {
 export default function setupTools(store) {
   const toolLabellingFlowCallback = getToolLabellingFlowCallback(store);
   const availableTools = [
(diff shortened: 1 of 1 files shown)


dannyrb:
This is a good quick fix, but that's an odd restriction to have in place. I think the code that does the filtering you've described is located here:

https://github.com/cornerstonejs/react-cornerstone-viewport/blob/master/src/CornerstoneViewport/CornerstoneViewport.js#L775-L797

I'll try to put together a new issue that would address this behavior.

galelis:
@dannyrb I totally agree with you, when I was debugging, I've noticed that. 
Maybe adding a new option in the availableTools to say which one should be by default activated on each buttonMask.

Issue #725, opened 2019-07-29
Replace window.config usage with Context or Redux


swederik:
We are currently setting window.config to store the viewer configuration and accessing it all over the place. For script-tag usage, we should be able to set the config using a prop to the top-level application, and this config should be used throughout the viewer. We could probably do this with Context or possibly Redux. I'm not sure which one is more appropriate.

https://github.com/OHIF/Viewers/search?q=window.config&unscoped_q=window.config

PR #852, opened 2019-09-03
refactor: 💡 React components to consume appConfig using Context

ladeirarodolfo:
### Includes
- React components to consume app configuration using React Context and
React Hooks
- Non React components to continue using global variable
window.config.
- Related documentation also changed.

Closes: #725


platform/viewer/src/studylist/StudyListWithData.js
```diff
@@ -204,7 +210,7 @@ class StudyListWithData extends Component {
     let healthCareApiWindows = null;
 
     // TODO: This should probably be a prop
```
swederik: I think you can remove the TODO :-)

platform/viewer/src/App.js
```diff
@@ -31,12 +31,15 @@ import { OidcProvider } from 'redux-oidc';
...
 import { getActiveContexts } from './store/layout/selectors.js';
 import i18n from '@ohif/i18n';
 import setupTools from './setupTools.js';
 import store from './store';
-import UserManagerContext from './UserManagerContext';
+
+// Contexts
+import WhiteLabellingContext from './context/WhiteLabellingContext';
+import UserManagerContext from './context/UserManagerContext';
+import AppContext from './context/AppContext';
```
dannyrb: 👍

platform/viewer/src/App.js
```diff
@@ -114,52 +117,61 @@ class App extends Component {
       );
     }
 
-    _initExtensions(this.props.extensions);
-    _initServers(this.props.servers);
+    this.appConfig = props || {};
```
dannyrb: Would props ever be `undefined`? If so, the app wouldn't be able to run without at least some default values (server). Should we check for those and throw/show warnings that are helpful to the user so they know where to look for configuration options?

platform/viewer/src/App.js
```diff
@@ -114,52 +117,61 @@ class App extends Component {
...
+      </AppContext.Provider>
     );
   }
 }
 
 /**
  * @param
  */
-function _initExtensions(extensions) {
+function _initExtensions(extensions, appConfig) {
```
dannyrb: `appConfig` should contain extensions? Can we just pass in `appConfig` and destructure?

`function _initExtensions({ extensions, hotkeys } = {}) {`

platform/viewer/src/studylist/StudyListWithData.js
```diff
@@ -51,9 +55,10 @@ class StudyListWithData extends Component {
   };
 
   componentDidMount() {
+    const { appConfig = {} } = this.context;
```
dannyrb: Can we make `this.context` equal to the `appConfig` instead of `this.context.appConfig`?
That way we don't need to destructure and provide a default in each of these instances.

When we first set it, just make `context` equal to `appConfig || {}`

platform/viewer/src/studylist/StudyListWithData.js
ladeirarodolfo: ok :D

platform/viewer/src/App.js
ladeirarodolfo: yes sure

platform/viewer/src/studylist/StudyListWithData.js
ladeirarodolfo: We could but, AppContext could have more than appConfig prop, plus, we could/should better organize appConfig/extension/or anything else when initially setting AppContext Provider. This is the first step for this improvement and at end we wont pass unnecessary properties into appConfig or AppContext

ladeirarodolfo: plus I didnt do the whole change because it will involve changing something not covered on this scope (i.e extensions/servers...)

dannyrb: That works for me. Thanks for providing good and thoughtful comments ^_^

platform/viewer/src/App.js
ladeirarodolfo: alright.. I will remove it! When time comes and we split passed props into (properties appConfig, servers, extensions...) that part will make more sense..

PR #862, opened 2019-09-04
fix: 🐛 Local file: failing when retrieving segmentation data

ladeirarodolfo:
### Includes
Fix Dicom Seg data retrieval for local file.

BREAKING CHANGE: DICOM Seg

More changes to come...

part of #838


extensions/cornerstone/src/handleSegmentationStorage.js
```diff
@@ -73,8 +74,12 @@ async function handleSegmentationStorage(
     studyInstanceUid,
     displaySetInstanceUid
   );
-  const segWadoUri = displaySet.images[0].getData().wadouri;
-  const arrayBuffer = await retrieveDicomData(segWadoUri);
+
+  const imageInstace =
+    displaySet && displaySet.images && displaySet && displaySet.images[0];
```
dannyrb: `const hasImageInstance = displaySet && displaySet.images && displaySet.images.length` ?

extensions/cornerstone/src/handleSegmentationStorage.js
```diff
@@ -1,8 +1,8 @@
-import * as dcmjs from "dcmjs";
+import * as dcmjs from 'dcmjs';
 
-import OHIF from "@ohif/core";
-import cornerstone from "cornerstone-core";
-import cornerstoneTools from "cornerstone-tools";
+import OHIF from '@ohif/core';
```
swederik: @dannyrb Are these expected changes from the linter? If so, can we just run the linter on everything at once so we don't see anything in the PRs anymore?

swederik:
Added some changes to make it work on Google Cloud too, but it is a bit ugly.

extensions/cornerstone/src/handleSegmentationStorage.js
dannyrb: Double quotes to single quotes is a change made by `prettier`.

1. Verify husky pre-commit is working for prettier/eslint autofix for modified files
    a. `lint-staged` helps with this
2. Apply changes to all files

_Most_ files should be linted already. Some slight rule discrepancies may have existed across repositories, but now that we have unified rules, it should allow us to be more consistent.

platform/viewer/src/connectedComponents/ViewerLocalFileData.js
```diff
@@ -41,6 +41,10 @@ class ViewerLocalFileData extends Component {
         studyMetadata.createDisplaySets(sopClassHandlerModules);
       studyMetadata.setDisplaySets(study.displaySets);
 
+      studyMetadata.forEachDisplaySet(displayset => {
+        displayset.localFile = true;
+        displayset.studyIndex = studyIndex;
```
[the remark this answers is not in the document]
swederik: The array is almost always going to have < 5 entries. 99% of the time it will have only one entry, so the find will be basically instantaneous anyway. Cleaner to have one less property.

platform/viewer/src/lib/filesToStudies.js
```diff
@@ -1,160 +1,20 @@
...
-    const dataset = dcmjs.data.DicomMetaDictionary.naturalizeDataset(
-      dicomData.dict
-    );
-    dataset._meta = dcmjs.data.DicomMetaDictionary.namifyDataset(
-      dicomData.meta
-    );
-    dataset.imageId = image.imageId;
-
-    datasets.push(dataset);
+  const processFilesPromises = files.map(file => {
```
swederik: I think you can just write `files.map(processFile)`

platform/viewer/src/lib/localFileLoaders/dicomFileLoader.js
```diff
@@ -0,0 +1,206 @@
...
+    dataset.imageId = image.imageId || imageId;
+
+    return dataset;
+  }
+
+  getStudies(dataset, imageId) {
+    return this.datasetsToStudies([dataset]);
+  }
+
+  datasetsToStudies(datasets) {
```
swederik: I think the PR is pretty good. My only reservation is that this function is overly complicated right now since it is only ever called on one dataset. One dataset can only produce one series and instance, so all the looping and filtering logic inside here is no longer necessary.

extensions/cornerstone/src/handleSegmentationStorage.js
```diff
@@ -1,8 +1,11 @@
...
+import * as dcmjs from 'dcmjs';
+import { api } from 'dicomweb-client';
 
-import OHIF from "@ohif/core";
-import cornerstone from "cornerstone-core";
-import cornerstoneTools from "cornerstone-tools";
+import OHIF from '@ohif/core';
+import cornerstone from 'cornerstone-core';
+import cornerstoneTools from 'cornerstone-tools';
+import DICOMWeb from '@ohif/core/src/DICOMWeb';
```
swederik: @dannyrb Is this type of import across packages OK?

ladeirarodolfo: ok I will change the approach :D

platform/viewer/src/lib/filesToStudies.js
ladeirarodolfo: yeah.. thank you

platform/viewer/src/lib/localFileLoaders/dicomFileLoader.js
ladeirarodolfo: ok .. thnk you

platform/viewer/src/connectedComponents/ViewerLocalFileData.js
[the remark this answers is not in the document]
ladeirarodolfo: ok done

Issue #724, opened 2019-07-29
Include Roboto font via WebPack


swederik:
In the Meteor version we eventually switched from Google fonts to locally including the Roboto font. We should do this again. Here was the commit originally: https://github.com/OHIF/Viewers/commit/87d61298b438604140a671cfb03449ae36498f58. This was helpful because otherwise testing the viewer when offline meant that it would load very slowly since it kept trying to reach out to Google to get the fonts.

WebPack can be used to import the font:

https://webpack.js.org/guides/asset-management/#loading-fonts

This will simplify script-tag usage, since this font will no longer need to be pulled from Google Fonts.

dannyrb:
@swederik, I would actually prefer we don't bundle the font, unless it's for script tag usage. For PWA, having it as a separate include allows us to offload the request to Google's lightning fast CDN and leverage caching that many users will have from having requested that same font for a different website (and Roboto is popular).

For PWA offline scenarios, we can make sure our service worker has a copy of assets we load from Google's CDN. If the asset is fresh enough, or if we're offline, we can use the locally cached copy instead.

This keeps our bundle size lower, improves load times, and makes it easier to not lock users into a specific font.

swederik:
That's fine with me too, I just don't want the end user for script-tag usage to have to include the fonts in their HTML page, I don't really have a strong opinion on the solution.

dannyrb:
**For Script Tag:**

Let's embed. It's one less thing for the user to care about, and we're not as concerned with performance tuning there.

**For Embeded:**

We'll need to modify it's WebPack build.

- Look at [`webpack-workbox-plugin`](https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin)
- Here is a [recipe for caching google fonts](https://developers.google.com/web/tools/workbox/guides/common-recipes)

Caching WebPack application files is probably outside the scope of this issue. There are a few hurdles to navigate carefully, and we don't want to over-complicate this ticket.

dannyrb:
**Additional guidance:**

For bundling w/ the commonjs/umd build, import the font in `index-umd.js`. The entry point file for that bundle.

It _should_ load and bundle. If not, you'll need to add a webpack loader that supports the font's file format.

For PWA, look at my most recent comment on that issues. It's mostly about setting the correct caching rules for the service-worker.

PR #906, opened 2019-09-10
#724 Fonts refactor

galelis:


dannyrb:
Notes: `app.bundle.[hash].js` is not cached. It's too large for our current max file size filter.

## Console:

image

## Application's Tab

#### Registration
image

#### Precache Assets:
image

#### Runtime: Google Fonts Stylesheets
image

#### Runtime: Google Fonts Webfonts
image

dannyrb:
PWA is looking good, but `yarn run build:package` is not bundling the font files:

image

dannyrb:
@galelis, it may not be worth bundling for UMD, but we'll need to run that by @swederik.

Most loaders simply split and spit out chunks in `dist`.
UMD may be able to import them from relative location, but would assume they're hosted alongside one-another.
This would work okay for `unpkg`, but not offline scenarios.

- https://survivejs.com/webpack/loading/fonts/

We could `base64` inline them, but it feels like a lot of bloat, and the loaders appear out of date:

- https://stackoverflow.com/questions/41657087/webpack-inline-font-with-url-loader
- https://webpack.js.org/loaders/url-loader/ (may not support)

`file-loader` does not appear to support what we're trying to do:

- https://webpack.js.org/loaders/file-loader/

### Recommended:

Either import using `url-loader` and publish alongside `index.umd.js`
Document that `roboto` needs to be loaded if using a source other than `unpkg` or your own bundler from the npm dependency

OR

Assume `roboto` is available globally.
Document that users should load roboto.
If they don't we fallback to the system font-stack.

dannyrb:
CC: @swederik, review on Monday?

PR #918, opened 2019-09-16
Fix: Update cornerstone-tools to 4.0.7

### LINKED ISSUE #917: Probe tool not working
Hi Probe Tool is not working. Reproducible at https://viewer.ohif.org/

Earlier it use to come like this: 

Screenshot 2019-09-16 at 2 44 58 PM

Now the values are not coming:

Screenshot 2019-09-16 at 2 45 38 PM

swederik:
Closes #917

### DIFF
--- a/extensions/cornerstone/package.json
+++ b/extensions/cornerstone/package.json
@@ -32,7 +32,7 @@
     "@ohif/ui": "^0.50.0",
     "cornerstone-core": "^2.2.8",
     "cornerstone-math": "^0.1.8",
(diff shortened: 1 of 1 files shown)


swederik:
Don't merge this yet, it doesn't actually fix the issue. I guess we need to update it everywhere + the lock files.

dannyrb:
@swederik, bump it in `/platform/viewer` with the yarn command? We install from frozen lockfile, so the lockfile does need to be bumped.

PR #919, opened 2019-09-16
fix: 🐛 Fix issue on not loading gcloud

ladeirarodolfo:
Fixed issue on not loading gcloud. Error on name variable declaration

### Included items
The correct name variable is _appConfig and not appConfig


dannyrb:
@ladeirarodolfo, if you have a high degree of confidence that a small change like this will fix a bug and have no side effects, no need to request a review.

PR #921, opened 2019-09-17
Check for Value in 00081155 sequence

bhavesh-pandey:
(Few patient protocol images doesn't have this value) and removed a duplicate declaration


dannyrb:
Is `00081155` a required value?

Issue #836, opened 2019-09-02
Change approach for loading study metadata from WADO to QIDO+WADO


swederik:
This ticket addresses a metadata-loading performance issue that has come up a few times lately. @salimkanoun reported it over email with a set of PET/CT studies and @fredericcervenansky have also run into it (https://github.com/OHIF/Viewers/issues/824) when deploying against their hospital PACS. 

The problem is that when we load a study right now we first make a WADO-RS call to retrieve all of the metadata for every instance related to that StudyInstanceUID. When a study has a large number of instances, this initial call can become very slow. Large studies e.g. PET/CT and DICOM Whole-slide Image studies suffer from performance issues because of this issue.

The proposed change is to switch from a single initial WADO-RS RetrieveMetadata call at the Study level to a set of calls: First, a QIDO-RS SearchForSeries call at the Study Level, and then a set of WADO-RS RetrieveMetadata calls at the Series level.

### Steps:

1. User loads a study into the Viewer
2. Viewer makes QIDO-RS SearchForSeries request given the StudyInstanceUID to retrieve the list of series. If this fails, the viewer displays an error that the study was not found.
3. Viewer starts making WADO-RS RetrieveMetaData requests at the Series level for each of the series listed in the study.
4. As each series is retrieved, createDisplaySets should be run on the series (https://github.com/OHIF/Viewers/blob/8f758186c1a096a80b92596c4c796f927d1d9a8b/platform/core/src/classes/metadata/StudyMetadata.js#L115-L190). The display sets should be concatenated to the current set of display sets in the study.
5. Whenever a new display set is added to the list of display sets for the study, the list should be sorted by series number to ensure the proper ordering in the study.

These calls can be tested on our public DICOMWeb server: https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs/

**QIDO-RS SearchForSeries:**
Docs: https://dicomcloud.azurewebsites.net/swagger/#!/QidoRS/QidoRS_SearchForSeries
GET ${qidoRoot}/studies/{StudyInstanceUID}/series

**WADO-RS RetrieveMetadata at the Series Level:**
Docs: https://dicomcloud.azurewebsites.net/swagger/#!/WadoRS/WadoRS_GetSeries
GET ${wadoRsRoot}/studies/{StudyInstanceUID}/series/{SeriesInstanceUID}

### Notes:
[...]

PR #922, opened 2019-09-17
Adding support for QIDO+WADO call for study metadata load

emfol:
https://github.com/OHIF/Viewers/issues/836

### DIFF
--- a/platform/core/src/classes/metadata/StudyMetadata.js
+++ b/platform/core/src/classes/metadata/StudyMetadata.js
@@ -204,11 +204,11 @@ export class StudyMetadata extends Metadata {
         )
     );
 
-    return displaySets;
+    return sortDisplaySetList(displaySets);
   }
 
   sortDisplaySets() {
-    this._displaySets.sort(_sortBySeriesNumber);
+    sortDisplaySetList(this._displaySets);
   }
 
   createAndAddDisplaySetsForSeries(sopClassHandlerModules, series) {
@@ -228,8 +228,7 @@ export class StudyMetadata extends Metadata {
   setDisplaySets(displaySets) {
     if (Array.isArray(displaySets) && displaySets.length > 0) {
       displaySets.forEach(displaySet => this.addDisplaySet(displaySet));
-      // @TODO Reenable after fix
-      // this.sortDisplaySets();
+      this.sortDisplaySets();
     }
   }
 
@@ -684,9 +683,13 @@ function _getDisplaySetFromSopClassModule(
  * @param {*} a - DisplaySet
  * @param {*} b - DisplaySet
  */
-function _sortBySeriesNumber(a, b) {
+function sortBySeriesNumber(a, b) {
   const seriesNumberAIsGreaterOrUndefined =
     a.seriesNumber > b.seriesNumber || (!a.seriesNumber && b.seriesNumber);
 
   return seriesNumberAIsGreaterOrUndefined ? 1 : -1;
+}
+
+function sortDisplaySetList(list) {
+  return list.sort(sortBySeriesNumber);
 }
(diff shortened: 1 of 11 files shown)


platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -88,6 +88,92 @@ export class StudyMetadata extends Metadata {
     return this._displaySets.slice();
   }
 
+  _createDisplaySetsForSeries(
```
emfol: the bulk of the work originally carried on by the `createDisplaySets` method but to a single series object

platform/core/src/studies/services/wado/seriesLoadQueue.js
```diff
@@ -0,0 +1,89 @@
...
+ * @param {string} givenSeriesIds A list of Series Instance UIDs which will be added to the load queue;
+ */
+function makeSeriesLoadQueue(dicomWebClient, studyId, givenSeriesIds) {
+  if (
+    dicomWebClient instanceof api.DICOMwebClient &&
+    isValidId(studyId) &&
+    Array.isArray(givenSeriesIds)
+  ) {
+    const queue = Object.create(seriesLoadQueueProto);
+    const seriesIds = [];
```
JamesAPetts: Could you replace `seriesIds` with `seriesInstanceUids` throughout (including function names), to be consistent with DICOM naming?

platform/core/src/studies/services/wado/seriesLoadQueue.test.js
```diff
@@ -0,0 +1,96 @@
...
+  let seriesLoadQueue;
+  const studyId = '1.2.3.4';
+  const seriesIds = ['1.2.3.4.1', '1.2.3.4.2'];
+  const dicomWebClient = new api.DICOMwebClient();
+
+  beforeEach(function() {
+    seriesLoadQueue = makeSeriesLoadQueue(dicomWebClient, studyId, seriesIds);
+  });
+
+  it('should create a series load queue instance', function() {
```
JamesAPetts: This particular test only checks if the properties defined on the object exist. I don't think this is useful test to have. 

The tests below that check for event emission and consumption work as intented are nice though.

platform/core/src/utils/updateMetaDataManager.js
```diff
@@ -70,32 +70,40 @@ const getWadoRsInstanceMetaData = (study, series, instance) => {
...
-        );
-      }
-    });
+      // TODO Drop dependency on this
+      cornerstoneWADOImageLoader.wadors.metaDataManager.add(imageId, metaData);
+    }
   });
 }
+
+export default function updateMetaDataManager(study, seriesInstanceUid) {
```
JamesAPetts: I'm a big fan of this "additional parameters for specific, or-do-all" approach, looks nice :+1:.

platform/core/src/studies/services/wado/seriesLoadQueue.js
```diff
@@ -0,0 +1,89 @@
...
+  return typeof subject === 'string' && subject.length > 0;
+}
+
+/**
+ * Creates a series load queue instance for the given study instance UID
+ * @param {api.DICOMwebClient} dicomWebClient The instance of the DICOMWebClient instance which will be used to load the series metadata.
+ * @param {string} studyId The Study Instance UID from which the series will be loaded
+ * @param {string} givenSeriesIds A list of Series Instance UIDs which will be added to the load queue;
+ */
+function makeSeriesLoadQueue(dicomWebClient, studyId, givenSeriesIds) {
```
JamesAPetts: I understand that pubSub must have been pulled in from the source of some external lib. But it feels like both this file and `pubSub.js` could be rewritten as ES6 classes and it'd be a lot easier to follow.

I think this is warranted for such core code. But it does _work_ in prototypical ES5 style.

JamesAPetts: What library did you pull pubSub from? It looks like it could be used as inspiration and just be refactored into one class.

dannyrb:
Closing. Feel free to follow this guidance on the other PR.

PR #927, opened 2019-09-18
fix: Use HTML5Backend for drag-drop if the device does not support touch

swederik:
fix: Use HTML5Backend for drag-drop if the device does not support touch, since it's less buggy

Note: still need to test the deploy preview on a real device, which is why this is marked as a draft.

### DIFF
--- a/platform/ui/package.json
+++ b/platform/ui/package.json
@@ -48,8 +48,9 @@
     "prop-types": "15.6.2",
     "react-bootstrap-modal": "4.2.0",
     "react-dates": "18.4.1",
-    "react-dnd": "8.0.3",
-    "react-dnd-touch-backend": "^0.8.3",
+    "react-dnd": "9.3.4",
+    "react-dnd-html5-backend": "^9.3.4",
+    "react-dnd-touch-backend": "^9.3.4",
     "react-i18next": "^10.11.0",
     "react-with-direction": "1.3.0"
   },
--- a/platform/ui/src/utils/viewerbaseDragDropContext.js
+++ b/platform/ui/src/utils/viewerbaseDragDropContext.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import { DndProvider } from "react-dnd";
 import HTML5Backend from 'react-dnd-html5-backend';
 import TouchBackend from 'react-dnd-touch-backend';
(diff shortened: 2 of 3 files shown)


platform/ui/package.json
```diff
@@ -48,8 +48,9 @@
     "prop-types": "15.6.2",
     "react-bootstrap-modal": "4.2.0",
     "react-dates": "18.4.1",
-    "react-dnd": "8.0.3",
-    "react-dnd-touch-backend": "^0.8.3",
+    "react-dnd": "9.3.4",
+    "react-dnd-html5-backend": "^9.3.4",
+    "react-dnd-touch-backend": "^9.3.4",
```
dannyrb: That's a huge version bump

platform/ui/src/utils/viewerbaseDragDropContext.js
```diff
@@ -1,13 +1,20 @@
-import { DragDropContext } from 'react-dnd';
+import React from 'react';
+import { DndProvider } from "react-dnd";
```
dannyrb: How'd you slip in a double quote? 🤔

platform/ui/src/utils/viewerbaseDragDropContext.js
```diff
@@ -1,13 +1,20 @@
...
 // See https://github.com/react-dnd/react-dnd/issues/186#issuecomment-335429067
 // https://github.com/react-dnd/react-dnd/issues/186#issuecomment-282789420
-// TODO: Find a way for this context to be used in the parent application as well.
-
+//
 // http://react-dnd.github.io/react-dnd/docs/api/drag-drop-context
 export default function viewerbaseDragDropContext(DecoratedClass) {
-  return DragDropContext(TouchBackend({ enableMouseEvents: true }), null, true)(
-    DecoratedClass
+  const backend = isTouchDevice ? TouchBackend({ enableMouseEvents: true }) : HTML5Backend;
```
dannyrb: Curious the impact this would have on devices that support touch and mouse. IE. touchscreen laptop

dannyrb:
`CircleCi` issue is caused by the Cypress "Orb". Here's the related issue:
https://github.com/cypress-io/circleci-orb/issues/178

platform/ui/package.json
swederik: I think 0.8.3 was turned into 8.0.3 for them somehow. It's actually not much different

PR #928, opened 2019-09-18
DICOMWeb auth function

jody-zeitler:
Our DICOMWeb server requires an Authorization token that is not derived from a "username:password" string. To make this as flexible as possible, `server.requestOptions.auth` can be a function that returns the entire Authorization header value.

As I added a test case for this feature, I also fixed up the input for the other test cases which passed an object instead of a string for the `auth` value.

### DIFF
--- a/platform/core/src/DICOMWeb/getAuthorizationHeader.js
+++ b/platform/core/src/DICOMWeb/getAuthorizationHeader.js
@@ -9,7 +9,7 @@ import user from '../user';
  * @export
  * @param {Object} [server={}]
  * @param {Object} [server.requestOptions]
- * @param {string} [server.requestOptions.auth]
+ * @param {string|function} [server.requestOptions.auth]
  * @returns {Object} { Authorization }
  */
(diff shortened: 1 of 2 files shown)


swederik:
Just commenting to mention that this is related: https://github.com/OHIF/Viewers/issues/896

dannyrb:
Thanks, @jody-zeitler!
This feels like a good solution until we have more time to improve our customization around auth and user access control.

MuriloSchaefer:
Hi,

I was trying to use this solution to get a token for the requests, but then I realize that we can't store a function in sessionStorage.

Today the server config is saved in `sessionStorage.js` using `JSON.stringify`. The problem is that `stringify` ignores functions, so when you define `requestOptions.auth` as a function, this function isn't saved in sessionStorage.

What would be a good solution for that?

dannyrb:
@MuriloSchaefer, can you open a new issue so we can discuss potential solutions?

PR #935, opened 2019-09-19
feat: 🎸 React custom component on toolbar button

ladeirarodolfo:
### Included items
- Implement custom react component solution for toolbar elements
- Changes on docs

### Closes issue:
Closes: 924

### Screenshoot
The following screenshoot is a sample of its usage and it is NOT part of this PR
Sep-19_-2019-2_11-PM

### DIFF
--- a/platform/viewer/src/connectedComponents/ToolbarRow.js
+++ b/platform/viewer/src/connectedComponents/ToolbarRow.js
@@ -153,32 +153,19 @@ class ToolbarRow extends Component {
 function _getCustomButtonComponent(button, activeButtons) {
   const CustomComponent = button.CustomComponent;
 
-  function isFunctionalComponent(Component) {
-    return (
-      typeof Component === 'function' &&
-      !(Component.prototype && Component.prototype.isReactComponent)
-    );
+  function isValidComponent(Component) {
+    return (typeof Component === 'function');
   }
 
-  function isClassComponent(Component) {
-    return !!(
-      typeof Component === 'function' &&
-      Component.prototype &&
-      Component.prototype.isReactComponent
-    );
-  }
   // Check if its a valid customComponent. Later on an CustomToolbarComponent interface could be implemented.
-  if (
-    isClassComponent(CustomComponent) ||
-    isFunctionalComponent(CustomComponent)
-  ) {
+  if (isValidComponent(CustomComponent)) {
     const parentContext = this;
     const isActive = activeButtons.includes(button.id);
 
     return (
       <CustomComponent
         parentContext={parentContext}
-        toolbarClickCallback={_handleToolbarButtonClick}
+        toolbarClickCallback={_handleToolbarButtonClick.bind(this)}
         button={button}
         key={button.id}
         activeButtons={activeButtons}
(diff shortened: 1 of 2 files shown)


platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -150,46 +150,98 @@ class ToolbarRow extends Component {
...
 
-        return childButton;
-      });
+  function isClassComponent(Component) {
+    return !!(
+      typeof Component === 'function' &&
+      Component.prototype &&
+      Component.prototype.isReactComponent
+    );
+  }
```
dannyrb: I don't know that we care if it's a functional component or a class component. We only really need it to be defined, and we can assume that if it is, that they've supplied the right type of arg. These could both be swapped with:

`const shouldUseCustomComponent = typeof CustomComponent === 'function';`

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -150,46 +150,98 @@ class ToolbarRow extends Component {
...
+
+      if (button.CustomComponent) {
+        getButtonComponentMethod = _getCustomButtonComponent;
+      } else if (button.buttons && button.buttons.length) {
+        getButtonComponentMethod = _getExpandableButtonComponent;
+      }
+
+      return getButtonComponentMethod.call(_this, button, activeButtons);
+    })
+    .filter(component => !!component);
```
dannyrb: This is protecting against empty components? I think I'd rather it blew up so the user was aware of their configuration error.

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -150,46 +150,98 @@ class ToolbarRow extends Component {
...
+    .map(button => {
+      let getButtonComponentMethod = _getDefaultButtonComponent;
+
+      if (button.CustomComponent) {
+        getButtonComponentMethod = _getCustomButtonComponent;
+      } else if (button.buttons && button.buttons.length) {
+        getButtonComponentMethod = _getExpandableButtonComponent;
+      }
+
+      return getButtonComponentMethod.call(_this, button, activeButtons);
```
dannyrb: I think this is easier to read:
Feel free to push back.

```js
      const hasCustomComponent = button.CustomComponent;
      const hasNestedButtonDefinitions = button.buttons && button.buttons.length;

      if (hasCustomComponent) {
        return _getCustomButtonComponent(button, activeButtons);
      }
      if (hasNestedButtonDefinitions) {
        return _getExpandableButtonComponent(button, activeButtons);
      }

      return _getDefaultButtonComponent(button, activeButtons);
```

If it's important to maintain `this`, you can use `.call()`. I'm not sure it's needed, as it feels like an odd lexical scope to maintain.

platform/viewer/src/connectedComponents/ToolbarRow.js
ladeirarodolfo: ok

platform/viewer/src/connectedComponents/ToolbarRow.js
ladeirarodolfo: ok

platform/viewer/src/connectedComponents/ToolbarRow.js
ladeirarodolfo: Yeah.. just changed to what you proposed.. about *this* its actually required considering current implementation of *_handleToolbarButtonClick* so I kept it...

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -150,46 +150,88 @@ class ToolbarRow extends Component {
   }
 }
 
+function _getCustomButtonComponent(button, activeButtons) {
+  const CustomComponent = button.CustomComponent;
+
+  function isValidComponent(Component) {
+    return (typeof Component === 'function');
+  }
```
dannyrb: I don't understand the advantage of creating a temporary function instead of simple variable that explains the result/check?

platform/viewer/src/connectedComponents/ToolbarRow.js
dannyrb: Works for me. We can continue to pass the scope for the time being.

platform/viewer/src/connectedComponents/ToolbarRow.js
ladeirarodolfo: The unique advantage is for maintainability purpose.. but in other hand it will cost a bit more of memory usage.. Changed it to variable declaration instead

PR #937, opened 2019-09-20
fix: Add some code splitting for PWA build

swederik:
TODO: Use React.Lazy and a Suspense wrapper thing: https://reactjs.org/docs/code-splitting.html#reactlazy


swederik:
Just going to merge this now and come back to the React.Lazy thing later. Hopefully this will help fix the timeouts we are seeing on the CircleCI demo viewer build step.

PR #945, opened 2019-09-24
fix: 🐛 Set series into active viewport by clicking on thumbnail

galelis:
Users should be able to set the series for the active viewport by
clicking it's thumbnail into Study browser

Scheenshot:
OHIF-Viewer

Closes: #895

### DIFF
--- a/platform/ui/src/components/studyBrowser/ThumbnailEntry.js
+++ b/platform/ui/src/components/studyBrowser/ThumbnailEntry.js
@@ -99,13 +99,13 @@ class ThumbnailEntry extends Component {
 
   onClick = () => {
     if (this.props.onClick) {
-      this.props.onClick();
+      this.props.onClick(this.props);
     }
   };
 
   onDoubleClick = () => {
     if (this.props.onDoubleClick) {
-      this.props.onDoubleClick();
+      this.props.onDoubleClick(this.props);
     }
   };
 }
--- a/platform/ui/src/components/studyBrowser/ThumbnailEntry.js
+++ b/platform/ui/src/components/studyBrowser/ThumbnailEntry.js
@@ -97,23 +97,15 @@ class ThumbnailEntry extends Component {
     );
   }
 
-  getCallbackData() {
-    const { studyInstanceUid, displaySetInstanceUid } = this.props;
-    return {
-      studyInstanceUid,
-      displaySetInstanceUid,
-    };
-  }
-
   onClick = () => {
     if (this.props.onClick) {
-      this.props.onClick(this.getCallbackData());
+      this.props.onClick(this.props.displaySetInstanceUid);
     }
   };
 
   onDoubleClick = () => {
     if (this.props.onDoubleClick) {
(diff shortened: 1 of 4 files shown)


TZubiri:
Neat!

platform/ui/src/components/studyBrowser/ThumbnailEntry.js
```diff
@@ -99,13 +99,13 @@ class ThumbnailEntry extends Component {
 
   onClick = () => {
     if (this.props.onClick) {
-      this.props.onClick();
+      this.props.onClick(this.props);
```
dannyrb: I'm not sure I follow why we need to pass all of `props` instead of just the click event?

platform/viewer/src/connectedComponents/ConnectedStudyBrowser.js
```diff
@@ -28,13 +31,55 @@ const mapStateToProps = (state, ownProps) => {
...
+    }
+
+    return study.displaySets.find(displaySet => {
+      return displaySet.displaySetInstanceUid === displaySetInstanceUid;
+    });
+  };
+
+  return {
+    ...propsFromState,
+    onThumbnailClick: data => {
```
dannyrb: Is this just a previously unused prop?

platform/viewer/src/connectedComponents/ConnectedStudyBrowser.js
```diff
@@ -28,13 +31,55 @@ const mapStateToProps = (state, ownProps) => {
...
+      const displaySet = findDisplaySet(
+        ownProps.studies,
+        data.studyInstanceUid,
+        data.displaySetInstanceUid
+      );
+
+      propsFromDispatch.setViewportSpecificData(
+        propsFromState.activeViewportIndex,
+        displaySet
+      );
```
dannyrb: Do we have a way to change the active displaySet without passing it in it's entirety? Ideally, we should be able to just set the stack by index, study id, displayset id, etc.

It feels like overkill to surface this amount of information in a click event.

platform/ui/src/components/studyBrowser/ThumbnailEntry.js
galelis: I've changed taht to pass the basic info about the item, as we need its displaySetUID to perform the viewport update.

platform/viewer/src/connectedComponents/ConnectedStudyBrowser.js
galelis: It was actually mergin a not used props `activeViewportIndex` that was only used inside mergeProps functions. We just need studies. 
Code is updated.

platform/viewer/src/connectedComponents/ConnectedStudyBrowser.js
galelis: That was actually a very good callout @dannyrb thanks for that
It called out my attention on other reducers and also to test how many data do I need to update the specific data on the viewport. Turns out, only the displaySet works fine.

PR #946, opened 2019-09-24
Changes to accomodate react-cornerstone-viewport updates

dannyrb:
## This I've Broken

- ~setViewportActive; should be managed by viewer, not extension~
- ~fix: don't configure tools if undefined/empty array (react-cornerstone-viewport)~
- ~Fix MeasurementsPanel Extension event listeners~
- Fix DICOM Seg loading
  - `SEGMENTATION_STORAGE: '1.2.840.10008.5.1.4.1.1.66.4',`
  - `extensions/cornerstone/src/handleSegmentationStorage.js`
- CINE Player needs to use new props on `react-cornerstone-viewport`
- Double check stack prefetch is working
- Double check each extension module is working as expected

## Follow-up

- `computedViewportPanes` should eventually be replaced by simplified/correct state
    - Marrying `viewportSpecificData` and `viewportPanes`
    - Until then, could use a "selector" instead of updating in component
- Measurements... Should be pushed from `cornerstone` extension to a service provided by `viewer`
- Measurements extension should consume measurements service to show data in sidepanel

Related: https://github.com/OHIF/Viewers/pull/1012


platform/ui/src/components/studyBrowser/StudyBrowser.js
```diff
@@ -1,62 +1,86 @@
...
+      ),
+    })
+  ).isRequired,
+  onThumbnailClick: PropTypes.func,
+  onThumbnailDoubleClick: PropTypes.func,
+};
+
+StudyBrowser.defaultProps = {
+  studies: [],
+  onThumbnailClick: noop,
```
swederik: What is the point of the noop if the prop is not required?

platform/viewer/src/components/ViewportGrid/ViewportGrid.js
```diff
@@ -0,0 +1,186 @@
...
+ * @param {string} displaySetInstanceUid
+ * @returns
+ */
+function _getDisplaySetStackFromStudies(
+  studies,
+  studyInstanceUid,
+  displaySetInstanceUid
+) {
+  if (!studyInstanceUid || !displaySetInstanceUid) {
+    return undefined;
```
swederik: `return;` is identical, though maybe we should just throw an error

platform/viewer/src/components/ViewportGrid/ViewportGrid.js
```diff
@@ -0,0 +1,186 @@
...
+  if (storedStack) {
+    return {
+      // studyInstanceUid,
+      // displaySetInstanceUid,
+      imageIds: storedStack.imageIds,
+      frameRate: storedStack.frameRate,
+    };
+  }
+
+  return undefined;
```
swederik: Useless return?

platform/ui/src/components/studyBrowser/StudyBrowser.js
dannyrb: https://github.com/OHIF/Viewers/pull/946/files/097af230eed60c98f43ca5269a9fab283bd79199#diff-d586a02af305f42f8e214dd9e355cf93R44

It makes the binding prettier. Spreading props to set them for a component is a code smell, and this allows us to avoid some conditional logic.

**Previous:**
```
<Thumbnail
  key={`${studyIndex}_${thumbIndex}`}
  {...this.props}
/>
```

**Without Spread:**
```
<Thumbnail
  key={`${studyIndex}_${thumbIndex}`}
  onClick={e => {
    if (this.props.onThumbnailClick) {
      this.props.onThumbnailClick(e);
    }
  }
  onDoubleClick={e => {
    if (this.props.onThumbnailDoubleClick) {
      this.props.onThumbnailDoubleClick(e);
    }
  }
/>
```

**Without Spread + Noop:**
```
const noop = () => {}; // no operation

<Thumbnail
  key={`${studyIndex}_${thumbIndex}`}
  onClick={onThumbnailClick}
  onDoubleClick={onThumbnailDoubleClick}
/>
```

FWIW, this is a pattern used in the React codebase. Lodash even has a `noop` function, haha.

platform/viewer/src/components/ViewportGrid/ViewportGrid.js
dannyrb: There is a fair amount here that we need to talk about. I have a running list of questions around loading data and passing them in an appropriate/consistent way to viewports.

platform/viewer/src/components/ViewportGrid/ViewportGrid.js
dannyrb: I could keep it or leave it. It's reminder that we're returning a value (undefined) if `storedStack` is undefined

platform/ui/src/components/studyBrowser/StudyBrowser.js
swederik: That makes a lot of sense!

dannyrb:
The bulk of this has been merged over various other PRs. It was a nice proof-of-concept spike, but the smaller PRs made it easier to more tightly control quality and identify issues caused by smaller sets of changes as they arose.

PR #949, opened 2019-09-24
ui fix: Made mpr more accessible from toolbar.

TZubiri:
The Toolbar was showing 2 expandable buttons with an "etc..." icon, one of them had only 1 button, MPR.

This way, mpr is better accessible with fewer clicks to get to it.

Codewise, I removed some commented out code (don't worry, git will remember), and reduced the amount of state shared between 2 components to the bare minimum.


dannyrb:
Great value add! Thanks, @TZubiri! 🎉

PR #953, opened 2019-09-26
OHIF-836 Change approach for loading study metadata from WADO to QIDO…

emfol:
…+WADO

@ This is a combination of 7 commits.
 - OHIF-836 [WIP] Adding support for QIDO+WADO study load
 - OHIF-836 [WIP] Adding support for lazy initialization of study object
 - OHIF-836 [WIP] Adapting view to support lazy loading of series...
 - OHIF-836 [WIP] Disabling sorting...
 - OHIF-836 [WIP] Fixing issue with display set thumbnails and re-enabling display set sorting
 - OHIF-836 Adding sorting by series number before initializing series load queue
 - OHIF-836 Adding unit tests for pubSub and seriesLoadQueue modules + Fixing pubSub nitification routine to also include topic name


platform/viewer/public/config/default.js
```diff
@@ -13,6 +13,7 @@ window.config = {
         qidoSupportsIncludeField: true,
         imageRendering: 'wadors',
         thumbnailRendering: 'wadors',
+        enableStudyLazyLoad: true,
```
emfol: This new config allows to enable/disable the QIDO+WADO study metadata load. Setting this to false will make the viewer behave exactly as before (a single WADO call for the entire study metadata);

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -270,30 +273,138 @@ function getRadiopharmaceuticalInfo(instance) {
...
+      instances: [],
+    };
+    study.seriesMap[seriesInstanceUid] = series;
+    study.seriesList.push(series);
   }
 
-  const anInstance = resultData[0];
-  if (!anInstance) {
-    return;
+  const sopInstanceUid = DICOMWeb.getString(instance['00080018']);
```
emfol: This file seems like having many changes but it was mostly moving portions of code around (splitting very large functions into smaller ones) in order to reuse specific steps of those functions.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -1,77 +1,118 @@
...
+      studies: null,
+      error: null,
+    };
+  }
 
-      // Render the viewer when the data is ready
-      // TODO: CLEAR THIS SOMEWHERE ELSE
-      studyMetadataManager.purge();
+  triggerDataLoad() {
+    const { props } = this;
```
dannyrb: Why not destructure this to:

`const { server, studyInstanceUids, seriesInstanceUids } = this.props`

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -1,77 +1,118 @@
...
+    retrieveStudiesMetadata(
+      props.server,
+      props.studyInstanceUids,
+      props.seriesInstanceUids
+    ).then(
+      studies => {
+        if (this.isReady) {
```
dannyrb: We shouldn't need an `isReady` flag to indicate that the DOM is ready. Update the data/state and let React care about that bit for you?

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -1,77 +1,118 @@
...
+      error: null,
+    };
+  }
 
-      // Render the viewer when the data is ready
-      // TODO: CLEAR THIS SOMEWHERE ELSE
-      studyMetadataManager.purge();
+  triggerDataLoad() {
+    const { props } = this;
+    retrieveStudiesMetadata(
```
dannyrb: Can we use `async/await` here?

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -1,77 +1,118 @@
...
         // Updates WADO-RS metaDataManager
         updateMetaDataManager(study);
-
         studyMetadataManager.add(studyMetadata);
-
+        // Attempt to load remaning series if any
+        this._attemptToLoadRemainingSeries(study, studyMetadata);
         return study;
       });
+      this.setState({ studies });
```
dannyrb: This is a very large, complex, array of objects to jam into reactive state. Do we really need all of this data to be reactive? It looks like we were already doing this, so probably outside the scope of this PR, but something to think on as we move forward.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -1,77 +1,118 @@
...
+        seriesLoadQueue.dequeue();
+      }
     }
   }
 
+  componentDidMount() {
+    // TODO: CLEAR THIS SOMEWHERE ELSE
+    studyMetadataManager.purge();
+    this.isReady = true;
+    this.triggerDataLoad();
```
dannyrb: Wait. So `isReady` is always true when `triggerDataLoad()` is called? This would only _not_ be true if the component was destroyed before the query resolved.

- Request the data, and create a cancellation token
- When the request resolves, clear the token
- In `willUnmount`, if the token is set, use it to cancel the request

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -1,77 +1,118 @@
...
+  setStudies(givenStudies) {
+    if (Array.isArray(givenStudies) && givenStudies.length > 0) {
+      const sopClassHandlerModules =
+        extensionManager.modules['sopClassHandlerModule'];
       // Map studies to new format, update metadata manager?
-      const updatedStudies = studies.map(study => {
+      const studies = givenStudies.map(study => {
         const studyMetadata = new OHIFStudyMetadata(
           study,
           study.studyInstanceUid
```
dannyrb: I know you didn't write this, but I don't understand why we need to pass `studyInstanceUid` when it already exists on `study`

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -1,77 +1,118 @@
...
         );
-        const sopClassHandlerModules =
-          extensionManager.modules["sopClassHandlerModule"];
-
-        study.displaySets =
-          study.displaySets ||
-          studyMetadata.createDisplaySets(sopClassHandlerModules);
+        if (!study.displaySets) {
```
dannyrb: Did you not like the previous syntax?

```js
study.displaySets =
          study.displaySets ||
          studyMetadata.createDisplaySets(sopClassHandlerModules);
```

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -1,77 +1,118 @@
...
+    );
+  }
 
+  setStudies(givenStudies) {
+    if (Array.isArray(givenStudies) && givenStudies.length > 0) {
+      const sopClassHandlerModules =
+        extensionManager.modules['sopClassHandlerModule'];
       // Map studies to new format, update metadata manager?
-      const updatedStudies = studies.map(study => {
+      const studies = givenStudies.map(study => {
```
dannyrb: It's not super clear what's happening in this map.

It looks like we're...
- creating/setting display sets, if needed
- creating a new `OHIFStudyMetadata` object
    - Giving it the display sets
    - Maybe appending it to the `study` somewhere with a hidden modify by reference?
- Use our work to update our `metadataManager`

After all of that, we do some `PubSub` magic to listen for new series being added to the study (when/why would that happen?)?  If one's added, we hydrate it using a more granular approach to what's spelled out above, and update our entire studies array in our reactiveState?

This isn't on you -- I realize that this is largely how this has worked up until now; I'm just trying to read what's happening, and it feels non-obvious.

platform/core/src/utils/updateMetaDataManager.js
```diff
@@ -70,32 +70,40 @@ const getWadoRsInstanceMetaData = (study, series, instance) => {
...
+    }
   });
 }
+
+export default function updateMetaDataManager(study, seriesInstanceUid) {
+  if (seriesInstanceUid) {
+    const series = study.seriesMap[seriesInstanceUid];
+    if (series) {
+      updateMetaDataManagerForSeries(study, series);
+    }
```
dannyrb: 1. Add JSDoc to indicate that `seriesInstanceUid` is optional?
2. Add JSDoc to indicate the type for "study"?
3. Reduce if/else nesting:

```js
const shouldUpdateSeriesMeta = seriesInstanceUid;
if (shouldUpdateSeriesMeta) {
    const series = study.seriesMap[seriesInstanceUid];
    updateMetaDataManagerForSeries(study, series);
} else {
    // ...
}
```

You shouldn't need to verify that `series` exists. If they're passing in a bad `seriesInstanceUid`, which they probably pulled off the same `study` object they're passing in, then that's on them.

platform/viewer/public/config/default.js
dannyrb: I'm struggling to think of a reason we would need to preserve the old behavior.

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -653,9 +683,13 @@ function _getDisplaySetFromSopClassModule(
  * @param {*} a - DisplaySet
  * @param {*} b - DisplaySet
  */
-function _sortBySeriesNumber(a, b) {
+function sortBySeriesNumber(a, b) {
```
dannyrb: I believe we used an underscore to indicate that the method lived in this module and was not exported.

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -270,30 +273,138 @@ function getRadiopharmaceuticalInfo(instance) {
...
+      sopInstance.greenPaletteColorLookupTableDescriptor = greenPaletteColorLookupTableDescriptor;
+      sopInstance.bluePaletteColorLookupTableDescriptor = bluePaletteColorLookupTableDescriptor;
+    }
   }
 
+  series.instances.push(sopInstance);
+  return sopInstance;
+}
+
+function createStudy(server, anInstance) {
```
dannyrb: `anInstance`?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -309,139 +420,36 @@ async function resultDataToStudyMetadata(server, studyInstanceUid, resultData) {
...
+ * series list.
+ *
+ * @param server
+ * @param instanceList
+ * @resolves {{seriesList: Array, patientName: *, patientId: *, accessionNumber: *, studyDate: *, modalities: *, studyDescription: *, imageCount: *, studyInstanceUid: *}}
+ */
+async function createStudyFromInstanceList(server, instanceList) {
+  if (Array.isArray(instanceList) && instanceList.length > 0) {
+    const anInstance = instanceList[0];
+    if (anInstance) {
```
dannyrb: Is it even possible for this to be undefined?
What is "an instance"? If that's the technical name, can we detail it a bit in the JSDoc?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -309,139 +420,36 @@ async function resultDataToStudyMetadata(server, studyInstanceUid, resultData) {
...
+      const study = createStudy(server, anInstance);
+      await addInstancesToStudy(server, study, instanceList);
+      return study;
+    }
+  }
+  return null;
```
dannyrb: It looks like previously we didn't return `null`. In some instances, it's good if an exception/error is thrown, because it tells us our assumptions were wrong.

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -452,18 +460,101 @@ async function resultDataToStudyMetadata(server, studyInstanceUid, resultData) {
...
+    .then(result => createStudyFromInstanceList(server, result));
+}
+
+async function lazyLoadStudyMetadata(server, studyInstanceUid) {
+  const seriesIds = await searchStudySeries(server, studyInstanceUid);
+  if (seriesIds) {
+    const dicomWeb = new api.DICOMwebClient({
+      url: server.wadoRoot,
+      headers: DICOMWeb.getAuthorizationHeader(server),
+    });
```
dannyrb: Both load methods can use the same client instance. No need to create it in both implementations?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -452,18 +460,101 @@ async function resultDataToStudyMetadata(server, studyInstanceUid, resultData) {
...
+        server,
+        firstSeries.instances
+      );
+      if (study) {
+        setQueueListeners(server, study, seriesLoadQueue);
+        return study;
+      }
+    }
+  }
+  return null;
```
dannyrb: Why?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -452,18 +460,101 @@ async function resultDataToStudyMetadata(server, studyInstanceUid, resultData) {
...
+        }
+      );
+    }
+  });
+}
+
+async function searchStudySeries(server, studyInstanceUID) {
+  const dicomWeb = new api.DICOMwebClient({
+    url: server.qidoRoot,
+    headers: DICOMWeb.getAuthorizationHeader(server),
```
dannyrb: This is also using a DICOMWebClient configured like the other two?

JamesAPetts:
I left comments in #922 , as I didn't realise you made a new PR.

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -452,18 +460,101 @@ async function resultDataToStudyMetadata(server, studyInstanceUid, resultData) {
...
+      studyInstanceUid,
+      seriesIds
+    );
+    const firstSeries = await seriesLoadQueue.dequeue();
+    if (firstSeries && firstSeries.instances) {
+      const study = await createStudyFromInstanceList(
+        server,
+        firstSeries.instances
+      );
+      if (study) {
```
dannyrb: When would this be undefined?

platform/core/src/studies/services/wado/seriesLoadQueue.js
```diff
@@ -0,0 +1,89 @@
...
+
+function enqueue(queue, seriesId) {
+  if (isValidId(seriesId)) {
+    queue[SERIES_IDS].unshift(seriesId);
+    return true;
+  }
+  return false;
+}
+
+function dequeue(queue) {
```
dannyrb: The naming here is interesting.
I don't know that I would expect `dequeue` to `retrieveSeriesMetadata`
It feels like a `queue` implementation that has a side-effect
That then also publishes an event that causes a different side-effect.
Very hard to follow

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: Agreed...

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: Yup!

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: Good point!

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: Good point!

platform/viewer/public/config/default.js
swederik: Maybe if QIDO-RS for study isn't supported, or if it is worse for performance on some servers. I would keep the flag but enable it by default.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: The semantic of:
`if (!a) { a = b }`
is different from:
`a = a || b`
as the former is a conditional assignment operation and the later is an unconditional one.
This changes though when we have a scenario like:
`if (b) { a = b } else { a = c }`
where the assignment to the variable "a" always takes place making it equivalent to:
`a = b || c`
In this specific situation it does not make much difference but if study.displaySets was a setter instead of a regular property this change would result in no additional calls that setter.
The following code snippet illustrates this:
`
var o = {
  count: 0,
  set a(v) {
    this._a = v;
    this.count++;
  },
  get a() {
    return this._a;
  }
};
if (!o.a) o.a = 'a';
if (!o.a) o.a = 'b';
if (!o.a) o.a = 'c';
if (!o.a) o.a = 'd';
console.log(o.a); // 'a'
console.log(o.count); // 1
o.a = o.a || 'a';
o.a = o.a || 'b';
o.a = o.a || 'c';
o.a = o.a || 'd';
console.log(o.a); // 'a'
console.log(o.count); // 5
`

dannyrb: Ooh. That's interesting.
Is this an optimization JS makes under the hood?

emfol: No... This cannot be optimized because as I said the semantic is different.
Changing from `if (!o.a) o.a = b` to `o.a = o.a || b` and vice-versa may have different side effects as the gist in the link above demonstrates.

dannyrb: I follow -- I was just trying to understand how/why we see benefits with `if (!o.a) o.a = b`
I get that there is a benefit; it's just not immediately obvious to me _why_ it's only called once if repeatedly checked.

platform/core/src/utils/updateMetaDataManager.js
emfol: Done!

platform/core/src/classes/metadata/StudyMetadata.js
emfol: Makes sense but since things are not exported by default it's unnecessary. Please let me know if I should return to what it was (in which case I'll have to adopt the leading underscore to the new local function I created such that all non-exported functions start with that instead of just that one)

platform/core/src/studies/services/wado/retrieveMetadata.js
emfol: Hummm... `aSopInstance` would be better?

platform/core/src/studies/services/wado/retrieveMetadata.js
emfol: Sorry too much defensive programming... :-)
I'll remove that conditional.
I think I'll change everything to `aSopInstance`...

platform/core/src/studies/services/wado/retrieveMetadata.js
emfol: one uses `wadoRoot` the other one uses `qidoRoot` (which might be different according to the config)...

platform/core/src/studies/services/wado/retrieveMetadata.js
emfol: This is the one using the `qidoRoot` instead of the `wadoRoot`

platform/viewer/public/config/default.js
emfol: Yes. I think the performance gain can only be noticed if the user has predominantly very large studies with many series. If the user only has studies with one or two series I'd advise to disable it as the performance will probably be worse since we'll be essentially replacing a single API call by two... :-/

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: Yes. But I already removed the `this.isReady` flag and added a `this.subscriptions` list which will take care of getting rid of the registered handlers if the component is unmounted.

platform/core/src/studies/services/wado/retrieveMetadata.js
emfol: I can throw an exception instead...

platform/core/src/studies/services/wado/seriesLoadQueue.js
emfol: As the name implies this is not a regular/simple queue but a `series load queue` instead which from my perspective implies that dequeueing its items would effectively load the queued series...
But no strong opinion on how I should name it. Totally open to suggestions.

emfol:
What should I use instead of Pub/Sub?

platform/core/src/studies/services/wado/retrieveMetadata.js
dannyrb: Thanks for the call-out! I think I need to take a closer look at DICOMWebClient. I assumed the URLs exposed APIs that adhered to different contracts (requests + responses). It's interesting that you can provide it either URL and it works.

I'll do my own research before poking you on this one again ^_^

emfol:
> I left comments in #922 , as I didn't realise you made a new PR.

@JamesAPetts, Replying to your questions:

> I understand that pubSub must have been pulled in from the source of some external lib. But it feels like both this file and pubSub.js could be rewritten as ES6 classes and it'd be a lot easier to follow. I think this is warranted for such core code. But it does work in prototypical ES5 style.

So, I did not take those from an external library. I wrote them myself. The reason I did not make them regular ES6 classes is that I wanted them to be "pluggable functionalities" instead of "fully fledged classes" which would constrain me from using them in regular objects/dictionaries and force me to use them only through extension (building class hierarchies).

But your question actually brought me another idea. I see how I can make them ES6 classes and still support plugging them into regular objects/dictionaries.

emfol:
As discussed with Danny, I removed the PubSub. I also simplified the approach to load remaining series.

emfol:
Last commit enables study lazy load for deploy preview and demo config files.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -66,51 +59,49 @@ class ViewerRetrieveStudyData extends Component {
...
+      seriesMetadata
+    );
+    study.displaySets = studyMetadata.getDisplaySets();
+    updateMetaDataManager(study, series.seriesInstanceUid);
+    this.setState(function(state) {
+      return { studies: state.studies.slice() };
+    });
+  }
+
+  _attemptToLoadRemainingSeries(studyMetadata) {
```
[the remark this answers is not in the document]
emfol: Nice! It was Danny's idea! :-)
Thanks for the feedback @JamesAPetts

emfol:
Added a fix for the segmentation issue (sort result of the initial QIDO request moving SEG series to the end of the queue) and merged my changes to StudyMetadata with the changes @swederik did. @swederik I used a single sorting criteria function instead of two and added additional modalities to the comparison (not only SEG but KO and PR as well since all of them have references to series in the same exam even though we still do not support that)
@dannyrb Let me know if you're OK with the changes related to the removal of pub/sub approach

emfol:
@JamesAPetts @swederik @dannyrb I just pushed the changes to add a loading indicator for image thumbnails. Please let me know if I missed anything.

swederik:
Thanks @emfol! Awesome work, I think we are safe to merge this now!

Issue #758, opened 2019-08-09
 functionality Save to Archive


kusiroll:
How do I implement the  functionality Save to Archive,  I am using the meteor version 1.x and would like to implement the save to archive. I can see it has been implemented as seen in the image below but the version i am currently running does not have this functionality. Can someone help me with this and exactly what packages to upgrade.

![image](https://user-images.githubusercontent.com/35647493/62751997-58f01580-ba2b-11e9-8f18-92281ef4d036.png)

![95636A4688AA4EB9A7602499F0CFC106](https://user-images.githubusercontent.com/35647493/62751842-d6ffec80-ba2a-11e9-804d-07e96ca995bb.png)

dannyrb:
There is no easy way to upgrade from v1 to v2. For most, it will be a matter of adopting v2 entirely, and then porting any "secret sauce" you added for your v1 fork to work for/with version 2's extensions.

"Save to Archive" is not currently supported, but we would be willing to review and merge PRs that add support. A solution would likely need to use DICOMWeb's STOW-RS to send DICOM Structured Reports back.

Depending on your needs, we could add the "Save to Archive" button, and provide a hook that receives application data when pressed, and you can then persist it with your mechanism of choice. This would also be possible with a custom Panel Extension.

Related:
- https://github.com/OHIF/Viewers/issues/68
- https://github.com/OHIF/Viewers/issues/591
- https://github.com/OHIF/Viewers/issues/729

kusiroll:
Thanks, @dannyrb was this implemented in version 1 meteor? if yes where can I find it?

dannyrb:
I actually joined the project right around the time v1 was deprecated, so I'm not positive. I "think" this is where the logic lives:

https://github.com/OHIF/Viewers/blob/v1.x/Packages/ohif-measurement-table/client/configuration/dataExchange.js#L15-L30

Note: This same repository and in the `v1.x` branch

gmer16:
Hi guys, 

congratulations for the work done so far with v2, this is already an amazing piece of software.
I have a question regarding the "Save to Archive" functionality.
I've been doing some work to try to migrate some of the SR ROI functionality from v1 to v2.
I've had some success with the Length annotations (I'm able to save the SR to Orthanc), even if there are still some details that need fixing. I've tried to do the same with the Polyline annotations, that I previously added to the viewer. 
[...]

PR #954, opened 2019-09-26
feat: 🎸 DICOM SR STOW on MeasurementAPI

galelis:
### User cases
As an user: 
- I should be able to add measurements and store as DICOM Structured Report
- I should see a visual feedback after saving with success
- I should see a visual feedback after saving with error (e.g. saving without any measurement)
- I should be able to retrieve saved measurements from DICOM Structured Report
- I should be able to see the retrieved measurements into the measurement table
- I should be able to see the retrieved measurements drawn into the viewport
- I should be able to see warnings on measurements of tools that are not supported by DICOMSR
- I should be able to refresh the page and get all measurements saved restored
- If I am not using a server type DICOMWeb, I shoulod not be able to see the save button neither retrieve measurements saved on DICOM Structured Report

### Limitations
- As an user, I can only save/restore measurements of LenghTool as a limitation on DCMJS, all other tools are going to get a warning into the measurement table
- As an user, I can not save an empty list of measurements, as a limitation on DCMJS, I should see a error feedback
- As an user, I can only use this feature if my server is a DICOMWeb server. 

### Screenshot
OHIF Viewer (1)

Closes: #758


platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
```diff
@@ -344,6 +344,21 @@ const mergeProps = (propsFromState, propsFromDispatch, ownProps) => {
         },
       });
     },
+    saveToSR: () => {
+      const successHandler = () => {
+        alert('Success');
```
swederik: @dannyrb We should consider adding some generic notification toast which can be called for stuff like this

platform/viewer/src/lib/DICOMSR/dataExchange.js
```diff
@@ -0,0 +1,33 @@
...
+  if (!latestSeries) return Promise.resolve({});
+
+  return retrieveMeasurementFromSR(latestSeries);
+};
+
+export const storeMeasurements = (measurementData, server) => {
+  console.info('[DICOMSR] storeMeasurements');
+
+  if (!server || server.type !== 'dicomWeb') {
+    return Promise.resolve({});
```
swederik: We should throw an error here instead

platform/viewer/src/lib/DICOMSR/handleSR.js
```diff
@@ -0,0 +1,63 @@
+import * as dcmjs from 'dcmjs';
+import OHIF from '@ohif/core';
+import retrieveDataFromSR from './retrieveDataFromSR';
+import retrieveDataFromMeasurements from './retrieveDataFromMeasurements';
+
+import { api } from 'dicomweb-client';
+
+const retrieveMeasurementFromSR = async series => {
```
swederik: Can you talk to @ladeirarodolfo about using the loader service he added?

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
dannyrb: https://github.com/OHIF/Viewers/issues/985

tiimoS:
Generally seems to work. I observed three obstacles though: 

1. Deleting the last remaining measurement causes errors
2. Not all measurement types are being supported (by dcmjs library)
3. Seems to generate a new SR for each measurement instead of one SR for all measurements

**1. Deleting the last remaining measurement causes errors**
There seem to be some problems when deleting the only remaining measurement using the context menu (right mouse click on measurement). 

**2. Not all measurement types are being supported (by dcmjs library)**
Furthermore, as already mentioned by #988 it does not support saving measurements such as ArrowAnnotate or angle but throws an error: 
`Unhandled Promise Rejection: Error: TypeError: undefined is not an object (evaluating 'toolClass.getTID300RepresentationArguments')`
This is mainly due to the fact that the dcmjs library only supports Length, Polyline and Bidirectional.   Is there a way to edit the dcmjs library to also accept other types of measurements. I would like to prevent adding a database to store some measurements there, and others using the SR documents.  It would be much better to deal with them in a unified manner. 

**3. Seems to generate a new SR for each measurement instead of one SR for all measurements**
During my tests I have added quite some measurements. Each measurement added seems to have generated a separate SR document. These are not being deleted when you remove the measurements. Hence, I end up with a series consisting of about 200 SR documents (which I cannot delete in the client itself). 

Currently, point 2. seems to the be the most important issue to be addressed. I would be happy to support you.

dannyrb:
@tiimoS, number two is interesting. I'm sure we intend to eventually cover the tools from cornerstone tool's we include, but I'm not sure of the progress, or how to extend `dcmjs` to support other tools. @JamesAPetts would probably be the best person to weigh in at this time.

JamesAPetts:
> @JamesAPetts would probably be the best person to weigh in at this time.

Its simply an issue of adding definitions for how we translate the current measurements to/from SR. The SR implementation in dcmjs is certainly not exhaustive yet, but as Danny says, we will make sure to cover the core tools in due course in dcmjs.

>  I would be happy to support you.

If you want to help the effort by working on support for any other cornerstone SR adapters in dcmjs I'll be happy to review a PR!

platform/viewer/src/lib/DICOMSR/dataExchange.js
```diff
@@ -0,0 +1,33 @@
...
+  console.info('[DICOMSR] retrieveMeasurements');
+
+  const latestSeries = getLatestSRSeries();
+
+  if (!latestSeries) return Promise.resolve({});
+
+  return retrieveMeasurementFromSR(latestSeries);
+};
+
+export const storeMeasurements = (measurementData, server) => {
```
cmcgee-mac: When this is called with this patch, I'm finding that the server parameter here is really an object with timepointIds in it. The if check on the server type below always seems to fail and never attempt the stowSRFromMeasurements() below.

platform/viewer/src/lib/DICOMSR/retrieveDataFromMeasurements.js
```diff
@@ -0,0 +1,34 @@
+import * as dcmjs from 'dcmjs';
+import OHIF from '@ohif/core';
+
+const retrieveDataFromMeasurements = measurements => {
+  const { MeasurementReport } = dcmjs.adapters.Cornerstone;
+  const { getImageIdForImagePath } = OHIF.viewerbase;
```
cmcgee-mac: OHIF.viewerbase doesn't seem to have this function in my environment. Maybe it should be OHIF.measurements instead?

platform/core/src/measurements/tools/length.js
```diff
@@ -1,7 +1,7 @@
 const displayFunction = data => {
   let lengthValue = '';
   if (data.length) {
-    lengthValue = data.length.toFixed(2) + ' mm';
+    lengthValue = Number(data.length).toFixed(2) + ' mm';
```
cmcgee-mac: Unfortunately, the number conversion here isn't sufficient since it breaks in other places in the cornerstone-tools if it is left as a string. The reason this is here seems to be because the dcmjs parser is returning back a string for what should be a javascript/json number. Decimal Strings are most logically translated as numbers. The DICOM JSON data type mapping does just this. http://dicom.nema.org/medical/dicom/current/output/html/part18.html#table_F.2.3-1

Here is where the string value is parsed out of the DICOM in the dcmjs library: https://github.com/dcmjs-org/dcmjs/blob/2037c76b65b618e1fde57a50eb8fc94a99126dd1/src/ValueRepresentation.js#L436

We could work around in a slightly better way by converting all of the erroneous decimal strings to numbers when the data is read from the SR before it becomes tool state. Do you think that the dcmjs library might fix this problem since it is producing bad tool state data?

cmcgee-mac: I found a GH issue for this in the dcmjs project: https://github.com/dcmjs-org/dcmjs/issues/53

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,71 @@
...
+
+const retrieveDataFromSR = Part10SRArrayBuffer => {
+  const allDisplaySets = getAllDisplaySets();
+
+  // Get the dicom data as an Object
+  const dicomData = dcmjs.data.DicomMessage.readFile(Part10SRArrayBuffer);
+  const dataset = dcmjs.data.DicomMetaDictionary.naturalizeDataset(
+    dicomData.dict
+  );
+
```
cmcgee-mac: Here is where I managed to work around the string/number mapping problem in dcmjs like this:

```
+  // This is a hack for the problem where numeric values stored as DS (Decimal String)
+  //  are being parsed as strings instead of numbers. See https://github.com/dcmjs-org/dcmjs/issues/53
+  //  for more details.
+  function fixNumericValues(o) {
+    Object.keys(o).forEach(function (k) {
+        if (o[k] !== null && typeof o[k] === 'object') {
+            fixNumericValues(o[k]);
+        }
+
+        if (k == "NumericValue" && o[k] !== null && typeof o[k] === 'string') {
+            o[k] = Number(o[k]);
+        }
+    });
+  }
+  fixNumericValues(dataset);
```

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
galelis: We should use a notification toast, but in order to use it here, in an extension, we need to come up with a better approach on how to expose the notification into external extensions. 
Danny and I are working on that in a separeted branch, once we finish it, I can update this.

dannyrb: Definitely not a blocker for this PR

platform/core/src/measurements/tools/length.js
galelis: Yeah, good catch @cmcgee-mac, 
Steve just opened a PR in order to fix it on dcmjs
https://github.com/dcmjs-org/dcmjs/pull/79

platform/viewer/src/lib/DICOMSR/dataExchange.js
galelis: Yes, this server here is not right, its actually a filter that we are not using right now. 
Forget to remove.

platform/viewer/src/lib/DICOMSR/retrieveDataFromMeasurements.js
galelis: You are correct. 
At the momment I've create the draft PR I haven't tested store functionality.

cmcgee-mac: Store is working pretty well. I've been trying it out. :)

platform/viewer/src/lib/DICOMSR/handleSR.js
```diff
@@ -0,0 +1,63 @@
...
+    studyInstanceUID: instance.getStudyInstanceUID(),
+    seriesInstanceUID: instance.getSeriesInstanceUID(),
+    sopInstanceUID: instance.getSOPInstanceUID(),
+  };
+
+  return dicomWeb.retrieveInstance(options).then(retrieveDataFromSR);
+};
+
+const stowSRFromMeasurements = async measurements => {
+  const dataset = retrieveDataFromMeasurements(measurements);
```
cmcgee-mac: In order to prevent repeating the same measurements previous series in subsequent series it might be best to filter out the measurements that have already been saved. I have this simple filter that seems to do the job in my testing. It might be the wrong place for this though and there might be a better way to check the origin/dirty state of the measurement.

```
  // Exclude the measurements that are already stored
  // TODO detect whether those measurements were modified
  measurements.allTools = measurements.allTools.filter( (m) => { if (m.imageId) { return false;} return true; });
```

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,71 @@
...
+};
+
+const imagingMeasurementsToMeasurementData = (dataset, displaySets) => {
+  const { MeasurementReport } = dcmjs.adapters.Cornerstone;
+  const storedMeasurementByToolType = MeasurementReport.generateToolState(
+    dataset
+  );
+  const measurementData = {};
+  let measurementNumber = 0;
+
```
cmcgee-mac: This could be a good place to allocate different colors for each series and then you can distinguish between measurements made at different times and the ones in the current session. Once a color is chosen it can be set like ```color: "someCssColor"``` around line 52 below in the toolData.

platform/viewer/src/lib/DICOMSR/dataExchange.js
```diff
@@ -0,0 +1,33 @@
+import OHIF from '@ohif/core';
+import { retrieveMeasurementFromSR, stowSRFromMeasurements } from './handleSR';
+import { getLatestSRSeries } from './srUtils';
+
+export const retrieveMeasurements = () => {
+  console.info('[DICOMSR] retrieveMeasurements');
+
+  const latestSeries = getLatestSRSeries();
```
cmcgee-mac: Getting the latest SR series here causes some series to be skipped while others have their measurements duplicated depending on the random sequence that the series are loaded into the viewer. This happens when there are multiple SR's in the study, potentially from multiple saves of the measurements.

platform/viewer/src/lib/DICOMSR/handleSR.js
galelis: We actually should be picking up only the latest SR series into the study, that should contain the latest state of saved measurements. Somehow, the series are being full filled 1 by 1, so in this case, everytime this code runs, it goes thourgh all SRs and store it again in the viewer. 
I am currently working on this.

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
galelis: You are right. Althought I've talked with Erik and for now we want to have only one SR being used, the latest one.

platform/viewer/src/lib/DICOMSR/dataExchange.js
galelis: When using lazyLoading for the series, we were calling this retrieve measurements several times, because of that, we were loading all SRs. 
I've change the way we call the retrieveMeasrurements method and how we handle the async request of series with lazy load. So now we only call this once we get all series loaded, then we can get the latest one as disired by our team. 

It should be working now.

cmcgee-mac: I wonder if it might be worth getting the latest SR that's already loaded from the viewer rather than fetching them again. See the presentation state code here for the componentDidUpdate(). Maybe this case is different than PR or perhaps there is a problem with what I did there.

https://github.com/OHIF/Viewers/pull/1044/files#diff-acf605965e02491d822a993ef89426abR211

platform/viewer/src/lib/DICOMSR/dataExchange.js
galelis: You are right @swederik 
I've added an error here and the code is now preventing showing the save button in case the server is not dicomWeb type.

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
galelis: I've changed the measurementTable component to start using Snackbar for saving feedbacks.

platform/viewer/src/lib/DICOMSR/dataExchange.js
galelis: So, the viewer will load (completely or by lazy loading) all series metaData, here I will need the instance data, which will required this new request.

cmcgee-mac: Thanks, so the intent is that the viewer only loads the series metadata, not instance-level information? If so, I'll need to adjust the presentation state PR to match.

extensions/cornerstone/src/init.js
```diff
@@ -78,6 +78,16 @@ export default function init(configuration = {}) {
   ];
 
   tools.forEach(tool => csTools.addTool(tool));
+
+  const BaseAnnotationTool = csTools.importInternal('base/BaseAnnotationTool');
+  tools.forEach(tool => {
+    if (tool.prototype instanceof BaseAnnotationTool) {
+      // BaseAnnotationTool would likely come from csTools lib exports
+      const toolName = new tool().name;
```
JamesAPetts: I get your concern that it feels a bit wonky.
The rationale that they are not static properites of the class is that one may make multiple tools by just passing different configuration. 

The default tool names are just the classnames with the suffix removed `Tool`, so you can do an `Object.keys` to get the name, and `split` off the Tool.

JamesAPetts: Note: Perhaps it'd be good in CST if we expose a default name explictly on the class, and then use this in the constructor if the default isn't changed. Its all implicit right now.

platform/ui/src/components/measurementTable/MeasurementTable.js
```diff
@@ -75,10 +79,35 @@ class MeasurementTable extends Component {
...
+  saveFunction = async event => {
+    const { saveFunction, snackbarContext } = this.props;
+    if (saveFunction) {
+      try {
+        await saveFunction();
+        snackbarContext.show({
+          title: 'STOW SR',
+          message: 'Measurements were saved with success',
+        });
+      } catch (error) {
```
JamesAPetts: We need to log this error. Its getting lost. I got an Error whilst saving measurements whilst testing on the deploy and couldn't debug it without pulling locally.

dannyrb:
uASvRQkAxP

Steps to recreate:

- Draw X annotations on image
- Expand Measurements Panel
- Save
- Full page refresh
- Only first measurement shows (no value)
- Interacting with the measurements adds the value, and next saved measurements is rendered
- Repeat until all show

image

extensions/cornerstone/src/init.js
dannyrb: Heh. Gustavo and I worked together on this bit. At the time, I said it was "good enough", but there's a better solution out there. Cornerstone Tools tracks `GlobalTools` and `tools`. If we updated the "History Tracking" changes we discussed earlier this year (RE: @JamesAPetts), then you could just iterate the list of `GlobalTools` and use the names defined there instead of using this hack.

(Which would fail if any tool specified a custom name when initializing, this just grabs the default name.)

JamesAPetts: > this just grabs the default name,

So right now we can do get the default name also by taking the class and removing `Tool` from the class name. That still feels better than instantiating a tool for no reason, no?

dannyrb: For CornerstoneTools, that's fine. I'm at a bit of a loss as we open up the configuration to allow registering/adding custom tools.

JamesAPetts: Eh, I guess you are right. This is "better" for now.
The upgrade required for CST is super trivial though, just need to go through each core tool and update the default name in the constructor to be taken from a static `BlahTool.DEFAULT_TOOLNAME` prop. , could lump it as part of this PR's work. Then third party tools can do the same.

platform/viewer/src/lib/DICOMSR/dataExchange.js
galelis: So, as far as I understood, we do get all study/series metaData and store it inside the `studyMetadataManager`. The instances we only fetch them when we add the series into a viewport, which will have a cache based on imageId and thats a separated thing.

extensions/vtk/package.json
```diff
@@ -56,7 +56,6 @@
     "@ohif/ui": "^0.58.4",
     "cornerstone-tools": "^4.0.9",
     "cornerstone-wado-image-loader": "^3.0.0",
-    "dcmjs": "^0.6.1",
```
galelis: @JamesAPetts can you confirm if this line is really necessary? I have removed, but I can come back in case its needed.

JamesAPetts: Yeah that doesn't need to be in dev dependencies 👍 .

platform/ui/src/components/measurementTable/MeasurementTable.js
galelis: This should be better now, I can logging the error.

extensions/cornerstone/src/init.js
galelis: The issue reported in this comment related with `data.length.toFixed is not a function` should be fixed by the latest version of `dcmjs`. Thanks for helping me on that @JamesAPetts 

The other issue reported, related with the way we get the tool name, I've spoke with Danny and this is something that we are going to change in the future, in order to catch the actual name that each tool was instantiated with and not the default name in the tool. In the near future, this changes would be unacessary in our point of view. 
We can discuss this later as well.

dannyrb:
@galelis can you resolve conflicts? Are you waiting on updated reviews from @JamesAPetts and myself?

JamesAPetts:
Happy to review this one conflicts are resolved.

extensions/cornerstone/src/init.js
JamesAPetts: Yeah this certainly isn't blocking for now, but something we can clean up in the future. We do only do this once, so its not a huge overhead.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -55,17 +56,28 @@ class ViewerRetrieveStudyData extends Component {
...
         studyMetadata.setDisplaySets(study.displaySets);
+
         // Updates WADO-RS metaDataManager
         updateMetaDataManager(study);
         studyMetadataManager.add(studyMetadata);
+
         // Attempt to load remaning series if any
-        this._attemptToLoadRemainingSeries(studyMetadata);
+        this._loadRemainingSeries(studyMetadata).then(
+          this._studyDidLoad.bind(this)
```
JamesAPetts: Its a little cleaner to bind this in the constructor:

`this._studyDidLoad = this._studyDidLoad.bind(this);`

Then you can just call `this._studyDidLoad()` in the class methods.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -55,17 +56,28 @@ class ViewerRetrieveStudyData extends Component {
...
+
         return study;
       });
       this.setState({ studies });
     }
   }
 
+  _studyDidLoad() {
+    this.setState({
+      seriesLoaded: true,
```
JamesAPetts: This should be `studyLoaded`? Its for multiple series (and the same wherever else its used).

galelis: Yes, it can be studyLoaded. 
I've added seriesLoaded just because I was waiting all series promises.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
galelis: Good tip. Just updated the code.

galelis:
> This looks good to me. Other than my minor comments, my only last request is that if a user tries to save any measurements which currently aren't supported by dcmjs, a cleaner error is thrown. Something like "XYZ measurements aren't supported at this time". Also the presence of an ellipse (for example) anywhere in the toolState shouldn't prevent the saving of supported measurements (i.e. just length at this time).

I agree with you, but this would be a dcmJS feedback. Mostly beucase only dcmJS knows which tool it supports, same for saving an empty SR, it breaks and does not give back a good feedback about it. 

We can create issues/feature-requests on dcmJS
What you think?

pieper:
>We can create issues/feature-requests on dcmJS
What you think?

Can you suggest a better API for dcmjs could declare which conversion types are supported?  Currently you could do `toolName in dcmjs.adapters.Cornerstone` and know whether or not it's supported.  Or are you saying something lower level breaks?

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
JamesAPetts: I'm fine with `studyLoaded` or `allSeriesLoaded`, it was just a little confusing  as series it was a bit ambiguous as to whether `seriesLoaded` was singular or plural.

Since you already changed it, if you are happy with `studyLoaded` we can go with that.

JamesAPetts:
> I agree with you, but this would be a dcmJS feedback. Mostly becuause only dcmJS knows which tool it supports

Either way we go about this, I don't think we can merge this until there are either useful error messages, or only the lengths are consumed.

There are a few ways we could go about this at this time:
- Filter the toolState so we are only passing the length measurements to dcmjs for now, expand this filter as we get more support in dcmjs. (quick)
- Add a filter to dcmjs. (better)

At some level we need to do filtering anyway, as we could always have plugin anotation tools that could have arbitrary toolData shape. It makes sense to do it in dcmjs if you'd like to spike it.

I think you'll just need to [return here](https://github.com/dcmjs-org/dcmjs/blob/1e10ca74c284e0679189fb87dd872455dcc3aaa5/src/adapters/Cornerstone/MeasurementReport.js#L27) if toolClass is undefined. You shouldn't have to change too much. Ping Steve or I when you want a review and I'm sure one of us will be able to check and merge it.

> same for saving an empty SR, it breaks and does not give back a good feedback about it.

If you [look here](https://github.com/dcmjs-org/dcmjs/blob/1e10ca74c284e0679189fb87dd872455dcc3aaa5/src/adapters/Cornerstone/MeasurementReport.js#L55), this throws an error message : `"No measurements provided."`.

So you can `catch` this message and use the error message to populate the UI:

```
// Example:


try {
  MeasurementReport.generateReport(toolState, cornerstone.metaData);
 // Other code
} catch (error) {
 const errorMessage = error.message; 
 // Do something with this message.
}
```

dannyrb:
@galelis, do you have clear enough direction from James on this? What can we do to help push this PR forward?

galelis:
> @galelis, do you have clear enough direction from James on this? What can we do to help push this PR forward?

Yes Danny, working to get the errors better logged and filtering the tools that we support on dcmjs.

galelis:
I have updated the error treatment. 
- handle no measurements provided
- show a different message when we have unsupported tools in the measurement table. 
- log in the console a more detailed statuses

platform/viewer/src/lib/DICOMSR/dataExchange.js
```diff
@@ -0,0 +1,49 @@
...
+    const { notSupportedTools } = await stowSRFromMeasurements(
+      measurementData,
+      server
+    );
+    if (studyInstanceUid) {
+      OHIF.studies.deleteStudyMetadataPromise(studyInstanceUid);
+    }
+    const message =
+      notSupportedTools.length > 0
+        ? 'Measurements were parcially saved, some of the tools are not supported.'
```
JamesAPetts: @dannyrb , do you think its worth writing which tools aren't supported in the snackbar here? The list would currently be quite long and I'm not sure on the UX. Right now the user has no real information as to what was saved. Its a stop-gap until we have support for all core tools, but we will always need something like this for unsupported 3rd party annotation tools anyway.

They are currently logged to the console for devs, but I don't know how much we should expose to the user.

platform/viewer/src/lib/DICOMSR/dataExchange.js
```diff
@@ -13,23 +13,37 @@ export const retrieveMeasurements = options => {
...
+  const firstMeasurement = measurementData[firstMeasurementKey][0];
   const studyInstanceUid =
-    measurementData[Object.keys(measurementData)[0]][0].studyInstanceUid;
-
-  return stowSRFromMeasurements(measurementData, server).then(
-    () => {
+    firstMeasurement && firstMeasurement.studyInstanceUid;
+
+  try {
+    const { notSupportedTools } = await stowSRFromMeasurements(
```
JamesAPetts: `notSupportedTools` -> `unsupportedTools`

platform/viewer/src/lib/DICOMSR/dataExchange.js
dannyrb: Partially is spelled wrong.

Instead of checking measurements at time of save, can we highlight them in the measurements table or use the "warning" styling with a note at the top of the table that tools that look like X cannot be persisted at this time?

Notification could remain simple. Annotations saved/persisted successfully.

platform/viewer/src/lib/DICOMSR/dataExchange.js
galelis: liked it. Code is updated.

platform/viewer/src/lib/DICOMSR/dataExchange.js
galelis: I´ve reverted the snackbar messages, back to error/success and added the warings on measurements that can not be persisted at this point.

galelis: Screen Shot 2019-11-18 at 03 11 42

JamesAPetts: I like this, looks a lot cleaner. @dannyrb does this PR get your green flag?

JamesAPetts: Its minor, but it would be nice if the `Save measurements`button had some on hover/on click animation.

galelis: I've forget to change that. 
Just updated the cursor and hover/action style.

platform/core/src/measurements/classes/MeasurementApi.js
```diff
@@ -1,4 +1,5 @@
 import cornerstoneTools from 'cornerstone-tools';
+import cornerstone from 'cornerstone-core';
```
dannyrb: Long-term, we don't want the `MeasurementApi` to depend on `cornerstone` or `cornerstone-core`, but that's outside the scope of this ticket. I'll create a follow-up issue for the backlog before we merge this PR.

platform/core/src/measurements/classes/MeasurementApi.js
```diff
@@ -203,13 +204,14 @@ export default class MeasurementApi {
 
   retrieveMeasurements(patientId, timepointIds) {
     const retrievalFn = configuration.dataExchange.retrieve;
+    const { server } = this.options;
```
dannyrb: Can you add JSDoc to `initialize`? I'm not sure what this method expects for options, and if it's using `server`, what properties `server` needs to have for it to be a valid option

platform/core/src/measurements/classes/MeasurementApi.js
```diff
@@ -203,13 +204,14 @@ export default class MeasurementApi {
...
     if (typeof retrievalFn !== 'function') {
       log.error('Measurement retrieval function has not been configured.');
       return;
     }
 
     return new Promise((resolve, reject) => {
-      retrievalFn(patientId, timepointIds).then(measurementData => {
+      retrievalFn({ patientId, timepointIds, server }).then(measurementData => {
```
dannyrb: The `...` that sets this in configuration/dataExchange made it difficult to track down where `retrievalFn` is defined. Outside the scope of this PR -- just making a note for follow-up issue(s).

platform/ui/src/components/measurementTable/MeasurementTable.js
```diff
@@ -75,10 +78,36 @@ class MeasurementTable extends Component {
...
+        const result = await saveFunction();
+        snackbarContext.show({
+          title: 'STOW SR',
+          message: result.message,
+        });
+      } catch (error) {
+        snackbarContext.show({
+          title: 'STOW SR',
+          message: error.message,
+        });
```
dannyrb: Instead of tightly coupling this component to `snackbar`, it should have callbacks for `onError` and `onSuccess` (or something similar. Then, wherever we're using this in `viewer`, should provide the implementation to those callback props. This is a blocker, imo

platform/ui/src/components/measurementTable/MeasurementTable.styl
```diff
@@ -1,92 +1,116 @@
...
+
+    .saveBtn
+      border: 1px solid var(--ui-border-color-active)
+      border-radius: 25px
+      background-color: var(--active-color)
+      color: var(--ui-gray-dark)
+      line-height: 25px
+      padding: 0 20px
+      outline: none
+      cursor: pointer
```
dannyrb: We really need a good reusable button component, haha.
I feel like every component with a button has it's own styles.

Not a blocker.

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
```diff
@@ -3,6 +3,7 @@ import { MeasurementTable } from '@ohif/ui';
 import OHIF from '@ohif/core';
 import moment from 'moment';
 import cornerstone from 'cornerstone-core';
+import DICOMSR from './../../lib/DICOMSR';
```
dannyrb: Is there a good reason this can't live in `@ohif/core`?

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
```diff
@@ -58,6 +59,19 @@ function getDataForEachMeasurementNumber(
   return data;
 }
 
+function getWarningsForMeasurement(meansurementData) {
+  const { toolType } = meansurementData;
```
dannyrb: Should `meansurementData` be `measurementData`?
Also, why pass in the entire object if we only use `toolType`?
Is `toolType` a string? Can we add JSDoc?

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
```diff
@@ -142,8 +162,30 @@ function convertTimepointsToTableData(timepoints) {
   ];
 }
 
+function updateWarnings(measurementCollection) {
+  measurementCollection;
+}
```
dannyrb: I'm not sure I understand what this function is doing

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
```diff
@@ -142,8 +162,30 @@ function convertTimepointsToTableData(timepoints) {
...
+
+function getSaveFunction(servers) {
+  const activeServer = servers.servers.find(a => a.active === true);
+  let saveFunction = undefined;
+
+  if (activeServer.type === 'dicomWeb') {
+    saveFunction = () => {
+      const measurementApi = OHIF.measurements.MeasurementApi.Instance;
```
dannyrb: That's a little wonky. Added to my "we should clean this up" list. Not blocking.

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
```diff
@@ -142,8 +162,30 @@ function convertTimepointsToTableData(timepoints) {
   ];
 }
 
+function updateWarnings(measurementCollection) {
+  measurementCollection;
+}
+
+function getSaveFunction(servers) {
```
dannyrb: JSDoc please 🙏

platform/viewer/src/connectedComponents/ConnectedViewer.js
```diff
@@ -1,13 +1,20 @@
...
+import Viewer from './Viewer.js';
+import OHIF from '@ohif/core';
 
 const { setTimepoints, setMeasurements } = OHIF.redux.actions;
 
-const mapStateToProps = (state, ownProps) => {
+const getActiveServer = servers => {
+  const isActive = a => a.active === true;
+  return servers.servers.find(isActive);
+};
```
dannyrb: Servers should now be exposed via `AppContext`. I wonder if it's worth having a `getActiveServer` method for it, or leveraging an equivalent selector 🤔 (not blocking)

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -58,20 +59,22 @@ class Viewer extends Component {
   static propTypes = {
     studies: PropTypes.array,
     studyInstanceUids: PropTypes.array,
+    activeServer: PropTypes.object,
```
dannyrb: Can we do a `PropTypes.shape({ })`, with our expected fields spelled out more explicitly?

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -58,20 +59,22 @@ class Viewer extends Component {
...
     onTimepointsUpdated: PropTypes.func,
     onMeasurementsUpdated: PropTypes.func,
     // window.store.getState().viewports.viewportSpecificData
     viewports: PropTypes.object.isRequired,
     // window.store.getState().viewports.activeViewportIndex
     activeViewportIndex: PropTypes.number.isRequired,
+    studyLoaded: PropTypes.bool,
```
dannyrb: Can we do `has` or `isStudyLoaded`?

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -182,6 +173,7 @@ class Viewer extends Component {
 
     const measurementApi = new MeasurementApi(timepointApi, {
       onMeasurementsUpdated: this.onMeasurementsUpdated,
+      server: activeServer,
```
dannyrb: 🤔 I wonder if we'll ever want Measurements to persist despite returning to the study list

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -192,26 +184,30 @@ class Viewer extends Component {
       const patientId = studies[0] && studies[0].patientId;
 
       timepointApi.retrieveTimepoints({ patientId });
-      measurementApi.retrieveMeasurements(patientId, [currentTimepointId]);
-
+      if (this.props.studyLoaded) {
+        this.measurementApi.retrieveMeasurements(patientId, [
+          currentTimepointId,
+        ]);
+      }
```
dannyrb: I wonder if this makes more sense to do with promises/events instead of in mounted and updated w/ a reactive bool.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -218,6 +219,7 @@ function ViewerRetrieveStudyData({
     );
 
     setStudies([...studies, study]);
+    setIsStudyLoaded(true);
```
dannyrb: I get it, but I think this can be cleaner. Probably not blocking, but worth thinking about.

platform/viewer/src/lib/DICOMSR/dataExchange.js
```diff
@@ -0,0 +1,42 @@
+import OHIF from '@ohif/core';
+import { retrieveMeasurementFromSR, stowSRFromMeasurements } from './handleSR';
+import { getLatestSRSeries } from './srUtils';
+
+export const retrieveMeasurements = options => {
```
dannyrb: JS Doc?

platform/viewer/src/lib/DICOMSR/dataExchange.js
```diff
@@ -0,0 +1,42 @@
...
+
+  const { server } = options;
+  const latestSeries = getLatestSRSeries();
+
+  if (!latestSeries) return Promise.resolve({});
+
+  return retrieveMeasurementFromSR(latestSeries, server);
+};
+
+export const storeMeasurements = async (measurementData, filter, server) => {
```
dannyrb: JS Doc?

platform/viewer/src/lib/DICOMSR/dataExchange.js
```diff
@@ -0,0 +1,42 @@
+import OHIF from '@ohif/core';
```
dannyrb: Everything related to DICOMSR is pretty critical. It would be nice to see unit tests either before merge, or as an immediate next task.

platform/viewer/src/lib/DICOMSR/handleSR.js
```diff
@@ -0,0 +1,72 @@
...
+import OHIF from '@ohif/core';
+import retrieveDataFromSR from './retrieveDataFromSR';
+import retrieveDataFromMeasurements from './retrieveDataFromMeasurements';
+
+import { api } from 'dicomweb-client';
+
+const VERSION_NAME = 'dcmjs-0.0';
+const TRANSFER_SYNTAX_UID = '1.2.840.10008.1.2.1';
+
+const retrieveMeasurementFromSR = async (series, server) => {
```
dannyrb: JSDOC?

platform/viewer/src/lib/DICOMSR/handleSR.js
```diff
@@ -0,0 +1,72 @@
...
+  const options = {
+    studyInstanceUID: instance.getStudyInstanceUID(),
+    seriesInstanceUID: instance.getSeriesInstanceUID(),
+    sopInstanceUID: instance.getSOPInstanceUID(),
+  };
+
+  return dicomWeb.retrieveInstance(options).then(retrieveDataFromSR);
+};
+
+const stowSRFromMeasurements = async (measurements, server) => {
```
dannyrb: JSDoc?

platform/viewer/src/lib/DICOMSR/index.js
```diff
@@ -0,0 +1,10 @@
+import { retrieveMeasurements, storeMeasurements } from './dataExchange';
+import { isToolSupported } from './isToolSupported';
+
+const DICOMSR = {
+  retrieveMeasurements,
+  storeMeasurements,
+  isToolSupported,
+};
+
+export default DICOMSR;
```
dannyrb: This seems like a solid candidate for `@ohif/core`

platform/viewer/src/lib/DICOMSR/isToolSupported.js
```diff
@@ -0,0 +1,8 @@
+import * as dcmjs from 'dcmjs';
```
dannyrb: Does dcmjs allow us to scoped import `adapters`?

dannyrb: Does it automatically include all adapters, even if we're not using a library it may have an adapter for?

platform/viewer/src/lib/DICOMSR/retrieveDataFromMeasurements.js
```diff
@@ -0,0 +1,52 @@
+import * as dcmjs from 'dcmjs';
+import OHIF from '@ohif/core';
+import cornerstone from 'cornerstone-core';
+
+const retrieveDataFromMeasurements = measurements => {
```
dannyrb: What is the shape of measurements? Can we JSDoc this? (note: it supports a `@typedef`)
Any way to only import the portions of dcmjs we need instead of everything?

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,70 @@
+import * as dcmjs from 'dcmjs';
```
dannyrb: I'm not sure `retrieve` is the best word here.
And it may be a bit verbose, but expanding `SR` to `StructuredReport` may help out newcomers to DICOM-land

- `parseDicomStructuredReport()`, maybe?

platform/viewer/src/lib/DICOMSR/srUtils.js
```diff
@@ -0,0 +1,71 @@
+import OHIF from '@ohif/core';
```
dannyrb: Drop the `utils` name/file, and make each of these methods their own file w/ default export?

platform/viewer/src/lib/DICOMSR/srUtils.js
```diff
@@ -0,0 +1,71 @@
...
+  let allDisplaySets = [];
+
+  allStudies.forEach(study => {
+    if (study.getDisplaySets) {
+      allDisplaySets = allDisplaySets.concat(study.getDisplaySets());
+    }
+  });
+
+  return allDisplaySets;
+};
```
dannyrb: This isn't really an SR util. Should this be a part of the studyMetadataManager?

dannyrb: If not, can we pass in the `studyMetadataManager` instead of importing?

platform/viewer/src/lib/DICOMSR/srUtils.js
```diff
@@ -0,0 +1,71 @@
...
+
+const getInstanceMetadata = (displaySets, sopInstanceUid) => {
```
dannyrb: JSDoc?

dannyrb: Can we rename this to `findInstanceMetadataBySopInstanceUid`?

platform/viewer/src/lib/DICOMSR/srUtils.js
```diff
@@ -0,0 +1,71 @@
...
+  let instance;
+
+  // find the correct instance
+  displaySets.some(displaySet => {
```
dannyrb: is `some` faster than `find`?

platform/viewer/src/lib/DICOMSR/srUtils.js
```diff
@@ -0,0 +1,71 @@
...
+    // Search the display set to find the instance metadata for
+    return (
+      displaySet.images &&
+      displaySet.images.find(instanceMetadata => {
+        if (instanceMetadata._sopInstanceUID === sopInstanceUid) {
+          instance = instanceMetadata;
+
+          return true;
```
dannyrb: Using this as a kind of `break;` for `.some()` instead of using a loop?

platform/viewer/src/lib/DICOMSR/srUtils.js
```diff
@@ -0,0 +1,71 @@
...
+        }
+      })
+    );
+  });
+
+  return instance;
+};
+
+const getLatestSRSeries = () => {
```
dannyrb: `findMostRecentStructuredReport`?
Can we pass in `studyMetadataManager` instead of importing it?
Or if we know that this needs a studies array, can this thing just take a studies array? No need to pass it something so heavy when we need so little.

platform/viewer/src/lib/DICOMSR/srUtils.js
```diff
@@ -0,0 +1,71 @@
+import OHIF from '@ohif/core';
+
+const { studyMetadataManager } = OHIF.utils;
```
dannyrb: `studyMetadataManager` isn't really a "utility". It's a bit odd that we expose it this way. If any of these methods really need the whole metadataManager, can we pass it in as a parameter? If they don't, let's scope what we pass in to only the information each function needs.

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,70 @@
...
+import { getAllDisplaySets, getInstanceMetadata } from './srUtils';
```
dannyrb: Are these exposed outside of the DICOM SR logic? (are they public for consumers of `@ohif/core`?)

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,70 @@
...
+  const dicomData = dcmjs.data.DicomMessage.readFile(Part10SRArrayBuffer);
+  const dataset = dcmjs.data.DicomMetaDictionary.naturalizeDataset(
+    dicomData.dict
+  );
+
+  // Convert the SR into the kind of object the Measurements package is expecting
+  return imagingMeasurementsToMeasurementData(dataset, allDisplaySets);
+};
+
+const imagingMeasurementsToMeasurementData = (dataset, displaySets) => {
```
dannyrb: JSDoc?

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,70 @@
+import * as dcmjs from 'dcmjs';
+import { getAllDisplaySets, getInstanceMetadata } from './srUtils';
+
+const retrieveDataFromSR = Part10SRArrayBuffer => {
```
dannyrb: JSDoc?

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,70 @@
...
+  const allDisplaySets = getAllDisplaySets();
```
dannyrb: Could all displaySets be passed in?

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,70 @@
...
+        measurement.sopInstanceUid
+      );
+      const study = instanceMetadata._study;
+      const series = instanceMetadata._series;
+      const imagePath = [
+        study.studyInstanceUid,
+        series.seriesInstanceUid,
+        measurement.sopInstanceUid,
+        measurement.frameIndex,
+      ].join('_');
```
dannyrb: Might be a nice helper.

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,70 @@
...
+
+      const toolData = Object.assign({}, measurement, {
+        imageId,
+        imagePath,
+        seriesInstanceUid: series.seriesInstanceUid,
+        studyInstanceUid: study.studyInstanceUid,
+        patientId: study.patientId,
+        measurementNumber: ++measurementNumber,
+        timepointId: currentTimepointId,
+        toolType,
```
dannyrb: Should this be renamed to `toolName`? I think this is a holdover from cornerstone v2 naming.

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,70 @@
...
+
+  // Convert the SR into the kind of object the Measurements package is expecting
+  return imagingMeasurementsToMeasurementData(dataset, allDisplaySets);
+};
+
+const imagingMeasurementsToMeasurementData = (dataset, displaySets) => {
+  const { MeasurementReport } = dcmjs.adapters.Cornerstone;
+  const storedMeasurementByToolType = MeasurementReport.generateToolState(
+    dataset
+  );
```
dannyrb: Oh, interesting. Do we do anything to check that data in the SR maps to something our application can display? (not that this would be the ideal place to do that, just thinking out loud.) For example, how should we handle:

There is Length data in the structured report, but...

- No Length `cornerstone-tool` is "added" and "enabled"
- OR no Length `vtk.js` widget
- OR `cornerstone-tool`'s Length tool expects a different tool state "shape"
    - We modify it at a later date, or the user has a custom Length tool

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
```diff
@@ -0,0 +1,70 @@
...
+      if (!imageId) {
+        return;
+      }
+
+      // TODO: We need the currentTimepointID set into the viewer
+      const currentTimepointId = 'TimepointId';
+
+      const toolData = Object.assign({}, measurement, {
+        imageId,
+        imagePath,
```
dannyrb: Just curious, do we support annotations with points on multiple slices? (or placed within the volume)

platform/viewer/src/lib/DICOMSR/retrieveDataFromMeasurements.js
```diff
@@ -0,0 +1,52 @@
...
+    const report = MeasurementReport.generateReport(
+      toolState,
+      cornerstone.metaData
+    );
+    return {
+      dataset: report.dataset,
+      unsupportedTools,
+    };
+  } catch (error) {
+    throw error;
```
dannyrb: Does that make this a pointless try/catch?

platform/viewer/src/lib/DICOMSR/retrieveDataFromMeasurements.js
```diff
@@ -0,0 +1,52 @@
...
+  const { getImageIdForImagePath } = OHIF.measurements;
+
+  const toolState = {};
+  const unsupportedTools = [];
+
+  Object.keys(measurements).forEach(measurementType => {
+    const annotations = measurements[measurementType];
+
+    annotations.forEach(annotation => {
+      const { toolType, imagePath } = annotation;
```
dannyrb: Won't only measurements from existing SRs have an imagePath? Or is this something our MeasurementsApi has been tacking on and I've never noticed?

platform/viewer/src/lib/DICOMSR/retrieveDataFromMeasurements.js
```diff
@@ -0,0 +1,52 @@
...
+      if (cornerstoneAdpater[toolType]) {
```
dannyrb: It might be dangerous to assume a string name guarantees a match.

platform/core/src/measurements/classes/MeasurementApi.js
galelis: Totally agree with you Danny.

platform/ui/src/components/measurementTable/MeasurementTable.styl
galelis: Yeah, I tried to get similar style as those rouded buttons in the toolbar section, but had to clone the style here for now.

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
galelis: Ideally my idea was to make it work here as I can have easy access to the OHIF instance and later get it inside `@ohif/core`

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
galelis: Fixed the callouts. 
I think my idea was to get warnings for measurements not only related with the toolType, but end up only using the toolType. Just removed and fixed the callout.

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
galelis: Probably refactor leftover.  
Removed it.

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
galelis: 👍

platform/viewer/src/connectedComponents/Viewer.js
galelis: Sure, make sense. 
Code updated.

platform/viewer/src/lib/DICOMSR/retrieveDataFromMeasurements.js
galelis: I've added this because once we try to save an empty SR (no measurements) `generateReport` throws an error. Did this way, so I can grab the error and display it properly using (as of now) the snackbar.

dannyrb: That makes sense, but wouldn't the error already bubble? We're catching and rethrowing?

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
galelis: I believe we don't have this kind of support on neither cornerstoneTools or dcmjs.

platform/viewer/src/lib/DICOMSR/srUtils.js
galelis: I've update the code and remove part of the logic and make it more simple.

platform/viewer/src/lib/DICOMSR/srUtils.js
galelis: I've updated the logic.

platform/viewer/src/lib/DICOMSR/srUtils.js
galelis: Code is updated with suggested changes.

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
galelis: Not right now.

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
galelis: Added.

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
galelis: Added. :)

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
galelis: Changed 👍

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
galelis: Refactored. 👍

platform/viewer/src/lib/DICOMSR/retrieveDataFromMeasurements.js
galelis: You are right. 
Just fixed here and in another function that I saw the same happening. 

This might be here because I was trying to use the snackbar directly here. Should have removed it after refactor.

platform/viewer/src/lib/DICOMSR/retrieveDataFromMeasurements.js
galelis: We can think on something stronger, but for now, on dcmjs its the only way I thought. 

@JamesAPetts do you have any idea if we can check this in a better way? If a tool is supported by dcmjs?

platform/core/src/measurements/classes/MeasurementApi.js
```diff
@@ -203,13 +204,14 @@ export default class MeasurementApi {
 
   retrieveMeasurements(patientId, timepointIds) {
     const retrievalFn = configuration.dataExchange.retrieve;
+    const { server } = configuration;
```
dannyrb: I think this is a valid solution, but we need to indicate that we expect `server` to be defined.

Here: https://github.com/OHIF/Viewers/blob/f5657595a59a996a984ea5fc5ce54d21148ddc3c/platform/core/src/measurements/classes/MeasurementApi.js#L18-L20

Here: https://github.com/OHIF/Viewers/blob/f5657595a59a996a984ea5fc5ce54d21148ddc3c/platform/core/src/measurements/classes/MeasurementApi.js#L11-L13

The easiest way to do this is by:

1. Not spreading props
2. Adding JSDoc
3. Potentially warning or throwing if `server` is not set

You can skip number three if `server` is required to instantiate the `MeasurementApi`, and you have checks in the constructor.

dannyrb: That also means the shape of `server` should be defined/documented.

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
galelis: hmmm 
The only thing we do right now is to check if dcmjs support certain tool when we are saving, as we use dcmjs to retrieve as well, the code believes all that is comming is supported as well, not sure how to handle different than that right now. 

> No Length cornerstone-tool is "added" and "enabled"
and
> OR no Length vtk.js widget
I think ifs the tool is not enabled/added, the toolState will be added but nothing wll happen, right? 

> cornerstone-tool's Length tool expects a different tool state "shape"
We also dont treat this scenario right now, I believe we would just add a "wrong" data into toolState

platform/viewer/src/lib/DICOMSR/retrieveDataFromSR.js
galelis: I've started chaging this in DICOM SR code, but it would affect the whole code for measurementTable/measurementAPI can we have a different ticket so I can work on that?

dannyrb: Yes 👍 
Anything particularly arduous, just make a note of it.
Having this functionality in any non-brittle capacity is a big win. Anything else we can follow-up with.

platform/viewer/src/lib/DICOMSR/srUtils.js
galelis: Done 👍

platform/viewer/src/lib/DICOMSR/srUtils.js
galelis: Function renamed! 👍

platform/viewer/src/lib/DICOMSR/srUtils.js
galelis: Just did this refactor, way cleaned and all with JSDocs 👍

platform/core/src/measurements/classes/MeasurementApi.js
```diff
@@ -7,9 +8,7 @@ import guid from '../../utils/guid';
 import studyMetadataManager from '../../utils/studyMetadataManager';
 import { measurementApiDefaultConfig } from './../configuration.js';
 
-const configuration = {
-  ...measurementApiDefaultConfig,
-};
+const configuration = measurementApiDefaultConfig;
```
dannyrb: If we're not going to expand the config, spreading is safer. It prevents accidentally modifying the default config by reference.

platform/core/src/measurements/classes/MeasurementApi.js
dannyrb: Still need the shape of `server` and the `configuration` object defined somewhere.

platform/ui/src/components/measurementTable/MeasurementTable.js
```diff
@@ -10,6 +10,7 @@ import PropTypes from 'prop-types';
 import { ScrollableArea } from './../../ScrollableArea/ScrollableArea.js';
 import { TableList } from './../tableList';
 import { Tooltip } from './../tooltip';
+import { withSnackbar } from './../../contextProviders/SnackbarProvider';
```
dannyrb: This is still pulling in `withSnackbar`?

dannyrb:
Most CI is failing here:

>   1) OHIF Study Viewer Page "before all" hook for "checks if series thumbnails are being displayed":
     Uncaught ReferenceError: withSnackbar is not defined

sctkndl:
I'm interested in testing out this MeasurementAPI, can someone point me to any details on how to use it, please? I'm interested in using it mostly with Bidirectional Tool, and probably not with a Dicomweb server (I'm thinking storing in json somewhere maybe). With some guidance on getting started, I may be able to design an alternative that works for all tools.  Alternatively, I could look at dcmjs and expanding it's capability to store data from other tools, unless that's already been ruled out?

pieper:
Hi -

The plan is definitely to expand dcmjs to handle all the measurement types.  We plan to follow the pattern put forth in the white paper linked below.  Any help on that would be much appreciated.  It will be great once it's implemented but we also understand some one-off json is much more convenient sometimes.

https://docs.google.com/document/d/1bR6m7foTCzofoZKeIRN5YreBrkjgMcBfNA7r9wXEGR4/edit#heading=h.fgs65rsvrdy3

You might also be interested in this draft standard for simplifying the creation of SR documents:

ftp://medical.nema.org/medical/dicom/supps/Frozen/sup219_fz_JSONSR.pdf

galelis:
@sctkndl Hi, 

Working on DCMJS and adding support for more tools would be fantastic and for that we have another issue created and a big discussion happening around it. #1215 

If your issue is just to store measurements on OHIF, we use cornerstoneTools and you can access all measurements by doing grouped by ImageId and MeasurementTool: 
`cornerstoneTools.globalImageIdSpecificToolStateManager.toolState;`

If you want to work with the measurementAPI together with the MeasurementTable, on OHIF v1.x we had an implementation that we used to export measurementAPI data into a CSV file. You can see the implementation here: 
https://github.com/OHIF/Viewers/blob/v1.x/Packages/ohif-measurements/client/reports/reportMeasurementData.js

PR #968, opened 2019-09-27
Fix for test scripts that were failing in CI

mirnasilva:
Fixed issue with `OpenStudy()` customized command by adding `.then()` with a callback function that receives the element as its first argument for a retry-and-timeout functionality.


dannyrb:
@mirnasilva, this says Draft, but that you've requested my review. If it's ready, can you mark it as "Ready for Review"?

platform/viewer/cypress/support/commands.js
```diff
@@ -30,12 +30,73 @@ Cypress.Commands.add('openStudy', (patientName) => {
...
+
+
+//Command to reset the viewport changes throught the cornerstone method
+Cypress.Commands.add('resetViewport', () => {
+  cy.window()
+    .its('cornerstone')
+    .then(($cornerstone) => {
+      $cornerstone.reset($cornerstone.getEnabledElements()[0].element)
+    });
+});
```
dannyrb: This is why I hinted that we should probably use the UI to reset the viewport.
At first glance, this feels like a great solution, but it fails in a few places that the UI button should not:

- There's no guarantee that `cornerstone` will be available on the window. We expose it now for convenience/debugging, but may remove it in the future.
- Resetting a VTK (or other) viewport does not use this logic
- If you have multiple cornerstone viewports, this would only reset the first (not the active one)
- There's no guarantee that `getEnabledElements` ordering corresponds to the UIs viewport ordering

platform/viewer/cypress/support/commands.js
```diff
@@ -30,12 +30,73 @@ Cypress.Commands.add('openStudy', (patientName) => {
...
+    })
+});
+
+
+// Command to perform three clicks into three different positions. Each position must be (x, y).
+// The positions are considering the element as reference, therefore, top-left of the element will be (0, 0).
+Cypress.Commands.add('addAngle', (element, initPosition, midPosition, finalPosition) =>  {
+
+  cy.get(element)
+    .click(initPosition[0], initPosition[1], { force: true })
```
dannyrb: Does our element fail an [actionability check](https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Actionability)? What's causing us to use `force` here? Those checks are supposed to be in place to make sure we can only click things the user can actually click.

platform/viewer/cypress/support/commands.js
```diff
@@ -30,12 +30,73 @@ Cypress.Commands.add('openStudy', (patientName) => {
...
+    .then(() =>{
+      cy.get(element)
```
dannyrb: You shouldn't need to "get" the element again. You can leverage the pattern shown in Cypress's closure docs to clean this up: https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html#Closures

platform/viewer/cypress/support/commands.js
```diff
@@ -30,12 +30,73 @@ Cypress.Commands.add('openStudy', (patientName) => {
...
+Cypress.Commands.add('addLine', (element, initPosition, finalPosition) =>  {
+
+  cy.get(element)
+    .click(initPosition[0], initPosition[1], { force: true })
+    .then(() =>{
+      cy.get(element)
+        .trigger('mousemove', { clientX: finalPosition[0], clientY: finalPosition[1] })
+        .click({ force: true })
+    })
+});
```
dannyrb: To illustrate the points made below with `addAngle`, here is how you might rewrite this method:

```js
/**
 * Command to perform two clicks into two different positions. Each position must be [x, y].
 * The positions are considering the element as reference, therefore, top-left of the element will be (0, 0).
 *
 * @param {*} viewport - Selector for viewport we would like to interact with
 * @param {number[]} firstClick - Click position [x, y]
 * @param {number[]} secondClick - Click position [x, y]
 */
Cypress.Commands.add('addLine', (viewport, firstClick, secondClick) =>  {

  cy.get(viewport).then(($viewport) => {
    const [ x1, y1 ] = firstClick;
    const [ x2, y2 ] = secondClick;

    $viewport.click(x1, y1, { force: true});
    $viewport.click(x2, y2, { force: true });
  });
});
```

The goal here is readability. Functionally, these should accomplish the same thing. We're focused on trying to make the code read more like common language, or simple steps; as well as add clarity around the types/shape of the parameters required for this command.

platform/viewer/cypress/support/commands.js
```diff
@@ -30,12 +30,73 @@ Cypress.Commands.add('openStudy', (patientName) => {
...
     cy.get('#patientName')
       .type(patientName);
-    cy.get('.studylistStudy > .patientName')
-      .contains(patientName)
-      .click();
+
+    const patientResult = '.studylistStudy > .patientName';
+
+    cy.get(patientResult)
+      .then({ timeout: 5000 }, ($patientResult) => {
```
dannyrb: This would be a good place to leverage a Cypress `Alias` now that we have that tool in our toolbelt:

https://github.com/OHIF/Viewers/pull/968/files#diff-eee6caca0f3d24dfab61b837323f0192R83

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+        .trigger('mousedown', 'top', { which: 1 })
+        .trigger('mousemove', 'center', { which: 1 })
+        .trigger('mouseup');
+
+      const overlaySeriesInformation = 'div.ViewportOverlay > div.bottom-left.overlay-element > div';
+      const expectedText = 'Ser: 1Img: 14 14/26256 x 256Loc: 0.00 mm Thick: 5.00 mm';   
+      cy.get(overlaySeriesInformation)
+        .should('have.text', expectedText);
+
+      cy.resetViewport();
```
dannyrb: Similar to your `before(() => {});` hook above, you should have an `afterEach` you can leverage for this

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+
+      //Check on cornerstone if image was flipped vertically
+      cy.window()
+        .its('cornerstone')
+        .then(($cornerstone) => {
+          expect($cornerstone.getEnabledElements()[0].viewport.vflip).to.be.true
+        });
+
+      cy.resetViewport();
+    });
```
dannyrb: I think we're okay to remove this test.
We're trying to strike a balance between maintainability and increased confidence that the application works.
This test is difficult, because it has us relying on our window'd `cornerstone`.

We have less brittle tests we can rely on to assert that the `cornerstone` extension was registered successfully.
This is a bit nuanced, and certainly up for debate.

Phrased differently, the end-to-end tests are like smoke detectors.
We want enough to make sure we have advanced notice of fire, but too many and it will become a full time job to replace their batteries.
Let's place them strategically.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+
+      //Check on cornerstone if image was flipped horizontally
+      cy.window()
+        .its('cornerstone')
+        .then(($cornerstone) => {
+          expect($cornerstone.getEnabledElements()[0].viewport.hflip).to.be.true
+        });
+
+      cy.resetViewport();
+    });
```
dannyrb: Similarly, this one can probably be removed.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+
+      //Check on cornerstone if image was rotated
+      cy.window()
+        .its('cornerstone')
+        .then(($cornerstone) => {
+          expect($cornerstone.getEnabledElements()[0].viewport.rotation).to.equal(90)
+        });
+
+      cy.resetViewport();
+    });
```
dannyrb: Similarly, this one can probably be removed.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+
+      //check on cornerstone if image was inverted
+      cy.window()
+        .its('cornerstone')
+        .then(($cornerstone) => {
+          expect($cornerstone.getEnabledElements()[0].viewport.invert).to.be.true
+        });
+
+      cy.resetViewport();
+    });
```
dannyrb: This, similarly, falls under "visual testing". Worth giving this a read: https://docs.cypress.io/guides/tooling/visual-testing.html#Visual-diff-elements

Something like this: https://github.com/meinaart/cypress-plugin-snapshots
Or this: https://docs.percy.io/docs/cypress

Would be better suited for a test like this.
I think it's best we remember these tests, but earmark them as potential future visual tests.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+    it('checks if Layout tool will multiply the number of viewports displayed', () => {
+      //Click on Layout button and verify if overlay is displayed
+      cy.get('.btn-group > .toolbar-button').as('layoutBtn')
+        .click().then(() => {
+          cy.get('.layoutChooser')
+            .should('be.visible')
+            .find('td')
+            .its('length')
+            .should('be.eq', 9);
+        })
```
dannyrb: This one is a good assertion :+1:

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+
+        //verify if layout has changed to 2 viewports 
+        cy.get('tbody > :nth-child(1) > :nth-child(2)').click();
+        cy.get('.viewport-container').then(($viewport) =>{
+          cy.wrap($viewport)
+            .should('have.attr', 'style', 'height: 100%; width: 50%;')//checking if the layout is correct
+            .its('length')
+            .should('be.eq', 2);
+        })
```
dannyrb: I think we can assert `length` here, but the actual styling checks will break with changes that should be merged today.

Semantically:
- Select X Layout
- Verify layout length reflects selection
- Repeat a few times

Instead of style checking, we would eventually add visual checks that the layouts are consistently displayed.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+
+      //Verify if overlay is hidden
+      cy.get('.tooltip-toolbar-overlay')
+        .should('not.be.visible');
+    });
+
+
+    it('checks if Layout tool will multiply the number of viewports displayed', () => {
+      //Click on Layout button and verify if overlay is displayed
+      cy.get('.btn-group > .toolbar-button').as('layoutBtn')
```
dannyrb: We already have an alias for `layoutBtn`? Does it not work if we reuse it here? If so, how can we make those aliases reusable?

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+        })
+
+        //Check if More button is active and if it has same icon as the secondary tool selected
+        cy.get('.expandableToolMenu')
+        .click()
+        .then(($moreBtn) =>{
+          cy.wrap($moreBtn)
+            .should('have.class', 'active')
+            .find('svg').should(($icon2)=>{
+              expect($icon2.text()).to.contain(iconName)
```
dannyrb: You should be able to use something like `.contains()` to prevent drilling down to the text:
- https://docs.cypress.io/api/commands/contains.html#Syntax

Basically: `$moreBtn.contains(iconName)`

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+         $cornerstone.getEnabledElements()[0].viewport.rotation = 90;
+        });
+
+      //Click on reset button
+      cy.get('.toolbar-button:nth-child(9)').click()
+   
+      const overlaySeriesInformation = 'div.ViewportOverlay > div.bottom-right.overlay-element > div';
+      const expectedText = 'Zoom: 211%W: 820 L: 410Lossless / Uncompressed';
+      cy.get(overlaySeriesInformation)
+        .should('have.text', expectedText);
```
dannyrb: Using the overlay text here is a great approach 🥇

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+      //Make some changes by zooming in and rotating the image
+      cy.window()
+        .its('cornerstone')
+        .then(($cornerstone) => {
+          $cornerstone.getEnabledElements()[0].viewport.scale = 4;
+         $cornerstone.getEnabledElements()[0].viewport.rotation = 90;
+        });
+
+      //Click on reset button
+      cy.get('.toolbar-button:nth-child(9)').click()
```
dannyrb: Don't we have an existing alias for this? If the location on the toolbar changes, we would need to update this selector in 2 places now instead of one.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+
+
+    it('checks if Reset tool will reset all changes made on the image', () => {
+      //Make some changes by zooming in and rotating the image
+      cy.window()
+        .its('cornerstone')
+        .then(($cornerstone) => {
+          $cornerstone.getEnabledElements()[0].viewport.scale = 4;
+         $cornerstone.getEnabledElements()[0].viewport.rotation = 90;
+        });
```
dannyrb: We should probably make these modifications by hand. If you need to, you can create mini-commands that make them quicker and easier to redo.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+            .as('measurementsPanel')
+            .should('be.visible');
+            
+          cy.get('.measurementItem')
+            .its('length')
+            .should('be.eq', 2);
+
+          cy.wrap($measurementsBtn)
+          .click();
+        })
```
dannyrb: This is a really good assertion to test ^_^
I think if we apply the lessons learned in other tests, we can clean it up to read a little easier.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,509 @@
...
+      cy.get('@levelsBtn').should('be.visible').contains('Levels');
+      cy.get('@panBtn').should('be.visible').contains('Pan');
+      cy.get('@lengthBtn').should('be.visible').contains('Length');
+      cy.get('@annotateBtn').should('be.visible').contains('Annotate');
+      cy.get('@angleBtn').should('be.visible').contains('Angle');
+      cy.get('@resetBtn').should('be.visible').contains('Reset');
+      cy.get('@cineBtn').should('be.visible').contains('CINE');
+      cy.get('@moreBtn').should('be.visible').contains('More');
+      cy.get('@twodmprBtn').should('be.visible').contains('2D MPR');
+      cy.get('@layoutBtn').should('be.visible').contains('Layout');
```
dannyrb: We define a lot of aliases here, but we don't use them in the below tests

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
mirnasilva: Yes, we do! Replaced by the alias.
Also, once we get the `data-cy` elements, I'm planning to refactor these selectors for better maintainability.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
mirnasilva: Thanks for your help on this one! :D

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
mirnasilva: Great observation. I'll keep this in mind and will implement some visual testing conditions on our next tests.

platform/viewer/cypress/support/commands.js
mirnasilva: Replaced command to use the UI to reset the viewport

platform/viewer/cypress/support/commands.js
mirnasilva: Thanks for all the great suggestions!

platform/viewer/cypress/support/commands.js
mirnasilva: It's necessary to grab the element again. I replaced it with `cy.wrap()` which is more appropriate then `cy.get()` in this context.

platform/viewer/cypress/support/commands.js
mirnasilva: Yeah, unfortunately, it is necessary not only to `force` the click but also to `trigger` a mouse movement to the next position. Otherwise, the annotation was not being added.

mirnasilva:
Hi @dannyrb , changes we made according to your suggestions. Can you please take a look again on this PR? Thanks!

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -22,11 +22,10 @@ describe('OHIF Cornerstone Toolbar', () => {
       cy.get('.pull-right > .RoundedButtonGroup > .roundedButtonWrapper > .roundedButton').as('measurementsBtn');
       cy.get('.viewport-element').as('viewport');
       cy.get('section.sidepanel.from-right').as('measurementsPanel')
+      //Following best practices, reset should be done before each test
+      cy.resetViewport();
```
dannyrb: :+1:

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -0,0 +1,407 @@
...
+          .should('not.have.text', expectedText);
+
+        //Test SKIP TO FIRST IMAGE button
+        cy.get('[title="Skip to first Image"]')
+          .click()
+        cy.get(overlaySeriesInformation)
+          .should('contain.text', expectedText);
+
+        //Test NEXT IMAGE button
+        cy.get('[title="Next Image"]') //Title is wrong and was reported on bug #995: https://github.com/OHIF/Viewers/issues/995
```
dannyrb: This fix should be merged. Can remove the `todo` comment? (and the one below)

PR #989, opened 2019-10-02
feat: 🎸 Add gcloud route to application

ladeirarodolfo:
Added gcloud full route to select store/location/project/study/series
### Includes
- Viewer to load url based on gcloud API url pattern
- Uses static gcloud API url pattern
- Persist servers/gcloud info on session storage

### A step to
- issue #879 
- Generic route handler (based on app config)

### Screenshoot
Oct-10-2019-2_38-PM

Closes: #753


swederik:
In case anyone else is trying to test this, and has signed the TTA for access to the TCIA collections on Google Cloud Healthcare:

1. Checkout the branch
2. Run the viewer with your client ID in the config file 
3. Navigate to a URL like this:
http://localhost:3000/projects/chc-tcia/locations/us-central1/datasets/qin-headneck/dicomStores/qin-headneck/dicomWeb/study/1.3.6.1.4.1.14519.5.2.1.2744.7002.292382393135745002321454178686

platform/viewer/src/googleCloud/api/GoogleCloudApi.js
```diff
@@ -22,6 +22,13 @@ class GoogleCloudApi {
     return this.urlBase + `/projects`;
   }
 
+  getUrlBaseDicomWeb(project, location, dataset, dicomstore) {
+    return (
+      this.urlBase +
+      `/projects/${project}/locations/${location}/datasets/${dataset}/dicomStores/${dicomstore}/dicomWeb`
```
swederik: For me this PR is good, but I would remove the`/dicomWeb` at the end. Thoughts @pieper?

pieper: I agree, no need for `dicomWeb`.

Otherwise yes, LGTM 👍

ladeirarodolfo: Ok removed it..

ladeirarodolfo:
Updated PR with more improvements: 
Store gcloud info on session storage (reduced requirement to set gcloud data everytime)
Check it out screenshoot

swederik:
Latest changes are working well for me, and the code changes look good. This makes Google Cloud usage so much less frustrating. Nice work @ladeirarodolfo!

PR #990, opened 2019-10-02
Move Series Information to Separate Row

sedenardi:
When viewing series in the Study Browser, series with long descriptions cause the numeric information to get pushed out of view.
Screen Shot 2019-10-02 at 12 14 13

To fix this, I've moved the information to a separate row (similar to how it was styled if it were `info-only`):
Screen Shot 2019-10-02 at 12 27 29

Another part of this PR involves fixing the alignment of series when there's an image and when there's not. Before (look at the descriptions):
Screen Shot 2019-10-02 at 12 42 10

After:
Screen Shot 2019-10-02 at 12 42 23

Let me know if you have any questions/comments/concerns. Thank you.


swederik:
Related: https://github.com/OHIF/Viewers/issues/821

PR #996, opened 2019-10-04
Add externals config to webpack bundling configuration of core project

tunglt1810:


dannyrb:
@tunglt1810, I'll be tackling this near term. I appreciate the attempt, but we need slightly more explicit external mappings for this to work for all of our supported use cases

Issue #561, opened 2019-06-15
VTK Extension: Disable MPR button for unreconstructable volumes


swederik:
Currently the MPR mode is always available. This doesn't make sense in a lot of cases.

- If we have images of different dimensions within a display set
- If we have images with different numbers of components (e.g. grayscale and RGB) within a display set
- If we have images which are not in parallel orientations
- If we only have a single image
- If the SOP Class is not an image (e.g. structured report)
- If the SOP Class does not make sense to reconstruct (e.g. microscopy)

I'm sure there are more reasons why some stacks cannot be easily reformatted, and so we should add this logic somewhere. [Slicer](https://slicer.org) or [MITK](http://mitk.org) must have this logic somewhere which we can take a look at.

PR #1010, opened 2019-10-08
feat: 🎸 Only allow reconstruction of datasets that make sense

JamesAPetts:
Regarding #561 :

Only allow reconstruction of datasets that can be 3D volumes:
- Not SR/PDF/etc, not a microscopy slide.
- Has more than one frame.
- Has the same dimmensions on each frame.
- Has the same number of components (monochrome/RGB) on each frame.
- Has parallel frames.
- Has equal spacing between frames, however if it looks like a slice is simply missing, still allow 3D reconstruction.

### DIFF
--- a/platform/core/src/utils/isDisplaySetReconstructable.js
+++ b/platform/core/src/utils/isDisplaySetReconstructable.js
@@ -1,11 +1,36 @@
+/**
+ * Checks if a series is reconstructable to a 3D volume.
+ *
+ * @param {Object} series The `OHIFSeriesMetadata` object.
+ * @param {Object} instances The `OHIFInstanceMetadata` object
+ */
 export default function isDisplaySetReconstructable(series, instances) {
-  debugger;
-
   // Can't reconstruct if we only have one image.
-  if (instances.length === 1) {
-    return false;
+
+  const modality = series._data.modality; // TODO -> Is there a better way to get this?
+  const isMultiframe = instances[0].getRawValue('x00280008') > 1;
+
+  if (!constructableModalities.includes(modality)) {
+    return { value: false };
   }
 
+  if (!isMultiframe && instances.length === 1) {
+    return { values: false };
+  }
+
+  if (isMultiframe) {
+    return processMultiframe(instances[0]);
+  } else {
+    return processSingleframe(instances);
+  }
+}
+
+function processMultiframe(instance) {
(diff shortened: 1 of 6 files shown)


platform/core/src/utils/isDisplaySetReconstructable.js
```diff
@@ -0,0 +1,150 @@
+/**
+ * Checks if a series is reconstructable to a 3D volume.
+ *
+ * @param {Object} series The `OHIFSeriesMetadata` object.
+ * @param {Object} instances The `OHIFInstanceMetadata` object
```
swederik: Seems to be an array

platform/core/src/utils/isDisplaySetReconstructable.js
```diff
@@ -0,0 +1,150 @@
...
+    const averageSpacingBetweenFrames =
+      _getPerpindicularDistance(firstIpp, lastIpp) / (instances.length - 1);
+
+    let previousIpp = firstIpp;
+
+    for (let i = 1; i < instances.length; i++) {
+      const instance = instances[i];
+      const ipp = _getImagePositionPatient(instance);
+
+      const spacingBetweenFrames = _getPerpindicularDistance(ipp, previousIpp);
```
swederik: Perpendicular

platform/core/src/utils/isDisplaySetReconstructable.js
```diff
@@ -0,0 +1,150 @@
...
+        spacingBetweenFrames,
+        averageSpacingBetweenFrames
+      );
+
+      if (spacingIssue) {
+        const issue = spacingIssue.issue;
+
+        if (issue === 'missing') {
+          missingFrames += spacingIssue.missingFrames;
+        } else if (issue === 'different') {
```
swederik: Can we make 'missing' and 'different' (whatever that means?) into constants somewhere?

JamesAPetts: Going to renamed `different` to `irregular`. Missing is reconstructable uniform spacing but with missing slices, whereas `irregular` doesn't fit on a uniform grid (has multiple slice spacings).

dannyrb:
My only note would be that, instead of hiding the icon, provide a different one (or some other visual cue), that the active stack is a no-go.

PR #1011, opened 2019-10-08
feat: Notification Service

### LINKED ISSUE #985: Notification Service
Per https://github.com/OHIF/Viewers/pull/954#pullrequestreview-295401590

Creating a very simple notification service that allows users to

**Resources:**

- Check out this library's readme as a good example of the kind of API we would want to expose/provide for service consumers: https://github.com/igorprado/react-notification-system
- Here are some UI components I've been playing with as an additional source of design inspiration: https://codepen.io/dannyrb/pen/xyZEER
- Here is an example of `useContext` and a notification system: https://daveceddia.com/usecontext-hook/

**Acceptance Criteria**

- [ ] Creation of notification components in `@ohif/ui`
- [ ] Creation of a simple service in `@ohif/core`
- [ ] Consume the service [here](https://github.com/OHIF/Viewers/pull/954#pullrequestreview-295401590) (or in at least one place to show usage)

rodrigolabs:
# Description

This PR creates a Notification Service to be used in the application.

## Usage

First of all, import the `SnackbarProvider` and wrap it at the top level of your application:

```javascript
import { SnackbarProvider } from '@ohif/ui';

const App = () => {
  return <SnackbarProvider>{...}</SnackbarProvider>;
};
```

### Class Components

It's needed to wrap the exported component into a HOC created to pass the Snackbar methods as props.

e.g.:

```javascript
import { withSnackbar } from "@ohif/ui";

class HelloWorld extends Component {
  render() {
    return (
      <button
        onClick={() => this.props.snackbarContext.show({ message: "test" })}
      >
        Notify
      </button>
    );
  }
}

export default withSnackbar(HelloWorld);
```

### Functional Components

Use the exported hook `useSnackbarContext` to retrieve its methods.

```javascript
import { useSnackbarContext } from "@ohif/ui";

const HelloWorld = () => {
  const snackbarContext = useSnackbarContext();

  return <button onClick={() => snackbarContext.show({ message: "test" })}>Notify</button>;
};

export default HelloWorld;
```

## Snackbar API Details

The Snackbar Provider pass down the following methods:

### show()

`snackbarContext.show({ message: "Hello" })`

Create a new notification in the stack.

**Params:**

```
{
  title: string,
  message: string,
  duration: number,
  position: "topLeft" | "topCenter | "topRight" | "bottomLeft" | "bottomCenter" | "bottomRight,
  type: "info" | "error" | "warning" | "success",
  autoClose: boolean
}
```

The **required params** are only `title` or `message`.

**Default values:**

```
{
  duration: 5000,
  position: "bottomRight",
  type: "info",
  autoClose: true
}
```

___

### hide()

**It's used internally to automatically close a notification. We will probably never need to use it!**

`snackbarContext.hide(1)`

Hide a specific notification in the stack. 

**Params:**

it receives only one parameter, **which is required**:

```
id: number
```

___

### hideAll()

`snackbarContext.hideAll()`

Hide all notifications in the stack.

**Params:**

```
null
```

---

Fixes #985


dannyrb:
I had no idea that `Snackbar` was an industry name for components like this:
- https://material.io/components/snackbars/#anatomy

I assume "toast notifications" are called that, because they pop up like a piece of toast in the toaster. Is the Snack bar a tiny bar of snack sized info?

platform/viewer/src/components/Snackbar/Snackbar.css
```diff
@@ -0,0 +1,235 @@
...
+    top: 30px;
+  }
+  to {
+    opacity: 1;
+    top: 0;
+  }
+}
+
+/* Firefox < 16 */
+@-moz-keyframes fadein {
```
dannyrb: We _might_ have some preprocessing in place that takes care of these vendor prefixes for you. If we don't, we should. As a short-term, I think you would only need to keep the IE alternatives for this, as the others fall out of our range of support.

platform/viewer/src/components/Snackbar/Snackbar.css
```diff
@@ -0,0 +1,235 @@
...
+.sb-error {
+  background-color: rgba(236, 61, 61, 0.9);
+}
+
+.sb-warning {
+  background-color: rgba(235, 173, 23, 0.9);
+}
+
+.sb-info {
+  background-color: rgba(54, 156, 199, 0.9);
```
dannyrb: Can we add these colors to our `:root` css variables, and then pull them from there as well?

platform/viewer/src/components/Snackbar/Snackbar.css
```diff
@@ -0,0 +1,235 @@
...
+.sb-item a {
+  color: white;
+  text-decoration: underline;
+}
+
+.sb-hidden {
+  padding: 0 !important;
+  margin: 0 !important;
+  height: 0 !important;
+  opacity: 0 !important;
```
dannyrb: Why `!important`? Should be able to accomplish with higher specificity?

`.sb-item.sb-hidden`

platform/viewer/src/components/Snackbar/Snackbar.css
```diff
@@ -0,0 +1,235 @@
...
+}
+
+.sb-item {
+  position: relative;
+  transition: height 300ms ease;
+  -webkit-animation: fadein 1s;
+  -moz-animation: fadein 1s;
+  -ms-animation: fadein 1s;
+  -o-animation: fadein 1s;
+  animation: fadein 1s;
```
dannyrb: Man, I hate manually adding vendor prefixes. If we get a test failure on a different OS, we can create an issue to add something via PostCSS to rectify this. Please feel free to remove the prefixed attributes.

platform/viewer/src/components/Snackbar/Snackbar.css
```diff
@@ -0,0 +1,235 @@
+.sb-container {
+  position: fixed;
+  width: 420px;
+  padding: 20px;
+  z-index: 9998;
+  box-sizing: border-box;
```
dannyrb: I think we have a wildcard for this at the top level.

`* { box-sizing: border-box; }`, so you shouldn't need this.

platform/viewer/src/components/Snackbar/Snackbar.css
```diff
@@ -0,0 +1,235 @@
+.sb-container {
+  position: fixed;
+  width: 420px;
+  padding: 20px;
+  z-index: 9998;
```
dannyrb: If possible, it might be nice to also set/pull this as a css-variable, so we have better visibility into how our application is "layered"

platform/viewer/src/components/Snackbar/Snackbar.css
```diff
@@ -0,0 +1,235 @@
+.sb-container {
+  position: fixed;
+  width: 420px;
```
dannyrb: Just curious, why `420px`?

dannyrb:
These are isolated pretty well. If you want, I think the component portion would be fine to move to the UI project in the components directory. I think we'd keep the SnackbarContext in `viewer`. Probably... Under `store` for now?

Only reason for moving would be easier documentation for the awesome examples you've spelled out in your PR already.

platform/viewer/src/components/Snackbar/Snackbar.css
rodrigolabs: Cool! I've added in the `:root` as a CSS variable. I did a quick search in the application for `z-index` and there are a lot of different values. Maybe it would be good to sometime take them all and put in the CSS variable as well 😊

platform/viewer/src/components/Snackbar/Snackbar.css
rodrigolabs: I was trying to force `!important` when I was creating, but now it's not needed anymore. Thanks!

platform/viewer/src/components/Snackbar/Snackbar.css
rodrigolabs: Good question, It was just a random value, but looking at the material specifications, the default width is set to `344px`, I'll change to `344px`. Any thoughts?

rodrigolabs:
> These are isolated pretty well. If you want, I think the component portion would be fine to move to the UI project in the components directory. I think we'd keep the SnackbarContext in `viewer`. Probably... Under `store` for now?
> 
> Only reason for moving would be easier documentation for the awesome examples you've spelled out in your PR already.

@dannyrb I did the fixes and moved the Snackbar component to the UI project. SnackbarContext is still in `viewer`, under `context`, because it has its own exported context and `Provider` component as well, I think it's better to keep it in the `context`, what do you think? that's ok?

platform/ui/src/components/snackbar/SnackbarContainer.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import SnackbarItem from './SnackbarItem';
+import './Snackbar.css';
+import { useSnackbarContext } from '../../../../viewer/src/context/SnackbarContext';
```
dannyrb: The tricky thing, and this is less obvious in our monorepo setup, is that imports that cross package boundaries will cause the individual project builds to fail. If you need more clarification on this, let me know.

platform/viewer/src/context/SnackbarContext.js
```diff
@@ -0,0 +1,102 @@
+import React, { useState, createContext, useContext } from 'react';
+import SnackbarContainer from '../../../ui/src/components/snackbar/SnackbarContainer';
+import SnackbarTypes from '../../../ui/src/components/snackbar/SnackbarTypes';
```
dannyrb: Similarly, these will work okay in our PWA build, but it will cause our `@ohif/ui` and `@ohif/viewer` packages to fail.

PR #1012, opened 2019-10-08
feat: saner react cornerstone extension update

dannyrb:
@JamesAPetts I've isolated the changes required to update the `react-cornerstone-viewport` from my larger PR. After this is merged, I'll create smaller, incrememental PRs that include the other changes so we're not making any large, hard to review, jumps.

Let me know if you have any questions on this one.


dannyrb:
Not sure what's up with our lockfile not updating. Un momento.

JamesAPetts:
Tackling this beast in a sec.

extensions/cornerstone/src/ConnectedCornerstoneViewport.js
```diff
@@ -23,36 +23,41 @@ const MEASUREMENT_ACTION_MAP = {
 const mapStateToProps = (state, ownProps) => {
   let dataFromStore;
 
+  // TODO: This may not be updated anymore :thinking:
```
JamesAPetts: 👀

extensions/cornerstone/src/ConnectedCornerstoneViewport.js
```diff
@@ -23,36 +23,41 @@ const MEASUREMENT_ACTION_MAP = {
...
-  // onNewImage,
-  // onRightClick,
-  // onTouchPress,
-  // onTouchStart,
-  // onMouseClick,
-
   // If this is the active viewport, enable prefetching.
   const { viewportIndex } = ownProps; //.viewportData;
   const isActive = viewportIndex === state.viewports.activeViewportIndex;
+  const isStackPrefetchEnabled = isActive;
```
JamesAPetts: If this just equals `isActive`, why have two variables? Can't we just base the stack prefetching logic on whether the port is active or not?

extensions/cornerstone/src/ConnectedCornerstoneViewport.js
```diff
@@ -23,36 +23,41 @@ const MEASUREMENT_ACTION_MAP = {
...
   const viewportSpecificData =
     state.viewports.viewportSpecificData[viewportIndex] || {};
 
+  // CINE
+  let isPlaying = false;
+  let frameRate = 24;
+
+  if (viewportSpecificData && viewportSpecificData.cine) {
```
JamesAPetts: This isn't this PR's fault I think, but I just noticed the CINE dialog is no longer draggable, is that a known issue?

extensions/cornerstone/src/OHIFCornerstoneViewport.js
```diff
@@ -249,25 +249,40 @@ class OHIFCornerstoneViewport extends Component {
   render() {
     let childrenWithProps = null;
 
+    if (!this.state.viewportData) {
+      return <></>;
```
JamesAPetts: Just `return null`. No need for a fragment here? Returning `null` makes React ignore the component, returning a fragment _might_ (?) incur some unneeded overheads.

extensions/cornerstone/src/OHIFCornerstoneViewport.js
```diff
@@ -249,25 +249,40 @@ class OHIFCornerstoneViewport extends Component {
...
+    }
+    const { viewportIndex } = this.props;
+    const {
+      imageIds,
+      currentImageIdIndex,
+      frameRate = 0, // Sometimes available here
```
JamesAPetts: `frameRate` appears unused?

extensions/cornerstone/src/OHIFCornerstoneViewport.js
```diff
@@ -249,25 +249,40 @@ class OHIFCornerstoneViewport extends Component {
...
-            viewportData={this.state.viewportData}
-            viewportIndex={this.props.viewportIndex}
-            {...this.props.customProps}
-          />
-        )}
+        <ConnectedCornerstoneViewport
+          viewportIndex={viewportIndex}
+          imageIds={imageIds}
+          imageIdIndex={currentImageIdIndex}
+          // ~~ Connected (From REDUX)
```
JamesAPetts: I like this pattern of commenting what is included from Redux 👍 .

extensions/cornerstone/src/index.js
```diff
@@ -13,6 +18,9 @@ export default {
    */
   id: 'cornerstone',
 
+  preRegistration(configuration = {}) {
```
JamesAPetts: It might just be worth doccing that either an object can be passed here (global config), or an array of config for each `cornerstoneTools` module?

extensions/cornerstone/src/init.js
```diff
@@ -0,0 +1,83 @@
...
+  const metadataProvider = new OHIF.cornerstone.MetadataProvider();
+
+  cornerstone.metaData.addProvider(
+    metadataProvider.provider.bind(metadataProvider)
+  );
+
+  StackManager.setMetadataProvider(metadataProvider);
+
+  //
+  initCornerstoneTools({
```
JamesAPetts: You should use the `configuration` here. If `configuration` is just an object, merge it, if its an array merge/add it to the correct array element: https://github.com/cornerstonejs/cornerstoneTools/blob/3f5e46c5e243ab7f27759af0df184946df283c2c/src/init.js#L25

extensions/cornerstone/src/init.js
```diff
@@ -0,0 +1,83 @@
...
+    ArrowAnnotateTool,
+    BidirectionalTool,
+    LengthTool,
+    AngleTool,
+    FreehandRoiTool,
+    EllipticalRoiTool,
+    DragProbeTool,
+    // Segmentation
+    BrushTool,
+    RectangleRoiTool,
```
JamesAPetts: This should be under `// Annotations`. Maybe you were thinking of the `RectangleScissorsTool`?

platform/viewer/src/App.js
```diff
@@ -36,11 +34,6 @@ import UserManagerContext from './context/UserManagerContext';
 import AppContext from './context/AppContext';
 
 // ~~~~ APP SETUP
-initCornerstoneTools({
```
JamesAPetts: This feels less tangled now 👍 .

platform/viewer/src/App.js
```diff
@@ -175,8 +165,9 @@ class App extends Component {
 function _initExtensions(extensions, hotkeys) {
   const defaultExtensions = [
     GenericViewerCommands,
-    MeasurementsPanel,
     OHIFCornerstoneExtension,
+    // WARNING: MUST BE REGISTERED _AFTER_ OHIFCORNERSTONEEXTENSION
```
JamesAPetts: Food for thought, we need to worry about dependency management at some point in the future, as it'll only get worse. We'll get so many GitHub issues of X doesn't work with Y:  -> "...swap the order". If each extention listed their dependencies we could sort the order and the user won't have to care too much when pulling down a new shiny from the web. We could also throw a warning if a dependency isn't provided? I digress.

platform/viewer/src/appExtensions/MeasurementsPanel/index.js
```diff
@@ -1,7 +1,15 @@
 import ConnectedMeasurementTable from './ConnectedMeasurementTable.js';
+import init from './init.js';
```
JamesAPetts: Food for thought: Would the measurements table make more sense as part of the cornerstone extention? 🤔

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
```diff
@@ -0,0 +1,188 @@
...
+  csTools.removeTool('Length');
+  csTools.removeTool('Angle');
+  csTools.removeTool('FreehandRoi');
+  csTools.removeTool('EllipticalRoi');
+  csTools.removeTool('CircleRoi');
+  csTools.removeTool('RectangleRoi');
+  csTools.removeTool('ArrowAnnotate');
+
+  // Re-add each tool w/ our custom configuration
+  csTools.addTool(csTools.BidirectionalTool, {
```
JamesAPetts: All of this feels like something that should happen with the tool configuration from the cornerstone extention?
Right now if I add a third party tool that'd like to register to the measurements table I'd have to define it in two places?

What if I put configuration for a tool that my app relies on in the cornerstone extention? It'll get nuked here and I'll be super confused as a user.

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
```diff
@@ -0,0 +1,188 @@
...
+  });
+
+  // TODO: MEASUREMENT_COMPLETED (not present in initial implementation)
+  const onMeasurementsChanged = (action, event) => {
+    return MEASUREMENT_ACTION_MAP[action](event);
+  };
+  const onMeasurementAdded = onMeasurementsChanged.bind(this, 'added');
+  const onMeasurementRemoved = onMeasurementsChanged.bind(this, 'removed');
+  const onMeasurementModified = onMeasurementsChanged.bind(this, 'modified');
+  const onLabelmapModified = onMeasurementsChanged.bind(
```
JamesAPetts: 👀  Are we going to have labelmap related stuff in the measurement table or will it live elsewhere? Its a whole other beast of complexity. P.S. since `labelmapModified` is debounced to the end of brush strokes or SEG tool usage this isn't a bottleneck, but just curious.

platform/viewer/src/appExtensions/MeasurementsPanel/labelingFlowCallbacks.js
```diff
@@ -0,0 +1,144 @@
...
+    };
+
+    // setToolContextMenuData(viewportIndex, toolContextMenuData);
+    setToolContextMenuData(0, toolContextMenuData);
+  };
+}
+
+function getOnTouchPressCallback(store) {
+  const setToolContextMenuData = (viewportIndex, toolContextMenuData) => {
+    store.dispatch(resetLabellingAndContextMenuAction());
```
JamesAPetts: These two actions are often dispatched together, could we bake them into one action? Probably a bit nitpicky.

platform/viewer/src/connectedComponents/ToolContextMenu.js
```diff
@@ -2,7 +2,8 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import cornerstone from 'cornerstone-core';
 import cornerstoneTools from 'cornerstone-tools';
-import getMeasurementLocationCallback from '../lib/getMeasurementLocationCallback';
+// This whole component should live in the Measurements Extension :thinking:
```
JamesAPetts: 👍

extensions/cornerstone/src/ConnectedCornerstoneViewport.js
dannyrb: Changing state/props to be cleaner is the rabbit hole of the larger PR. On my list, but just need this fixes merged first.

extensions/cornerstone/src/ConnectedCornerstoneViewport.js
dannyrb: Shot you a note. It _was_ working, but stopped after I force updated our yarn lock. I think this is a loose versioning issue. Will hold on merging until resolved.

extensions/cornerstone/src/OHIFCornerstoneViewport.js
dannyrb: I think you're right. We have a lot of styling that targets diva inside named elements. Wasn't sure if this would have an impact. (Do fragments function as empty divs, or as fake containers to appease vdom?) I'll test with null, and if it works well, then I will go that route.

extensions/cornerstone/src/OHIFCornerstoneViewport.js
dannyrb: Not sure if this derived from metaData and should override the default/initial value. It can be hard to trace where data is coming from, so I left it for now. I can remove it

swederik: It would come from the FrameTime in the instance, for example.

platform/viewer/src/appExtensions/MeasurementsPanel/index.js
dannyrb: Great question. I can see the pros and cons. As extensions work today, I prefer them separate. It would also be nice if the measurements panel could show vtk measurements. Kind of need an `@ohif/core` measurements service with providers/consumers to make that work though.

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
dannyrb: At some point, very likely.

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
dannyrb: You've never been able to reconfigure without modifying code and a rebuild. As a low hanging fruit, next enhancement, this extension provides a default configuration, but the user can provide an alternative one at runtime, or an `onModified` one that merges the default config?

This change gets us one step closer by moving all of this logic into a preregistration hook. Before it was buried in app code.

platform/viewer/src/appExtensions/MeasurementsPanel/labelingFlowCallbacks.js
dannyrb: Either as one action, or using "redux thunk". I would be okay with that.

Issue #1013, opened 2019-10-08
Is not possible to add Measurement Description if Relabel was not previously selected


mirnasilva:
### Description

On measurement panel, user is not able to add a measurement Description if a Relabel was not previously selected.

![bug-description](https://user-images.githubusercontent.com/13245529/66440171-8267f600-ea08-11e9-8d78-3a178a5982ae.gif)

### Steps to Reproduce

1. Load view studies page
2. Add a Length measurement
3. Click on Measurements button
4. Select the measurement item and click on Description
5. Add "Text test" and click on Confirm button

### Actual Result

The description text was not added to the panel. If user clicks again on 'Description' button, he/she will see the entered text there.

### Acceptance Criteria

User should be able to add a Description text on the measurement, even if Relabel was not selected yet.

dannyrb:
@evren217 want to double check and see if this was a conscious decision. Description does not show unless a label is set.

evren217:
We should be able to show description only if entered without relabeling.

dannyrb:
Thanks for confirming that this is a bug @evren217

Issue #999, opened 2019-10-04
Current viewport is not being set as active after changing Layout selection back to one viewport


mirnasilva:
### Description

If user changes the viewport selection to more than one viewport and interacts with those viewports, once he/she switches back to only 1 viewport, the current viewport is not being set as active. This is causing the application to throw an exception when the user selects another tool at the toolbar menu.

![layout-bug](https://user-images.githubusercontent.com/13245529/66244550-68b07100-e6df-11e9-8d7c-f76b79aabb06.gif)

### Steps to reproduce

1. Load view studies page
2. Select the Layout tool on the toolbar
3. Change the layout selection to 3 (or more)
4. Click on any viewport different from the 1st one
5. Change the layout selection back to 1
6. Click on Reset tool on the toolbar
7. Check that exception `element not enabled` was thrown 

### Actual result

Exception `element not enabled` is being thrown and user is not able to use the selected tool

### Acceptance criteria

After changing the viewport number selection and returning back to only 1 viewport, the current viewport should be set as `active` by default.

dannyrb:
👋 Take it away, @igoroctaviano

dannyrb:
Resolved here: https://github.com/OHIF/Viewers/pull/1018

PR #1014, opened 2019-10-08
Test cases implemented: Cornerstone tools Clear and Eraser; Measurement panel Relabel and Description

mirnasilva:
### Test cases implemented

- Cornerstone tools: 
  - Clear;
  - Eraser; 
- Measurement panel: 
  - Relabel;
  - Description (this test is currently commented due to issue #1013 );

### Aliases declaration

Created an Aliases file to facilitate the maintenance of the selectors: `./support/aliases.js`

Now we can use commands: `cy.initCornerstoneToolsAliases()` to initialize the aliases for Cornerstone tools and `cy.initCommonElementsAliases()` to initialize the aliases for common page elements. The initialization of the aliases should be added inside the `beforeEach` hook of each test suite.

**Note:** Keep in mind that currently, cypress alias works as a constant of the element. Therefore, if you're manipulating an element that needs to retrieve an updated value, even if you perform a new `cy.get('@alias')`, the storage value will be the first previously retrieved. In this case, you may need to work with the selector itself, instead of the alias.

**Note 2:** Cypress is also not able to create aliases of elements that are not currently displayed.


platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -6,22 +6,8 @@ describe('OHIF Cornerstone Toolbar', () => {
...
-      cy.get('.toolbar-button:nth-child(9)').as('resetBtn');
-      cy.get('.toolbar-button:nth-child(10)').as('cineBtn');
-      cy.get('.expandableToolMenu').as('moreBtn');
-      cy.get('.PluginSwitch > .toolbar-button').as('twodmprBtn');
-      cy.get('.btn-group > .toolbar-button').as('layoutBtn');
-      cy.get('.pull-right > .RoundedButtonGroup > .roundedButtonWrapper > .roundedButton').as('measurementsBtn');
-      cy.get('.viewport-element').as('viewport');
-      cy.get('section.sidepanel.from-right').as('measurementsPanel')
+      cy.initCornerstoneToolsAliases();
+      cy.initCommonElementsAliases();
```
dannyrb: That's interesting. So these persist if set in `beforeEach`, but not if done in a separate test case?

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -58,9 +44,8 @@ describe('OHIF Cornerstone Toolbar', () => {
         .trigger('mousemove', 'center', { which: 1 })
         .trigger('mouseup');
 
-      const overlaySeriesInformation = 'div.ViewportOverlay > div.bottom-left.overlay-element > div';
       const expectedText = 'Ser: 1Img: 14 14/26256 x 256Loc: 0.00 mm Thick: 5.00 mm';   
-      cy.get(overlaySeriesInformation)
+      cy.get('@viewportInfoBottomLeft')
```
dannyrb: Good job cleaning up selectors :+1:

platform/viewer/cypress/support/commands.js
```diff
@@ -118,13 +130,24 @@ Cypress.Commands.add('waitDicomImage', (timeout = 10000) => {
...
 Cypress.Commands.add('resetViewport', () => {
-  cy.get('@resetBtn').click()
+  cy.initCornerstoneToolsAliases();
+  cy.get('@resetBtn').click();
+  //Click on More button
+  cy.get('@moreBtn')
+    .click();
+  //Verify if overlay is displayed
+  cy.get('.tooltip-toolbar-overlay')
+    .should('be.visible');
```
dannyrb: Hmm.. If it's not, we could toggle it again?

platform/viewer/cypress/integration/common/OHIFStudyViewer.spec.js
```diff
@@ -15,33 +22,80 @@ describe('OHIF Study Viewer Page', () => {
...
+                .then(() =>{
+                  cy.get('.treeInputs > .wrapperLabel') //select 'Bone' result
+                    .contains('Bone')
+                    .click()
+                    .then(() => {
+                      cy.get('.checkIconWrapper').should('be.visible').click(); //confirm the selection on overlay
+                    });
+                });
+            });
+        });
```
dannyrb: My biggest hang-up is the nesting here. Does something like this work?

```js
cy.addLengthMeasurement(); //Adding measurement in the viewport
      cy.get('@measurementsBtn').click();
      cy.get('.measurementItem').click();

      // Click "Relabel"
      cy.get('.btnAction')
        .contains('Relabel')
        .click();

      // Search for "Bone"
      cy.get('.searchInput')
        .should('be.visible')
        .type('Bone');

      // Select "Bone" Result
      cy.get('.treeInputs > .wrapperLabel')
         .contains('Bone')
         .click();

       // Confirm Selection
       cy.get('.checkIconWrapper')
          .should('be.visible')
          .click();
```

Alternatives for selecting if you feel locked in to wrap:

- https://github.com/cypress-io/cypress/issues/3745#issuecomment-474211567

Are we using the `should('be.visible')` checks to wait for the animations?

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -401,6 +383,114 @@ describe('OHIF Cornerstone Toolbar', () => {
         //      .its('length')
         //      .should('be.eq', 1);
         //  })
+        cy.reload();
```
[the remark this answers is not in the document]
mirnasilva: Yes, we are going to keep the reload just while we don't have the fix for https://github.com/OHIF/Viewers/issues/999

platform/viewer/cypress/support/commands.js
```diff
@@ -118,13 +130,24 @@ Cypress.Commands.add('waitDicomImage', (timeout = 10000) => {
...
+  cy.get('@resetBtn').click();
+  //Click on More button
+  cy.get('@moreBtn')
+    .click();
+  //Verify if overlay is displayed
+  cy.get('.tooltip-toolbar-overlay')
+    .should('be.visible');  
+  //Click on Clear button
+  cy.get('.tooltip-inner > :nth-child(10)')
+    .click();
```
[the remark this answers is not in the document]
mirnasilva: Yes, but only on this function scope because by the moment we initialize all aliases, this element is not visible. So a replacement would be:
```
cy.get('.tooltip-inner > :nth-child(10)').as('clearBtn')
    .click();
```

platform/viewer/cypress/support/commands.js
mirnasilva: Yep, we can! I added a conditional test for that.

platform/viewer/cypress/integration/common/OHIFStudyViewer.spec.js
mirnasilva: Great callout! I replaced the nesting for your suggestion, they both work fine.
You have nice point for the `should('be.visible')` checks. I was using it to make sure that we are clicking on the right-visible element and not on something behind in case it was not visible, but I just realized that the element MUST be visible in order to be retrieved by `cy.get`, therefore, `should('be.visible')` is redundant here. We can remove them. :)

dannyrb:
It looks like we're getting a failure: https://circleci.com/gh/OHIF/Viewers/4406?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

```
OHIF Cornerstone Toolbar checks if Eraser tool will remove the measurements added in the viewport:
     CypressError: Timed out retrying: expected 2 to equal 1
```

OHIF Cornerstone Toolbar -- checks if Eraser tool will remove the measurements added in the viewport (failed) (1)

OHIF Cornerstone Toolbar -- checks if Eraser tool will remove the measurements added in the viewport (failed)

Could be caused by a difference in screen resolution/size between you and the CI server for erase clicks? I'm not sure what it uses by default, or how you're testing.

mirnasilva:
@dannyrb , just pushed the fix for that!
Actually, it was failing because of the right panel. Once the panel is expanded the viewport became smaller and the position of each measurement also changed. We just need to collapse the panel before clicking to erase.

dannyrb:
Great catch @mirnasilva! :D

PR #1018, opened 2019-10-09
fix: 🐛 set current viewport as active when switching layouts

igoroctaviano:
check if current viewport index is less than the current layout lenght
and set 0 if so or keep current selected cell index as active viewport

Closes: 999


dannyrb:
includes workaround for: https://github.com/OHIF/Viewers/issues/1019

platform/viewer/src/connectedComponents/ConnectedLayoutButton.js
```diff
@@ -36,17 +37,23 @@ const mapDispatchToProps = dispatch => {
         viewports,
       };
 
+      const { row, col } = selectedCell;
+      const selectedCellIndex = (row * currentLayout.viewports.length + col) || 0;
+      if (selectedCellIndex < activeViewportIndex) {
+        dispatch(setViewportActive(0));
+      }
```
dannyrb:
```
      const maxActiveIndex = rows * columns - 1;
      if (activeViewportIndex > maxActiveIndex ) {
        dispatch(setViewportActive(0));
      }
```

igoroctaviano: done!

PR #1020, opened 2019-10-09
fix: 🐛 set tools bidirectional, eraser and annotate command

igoroctaviano:
annotate, bidirectional and eraser had types as command but they should
be typed as setToolActive

Closes: 981


dannyrb:
This looks good. I'll sneak it in before we merge your other one 🎉

PR #1026, opened 2019-10-10
refactor: layout manager to viewer

dannyrb:
- `LayoutManager` out of UI and into `viewer`
    - It leverages core's extension manager to grab viewports, and provides minimal reusable code. As an extension point, it makes more sense in `viewer`
- `LayoutManager` renamed `ViewportGrid`
- `ViewportGrid` refactored to use css-grid
- `ViewportGrid` refactored to use `numRows` and `numColumns` for creating grid
    - Future viewports can specify `colSpan` and `rowSpan` to take up multiple cells
- Functional Components and Hooks where possible to clean things up a bit

There will likely be no discernable difference between this and the `master` viewer. It's mostly clean-up. Additional changes on top of this will improve drag-n-drop performance.


docs/latest/advanced/extensions.md
```diff
@@ -140,8 +140,8 @@ children: PropTypes.node,
 customProps: PropTypes.object
 ```
 
-Viewport components are managed by the `LayoutManager`. Which Viewport component
```
JamesAPetts: Why rename something that's existed for such a long time?

You may have refactored how it operates, but the `ViewportGrid` is essentially the replacement for the layout manager. It manages layout, so why not keep the old name?

JamesAPetts:
Clicking mpr2d bring up one vtkjs viewport.

PR #1027, opened 2019-10-10
fix: E2E Series Metadata

rodrigolabs:
This is an implementation to wait for the study series to be loaded.

The current implementation we're using in `OHIFStudyViewer` waits for the image to be rendered. So, it's possible for the first viewport to render an image before any other thumbnails are showing (if the PACS has really slow metadata requests).

https://github.com/OHIF/Viewers/blob/a36cb63926d2cf439e7e68f29be160cc4799eaa3/platform/viewer/cypress/integration/common/OHIFStudyViewer.spec.js#L4

`waitDicomImage` works now because it buys us more time, and will probably continue to work so long as we have a PACS that isn't super slow, but could still cause the occasional intermittent failure, e.g.:

- First series loads
- All other series are requested asynchronously (running in background)
- First image is displayed
- Test fails ❌
- Other requested series resolve
- Thumbnails 2+ are shown


dannyrb:
Slack notes:

- We can't know number of thumbnails that will be rendered based on meta
- We _will_ know how many we care about at time of testing (context)
- Have "minimum" thumbnails to wait for as part of command
- Allow user to pass in a custom number

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -79,11 +80,11 @@ describe('OHIF Cornerstone Toolbar', () => {
 
     //drags the mouse inside the viewport to be able to interact with series
     cy.get('@viewport')
-      .trigger('mousedown', 'top', { which: 1 })
-      .trigger('mousemove', 'center', { which: 1 })
+      .trigger('mousedown', 'center', { which: 1 })
```
rodrigolabs: changed click from `center` to `top` because it seems sometimes the click was not working properly. Clicking first in the center is more assertive.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -431,11 +433,15 @@ describe('OHIF Cornerstone Toolbar', () => {
     cy.addAngle('.cornerstone-canvas', initPos, midPos, finalPos);
 
     //Verify if measurement annotation was added into the measurements panel
-    cy.get('@measurementsBtn').click();
+    cy.get('@measurementsBtn')
+      .click()
+      .wait(1000);
```
mirnasilva: Can you try replacing this `wait` for: `.should('be.visible')`?

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -431,11 +433,15 @@ describe('OHIF Cornerstone Toolbar', () => {
...
     cy.get('.measurementItem')
       .its('length')
       .should('be.eq', 2);
-    cy.get('@measurementsBtn').click();
+    cy.get('@measurementsBtn')
+      .click()
+      .wait(1000);
```
mirnasilva: Can you try replacing this `wait` for: `.should('not.exist')`?

dannyrb:
Closing, as discussed

PR #1030, opened 2019-10-11
fix: Switch token storage back to localStorage because in-memory was …

swederik:
…annoying for end users


platform/viewer/src/utils/getUserManagerForOpenIdConnectClient.js
```diff
@@ -21,17 +20,12 @@ export default function(store, oidcSettings) {
     return;
   }
 
-  // Do not store tokens in localStorage or sessionStorage
-  // https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/HTML5_Security_Cheat_Sheet.md#local-storage
-  const userStore = new WebStorageStateStore({ store: new InMemoryWebStorage() });
```
dannyrb: 🙌

PR #1031, opened 2019-10-11
ci: Set cache key to only use yarn.lock checksum

swederik:
We make a lot of changes but don't touch the lockfile on every PR. I don't think we need to include the branch name in the cache key. I'm checking this because the initial Checkout step in the CI process is taking 6 minutes to restore and then reupload the cache, which seems crazy: https://circleci.com/workflow-run/3938d5fd-3c82-4ff1-957e-6a04dd577df7

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -295,7 +295,7 @@ workflows:
           store_artifacts: false
           working_directory: platform/viewer
           build: yarn run build:e2e
-          start: yarn run test:e2e:dist
+          start: yarn run test:e2e:serve
           wait-on: 'http://localhost:3000'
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -90,9 +151,9 @@ jobs:
       - codecov/upload:
           file: '/home/circleci/repo/platform/core/coverage/reports'
           flags: 'core'
-
+      
       # Persist :+1:
-      - persist_to_workspace:
+      - persist_to_workspace: &persist_to_workspace
```
dannyrb: Can probably remove this anchor

PR #1034, opened 2019-10-11
feat: 🎸 Load spinner when selecting gcloud store. Add key on td

### LINKED ISSUE #1029: Study List data formatting
@pieper has a DICOM store on Google Cloud which fails every time it gets to Page 2 of the study list. It looks like it's because React is trying to render an Object inside one of the table cells in the study list. We need to ensure that each of the fields in the study object used below are renderable as Strings by React.

https://github.com/OHIF/Viewers/blob/659649e10b3732d4ce9a0d851390bb75786727eb/platform/ui/src/components/studyList/StudyList.js#L224-L253

> Uncaught Invariant Violation: Objects are not valid as a React child (found: object with keys {}). If you meant to render a collection of children, use an array instead.
    in td (created by StudyList)
    in tr (created by StudyList)
    in tbody (created by StudyList)
    in table (created by StudyList)
    in div (created by StudyList)
    in div (created by StudyList)
    in StudyList (created by TranslatedComponent)
    in TranslatedComponent (created by withI18nextTranslation(TranslatedComponent))
    in withI18nextTranslation(TranslatedComponent) (created by WrapperComponent)
    in I18nextProvider (created by WrapperComponent)

ladeirarodolfo:
### Includes
- Added loader spinner into study list when searching/filtering list
- Added React keys to table td tags
- Robust implementation preventing errors when cell value is not a string(it fixes the issue 1029)
- Reduced amount of component render
- Minor improvement on Cypress implementation

It closes #1029


platform/ui/src/components/studyList/StudyList.js
```diff
@@ -16,6 +16,69 @@ import { withTranslation } from '../../utils/LanguageProvider';
...
+
+function getNoListFragment(studies, error, loading) {
+  if (loading) {
+    return (
+      <div className="loading">
+        <StudyListLoadingText />
+      </div>
+    );
+  } else if (error) {
+    return <div className="notFound">There was an error fetching studies</div>;
```
swederik: Can we wrap these in translation strings (this.props.t()) so they can be internationalized?

platform/ui/src/components/studyList/StudyList.js
```diff
@@ -399,25 +441,23 @@ class StudyList extends Component {
               </tr>
             </thead>
             <tbody id="studyListData">
-              {this.props.studies.map(study => {
-                return this.renderTableRow(study);
-              })}
+              {!noListFragment
+                ? this.props.studies.map((study, index) => {
```
swederik: I think you can just write `this.props.studies.map(this.renderTableRow)`

dannyrb:
Netlify:

image

I have the branch checked out locally -- taking a peek.

ladeirarodolfo:
I have the fix for CI failing.. I will push soon, after some technical discussion with QA team

platform/ui/src/components/studyList/StudyList.js
```diff
@@ -110,25 +177,27 @@ class StudyList extends Component {
 
   getBlurHandler(key) {
     return event => {
+      debugger;
```
swederik: Forgot a debugger

platform/ui/src/components/studyList/StudyList.js
```diff
@@ -177,17 +177,38 @@ class StudyList extends Component {
...
     return event => {
-      debugger;
       this.delayedSearch.cancel();
       this.setSearchData(key, event.target.value);
     };
   }
 
+  // Should update state method.
+  // Used to prevent relying on shouldComponentUpdate React method or Pure component implementations of shallow comparison.
+  shouldUpdateStateValue(stateValue, stateValueToCompare) {
```
swederik: Can / should we use something like lodash isEqual instead?

dannyrb:
CI is passing

Issue #1024, opened 2019-10-10
Exit MPR button


bhavesh-pandey:
Hi,

Can we have a specific MPR exit button on MPR view?

Thanks

dannyrb:
We've been having internal talks about being able to register different layouts, and the UI/UX required to make them work well. Do you have any suggestions? We're very receptive to sketches, comments, and other applications you think handle this well (as an example).

karthikvrao:
Hi, @bhavesh-pandey @dannyrb. My suggestion is that to display a button to exit MPR when MPR is clicked. So, in effect it will be a toggle. I will submit a PR soon for this.

PR #1037, opened 2019-10-12
feat: Implement a 'Exit 2D MPR' button in the toolbar to act as a tog…

karthikvrao:
Refers to issue #1024.


platform/viewer/src/connectedComponents/ConnectedExitPluginSwitch.js
```diff
@@ -0,0 +1,64 @@
...
+      height: `${100 / rows}%`,
+      width: `${100 / columns}%`,
+      plugin,
+    });
+
+    const layout = {
+      viewports,
+    };
+
+    setLayout(layout);
```
dannyrb: I believe setLayout takes:

```js
{
  numRows: 1,
  numColumns: 1,
  viewports: [{ plugin: 'plugin-type' }],
}```

File: platform/viewer/src/connectedComponents/ToolbarRow.js [hunk 2]
```diff
@@ -128,7 +134,15 @@ class ToolbarRow extends Component {
...
           <ConnectedLayoutButton />
-          <ConnectedPluginSwitch />
+          {this.state.isPlugSwitchOn ?
+            <ConnectedExitPluginSwitch
+              togglePlugSwitchStatus={this.togglePlugSwitchStatus}
+            />
+            :
+            <ConnectedPluginSwitch
+              togglePlugSwitchStatus={this.togglePlugSwitchStatus}
+            />
```
dannyrb: Instead of toggling the component, can we toggle a prop on the component that changes the icon / label text / action?

platform/viewer/src/connectedComponents/ConnectedPluginSwitch.js
```diff
@@ -35,11 +35,13 @@ const mapStateToProps = state => {
 const mergeProps = (propsFromState, propsFromDispatch, ownProps) => {
   //const { activeViewportIndex, layout } = propsFromState;
   //const { setLayout } = propsFromDispatch;
+  const { togglePlugSwitchStatus } = ownProps;
 
   // TODO: Do not display certain options if the current display set
   // cannot be displayed using these view types
   const mpr = () => {
-    commandsManager.runCommand('mpr2d');
+        commandsManager.runCommand("mpr2d");
```
dannyrb: We could fire a different command if we know the togglePlugSwitchStatus? One that exits that viewing mode? Or, I think we can pass an object of options to the command we're calling "runCommand" for? If not, that shouldn't be a hard change, and you can have a single command that could handle the exit.

Just thinking of ways to reuse the code instead of adding a whole new component to handle this.

platform/viewer/src/connectedComponents/ConnectedExitPluginSwitch.js
karthikvrao: I just borrowed the code related to setLayout from ConnectedLayoutButton.js as it checked for `currentLayout.viewports[i]` and then set the plugin. If you suggest that this is unnecessary and something like below is sufficient, I will make that change.
```
{
  numRows: 1,
  numColumns: 1,
  viewports: [{ plugin: 'cornerstone' }],
}
```

platform/viewer/src/connectedComponents/ToolbarRow.js
karthikvrao: I agree. I had thought of keeping the toggleStatus private to the ConnectedPluginSwitch component and let it change the onClick action and icon/label, so as to become a toggling button in itself. But, that would directly alter the function of ConnectedPluginSwitch and I wasn't sure if that was acceptable.

karthikvrao:
@dannyrb please check the updated code.

dannyrb:
@karthikvrao thanks for the high quality PR! ^_^

There are still a few small issues with navigating between vtk and cornerstone viewports, but I believe they were existing issues and outside the scope of this ticket. If you're interested in shoring those up as well, let me know, and I'm more than happy to provide guidance/details.

_pulls merge lever_

karthikvrao:
Thanks @dannyrb. I am happy to contribute. 😄

PR #1042, opened 2019-10-14
Reliably show SR (structured report) series

cmcgee-mac:
Before this fix, SR series for a study show up sometimes, other times they cannot be seen in the left panel of the viewer.

The documentation for the _createDisplaySetsForSeries function says that it should return the display sets, however there is one case where it doesn't do that. This fixes the case so that it returns the display sets.

### DIFF
--- a/platform/core/src/classes/metadata/StudyMetadata.js
+++ b/platform/core/src/classes/metadata/StudyMetadata.js
@@ -121,7 +121,7 @@ export class StudyMetadata extends Metadata {
       if (displaySet) {
         displaySet.sopClassModule = true;
         displaySets.push(displaySet);
-        return;
+        return displaySets;
       }
     }
 


dannyrb:
@cmcgee-mac, thanks for this! I'll merge this, but tagging @emfol to make sure I'm not breaking anything he intended with the `return;`

cmcgee-mac:
@dannyrb I hope that the fix is correct. It took me a while to get a reproducible case where the SR was consistently not showing up. Sometimes it would show up, sometimes not.

dannyrb:
@cmcgee-mac, I think that's behavior I would expect. `sopClassHandlerModule`'s should have the ability to determine how each series is split into display sets. If it returns a `displaySet`, but we just return there, then it only works if we modify the `givenDisplaySets` array by reference (which we probably shouldn't do anyway).

We really should have unit tests for this portion of our code.

PR #1044, opened 2019-10-14
Initial Prototype Draft: Presentation State to Measurements

cmcgee-mac:
Read presentation state data for circle and simple two point lines and convert them into CircleRoi and Length measurements on the viewer. This is a draft PR that demonstrates the approach behind the prototype. The code structure is not yet in a production ready state. There are no tests yet. Much of the presentation state capabilities aren't there yet.


swederik:
Thanks for posting this!

Tagging @pieper because we just discussed PR recently and @JamesAPetts because he is going to deal with "non-image DICOMs (SR, PR, SEG, RTSTRUCT) which will provide data to some measurement service" soon.

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -395,6 +395,72 @@ async function makeSOPInstance(server, study, instance) {
...
+    //  text tool.
+
+    const graphics = [];
+    DICOMWeb.getSequence(layer['00700009']).forEach((g) => {
+       if (DICOMWeb.getString(g['00700005']) != "PIXEL") {
+         // Only PIXEL coordinates are supported (not FRACTION)
+         return;
+       }
+       if (DICOMWeb.getString(g['00700024']) != "N") {
+         // There's no support for filled shapes in the viewer
```
swederik: @JamesAPetts Maybe as a label map?

pieper:
This is a great addition 👍   Can you provide testing data?  Maybe a description of what you do/don't support or plan to?

cmcgee-mac:
I have a tool that will generate some GSPS examples for a study. Is there a place where we can put OHIF viewer sample studies? Maybe they can be placed on the OHIF viewer demo site?

This prototype only handles circles and two-point simple lines. It should be straight forward to add ellipse, rectangle and n-point polylines. Adding text annotations, layers, points, translucent color filled shapes and interpolated (curved) lines would require some additional support from cornerstone.

swederik:
> I have a tool that will generate some GSPS examples for a study. Is there a place where we can put OHIF viewer sample studies? Maybe they can be placed on the OHIF viewer demo site?
> 

Yeah sure just send them to me and I will put them up.

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -206,7 +206,119 @@ class Viewer extends Component {
       const patientId = studies[0] && studies[0].patientId;
       const currentTimepointId = this.currentTimepointId;
 
+      // Translate presentation state information into measurement/tool state
+      studies.forEach((study) => {
+      study.seriesList.forEach((series) => {
```
swederik: We needs to figure out where this logic should go. Right now we have some of it in the Cornerstone extension (for DICOM Seg), and some in the measurement extension (in an ongoing PR for DICOM SR).

swederik:
So a while back with @pieper and @pavertomato we had initially discussed it as follows:
- We would do import/export of measurements with DICOM SR. If a TID1500 SR was ingested, the Cornerstone Tools would be re-hydrated and the user could move them around and re-save them.
- If we encountered a DICOM PR, we would just render each of the graphics, but not allow them to be re-hydrated as Cornerstone Tools (Length, Ellipse, etc..) and manipulated.

Looking at your PR, though, it seems like we could attempt to create CornerstoneTools measurements using either SR or PR.

So I guess we should do this:
- If we encounter a PR, we check each of the graphic types and number of points to identify if it has a matching tool that we support (e.g. length, ellipse, free text)
- If it doesn't match a tool that we support, or it's ambiguous, we render the graphic as a generic polyline + free text
- Only rehydrate measurements from the most recent series (applies to SEG, SR, PR, RTSTRUCT). The others will be done on demand by the user

So I guess we need to create Cornerstone Tools for polyline (just renders an array of points), and free text (text at a specific location).

Questions:
- Do we display something in the Measurement table for something that doesn't match one of our known tools?
- What are we doing about Graphic Layer Order if we are rehydrating things into measurements? That isn't supported in Cornerstone. Maybe if this is set, we just make everything generically rendered?
- Is there anything salvageable from this PR (https://github.com/OHIF/Viewers/pull/325/commits) where Presentation State support was briefly available in the Meteor version of OHIF? Note: you probably have to go back through the commits because we removed it prior to merging because it wasn't quite finished yet.
 
We should look at how Weasis does this, they do a pretty good job. We can also use it to generate some sample data.

swederik:
> I didn't know until now that SR's could have graphics in them. :)

Yeah there is a template (TID1500 Measurement Report) for quantitative imaging which will allow us to specify these things correctly. We have some support for it in dcmjs (e.g. https://github.com/dcmjs-org/dcmjs/blob/master/src/utilities/TID300/Length.js) and plan to continue developing it.

- As far as I know, the Cornerstone Freehand ROI tool requires that the form is a closed shape, so I think we will need another (very similar) tool for a polyline. @JamesAPetts would know more though.

> For example, the way that rectangles are defined the points are upper-left and bottom-right. 

- Do you mean in PR or in Cornerstone? We can write an adapter to save four lines to the PR if necessary. 

> What's missing is support for interpolated (curved) lines.

Yeah that is not present in Cornerstone Tools yet. Horos has a good implementation though someone could use that for inspiration to develop a contour-based tool.

> It does seem to perform a little slowly in cases that I've tried.

Slowly in what way?

Thanks for digging up the Osirix approach. It seems like there is a bit a low bar for presentation state graphic support based on your research. I think Weasis does it better. I don't know about commercial PACS though.

> My preference based on the project I'm working on now would be to at least be able to capture any new measurements that the user makes and save those back as PR's into the study using STOW-RS.

Great! That's exactly our plan with DICOM SR. See https://github.com/OHIF/Viewers/pull/954

> For existing measurements, I was thinking of marking them with a dull colour (unless the colour is specified in the PR) and potentially marking them unmodifiable, but then there would need to be a way to hide them in case they are in the way. Maybe a show/hide functionality for measurements that didn't originate with the user? Alternatively, save only save measurements modified or created by the user in the current session.

That sounds OK but I suppose it will quickly result in tons of dull measurements whenever a user modifies something, saves, and reloads. You may note that some of the studies on viewer.ohif.org have dozens of SRs in them because this STOW-RS behaviour was introduced in the Meteor OHIF version and a lot of people were testing it.

cmcgee-mac:
> * As far as I know, the Cornerstone Freehand ROI tool requires that the form is a closed shape, so I think we will need another (very similar) tool for a polyline. @JamesAPetts would know more though.

True, that would be a mismatch between PR and the freehand tool. Whereas polyline will allow open shapes unless you want to fill them.

> > For example, the way that rectangles are defined the points are upper-left and bottom-right.
> 
> * Do you mean in PR or in Cornerstone? We can write an adapter to save four lines to the PR if necessary.

I was referring to the idea of falling back to freehand based on the polyline data in PR if we can't find another more suitable measurement tool in Cornerstone. I don't know if it will work in all cases. Plus, we have the non-closed problem. I wonder if it may be best to skip unsupported PR graphics and report that we skipped it somewhere that can be checked if there's a missing shape.

> > It does seem to perform a little slowly in cases that I've tried.
> 
> Slowly in what way?

I noticed that when I traced the whole contour of a tibia in one of my scans that when closing the shape that there's a noticeable 2 second delay while cornerstone updates. My guess is that I drew alot of points and there's some heavy computation in there to calculate the area, maybe?

> Great! That's exactly our plan with DICOM SR. See #954

I wonder how it will be decided whether to save the measurement data to SR, PR or one of the other formats you have mentioned? Maybe different save buttons for each type?

> That sounds OK but I suppose it will quickly result in tons of dull measurements whenever a user modifies something, saves, and reloads. You may note that some of the studies on viewer.ohif.org have dozens of SRs in them because this STOW-RS behaviour was introduced in the Meteor OHIF version and a lot of people were testing it.

Yes, I see that the online demo is a pathological case, but one that needs to be supported in the viewer. :)

In PR at least, not sure yet about SR and others, there's a way to store the colour of each layer. I wonder if a calculated unique colour could be found based on existing layers for the "session." Then at least each "session" could be discerned on the image.

cmcgee-mac:
I was thinking about this a bit more and re-read some of the DICOM PR specification.

The spec mentions that not all imaging equipment will support first class layers. I think that we can get away without supporting them like OsiriX and technically be compliant.

There are certain graphic elements that we won't be able to support in the near and even in the long term. For example, open polylines are not yet supported by FreehandRoi. Meanwhile, the spec says that vendors can have their own custom graphics not described in the specification.

I propose that for all non-supported graphic elements that we calculate a bounding box from the points in the graphic and draw a RectangleRoi with text indicating that there is an unsupported graphic there. Details are logged in the web browser console if someone needs it.

There is the problem of defining measurements to be saved and ones that should not. Also, the related problem of displaying where they come from.

My proposal is the following. Faithfully colour each graphic in what has been specified in the PR, even if it is greyscale and mark it as "saved." However, when the user draws new measurements we pick an unused colour from a palette of colours (maybe 15-20) with reasonable contrast to each other and those measurements are marked as "not saved." If the user modifies an existing measurement then the colour changes to the current colour and is marked as not saved. Measurements can be deleted from either clicking on them in the viewer or from the measurements panel so that they can be removed from view.

For saved measurements we put the series description in the measurement panel so that the user can see from a glance where each one comes and click to highlight it in the view. For PR we can even use heuristics to discover the name of the person who made it, which could also be displayed.

When the user clicks "Save as PR" on the measurements panel all "not saved" measurements are saved as a new series in the study with the description "Measurements by <username>" in a single layer in the chosen colour. The username of the logged-in user is stored in the DICOM metadata. All measurements are marked as "saved" and a new colour is chosen for any new measurements after the save.

How does this sound?

dannyrb:
@cmcgee-mac, @swederik is a little tied up at the moment. In your opinion, what needs discussed/actioned before this PR can be merged? I want to make sure it doesn't become too outdated and we miss out on this valuable work ^_^

cmcgee-mac:
@dannyrb thanks, I think there are several design items that need to be worked out before this can be merged.

I realize now that the PR annotations are not intended to be measurements, but this patch translates them into cornerstone tools that are intended for measurement. There are dimensions displayed on the cornerstone for example depending on the type of tool and this isn't usually the case for PR in other viewers.

This patch intends to use a color map to show multiple PR series on the same image so that they can be recognized. To complete this capability there needs to be some kind of legend so that the color can be related back to the series where it came. Otherwise, all graphics should use the one specified in the PR layer.

There is a still some incomplete PR shapes and some mismatches. I think that for shapes that cannot be represented in cornerstone at the moment we could probably get away way drawing a bounding box rectangle with "unsupported shape" text in it. The only problem with this is that the cornerstone rectangle roi tool doesn't support inner text that isn't a measurement of the shape drawn.

Floating text is not yet supported, but could probably be treated like other unsupported shapes with a bounding box.

There is still a mismatch between open and closed polygons, the former of which is not supported by the cornerstone freehand tool. Again, this can probably be rendered with a bounding box for now.

cmcgee-mac:
The way that this patch retrieves the data necessary from the PR series is different that what is being used for the TID 300 pull request, which directly fetches the DICOM later on. It probably makes sense to use the same approach for both PR and SR.

dannyrb:
SR POC has been merged, but we're aware of plenty of edge cases that need to be addressed. Commenting mostly to scare away the `stale` bot ^_^

dannyrb:
We're working on a measurement service, here: https://github.com/OHIF/Viewers/pull/1314

As it's difficult to map DICOM SR from an unknown source to edit/update tools, the fallback behavior is to map the measurements to a render only tool that supports all of the DICOM SR value types.

If we can map the PR to those value types, we can add a concept of layers/color to the service and potentially achieve the rendering required. If there is a way to reliably map graphic types to DICOM SR graphic types, let me know. We could also expand our list, and account for "unknowns" using the bounding box approach @cmcgee-mac has outlined.

sedghi:

sedghi:

PR #1055, opened 2019-10-16
Progressive loading of vtkjs viewport

### LINKED ISSUE #1051: Add progressive loading for Volume for MPR mode
Right now the user has to wait for Promise.all() for every image to be downloaded, parsed, decoded, and inserted into the volume before anything is rendered. This is unnecessary. Since we are forced to create the 3D volume buffer ahead of time anyway, we can start rendering it and just add the slice data into the volume as it comes in. We will need to make sure that re-rendering isn't triggered too often (e.g. not every slice) though so it's not too slow.

Overall we don't really care if the user can start interacting with the volume immediately, we just care that there is some indication that progress is occurring. Right now we just have a 'Loading...' which can take 15-20 seconds in some cases. This makes the user wonder if the app has crashed or something.

JamesAPetts:
Closes #1051

- Progressively load volumes in vtkjs. The volume and UI is still interactable whilst waiting on frames to come from the data source. The volume is re-loaded into the GPU every 20% of frames that are loaded.
- WWWC on viewport overlay correctly updates on change.
- Auto window level based on leveling of cornerstone viewport on mode switch.
- PT leveled from 0-5 SUV.


dannyrb:
@JamesAPetts is this ready for review?

extensions/vtk/src/OHIFVTKViewport.js
```diff
@@ -248,19 +253,50 @@ class OHIFVTKViewport extends Component {
...
+
+          // TODO: There must be a better way to do this.
+          // We do this so that if all the data is available the react-vtkjs-viewport
+          // Will render _something_ before the volumes are set and the volume
+          // Construction that happens in react-vtkjs-viewport locks up the CPU.
+          setTimeout(() => {
+            this.setState({
+              volumes: [volumeActor],
+            });
+          }, 200);
```
dannyrb: ☝️ 👉 👇 👈

JamesAPetts: I know, feel free to fix it!

extensions/vtk/src/LoadingIndicator.js
```diff
@@ -35,7 +35,8 @@ class LoadingIndicator extends PureComponent {
           <div className="imageViewerLoadingIndicator loadingIndicator">
             <div className="indicatorContents">
               <p>
-                Loading... <i className="fa fa-spin fa-circle-o-notch fa-fw" />{' '}
+                Reformatting...{' '}
```
swederik: Use the translation function for i18n, and remove that {' '} thing

extensions/vtk/src/OHIFVTKViewport.js
```diff
@@ -172,47 +172,53 @@ class OHIFVTKViewport extends Component {
       default:
         imageDataObject = getImageData(stack.imageIds, displaySetInstanceUid);
 
-        const loadImageDataPromise = loadImageData(imageDataObject);
+        //loadImageData(imageDataObject);
```
swederik: I guess you can remove this line

extensions/vtk/src/OHIFVTKViewport.js
```diff
@@ -248,19 +253,50 @@ class OHIFVTKViewport extends Component {
...
+      displaySetInstanceUid
+    );
+
+    this.setState(
+      {
+        paintFilterBackgroundImageData: imageDataObject.vtkImageData,
+        paintFilterLabelMapImageData: null, // TODO
+        percentComplete: 0,
+      },
+      () => {}
```
swederik: useless callback

extensions/vtk/src/OHIFVTKViewport.js
```diff
@@ -285,6 +321,56 @@ class OHIFVTKViewport extends Component {
...
+          if (percentComplete !== this.state.percentComplete) {
+            this.setState({
+              percentComplete,
+            });
+          }
+
+          // TODO -> Just do this higher up, call when loadImageData is first
+          // called and then do this once after all apis are made.
+          if (
+            this.props.viewportIndex === 0 &&
```
swederik: Why does viewportIndex matter?

extensions/vtk/src/utils/setMPRLayout.js
```diff
@@ -1,11 +1,21 @@
...
+  numColumns = 3
+) {
   return new Promise((resolve, reject) => {
     const viewports = [];
-    const numRows = 1;
-    const numColumns = 3;
     const numViewports = numRows * numColumns;
+
+    if (viewportPropsArray && viewportPropsArray.length !== numViewports) {
+      reject(
```
swederik: Can we reject with `new Error('viewportProps is supplied but its length is not equal to numViewports')` instead of just a string?

extensions/vtk/src/utils/setMPRLayout.js
```diff
@@ -42,3 +53,65 @@ export default function setMPRLayout(displaySet) {
     );
   });
 }
+
+/*
```
swederik: Just remove this instead of leaving a massive comment block

extensions/vtk/src/OHIFVTKViewport.js
JamesAPetts: Moved down here: https://github.com/OHIF/react-vtkjs-viewport/pull/56

extensions/vtk/src/utils/setMPRLayout.js
JamesAPetts: Oops!

JamesAPetts:
Addressed issues and removed RAF successfully.

Just need to wait for `react-vtkjs-viewport` to publish and I'll merge this.

PR #1059, opened 2019-10-17
Adding Percy Snapshot on VTK, PDF and Microscopy tests

mirnasilva:
Adding Percy Snapshot on VTK, PDF and Microscopy tests.


platform/viewer/cypress/support/commands.js
```diff
@@ -54,6 +54,21 @@ Cypress.Commands.add('openStudy', patientName => {
...
+ */
+Cypress.Commands.add('openStudyModality', modality => {
+  cy.initRouteAliases();
+  cy.visit('/');
+  cy.get('#modalities').type(modality);
+
+  cy.get('#studyListData > tr:nth-child(1)', { timeout: 5000 })
+    .contains(modality)
+    .click();
+});
```
dannyrb: Oh, this is neat :+1:

mirnasilva:
> Some issues clicking a study. Could be related to `onBlur` issue. I'll look into it:
> 
> * https://5033-44164233-gh.circle-artifacts.com/0/root/project/platform/viewer/cypress/screenshots/pwa/OHIFExtensionVTK.spec.js/OHIF%20VTK%20Extension%20--%20checks%20if%20VTK%20buttons%20are%20displayed%20on%20the%20toolbar%20--%20before%20all%20hook%20%28failed%29.png
> * https://5033-44164233-gh.circle-artifacts.com/0/root/project/platform/viewer/cypress/screenshots/pwa/OHIFExtensionMicroscopy.spec.js/OHIF%20Microscopy%20Extension%20--%20checks%20if%20series%20thumbnails%20are%20being%20displayed%20--%20before%20all%20hook%20%28failed%29.png

I noticed that it's failing only when we have more than 1 result. Could be a coincidence, but just in case I added a `.first()` to make sure we are clicking on the 1st result.

mirnasilva:
`VTK` test is failing, since `VTK toolbar` is not being displayed. If this behavior is intended, we may want to disable all `VTK` tests for now.

DONg1Ulsig

Once this is solved, PR_Checks should pass with success.

dannyrb:
I reverted the PR that was causing the build failure, but it seems we're getting a new one: [Example Image](https://5152-44164233-gh.circle-artifacts.com/0/root/project/platform/viewer/cypress/screenshots/pwa/OHIFExtensionVTK.spec.js/OHIF%20VTK%20Extension%20--%20checks%20if%20VTK%20buttons%20are%20displayed%20on%20the%20toolbar%20--%20before%20each%20hook%20%28failed%29.png)

platform/viewer/cypress.json
```diff
@@ -2,5 +2,9 @@
   "baseUrl": "http://localhost:3000",
   "video": false,
   "chromeWebSecurity": false,
+  "waitForAnimations": true,
```
dannyrb: Ooh. Didn't know this was a flag. Good catch.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -216,26 +219,49 @@ describe('OHIF Cornerstone Toolbar', () => {
       .click();
 
     let expectedText = 'Img: 1 1/26';
-    cy.get('@viewportInfoBottomLeft').should('not.have.text', expectedText);
+    cy.get('@viewportInfoBottomLeft', { timeout: 15000 }).should(
```
dannyrb: Is there a reason these each need 15 second timeouts? That seems really high.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -216,26 +219,49 @@ describe('OHIF Cornerstone Toolbar', () => {
...
+      'not.have.text',
+      expectedText
+    );
 
     //Test SKIP TO FIRST IMAGE button
-    cy.get('[title="Skip to first Image"]').click();
-    cy.get('@viewportInfoBottomLeft').should('contain.text', expectedText);
+    cy.get('[title="Skip to first Image"]')
+      .click()
+      .wait(1000);
```
dannyrb: We probably don't need this wait if the `get` has a timeout that will continue to revalidate expected text for its duration?

platform/viewer/cypress/support/commands.js
```diff
@@ -41,17 +41,45 @@ import {
...
+ * Command to search for a modality and open the study.
+ *
+ * @param {string} modality - Modality type that we would like to search for
+ */
+Cypress.Commands.add('openStudyModality', modality => {
+  cy.initRouteAliases();
+  cy.visit('/');
+  cy.get('#modalities')
+    .type(modality)
+    .wait(2000);
```
dannyrb: Instead of an explicit wait here, we probably want a timeout for the below `cy.get` to contain the modality. If that's failing, it could be related to the on blur issue we identified earlier, but that it a "bug" (as far as I'm concerned) that we should work to resolve.

In general, if we have an explicit wait, we should try to identify why, as it's usually an indicator that we have something we need to create an issue for and resolve on the development side.

platform/viewer/cypress/integration/pwa/OHIFExtensionPDF.spec.js
```diff
@@ -1,23 +1,26 @@
 describe('OHIF PDF Extension', () => {
   before(() => {
     cy.openStudy('Dummy');
-    cy.expectMinimumThumbnails(1);
+    cy.expectMinimumThumbnails(3);
   });
 
   it('checks if series thumbnails are being displayed', () => {
-    cy.get('[data-cy="thumbnail-list"]')
+    cy.get('[data-cy="thumbnail-list"]', { timeout: 15000 })
```
dannyrb: If we're using the `cy.expectMinimumThumbnails` in the `before` hook, we may no longer need a timeout for these tests.

platform/viewer/cypress/integration/pwa/OHIFExtensionMicroscopy.spec.js
```diff
@@ -0,0 +1,27 @@
+describe('OHIF Microscopy Extension', () => {
+  before(() => {
+    cy.openStudyModality('SM');
+    cy.expectMinimumThumbnails(2);
+  });
+
+  it('checks if series thumbnails are being displayed', () => {
+    cy.get('[data-cy="thumbnail-list"]', { timeout: 15000 })
```
dannyrb: Same thing here. We're leveraging `cy.expectMinimumThumbnails`, so we should no longer need a `timeout` for these gets? They each already have a minimum timeout time applied; no need to expand it further.

platform/viewer/cypress/integration/pwa/OHIFExtensionMicroscopy.spec.js
```diff
@@ -0,0 +1,27 @@
...
+  it('drags and drop a SM thumbnail into viewport', () => {
+    cy.get('[data-cy="thumbnail-list"]', { timeout: 15000 })
+      .contains('SM', { timeout: 15000 })
+      .drag('.viewport-drop-target');
+
+    cy.get('.DicomMicroscopyViewer', { timeout: 15000 })
+      .its('length')
+      .should('be.eq', 1);
+
+    cy.wait(3000);
```
dannyrb: I think this wait actually makes sense, but a small comment above it to indicate we're waiting for the "renderer" to display our image, and we're using this in lieu of something more exact/formal.

platform/viewer/cypress/integration/pwa/OHIFExtensionHTML.spec.js
```diff
@@ -1,24 +1,26 @@
...
+    cy.get('[data-cy="thumbnail-list"]', { timeout: 10000 })
+      .contains('SR', { timeout: 10000 })
       .its('length')
       .should('to.be.at.least', 1);
   });
 
   it('drags and drop a SR thumbnail into viewport', () => {
-    cy.get('[data-cy="thumbnail-list"]')
-      .contains('SR')
+    cy.get('[data-cy="thumbnail-list"]', { timeout: 10000 })
```
dannyrb: Are these timeouts still needed?

dannyrb:
```
OHIF VTK Extension "before each" hook for "checks if VTK buttons are displayed on the toolbar":
     Uncaught TypeError: Cannot read property 'blendFuncSeparate' of null
```

dannyrb:
This appears to be a method that should live on the GL context?

https://github.com/Kitware/vtk-js/search?q=blendFuncSeparate&unscoped_q=blendFuncSeparate

dannyrb:
@mirnasilva tests are passing :+1:
After some light cleanup these should be good for merge.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
mirnasilva: Yeah, unfortunately, every interaction with cine controls demands a new image to load in the viewport. Most of the times it is taking a long time to fully load the image.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
mirnasilva: Not actually. We are changing to another image in the viewport and we need to wait the image to be fully loaded. The `wait` here is actually waiting for the image to load, it is not related to the `get`.

platform/viewer/cypress/integration/pwa/OHIFExtensionHTML.spec.js
mirnasilva: We can remove these

platform/viewer/cypress/integration/pwa/OHIFExtensionMicroscopy.spec.js
mirnasilva: Hmm. Here the SM thumbnails are taking a really long time to load. Either we increase the number of expected thumbnails to `5` or maybe `7`, or we leave the `15s` timeout. Either way, we need to wait more time here.

platform/viewer/cypress/integration/pwa/OHIFExtensionPDF.spec.js
mirnasilva: Hmm. Here the PDF (`DOC`) thumbnails are taking a really long time to load. Either we increase the number of expected thumbnails to `5` or maybe `7`, or we leave the `15s` timeout. Either way, we need to wait more time here.

platform/viewer/cypress/support/commands.js
mirnasilva: We can say that the problem here to the asynchronism. If we only add the `timeout` on the `cy.get('#studyListData')`, we will either get an empty element (since the element is present in the DOM) or a full element (a full list of the patients listed in that page). If we get an empty element, it will definitely fail. If we get the full page list, it may fail (if the searched patient were not initially listed on the 1st page) or not. This may occur failures if we change/add more patient data.

With this `2s` wait, we are making sure that the captured element will contain the search result.

PR #1062, opened 2019-10-18
fix: MPR initialization

rodrigolabs:
Fixes #1022

Issue #1022, opened 2019-10-09
MPR Fails if initiated from viewport that was not populated by drag-n-drop


caoer:
> **Before Creating an issue**
>
> - Are you running the latest version? YES
> - Are you reporting to the correct repository? YES
> - Did you search existing issues? YES

## Bug Report

I'm using the official web url [https://viewer.ohif.org/viewer/1.3.6.1.4.1.14519.5.2.1.2744.7002.373729467545468642229382466905](https://viewer.ohif.org/viewer/1.3.6.1.4.1.14519.5.2.1.2744.7002.373729467545468642229382466905)

### Describe the Bug

trying to adjust slab thinkness, screen became black. 

### What steps can we follow to reproduce the bug?

please see screen recording at: https://kapture-fugspsudqf.now.sh

```js
Please use code blocks to show formatted errors or code snippets
```

> :warning: Reports we cannot reproduce are at risk of being marked stale and
> closed. The more information you can provide, the more likely we are to look
> into and address your issue.

dannyrb:
@caoer, thanks for the report. It looks like if you click `2D MPR`, what we're seeing are actually `cornerstone` viewports and not `vtk` viewports. 

If I drag and drop that same series into the first viewport, and then click `2D MPR`, everything works as expected. We don't notice this in most of our other studies because they tend to have a pilot (single image series, can't reconstruct) so you're forced to drag a series before initiating MPR.

The thumbnail drag-n-drop probably updates our `viewportSpecificData` with properties that our viewport "autofill" (for empty viewports) does not.

This is related to: https://github.com/OHIF/Viewers/issues/1005

PR #1065, opened 2019-10-19
Revert "Revert "fix: MPR initialization""

dannyrb:
Reverts OHIF/Viewers#1064

This is a Revert _of a revert_ PR. IE. It reapplies a PR we had to revert. Here are the reasons why we originally had to revert this code, and the items that need investigation:

> Reverts OHIF/Viewers#1062
>  
> We begin to see a new failure on  the CI Server when this fix was previously merged.
> 
> - [Example Failure Build](https://circleci.com/gh/OHIF/Viewers/5123)
> - [Example Screenshot](https://5123-44164233-gh.circle-artifacts.com/0/root/project/platform/viewer/cypress/screenshots/pwa/OHIFExtensionVTK.spec.js/OHIF%20VTK%20Extension%20--%20checks%20if%20VTK%20buttons%20are%20displayed%20on%20the%20toolbar%20--%20before%20each%20hook%20%28failed%29.png)
> 
> This impacts the bug that was fixed by PR #1062, here: [Issue #1022](https://github.com/OHIF/Viewers/issues/1022)
> 
> Interestingly, I've witnessed this issue before, but I've been unable to recreate it in the #1062's [deploy preview](https://deploy-preview-1062--ohif.netlify.com/)
> 
> IE. When kicking of MPR viewing, all of the viewports are `cornerstone` viewports instead of `vtk` viewports.


platform/viewer/cypress/support/commands.js
```diff
@@ -44,7 +44,8 @@ Cypress.Commands.add('openStudy', patientName => {
   cy.wait('@getStudies');
   cy.get('#studyListData .studylistStudy', { timeout: 5000 })
     .contains(patientName)
-    .first().click();
+    .first()
+    .click({ force: true });
```
rodrigolabs: I was getting some E2E errors while trying to find and click in a study. The study was being displayed but click was not working. `{ force: true }` solved this problem.

dannyrb: That's odd. I would have guessed it was because the `contains` should be wrapped in a should so it can leverage the timeout in the above `cy.get`. I wonder what's going on here that requires us to force a click. That usually means the found element wouldn't be clickable by the user (off screen, invisible, etc.)

platform/viewer/src/components/ViewportGrid/ViewportGrid.js
```diff
@@ -32,6 +32,11 @@ const ViewportGrid = function(props) {
 
   const ViewportPanes = layout.viewports.map((layout, viewportIndex) => {
     const displaySet = viewportData[viewportIndex];
+
+    if (!displaySet) {
+      return null;
+    }
```
dannyrb: I wonder if this is possible because we don't clear `viewportSpecificData` when our `numRows` or `numColumns` are updated?

platform/viewer/src/components/ViewportGrid/ViewportGrid.js
```diff
@@ -108,6 +113,9 @@ ViewportGrid.propTypes = {
   setViewportData: PropTypes.func.isRequired,
   studies: PropTypes.array,
   children: PropTypes.node,
+  defaultPlugin: PropTypes.string,
+  numRows: PropTypes.number,
+  numColumns: PropTypes.number,
```
dannyrb: numRows, and numColumns are probable `isRequired`?

platform/viewer/src/connectedComponents/ViewerMain.js
```diff
@@ -65,63 +63,75 @@ class ViewerMain extends Component {
...
+
+      this.fillEmptyViewportPanes();
     }
   }
 
   componentDidUpdate(prevProps) {
-    if (this.props.studies !== prevProps.studies) {
+    const prevViewportAmount = prevProps.layout.viewports.length;
+    const viewportAmount = this.props.layout.viewports.length;
+    const isVtk = this.props.layout.viewports.some(vp => !!vp.vtk);
```
dannyrb: So if any `layout.viewports` has a `vtk` property (what we cram our public API on to)
Then DO NOT `fillEmptyViewportPanes`

That's interesting. I think I get the gist of why, but it's a bit muddy.
If this works, great :+1:
I think we'll still have a fair bit of clean-up we can do as we iterate on this layout logic.

PR #1066, opened 2019-10-20
[LEGACY] feat: MR series stack sorting

alcheb:
Hi there. This is my first PR! :tada:
I hope I'm not doing anything wrong. This is a proposal for a slight modification of the stack sorting functions. I use it to work on liver MR and thought it would be cool to have this feature out-of-the-box. I took the liberty of modifying the code in-line. I'll let you judge if this feature needs its own separate file. My skills are limited for this and I appreciate your help.

# Overall purpose of this PR
To implement stack sorting in the case of multiple MR sequences or phases bundled in the same series.

# Description of the problem
In affected series, similar yet different MR sequences (different timing or acquisition parameters) are bundled together resulting in a "flickering" effect as all phases are viewed as one.

An example might be useful at this stage: https://viewer.ohif.org/viewer/2.25.63009447009646876712620988718732709379

All three series in this exam are affected by the problem :
  - in diffusion series, there is three different sequences with different B values
  - in the T1 sequence, three different timed acquisitions are mixed together

Can you see it?

# Modalities affected by the problem
MR exams (possibly a good number of those). I only encountered it in liver MR.
The problem may arise in:
  - diffusion series in which the diffusion factor (B) is changed;
  - timed acquisition of T1 sequences with contrasting agent (phases).

Other modalities might be affected (CT, PT).

# Why is this important to fix?
Series affected by this are flat-out unusable without a great deal of eye squinting. Diffusion series and timed acquisition with contrast are super important for diagnosing liver tumors (for investigational purposes of course :stuck_out_tongue_winking_eye:). Without a fix, we need another viewer to correctly interpret the exam.

# How other viewers fix it
I won't cite any names here as I don't remember the particulars. Some split these series into as many stacks as there are phases. Some others just sort them and let you scroll one phase after another.
These changes are made either automatically or to the end-user's input, depending on the viewer.

# Proposed solution
The most simple of those solutions would be sorting the series without splitting the stacks, because:
  - this is the first one I tried and managed to get working
  - we would need (would we?) to give appropriate names to the new stacks
  - there is already a sorting function built into OHIF ^_^

There isn't much code needed to implement this feature. It works like this :
  1. An MR modality is requested (imageSet.modality)
  2. All MR series are sorted by TemporalPositionIdentifier
  3. Tag _instance.DiffusionBValue is checked: if not undefined, this is possibly a diffusion series and needs sorting according to this tag.

To be able to do this, we just need to request those tags when retrieving series metadata.
You can see it working here:
https://deploy-preview-1066--ohif.netlify.com/pwa/viewer/2.25.63009447009646876712620988718732709379 :smile: 

# Liver MR specifics
Liver MR exams are subject to substantial changes in acquisition techniques based on machine used, centre, and dare I say fashion trends in the radiology world :smile:.
Hence liver MR are often made of a butt-load of different series with new/unknown naming conventions. Affected series are fortunately somewhat predictable (diffusion/T1 with contrast) but we may encounter now or in the future the occasional odd one that has this problem. This is why I'm suggesting using this on every MR series.

I hope this helps in understanding the reasoning behind this PR. I'm eagerly awaiting your comments.

Cheers,
Al


platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -615,6 +615,21 @@ const makeDisplaySet = (series, instances) => {
     });
   }
 
+  // If MR series, sort the images by TemporalPositionIdentifier
+  if (imageSet.modality == 'MR') {
```
dannyrb: I think we're okay to use `===` here

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -615,6 +615,21 @@ const makeDisplaySet = (series, instances) => {
...
+    imageSet.sortBy((a, b) => {
+      // Sort by TemporalPositionIdentifier
+      if (a._instance.TemporalPositionIdentifier > b._instance.TemporalPositionIdentifier) return 1;
+      if (a._instance.TemporalPositionIdentifier < b._instance.TemporalPositionIdentifier) return -1;
+
+      // Sort by DiffusionBValue if defined (Diffusion series)
+      if (typeof(instance._instance.DiffusionBValue) != 'undefined') {
```
dannyrb: Similarly, `!==`

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -615,6 +615,21 @@ const makeDisplaySet = (series, instances) => {
...
+        if (a._instance.DiffusionBValue > b._instance.DiffusionBValue) return 1;
+        if (a._instance.DiffusionBValue < b._instance.DiffusionBValue) return -1;
+      }
+    });
```
dannyrb: The goal here being:

- If `TemporalPositionIdentifier` is available, sort by it.
- If one or both `TemporalPositionIdentifier`s are undefined, or if they are equal, defer to `DiffusionBValue`

dannyrb:
First and foremost, thanks for the PR and for the amazing write-up ^_^

Second, and this is mostly for any reviewers, the E2E CI tests are failing, but it looks like the failures are caused by an assumption of MR slice ordering in the stack. These will likely need to be updated to work with the new ordering if/when we approve these changes.

While this is the easier implementation, as the end user, would you prefer that the stacks be split?

@swederik, do you have a preference on sort versus split? To codify this a bit, should we back these with some simple unit tests?

platform/core/src/classes/metadata/StudyMetadata.js
alcheb: Ok, will do

platform/core/src/classes/metadata/StudyMetadata.js
alcheb: Ok :+1:

platform/core/src/classes/metadata/StudyMetadata.js
alcheb: Yes! It makes me think:
 - in diffusion series `TemporalPositionIdentifier` would not be of any use (it is usually defined and equals 1)
- in non-diffusion series, `DiffusionBValue` is either `undefined` or defined AND equal to 0

There is another DICOM tag named `0020,0105 (NumberOfTemporalPositions)` that could prevent unnecessary sorting.

We could do something like: 
```
if (imageSet.modality === 'MR') {
  imageSet.sortBy((a, b) => {

    // Sort by TemporalPositionIdentifier
    if (instance._instance.NumberOfTemporalPositions !== 1) {      
      if (a._instance.TemporalPositionIdentifier > b._instance.TemporalPositionIdentifier) return 1;
      if (a._instance.TemporalPositionIdentifier < b._instance.TemporalPositionIdentifier) return -1;
    }

    // Sort by DiffusionBValue if defined (Diffusion series)
    if (typeof(instance._instance.DiffusionBValue) !== 'undefined') {
      if (a._instance.DiffusionBValue > b._instance.DiffusionBValue) return 1;
      if (a._instance.DiffusionBValue < b._instance.DiffusionBValue) return -1;
    }
  });
```

And in retrieveMetadata.js:
```
   // Get DiffusionBValue and TemporalPositionIdentifier for MR stack sorting
    DiffusionBValue: DICOMWeb.getNumber(instance['00189087']),
    TemporalPositionIdentifier: DICOMWeb.getNumber(instance['00200100']),
    NumberOfTemporalPositions: DICOMWeb.getNumber(instance['00200105']),
```

What do you think?

alcheb:
> While this is the easier implementation, as the end user, would you prefer that the stacks be split?

I don't have a personal preference, but:
  - in favor of splitting stacks:
    - it avoids unintentional skipping of a whole series (the end user would only scroll through the first phase and think this is it)
    - it avoids bad window levels especially in diffusion series as the images get darker with high `DiffusionBValues` (see example)
  - in favor of sorting :
    - it avoids over-crowding MR exams that already have too many series
    - I can do it :sweat_smile: 

Either way I'll be happy!
Cheers
Al

JoaoSantinha:
@alcheb very cool PR!

I would like to just add that most times the diffusion b-value tag are vendor specific in which only Philips fills the corresponding tag for the b-value.
Here are the tags normally used by Philips, GE and Siemens:
Siemens = (0018, 0024) or (0019; 100C)
GE = (0043, 1039)
Philips = (0018, 9087)

As for Philips, sometimes the files might be exported in enhanced DICOM leading to this tag to be
included within (5200, 9230) -> (0018, 9117)

Regarding the viewing of 4D series, Horos and OsiriX do something very interesting which is to load them as 4D series, where you will scroll through a 3D volume (e.g. b=0 or time = 0) and by using the up or down arrows you change to another b-value or time point 3D volume. 

Hope this helps.

pieper:
This is an important topic - thanks for contributing! 👍 

There's a great set of testing data on this page:

https://www.nitrc.org/plugins/mwiki/index.php/dcm2nii:MainPage#Diffusion_Tensor_Imaging

Would be good to know how this sorting methodology works with the various datasets.

alcheb:
> @alcheb very cool PR!

Thanks! :call_me_hand: 

> Regarding the viewing of 4D series, Horos and OsiriX do something very interesting which is to load them as 4D series, where you will scroll through a 3D volume (e.g. b=0 or time = 0) and by using the up or down arrows you change to another b-value or time point 3D volume.

Yes! Totally forgot about that. I've seen it as well in Telemis. There were numbers at the bottom indicating which stack you were viewing. Moving from stack to stack in the same series was done with left and right arrows or sliding left or right with the mouse. Thank you for putting a name on those things! :smile: 

With all that said I reworked a bit the PR (commit to follow). This proposal has better scalability and the sorting is done separately giving the opportunity to develop at some later time a splitting version or a "multi-stack-series-scrolling" scheme as in Horos/OsiriX/Telemis.

The code is I hope sufficiently commented for a quick understanding of what's going on. There is some monitoring-specific code that outputs to the console.

We will need to extensively test this against sample exams. I have some other MR images Siemens that I'll anonymize and upload when I find the time.

Thank you @JoaoSantinha @pieper !
Cheers,
Al

swederik:
> > Regarding the viewing of 4D series, Horos and OsiriX do something very interesting which is to load them as 4D series, where you will scroll through a 3D volume (e.g. b=0 or time = 0) and by using the up or down arrows you change to another b-value or time point 3D volume.
> 
> Yes! Totally forgot about that. I've seen it as well in Telemis. There were numbers at the bottom indicating which stack you were viewing. Moving from stack to stack in the same series was done with left and right arrows or sliding left or right with the mouse.

This is definitely something we want to add!

Some other things to think about:
- How do we prefetch for a stack of stacks in Cornerstone? Across time (or 4th dimension) first, then within the stack? We will need a new prefetch strategy for this.
- What do we do for the VTK extension? Build multiple volumes and swap them out for 4d? This could become very memory intensive. Maybe we should prevent reconstruction of 4d stacks for now for VTK?

alcheb:
Hi! I cleaned up some redundancies in the code. No need to have 2 indentical tag lists in two separate files.

Others things I did : 
  - limited the number of stacks within a series to 4 (arbitrarily). The AcquisitionTime tag was causing some problems on a prostate MR with 19 different values ; now, if there are 5 or more unique values in the series the tag tested will be ignored ;
  - added some information in the series description about tags used in case of sorting ;
  - tested this with the DTI DICOM files provided by @pieper : brain MR exams, no problems with sorting ; some series are very very large (more than 2000 images!) ; 
  - tested Siemens and GE-made liver MR with no problems (will upload them shortly).

> Some other things to think about:
> 
> * How do we prefetch for a stack of stacks in Cornerstone? Across time (or 4th dimension) first, then within the stack? We will need a new prefetch strategy for this.
> * What do we do for the VTK extension? Build multiple volumes and swap them out for 4d? This could become very memory intensive. Maybe we should prevent reconstruction of 4d stacks for now for VTK?

I wrote a quick override to consider all 4D series as non reconstructable for the time being. 

Sadly I think I am unable to go much further than that. Any tinkering with cornerstone, viewports and overlay elements in order to implement a true multi-stack series scrolling will have to be done by someone else! :sweat_smile:

Cheers
Al

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -585,6 +585,38 @@ const isMultiFrame = instance => {
   return instance.getRawValue('x00280008') > 1;
 };
 
+const is4DSeries = instances => {
+  if (instances[0].getRawValue('x00080060') !== 'MR') return 0; // only MR exams
```
swederik: Can you return false instead of 0?

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -585,6 +585,38 @@ const isMultiFrame = instance => {
...
+
+  // The goal here is to identify 4D series and the tags we'll use to sort them
+  // List tags used to identify and sort 4D series
+  const sortTagsList = Object.keys(instances[0]._instance.sort4DSeriesTags);
+
+  // Among those tags, identify those that have multiple values
+  var sortTags = Array(); // Declare an array to store tags to sort by
```
swederik: You can use let instead of var

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -585,6 +585,38 @@ const isMultiFrame = instance => {
...
+  var sortTagsVal = Array(); // for debug purposes
+
+  sortTagsList.map(testedTag => { // loop through the tag list
```
swederik: I think you can use forEach instead if you aren't using the result of the map

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -585,6 +585,38 @@ const isMultiFrame = instance => {
...
+      // And less than 5 (i.e. not too many)
+      sortTags.push(testedTag); // Push to our new array
+      sortTagsVal.push(tagsUniqueValues); // debug
+    }
+  });
+  // If none of them do : return 0
+  if (sortTags.length === 0) return 0;
+
+  // Debug: report on tags for each series
+  console.log('%c4D series! Sorting tags:', 'color: black; background: #9ccef9; padding: 4px; font-weight: bold;');
```
swederik: You can use log.debug here:

https://github.com/OHIF/Viewers/blob/b59c8c277af38ab601c2f3baa5a0cef8841c4438/platform/core/src/log.js

Example usage:

https://github.com/OHIF/Viewers/blob/b59c8c277af38ab601c2f3baa5a0cef8841c4438/platform/core/src/ui/handleError.js#L1

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -597,11 +629,12 @@ const makeDisplaySet = (series, instances) => {
     seriesTime: seriesData.seriesTime,
     seriesInstanceUid: series.getSeriesInstanceUID(),
     seriesNumber: instance.getRawValue('x00200011'),
-    seriesDescription: instance.getRawValue('x0008103e'),
+    seriesDescription: instance.getRawValue('x0008103e') || '',
```
swederik: Why do we need an empty string?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -355,6 +355,32 @@ async function makeSOPInstance(server, study, instance) {
     lossyImageCompressionMethod: DICOMWeb.getString(instance['00282114']),
     echoNumber: DICOMWeb.getString(instance['00180086']),
     contrastBolusAgent: DICOMWeb.getString(instance['00180010']),
+    sort4DSeriesTags: {
```
swederik: Can we take these out of this sub object here in the instance and then organize them into the sort4DSeriesTags object in the is4DSeries function which actually does the sorting?

dannyrb:
@alcheb are you able to action @swederik's suggested changes, or do you require assistance?

dannyrb:
@swederik would you like me to assign a Core Team developer to carry this across the finish line?

dannyrb:
@swederik would you like me to assign a Core Team developer to carry this across the finish line?

Last poke, or I'll let stale bot close this on it's next pass through :P

swederik:
Don't close it. We want / need this functionality it just keeps getting pushed down my list.

sedghi:

sedghi:

PR #1068, opened 2019-10-21
feat: responsive study list

dannyrb:
## Changes

- Remove version "pill" at Erik's request
- Decrease component nesting
- Decrease use of prop drilling
- Update many components to be "function" components
- Utilize `useMedia` and `useDebounce` hooks
- Page no longer has "horizontal scrollbar" at low widths
- Page no longer throws percy false positives for text placement
- Page no longer uses a default all-inclusive date range
- `react-dates` component updated to latest
- `react-dates` component uses single month, instead of 2, to better support smaller displays
- Support for searching multiple fields given a single input value
- Support for different table row templates at varying breakpoints

## Suggested Next steps

- Expose table meta to allow for custom filters
- Expose table templates to allow for custom table templates at different breakpoints
- Expose search hook
    - Takes raw values from filters
    - Makes a request for data
    - Must map data back to a display representation that can be consumed by table templates
- Logo text to two lines, so we can show it longer, or `OHIF` on very small displays
- Hamburger menu button that launches "popover" or "drawer" for more menu options
- Quick date filters on smallest table layout
- Expand or sidepanel advanced filter options for small and medium layouts
- Expand row to see a "study detail" or "preview"
    - Series images, reports, related studies, etc.
- Better support for including "actions" for each row
- Potentially use `select2` for enhanced study seeking in mobile views
- See: https://uxdesign.cc/design-better-data-tables-4ecc99d23356

### Mobile

image

### Tablet

image

### Desktop

image


platform/ui/src/components/studyList/TableSearchFilter.js
```diff
@@ -23,7 +23,9 @@ function TableSearchFilter(props) {
   const today = moment();
   const lastWeek = moment().subtract(7, 'day');
   const lastMonth = moment().subtract(1, 'month');
-  const defaultStartDate = moment().subtract(365 * 10, 'days'); // this.props.studyListDateFilterNumDays,
+  // TODO: Does this need to be preserved? Default sort could be nice for large data sources
```
swederik: Yes we need some sort of default. Large PACS will need this set to e.g. 1 day

platform/ui/src/components/studyList/StudyList.styl
```diff
@@ -76,14 +100,106 @@ placeholder-color(c)
...
+  &.emptyCell
+    color: var(--ui-gray-light)
+
+.study-list-container > table.table > thead > tr > th.studyDate
+  min-width: 230px
+
+.study-list-container
+  .filters
+    label
+      display: flex;
```
rodrigolabs: image


The sorting arrow is a little bit misaligned, you can fix that by adding `align-items: center;`:

```suggestion
      display: flex;
      align-items: center;
```

rodrigolabs:
We had a line-break on the calendar buttons, maybe we can decrease `padding/margin` to keep them in the same line, what do you think?

platform/ui/src/components/studyList/CustomDateRangePicker.js
```diff
@@ -2,13 +2,12 @@
 //  https://github.com/airbnb/react-dates#initialize
 import 'react-dates/initialize';
 import 'react-dates/lib/css/_datepicker.css';
+import './CustomDateRangePicker.css';
```
rodrigolabs: quick question: In some places, I see `.css` being used and in some other `.styl` as well as the use of CSS standards like BEM or just hyphenated lowercase words. Do we have any standards for this or is it something we can think about in the future? Just a curiosity to see how we can move forward in the future about this :)

platform/ui/src/components/studyList/TableSearchFilter.js
```diff
@@ -0,0 +1,117 @@
...
+  const today = moment();
+  const lastWeek = moment().subtract(7, 'day');
+  const lastMonth = moment().subtract(1, 'month');
+  // TODO: Does this need to be preserved? Default sort could be nice for large data sources
+  // this.props.studyListDateFilterNumDays,
+  const defaultStartDate = moment().subtract(365 * 10, 'days');
+  const defaultEndDate = today;
+  const studyDatePresets = [
+    {
+      text: 'Today',
```
rodrigolabs: This label can be translated?

platform/ui/src/components/studyList/TableSearchFilter.js
```diff
@@ -0,0 +1,117 @@
...
+      start: today,
+      end: today,
+    },
+    {
+      text: 'Last 7 days',
```
rodrigolabs: same as above `i18n`

platform/ui/src/components/studyList/TableSearchFilter.js
```diff
@@ -0,0 +1,117 @@
...
+      start: lastWeek,
+      end: today,
+    },
+    {
+      text: 'Last 30 days',
```
rodrigolabs: same as above `i18n`

platform/viewer/src/components/Header/Header.css
```diff
@@ -97,26 +99,45 @@
   margin-right: 1rem;
 }
 
-.header-versionInfo {
```
rodrigolabs: Why is this being removed? Are we gonna keep the version info only in the about modal?

rodrigolabs:
The plus button for `addNewStudy` is logging a message in the console:



image


Was that expected?

platform/ui/src/components/studyList/StudyList.js
```diff
@@ -1,428 +1,400 @@
...
+};
+
+TableRow.defaultProps = {
+  isHighlighted: false,
+};
+
 const connectedComponent = withTranslation('StudyList')(StudyList);
 export { connectedComponent as StudyList };
+
+// function getNoListFragment(translate, studies, error, loading) {
```
rodrigolabs: Will this function be used? If not, maybe we could remove these lines.

rodrigolabs:
@dannyrb I made a few simple questions and I think we would need only align the sort indicators, some i18n translations, and the calendar buttons :)
Great job on making study list responsive!

platform/ui/src/components/studyList/StudyList.styl
dannyrb: Ooh. Using GitHub `suggestions` feature.
Thanks for the tip 🙏

platform/ui/src/components/studyList/CustomDateRangePicker.js
dannyrb: This is a great point! This is a conversation I want to have with the team at some point. Currently, everyone more or less employs their particular preference, or tries to adhere as closely to what they believe the current file's style is using.

Long-term, I think we'll either land on something like `tailwindcss`
See: https://tailwindcss.com/

Or CSS Modules, as we can scope classes to the component, and whatever naming convention the team prefers; but enforced with something like `stylelint`

platform/viewer/src/components/Header/Header.css
dannyrb: This is a request from @swederik.
Yes, we'll still have the version in the About Modal, but it will no longer be visible in the header.

platform/ui/src/components/studyList/TableSearchFilter.js
dannyrb: I've updated to use translation; although, I think this is the first functional component that does so. It's also my first time messing with them outside of some light research. Can you double check my implementation?

dannyrb:
@rodrigolabs and @swederik; doubled back to clean-up based on feedback. Let me know if I missed anything.

PR #1071, opened 2019-10-22
fix: E2E Intermittent Failures

rodrigolabs:
As discussed with @dannyrb, this PR is a quick fix in E2E tests that are causing intermittent failures, basically, by changing the mouse click from `top/center` to `center/top`.


dannyrb:
LGTM :+1:

Issue #1072, opened 2019-10-22
Single-click to load thumbnail only works after performing drag-n-drop


mirnasilva:
Related to #895. Issue still happening.

### Description

User is not able to load a new series by single-clicking on a thumbnail. This is only possible after performing an initial drag-n-drop.
This issue can also be reproduced with multiple layouts.

![bug-singleclick-thumbnail](https://user-images.githubusercontent.com/13245529/67342168-f4136a00-f507-11e9-9654-a38a0d683271.gif)

### Steps to reproduce

1. Load a study
2. Single-click on any thumbnail
3. Verify that the series was not loaded
4. Drag-n-drop any thumbnail into the viewport
5. Single-click again on any thumbnail
6. Verify that the series was loaded with success

### Acceptance Criteria

User should be able to load a new series by single-clicking on the thumbnail. This should be valid for single viewport and multiple viewports.

dannyrb:
This is resolved; We've created separate smaller issues to address the other incorrect behaviors around this.

PR #1073, opened 2019-10-23
E2E tests: Stack Scroll, Right-click to zoom, Middle-click to Pan

mirnasilva:
Tests implemented:
- Stack scrolling
- Right-click to zoom
- Middle-click to pan image
- Single-click to load thumbnail (bug found #1072)
- About Menu overlay

Added some `data-cy` selectors for some elements on About overlay.


platform/viewer/cypress/integration/common/OHIFStudyViewer.spec.js
```diff
@@ -206,4 +206,85 @@ describe('OHIF Study Viewer Page', () => {
...
+
+    //pan image with middle click
+    cy.get('@viewport')
+      .trigger('mousedown', 'center', { which: 2 })
+      .trigger('mousemove', 'bottom', { which: 2 })
+      .trigger('mouseup', 'bottom')
+      .then(() => {
+        expect(currentPan().y > 0).to.eq(true);
+      });
+  });
```
dannyrb: We can probably keep this for now. This might be a better candidate for percy visual testing though. Maybe add a comment at the top of the test to indicate as much.

platform/viewer/cypress/integration/common/OHIFStudyViewer.spec.js
```diff
@@ -1,16 +1,18 @@
-describe('OHIF Study Viewer Page', () => {
-  before(() => {
+import { Youtube } from 'react-feather';
```
dannyrb: Accidental import? Can we remove this?

mirnasilva: That's crazy... Removing right now!

platform/ui/src/components/userPreferencesModal/AboutModal.js
```diff
@@ -63,8 +75,8 @@ class AboutModal extends Component {
   renderTableRow(item) {
     return (
       <tr key={item.name}>
-        <td>{item.name}</td>
-        <td>
+        <td data-cy={item.testLabel}>{item.name}</td>
+        <td data-cy={item.testId}>
```
dannyrb: For most of these tests, we only super care that the body of the modal contains some expected text. We can create a single `data-cy="about-modal-body"` on the Modal's body, and then assert that it contains our expected text. This has a few benefits:

- We don't need to update our row render logic to include extra values that are only meant for testing
- The entire modal body layout can change, but the tests will still pass
    - IE. we don't care about the layout, we only care about the content

platform/ui/src/components/userPreferencesModal/AboutModal.js
```diff
@@ -112,6 +126,7 @@ class AboutModal extends Component {
             {` `}
             <a
               className="btn btn-default"
+              data-cy="btn-more-details"
```
dannyrb: Similarly, for these buttons, we can probably look for them by scoping our query to the modal (`data-cy="about-modal"`), and then just verifying that the text content of the button exists anywhere in the modal (contains text).

This allows us to scope where we're looking, but allows for things to move around, without breaking our tests.

Issue #1075, opened 2019-10-23
Returning to the Study List before all series have finished loading throws multiple errors in console


emfol:
### Reproduction Steps

1. Make sure lazy load of series is enabled
2. Load a study with multiple series;
3. Before all series finish loading, click the `Study List` button to return to the study list;
4. Check the errors in the console;

![console-errors](https://user-images.githubusercontent.com/1145367/67397634-db469b00-f57f-11e9-9cfe-ce664ebddf7f.png)

dannyrb:
PR: https://github.com/OHIF/Viewers/pull/1090

PR #1090, opened 2019-10-25
fix: #1075 Returning to the Study List before all series have finishe…

emfol:
These changes fix the issues for  ImageThumnail and ViewerRetrieveStudyData
There is still one exception being thrown but that one is in react-cornerstone-viewport project [WIP]

### DIFF
--- a/platform/ui/src/components/studyBrowser/ImageThumbnail.js
+++ b/platform/ui/src/components/studyBrowser/ImageThumbnail.js
@@ -32,7 +32,7 @@ export default class ImageThumbnail extends PureComponent {
 
   constructor(props) {
     super(props);
-
+    this.abortAsyncOperations = false;
     this.canvas = React.createRef();
     this.state = {
       loading: this.shouldRenderToCanvas(),
@@ -48,19 +48,27 @@ export default class ImageThumbnail extends PureComponent {
       cornerstone
         .loadAndCacheImage(this.props.imageId)
         .then(image => {
-          cornerstone.renderToCanvas(this.canvas.current, image);
-          this.setState({
-            loading: false,
-          });
+          if (!this.abortAsyncOperations) {
+            cornerstone.renderToCanvas(this.canvas.current, image);
+            this.setState({
+              loading: false,
+            });
+          }
         })
         .catch(error => {
-          this.setState({
-            loading: false,
-            error: true,
-          });
-          throw new Error(error);
+          if (!this.abortAsyncOperations) {
+            this.setState({
+              loading: false,
+              error: true,
+            });
+            throw new Error(error);
+          }
         });
     }
+  }
+
+  componentWillUnmount() {
+    this.abortAsyncOperations = true;
   }
 
   render() {
--- a/platform/ui/src/components/studyBrowser/ImageThumbnail.js
+++ b/platform/ui/src/components/studyBrowser/ImageThumbnail.js
@@ -1,6 +1,7 @@
 /* global cornerstone */
 import './ImageThumbnail.styl';
 
+import { utils } from '@ohif/core';
 import React, { PureComponent } from 'react';
 
 import PropTypes from 'prop-types';
@@ -32,7 +33,7 @@ export default class ImageThumbnail extends PureComponent {
 
   constructor(props) {
     super(props);
-    this.abortAsyncOperations = false;
+    this.cancelablePromises = [];
     this.canvas = React.createRef();
     this.state = {
       loading: this.shouldRenderToCanvas(),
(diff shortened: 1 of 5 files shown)


platform/ui/src/components/studyBrowser/ImageThumbnail.js
```diff
@@ -48,21 +48,29 @@ export default class ImageThumbnail extends PureComponent {
...
-            error: true,
-          });
-          throw new Error(error);
+          if (!this.abortAsyncOperations) {
+            this.setState({
+              loading: false,
+              error: true,
+            });
+            throw new Error(error);
+          }
```
dannyrb: Ahh. So these requests have already been initiated; we just need to make sure we're not trying to render to a removed canvas if we've unmounted. We can probably clean this up a bit w/ `async/await`, and just returning early if we have our result.

```js
const { imageId } = this.props;
try {
  const image = await cornerstone.loadAndCacheImage(imageId);

  if (!this.canvas.current) return;
  cornerstone.renderToCanvas(this.canvas.current, image);
  this.setState({ loading: false });
} catch(error) {
  this.setState({ loading: false, error: true });
  throw new Error(error);
}
```

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -83,18 +85,56 @@ class ViewerRetrieveStudyData extends Component {
...
+    for (const studyInstanceUid in this.seriesLoadStats) {
+      const seriesLoadStats = this.seriesLoadStats[studyInstanceUid];
+      if (
+        seriesLoadStats &&
+        (seriesLoadStats.count > 0 || seriesLoadStats.errors > 0)
+      ) {
+        deleteStudyMetadataPromise(studyInstanceUid);
+        studyMetadataManager.remove(studyInstanceUid);
+        log.info(`Purging incomplete study data: ${studyInstanceUid}`);
+      }
```
dannyrb: Similar thing here, right? We need to:

- Safely handle in-flight promises from trying to render to now removed DOM
- Delete all queued promises that are now no longer relevant

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -83,18 +85,56 @@ class ViewerRetrieveStudyData extends Component {
     });
   }
 
+  _handleSeriesLoadResult(error, studyMetadata, series) {
```
dannyrb: This signature is weird. Shouldn't it be `studyMetadata`, `series`, `error` as error is optional?

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -83,18 +85,56 @@ class ViewerRetrieveStudyData extends Component {
...
     const { seriesLoader } = studyMetadata.getData();
     if (!seriesLoader) {
       return;
     }
+    const seriesLoadStats = (this.seriesLoadStats[
+      studyMetadata.getStudyInstanceUID()
+    ] = {
+      errors: 0,
+      count: 0,
+    });
```
dannyrb: Track each success/fail for each study's series metadata load?

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -83,18 +85,56 @@ class ViewerRetrieveStudyData extends Component {
     });
   }
 
+  _handleSeriesLoadResult(error, studyMetadata, series) {
+    if (this.abortSeriesLoad) return;
+    const seriesLoadStats = this.seriesLoadStats[
```
dannyrb: I'm wary of reusing the same name for `local` variables and object properties. It's akin to "variable shadowing", and can lead to confusing code. Can we pick out better names here?

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -18,6 +18,8 @@ class ViewerRetrieveStudyData extends Component {
 
   constructor(props) {
     super(props);
+    this.abortSeriesLoad = false;
+    this.seriesLoadStats = Object.create(null);
```
dannyrb: An object, where each property is a study Uid?

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -83,18 +85,56 @@ class ViewerRetrieveStudyData extends Component {
     });
   }
 
+  _handleSeriesLoadResult(error, studyMetadata, series) {
+    if (this.abortSeriesLoad) return;
```
dannyrb: Are we aborting here because the studyMetadataManager may be dead? If not, why are we?

dannyrb:
Also, can you provide more info around this?

> There is still one exception being thrown but that one is in react-cornerstone-viewport project [WIP]

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: I tried to follow the old JS standard for callbacks hehehe...
If there is an error, none of the other parameters should be used.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: yes!

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: no, because the view might have been unmounted....

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: yup! good point.

emfol:
> Also, can you provide more info around this?
> 
> > There is still one exception being thrown but that one is in react-cornerstone-viewport project [WIP]

Yes. The same kind of exception is being thrown from the viewport if we get back to the study list before the image finishes loading and gets rendered into the viewport. I'm likely to come up with a similar fix for the viewport as well.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: yes!

platform/ui/src/components/studyBrowser/ImageThumbnail.js
emfol: Not only this, we must also make sure we do not call `setState` when the component has been unmounted. I found an article about preventing this kind of error and I changed the code to use the _cancelable_ pattern described there:
https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
emfol: fixed!

dannyrb:
image
https://app.circleci.com/jobs/github/OHIF/Viewers/6344

PR #1091, opened 2019-10-25
fix: set SR in ActiveViewport by clicking thumb

### LINKED ISSUE #1076: Unable to set SR in `ActiveViewport` by clicking thumbnail
You can click a thumbnail to set it as the dataSet for the activeViewport; however, we get odd results if we **click** an SR thumbnail.

CYUL2Q7noe

rodrigolabs:
Fixes #1076 

---

## Preview 

ezgif com-optimize

### DIFF
--- a/platform/core/src/redux/reducers/viewports.js
+++ b/platform/core/src/redux/reducers/viewports.js
@@ -34,7 +34,6 @@ const defaultState = {
  * @param {Object} [action.viewportSpecificData]
  */
 const viewports = (state = defaultState, action) => {
-  let currentData;
   let viewportSpecificData;
   let useActiveViewport = false;
   switch (action.type) {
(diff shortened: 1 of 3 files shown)


platform/core/src/redux/reducers/viewports.js
```diff
@@ -85,14 +84,24 @@ const viewports = (state = defaultState, action) => {
...
         ? state.activeViewportIndex
         : action.viewportIndex;
-      currentData = cloneDeep(state.viewportSpecificData[viewportIndex]) || {};
+      const { dom } = state.viewportSpecificData[viewportIndex];
+
       viewportSpecificData = cloneDeep(state.viewportSpecificData);
-      viewportSpecificData[viewportIndex] = merge({}, currentData, action.data);
+      viewportSpecificData[viewportIndex] = {
+        dom,
+        ...action.data,
```
dannyrb: I know that `action.data` is everything, but what is `action.data`?
Can we be more explicit here? Or are we SOL with the current setup?

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -295,7 +295,10 @@ class Viewer extends Component {
                 activeIndex={this.props.activeViewportIndex}
               />
             ) : (
-              <ConnectedStudyBrowser studies={this.state.thumbnails} />
+              <ConnectedStudyBrowser
+                studies={this.state.thumbnails}
+                studyMetadata={this.props.studies}
```
dannyrb: Man, we really need to knock out a better way to do this

platform/core/src/redux/reducers/viewports.js
rodrigolabs: This is what we're calling `viewportSpecificData`, e.g.:
```javascript
{
  displaySetInstanceUid,
  seriesDate,
  seriesTime,
  seriesInstanceUid,
  seriesNumber,
  plugin,
  seriesDescription,
  modality,
  [etc...]
}
```

dannyrb: Right, but we need to start destructuring "data" so it's clearer at a glance what's going into reactive state, and to make sure nothing new sneaks in unless we deliberately add it.

platform/viewer/src/connectedComponents/Viewer.js
rodrigolabs: Totally agree! In the next steps, we can probably add the study list in the redux store to avoid that.

Issue #1089, opened 2019-10-25
JSON launch not working properly after QIDO/WADO switch


swederik:
@spalte reported this. I thought disabling study lazy load would fix it, but he says it doesn't.

Here's an example:

https://viewer.ohif.org/viewer?url=https://s3.eu-central-1.amazonaws.com/ohif-viewer/JSON/PTCTStudy.json

PR #1093, opened 2019-10-25
fix: 🐛 JSON launch not working properly (#1089)

emfol:
JSON launch not working properly after QIDO/WADO switch #1089

Closes: #1089

TEST URL: https://deploy-preview-1093--ohif.netlify.com/pwa/viewer?url=https://s3.eu-central-1.amazonaws.com/ohif-viewer/JSON/PTCTStudy.json


emfol:
@swederik I don't know how this got broken but as you can see from the changes, it has nothing to do with QIDO+WADO feature...

dannyrb:
`?url=https://s3.eu-central-1.amazonaws.com/ohif-viewer/JSON/PTCTStudy.json` -->

platform/viewer/src/connectedComponents/ViewerMain.js
```diff
@@ -60,11 +60,12 @@ class ViewerMain extends Component {
...
 
-      this.setState({
-        displaySets,
-      });
-
-      this.fillEmptyViewportPanes();
+      this.setState(
+        {
+          displaySets,
+        },
```
dannyrb: Do we have different lint settings somehow? I don't get this same change on save.

emfol: I didn't change my linting settings... no matter what I do, after saving I only get this... :-/
I'm curious how it looks on your end...

emfol:
When I first tried to load the JSON URL on my local environment, the only exception I got was coming from the `<Viewer>` component and was related to the fact that the `viewports` property (which is required) was absent:

https://github.com/OHIF/Viewers/blob/bef50f929853ee70c519ece87d5914b8fd96414f/platform/viewer/src/connectedComponents/Viewer.js#L64

After changing the render method to use `<ConnectedViewer>` instead of `<Viewer>` the exception was gone but still the viewport wouldn't render.

Comparing how `<ConnectedViewer>` was invoked inside `<ViewerRetrieveStudyData>` I noticed the `studyInstanceUids` property was absent so I added that to make sure it was not interfering with anything.

While debugging downstream (inside `<ViewerMain>`) I realized that the viewport was not being initialized because the `fillEmptyViewportPanes` method was being called before the component state was properly updated (from `componentDidMount` method);

So, the fix was:
1. Replace `<Viewer>` by `<ConnectedViewer>`;
2. Make sure `fillEmptyViewportPanes` method was only called after the component state was properly updated;

As I said, I don't know how or when this feature got broken but I couldn't find any relationship between this issue and the QIDO+WADO changes...

platform/viewer/src/connectedComponents/ViewerMain.js
emfol: Actually, if I keep everything in a single line, the linter does not frown on me... :-)
I'll be updating the review soon...

dannyrb:
@emfol, I appreciate the explanation! It helps me wrap my head around what's going on much better.

PR #1106, opened 2019-10-26
Revert "feat: Snapshot Download Tool"

dannyrb:
Reverts OHIF/Viewers#840


dannyrb:
Using revert to confirm that the `Download` tool changes broke the MIP functionality:

image

Issue #1086, opened 2019-10-25
Add ModalProvider/Context


dannyrb:
Recently, we added a "notification service" in the form of a new Snackbar Component, container, and context. You can see the PR for it here:

https://github.com/OHIF/Viewers/pull/1011

We would like to implement something similar, but for modals. There should only ever be a single modal open in our UI; and having the ability to slot in different headers, body content, actions, etc. allows for a very reusable component, and for us to easily trigger a modal from any other component.

Ideally, this will be implemented in `@ohif/ui` (`<root>/platform/ui/`), and export it so it can be consumed by `@ohif/viewer` (`<root>/platform/viewer/`) and you can wrap any external dependency that may provide the bulk of the functionality that you're looking for.

At a minimum it should provide the same functionality that our current Modal does. We're using it in very few places:

- Options --> About
- User Preferences (commented out)
- Download / Snapshot Tool (in-progress)

After this is implemented, we'll need to update it's usage. I see this as taking place over three additional issues that can be completed separately:

- Update `About OHIF` modal
- Update `User Preferences` modal (and re-add)
- Expose as a service to extensions (see me)

PR #1110, opened 2019-10-28
feat: 🎸 New modal provider

igoroctaviano:
A new modal provider/context component to unify modal related code and
simplify modal use

Closes: #1086


platform/ui/src/index.js
```diff
@@ -52,6 +52,11 @@ import SnackbarProvider, {
   useSnackbarContext,
   withSnackbar,
 } from './utils/SnackbarProvider';
+import ModalProvider, {
+  useModalContext,
```
dannyrb: I realize this matches `useSnackbarContext`, but would it make more sense if this was simple `useModal`? Or am I misconstruing what this is?

platform/viewer/src/App.js
```diff
@@ -29,6 +29,7 @@ import { getActiveContexts } from './store/layout/selectors.js';
 import i18n from '@ohif/i18n';
 import store from './store';
 import { SnackbarProvider } from '@ohif/ui';
+import { ModalProvider } from '@ohif/ui';
```
dannyrb:
```
import { SnackbarProvider } from '@ohif/ui';
import { ModalProvider } from '@ohif/ui';
```

to

```
import { ModalProvider, SnackbarProvider } from '@ohif/ui';
```

platform/ui/src/utils/ModalProvider.js
```diff
@@ -0,0 +1,91 @@
...
+const ModalProvider = ({ children }) => {
+  const DEFAULT_OPTIONS = {
+    component: null,
+    backdrop: false,
+    keyboard: false,
+    show: true,
+    large: true,
+    closeButton: true,
+    title: null,
+    customClassName: null,
```
dannyrb: Can we get inline comments for these to better indicate what they do?

platform/ui/src/utils/ModalProvider.js
```diff
@@ -0,0 +1,91 @@
...
+  const [options, setOptions] = useState(DEFAULT_OPTIONS);
+
+  const show = (component, props = {}) =>
+    setOptions({
+      ...options,
+      component,
+      ...props,
+    });
+
+  const hide = () => setOptions(DEFAULT_OPTIONS);
```
dannyrb: This is neat. So `hide` also "clears" the component.

platform/ui/src/utils/ModalProvider.js
```diff
@@ -0,0 +1,91 @@
...
+    customClassName: null,
+  };
+
+  const [options, setOptions] = useState(DEFAULT_OPTIONS);
+
+  const show = (component, props = {}) =>
+    setOptions({
+      ...options,
+      component,
+      ...props,
```
dannyrb: Can we JSDOC `show`?

Show a component, and configure the modal's options. Got it.
How does `...` spread work if `options` has conflicting keys with `props`? Do the ones specified by `props` take priority?

platform/ui/src/utils/ModalProvider.js
```diff
@@ -0,0 +1,91 @@
...
+    });
+
+  const hide = () => setOptions(DEFAULT_OPTIONS);
+
+  return (
+    <Provider value={{ ...options, show, hide }}>
```
dannyrb: Still a little new to React. Is it possible to use `useContext` here for the ModalContext instead of a provider? If both are possible, which one is preferable? Why?

platform/ui/src/utils/ModalProvider.js
```diff
@@ -0,0 +1,91 @@
...
+                <Modal.Footer>
+                  {' '}
+                  <Footer hide={hide} />
+                </Modal.Footer>
+              )}
+            </Modal>
+          ) : null;
+        }}
+      </Consumer>
+      {children}
```
dannyrb: So `children` would render outside of the modal?

igoroctaviano: Yes, this children is the rest of the app that goes inside the provider itself.

dannyrb: Ahh! That makes a lot of sense.
Instead of baking the Modal into the provider, can we keep it as a separate `@ohif/ui` modal that we place beside the `<OHIFStandaloneViewer>` in `@ohif/viewer`?

https://github.com/OHIF/Viewers/pull/1110/files#diff-d14361ccec1e91c0c2716e30087fe0f2R106

This allows for anyone reusing our `@ohif/ui` but with a custom viewer project to swap out the implementation with anything that's API compatible.

platform/ui/src/utils/ModalProvider.js
igoroctaviano: So, let me try to explain... The provider passes the show function to all children components, the children then call the show method with a modal component as argument, but the Modal itself (inside the provider) also needs the Modal context which have a component state that the user just set (the new component), so the modal consumes the same context to show the content. This is why theres a ModalContext being used inside the provider itself. The provider is required in order to create the context, and the context can be accessed through props or through the useContext.

platform/ui/src/utils/ModalProvider.js
```diff
@@ -0,0 +1,91 @@
...
+            >
+              {(Header || options.title) && (
+                <Modal.Header closeButton={options.closeButton}>
+                  {options.title && <Modal.Title>{options.title}</Modal.Title>}
+                  {Header && <Header hide={hide} />}
+                </Modal.Header>
+              )}
+              <Modal.Body>
+                {Body && <Body hide={hide} />}
+                {Component && <Component hide={hide} />}
```
dannyrb: Do we need `Body` and `Component` here?

platform/ui/src/index.js
igoroctaviano: It could be renamed for sure, this is just a hook, a component calling a useContext will always rerender when the context value changes

platform/ui/src/utils/ModalProvider.js
igoroctaviano: Yeah.. the idea was to have the body prop independent of the first argument of the show function but i think it's better to just remove it and have the first arg (which represents the body) as required.

platform/ui/src/utils/ModalProvider.js
igoroctaviano: The props will override the others, so the user props take priority.

platform/ui/src/utils/ModalProvider.js
igoroctaviano: Done! thx.

platform/ui/src/utils/ModalProvider.js
igoroctaviano: I'm passing it as a prop to the provider now. Thx!

platform/ui/src/utils/ModalProvider.js
igoroctaviano: Sure! done, thx.

platform/ui/src/utils/ModalProvider.js
igoroctaviano: Sure! done! thx.

platform/viewer/src/App.js
igoroctaviano: Done! thx.

platform/ui/src/utils/ModalProvider.js
```diff
@@ -25,11 +25,7 @@ const ModalProvider = ({ children, modal: Modal }) => {
    * @returns void
    */
   const show = (component, props = {}) =>
-    setOptions({
-      ...options,
-      component,
-      ...props,
-    });
+    setOptions(Object.assign({}, options, props, { component }));
```
dannyrb: Thank you. I think this is much clearer 🙏

platform/ui/src/components/modal/Modal.js
```diff
@@ -0,0 +1,42 @@
...
+    {Footer && (
+      <Modal.Footer>
+        {' '}
+        <Footer hide={onHide} />
+      </Modal.Footer>
+    )}
+  </Modal>
+);
+
+export default CustomModal;
```
dannyrb: Ah. I see you used `CustomModal` to avoid name conflicts with `react-bootstrap-modal`. Can we alias one or the other?

Either `react-bootstrap-modal` as `ReactModal`, or our modal as `OhifModal`, or both renamed (to provide clarity)

platform/ui/src/components/modal/Modal.js
```diff
@@ -0,0 +1,42 @@
...
+  closeButton,
+  backdrop,
+  keyboard,
+  show,
+  large,
+  title,
+  onHide,
+  footer: Footer,
+  header: Header,
+  children: Component,
```
dannyrb: Do we think we'll need/use the `large` prop? Are all of these worth passing through? Can we add `propTypes`?

platform/ui/src/components/modal/Modal.js
```diff
@@ -0,0 +1,42 @@
...
+}) => (
+  <Modal
+    className={`modal fade themed in ${className}`}
```
dannyrb: can we use our `classnames` dependency here to clean this up instead of using string interpolation?

platform/viewer/src/App.js
```diff
@@ -118,10 +119,12 @@ class App extends Component {
...
           <I18nextProvider i18n={i18n}>
-            <Router basename={this.props.routerBasename}>
-              <WhiteLabellingContext.Provider value={this.props.whiteLabelling}>
+            <Router basename={routerBasename}>
+              <WhiteLabellingContext.Provider value={whiteLabelling}>
                 <SnackbarProvider>
-                  <OHIFStandaloneViewer />
+                  <ModalProvider modal={Modal}>
+                    <OHIFStandaloneViewer />
+                  </ModalProvider>
```
dannyrb: I originally meant something like this:

```jsx
<ModalProvider>
    <Modal />
    <OHIFStandaloneViewer />
</ModalProvider>
```

But I think your implementation hammers home that there should be a single `Modal` component. Let's keep what you have 👍

platform/ui/src/utils/ModalProvider.js
```diff
@@ -0,0 +1,78 @@
+import React, { useState, createContext, useContext } from 'react';
+
+const ModalContext = createContext(null);
+const { Provider, Consumer } = ModalContext;
+
+export const useModal = () => useContext(ModalContext);
+
+const ModalProvider = ({ children, modal: Modal }) => {
```
dannyrb: Do `propTypes` work here?

igoroctaviano: Done! thx.

platform/viewer/src/App.js
igoroctaviano: Done! thx.

platform/ui/src/components/modal/Modal.js
igoroctaviano: Done! thx.

platform/ui/src/components/modal/Modal.js
igoroctaviano: Done! thx.

platform/ui/src/components/modal/Modal.js
igoroctaviano: Done! thx.

Issue #1099, opened 2019-10-26
Snapshot download tool downloads empty images if width/height increased above large threshold


dannyrb:
To reproduce:

- Select an image
- Draw one line annotation
- Increase width above `500000`
- Attempt to download image
- Open downloaded image file (empty/blank)

Fix:

- Enforce a reasonable max value (10,000?)

PR #1112, opened 2019-10-28
fix: 🐛 Limit image download size to avoid browser issues

igoroctaviano:
This fix adjusts the max texture size to allow browser compatibility
(blob size limit)

Closes: #1099


platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -75,24 +76,31 @@ const DownloadDialog = ({
   }, [defaultSize, disableViewport, enableViewport, viewportElement]);
 
   useEffect(() => {
+    const getValidMeasure = value => {
+      const min = minimumSize - 1;
```
dannyrb: Could just use `value >= minimumSize`?

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -75,24 +76,31 @@ const DownloadDialog = ({
...
+      return value > min && value < maximumSize
+        ? value
+        : value < min
+        ? minimumSize
+        : maximumSize;
```
dannyrb: Let's avoid a nested ternary if we can help it

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -212,18 +224,22 @@ const DownloadDialog = ({
...
             </div>
             <div className="height">
               <TextInput
                 type="number"
                 min={minimumSize}
+                max={maximumSize}
                 value={height}
                 label={t('Image height (px)')}
                 onChange={onHeightChange}
+                onBlur={onHeightChange}
```
dannyrb: Does `onBlur` catch anything not already caught by `onChange`?

platform/viewer/src/connectedComponents/ConnectedDownloadDialog.js
```diff
@@ -108,7 +109,15 @@ const mapStateToProps = (state, ownProps) => {
         return window.navigator.msSaveBlob(blob, file);
       }
 
-      return cornerstoneTools.SaveAs(viewportElement, file, mimetype);
+      viewportElement.querySelector('canvas').toBlob(blob => {
+        const URLObj = window.URL || window.webkitURL;
+        let a = document.createElement('a');
```
dannyrb: `a` is not reassigned. This can probably be `const`

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -75,24 +76,31 @@ const DownloadDialog = ({
   }, [defaultSize, disableViewport, enableViewport, viewportElement]);
 
   useEffect(() => {
+    const getValidMeasure = value => {
```
dannyrb: I don't follow why this needs to be wrapped in `useEffect`. Any of the `set` calls will trigger a re-render. You can have this be a normal func declared beneath `DownloadDialog`'s function.

platform/ui/src/components/downloadDialog/DownloadDialog.js
```diff
@@ -125,10 +133,13 @@ const DownloadDialog = ({
...
     fileType,
     downloadCanvas.ref,
+    minimumSize,
+    maximumSize,
   ]);
 
-  const onHeightChange = () => {
-    const newHeight = event.target.value;
+  const onHeightChange = event => {
+    const newHeight =
```
swederik: You can also use Math.min(event.target.value, maximumSize) if you want to easily clamp something.

platform/ui/src/components/downloadDialog/DownloadDialog.js
igoroctaviano: This is because the maximumSize and minimumSize properties are being used to process the valid measure, the useEffect have those being watched. I have simplified this function though since im clamping the value in the onChange for max values. 

Done!, thx.

platform/viewer/src/connectedComponents/ConnectedDownloadDialog.js
igoroctaviano: Done! thx.

platform/ui/src/components/downloadDialog/DownloadDialog.js
igoroctaviano: It wasn't, I forgot to take this out. Thank you.

platform/ui/src/components/downloadDialog/DownloadDialog.js
igoroctaviano: Sure! thx, I have simplified this function because of the clamp for max values in the onChange event.

platform/ui/src/components/downloadDialog/DownloadDialog.js
igoroctaviano: Done. Thx.

platform/ui/src/components/downloadDialog/DownloadDialog.js
igoroctaviano: That's a lot better, thx.

PR #1117, opened 2019-10-28
feat: 🎸 Filter by url query param for seriesInstnaceUID

ladeirarodolfo:
### Includes
- Filter series by query param (works for default and gcloud). IT FILTERS ONLY SERIES (no sopInstance yet)
- Filter applies to lazy or not study loading
- Uses Snackbar to toast user messages.

#### Related
- There will be another version (PR to be opened soon), which will promote instead of filter series.

### Screnshoots
undefined-_2_
Untitled_-Oct-28_-2019-6_25-PM


dannyrb:
@swederik is there a simple way for me to test this? If not currently, is there a simple/sane approach to make it possible?

platform/core/src/studies/retrieveStudiesMetadata.js
```diff
@@ -13,19 +13,15 @@ import { retrieveStudyMetadata } from './retrieveStudyMetadata';
 export default function retrieveStudiesMetadata(
   server,
   studyInstanceUids,
-  seriesInstanceUids
+  filters
```
dannyrb: Can we destructure this at some point, so it's clearer what possible filters can be?

platform/core/src/studies/retrieveStudyMetadata.js
```diff
@@ -11,7 +11,7 @@ const StudyMetaDataPromises = new Map();
  * @param {string} studyInstanceUid The UID of the Study to be retrieved
  * @returns {Promise} that will be resolved with the metadata or rejected with the error
  */
-export function retrieveStudyMetadata(server, studyInstanceUid) {
+export function retrieveStudyMetadata(server, studyInstanceUid, filters) {
```
dannyrb: Can we update the JSDoc and destructure here so it's clearer which filters can be passed?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -468,37 +468,70 @@ async function createStudyFromSOPInstanceList(server, sopInstanceList) {
  * @param {string} studyInstanceUid The Study Instance UID of the study which needs to be loaded
  * @returns {Object} A study descriptor object
  */
-async function RetrieveMetadata(server, studyInstanceUid) {
-  return (server.enableStudyLazyLoad !== false
+async function RetrieveMetadata(server, studyInstanceUid, filters) {
+  return (server.enableStudyLazyLoad !== false && false
```
dannyrb: `&& false` makes this always `false`?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -468,37 +468,70 @@ async function createStudyFromSOPInstanceList(server, sopInstanceList) {
  * @param {string} studyInstanceUid The Study Instance UID of the study which needs to be loaded
  * @returns {Object} A study descriptor object
  */
-async function RetrieveMetadata(server, studyInstanceUid) {
-  return (server.enableStudyLazyLoad !== false
+async function RetrieveMetadata(server, studyInstanceUid, filters) {
```
dannyrb: Update `JSDoc`, and what can be in `filters`?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -468,37 +468,70 @@ async function createStudyFromSOPInstanceList(server, sopInstanceList) {
...
 }
 
 /**
  *
  * @param {*} server
  * @param {*} studyInstanceUID
  */
-async function loadStudyMetadata(server, studyInstanceUID) {
-  const dicomWeb = new api.DICOMwebClient({
+async function loadStudyMetadata(server, studyInstanceUID, filters) {
```
dannyrb: Update JSDOC

If it feels like I'm asking "what can be in `filters`"? too much, then that might be indicative of two things:

1. Why are we passing filters so far down the callstack? Does every layer of this beast have a good purpose? Does each actually need to know our filters?
2. Is there a better pattern for passing complex values?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -468,37 +468,70 @@ async function createStudyFromSOPInstanceList(server, sopInstanceList) {
...
   });
-  return dicomWeb
-    .retrieveStudyMetadata({ studyInstanceUID })
-    .then(result => createStudyFromSOPInstanceList(server, result));
+
+  const options = {
+    studyInstanceUID,
+  };
+
+  const { seriesInstanceUID } = filters;
```
dannyrb: Ah. Here it is :Wave:

platform/viewer/src/routes/ViewerRouting.js
```diff
@@ -36,7 +36,7 @@ function ViewerRouting({ match: routeMatch, location: routeLocation }) {
     return (
       <ConnectedViewerRetrieveStudyData
         studyInstanceUids={studyUIDs}
-        seriesInstanceUids={seriesUIDs}
+        seriesInstanceUIDs={seriesUIDs}
```
dannyrb: Why the change in casing? I feel like we more consistently use `Uids`.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -118,4 +155,4 @@ class ViewerRetrieveStudyData extends Component {
   }
 }
 
-export default ViewerRetrieveStudyData;
+export default withSnackbar(ViewerRetrieveStudyData);
```
dannyrb: This is a pretty simple component. We could just make it a functional component?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -468,37 +468,70 @@ async function createStudyFromSOPInstanceList(server, sopInstanceList) {
...
     url: server.wadoRoot,
     headers: DICOMWeb.getAuthorizationHeader(server),
   });
+
   const seriesLoader = makeSeriesLoader(
     dicomWeb,
-    studyInstanceUid,
-    seriesInstanceUids
+    studyInstanceUID,
+    seriesInstanceUIDsMap
```
dannyrb: casing change?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -468,37 +468,70 @@ async function createStudyFromSOPInstanceList(server, sopInstanceList) {
...
+  try {
+    studyMetadata = await retrieveMethod.call(client, options);
+  } catch (e) {
+    // in case occurred error when retrieveSeriesMetadata, retrieve again from retrieveStudyMetadata
+    if (seriesInstanceUID) {
+      studyMetadata = await client.retrieveStudyMetadata({ studyInstanceUID });
+    } else {
+      throw e;
+    }
+  }
```
dannyrb: Wondering if we should clean this up a little to make it clearer what's going on?

0. Identify request + options
1. Primary request
2. Fallback request
3. ???
4. Profit

platform/core/src/studies/services/wado/retrieveMetadata.js
ladeirarodolfo: Definitely we can revisit and discuss about these many layers.. I dont see a purpose of having that many.
"filters" must be passed to the very end layer which does the call to retrieve series (in case of filter), because lazy load uses the first one (in our case would be the filtered) to trigger the process.

There is another logic which  we can place the filtering/promoting at the very beginning layer, but from my understanding this wouldnt be a good user experience at all (page will full load, than filter will be applied). I.e on a lazy env.. studies will be listed then filtered.

Let me know what you think?

platform/core/src/studies/retrieveStudiesMetadata.js
ladeirarodolfo: ok

platform/core/src/studies/services/wado/retrieveMetadata.js
ladeirarodolfo: debugging flag :D removing soon

platform/core/src/studies/services/wado/retrieveMetadata.js
ladeirarodolfo: ok

platform/core/src/studies/services/wado/retrieveMetadata.js
ladeirarodolfo: ok changed a bit the approach

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
ladeirarodolfo: yes.. sure.. can we cover on a separate PR?

platform/core/src/studies/services/wado/retrieveMetadata.js
ladeirarodolfo: :D updated JSDOCs

platform/viewer/src/routes/ViewerRouting.js
ladeirarodolfo: I spoke to Erik and he mentioned UID would be more appropriate.
But again.. I see places with different versions.. I`m not that confident using one or other..
totally opened to implement what you think is the best one

dannyrb: I don't have a strong preference, I just want us to be consistent. I'll poke Erik tomorrow.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -12,7 +13,7 @@ const { studyMetadataManager, updateMetaDataManager } = utils;
 class ViewerRetrieveStudyData extends Component {
   static propTypes = {
     studyInstanceUids: PropTypes.array.isRequired,
-    seriesInstanceUids: PropTypes.array,
+    seriesInstanceUIDs: PropTypes.array,
```
dannyrb: It's still confusing/odd to me that we're keeping `studyInstanceUids` but changing the casing of `seriesInstanceUIDs`. Unless we're changing the usage throughout, I'd rather we avoid doing this. At a minimum, we should use the same casing rules within the same functions/scope.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -26,19 +27,55 @@ class ViewerRetrieveStudyData extends Component {
 
   async loadStudies() {
     try {
-      const { server, studyInstanceUids, seriesInstanceUids } = this.props;
+      const { server, studyInstanceUids, seriesInstanceUIDs } = this.props;
+      const filters = {};
+
+      // Use the first, discard others
+      const seriesInstanceUID = seriesInstanceUIDs && seriesInstanceUIDs[0];
```
dannyrb: Where `seriesInstanceUIDs` is set by the route?

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -26,19 +27,55 @@ class ViewerRetrieveStudyData extends Component {
...
+    // skip in case no filter or no toast manager
+    if (!seriesInstanceUID || !snackbarContext) {
+      return;
+    }
+
+    const firstStudy = studies[0] || {};
+    const { seriesList = [] } = firstStudy;
+    const firstSeries = seriesList[0];
+
+    if (!firstSeries || firstSeries.seriesInstanceUid !== seriesInstanceUID) {
```
dannyrb: Our "filter validation" is really a test to see that _any_ results were returned after we've applied our filters to our query, right? It might be easier to just check that we returned a study with one or more series?

platform/core/src/studies/services/wado/retrieveMetadata.js
```diff
@@ -1,609 +1,30 @@
...
-  if (Array.isArray(sopInstanceList) && sopInstanceList.length > 0) {
-    const firstSopInstance = sopInstanceList[0];
-    const study = createStudy(server, firstSopInstance);
-    await addInstancesToStudy(server, study, sopInstanceList);
-    return study;
-  }
-  throw new Error('Failed to create study out of provided SOP instance list');
-}
+import RetrieveMetadataLoaderSync from './retrieveMetadataLoaderSync';
+import RetrieveMetadataLoaderAsync from './retrieveMetadataLoaderAsync';
```
dannyrb: Ah, okay. So just splitting up this logic to isolate it a bit.

platform/core/src/studies/services/wado/retrieveMetadataLoader.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Class to define inheritance of load retrieve strategy.
+ * The process can be async load (lazy) or sync load
+ *
+ * There are methods that must be implemented at consumer level
+ * To retrieve study call execLoad
+ */
+export default class RetrieveMetadataLoader {
```
dannyrb: 👋 Hello Mr. Abstract Class

platform/core/src/studies/studyUtils.js
```diff
@@ -0,0 +1,413 @@
+import DICOMWeb from '../DICOMWeb/';
```
dannyrb: If this is the shared logic for the metadata loaders, would it make more sense as "internals" that live in the same location next to those files?

dannyrb: Naming files "utilities" is a bit dangerous, in my opinion. I often think of utilities as general purpose static methods. These feel like data mappings, and a few methods for parsing data, building URLs, etc.

If it's simply shared logic you've tried to factor out, you can name it that, and put it close to the files that use it (if they're co-located). If it's generic data "mapping", then maybe we need to have a discussion (as a team) about creating a "mapping" layer where we discuss a more formal "Aggregate Root", "Viewmodel", and "Data Transfer Object" setup.

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
```diff
@@ -0,0 +1,159 @@
...
+import * as StudyUtils from '../../studyUtils';
+
+/**
+ * Search series of a given study
+ * @param {DICOMwebClient} client Dicomweb client api
+ * @param {string} studyInstanceUID The Study Instance UID to search series from;
+ * @returns {Arrays} A list of Series Instances
+ */
+async function searchStudySeries(client, studyInstanceUID) {
+  const seriesList = await client.searchForSeries({ studyInstanceUID });
```
dannyrb: What data is fetched from this? Is there anyway to apply our series specific filter with this request?

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Filter seriesList
+ * @param {Arrays} seriesList list of Series Instance UIDs
+ * @param {Object} filters object containing filter to be applied.
+ * @returns {Arrays} A list of Series Instances
+ */
+function filterStudySeries(seriesList = [], filters) {
+  let result = [...seriesList];
+  const { seriesInstanceUID: toCompare } = filters;
```
dannyrb: So at this point, we at least have _some_ metadata for _every_ series in our study. I know in our `studyList` search, we're able to apply filters to fields. Can we apply a filter to the seriesInstanceUid field when we query?

See: https://www.dicomstandard.org/dicomweb/query-qido-rs/
First `key`/`value` matching

Once we have a list of all the `seriesInstanceUids` that qualify for our study + filter, we can then do the heavier metadata and instance calls? Also lets us skip filtering on the client (less data transmission)

platform/core/src/studies/services/wado/retrieveMetadataLoaderSync.js
```diff
@@ -0,0 +1,78 @@
...
+   */
+  *getLoaders() {
+    const loaders = [];
+    const { filters: { seriesInstanceUID } = {}, client } = this;
+
+    if (seriesInstanceUID) {
+      loaders.push(client.retrieveSeriesMetadata);
+    }
+
+    loaders.push(client.retrieveStudyMetadata);
```
dannyrb: I think I see where/how we're ultimately filtering our result set in the `async` implementation. I'm pretty sure this is how/where we accomplish this here. I'm not 100% sure I understand why it works here. Can you give me a high level overview?

platform/core/src/studies/getSeriesInfo.js
```diff
@@ -0,0 +1,25 @@
...
+    info = Object.freeze({
+      modality,
+      isLowPriority: isLowPriorityModality(modality),
+      seriesInstanceUid: DICOMWeb.getString(series['0020000E']),
+      seriesNumber: DICOMWeb.getNumber(series['00200011'], 0) || 0,
+    });
+    series[INFO] = info;
+  }
+  return info;
+}
```
dannyrb: Is this new code? What are we using this for? Is this a common approach for mapping/persisting reusable models?

Is there a better, more descriptive name, than `getSeriesInfo`?

platform/core/src/studies/sortStudy.js
```diff
@@ -1,17 +1,96 @@
+import getSeriesInfo from './getSeriesInfo';
```
dannyrb: Can you walk me through the changes in this file? It looks like we added a fair amount of code here, and I want to understand the benefits + usage

platform/viewer/src/routes/ViewerRouting.js
ladeirarodolfo: I will change it to seriesInstanceUids!

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
ladeirarodolfo: or from ViewerRouting or from useLocation hooks. I`m using the first strategy because it was already present on the code.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
ladeirarodolfo: not actually because as per requirement, it will always return series (filtered or not)..

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
ladeirarodolfo: changed to seriesInstanceUids

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
dannyrb: Ah. Okay. So this is further down the pipe. We've hit our fallback (study retrieve), so we're more or less guaranteed to have data. We have to check the prop to see if filtering was applied.

What if the first series in the list just happens to match our filter?

platform/core/src/studies/studyUtils.js
ladeirarodolfo: ok changed it a bit..

platform/core/src/studies/getSeriesInfo.js
ladeirarodolfo: this is not new code.. just moved to a shared place where sortStudy and retrieveMetadaAsync can use...

platform/core/src/studies/sortStudy.js
ladeirarodolfo: actually there is nothing new.. I preserved old implementation of sortStudy and exposed what has being used by retrieveMetadaAsync file.. but its always a pleasure to walk you through :D

platform/core/src/studies/services/wado/retrieveMetadataLoaderSync.js
ladeirarodolfo: yes.. lets do it..

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
ladeirarodolfo: ok changed it a bit.. using api to fetch series (filtered)

ladeirarodolfo:
To test it:

Filter
Open any study with valid query param ?SeriesInstanceUID
https://viewer.ohif.org/viewer/1.3.6.1.4.1.25403.345050719074.3824.20170126085406.1?SeriesInstanceUID=2.25.856992923281239003733479634608709215918

Not Filter
Open any study without query param ?SeriesInstanceUID

Fail scenarios
Open any study with invalid query param ?SeriesInstanceUID


Studies
Study 1
https://viewer.ohif.org/viewer/1.3.6.1.4.1.25403.345050719074.3824.20170126085406.1

Valid SeriesInstanceUID values
2.25.856992923281239003733479634608709215918
2.25.374532628115939034791691984300357809891
2.25.260236938056495981644704674705700826124
2.25.216660299853408893926642702413523431206
2.25.810301644129899623236282516055099191586
1.3.6.1.4.1.25403.345050719074.3824.20170126085406.5


Study 2
https://viewer.ohif.org/viewer/1.2.840.113619.2.30.1.1762295590.1623.978668949.886

Valid SeriesInstanceUID values
1.2.840.113619.2.30.1.1762295590.1623.978668949.890
1.2.840.113619.2.30.1.1762295590.1623.978668949.887
2.25.362578567227550924517643184276356001107
2.25.930912967783862238320235018609232362156
2.25.335260806496828400742633309031250629174


How to get a valid seriesInstanceUID for given study?? (valid in case lazy load only. env has set it as true by default)
- Go to study list
- Open any study
- Go to viewer page
- Open devtools/network tab. Look for request `/studies/{studyInstanceUID}/series`
- Each item will be a series; Get related seriesInstanceUID  `Value` on key `0020000E`

PR #1118, opened 2019-10-28
fix: Set SR viewport as active by interaction

### LINKED ISSUE #1076: Unable to set SR in `ActiveViewport` by clicking thumbnail

### LINKED ISSUE #1079: User is unable to set the SR viewport as the ActiveViewport by interaction.
- Open a study containing an SR
- Create a 3x3 layout
- Add SR to an inactive layout
- Attempt to interact with Viewport
- Viewport does not become "active" (see border around viewport)

- [ ] **Should have E2E test**

n2yneRSJjA

rodrigolabs:
Fixes #1079 

## Preview

sr-active-viewport

___

✅E2E Test added


.eslintrc.json
```diff
@@ -16,6 +16,7 @@
   },
   "globals": {
     "cy": true,
+    "before": true,
```
dannyrb: 👍

platform/viewer/cypress/support/commands.js
```diff
@@ -328,6 +328,21 @@ Cypress.Commands.add('percyCanvasSnapshot', (name, options = {}) => {
   });
 });
 
+Cypress.Commands.add('setLayout', (columns = 1, rows = 1) => {
```
dannyrb: 👍

platform/viewer/cypress/integration/pwa/OHIFExtensionHTML.spec.js
```diff
@@ -22,4 +22,23 @@ describe('OHIF HTML Extension', () => {
       'Imaging Measurement Report'
     );
   });
+
+  it.only('checks if the HTML viewport has been set to active by interaction', () => {
```
dannyrb: `it.only` -- haven't seen that before. How many flavors of this are there?

extensions/dicom-html/src/DicomHtmlViewport.js
```diff
@@ -225,15 +233,34 @@ class DicomHtmlViewport extends Component {
...
 
+  setViewportActiveHandler = () => {
+    const {
+      setViewportActive,
+      viewportIndex,
+      activeViewportIndex,
+    } = this.props;
+
+    if (viewportIndex !== activeViewportIndex) {
+      setViewportActive();
```
dannyrb: If you're checking if the viewportIndexes are equal, is a throttle still required? This would short-circuit things pretty quickly, and may actually be more performant?

platform/viewer/cypress/integration/pwa/OHIFExtensionHTML.spec.js
rodrigolabs: Oh, good catch! `.only` runs only this specific test. I'm gonna remove it, my mistake!

extensions/dicom-html/src/DicomHtmlViewport.js
rodrigolabs: I was afraid to let all scroll events happen and so I used the throttle only on `onScroll`, but in fact, with this condition, we get a quick escape if the method doesn't have to be executed. Maybe I can remove this? @dannyrb

platform/viewer/cypress/support/commands.js
rodrigolabs: cc @mirnasilva this custom command can be helpful 😄

Issue #1120, opened 2019-10-29
Study Date: no mask, no validation and clear button disappears


mirnasilva:
### Description

**A couple of issues were found on Study Date fields:**
1. Once user clicks on `Clear dates` button, the button disappears and user is not able to clear the fields again if he/she adds the dates manually. The button only shows again if the user selects a date on the Date Picker.
2. User is able to insert more characters than 8, including letters. There is no validation when user inserts the value manually. i.e. `12341231234`or `asdfasdfasdf`.
3. When user manually inserts a date (typing it) the field does not add a mask into the added value. i.e. User adds `10102000`, the field should mask it as `10/10/2000`.
4. When user manually inserts an invalid date (i.e. future date), there is no warning/error message to inform that the date is invalid. Although the results are not filtered by the invalid date, it may confuse the user and mislead he/she to think that the results are correctly filtered. _**Suggestion:** display an error message after field validation and display 0 study results on the table list._
5. When user manually inserts a valid start and end dates, the search is returning 0 results. i.e. `01/01/2000 and 10/10/2019`

**Improvement:**
1. User is not able to navigate between `years` on the date picker. If the user wants to select a start date from year 2000, he/she has to navigate month-by-month until he/she gets to the desired year. 
_**Suggestion:** can we also add a year range selection? Once user clicks on the date picker header (where is placed `Month Year`), then a year range could be displayed like in [this example](http://jsfiddle.net/lemonkazi/ucfxywh6/)_ 
2. Once user opens the date picker and selects the Start Date, the selected range follows the mouse cursor on the calendar and the dates inside the range become white/light gray. The light gray color can mislead the user to think that those are invalid/unavailable dates instead of the selected range. _**Suggestion:** can we change the color of the selected range? Maybe use the same light blue of the application version label in the cell background (`#9ccef9`)._

![Screen Shot 2019-10-28 at 21 09 06](https://user-images.githubusercontent.com/13245529/67727216-4a305380-f9c7-11e9-8a03-39b75e5fba9a.png)

![Screen Shot 2019-10-28 at 21 21 38](https://user-images.githubusercontent.com/13245529/67727677-12c2a680-f9c9-11e9-9ec0-88a3f12ac409.png)

dannyrb:
I think these are good UX improvements, and some are clearly small bugs that need fixed (for our usage, anyway). Thanks for taking the time to right up a report :+1:

dannyrb:
[...]

Issue #1114, opened 2019-10-28
Search fields only searches for full string and are case sensitive


mirnasilva:
### Description

`Patient Name`, `MRN`,  `Accession` and `Description` fields are case sensitive and only searches for complete string (partial string i.e: First Name, doesn't find results).

**No results found by searching partial string:**
![bug-search-name3](https://user-images.githubusercontent.com/13245529/67709911-b39b6c80-f99d-11e9-8482-3ca2d3159724.gif)

**No results found by searching lower case:**
![bug-search-name2](https://user-images.githubusercontent.com/13245529/67707269-f149c680-f998-11e9-8edd-260f2c47f4a7.gif)

### What steps can we follow to reproduce the bug?

Scenario 1:
1. On study list page, search for Patient Name: `neptune`
2. Verify that no results were found
3. Search for Patient Name: `Neptune`
4. Verify that 3 results were found

Scenario 2:
1. Search for Patient Name: `Fall`
2. Verify that no results were found
3. Search for Patient Name: `Fall 1`
4. Verify that 1 result was found

Scenario 3:
1. Search for Description: `ct`, `CT` or `CT abdomen`
2. Verify that no results were found
3. Search for Description: `CT ABDOMEN` or `CT Abdomen`
4. Verify that 1 result was found

Scenario 4:
1. Search for Patient Name: `Mister` or `mister`
2. Verify that no results were found
3. Search for Patient Name: `MISTER`
4. Verify that 5 different results were found. In this case, substring was returned.

### Actual result

The search only returns results when we complete string matches the database. The fields are case sensitive and don't return results if the case is not the same as the database.

### Expected result

1. All search fields should not be case sensitive. User should be able to find the same results by searching with lower case, upper case, camel case, etc.

2. `Patient Name` and `Description` fields should return values that match a partial string (substring). User should be able to search patients only by their First Name, Last Name, or a Part of the description.
2.1. `MRN` and `Accession` fields should not search for substrings, only complete strings.

dannyrb:
This is tricky, because we're testing two things here.

1. The client's ability to create requests.
2. The server's ability to search/filter/return our expected result sets.

By default, our application communicates with DICOMWeb conformant APIs. DICOMWeb allows for servers to optionally support a parameter named `fuzzymatching`. You can read more about what we have the ability to [query here (QIDO-RS)](https://www.dicomstandard.org/dicomweb/query-qido-rs/).

[...]

Issue #1100, opened 2019-10-26
Test cases for study list:


dannyrb:
- Search filters
- Sort results
- Pagination
- Show rows per page x Study count
- Study Date (date picker, type date, invalid date, future date)

PR #1121, opened 2019-10-29
#1100: Study list test cases: Patient Name, MRN, Modality, Description

mirnasilva:
Test cases for issue #1100 
The majority of the tests are blocked by issues: #1114 #1120 

Hardcoded tests variations for case matching (camel, upper, lower and mixed), to guarantee that we will get always the same expected result.


platform/viewer/cypress/integration/common/OHIFStudyList.spec.js
```diff
@@ -0,0 +1,181 @@
+const expectExport = require('expect');
```
mirnasilva: need to remove this

platform/viewer/cypress/integration/common/OHIFStudyViewer.spec.js
```diff
@@ -72,30 +72,26 @@ describe('OHIF Study Viewer Page', function() {
...
-  //     .type(descriptionText);
-  //
-  //   // Confirm
-  //   cy.get('.btn-confirm').click();
-  //
-  //   //Verify if descriptionText was added
-  //   cy.get('.measurementLocation')
-  //     .should('contain.text', descriptionText);
-  // });
+  it('checks if Description can be added to measurement item under Measurements panel', () => {
```
mirnasilva: issue #1013 was fixed. Uncommenting this test.

mirnasilva:
Pending tests for Sort Results and Pagination.

dannyrb:
@mirnasilva, studylist changes were merged yesterday. You may need to pull latest and verify your selectors are still accurate.

mirnasilva:
- Fixed failures on CI
- Added tablet tests
- Refactored aliases and desktop tests
- Added `cy.waitStudyList()` custom command to wait for the search results

platform/viewer/cypress/support/commands.js
```diff
@@ -76,12 +78,18 @@ Cypress.Commands.add('isPageLoaded', (url = '/viewer/') => {
...
-Cypress.Commands.add('openStudyList', patientName => {
+Cypress.Commands.add('openStudyList', () => {
   cy.initRouteAliases();
   cy.visit('/');
   cy.wait('@getStudies');
 });
 
+Cypress.Commands.add('waitStudyList', () => {
+  cy.get('@searchResult').should($list => {
+    expect($list).to.not.have.class('no-hover');
```
dannyrb: I think checking the list length may be more reliable (if this class changes).

platform/viewer/cypress/support/aliases.js
```diff
@@ -52,3 +52,28 @@ export function initVTKToolsAliases() {
...
+  cy.get('#start-date').as('studyListStartDate');
+  cy.get('#end-date').as('studyListEndDate');
+  cy.get('#filter-modalities').as('modalities');
+  cy.get('#filter-studyDescription').as('studyDescription');
+  cy.get('[data-cy="study-list-results"] > tr').as('searchResult');
+}
+
+//Creating aliases for Study List page elements on Tablet experience
+export function initStudyListAliasesOnTablet() {
+  cy.get('.study-count').as('studyCount');
```
dannyrb: You could use these methods to resize to the correct width before setting aliases?

platform/viewer/cypress/integration/common/OHIFStudyList.spec.js
```diff
@@ -0,0 +1,328 @@
+describe('OHIF Study List', function() {
+  context('Desktop resolution', function() {
+    beforeEach(function() {
+      cy.viewport(1750, 720);
+      cy.openStudyList();
+      cy.initStudyListAliasesOnDesktop();
+      // run these tests as if in a desktop
```
dannyrb: I think you're okay to remove this comment :+1:

platform/viewer/cypress/integration/common/OHIFStudyList.spec.js
```diff
@@ -0,0 +1,328 @@
...
+    });
+
+    // //TO-TO: This test should be uncommented once issue #1114 is fixed: https://github.com/OHIF/Viewers/issues/1114
```
dannyrb: I've responded to your issue with some additional context here:

- https://github.com/OHIF/Viewers/issues/1114#issuecomment-549216117

We may want to discuss how best to test given the new information, and if we should create new tickets/stories around improving the application to clear up any misconceptions users may have with search + search capabilities.

platform/viewer/cypress/integration/common/OHIFStudyList.spec.js
```diff
@@ -0,0 +1,328 @@
...
+    });
+
+    it('searches Modality with upper case', function() {
+      cy.get('@modalities').type('MR');
+      //Wait result list to be displayed
+      cy.waitStudyList();
+      cy.get('@searchResult').should($list => {
+        expect($list.length).to.be.eq(17);
+      });
+    });
```
dannyrb: We can probably cut back to a single modality test.

platform/viewer/cypress/integration/common/OHIFStudyList.spec.js
```diff
@@ -0,0 +1,328 @@
...
+    //   //Wait result list to be displayed
+    //   cy.waitStudyList();
+    //   cy.get('@searchResult').should($list => {
+    //     expect($list.length).to.be.eq(42);
+    //   });
+    // });
+  });
+
+  context('Tablet resolution', function() {
+    beforeEach(function() {
```
dannyrb: If any of these tests are identical, but with different selectors, we may want to look into a better pattern. Not a blocker for this PR; just something to think about.

platform/viewer/cypress/support/commands.js
mirnasilva: The problem with length is that if we don't find results or the result is loading, the length will be 1. This can be misled if our search result also returns 1.

platform/viewer/cypress/support/aliases.js
mirnasilva: Yes, the resizing is being done before calling this function.

platform/viewer/cypress/integration/common/OHIFStudyList.spec.js
mirnasilva: Got it. In order to do a clean up here, I'll remove all the commented tests related to `fuzzymatching`. Once we decide to activate that, we can always go back and write one simple test to check it.

platform/viewer/cypress/integration/common/OHIFStudyList.spec.js
mirnasilva: I believe that leaving it separated will make it easier to identify possible failures. :)

PR #1128, opened 2019-10-29
fix: study list translations

dannyrb:
1. Translations are working locally, but they pull from static files that are synced manually from `locize`.
2. Hosted solution displays `key` names instead of translated text.
  - When we use `useTranslation`, we're not guaranteed that our translated text is available
  - `Locize` is enabled, so we fetch translations from its API
  - We're not using suspense, so we need to wait for the `useTranslation` hook's `ready` state

image


dannyrb:
I'm a bit stumped. This works locally, but fails when `USE_LOCIZE` is set to true in our `i18n` config.

dannyrb:
@ladeirarodolfo beat me to the solution on this one ^_^
See in-progress work here:

https://github.com/OHIF/Viewers/pull/1150

PR #1134, opened 2019-10-30
fix: get adapter store picker to show

dannyrb:
image

image

### DIFF
--- a/platform/viewer/src/studylist/StudyListRoute.js
+++ b/platform/viewer/src/studylist/StudyListRoute.js
@@ -1,9 +1,9 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useContext } from 'react';
 import PropTypes from 'prop-types';
 import Dropzone from 'react-dropzone';
 import OHIF from '@ohif/core';
 import { withRouter } from 'react-router-dom';
-import { withTranslation } from 'react-i18next';
+import { useTranslation } from 'react-i18next';
 import {
   StudyList,
   PageToolbar,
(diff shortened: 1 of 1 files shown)


pieper:
Thanks! LGTM - do you want me to try it out on my sandbox or will you just merge to master?

pieper:
Also when I try to go to this dicom store I get the error pasted below.

https://idc-sandbox-000.firebaseapp.com/projects/chc-tcia/locations/us-central1/datasets/breast-diagnosis/dicomStores/breast-diagnosis

image

dannyrb:
@pieper, haha, losanges. I can remove the styling.

I'll try to reproduce and diagnose. You see this issue when selecting with the picker, or only when attempting to visit the route directly?

dannyrb:
@pieper, I must not have access to the same data as you. Despite a different app token than what I use for local debugging, I see the same datasets:

Project: `OHIF Cloud Healthcare` (ohif-cloud-healthcare)

FWIW, I'm able to switch between datasets with what I do have access to. When I try to visit the route you've provided directly, I get a `403` (which we could handle more gracefully as well)

pieper:
@dannyrb you probably need to fill out the form to access the TCIA data:

https://cloud.google.com/healthcare/docs/resources/public-datasets/tcia#cloud-healthcare-api

This will give you access to chc-tcia project dicom stores.  I think it's worth doing since there are a lot of good test datasets there (MR, CT, PET, 4d, etc).

Thanks again for your help with this 👍

dannyrb:
@pieper a couple of odd things are causing that particular dataset to fail. The main culprit being that `getName` for `patientName` is returning an object instead of a string value.

We're also seeing issues where both `modality` and `modalitiesInStudy` are undefined. I'm drafting a small PR that should allow us to handle these situations gracefully. I'll ping you when it's up/merged.

pieper:
@dannyrb did you have any trouble signing up for the tcia datasets?  Andrey also says he can add people to that project directly through our contact at Google.

PR #1141, opened 2019-10-31
Issue 879 viewer route query param not filtering but promoting

ladeirarodolfo:
### Included items
- Add config flag filterQueryParam:
   - true: filter series in case there is a valid query param value
   - false: does NOT filter but instead promote (bring to top) the series of given query param value
- Promote implementation
- Refactor ViewerRetrieveStudyData to be React functional component
- It uses implementation of #1117

#### DO NOT MERGE UNTIL
#1117 

### Screenshoot
Untitled_-Oct-31_-2019-5_42-PM


dannyrb:
Will review after #1117 is merged (fyi)

Issue #1101, opened 2019-10-26
Simple E2E test(s) to verify JSON URL routing continues to work


dannyrb:
https://deploy-preview-1093--ohif.netlify.com/pwa/viewer?url=https://s3.eu-central-1.amazonaws.com/ohif-viewer/JSON/PTCTStudy.json

or

http://localhost:3000/viewer?url=https://s3.eu-central-1.amazonaws.com/ohif-viewer/JSON/PTCTStudy.json

PR #1145, opened 2019-11-01
Tests for routes: PT/CT json, studies, series, frames, metadatas

mirnasilva:
Tests cases for issue #1101 

Verifying if the following routes are being requested:
```
**/PTCTStudy/**
**/studies/**
**/studies/**/series
**/studies/**/series/**/metadata
**/studies/**/series/**/instances/**/frames/1
```


platform/viewer/cypress/integration/common/OHIFRoutes.spec.js
```diff
@@ -0,0 +1,42 @@
+// type definitions for Cypress object "cy"
+/// <reference types="cypress" />
```
dannyrb: Why typedef comments?

platform/viewer/cypress/integration/common/OHIFRoutes.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    cy.visit(
+      '/viewer?url=https://s3.eu-central-1.amazonaws.com/ohif-viewer/JSON/PTCTStudy.json'
+    );
+    cy.server();
+    cy.route('GET', '**/PTCTStudy/**').as('getPTCTStudy');
+
+    cy.wait('@getPTCTStudy.all');
+    cy.get('@getPTCTStudy').should($route => {
+      expect($route.status).to.be.eq(200);
+    });
```
dannyrb: I tried to think about how I might test this, and I can't think of anything more resilient, except maybe creating a `json` file that's part of our app we use specifically for this test (instead of an external resource).

Can we drop a PercySnapshot in here? Probably after a minimum thumbnail load or something similar?

platform/viewer/cypress/integration/common/OHIFRoutes.spec.js
```diff
@@ -0,0 +1,42 @@
...
+
+    cy.wait(['@getStudies', '@getSeries', '@getMetadata', '@getFrames']).spread(
+      (getStudies, getSeries, getMetadata, getFrames) => {
+        expect(getStudies.status).to.be.eq(200);
+        expect(getSeries.status).to.be.eq(200);
+        expect(getMetadata.status).to.be.eq(200);
+        expect(getFrames.status).to.be.eq(200);
+      }
+    );
+  });
```
dannyrb: I think we already have this route covered with our other tests. This is interesting, because it forces us to test that we're hitting DICOMWeb endpoints, but I don't know that it buys me more confidence that this route is working as expected.

We can probably skip the basic `studies` route -- maybe even add a comment at the top of the file that we exclude tests for it because the bulk/all of our other study/viewer tests use this route?

platform/viewer/cypress/integration/common/OHIFRoutes.spec.js
mirnasilva: This typedef automatically activates an Intelligent Code Completion [(check it here)](https://docs.cypress.io/guides/tooling/intelligent-code-completion.html#Writing-Tests). We can remove it if it doesn't make sense to keep it. (but is pretty cool :) )

platform/viewer/cypress/integration/common/OHIFRoutes.spec.js
mirnasilva: Hmm. Not sure exactly how we can navigate/load the local `json` file. The following code isn't working, any thoughts?

```
cy.fixture('./../fixtures/PTCTStudy.json').as('PTCTStudyJSON');

cy.visit('/');
cy.request('viewer?url=PTCTStudyJSON');
```

dannyrb:
@mirnasilva, tests are failing

PR #1148, opened 2019-11-04
feat: 🎸 Change ImageThumb to functional and change thumb on upd

ladeirarodolfo:
### Includes
- It changes ImageThumbnail to functionalComponent
- It set Thumb component key to be based on displayInstanceUid value (old implementation)

### A complement of this impl will go on (but its not a blocker for current one):
#1141


platform/ui/src/components/studyBrowser/ImageThumbnail.js
```diff
@@ -12,100 +12,106 @@ import ViewportLoadingIndicator from '../../viewer/ViewportLoadingIndicator';
...
-    stackPercentComplete: PropTypes.number.isRequired,
-  };
-
-  static defaultProps = {
-    error: false,
-    stackPercentComplete: 0,
-    width: 217,
-    height: 123,
-  };
+  const [loading, setLoading] = useState(false);
```
dannyrb: Can we do `isLoading` and `setIsLoading`?

PR #1150, opened 2019-11-05
Issue branch from danny experimental changes pr 1128

dannyrb:
### It includes
- Created a hooks for managing mediaQuery
- Wrapper components to render only if useTranslation is ready


dannyrb:
Per Slack:

> I'm okay with leveraging this trick. I created a PR from your work. Let's clean up the bits that were just testing and update useMedia to support passing in one or more values to watch for it's useEffect?

dannyrb:
image

platform/ui/src/components/studyList/StudyList.js
```diff
@@ -104,10 +104,13 @@ function StudyList(props) {
     smallTableMeta
   );
 
+  if (!ready) {
+    return null;
+  }
```
dannyrb: Can we do this in-line by the `return` for this component? Potentially aliased as `translationsAreReady` or something similar?

platform/ui/src/components/studyList/TableSearchFilter.js
```diff
@@ -49,56 +51,58 @@ function TableSearchFilter(props) {
     },
   ];
 
-  return meta.map((field, i) => {
-    const { displayText, fieldName, inputType } = field;
-    const isSortField = sortFieldName === fieldName;
-    const sortIcon = isSortField ? sortIconForSortField : sortIcons[0];
+  return ready
```
dannyrb: Similar aliasing here would make it clearer what we're waiting for to be ready.

platform/ui/src/hooks/useMedia.js
```diff
@@ -17,36 +33,37 @@ import { useState, useEffect } from 'react';
  * @returns
  */
 function useMedia(queries, values, defaultValue) {
-  // Array containing a media query list for each query
-  const mediaQueryLists = queries.map(q => window.matchMedia(q));
+  const memoizedQueryMap = useMemo(() => getMediaQueryMap(queries), [queries]);
+  const memoizedValue = useMemo(
+    () => getValue(memoizedQueryMap, values, defaultValue),
+    [mediaQueryLists, values]
+  );
```
dannyrb: So we're trying to be a bit more efficient.
Were we seeing a lot of re-renders? Does this prevent renders unless:

1. A new query is added to the map
2. One or more queries returns a new value (true/false) for matching the current window width

If so, might be worth adding to the JSDoc for `useMedia` -- this is a neat trick

platform/ui/src/hooks/useMedia.js
```diff
@@ -17,36 +33,37 @@ import { useState, useEffect } from 'react';
...
-    const index = mediaQueryLists.findIndex(mql => mql.matches);
+  const addMediaQueryListeners = mediaQueryLists => {
+    // Set a listener for each media query with above mediaQueryListener as callback.
+    mediaQueryLists.forEach(mql => mql.addListener(mediaQueryListener));
+  };
 
-    // Return related value or defaultValue if none
-    return typeof values[index] !== 'undefined' ? values[index] : defaultValue;
+  const removeMediaQueryListeners = (mediaQueryLists = []) => {
+    mediaQueryLists.forEach(mql => mql.removeListener(mediaQueryListener));
```
dannyrb: Ah. Good call on adding/removing. I could see this causing issues if these weren't removed as we navigate routes. What currently triggers clean-up?

platform/ui/src/hooks/useMedia.js
```diff
@@ -17,36 +33,37 @@ import { useState, useEffect } from 'react';
...
-    [] // Empty array ensures effect is only run on mount and unmount
-  );
+  useEffect(() => {
+    addMediaQueryListeners(mediaQueryLists);
+  }, [mediaQueryLists]);
+
+  useEffect(() => {
+    // Remove listeners on cleanup
+    return () => removeMediaQueryListeners(mediaQueryLists);
+  }, []);
```
dannyrb: `[]` means this is only run once, but `[mediaQueryLists]` above means the effect is run each time `mediaQueryLists` is modified.

I'm not 100% sure I follow the logic of this refactor -- I may need a walkthrough. I've only used `useMemo` once or twice, so I'm likely missing something obvious.

platform/ui/src/components/studyList/StudyList.js
ladeirarodolfo: yes :D

dannyrb:
Relayed feedback that `useMedia` breakpoints are not triggering at appropriate times:


image

platform/ui/src/hooks/useMedia.js
```diff
@@ -1,54 +1,223 @@
...
+    return content;
+  };
 
-  return value;
-}
+  return getContent();
+};
 
-export default useMedia;
+export { useDisplayMediaSize, useDisplayMediaContent };
```
dannyrb: This feels significantly more complex than the original implementation. Is all of this necessary? Can you walk me through:

- Issues with the original `useMedia` implementation
- Issues with the second iteration
- How the current solution is optimal for solving `useMedia`'s original intent?

platform/viewer/src/studylist/StudyListRoute.js
```diff
@@ -58,8 +58,12 @@ function StudyListRoute(props) {
   const [pageNumber, setPageNumber] = useState(0);
   const appContext = useContext(AppContext);
   // ~~ RESPONSIVE
-  const displaySize = useMedia(
-    ['(min-width: 1750px)', '(min-width: 1000px)', '(min-width: 768px)'],
+  const displaySize = useDisplayMediaSize(
```
dannyrb: So `useDisplayMediaSize` is the previous implementation of `useMedia`

dannyrb: At least from an API standpoint

ladeirarodolfo: yes.. but I`m ok if you believe we should preserve it as it was

platform/ui/src/hooks/useMedia.js
```diff
@@ -1,54 +1,235 @@
...
+  const getContent = () => {
+    const content =
+      displaySize in contentArrayMap
+        ? contentArrayMap[displaySize]
+        : defaultContent;
+    return content;
+  };
+
+  return getContent();
+};
```
dannyrb: Instead of calling `useDisplayMediaSize` internally, we should probably setup the JSDOC Example to show:

```js
const size = useMedia(
    // Media queries
    ['(min-width: 1500px)', '(min-width: 1000px)', '(min-width: 600px)'],
    // Value to return for matched media query
    ['large', 'medium', 'small'],
    // Default value
    'medium');

// Pass size to a child component, and in child component:
const content = useContent(
    props.size,
    {"small": ObjectA, "medium": ObjectB, "large": ObjectB },
    "medium");
```

dannyrb: I'm not 100% sure `useContent` is necessary though. As you could just do:

```js
const contentTemplates = { /* ... */ };
const defaultTemplate = contentTemplates ["medium"];

const content = props.size ? contentTemplates[props.size] : defaultTemplate;
```

platform/ui/src/components/studyList/StudyList.js
```diff
@@ -208,6 +221,7 @@ StudyList.propTypes = {
   }).isRequired,
   onFilterChange: PropTypes.func.isRequired,
   studyListDateFilterNumDays: PropTypes.number,
+  displaySize: PropTypes.string,
```
dannyrb: Can we do a value, oneOf for each of the options?

dannyrb:
@ladeirarodolfo, I'm not sure why, but these changes cause the DateRange picker to fail. You may want to connect with @rodrigolabs to see what's up. 

https://viewer-dev.ohif.org/ continues to show the expected behavior with @rodrigolabs latest changes.

ladeirarodolfo:
> @ladeirarodolfo, I'm not sure why, but these changes cause the DateRange picker to fail. You may want to connect with @rodrigolabs to see what's up.
> 
> https://viewer-dev.ohif.org/ continues to show the expected behavior with @rodrigolabs latest changes.

There was a merge conflict on that file.. I will double check it

platform/ui/src/components/studyList/StudyList.js
ladeirarodolfo: should I? The idea was not to allow dev set any value he wants? Not necessary would be a value oneOf ['small','medium','large'].. Or should I restrict into it? If thats the case I would say to turn it into a app config and use from there

dannyrb: You're correct. Sleepy brain was lying to me here. Thank you!

PR #1153, opened 2019-11-06
improvement: Date Picker

rodrigolabs:
There are some UX improvements and bugs to fix in the Study Date filter (DatePicker).

- [x] Add i18n in the DatePicker (for labels and date format)
- [x] Create a `select` component that allows users to navigate between months/years easily.
- [x] CSS adjustments
- [ ] Clear dates button is not being displayed when typing -- **`react-dates` doesn't show the clear button when user type a date. `need to find an alternative solution`.**
- [ ] Apply mask when typing: `MM/DD/YYYY` -- **`react-dates` doesn't allow us to use a mask in the input. `need to find an alternative solution`.**
- [ ] User is able to manually type an invalid date `Display an error message after field validation`.
- [x] When the user manually inserts a valid start and end dates, the search is returning 0 results. i.e. `01/01/2000` and `10/10/2019`
- [ ] User is able to insert more characters than 8, including letters. There is no validation when user inserts the value manually. i.e. `12341231234` or `asdfasdfasdf`. -- **we have no control over the `input` events. `need to check`**

---

## Preview

---

Related #1120


platform/i18n/src/locales/en-US/Common.json
```diff
@@ -1,4 +1,5 @@
 {
+  "Close": "Close",
```
dannyrb: Were these changes made in `Locize` and then pulled?

dannyrb:
@rodrigolabs, taking a look at your changes, these may be good enough for now. We may want to style the dropdowns slightly so they are consistent across browsers.

platform/i18n/src/locales/en-US/Common.json
rodrigolabs: Yes -- all ` i18n` changes in this PR were pulled from Locize.

rodrigolabs:
> @rodrigolabs, taking a look at your changes, these may be good enough for now. We may want to style the dropdowns slightly so they are consistent across browsers.

I've just pushed an update about the dropdown style, you can see the updated dropdown preview in the PR description :)

platform/viewer/src/studylist/StudyListRoute.js
```diff
@@ -254,7 +256,6 @@ function StudyListRoute(props) {
           onSort={handleSort}
           filterValues={filterValues}
           onFilterChange={handleFilterChange}
-          studyListDateFilterNumDays={appConfig.studyListDateFilterNumDays}
```
dannyrb: Why remove this?

platform/ui/src/components/studyList/TableSearchFilter.js
```diff
@@ -76,14 +69,14 @@ function TableSearchFilter(props) {
...
             startDateId="start-date"
-            endDate={studyListDateFilterNumDays ? defaultEndDate : null}
+            endDate={values.studyDateTo || null}
             endDateId="end-date"
             // TODO: We need a dynamic way to determine which fields values to update
             onDatesChange={({ startDate, endDate, preset = false }) => {
-              onValueChange('studyDateTo', startDate);
-              onValueChange('studyDateFrom', endDate);
+              onValueChange('studyDateFrom', startDate);
+              onValueChange('studyDateTo', endDate);
```
dannyrb: Hah. Good catch. Thank you.

platform/ui/src/components/studyList/CustomDateRangePicker.js
```diff
@@ -7,39 +7,25 @@ import './CustomDateRangePicker.css';
...
-    if (this.props.onDatesChange) {
-      this.props.onDatesChange({ startDate, endDate, preset });
-    }
-  }
+  const {
+    onDatesChange,
+    startDate,
+    endDate,
+    presets,
+    ...dateRangePickerProps
```
dannyrb: I normally don't love spreading props; but because this is a pass-through component, I think we're okay :+1:

platform/viewer/src/studylist/StudyListRoute.js
rodrigolabs: My mistake, already fixed!

Issue #1140, opened 2019-10-31
Blank MPR windows


pieper:
### Describe the Bug

MPR mode doesn't work for this study:

https://idc-sandbox-000.firebaseapp.com/projects/idc-sandbox-000/locations/us-central1/datasets/sdp-dev/dicomStores/scratch/study/61.7.22285965616260355338860879829667630274

### What steps can we follow to reproduce the bug?

* Open the study
* click 2D MPR mode
* reformatting progress messages appear, but the windows remain black
* tools appear to be working, but nothing ever shows up

pieper:
Note that this is a 661 slice CT, so that's 338MB of texture memory.

dannyrb:
Is there a separate form or approval process to gain access to the `idc-sandbox-000` project? If it's not available at this time, is there another way for you to share this study?

pieper:
Sorry, yes, I'll request access for you.

This gets you to the same study pulled from TCIA:

https://idc-sandbox-000.firebaseapp.com/projects/chc-tcia/locations/us-central1/datasets/ct-lymph-nodes/dicomStores/ct-lymph-nodes/study/61.7.22285965616260355338860879829667630274

dannyrb:
In `getOrCreateVolume`, we see that `imageMetadata0` has an undefined width and center. Setting `lower`/`upper` range to `NaN` doesn't throw an error or warning, but it does appear to be the issue.

If we instead use any number as a fallback, everything begins working again.

PR #1157, opened 2019-11-07
fix: mpr2d vtkjs viewport does not render if range is set to `NaN` values

dannyrb:
fixes: https://github.com/OHIF/Viewers/issues/1140

For our example study, the `windowWidth` and `windowRange` were undefined. This had us calling:

```js
    volumeActor
      .getProperty()
      .getRGBTransferFunction(0)
      .setRange(lower, upper);
```

With a `NaN` `lower` and `upper` bounds.

VTK does not throw or warn, and subsequent calls to update the range do not fix the issue. The only way to mitigate is to make sure we only set valid values.

We currently set the range for the `RGBTransferFunction` in two places in the extension's code:

- `vtk/src/commandsModule.js`
- `vtk/src/OHIFVTKViewport.js`

As a follow-up PR, to make things a bit more resilient, we could:

- Move `setVoi` in `vtk/src/commandsModule.js` to the PublicAPI of our `react-vtkjs-viewport`
    - This would become the API for updating a viewer components window levels
        - Display value, and RGBTransferFunction

OR

- Create a context for viewport levels in `@ohif/viewer`
- Update the context to set level values on the viewport
- Have the viewport implement the `RGBTransferFunction` logic

Up for discussion, but I didn't want to blaze a trail without discussion. We haven't landed on the best way to surface API just yet.


dannyrb:
CC: @pieper for awarness

swederik:
Here's a test volume: https://deploy-preview-1157--ohif.netlify.com/pwa/viewer/61.7.102432584850582937613178565576727689438

JamesAPetts:
> Move setVoi in vtk/src/commandsModule.js to the PublicAPI of our react-vtkjs-viewport

I'm happy to do this, but we need to make sure we only call it once and not once per API, as the volume is accessed via the API, and the rgbTransferFunction set for the volume, which is a bit hairy, but still makes sense.

We'll have to keep track of what volumes we feed to each viewport if/when we get to displaying multiple different vtk volumes in multiple viewport simultaneously. As then we would have to potentially sync them differently.

PR #1158, opened 2019-11-08
fix: Add a fallback metadata provider which pulls metadata from WADO-…

swederik:
…RS cache when WADO-URI is used.

Marked as draft because this is just a proof of concept fix, the code needs to be cleaned up.


platform/viewer/public/config/default.js
```diff
@@ -11,8 +11,8 @@ window.config = {
         qidoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
         wadoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
         qidoSupportsIncludeField: true,
-        imageRendering: 'wadors',
-        thumbnailRendering: 'wadors',
+        imageRendering: 'wadouri',
```
swederik: didn't mean to commit this but it can be used for testing

JamesAPetts:
remove your debugger comment and this looks like a good stop-gap solution to me.

Issue #1049, opened 2019-10-15
PDF viewer frame can't be closed


alcheb:
Hello there, first timer here. Apologies in advance. I am not a developer and my contributing skills are close to inexistent. I use OHIF primarily to look at liver imagery (CT/MR).

## Bug Report
When opening a PDF file in the viewer (DOC modality) it is not possible to close it.

### Describe the Bug
Replacing the open PDF viewer frame with another stack by drag-and-drop is not possible. Changing the layout doesn't help either. The only way around I found, besides reloading the page, is entering MPR mode on another stack and then exiting it.

### What steps can we follow to reproduce the bug?

1. Open a study containing a PDF file/DOC modality (*e.g.* https://viewer.ohif.org/viewer/1.2.826.0.13854362241694438965858641723883466450351448)
2. Drag-and-drop the DOC stack to open it
3. Try to close it

Thanks for your help!
Al

dannyrb:
Thanks, @alcheb! No need to be a developer to help out. Reproducible bug reports like this are incredibly valuable ^_^

mirnasilva:
Hi team! Amazing work!
Great job, @igoroctaviano ! I just did a round of tests and everything is working fine. 
The issue was fixed and I'm moving it to Done!
Thanks!

PR #1162, opened 2019-11-11
fix: 🐛 Update ohif pdf extension to optionally use pdfjs

igoroctaviano:
Update ohif pdf extension of optionally use pdfjs

Closes: #1049


igoroctaviano:
Hey @dannyrb, maybe you want to take a look and test it to see if this solution is a fit for the issue #1049 ?

dannyrb:
Haha, the hoops we jump through for drag-n-drop. I think something like this is probably the better long-term solution. We can't guarantee all browsers will render PDFs the same way, or with the same features exposed, if we rely on native functionality.

That being said, the default features and rendering aren't as great as what FireFox or Chrome give me out of the box (go figure).

If you've tried the obvious solutions for drag-n-drop and are hitting a wall, I think this is the sacrifice I'd rather make. @swederik should probably be the one to pull the trigger though.

I'll queue up some time to ask him about this one.

dannyrb:
👢 💨 🤖 
Let's action this before EOW

swederik:
I'm fine with this PR, even though I sort of hate the UI. 

Some minor changes I'd like to see:
- My mousewheel won't scroll the PDF by default for some reason. The only way to scroll is to manually grab the scrollbar in the div. That doesn't work at all (bar doesn't respond) if the PDF is too zoomed in.
- We should dynamically import this component like we do with the DICOM Microscopy Viewer: https://github.com/OHIF/Viewers/blob/master/extensions/dicom-microscopy/src/index.js#L4

dannyrb:
@igoroctaviano, can you look into @swederik's suggestions? Let me know if you have any questions.

igoroctaviano:
> * https://github.com/OHIF/Viewers/blob/master/extensions/dicom-microscopy/src/index.js#L4

Alright. I'll check your suggestions asap, they are great, thanks.

igoroctaviano:
> @igoroctaviano, can you look into @swederik's suggestions? Let me know if you have any questions.

@dannyrb Sure! ill let you know if anything comes up. 👍

extensions/dicom-html/src/index.js
```diff
@@ -3,7 +3,7 @@ import OHIFDicomHtmlSopClassHandler from './OHIFDicomHtmlSopClassHandler.js';
 
 const OHIFDicomHtmlViewport = asyncComponent(() =>
   import(
-    /* webpackChunkName: "OHIFDicomHtmlViewport" */ './OHIFDicomHtmlViewport.js'
+    /* webpackChunkName: "ConnectedOHIFDicomPDFViewport" */ './OHIFDicomHtmlViewport.js'
```
swederik: This shouldn't have been changed, it's in the wrong extension.

igoroctaviano: I was removing those comment from all imports but I have added them back, missed this one. Ill update asap, thanks!

igoroctaviano: Done.

extensions/dicom-pdf/package.json
```diff
@@ -37,7 +37,8 @@
   "dependencies": {
     "@babel/runtime": "^7.5.5",
     "classnames": "^2.2.6",
-    "lodash.isequal": "^4.5.0"
+    "lodash.isequal": "^4.5.0",
+    "pdfjs-dist": "^2.2.228"
```
dannyrb: 👀

dannyrb: Had to sanity check for myself. Much more usage than `pdfjs` package.
`13.7 MB` -- Holy cow

dannyrb: This one comes in at `400kb` https://github.com/wojtekmaj/react-pdf
?

dannyrb:
I'm seeing the following issues on the Netlify Preview:

image
image

extensions/dicom-pdf/src/ConnectedOHIFDicomPDFViewer.js
```diff
@@ -0,0 +1,33 @@
...
+const { setViewportActive } = OHIF.redux.actions;
+
+const mapStateToProps = (state, ownProps) => {
+  const { viewportIndex, byteArray } = ownProps;
+  const { activeViewportIndex } = state.viewports;
+
+  return {
+    viewportIndex,
+    activeViewportIndex,
+    byteArray,
```
dannyrb: advantage of mapping `byteArray` to `byteArray`?

extensions/dicom-pdf/package.json
igoroctaviano: > This one comes in at `400kb` https://github.com/wojtekmaj/react-pdf
> ?

@dannyrb I'm not so sure, it uses pdf-dist under the hood:

Screen Shot 2019-12-13 at 09 53 39

extensions/dicom-pdf/src/ConnectedOHIFDicomPDFViewer.js
igoroctaviano: Not at all, removed. ⭐️

extensions/dicom-pdf/package.json
dannyrb: Ah. I was going off the value I found on NPM's website: https://www.npmjs.com/package/react-pdf

It would be odd if it didn't factor in dependencies. I'm curious if it does anything fancy to reduce the bundle's size, or handles code splitting/workers on our behalf.

igoroctaviano:
> I can review after the netlify preview is working 👍

I have added the worker src, now it should work ;)

extensions/dicom-pdf/src/DicomPDFViewport.js
```diff
@@ -1,23 +1,135 @@
...
+        pdf: null,
+        scale: 1,
+        canvas: null,
+      },
+    };
+  }
 
   static propTypes = {
     byteArray: TypedArrayProp.uint8,
+    useNative: PropTypes.bool,
```
dannyrb: Ah! This is clever. If we're going to `useNative` to toggle behavior, we could also code-split the `pdfjs` imports in mounted or updated? Then, if we're using the `useNative` prop, we never pull in pdfjs

extensions/dicom-pdf/src/DicomPDFViewport.js
```diff
@@ -1,23 +1,135 @@
...
 class DicomPDFViewport extends Component {
-  state = {
-    fileURL: null,
-    error: null,
-  };
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      [this.props.viewportIndex]: {
```
dannyrb: If each component is only ever tied to one viewport index, why would we use this as a key?

extensions/dicom-pdf/src/DicomPDFViewport.js
```diff
@@ -36,50 +148,156 @@ class DicomPDFViewport extends Component {
...
+      setViewportActive,
+      viewportIndex,
+      activeViewportIndex,
+    } = this.props;
 
-    this.renderPDF(dataSet, this.props.byteArray);
-  }
+    if (viewportIndex !== activeViewportIndex) {
+      setViewportActive(viewportIndex);
+    }
```
dannyrb: You could probably avoid passing `activeViewportIndex` as a prop, if you instead pass in a `setViewportActive.bind(activeViewportIndex)` as the `setViewportActive` prop.

extensions/dicom-pdf/src/DicomPDFViewport.js
```diff
@@ -1,23 +1,135 @@
...
+    super(props);
+
+    this.state = {
+      [this.props.viewportIndex]: {
+        fileURL: null,
+        error: null,
+        currentPageIndex: 1,
+        pdf: null,
+        scale: 1,
+        canvas: null,
```
dannyrb: We track this, but it looks like we don't use it's value after it's set?

extensions/dicom-pdf/src/DicomPDFViewport.js
```diff
@@ -1,23 +1,135 @@
...
+  };
+
+  async componentDidMount() {
+    const dataSet = this.parseByteArray(this.props.byteArray);
+    const fileURL = this.getPDFFileUrl(dataSet, this.props.byteArray);
+
+    this.setState(state => ({
+      ...state,
+      [this.props.viewportIndex]: {
+        ...state[0],
```
dannyrb: I'm not sure I understand this spread. When would we have a `state[0]` other than when the `viewportIndex` is already `0`?

dannyrb:
@mirnasilva, can you give this a look and compare it to production?

- https://deploy-preview-1162--ohif.netlify.com/pwa/viewer/1.2.826.0.13854362241694438965858641723883466450351448

This is mostly a bug fix for the "unable to drag-n-drop a thumbnail to switch from an active PDF Viewport to a Cornerstone Viewport"

extensions/dicom-pdf/src/DicomPDFViewport.js
igoroctaviano: Thats right, updated, thanks.

extensions/dicom-pdf/src/DicomPDFViewport.js
igoroctaviano: Not necessary now, removed, thankyou.

extensions/dicom-pdf/src/DicomPDFViewport.js
igoroctaviano: Totally right, updated. 🏅

igoroctaviano:
> A couple of small questions, mostly for my own understanding.
> The behavior and code look clean enough for merge.
> 
> I encountered one small issue when doing `9x9` grid of PDF documents, and zooming any viewport other than the first. After many zooms, the scrollbars of the first viewport expanded to match. I'm not sure it's worth investigating, just making a note.

Hi @dannyrb, good points. Updated.
About the 9x9 grid, this happens because the canvas 

> A couple of small questions, mostly for my own understanding.
> The behavior and code look clean enough for merge.
> 
> I encountered one small issue when doing `9x9` grid of PDF documents, and zooming any viewport other than the first. After many zooms, the scrollbars of the first viewport expanded to match. I'm not sure it's worth investigating, just making a note.

@dannyrb thanks for catching that. I have fixed it using refs. 🎖

dannyrb:
Thanks, @igoroctaviano! If someone masochist wants to view 9 PDFs at the same time, they can now do so without even a minor inconvenience:

image

dannyrb:
@mirnasilva, I'm merging this. We can review and assess impact to existing test suite and deploy-ability before adding this to the two-weeks-from-now-release

@igoroctaviano, this may be specific to the PDF we're trying to view, but it's worth noting I see these (non-breaking) console exceptions:

image

mirnasilva:
Hi @dannyrb  and @igoroctaviano ! Amazing work!
Great job, @igoroctaviano ! I just did a round of tests and everything is working fine.
Thanks!

PR #1165, opened 2019-11-12
refactor: 💡 Improve hotkeyPreferences proptypes / add lowercase

igoroctaviano:
Improve hotkeyPreferences proptypes and lowercase event keys to guard
against different browsers casings

### DIFF
--- a/platform/ui/src/components/userPreferencesForm/HotKeysPreferences.js
+++ b/platform/ui/src/components/userPreferencesForm/HotKeysPreferences.js
@@ -10,7 +10,13 @@ import PropTypes from 'prop-types';
 
 export class HotKeysPreferences extends Component {
   static propTypes = {
-    hotkeyDefinitions: PropTypes.array.isRequired,
+    hotkeyDefinitions: PropTypes.arrayOf(
+      PropTypes.shape({
+        commandName: PropTypes.string,
+        keys: PropTypes.arrayOf(PropTypes.string),
+        label: PropTypes.string,
+      })
+    ).isRequired,
   };
 
   constructor(props) {
@@ -76,7 +82,7 @@ export class HotKeysPreferences extends Component {
         specialKeyName ||
         keyDownEvent.key ||
         String.fromCharCode(keyDownEvent.keyCode);
-      pressedKeys.push(keyName);
+      pressedKeys.push(keyName.toLowerCase());
     }
 
     this.updateHotKeysState(commandName, pressedKeys.join('+'));
@@ -123,7 +129,7 @@ export class HotKeysPreferences extends Component {
     const hotKey = this.state.hotKeys[hotKeyIndex];
     const keys = hotKey.keys[0];
     const pressedKeys = keys.split('+');
-    const lastPressedKey = pressedKeys[pressedKeys.length - 1];
+    const lastPressedKey = pressedKeys[pressedKeys.length - 1].toLowerCase();
 
     // clear the prior errors
     this.setState({ errorMessages: {} }, () => {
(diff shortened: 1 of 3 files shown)


platform/ui/src/components/userPreferencesForm/HotKeysPreferences.js
```diff
@@ -10,7 +10,13 @@ import PropTypes from 'prop-types';
...
 export class HotKeysPreferences extends Component {
   static propTypes = {
-    hotkeyDefinitions: PropTypes.array.isRequired,
+    hotkeyDefinitions: PropTypes.arrayOf(
+      PropTypes.shape({
+        commandName: PropTypes.string,
+        keys: PropTypes.arrayOf(PropTypes.string),
+        label: PropTypes.string,
+      })
+    ).isRequired,
```
dannyrb: Just curious, what happens if `hotkeyDefinitions` is an empty array?

platform/ui/src/components/userPreferencesForm/UserPreferences.js
```diff
@@ -15,7 +15,13 @@ export class UserPreferences extends Component {
...
   // TODO: Make this more generic. Tabs should not be restricted to these entries
   static propTypes = {
-    hotkeyDefinitions: PropTypes.array.isRequired,
+    hotkeyDefinitions: PropTypes.arrayOf(
+      PropTypes.shape({
+        commandName: PropTypes.string,
+        keys: PropTypes.arrayOf(PropTypes.string),
+        label: PropTypes.string,
+      })
+    ),
```
dannyrb: Why required above, but not here?

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
```diff
@@ -15,7 +15,13 @@ class UserPreferencesForm extends Component {
...
     onResetToDefaults: PropTypes.func,
     windowLevelData: PropTypes.object,
-    hotkeyDefinitions: PropTypes.array,
+    hotkeyDefinitions: PropTypes.arrayOf(
+      PropTypes.shape({
+        commandName: PropTypes.string,
+        keys: PropTypes.arrayOf(PropTypes.string),
+        label: PropTypes.string,
+      })
+    ),
```
dannyrb: Why required above, but not here?

platform/ui/src/components/userPreferencesForm/HotKeysPreferences.js
igoroctaviano: The component will be rendered without any hotkey field. Do you think it's worth to conditionally render an informative text that no hotkey defs were found?

dannyrb: I think something like:

> No hotkeys are configured for this application. Hotkeys can be configured in the application's `app-config.js` file.

Should be good enough for now. We can think of something better as a future enhancement.

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
igoroctaviano: Done! thx.

platform/ui/src/components/userPreferencesForm/UserPreferences.js
igoroctaviano: Done! thx.

platform/ui/src/components/userPreferencesForm/HotKeysPreferences.js
igoroctaviano: Im using the message directly for now, I have added the key/value to locize as well but I'm not sure about locals update.

dannyrb: If it works locally, and it works in the deploy preview, then I think we're fine.

PR #1171, opened 2019-11-13
feat: expose UiNotifications service

dannyrb:
CC: @igoroctaviano


dannyrb:
Addressed here: https://github.com/OHIF/Viewers/pull/1172

PR #1172, opened 2019-11-13
feat: expose UiNotifications service

igoroctaviano:


extensions/cornerstone/src/init.js
```diff
@@ -4,30 +4,32 @@ import csTools from 'cornerstone-tools';
...
 cornerstone.metaData.addProvider(fallbackMetaDataProvider, -1);
 
-
 /**
  *
  * @param {object} configuration
  * @param {Object|Array} configuration.csToolsConfig
  */
-export default function init(configuration = {}) {
+export default function init(services, configuration = {}) {
```
dannyrb: I'm wondering if we should change-up this signature.

Should we inject the ServicesManager, or all services?
Should we pass in an object instead of named parameters to reduce the chances of introducing breaking changes to extensions?

extensions/cornerstone/src/init.js
```diff
@@ -4,30 +4,32 @@ import csTools from 'cornerstone-tools';
...
+  const { UINotificationService } = services;
```
dannyrb: We should probably do: `UiNotificationService`

With `UINotificationService`, it makes it more ambigous when we should break from pascal casing. We had an issue with this earlier for `SeriesInstanceUid` (should it be `UID`, `UId`, or `Uid`?)

extensions/cornerstone/src/init.js
dannyrb: Once we land on an ideal solution here, we can make sure the other Extension Module's have access to the services with a similar pattern

platform/core/src/services/index.js
```diff
@@ -0,0 +1,4 @@
+import ServicesManager from './ServicesManager.js';
+import createUiNotificationService from './uiNotificationService';
```
dannyrb: I'm not sure if `uiNotificationService` should have the first letter capitalized or not. Factory functions are similar to classes, and I would normally capitalize a class.

platform/ui/src/utils/SnackbarProvider.js
```diff
@@ -1,12 +1,18 @@
...
+} from 'react';
 import SnackbarContainer from '../components/snackbar/SnackbarContainer';
 import SnackbarTypes from '../components/snackbar/SnackbarTypes';
 
 const SnackbarContext = createContext(null);
 
 export const useSnackbarContext = () => useContext(SnackbarContext);
 
-const SnackbarProvider = ({ children }) => {
+const SnackbarProvider = ({ children, service }) => {
```
dannyrb: PropType for `service` and the one method it needs to expose?

extensions/cornerstone/src/init.js
igoroctaviano: I think we should change it to be an object instead of named parameters... About the manager instead of all services, do you think that the extension could leverage the manager itself somehow?

platform/ui/src/utils/SnackbarProvider.js
igoroctaviano: For sure! actually snackbar has 0 proptypes now, I should add it in this branch?

extensions/cornerstone/src/init.js
dannyrb: I'm not sure.

If I receive `{ ServiceName, AnotherService, AndAnotherService, Configuration }`
Versus: `{ ServiceManager, Configuration }`

We can probably add some convenience methods to `ServiceManager` that verify if a service exists, or if it's implementation has been set, etc.

platform/ui/src/utils/SnackbarProvider.js
dannyrb: I think that's fine :+1:

extensions/cornerstone/src/init.js
igoroctaviano: UINotificationService => User Interface Notification Service is more natural to me because im actually contracting each word in my mind and remembering that people usually write UI/UX but we can definitely change this to use a different casing.

dannyrb: @igoroctaviano, I'm generally cool with whatever (with regards to naming), so long as we are consistent 👍

platform/core/src/services/index.js
igoroctaviano: That was my feeling when I changed the service name to be capitalized. I'm used to capitalizing not only because its a class type but services, controllers, classes, managers... so i can distinguish methods from group of methods grouped by a single purpose.. what do you think?

dannyrb: I agree with your thoughts here 👍

Issue #1163, opened 2019-11-12
Why does the URL of each interface point to the same in the runtime?


JiangWeian:
As shown in the picture:
![11](https://user-images.githubusercontent.com/56569509/68648519-86cd8480-055b-11ea-8ad0-fd9e2ae59b42.png)
Each interface URL points to 'qidors'. I thick it's unsafe in the runtime.

dannyrb:
Are you experiencing any issues? (Failing requests in the app)

What does your app configuration file look like? Do you see the same values when using the default configuration file?

JiangWeian:
I have find the reasons. It is caused by incorrect use of dicomweb-client. Line in 'Viewers-master\platform\core\src\studies\services\wado\retrieveMetadataLoaderAsync.js', should be modified as follows:
export default class RetrieveMetadataLoaderAsync extends RetrieveMetadataLoader {
  configLoad() {
    const { server } = this;
	
	var options = {
		url: server.baseUrl,
        headers: DICOMWeb.getAuthorizationHeader(server),
		qidoURLPrefix: server.qidoURLPrefix,
		wadoURLPrefix: server.wadoURLPrefix,
	};
	
	const client = new api.DICOMwebClient(options);
    this.client = client;
  }

the value of 'options' should contain qidoURLPrefix and wadoURLPrefix, even stowURLPrefix. If not, the same values will be assigned to qidoURL, wadoURL and stowURL in dicomweb-client through else statement. The relevant code in dicomweb-client as follow:
      if ('qidoURLPrefix' in options) {
        console.log("use URL prefix for QIDO-RS: ".concat(options.qidoURLPrefix));
        this.qidoURL = this.baseURL + '/' + options.qidoURLPrefix;
      } else {
        this.qidoURL = this.baseURL;
      }

      if ('wadoURLPrefix' in options) {
        console.log("use URL prefix for WADO-RS: ".concat(options.wadoURLPrefix));
        this.wadoURL = this.baseURL + '/' + options.wadoURLPrefix;
      } else {
        this.wadoURL = this.baseURL;
      }

      if ('stowURLPrefix' in options) {
        console.log("use URL prefix for STOW-RS: ".concat(options.stowURLPrefix));
        this.stowURL = this.baseURL + '/' + options.stowURLPrefix;
      } else {
        this.stowURL = this.baseURL;
      }

JiangWeian:
@dannyrb, Do you have read my problem solving code ?

[...]

PR #1175, opened 2019-11-14
For reference ticket #1163

JiangWeian:
Configuration added in default.js of 'qidoURLPrefix' and 'wadoURLPrefix' can control the difference between the two values


platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
```diff
@@ -66,12 +66,16 @@ function makeSeriesAsyncLoader(
...
-    });
-
+	
+	
+	var options = {
+		url: server.baseUrl,
+        headers: DICOMWeb.getAuthorizationHeader(server),
+		qidoURLPrefix: server.qidoURLPrefix,
+		wadoURLPrefix: server.wadoURLPrefix,
+	};
```
dannyrb: Is this in line with the options DICOMwebClient expects?

We should probably preserve the old configuration if `qidoURLPrefix`, `wadoURLPrefix`, and `baseUrl` are not specified. If we don't, this becomes a major/breaking change.

Mixed casing is a bit confusing: `qidoURLPrefix`, but `url` in `baseUrl` is not capitalized?

Note: incorrect indentation and some lint rules not applied.

platform/viewer/public/config/default.js
```diff
@@ -10,6 +10,9 @@ window.config = {
         wadoUriRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/wado',
         qidoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
         wadoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
+		baseUrl: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE',
+	    qidoURLPrefix: 'rs',
+	    wadoURLPrefix: 'rs',
```
dannyrb: These were added, but the `wadoRoot` and `qidoRoot` are still specified?

JiangWeian:
@dannyrb ,OK, I will reconsider your question.

dannyrb:
@JiangWeian, that would do the trick.
That change, and a deprecation warning for the old format, if the new keys aren't specified, when we parse the configuration (`OHIF.log.warn`)

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
```diff
@@ -66,12 +66,16 @@ function makeSeriesAsyncLoader(
...
-      headers: DICOMWeb.getAuthorizationHeader(server),
-    });
-
+	
+	
+	var options = {
+		url: server.baseUrl,
+        headers: DICOMWeb.getAuthorizationHeader(server),
+		qidoURLPrefix: server.qidoURLPrefix,
+		wadoURLPrefix: server.wadoURLPrefix,
```
dannyrb: Does it make more sense to not have a baseUrl, and instead specify the `qidoRoot` and `wadoRoot`? We even already have keys for these. What value does this more granular option setting buy us?

dannyrb:
Some additional comments here: https://github.com/OHIF/Viewers/issues/1163

We specify:

    A base URL
    QIDO Prefix URL
    WADO Prefix URL
    QIDO Root (Base + prefix)
    WADO Root (Base + prefix)

Can we safely say the base is always the same? Or should we ditch "prefix" and "base" and simply always use the Root key values? Then add deprecation warnings for the other keys until the next major bump.

sedghi:

sedghi:

PR #1176, opened 2019-11-14
[UI Services to Extensions]: Add new modal service

igoroctaviano:


platform/ui/src/utils/SnackbarProvider.js
```diff
@@ -25,6 +25,11 @@ const SnackbarProvider = ({ children, service }) => {
   const [count, setCount] = useState(1);
   const [snackbarItems, setSnackbarItems] = useState([]);
 
+  /**
+   * Sets the implementation of a notification service that can be used by extensions.
+   *
+   * @returns void
+   */
```
dannyrb: Oops. We should probably make sure that if `service` is optional that we don't call `service.setServiceImplementation({ hide, show });` if it's not set

platform/ui/src/utils/ModalProvider.js
```diff
@@ -65,6 +82,9 @@ const ModalProvider = ({ children, modal: Modal }) => {
 ModalProvider.propTypes = {
   children: PropTypes.node,
   modal: PropTypes.node,
+  service: PropTypes.shape({
+    setServiceImplementation: PropTypes.func,
```
dannyrb: if service is defined, then this property of it's shape can be required

platform/ui/src/utils/ModalProvider.js
```diff
@@ -65,6 +82,9 @@ const ModalProvider = ({ children, modal: Modal }) => {
 ModalProvider.propTypes = {
   children: PropTypes.node,
   modal: PropTypes.node,
```
dannyrb: Not part of your PR, but this can be required as well?

platform/core/src/services/UIModalService/index.js
```diff
@@ -0,0 +1,89 @@
...
+function createUiModalService() {
+  return uiModalServicePublicApi;
+}
+
+/**
+ * Show a new UI modal;
+ *
+ * @param {Modal} component
+ * @param {ModalProps} props { backdrop, keyboard, show, closeButton, title, customClassName }
+ */
```
dannyrb: `component` - Probably a comment that indicates this should be a React component (as redundant as that may seem)

`ModalProps` - Neat use of typdef

platform/core/src/services/UIModalService/index.js
```diff
@@ -0,0 +1,89 @@
...
+ * @property {string} [backdrop=false] -
+ * @property {string} [keyboard=false] -
+ * @property {number} [show=true] -
+ * @property {string} [closeButton=true] -
+ * @property {string} [title=null] - 'Modal Title'
+ * @property {boolean} [customClassName=null] - '.ModalClass'
+ */
+
+const uiModalServicePublicApi = {
+  name: 'UIModalService',
```
dannyrb: If we're adding a `name`, how do we see it being used?
Is it worth allowing it to be overridden in `createUiModalService`?
Should we add this functionality to our other UI service?

platform/core/src/services/UIModalService/index.js
```diff
@@ -0,0 +1,89 @@
...
+  show,
+  setServiceImplementation,
+};
+
+const uiModalServiceImplementation = {
+  _hide: () => console.warn('hide() NOT IMPLEMENTED'),
+  _show: () => console.warn('show() NOT IMPLEMENTED'),
+};
+
+function createUiModalService() {
```
dannyrb: Casing?

`createUIModalService`?

platform/core/src/services/UIModalService/index.js
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+const uiModalServicePublicApi = {
+  name: 'UIModalService',
+  hide,
+  show,
+  setServiceImplementation,
+};
+
+const uiModalServiceImplementation = {
```
dannyrb: What should the casing be here?

`UiModalServiceImplementation`?
`UIModalServiceImplementation`?

dannyrb: Same goes for above `uiModalServicePublicApi` (and should `API` be capitalized?)

platform/core/src/index.test.js
```diff
@@ -10,7 +10,8 @@ describe('Top level exports', () => {
       'HotkeysManager',
       'ServicesManager',
       //
-      'createUiNotificationService',
+      'createUINotificationService',
+      'createUIModalService',
```
dannyrb: Note: this doesn't match the casing of the same method within `UIModalService/index.js`

extensions/cornerstone/src/index.js
```diff
@@ -24,8 +24,8 @@ export default {
    * @param {object} [configuration={}]
    * @param {object|array} [configuration.csToolsConfig] - Passed directly to `initCornerstoneTools`
    */
-  preRegistration({ serviceManager, configuration = {} }) {
-    init({ serviceManager, configuration });
+  preRegistration({ servicesManager, configuration = {} }) {
```
dannyrb: I feel like the single most difficult thing in programming is "what should I name this thing?"
I'm okay with `servicesManager` -- I constantly struggle with if things like this should be plural or not.

igoroctaviano: Me too! In this case the manager is handling all services, so I think it's ok to use the plural.

platform/core/src/services/UIModalService/index.js
igoroctaviano: The other one is also using the name property, this property was actually used inside the service manager when registering services. We can definitely allow it to be overridden, I'm not sure why... do you have something in mind?

platform/core/src/services/UIModalService/index.js
igoroctaviano: in this case, it's ok to use camelcase because it's just a private object.

platform/core/src/index.test.js
igoroctaviano: Done! thank you.

platform/ui/src/utils/SnackbarProvider.js
igoroctaviano: Done! thank you.

platform/ui/src/utils/ModalProvider.js
igoroctaviano: Done! thank you.

platform/core/src/services/UIModalService/index.js
igoroctaviano: Done! thank you.

PR #1187, opened 2019-11-15
Vtkjs view segmentations

JamesAPetts:
- Filter segmentations out of series list and fetch + load the most recent one when a scan is pulled into a viewport. 
- View segmentations in the 2D MPR view.
- Add metadata to vtkjs viewport overlay.

We can't merge this until we have UI to select a different segmentations.

Currently fetches segmentations in an unlazy way, can easily make it it lazy load once #954 is merged.

TODO: Super draft for live demo purposes. Need to update tests etc.


dannyrb:
@JamesAPetts do you need a Product Team member to knock out the UI on top of this PR?

platform/viewer/src/index.js
```diff
@@ -24,6 +24,7 @@ import ReactDOM from 'react-dom';
  */
 import OHIFVTKExtension from '@ohif/extension-vtk';
 import OHIFDicomHtmlExtension from '@ohif/extension-dicom-html';
+import OHIFDicomSegExtension from '@ohif/extension-dicom-seg';
```
dannyrb: Wondering if we need to split so that we have "Segmentation Loading" as one extension, and "Segmentation Tools" as a separate extension.

We've also discussed use `sopClassHandler` for "filtering the thumbnails" and a new hook, when the visible viewports show new displaySets/studyInstanceUids, that extensions should be given a hook to load new data.

Can we offload this to `@ohif/core` in some way? Is this part of the slicer `IO Module` pattern we've seen?

platform/viewer/src/components/ViewportGrid/ViewportGrid.js
```diff
@@ -30,6 +33,12 @@ const ViewportGrid = function(props) {
     return null;
   }
 
+  useEffect(() => {
+    viewportData.forEach(displaySet => {
+      loadAndCacheDerivedDisplaySets(displaySet, studies);
```
dannyrb: I think this is the "IO" to load segs as we switch viewports around

platform/viewer/package.json
```diff
@@ -50,6 +50,7 @@
     "@ohif/extension-dicom-html": "^1.0.1",
     "@ohif/extension-dicom-microscopy": "^0.50.6",
     "@ohif/extension-dicom-pdf": "^1.0.0",
+    "@ohif/extension-dicom-seg": "^1.0.0",
```
dannyrb: Can probably use pre-v1

extensions/vtk/src/OHIFVTKViewport.js
```diff
@@ -6,12 +6,14 @@ import LoadingIndicator from './LoadingIndicator.js';
...
 import PropTypes from 'prop-types';
 import cornerstone from 'cornerstone-core';
-import handleSegmentationStorage from './handleSegmentationStorage.js';
+import cornerstoneTools from 'cornerstone-tools';
 import vtkDataArray from 'vtk.js/Sources/Common/Core/DataArray';
 import vtkImageData from 'vtk.js/Sources/Common/DataModel/ImageData';
 import vtkVolume from 'vtk.js/Sources/Rendering/Core/Volume';
 import vtkVolumeMapper from 'vtk.js/Sources/Rendering/Core/VolumeMapper';
 
+const segmentationModule = cornerstoneTools.getModule('segmentation');
```
dannyrb: Need to be careful we're not accidentally creating a hard dependency on `dicom-seg` extension

extensions/cornerstone/src/OHIFCornerstoneViewport.js
```diff
@@ -196,13 +168,10 @@ class OHIFCornerstoneViewport extends Component {
       );
     }
 
-    const sopClassUid = sopClassUids && sopClassUids[0];
-
     this.getViewportData(
       studies,
       studyInstanceUid,
       displaySetInstanceUid,
-      sopClassUid,
```
dannyrb: Where is this defined? Did the API change?

JamesAPetts: We originally had handlers for dragging SEG into cornerstone viewports. Since that approach is being changed, it is now unused so I removed it. Its only API internal to the component.

platform/viewer/src/index.js
JamesAPetts: Splitting them up may not be a bad idea. Someone may want to be able to view and not edit.

> that extensions should be given a hook to load new data.

Yeah it still feels a bit wonky to do it this way, but that seemed like the easiest way to keep the loading encapsulated within the plugin to some degree.

We already have `loadAndCachedDerivedDisplaySets` in `@ohif/core`, what else do you think should be pushed down there? We don't have much more in the seg extension other than the loader itself.

platform/viewer/src/components/ViewportGrid/ViewportGrid.js
JamesAPetts: It is the IO to switch any derived displayset on the study. Note that the only dependency is `@ohif/core`, and if a displaySet has been created by an extension's sopClassHandler, then the `load` callback from the extension will be called from the displaySet.

platform/viewer/package.json
JamesAPetts: Sure, I'm indifferent.

extensions/vtk/src/OHIFVTKViewport.js
JamesAPetts: We aren't here, we have a hard dependency on cornnerstone-tools, which is just required to interoperate between `cornerstonejs` and `vtkjs`.

If a `cornerstoneTools` segmentation exists on the displaySet it will be loaded. This could be true whether or not you have a DICOM Seg handler yourself. The segmentation might be created first hand.

Splitting the segmentation handler and the segmentation tools into different extensions solves this problem as you have no possible input to the `cornerstoneTools` segmentation module. If you remove both the vtk plugin will gracefully sweep over the vtk labelmap generation when it simply finds no segmentations in the module.

fedorov:
How do I preview this? I get black screen when I try to open the preview using the netlify link from the above (I do get the study list, but cannot open any of the image series).

dannyrb:
@fedorov, not sure what happened to the previous preview. This one should work. Let me know if you run into issues: https://deploy-preview-1187--ohif.netlify.com/pwa/viewer/1.3.6.1.4.1.14519.5.2.1.2744.7002.373729467545468642229382466905

fedorov:
@dannyrb it still does not work for me. The link you provided is to a study that does not have any SEGs. If I open this one that has SEGs https://deploy-preview-1187--ohif.netlify.com/pwa/viewer/1.3.6.1.4.1.14519.5.2.1.3671.7001.133687106572018334063091507027, I see OHIF blinking for a moment, but then I get black screen. I tried in incognito, and it doesn't make any difference. Here's the console in case it helps.

image

dannyrb:
Screenshot_20200114-170441_Chrome

I'm on Mobile, but I see these labelmaps and segs for the study I linked

pieper:
What platform is that @dannyrb ?  I get the same blank screen as Andrey on chrome and safari on mac.  On iphone I get a message that webgl2 is needed.

fedorov:
I am afraid I cannot explain what is going on or fix it, but I made a screncast demonstrating the problem in my side. This was shot in Chrome 79.0.3945.117 "incognito" mode on macOS 10.14.6: https://www.screencast.com/t/quQDMhmf8x.

JamesAPetts:
@fedorov , it appears that particular study you choose has issues. I'm not sure why this is at the moment, but it also has loading issues on master.

The scan Danny referenced does have SEGs, its just the default SEG loaded (most recent) is rather featureless. Open the right side panel and choose `tumor segmentation - User2`, alternatively here is a nice working brain scan:

https://deploy-preview-1187--ohif.netlify.com/pwa/viewer/1.3.12.2.1107.5.2.32.35162.30000015050317233592200000046

There is definitely some odd data on the PACs, but this is good and should be used for testing fallbacks + implementing warnings.

dannyrb:
@JamesAPetts, okay to close this?

pieper:
Does this functionality exist anywhere else?  Right now I think this netlify link is the only live example we have of the vtkjs seg support (or it also somewhere else?).

pieper:
Okay, I see now - yes, much better thanks 👍

dannyrb:
Closing, but keeping the branch available until the other PR is merged. The Netlify link should stay active. You can always open a new PR with this branch if it dies on you.

Issue #1193, opened 2019-11-15
UserPreferences issues with translation and hotkeys


mirnasilva:
## Bug Report

A couple of bugs were found and they will be described below:

### Issue 1: Translation not cancelled

On UserPreferences modal, if user selects another language (i.e. Spanish) and clicks on `Cancel`, the toolbar and some of the other page elements will be translated into the selected language, even though the user cancelled the language selection. This is happening on **StudyList page** and on **Viewer page**.

**Steps to Reproduce:**

1. Open Preferences modal
2. Go to General tab
3. Select `Spanish` language
4. Verify that the header and toolbar gets translated behind the modal
5. Click on `Cancel` button
6. Verify that the header and toolbar remain in Spanish, even though the user cancelled the language selection.

### Issue 2: Options overlay not being translated

1. On StudyList page, if user selects another language on the Preferences modal, the Options overlay is not being translated. After selecting Spanish and navigating to Viewer page, the Options overlay gets partially translated to Spanish and won't change back to English, even if user selects English on Preferences modal. 

2. The option `Preferences` is being displayed only in English and is not being translated.

**Study List page**
<img width="506" alt="Screen Shot 2019-11-15 at 18 30 35" src="https://user-images.githubusercontent.com/13245529/68977519-31051080-07d7-11ea-939a-96a4cc0fe953.png">

**Viewer page**
<img width="431" alt="Screen Shot 2019-11-15 at 18 31 28" src="https://user-images.githubusercontent.com/13245529/68977516-2e0a2000-07d7-11ea-82ff-b5e2df35b468.png">

### Issue 3: "Reset to Defaults" button closes the modal

If user clicks on "Reset to Defaults" button, the Preferences modal is closed and the user has to open it again to make sure it restored default values. 

**Acceptance Criteria:** 

1. After user clicks on "Reset to Defaults" button, the values should be restored and the modal should remain open.
a) If user clicks on "Save" button, then the default values are saved and the modal is closed.
b) If user clicks on "Cancel" button, then the default values are **not saved**, the previous values are kept and the modal is closed.

### Issue 4: No Zoom In limit for hotkeys

When user is zooming in through the mouse he/she can reach a limit of zoom. When user is zooming in by using hotkeys, there is no zooming limit.

**Expected result**
The hotkey zooming in should respect the same limit as the mouse zooming.

[...]

PR #1195, opened 2019-11-17
Test Cases for User Preferences

mirnasilva:
### Test Cases for User Preferences modal

Some of the tests are blocked by bugs from #1193 

Added `data-cy` selectors for buttons inside the Preferences modal and for hotkeys shortcut input fields. This really improved the readability of the tests, since the shortcut inputs selector were getting too messy.

Test cases were divided into Study List Page scenarios and Study Viewer Page scenarios:

**Study List Page**
- opens User Preferences modal and verify the displayed information
- checks translation by selecting Spanish language
- checks if user can cancel the language selection and application will be in English
- checks if Preferences set in Study List Page will be consistent on Viewer Page

**Study Viewer Page**
- opens User Preferences modal and verify the displayed information
- checks translation by selecting Spanish language
- checks if user can cancel the language selection and application will be in English
- checks new hotkeys for "Rotate Right" and "Rotate Left"
- checks new hotkeys for "Next" and "Previous" Image on Viewport
- checks error message when duplicated hotkeys are inserted
- checks error message when invalid hotkey is inserted
- checks error message when only modifier keys are inserted
- checks if user can cancel changes made on User Preferences Hotkeys tab
- checks if user can reset to default values on User Preferences Hotkeys tab


dannyrb:
Potential test flakiness here:

image

Re-running workflow

platform/ui/src/components/userPreferencesForm/HotKeysPreferences.js
```diff
@@ -201,11 +201,12 @@ export class HotKeysPreferences extends Component {
               value={keys[0]}
               vali="true"
               className="form-control hotkey text-center"
+              data-cy={label}
               onKeyDown={event => this.onInputKeyDown(event, commandName)}
               onKeyUp={event => this.onChange(event, commandName)}
             />
             <span className="wrapperText" />
-            <span className="errorMessage">
+            <span className="errorMessage" data-cy={`error-${label}`}>
```
dannyrb: This is tricky. I don't love `dynamic` data-cy attributes, but I'm trying to think about how else I would do this. I'll circle back after I see how you're leveraging these.

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
```diff
@@ -0,0 +1,571 @@
...
+  context('Study List Page', function() {
+    before(() => {
+      cy.visit('/');
+    });
+
+    beforeEach(() => {
+      // Open User Preferences modal
+      cy.openPreferences().then(() => {
+        cy.initPreferencesModalAliases();
+      });
```
dannyrb: You can probably forego using a `then` here:

```js
cy.openPreferences();
cy.initPreferencesModalAliases();
```

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
```diff
@@ -0,0 +1,571 @@
...
+    it('opens User Preferences modal and verify the displayed information', function() {
+      cy.get('@preferencesModal').should('be.visible');
+      cy.get('.HotKeysPreferences').should('be.visible');
+      cy.get('@hotkeysLink').should('have.class', 'active');
+      cy.get('@generalLink').should('be.visible');
+      cy.get('@restoreBtn')
+        .scrollIntoView()
+        .should('be.visible');
+      cy.get('@cancelBtn').should('be.visible');
+      cy.get('@saveBtn').should('be.visible');
```
dannyrb: For the buttons, maybe instead of `should be visible`, we can make sure they contain the text we expect?

What is `@generalLink`? What is `@hotkeysLink`?
What is `.HotkeytsPreferences`?

The text says "opens user preferences modal", but I think that happens in the `beforeEach`
We could close it after initing aliases, or update the text here?

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
```diff
@@ -0,0 +1,571 @@
...
+
+      // Visual comparison
+      cy.screenshot(
+        'User Preferences Modal - Hotkeys tab initial state in Study Viewer page'
+      );
+      cy.percyCanvasSnapshot(
+        'User Preferences Modal - Hotkeys tab initial state in Study Viewer page'
+      );
+      cy.get('.close').click();
```
dannyrb: We could add a `data-cy="modal-close"`?

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
```diff
@@ -0,0 +1,571 @@
...
+    });
+
+    it('checks translation by selecting Spanish language', function() {
+      cy.get('@generalLink')
```
dannyrb: Ah. Okay. Instead of `@generalLink` can we do something more like `@userPreferencesGeneralTab` or something similarly descriptive?

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
```diff
@@ -0,0 +1,571 @@
...
+    beforeEach(() => {
+      cy.initCommonElementsAliases();
+      cy.resetViewport();
+      cy.wait(200);
+      cy.resetHotkeysPreferences();
+      // Open User Preferences modal
+      cy.openPreferences();
+    });
+
+    it('opens User Preferences modal and verify the displayed information', function() {
```
dannyrb: This test description is duplicated on line 14

platform/ui/src/components/userPreferencesForm/HotKeysPreferences.js
dannyrb: Okay. Instead of dynamic `data-cy`, can we try...

[.closest()](https://docs.cypress.io/api/commands/closest.html#Syntax)

So, something like:

```js
// Within scopes all `.get` and `.contains` to within the matched elements
// dom instead of checking from document
cy.get('.modal-body').within(() => {

  cy.contains('Next Image Viewport') // label we're looking for
     .closest('input') // closest input to that label
     .type('c');

  cy.contains('Previous Image')
     // etc.
})
```

This should feel similar to the code you're using below to find input elements. You're still using the label, but now we're not generating dynamic cypress attributes that match text we can already seek on.

If the `data-cy` attributes made things less brittle, I would be more keen on keeping them. They're nice for "container" elements, decoupling from `class name` dependencies, etc.

`.row > .col-sm-6 > div:first-child` is much harrier than `data-cy="page-footer"`

platform/viewer/cypress/support/commands.js
```diff
@@ -410,3 +409,33 @@ function canvasToImage(selectorOrEl) {
...
+    .scrollIntoView()
+    .click()
+    .then(() => {
+      cy.get('[data-cy="about-item-menu"]')
+        .last()
+        .click();
+    });
+});
+
+Cypress.Commands.add('resetHotkeysPreferences', () => {
```
dannyrb: This opens user preferences but does not close them. Can we rename it to `resetHotkeyBindings` and make sure it closes the `UserPreferences` modal?

dannyrb: Or if you prefer "preferences": `resetUserHoktkeyPreferences`

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
```diff
@@ -0,0 +1,571 @@
...
+      // Check if new hotkey is working
+      cy.get('body').type('{shift}Q', { release: false });
+      cy.get('@viewportInfoMidTop').should('contains.text', 'R');
+    });
+  });
+
+  context('Study Viewer Page', function() {
+    before(() => {
+      cy.openStudy('MISTER^MR');
+      cy.waitDicomImage();
```
dannyrb: I'm not 100% sure why this is blowing up on you.
This is wrapped in the same `decribe` as above. When running locally, do these pass?
Does it work if you close the previous suite's `describe` block and wrap this in a new one?

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
```diff
@@ -0,0 +1,571 @@
...
+      // Go go hotkeys tab
+      cy.get('@hotkeysLink').click();
+
+      // Set invalid hotkey for 'Rotate Right' function
+      cy.get('[data-cy="Rotate Right"]')
+        .click()
+        .type('{ctrl}Z', { force: true });
+
+      // Check error message
+      cy.get('[data-cy="error-Rotate Right"]').should(
```
dannyrb: Similarly, I think you can use `closest()` here to avoid a new `data-cy`

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
mirnasilva: Yes and no. On line 14 I'm checking if everything is displayed as expected when we open User Preferences from the StudyList page. Here we are opening it from Study Viewer page. SO we are checking the same thing but from different origins.

platform/viewer/cypress/support/commands.js
mirnasilva: On the current flow, once user clicks on "Reset to Defaults" button, the modal is automatically closed.

platform/ui/src/components/userPreferencesForm/HotKeysPreferences.js
mirnasilva: Thanks for this suggestion! I just had to do a small change, because `.closest()` only searches in its ancestors and the `input` is a label's sibling. Instead of that, I used `.parent().find()` and it did the trick.

```
cy.get('.modal-body').within(() => {
      cy.contains(function_label) // label we're looking for
        .parent()
        .find('input') // closest input to that label
        .type(shortcut, { force: true }); // Set new shortcut for that function
    });
```

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
mirnasilva: I ended up removing the `cy.waitDicomImage();`. Since we are not interacting with the viewport right away, the application has plenty of time to load the image.
But I'm still not sure why it passes locally and not on script-tag server.

platform/viewer/cypress/integration/common/OHIFCornerstoneHotkeys.spec.js
```diff
@@ -85,6 +85,7 @@ describe('OHIF Cornerstone Hotkeys', () => {
     cy.get('@layoutBtn').click();
     //Select 3 viewports
     cy.get('tbody > :nth-child(1) > :nth-child(3)').click();
+    cy.wait(1000);
```
dannyrb: Why the wait? And this feels brittle (above line). We should probably at least add a `data-cy` selector for our viewports.

mirnasilva: Replaced the `wait` for a more fancier condition that checks if message 'Loading...' is being displayed on the viewports.

`Line 87` is selecting the layout style on that table overlay. This one is selecting [1,3]. I can replace this eventually with `setLayout()`, but for now let's just get this code merged, since this is not directly related to the new Preferences Modal tests.

PR #1202, opened 2019-11-18
feat: New dialog service

igoroctaviano:


igoroctaviano:
> _No description provided._

@dannyrb This PR includes two unrelated changes: 
1. Modal refactoring to use react-modal instead and remove dependency upon current global modal styles.
2. (Small bugfix) Download viewport preview image getting with weird aspect when resizing.

dannyrb:
CI is failing for PWA and Script-tag

It looks like we had one test that used the `.modal-content` class that has been removed:

image

Can you add a `data-cy="modal"` or "modal-body" to the new modal, and update the test accordingly so it will pass?

dannyrb:
It's really nice to see `draggable` in play! There are a few items we need to account for now that it's in use:

ESEDZHnmxM

- Show `drag` icon when modal dialog is draggable
- Modal should not be `draggable` on buttons or range slider
- Dragging the modal far right or far down creates scrollbars and extends screen
  - Modal should be bound to viewport height/width
  - and/or we should hide vertical/horizontal overflow for the `viewer` routes

There are a few other minor issues that aren't blocking, but if we don't account for them here, we should create minor bugs:

- UI state for `CINE` button should remain `active` while dialog is open (helps indicate it can be clicked again to close)
- When CINE is opened, it should open to most recent dragged location (unless that location is off-screen)

platform/core/src/services/UIDialogService/index.js
```diff
@@ -0,0 +1,85 @@
...
+  return uiDialogServicePublicAPI;
+}
+
+/**
+ * Show a new UI dialog;
+ *
+ * @param {DialogContent} component React component
+ * @param {DialogProps} props { id, content, ...props }
+ */
+function create({ id, content, ...props }) {
```
dannyrb: in `UIModalService`, we defined the props here. I think being explicit makes for a better contract, but I understand the want of something that's more flexible. I think we want extension authors to program to these contracts, and not toward any one implementation. By not defining props, we're saying, look to the underlying implementation to figure out what should be passed.

platform/core/src/services/UIModalService/index.js
```diff
@@ -38,19 +35,16 @@ function createUIModalService() {
...
-    footer: null,
-    backdrop: false,
-    keyboard: false,
-    show: true,
+    shouldCloseOnEsc: false,
+    isOpen: true,
     closeButton: true,
     title: null,
-    customClassName: null,
+    customClassName: '',
```
dannyrb: Why empty string? When you use `classnames()`, you can toss `null` in and everything works alright. Easier to check than empty string as well

platform/core/src/services/UIModalService/index.js
```diff
@@ -38,19 +35,16 @@ function createUIModalService() {
...
+ * @param {ModalProps} props { shouldCloseOnEsc, isOpen, closeButton, title, customClassName }
  */
 function show(
   component,
   props = {
-    header: null,
-    footer: null,
-    backdrop: false,
-    keyboard: false,
-    show: true,
```
dannyrb: 👀 Goodbye? 👋 
Not disagreeing, but curious to hear reasoning.

dannyrb: My first guess is simply that we're not using these (yet) (KISS) (YAGNI)

platform/ui/package.json
```diff
@@ -52,7 +52,9 @@
     "react-dnd": "9.4.0",
     "react-dnd-html5-backend": "^9.4.0",
     "react-dnd-touch-backend": "^9.4.0",
+    "react-draggable": "^4.1.0",
     "react-i18next": "^10.11.0",
+    "react-modal": "^3.11.1",
```
dannyrb: Can we nuke `"react-bootstrap-modal"`?

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.styl
```diff
@@ -3,13 +3,8 @@
...
 
 .ViewportDownloadForm
-  color: var(--text-secondary-color);
-  filter: drop-shadow(0 0 3px var(--ui-gray-darkest));
-  border: none;
-  border-radius: 8px;
-  width: inherit;
-  padding: 15px;
-  background: transparent;
+  display: flex;
```
dannyrb: Flexbox <3

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.styl
```diff
@@ -90,7 +86,7 @@
 
   .actions
     display: flex;
-    height: 60px;
+    height: auto;
```
dannyrb: Do we even need `auto`?

platform/ui/src/components/measurementTable/MeasurementTableItem.js
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { withTranslation } from '../../utils/LanguageProvider';
+import { withTranslation } from '../../utils/providers';
```
dannyrb: Is it worth moving all `providers` to a root folder named `contextProviders` instead of jamming them into `utils`?

platform/ui/src/components/ohifModal/OHIFModal.js
```diff
@@ -1,64 +1,65 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import ReactBootstrapModal from 'react-bootstrap-modal';
+import Modal from 'react-modal';
```
dannyrb: Much lighter dependency 👍

platform/ui/src/components/ohifModal/OHIFModal.styl
```diff
@@ -0,0 +1,56 @@
...
+  border: 0
+  color: var(--text-primary-color)
+  position: relative
+  -webkit-box-shadow: 0 3px 9px rgba(0,0,0,.5);
+  box-shadow: 0 3px 9px rgba(0,0,0,.5);
+  background-clip: padding-box;
+  outline: 0;
+
+  @media (min-width: 768px)
+    width: 600px;
```
dannyrb: May be worth setting a percentage width here and a `min-width`

platform/ui/src/components/userPreferencesForm/HotKeysPreferences.styl
```diff
@@ -1,7 +1,5 @@
 .HotKeysPreferences
   display: flex;
-  margin-right: -15px;
-  margin-left: -15px;
```
dannyrb: The negative margin trick is how the bootstrap "row" class works. Bye bye negative margin 👋

platform/ui/src/components/measurementTable/MeasurementTableItem.js
igoroctaviano: Beyond simplifying all imports I think a new folder just keep things more organized. Do you think its not worth it?

platform/ui/package.json
igoroctaviano: Not yet I think... It's being used in DicomFileUploaderModal and DicomStorePickerModal, these can be refactored in another task and then we can nuke it.

platform/ui/src/components/measurementTable/MeasurementTableItem.js
dannyrb: Oh, I was agreeing that they should have their own place to live. I'm just not sure if under `utils` makes sense. I wouldn't think of a ContextProvider as a "utility".

I could see them either:

- Living next to the component they provide the context for
- or living under `src/contextProviders`, as a quick way to see all of the different context providers for our UI library

platform/core/src/services/UIModalService/index.js
igoroctaviano: Yes, we're not using them yet but I think header a footer being passed as props is just weird, header and footer are content's responsibility now.

platform/ui/package.json
dannyrb: Created: https://github.com/OHIF/Viewers/issues/1204

platform/ui/src/components/measurementTable/MeasurementTableItem.js
igoroctaviano: > Oh, I was agreeing that they should have their own place to live. I'm just not sure if under `utils` makes sense. I wouldn't think of a ContextProvider as a "utility".
> 
> I could see them either:
> 
> * Living next to the component they provide the context for
> * or living under `src/contextProviders`, as a quick way to see all of the different context providers for our UI library

oh! yeah, I was wondering why they were in utils as well. Ill move them fore sure! thx.

platform/viewer/src/App.js
```diff
@@ -10,52 +21,56 @@ import {
...
-import { SnackbarProvider, ModalProvider, OHIFModal } from '@ohif/ui';
 
-// Contexts
+/** Contexts */
 import WhiteLabellingContext from './context/WhiteLabellingContext';
 import UserManagerContext from './context/UserManagerContext';
 import AppContext from './context/AppContext';
 
-// ~~~~ APP SETUP
+/** ~~~~~~~~~~~~~ Application Setup */
```
dannyrb: Little dissapointed the commit for this didn't say "added more squiggles"

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -279,21 +267,38 @@ function _getVisibleToolbarButtons() {
...
+        .getBoundingClientRect();
+      const newDialogId = dialog.create({
+        content: ConnectedCineDialog,
+        defaultPosition: {
+          x: x + spacing || 0,
+          y: y + spacing || 0,
+        },
+      });
+      this.setState({ dialogId: newDialogId });
+    }
```
dannyrb: This is great. Here is something we need to think about:

How can we make extensions powerful enough that this logic can live in them? Having "Built in" viewer behavior lets us offload some concerns to `viewer`, but it also makes us less flexible.

Ideally, we end up at something like:

- ToolbarButton is of type `Toggle`
  - When clicked, is `enabled`, and extension's logic is called w/ `enabled: true`
  - When clicked again, enabled is `false`, and extension's logic is called
- The toolbar can use the now exposed `Dialog` service to hide/show dialog
- **As the user interacts, the extension is able to update the ViewportComponent based on that interaction**

I think that last part is the only thing missing, and what we need a good solution for (outside the scope of this PR, but an upcoming issue for sure). If we wanted to update... cineIsPlaying, stack index, slice index, etc. What would that look like for an extension?

I'll create a `discussion` issue for this.

dannyrb: Today, we only have two viewport extensions that can leverage CINE like behavior. I could see that growing AND shrinking. I am a little torn on if this **specific** use case makes more sense to configure in `@ohif/viewer` or `@ohif/extension-*`. Our current implementation updates viewport props, and it's up to registered viewports to leverage those prop updates to show CINE behavior.

The changes outlined above would make it easier for extensions to:

1. Show custom dialog/modal UI
2. Handle how input to that UI influences behavior _like_ viewport CINE behavior

But this could also extend to modifying any state the `@ohif/viewer` has traditionally managed so far. One thing we have discussed is creating a `MeasurementsService`.

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -279,21 +267,38 @@ function _getVisibleToolbarButtons() {
...
   }
 
   if (behavior === 'DOWNLOAD_SCREEN_SHOT') {
-    this.props.modal.show(ConnectedViewportDownloadForm, {
-      title: this.props.t('Download High Quality Image'),
-      customClassName: 'ViewportDownloadForm',
+    modal.show(ConnectedViewportDownloadForm, {
```
dannyrb: Similarly, I think we can move this to the `@ohif/extension-cornerstone` and ditch the built in here (outside scope of this PR). This one's a bit closer/easier.

We use `state.viewports` to grab `viewportSpecificData` and `activeViewportIndex`
We use those to grab `dom` off the `viewportSpecificData`'s `activeViewport`

DOM is just a reference to the active viewport's `enabledElement`
`dom` is actually set/added by the `@ohif/extension-cornerstone`

The only bit of data we need exposed to refactor the `ConnectedViewportDownloadForm` _should_ be the `activeViewportIndex` which we already have exposed for extensions (albeit, that method could be cleaned up).

platform/viewer/src/connectedComponents/ToolbarRow.js
dannyrb: Follow-up issue created here: https://github.com/OHIF/Viewers/issues/1205

platform/viewer/src/connectedComponents/ToolbarRow.js
dannyrb: Backlogged here: https://github.com/OHIF/Viewers/issues/1206

platform/ui/src/utils/providers/DialogProvider.js
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @returns id
+   */
+  const create = useCallback(
+    ({ id, content, ...props }) => {
+      let dialogId = id;
+      if (!dialogId) {
+        dialogId = Math.random()
+          .toString(36)
+          .substr(2, 5);
```
dannyrb: Oh, interesting. Allow them to provide their own ID. I like it.
For random, do we have a GUID generator? Could use incrementing count, but would need to note that custom IDs should not be numbers as string to prevent conflicts.

platform/ui/src/utils/providers/DialogProvider.js
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }, [create, dismiss, service]);
+
+  /**
+   * Creates a dialog and return its id.
+   *
+   * @returns id
+   */
+  const create = useCallback(
+    ({ id, content, ...props }) => {
```
dannyrb: Probably should be more explicit of props. See earlier comment.

platform/ui/src/utils/providers/DialogProvider.js
```diff
@@ -0,0 +1,164 @@
...
+    setDialogs(dialogs => dialogs.filter(dialog => dialog.id !== id));
+  }, []);
+
+  /**
+   * Dismisses all dialogs.
+   *
+   * @returns void
+   */
+  const dismissAll = () => {
+    setCount(1);
```
dannyrb: In `dismiss`, if dialogs length is `0`, could also reset this.

platform/ui/src/utils/providers/DialogProvider.js
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Expose dialog methods to window for debug purposes.
+   */
+  window.dialog = {
+    create,
+    dismiss,
+    dismissAll,
+    dialogs,
+  };
```
dannyrb: Can remove this for PR?

platform/ui/src/utils/providers/DialogProvider.js
```diff
@@ -0,0 +1,164 @@
...
+            position={position}
+            defaultPosition={defaultPosition}
+            onStop={onStop}
+            onDrag={event => {
+              _reorder(id);
+              onDrag(event);
+            }}
+          >
+            <div
+              style={{ zIndex: '999', position: 'absolute' }}
```
dannyrb: As we add more and more things with `z-index`, we may want to formalize which elements go on which layers

dannyrb: (not blocking)

platform/ui/src/components/ohifModal/OHIFModal.styl
igoroctaviano: I'm using the same media from modal.styl. Honestly, I find percentages to be harder to define, do you have something in mind?

platform/ui/src/components/measurementTable/MeasurementTableItem.js
igoroctaviano: Done, thx.

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.styl
igoroctaviano: Done, thx.

platform/core/src/services/UIModalService/index.js
igoroctaviano: Done, thx.

platform/core/src/services/UIDialogService/index.js
igoroctaviano: Done, thx.

platform/ui/src/components/ohifModal/OHIFModal.styl
dannyrb: I only suggested percentage so that this breakpoint would continue to work as we approach the lower mobile threshold of ~320px. You could do that with `width: 78%; min-width: 300px`

While we're not there yet, I could see a Modal becoming an "Action Sheet" or something similar at mobile resolutions; this would just buy us a bit of grace until we get there / while we transition.

PR #1211, opened 2019-11-20
feat: Add new annotate tool using new dialog service

igoroctaviano:


platform/viewer/src/components/Header/Header.js
```diff
@@ -36,18 +36,14 @@ class Header extends Component {
   }
 
   loadOptions() {
-    const {
-      t,
-      user,
-      userManager,
-      modal: { show },
-    } = this.props;
+    const { t, user, userManager, modal } = this.props;
```
dannyrb: Can we use `withTranslations`?

igoroctaviano: Which one? currently its using withTranslation from react-i18next

dannyrb: Pointing out that we may not need to pass `t` as a prop

platform/core/src/services/UIDialogService/index.js
```diff
@@ -51,29 +45,29 @@ function createUIDialogService() {
...
+  id = null,
+  content = null,
+  contentProps = null,
+  onStart = null,
+  onDrag = null,
+  onStop = null,
+  isDraggable = true,
+  showOverlay = false,
+  defaultPosition = null,
+  position = null,
```
dannyrb: I'm not sure there's much value in defaulting these to `null`. Not a blocker, just confused.

platform/core/src/services/UIModalService/index.js
```diff
@@ -34,20 +30,26 @@ function createUIModalService() {
...
-) {
-  return uiModalServiceImplementation._show(component, props);
+function show({
+  content = null,
+  contentProps = null,
+  shouldCloseOnEsc = false,
+  isOpen = true,
+  closeButton = true,
+  title = null,
+  customClassName = null,
```
dannyrb: Same here. I don't understand why `undefined` is a bad value. This implies that we're using `null` to indicate something a bit more intentional?

platform/ui/src/components/simpleDialog/SimpleDialog.js
```diff
@@ -21,6 +22,31 @@ class SimpleDialog extends Component {
...
+          <TextInput
+            type="text"
+            value={value}
+            onChange={event => setValue(event.target.value)}
+            label={label}
+          />
+        </SimpleDialog>
+      </div>
+    );
+  };
```
dannyrb: If we're updating the `SimpleDialog`, we need to do a quick manual regression test on the relabeling workflow.

platform/ui/src/components/simpleDialog/SimpleDialog.styl
```diff
@@ -1,6 +1,10 @@
 @import './../../design/styles/common/button.styl'
 @import './../../design/styles/common/global.styl'
 
+.InputDialog
+  .simpleDialog
+    position: relative
+
```
dannyrb: We nest a "dialog" within a "dialog"?

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -114,62 +125,84 @@ const DialogProvider = ({ children, service }) => {
...
+                  const e = event || window.event;
+                  const target = e.target || e.srcElement;
+                  const BLACKLIST = [
+                    'SVG',
+                    'BUTTON',
+                    'PATH',
+                    'INPUT',
+                    'SPAN',
+                    'LABEL',
+                  ];
```
dannyrb: Oh. This is an interesting approach.

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -114,62 +125,84 @@ const DialogProvider = ({ children, service }) => {
...
-                )}
-                style={{ zIndex: '999', position: 'absolute' }}
-                onClick={() => _reorder(id)}
+              <Draggable
+                disabled={!isDraggable}
+                position={position}
+                defaultPosition={lastDialogPosition || defaultPosition}
+                bounds="parent"
+                onStart={event => {
+                  const e = event || window.event;
```
dannyrb: When would this be `window.event`?

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -114,62 +125,84 @@ const DialogProvider = ({ children, service }) => {
     });
   };
 
+  const validCallback = callback => callback && typeof callback === 'function';
+
   return (
-    <DialogContext.Provider value={{ create, dismiss, dismissAll, dialogs }}>
+    <DialogContext.Provider value={{ create, dismiss, dismissAll, noDialogs }}>
```
dannyrb: `noDialogs` --> `hasNoDialogs`?

dannyrb: Or good ol' `isEmpty`

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -114,62 +125,84 @@ const DialogProvider = ({ children, service }) => {
...
       <div className="DraggableArea">
         {dialogs.map(dialog => {
           const {
-            id,
-            content: Dialog,
-            position /* Position of the dialog. {{x: 0, y: 0}} */,
-            defaultPosition,
+            id = null,
```
dannyrb: Is `id` even allowed to be empty? Why default to `null`?

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -114,62 +125,84 @@ const DialogProvider = ({ children, service }) => {
...
+            content: DialogContent = null,
+            contentProps = null,
+            position = null,
+            defaultPosition = null,
             isDraggable = true,
-            onStart = () => {},
-            onStop = () => {},
-            onDrag = () => {},
+            showOverlay = false,
```
dannyrb: If each dialog can `showOverlay`, what happens if multiple do?

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -114,62 +125,84 @@ const DialogProvider = ({ children, service }) => {
...
-                setIsDragging(false);
-                return;
-              }}
-              onDrag={event => {
-                setIsDragging(true);
-                _reorder(id);
-                _updateLastDialogPosition(id);
-                onDrag(event);
-              }}
+              className={classNames('Overlay', showOverlay && 'active')}
```
dannyrb: You may want "one" of these that is shown if any dialog is attempting to `showOverlay`? Unless we want to support a scenario where you have one dialog, that opens another dialog but hides the previous beneath the overlay to prevent interaction?

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -114,62 +125,84 @@ const DialogProvider = ({ children, service }) => {
...
-              </div>
-            </Draggable>
+                <div
+                  id={`draggableItem-${id}`}
+                  className={classNames(
+                    'DraggableItem',
+                    isDragging && 'dragging'
+                  )}
+                  style={{ zIndex: '999', position: 'absolute' }}
+                  onClick={() => _reorder(id)}
```
dannyrb: Maybe name `_bringToFront`?

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: This overlay is meant to be used if the user has to answer one dialog at a time, like when the user adds a new arrow annotation, we cannot allow the user to just keep creating dialogs before closes the last.

dannyrb: If you think we'll need to support more than one overlay, then keep it but document it? If you think we'll only ever need one overlay, bump it up a layer?

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: Nothing, the user will have to answer one at a time according to the call order.

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: Its better! updated, thanks.

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: Its not, updated, thanks.

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: Yeah, i think isEmpty is simpler, updated, thanks.

platform/ui/src/components/simpleDialog/SimpleDialog.styl
igoroctaviano: I have seen this approach in some React libs where they have extended versions of the component as static props of the component class. It's totally not necessary though.

platform/core/src/services/UIModalService/index.js
igoroctaviano: There's nothing is waiting for null so it's fine to take it out, thanks.

platform/core/src/services/UIDialogService/index.js
igoroctaviano: Habits when using conditional rendering / error boundaries in React render method since undefined will be an object instead of a null value which is renderable. But! it's not this case though :D so ill remove it, thanks.

dannyrb:
`.DraggableArea` has a width of `100vw` on FireFox. FireFox includes the scroll-bar in that calculation. This causes a horizontal scrollbar to appear. Using `100%` resolves the issue.

platform/ui/src/components/simpleDialog/SimpleDialog.js
dannyrb: Looks okay in labeling flow. Resolving.

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: Updated. Only render overlay if theres at least one dialog with showOverlay prop, thanks.

igoroctaviano:
> `.DraggableArea` has a width of `100vw` on FireFox. FireFox includes the scroll-bar in that calculation. This causes a horizontal scrollbar to appear. Using `100%` resolves the issue.

I tested here but it was not happening, maybe its my resolution. Updated and tested with width/height 100%. Thanks.

dannyrb:
image

image

platform/viewer/cypress/integration/pwa/OHIFExtensionVTK.spec.js
```diff
@@ -102,7 +103,7 @@ describe('OHIF VTK Extension', () => {
     // Visual comparison
     cy.screenshot('VTK WWWC tool - Canvas should be bright');
     cy.percyCanvasSnapshot('VTK WWWC tool - Canvas should be bright');
-  });
+  }); */
```
dannyrb: https://github.com/OHIF/Viewers/issues/1225

PR #1218, opened 2019-11-22
refactor: Viewports state

igoroctaviano:
- [x] Clear up circular references.
- [x] Structure and document the code.
- [x] Remove dom of active viewport from state. (Avoid crash of redux dev tool / big state)
- [x] Use commandsManager in connected components that were using dom references from store.
- [x] Improve jsDocs.


platform/viewer/src/connectedComponents/ConnectedCineDialog.js
```diff
@@ -16,7 +17,8 @@ const { setViewportSpecificData } = OHIF.redux.actions;
 const mapStateToProps = state => {
   // Get activeViewport's `cine` and `stack`
   const { viewportSpecificData, activeViewportIndex } = state.viewports;
-  const { cine, dom } = viewportSpecificData[activeViewportIndex] || {};
+  const { cine } = viewportSpecificData[activeViewportIndex] || {};
+  const dom = commandsManager.runCommand('getActiveViewportEnabledElement');
```
dannyrb: Ah. Yes. This right here.
Can we duplicate this in the `vtk` code?
(Maybe not until we inject the commandsManager?)

igoroctaviano: Question... We should wrap commandsModule like init and inject the commandsManager?

dannyrb: I think so. We can pass it in the same object we pass serviceManager

extensions/vtk/src/commandsModule.js
```diff
@@ -341,7 +342,7 @@ function getVOIFromCornerstoneViewport(displaySet) {
         windowCenter,
       };
     }
-  }
+  } */
```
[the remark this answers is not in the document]
igoroctaviano: Done! thx.

platform/viewer/src/connectedComponents/ConnectedCineDialog.js
igoroctaviano: Got it. Done! thx.

Issue #1217, opened 2019-11-22
Download/Snapshot Image tool: bugs and improvements


mirnasilva:
### Issue 1: Minimum width and height not defined

1. When user set the width or height to 0 and click on Download button, it doesn't open the download dialog and also doesn't show any error message. 
**Expected Behavior:** The application should display an error message to inform the user that the minimum valid size is 100px, and user should not be able to download the image.

2. When user set the width or height to less than 100px and click on Download button, the downloaded image has size 100x100. Can we set the minimum value to 100px and restrict the user to set values smaller than that? 
**Expected Behavior:**  The application should display an error message to inform the user that the minimum valid size is 100px, and user should not be able to download the image.

### Issue 2: Image is being resized while user still typing the dimensions (width or height)

When user adds a bigger value on height/width field (for example: `10000`), the added value takes some time to be displayed/updated. This is providing a bad user experience because it is updating the image preview at the same type user still typing the value.
**Suggestion:** The Image Preview should be updated only after the user removes the focus of the height/width field. This way we can avoid too many unnecessary on-demand updates on the Image Preview.

### Issue 3: Empty file name

When user deletes the file name no error message is being displayed. Although during the download, the file is displayed with the default name `image.jpg`, letting the user know that the file name is a mandatory field would improve the usability.

### Improvement 4: Width and Height are scaled proportionally

To improve user experience it would be nice to have a visual indication that the image size is being scaled proportionally, like the image below. This way we can avoid having the user trying to set different dimensions unsuccessfully. 

<img width="238" alt="Screen Shot 2019-11-21 at 21 16 38" src="https://user-images.githubusercontent.com/13245529/69445588-65298580-0d31-11ea-802c-74f142d00b16.png">

PR #1219, opened 2019-11-22
E2E Tests: Download Image File Modal

mirnasilva:
### E2E tests for Download Image File Modal

Issue #1217 was opened during the implementation and it's a blocker for some of the test cases.

The following test cases were implemented:
- checks displayed information for Tablet experience
- checks displayed information for Desktop experience
- downloads image file (Blocked)
- cancel changes on download modal
- checks if "Show Annotations" checkbox will display annotations


dannyrb:
@mirnasilva, we have test failures on this one.

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
```diff
@@ -284,17 +284,27 @@ const ViewportDownloadForm = ({
...
+          data-cy="viewport-preview-img"
         />
       </div>
 
       <div className="actions">
         <div className="action-cancel">
-          <button type="button" className="btn btn-danger" onClick={onClose}>
+          <button
+            type="button"
+            data-cy="cancel-btn"
```
rodrigolabs: We already have a `data-cy="cancel-btn"` in `UserPreferencesForm`, I think it's better to rename it to prevent conflicts:

https://github.com/OHIF/Viewers/blob/27e02825a456981158b227e3a26364b9469b8609/platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js#L76-L80

Maybe we can start using some pattern for naming these `data-cy`, what's your opinion @mirnasilva?

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
```diff
@@ -0,0 +1,188 @@
...
+    cy.get('.file-type')
+      .find('select')
+      .select('png');
+
+    // TO-DO: Implement a way to not trigger the button function to open the download browser dialog.
+    // Suggestion of approach: https://github.com/cypress-io/cypress/issues/949
+    // Once we can block the download dialog we can click on Download button.
+    // cy.get('[data-cy="download-btn"]')
+    //   .scrollIntoView()
+    //   .click();
```
dannyrb: Is this something @rodrigolabs can assist with?

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
```diff
@@ -0,0 +1,188 @@
...
+      .should('have.attr', 'src')
+      .and('include', 'data:image');
+    // Check buttons
+    cy.get('[data-cy="cancel-btn"]')
+      .scrollIntoView()
+      .should('have.text', 'Cancel');
+    cy.get('[data-cy="download-btn"]')
+      .scrollIntoView()
+      .should('have.text', 'Download');
+  });
```
dannyrb: Can we keep all of the `text` checks for `Desktop`, but only do a visual/snapshot test for the other resolutions?
If we eventually opt to have different templates at certain breakpoints, we can get more granular.

One thing to verify is that updating one field's value results in the other's updating as well.

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
dannyrb: Because they're in different containers that should never be visible at the same time, it's probably not a huge deal. If the form already has a `data-cy`, we could also key off the button's text after using the container to scope our search.

mirnasilva: I thought the same thing since `UserPreferencesForm` and `DownloadFile` are both modals, they won't be visible at the same.

mirnasilva:
CI is finally passing \o/
@rodrigolabs @dannyrb

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
mirnasilva: With @rodrigolabs 's help, now we are able to test the `Download` button.
We made the download url accessible for cypress and we also blocked the dialog event once we the `Download` button is clicked. Then we verify if the captured url contains the string`"blob:"`.

A better approach would be checking if this url's request returns 200, but due to [this cypress issue](https://github.com/cypress-io/cypress/issues/2918), it is not possible to implement it yet.

mirnasilva:
Hi @dannyrb and @rodrigolabs , I believe now we are good to merge :)
Please let me know if any other changes need to be made. 
Thanks

extensions/cornerstone/src/CornerstoneViewportDownloadForm.js
```diff
@@ -115,6 +115,12 @@ const CornerstoneViewportDownloadForm = ({ onClose, activeViewportIndex }) => {
       const URLObj = window.URL || window.webkitURL;
       const a = document.createElement('a');
       a.href = URLObj.createObjectURL(blob);
+      window.downloadUrl = URLObj.createObjectURL(blob);
+      // Do not open the download dialog if Cypress test is being executed
+      if (window.Cypress) {
+        console.log(window.downloadUrl);
+        return;
+      }
```
dannyrb: I'm hesitant to bake test specific code like this in.
If this is the only way to test this functionality, I'm okay with us foregoing it, or doing a simpler integration test.

platform/ui/src/components/ohifModal/OHIFModal.js
```diff
@@ -29,7 +29,7 @@ const OHIFModal = ({
   const renderHeader = () => {
     return (
       title && (
-        <div className="OHIFModal__header">
+        <div className="OHIFModal__header" data-cy="OHIFModal-header">
```
dannyrb: Personal preference, can we keep `data-cy` values lowercase, words separated by dashes?
Also, something more generic like `modal-header` would work well here :+1:

platform/ui/src/components/ohifModal/OHIFModal.js
```diff
@@ -44,14 +44,17 @@ const OHIFModal = ({
...
+      data-cy="OHIFModal"
       shouldCloseOnEsc={shouldCloseOnEsc}
       isOpen={isOpen}
       title={title}
       style={customStyle}
     >
       <>
         {renderHeader()}
-        <div className="OHIFModal__content">{children}</div>
+        <div className="OHIFModal__content" data-cy="OHIFModal-content">
```
dannyrb: Similarly, this can be `modal-content`

platform/viewer/cypress/support/commands.js
```diff
@@ -469,6 +469,18 @@ Cypress.Commands.add(
...
 );
 
+Cypress.Commands.add('openDownloadImageModal', () => {
+  // Click on More button
+  cy.get('.expandableToolMenu')
+    .as('moreBtn')
+    .click();
+
+  // Click on Download button
+  cy.get('.tooltip-inner > :nth-child(13)')
```
dannyrb: Instead of `:nth-child(13)`, are we able to "get" the button we want using text?
I think it's position is more likely to change than it's label/name

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
```diff
@@ -0,0 +1,152 @@
...
+    cy.waitDicomImage();
+    cy.expectMinimumThumbnails(5);
+  });
+
+  beforeEach(() => {
+    cy.openDownloadImageModal();
+  });
+
+  afterEach(() => {
+    cy.get('[data-cy="close-button"]')
```
dannyrb: Can we add a simple comment above this? `// close modal`

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
```diff
@@ -0,0 +1,152 @@
...
+      .should('have.value', 'jpg');
+    // Check image preview
+    cy.get('[data-cy="image-preview"]').should('contain.text', 'Image Preview');
+    cy.get('[data-cy="viewport-preview-img"]')
+      .should('have.attr', 'src')
+      .and('include', 'data:image');
+    // Check buttons
+    cy.get('[data-cy="cancel-btn"]')
+      .scrollIntoView()
+      .should('have.text', 'Cancel');
```
dannyrb: I don't think we care if it says `Cancel`. We probably mostly care that clicking it has the expected action occur.

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
```diff
@@ -0,0 +1,152 @@
...
+    cy.get('[data-cy="download-btn"]')
+      .scrollIntoView()
+      .should('have.text', 'Download');
```
dannyrb: If we can't test the actual download, verifying the button exists would be the next best thing

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
```diff
@@ -0,0 +1,152 @@
...
+    cy.get('[data-cy=OHIFModal-header]')
+      .as('downloadImageModal')
+      .should('contain.text', 'Download High Quality Image');
+    // Check input fields
+    cy.get('[data-cy=file-info-container]')
+      .should('contain.text', 'Image width (px)')
+      .should('contain.text', 'Image height (px)')
+      .and('contain.text', 'File name')
+      .and('contain.text', 'File type')
+      .and('contain.text', 'Show Annotations');
```
dannyrb: Similarly, by testing that the functionality works as expected, you'll be verifying (to some extent) that all of these fields exist. While the labels are important, losing them would get picked up in a visual test, or an Enzyme/Jest snapshot test

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
```diff
@@ -0,0 +1,152 @@
...
+    //For testing purposes, download dialog event is blocked during cypress execution
+    cy.get('[data-cy="download-btn"]')
+      .scrollIntoView()
+      .click();
+    //Get download url and verify if blob was created
+    cy.window()
+      .its('downloadUrl')
+      .then($url => {
+        expect($url).to.contain('blob:');
+      });
```
dannyrb: What happens when you don't use this workaround? If it exceptions, can we assert that?

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
```diff
@@ -0,0 +1,152 @@
...
+    cy.get('[data-cy="show-annotations"]').check();
+    // Check image preview
+    cy.get('[data-cy="image-preview"]').scrollIntoView();
+    // Visual comparison
+    cy.screenshot('Download Image Modal - Show Annotations checked');
+    cy.percyCanvasSnapshot('Download Image Modal - Show Annotations checked');
+    //Compare classes that exists on Image Preview with Annotations and Without Annotation
+    cy.get('[data-cy="OHIFModal-content"]')
+      .find('canvas')
+      .should('have.class', 'magnifyTool'); //Class "MagnifyTool" exists with annotations displayed on Image preview
```
dannyrb: Why would this have the class `magnifyTool`?

extensions/cornerstone/src/CornerstoneViewportDownloadForm.js
mirnasilva: Yes, that's the only way of testing the `Download` button. We need this not only to expose the url but also to block the dialog event which Cypress cannot handle.
Without this, we don't have any information (url, requests, ...) to verify that the `Download` button is working.

platform/viewer/cypress/support/commands.js
mirnasilva: This is tricky, because the label/name can also change if user picks another language. A better/safer approach would be adding the `data-cy` selectors on the toolbar. What do you think?

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
mirnasilva: The idea is to make sure all elements are displayed, but we can check the same thing with Percy Snapshot comparison. The action of closing the modal is being tested here: `cancel changes on download modal`. Having that said, we can remove this.

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
mirnasilva: We can only if we use the workaround:
```
       if (window.Cypress) {
        console.log(window.downloadUrl);
        return;
      }
```

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
mirnasilva: Actually, nothing happens. We don't have exceptions or requests to make sure `Download` was called.

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
mirnasilva: I have no idea why this class is here, but is the only distinct class that allows me to check if the Image Preview is displayed with/without annotations. This is possible by verifying if class `magnifyTool` is existent or not.

extensions/cornerstone/src/CornerstoneViewportDownloadForm.js
dannyrb: Unfortunately, I think that means we'll need to not test this specific functionality with Cypress.
We'll have to explore integration/unit tests, and assume the browser does it's job if the button is clicked here.

platform/viewer/cypress/support/commands.js
dannyrb: Hmmm... I'm not sure where I fall on this.
We could render `data-cy` using the english label, and have it not tied to translations.
I'm okay to keep this as `:nth-child(13)` for merge, but worth modifying the next time we touch this.

I'd prefer we implement it in `@ohif/viewer`'s toolbarRow without modifying the extension, if we do go down this route.

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
dannyrb: > The idea is to make sure all elements are displayed

I think you've hit the nail on the head when you say:

> we can check the same thing with Percy Snapshot comparison

Visual testing would be too much to maintain with Cypress alone, and very brittle. Percy does _just enough_ to give us a gut check without a huge maintenance overhead.

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
dannyrb: I would prefer we not use that workaround for now

platform/viewer/cypress/integration/common/OHIFDownloadSnapshotFile.spec.js
dannyrb: I can't find anywhere in our application code that does this. I think it may be a `cornerstone` implementation detail.

I wonder if we could snag the `canvas` and compare pixel data to verify instead 🤔 
Follow-up; not blocking.

extensions/cornerstone/src/CornerstoneViewportDownloadForm.js
```diff
@@ -115,7 +115,6 @@ const CornerstoneViewportDownloadForm = ({ onClose, activeViewportIndex }) => {
       const URLObj = window.URL || window.webkitURL;
       const a = document.createElement('a');
       a.href = URLObj.createObjectURL(blob);
-      window.downloadUrl = URLObj.createObjectURL(blob);
```
mirnasilva: Nice catch! Thanks! ;)

PR #1224, opened 2019-11-25
refactor: Shift ConnectedViewportDownloadForm to extension-cornerstone

igoroctaviano:


platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -308,9 +309,17 @@ function _handleBuiltIn(button) {
...
+    );
+
     modal.show({
-      content: ConnectedViewportDownloadForm,
+      content: CornerstoneViewportDownloadForm,
       title: t('Download High Quality Image'),
+      contentProps: {
+        activeViewportIndex,
+        onClose: modal.hide,
+      },
```
dannyrb: If we've moved the `CornerstoneViewportDownloadForm` to the `cornerstone` extension, can't we remove this "behavior" entirely? We should be able to replace the toolbar button's command with one that calls `modal.show` instead of using a "built in behavior"

I want us to slowly remove built in behaviors, because they become part of the contract that every "viewer" would need to provide. Offering the more generic, and optional, UI Services seem to be the more versatile pattern.

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -308,9 +309,17 @@ function _handleBuiltIn(button) {
   }
 
   if (options.behavior === 'DOWNLOAD_SCREEN_SHOT') {
+    const activeViewportIndex = commandsManager.runCommand(
+      'getActiveViewportIndex'
+    );
```
dannyrb: If you shift `modal.show` to inside a command, you can get rid of the `getActiveViewportIndex` command.

extensions/cornerstone/src/commandsModule.js
```diff
@@ -138,9 +138,17 @@ const actions = {
...
+    return viewports.activeViewportIndex;
+  },
 };
 
 const definitions = {
+  getActiveViewportIndex: {
+    commandFn: actions.getActiveViewportIndex,
+    storeContexts: ['viewports'],
+    options: {},
+  },
```
dannyrb: Commented below, but you should be able to remove this. You either have access to store contexts, or you're in the `viewer` project and can use context/redux; not sure there are many other use cases that would require something like this.

igoroctaviano: Done! thx.

platform/viewer/src/connectedComponents/ToolbarRow.js
igoroctaviano: Done! thx.

platform/viewer/src/connectedComponents/ToolbarRow.js
igoroctaviano: Totally get it. Updated! Thank you.

PR #1229, opened 2019-11-26
feat: Ability to configure cornerstone tools via extension configuration

igoroctaviano:
Won't fix:
- Make a more generic context menu with customizable options and actions when calling show.
- Refactor LabellingFlow to DialogService with DialogContent passed from this extension. (Maybe LabellingFlow doesn't need to an independent provider).
- Pass the ServicesManager the "factory function" instead of the plain object.(reducing chances of accidental service use in core app)
- Services should all have a 'common through all services' .create() method instead of a .createUIModalService() method.
- Remove brushTool from our default tools. Include this by default in the seg specific extension.
- UIContextMenuService and UILabelingFlowService look similar to Dialog. We should find a smarter way to say these a flavoured version of UIDialogService.
- Simplify LabellingFlow by absorving LabellingManager logic.
- Pass in rootConfig or applicationConfig to preInit/getModule.

User Cases:
- As a developer, I can configure cornerstone tools via extension configuration.
- As a developer, I can use a supported context menu provider and service which allows me to use or control the context menu dialog through a UI service in my extensions.
- As a developer, I can use a supported labelling flow provider and service which allows me to control the labelling flow dialog through a UI service in my extensions.
- As a developer, I can now use function type configuration which allows me to inject dependencies and have access to ServicesManager and CommandsManager.
- As a developer, I can now have CommandsManager available to my extensions.
- As a developer, I can now use LabellingFlow component independently of ToolContextMenu through a supported UI service.
- As a developer, I can now use a new cornerstone command to get nearby tool data.


platform/viewer/src/store/layout/reducers.js
```diff
@@ -1,33 +0,0 @@
-const defaultState = {
```
dannyrb: Happy to see this go 🔥

igoroctaviano:
@dannyrb I'll stop here :star:, let's discuss what we have in this branch. I'll avoid getting this pr too big so we can have a human reviewable pr. :ok_woman:

platform/viewer/src/store/layout/reducers.js
igoroctaviano: :haircut:

platform/viewer/src/index.js
```diff
@@ -35,15 +27,9 @@ const appDefaults = {
 
 if (window) {
   config = window.config || {};
-  config.extensions = [
-    OHIFVTKExtension,
-    OHIFDicomHtmlExtension,
-    OHIFDicomMicroscopyExtension,
-    OHIFDicomPDFExtension,
-  ];
```
dannyrb: If you remove these from here, and shift them to `App.js`, then they will be bundled in `@ohif/viewer`. We only want them in the PWA application, so we can keep `@ohif/viewer` lightweight (only cornerstone extension, with others explicitly added).

platform/viewer/src/index.js
```diff
@@ -13,19 +13,11 @@ import ReactDOM from 'react-dom';
  * EXTENSIONS
  * =================
  *
- * Importing and modifying the extensions our app uses HERE allows us to leverage
- * tree shaking and a few other niceties. However, by including them here they become
- * "baked in" to the published application.
- *
- * Depending on your use case/needs, you may want to consider not adding any extensions
+ * You may want to consider not adding any extensions
```
dannyrb: haha

platform/viewer/src/connectedComponents/ViewerMain.js
```diff
@@ -153,7 +152,6 @@ class ViewerMain extends Component {
             setViewportData={this.setViewportData}
           >
             {/* Children to add to each viewport that support children */}
-            <ConnectedToolContextMenu />
```
dannyrb: Happy to see this go to. It does make me wonder if we need to provide a blanket way to pass children to Viewports? I'm struggling to think of a valid use case though.

platform/viewer/src/index.js
igoroctaviano: Hahaha... I knew deep down that I would go back, I just wanna check with you how ill add those inside the config function before we call it inside app to inject the deps. This should go as a separate prop?

platform/viewer/public/config/default.js
```diff
@@ -1,72 +1,140 @@
...
-        name: 'DCM4CHEE',
-        wadoUriRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/wado',
-        qidoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
-        wadoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
-        qidoSupportsIncludeField: true,
-        imageRendering: 'wadors',
-        thumbnailRendering: 'wadors',
-        enableStudyLazyLoad: true,
+window.config = ({ servicesManager, dependencies }) => {
+  const { SimpleDialog, merge } = dependencies;
```
igoroctaviano: @dannyrb maybe we can find better ways to pass in components/deps than this one? I'm all :ear:

extensions/cornerstone/src/init.js
```diff
@@ -73,61 +52,40 @@ export default function init({ servicesManager, configuration = {} }) {
...
-        callInputDialog(data, eventDetails, callback),
-    },
-  });
+  if (!isEmpty(configuration.tools)) {
+    /* Add tools with its custom props through extension configuration. */
+    tools.forEach(tool => {
+      const toolName = tool.name.replace('Tool', '');
+      const props = configuration.tools[toolName] || {};
+      csTools.addTool(tool, props);
+    });
```
dannyrb: Are we setting `tools` to `configuration.tools`? I notice that, despite the check, we still hook up our locally defined ones? We may also want this to support the same config / abilities as the `react-cornerstone-viewport` tools prop.

extensions/cornerstone/src/init.js
```diff
@@ -73,61 +52,40 @@ export default function init({ servicesManager, configuration = {} }) {
...
+    csTools.BidirectionalTool,
+    csTools.LengthTool,
+    csTools.AngleTool,
+    csTools.FreehandRoiTool,
+    csTools.EllipticalRoiTool,
+    csTools.DragProbeTool,
+    csTools.RectangleRoiTool,
     // Segmentation
-    BrushTool,
+    csTools.BrushTool,
```
dannyrb: Unrelated, but we can remove `brushTool` from our default tools.
I think we'll only include this by default in the seg specific extension.

dannyrb: FYI, good call on nixing the destructure. This is cleaner.

platform/viewer/src/App.js
```diff
@@ -104,13 +110,30 @@ class App extends Component {
...
-    _initServices([UINotificationService, UIModalService, UIDialogService]);
-    _initExtensions(extensions, hotkeys);
+    _initServices([
+      UINotificationService,
+      UIModalService,
+      UIDialogService,
+      UIContextMenuService,
+      UILabellingFlowService,
+    ]);
+    _initExtensions([...defaultExtensions, ...extensions], hotkeys, tools);
```
igoroctaviano: @dannyrb I thought about adding them this way, what do you think? we can have the extensions coming from the config file and the defaultExtensions added using the index.js. Maybe more explicit? :mag_right:

extensions/cornerstone/src/init.js
```diff
@@ -73,61 +52,40 @@ export default function init({ servicesManager, configuration = {} }) {
...
-
-  csTools.addTool(csTools.ArrowAnnotateTool, {
-    configuration: {
-      getTextCallback: (callback, eventDetails) =>
-        callInputDialog(null, eventDetails, callback),
-      changeTextCallback: (data, eventDetails, callback) =>
-        callInputDialog(data, eventDetails, callback),
-    },
-  });
+  if (!isEmpty(configuration.tools)) {
```
dannyrb: isEmpty would just be checking an array? Could just use `.length` instead of `lodash`?

platform/viewer/public/config/default.js
```diff
@@ -1,72 +1,140 @@
...
+        label: 'New label',
+        measurementData: data ? { description: data.text } : {},
+        onClose: () => UIDialogService.dismiss({ id: dialogId }),
+        onSubmit: value => {
+          callback(value);
+          UIDialogService.dismiss({ id: dialogId });
+        },
+      },
+    });
+  };
```
dannyrb: I think the `cornerstone` extension should use our custom Dialog by default? This _should_ be possible to do here, but we would only do it to override the dialog the `cornerstone` extension uses by default.

platform/ui/src/contextProviders/DialogProvider.styl
```diff
@@ -1,8 +1,8 @@
-.DraggableItem
+.DraggableItem.draggable
   div
     cursor: grab !important
 
-.DraggableItem.dragging
+.DraggableItem.draggable.dragging
```
dannyrb: If it's a DraggableItem, why wouldn't it already be `.draggable`?

extensions/cornerstone/src/init.js
```diff
@@ -26,32 +26,11 @@ cornerstone.metaData.addProvider(fallbackMetaDataProvider, -1);
...
-        label: 'New label',
-        measurementData: data ? { description: data.text } : {},
-        onClose: () => UIDialogService.dismiss({ id: dialogId }),
-        onSubmit: value => {
-          callback(value);
-          UIDialogService.dismiss({ id: dialogId });
-        },
-      },
-    });
-  };
```
dannyrb: We likely want to keep this here. See comment in `default.js`

platform/ui/src/contextProviders/DialogProvider.styl
igoroctaviano: This is because a draggable item can have its dragging functionality deactivated for a while if needed.

platform/core/src/extensions/ExtensionManager.js
```diff
@@ -26,7 +26,7 @@ export default class ExtensionManager {
       const hasConfiguration = Array.isArray(extension);
 
       if (hasConfiguration) {
-        const [ohifExtension, configuration] = extensions;
+        const [ohifExtension, configuration] = extension;
```
dannyrb: Oops, haha. Good catch.

platform/core/src/services/UIContextMenuService/index.js
```diff
@@ -0,0 +1,63 @@
...
+}) {
+  if (showImplementation) {
+    uiContextMenuServiceImplementation._show = showImplementation;
+  }
+  if (hideImplementation) {
+    uiContextMenuServiceImplementation._hide = hideImplementation;
+  }
+}
+
+export default createUIContextMenuService;
```
dannyrb: `UIContextMenuService` and `UILabelingFlowService` look similar to Dialog. I'm wondering if there's a smart way to say these a flavored version of `UIDialogService`?

I'll keep reading, but we should at least think about other ways to structure this.

platform/core/src/services/UIDialogService/index.js
```diff
@@ -54,6 +56,8 @@ function create({
   onStart,
   onDrag,
   onStop,
+  centralize = false,
+  useLastPosition = true,
```
dannyrb: Are there instances where `isDraggable` would be true, but this would be `false`?
Maybe instead of having the service remember the dialog's last position, we should have it return it's position when it closes? Then the caller can track and use it as initial values when opening if it wants this behavior?

igoroctaviano: by this you mean lastPosition or centralize? isDraggable is independent of them, it just freezes  the component in a default position or last position if required.  I was wondering if the last position of a dialog would be better living outside... because this would require a new state in every component that needs this behaviour.

platform/core/src/services/UIContextMenuService/index.js
igoroctaviano: Yeah, they have the same service/factory interface... we can have the dialog service generic enough to be used to instantiate dialogs and flavored versions of dialogs but this wouldnt be flexible. Maybe migrate to class and use extends, something like BaseTool in cornerstone?

extensions/cornerstone/src/init.js
igoroctaviano: Updated.

extensions/cornerstone/src/init.js
igoroctaviano: Its checking for an empty object, but I have used the native version with object.keys length. Done! thanks.

extensions/cornerstone/src/init.js
igoroctaviano: configuration.tools is the default configuration, Inside configuration tools object, each key is a different tool and the user can add any prop that will be passed as arg to addTool.

platform/viewer/src/index.js
igoroctaviano: Just moved this code back and made some changes in the props (App.js), what about now?

platform/viewer/public/config/default.js
igoroctaviano: Makes sense! thanks, moved.

platform/ui/src/contextProviders/ContextMenuProvider.js
```diff
@@ -0,0 +1,136 @@
...
+              props: {
+                measurementData,
+                editDescriptionOnDialog: true,
+              },
+            }),
+        },
+        defaultPosition: _getDefaultPosition(event),
+      });
+    },
+    [create, dismiss, hide, showLabellingFlow]
```
dannyrb: We may need to do this gradually, but I would think `showLabelingFlow` would not exist directly in `ContextMenuProvider`. When creating the context menu, you would probably want to be able to set the options in the context menu, and the behavior when they're clicked.

Kicking off the labeling flow would be configuration/props?

platform/ui/src/contextProviders/ContextMenuProvider.js
```diff
@@ -0,0 +1,136 @@
...
+import { useDialog } from './DialogProvider';
+import { useLabellingFlow } from './LabellingFlowProvider';
+
+const ContextMenuContext = createContext(null);
+const { Provider } = ContextMenuContext;
+
+export const useContextMenu = () => useContext(ContextMenuContext);
+
+const ContextMenuProvider = ({ children, service }) => {
+  const { create, dismiss } = useDialog();
```
dannyrb: Okay, so it is using `useDialog` under the hood :+1:
We're just exposing it as a named service, and this separate context provider keeps the state separate 🤔 

Interesting stuff to think about.

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -20,7 +20,30 @@ export const useDialog = () => useContext(DialogContext);
...
+    const root = document.querySelector('#root');
+    const centerX = root.offsetLeft + root.offsetWidth / 2;
+    const centerY = root.offsetTop + root.offsetHeight / 2;
+    const item = document.querySelector(`#draggableItem-${id}`);
+    const itemBounds = item.getBoundingClientRect();
+    return {
+      x: centerX - itemBounds.width / 2,
+      y: centerY - itemBounds.height / 2,
+    };
+  };
```
dannyrb: If we shift tracking the "last position" out of the DialogProvider, we could say, if no position is set, default to center. That could simplify the API a bit.

dannyrb: Ah. Actually, not contingent on shifting last position out, but it still makes sense as a default position.

I'm not sure when/if we would want to snap the dialog back to center. It'd probably only be an initial state, and one that you'd override if dragged/re positioned.

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -118,6 +148,8 @@ const DialogProvider = ({ children, service }) => {
         contentProps,
         position,
         defaultPosition,
+        centralize = false,
+        useLastPosition = true,
```
dannyrb: `preservePosition` ?

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -118,6 +148,8 @@ const DialogProvider = ({ children, service }) => {
         contentProps,
         position,
         defaultPosition,
+        centralize = false,
```
dannyrb: nix, and use center as position if none is provided for `position` or `defaultPosition`? Could also allow `defaultPosition` to support `string` values like `left`, `center`, `topRight`, if you felt it was important.

dannyrb: Just curious, do we need `defaultPosition`? Do dialogs need to be able to return to an initial position?

For example:

- position: `null`; default to center
- position: `{ x: 200, y: 200 }`; initial
    - When dialog is closed: `{ position: { x, y } }`
    - Optionally use those values when opening same dialog if you want to preserve
    - OR, if it's draggable, track internally and key off `preservePosition` flag

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -128,8 +160,14 @@ const DialogProvider = ({ children, service }) => {
...
-          position={position}
-          defaultPosition={lastDialogPosition || defaultPosition}
+          position={
+            centralize
+              ? centerPositions.find(position => position.id === id)
+              : position
+          }
+          defaultPosition={
+            (useLastPosition && lastDialogPosition) || defaultPosition
+          }
```
dannyrb: Ah. Part of this is we're keying off `Draggable`'s API.
Their API is likely to be more generic, as it needs to support more use cases.
If we can narrow/simplify it for our needs in this specific situation, we should.

dannyrb: If we are conditionally draggable, we could forego wrapping our dialog in `<Draggable>` if we're not draggable? Saves you on some of the odd classnames around `.DraggableItem.draggable`

platform/ui/src/contextProviders/DialogProvider.js
```diff
@@ -169,7 +207,11 @@ const DialogProvider = ({ children, service }) => {
         >
           <div
             id={`draggableItem-${id}`}
-            className={classNames('DraggableItem', isDragging && 'dragging')}
+            className={classNames(
+              'DraggableItem',
+              isDragging && 'dragging',
+              isDraggable && 'draggable'
```
dannyrb: classNames supports a nice syntax like:

```js
classNames(
  'DraggableItem',
  {
    'is-dragging': isDragging,
    'is-draggable': isDraggable
   });
```

Your syntax makes sense; just caught me off-guard

platform/ui/src/contextProviders/LabellingFlowProvider.js
```diff
@@ -0,0 +1,131 @@
+import React, {
+  createContext,
+  useContext,
+  useEffect,
+  useCallback,
+} from 'react';
+import PropTypes from 'prop-types';
+
+import LabellingManager from '../../../viewer/src/components/Labelling/LabellingManager';
```
dannyrb: This import is across package boundaries?

platform/ui/src/contextProviders/ContextMenuProvider.js
```diff
@@ -0,0 +1,136 @@
+import React, {
+  createContext,
+  useContext,
+  useEffect,
+  useCallback,
+} from 'react';
+import PropTypes from 'prop-types';
+
+import ToolContextMenu from '../../../viewer/src/connectedComponents/ToolContextMenu';
```
dannyrb: This import is across package boundaries?

platform/ui/src/contextProviders/LabellingFlowProvider.js
```diff
@@ -0,0 +1,131 @@
...
+
+LabellingFlowProvider.propTypes = {
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
+  service: PropTypes.shape({
+    setServiceImplementation: PropTypes.func,
+  }),
+  commandsManager: PropTypes.object.isRequired,
```
dannyrb: Instead of taking in the entire commands manager, have this be a callback method? `onSelection` or something. Then, in the props for the provider, set the value as:

```js
onSelect={(val) => {
      commandsManager.runCommand(
        'updateTableWithNewMeasurementData',
        measurementData
      );
}}
```

Or something similar. Our generic UI shouldn't need to know or care about the commands manager

platform/ui/src/contextProviders/LabellingFlowProvider.js
igoroctaviano: Actually i think those providers should be located in viewer, right? :boom:

platform/viewer/src/connectedComponents/ToolContextMenu.js
```diff
@@ -17,232 +14,125 @@ const toolTypes = [
...
-      );
-      cornerstone.updateImage(element);
+const ToolContextMenu = ({
+  onSetLabel,
+  onSetDescription,
+  isTouchEvent,
+  eventData,
+  onClose,
+}) => {
+  const mainElement = useRef();
```
dannyrb: I feel like there's a better way to dynamically update position?

platform/viewer/src/connectedComponents/ToolContextMenu.js
```diff
@@ -1,9 +1,6 @@
-import React, { Component } from 'react';
+import React, { useRef } from 'react';
 import PropTypes from 'prop-types';
-import cornerstone from 'cornerstone-core';
-import cornerstoneTools from 'cornerstone-tools';
-// This whole component should live in the Measurements Extension :thinking:
-import getMeasurementLocationCallback from '../appExtensions/MeasurementsPanel/getMeasurementLocationCallback';
+import { commandsManager } from './../App.js';
```
dannyrb: Not as big of a deal, because this is in the viewer project, but might make more sense to identify the actions that trigger commands, expose them as props, and then bind props to appropriate command callbacks.

platform/viewer/src/connectedComponents/ToolContextMenu.js
igoroctaviano: mainElement is not necessary anymore, i can take it out.

platform/viewer/src/connectedComponents/ToolContextMenu.js
igoroctaviano: Makes sense, the same with providers usage of commandsManager.

dannyrb: Is `ToolContextMenu` a flavor of `ContextMenu` 🤔

platform/viewer/src/components/Labelling/LabellingManager.js
```diff
@@ -1,21 +1,16 @@
 import React, { Component } from 'react';
```
dannyrb: I assume the LabellingManager wraps the different labeling components and orchestrates their "flow"? 

If so, does it only render the `<LabelingFlow>` component? I think that would make this a container component we're trying to only have business logic in?

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
```diff
@@ -131,15 +40,40 @@ export default function init({ servicesManager, configuration = {} }) {
...
+  };
+
+  const onTouchStart = () => {
+    commandsManager.runCommand('hideContextMenu');
+    commandsManager.runCommand('hideLabellingFlow');
+  };
+
+  const onMouseClick = () => {
+    commandsManager.runCommand('hideContextMenu');
+    commandsManager.runCommand('hideLabellingFlow');
```
dannyrb: Commands added by this extension, and not service methods? Not arguing, this just caught me off-guard. I'll have to find their implementation before I have better thoughts.

platform/ui/src/contextProviders/LabellingFlowProvider.js
igoroctaviano: Yeah but im afraid that the provider will have a bunch of callbacks being passed through App.js. Do you have any different alternative in mind?

platform/viewer/src/components/Labelling/LabellingManager.js
igoroctaviano: Yeah, looks like it orchestrates to show edit description dialog or the labelling flow, I could level this code up to the provider and remove this component entirely.

platform/viewer/src/connectedComponents/ToolContextMenu.js
igoroctaviano: ContextMenu is the provider and ToolContextMenu the current component, its very related to tool but I changed the provider name to be just contextmenu, should i update to follow the component name?

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: ...me not reading their documentation :eyes: very cool!

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: If we want to open the dialog directly on top of a tool which is a very common the defaultPosition would be very handy.

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: Yeah, i think that would be better

platform/ui/src/contextProviders/ContextMenuProvider.js
igoroctaviano: Yeah, they are independent and they are using each other using context like other ui components use modal. Kicking off labelling flow would be through props when calling it through its provider show method or directly at component declaration.

platform/ui/src/contextProviders/ContextMenuProvider.js
igoroctaviano: Yeah the providers itself follows the same context rules as other components inside app, they can access different contexts to leverage the existent generic components, this way they can extend themselves.

platform/viewer/public/config/default.js
```diff
@@ -1,72 +1,116 @@
...
+  );
+
+  /* Add specific tools configuration here. */
+  const specificToolsProps = {
+    /* ArrowAnnotate: {
+      configuration: {
+        ...
+      },
+    }, */
+  };
```
dannyrb: While `allToolProps` and `specificToolProps` are nice guides for how to do this, now that we've offloaded to the extension, can these be removed?

platform/viewer/src/App.js
```diff
@@ -79,23 +87,20 @@ window.store = store;
...
+    defaultExtensions: PropTypes.array,
   };
 
   static defaultProps = {
     whiteLabelling: {},
     oidc: [],
-    extensions: [],
+    config: {
+      extensions: [],
+    },
```
dannyrb: `whiteLabelling`, `oidc`, etc. are config properties. Why the addition of a wrapper object? Everything this component receives is technically application configuration.

platform/viewer/public/html-templates/script-tag.html
```diff
@@ -32,7 +32,7 @@
       console.log('OHIF Viewer rendered/updated');
     };
     window.OHIFViewer.installViewer(
-      window.config, // See `/config/default.js`
+      { config: window.config }, // See `/config/default.js`
```
dannyrb: See `App.js` comment. I'm unsure of the value of this change.

platform/viewer/src/components/EditDescriptionDialog/EditDescriptionDialog.js
```diff
@@ -1,24 +1,20 @@
...
 
+import SimpleDialog from '../SimpleDialog/SimpleDialog.js';
 import bounding from '../../lib/utils/bounding.js';
-import { getDialogStyle } from './../Labelling/labellingPositionUtils.js';
-
 import './EditDescriptionDialog.css';
 
 export default class EditDescriptionDialog extends Component {
   static defaultProps = {
     componentRef: React.createRef(),
```
dannyrb: Ref still needed?

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
```diff
@@ -201,26 +198,20 @@ const mapDispatchToProps = dispatch => {
...
       });
 
-      const options = {
+      const props = {
         editDescriptionOnDialog: true,
+        measurementData: tool,
       };
 
-      getMeasurementLocationCallback(eventData, tool, options);
+      ownProps.commandsManager.runCommand('showLabellingFlow', {
```
dannyrb: Similar to other things we've discussed. I'm hesitant to tie a component directly to `commandsManager`
Because this would be isolated to an extension, that bothers me less, but still something we should discuss.

extensions/cornerstone/src/commandsModule.js
```diff
@@ -157,9 +157,149 @@ const commandsModule = ({ servicesManager }) => {
...
+      const { UIContextMenuService } = servicesManager.services;
+      UIContextMenuService.hide();
+    },
+    showLabellingFlow({ defaultPosition, centralize, props }) {
+      const { UILabellingFlowService } = servicesManager.services;
+      UILabellingFlowService.show({
+        defaultPosition,
+        centralize,
+        props,
+      });
```
dannyrb: If this is just a generic command that has no cornerstone dependency and is a pass-through call to a service, why can't we just call the service directly? This feels like an unnecessary layer?

platform/ui/src/contextProviders/DialogProvider.js
dannyrb: Wouldn't we just set `position`?

extensions/cornerstone/src/commandsModule.js
igoroctaviano: Done! thanks.

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
igoroctaviano: Totally agree. Done! thanks.

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
igoroctaviano: Done! thanks. Now we are using services. :)

platform/ui/src/contextProviders/LabellingFlowProvider.js
igoroctaviano: Done. Components are now passed through props. thanks.

platform/ui/src/contextProviders/ContextMenuProvider.js
igoroctaviano: Done. Components are now passed through props. thanks.

platform/viewer/src/components/EditDescriptionDialog/EditDescriptionDialog.js
igoroctaviano: 🥇:) Removed, thanks.

platform/viewer/public/html-templates/script-tag.html
igoroctaviano: Updated, now I think its clearer.

platform/viewer/src/App.js
igoroctaviano: Updated, now I think its clearer.

platform/viewer/public/config/default.js
igoroctaviano: You right! removed. Thanks. 👍

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: The defaultPosition will be set only once, position can be used for aync positioning (internally now). I have updated the code so that we only expose defaultPosition and use position only to centralize the component internally (async task since we want to calculate based on the rendered element).

platform/ui/src/contextProviders/DialogProvider.js
igoroctaviano: We still need them to be wrapped in Draggable because we are leveraging positioning/events. The odd classNames are basically to show or not the drag icon (dragging => closed hand and drag => openhand).

platform/viewer/src/index.js
```diff
@@ -27,26 +27,27 @@ import OHIFDicomHtmlExtension from '@ohif/extension-dicom-html';
...
+  defaultExtensions: [
     OHIFVTKExtension,
     OHIFDicomHtmlExtension,
     OHIFDicomMicroscopyExtension,
     OHIFDicomPDFExtension,
-  ];
-}
-
-const appProps = Object.assign({}, appDefaults, config);
+  ],
```
dannyrb: This is a breaking change. We need to document this and do a major version bump.

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
```diff
@@ -131,15 +45,38 @@ export default function init({ servicesManager, configuration = {} }) {
...
+
+  const onTouchStart = () => {
+    UILabellingFlowService.hide();
+    UIContextMenuService.hide();
+  };
+
+  const onMouseClick = () => {
+    UIContextMenuService.hide();
+    UILabellingFlowService.hide();
+  };
```
dannyrb: We have no guarantees that a service is registered; we need to check if they are before calling methods on them.

We should probably create follow-up issues to make a more generic contextmenu? Next step would be this extension provided context options + actions when calling `show`.

Similarly, labeling flow, in the future, could probably be a DialogService with DialogContent passed from this extension?

platform/viewer/src/appExtensions/MeasurementsPanel/index.js
```diff
@@ -7,10 +8,34 @@ export default {
    */
   id: 'measurements-table',
 
-  preRegistration({ servicesManager, configuration = {} }) {
-    init({ servicesManager, configuration });
+  preRegistration({ servicesManager, commandsManager, configuration = {} }) {
```
dannyrb: An aside -- We should maybe also pass in `rootConfig` or `applicationConfig`

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
```diff
@@ -201,26 +189,12 @@ const mapDispatchToProps = dispatch => {
...
       const tool = MeasurementApi.Instance.tools[toolType].find(measurement => {
         return measurement._id === measurementId;
       });
 
-      const options = {
-        editDescriptionOnDialog: true,
-      };
-
-      getMeasurementLocationCallback(eventData, tool, options);
+      ownProps.onEditDescription(tool);
```
dannyrb: Is this required/guaranteed to be defined? Same with other callbacks.

platform/viewer/public/config/public_dicomweb.js
```diff
@@ -1,5 +1,6 @@
 window.config = {
   routerBasename: '/',
+  whiteLabelling: {},
```
dannyrb: Is this required, or are we adding it just so users know it's an option?

igoroctaviano: Before it was working without it because we had a default prop for whiteLabelling, now we don't because of the config which its a function return now.

dannyrb: After the function return, we can still `Object.assign({}, outputConfig, defaultConfig)` if we need to support defaults.

extensions/cornerstone/src/commandsModule.js
```diff
@@ -157,9 +157,109 @@ const commandsModule = ({ servicesManager }) => {
...
 
   const definitions = {
+    getNearbyToolData: {
+      commandFn: actions.getNearbyToolData,
+      storeContexts: ['viewports'],
+      options: {},
+    },
+    removeToolState: {
+      commandFn: actions.removeToolState,
+      storeContexts: ['viewports'],
```
dannyrb: It looks like this isn't using the `viewports` context. You can probably remove.

extensions/cornerstone/src/commandsModule.js
```diff
@@ -157,9 +157,109 @@ const commandsModule = ({ servicesManager }) => {
...
+      options: {},
+    },
+    updateTableWithNewMeasurementData: {
```
dannyrb: Same fore these other new commands.

platform/viewer/public/html-templates/script-tag.html
```diff
@@ -31,8 +31,9 @@
     var componentRenderedOrUpdatedCallback = function () {
       console.log('OHIF Viewer rendered/updated');
     };
+    var props = { config: window.config };
     window.OHIFViewer.installViewer(
-      window.config, // See `/config/default.js`
+      props, // See `/config/default.js`
```
dannyrb: I think we can keep this passing in an object or a function. I'm not sure the `config` key adds value. Then `installViewer` can set it up appropriately for the App component?

dannyrb:
Let's put a bow on this:

1. Identify "wontfix" items for this PR that should result in new issues and follow-up items.
2. Create "User Cases" that tell the story of the features/functionality that are added here.
  - `As a developer, I can now add custom tools by changing the cornerstone extension's configuration`
  - `As a user, I can do fancy new thing.`
3. Anything breaking? Let's list those so we can draft a tiny statement.

platform/viewer/public/html-templates/script-tag.html
igoroctaviano: Got it. I thought to add the config key would allow more stuff beside config to be passed down as props since the installViewer have this param named as props. I'll update the installViewer props argument to config to be more precise about its usage.

extensions/cornerstone/src/commandsModule.js
igoroctaviano: Alright, done. Thanks.

extensions/cornerstone/src/commandsModule.js
igoroctaviano: Alright, done. Thanks.

platform/viewer/public/config/public_dicomweb.js
igoroctaviano: Alright, done. Thanks.

platform/viewer/src/appExtensions/MeasurementsPanel/ConnectedMeasurementTable.js
igoroctaviano: Alright, done. Thanks.

platform/viewer/src/appExtensions/MeasurementsPanel/index.js
igoroctaviano: Added in the comment. 👍

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
igoroctaviano: Alright, done. Added in the comment about (wontfix) this follow-up issue.

igoroctaviano:
> Let's put a bow on this:
> 
> 1. Identify "wontfix" items for this PR that should result in new issues and follow-up items.
> 2. Create "User Cases" that tell the story of the features/functionality that are added here.
> 
> * `As a developer, I can now add custom tools by changing the cornerstone extension's configuration`
> * `As a user, I can do fancy new thing.`
> 
> 1. Anything breaking? Let's list those so we can draft a tiny statement.


Won't fix:
- Make a more generic context menu with customizable options and actions when calling show.
- Refactor LabellingFlow to DialogService with DialogContent passed from this extension. (Maybe LabellingFlow doesn't need to an independent provider).
- Pass the ServicesManager the "factory function" instead of the plain object.(reducing chances of accidental service use in core app)
- Services should all have a 'common through all services' .create() method instead of a .createUIModalService() method.
- Remove brushTool from our default tools. Include this by default in the seg specific extension.
- UIContextMenuService and UILabelingFlowService look similar to Dialog. We should find a smarter way to say these a flavoured version of UIDialogService.
- Simplify LabellingFlow by absorving LabellingManager logic.
- Pass in rootConfig or applicationConfig to preInit/getModule.

User Cases:
- As a developer, I can configure cornerstone tools via extension configuration.
- As a developer, I can use a supported context menu provider and service which allows me to use or control the context menu dialog through a UI service in my extensions.
- As a developer, I can use a supported labelling flow provider and service which allows me to control the labelling flow dialog through a UI service in my extensions.
- As a developer, I can now use function type configuration which allows me to inject dependencies and have access to ServicesManager and CommandsManager.
- As a developer, I can now have CommandsManager available to my extensions.
- As a developer, I can now use LabellingFlow component independently of ToolContextMenu through a supported UI service.
- As a developer, I can now use a new cornerstone command to get nearby tool data.

dannyrb:
The closest thing to QA we have for user stories that impact developers are `unit tests`. I think we can forego the normal @mirnasilva approved process until we have a better way to run integration/e2e tests for extension and service changes.

It's interesting how much of this work hopefully has no impact on current functionality, but also improves the experience for others trying to tweak the viewer.

I think the only things we have to make sure we do before this can be merged are:

- [ ] Update any documentation and code around the updated `<App />` component props
- [ ] Verify the cornerstone extension config changes work
    - [ ] Update the `default` config for a single commit, and share the Deploy Preview for it (then revert in next commit)?
    - [ ] Draft at least example schema for the cornerstone extension in it's readme
- [ ] Make sure we note the breaking change

igoroctaviano:
> The closest thing to QA we have for user stories that impact developers are `unit tests`. I think we can forego the normal @mirnasilva approved process until we have a better way to run integration/e2e tests for extension and service changes.
> 
> It's interesting how much of this work hopefully has no impact on current functionality, but also improves the experience for others trying to tweak the viewer.
> 
> I think the only things we have to make sure we do before this can be merged are:
> 
> * [ ]  Update any documentation and code around the updated `<App />` component props
> * [ ]  Verify the cornerstone extension config changes work
>   
>   * [ ]  Update the `default` config for a single commit, and share the Deploy Preview for it (then revert in next commit)?
>   * [ ]  Draft at least example schema for the cornerstone extension in it's readme
> * [ ]  Make sure we note the breaking change
> 
> @igoroctaviano, feel free to push back on, or add to, this list.

https://5dee6d5c04b544000807e9db--ohif.netlify.com/pwa/viewer/1.2.826.0.13854362241694438965858641723883466450351448

docs/latest/configuring/data-source.md
```diff
@@ -130,6 +130,31 @@ window.config = {
 };
 ```
 
+The configuration can also be written as a JS Function in case you need to inject dependencies like external services:
+
+```js
+window.config = ({ randomService } = {}) => {
```
dannyrb: Wouldn't `randomService` actually be the `ServicesManager`?

extensions/_example/src/index.js
```diff
@@ -12,8 +12,9 @@ export default {
    */
 
   preRegistration({
-    servicesManager,
-    configuration: extensionConfiguration,
+    servicesManager = {},
+    commandsManager = {},
```
dannyrb: Why would `servicesManager` or `commandsManager` ever be undefined?

docs/latest/deployment/recipes/embedded-viewer.md
```diff
@@ -49,6 +49,33 @@ window.config = {
 };
 ```
 
+```js
```
dannyrb: Instead of duplicating config docs, we should link to `configuring` docs page.

platform/viewer/src/appExtensions/MeasurementsPanel/index.js
```diff
@@ -7,10 +8,42 @@ export default {
    */
   id: 'measurements-table',
 
-  preRegistration({ servicesManager, configuration = {} }) {
-    init({ servicesManager, configuration });
+  preRegistration({
+    servicesManager = {},
+    commandsManager = {},
```
dannyrb: Same here. If `servicesManager` and `commandsManager` are undefined here, I want everything to explode.

platform/viewer/src/App.js
```diff
@@ -213,30 +273,31 @@ function _initServices(services) {
 /**
  * @param
  */
-function _initExtensions(extensions, hotkeys) {
-  const defaultExtensions = [
+function _initExtensions(extensions, tools) {
+  const requiredExtensions = [
```
dannyrb: Ah. I see what you did here 🥇

platform/viewer/src/App.js
```diff
@@ -104,31 +119,50 @@ class App extends Component {
...
-    _initServices([UINotificationService, UIModalService, UIDialogService]);
-    _initExtensions(extensions, hotkeys);
+    _initServices([
+      UINotificationService,
+      UIModalService,
+      UIDialogService,
+      UIContextMenuService,
+      UILabellingFlowService,
+    ]);
+    _initExtensions([...defaultExtensions, ...extensions], tools);
```
dannyrb: Okay. I see why you did this. We are tightly coupled to `cornerstone` extension, so it's _always_ baked in. That makes it difficult to provide it's configuration in the `extensions` array.

Having a specific spot for it in the application config stinks, but it's our only option until we decouple.

My only ask is that we switch from providing something as specific as `tools`, and just pass in an entire `cornerstoneExtensionConfig` that has a `tools` key, that way we won't have a breaking change as we add more values we'd like to set.

docs/latest/configuring/data-source.md
igoroctaviano: Bad example =S, updated. Thanks.

platform/viewer/src/App.js
igoroctaviano: Updated.

platform/viewer/src/appExtensions/MeasurementsPanel/index.js
igoroctaviano: Updated.

docs/latest/deployment/recipes/embedded-viewer.md
igoroctaviano: Updated.

extensions/_example/src/index.js
igoroctaviano: Updated.

PR #1231, opened 2019-11-27
fix of undefined name of project

pavertomato:
seems like some of internal google's project may contain incorrect characters  and won't be parsed in json response

### DIFF
--- a/platform/viewer/src/googleCloud/ProjectsList.js
+++ b/platform/viewer/src/googleCloud/ProjectsList.js
@@ -63,9 +63,18 @@ class ProjectsList extends Component {
     if (loading) {
       return loadingIcon;
     }
+
+    const lowerCaseFilter = filter.toLowerCase();
+    const filteredProjects = projects.filter(project =>   
+      typeof project.name  === 'string' &&
+      (filter === "" || project.name.toLowerCase().includes(lowerCaseFilter))
+    );
+
     const body = (
       <tbody id="ProjectList">
-        {projects.filter(project => (project.name.toLowerCase().includes(filter.toLowerCase()) || filter=="")).map(this.renderTableRow)}
+        {
+          filteredProjects.map(this.renderTableRow)
+        }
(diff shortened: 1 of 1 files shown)


platform/viewer/src/googleCloud/ProjectsList.js
```diff
@@ -65,7 +65,11 @@ class ProjectsList extends Component {
     }
     const body = (
       <tbody id="ProjectList">
-        {projects.filter(project => (project.name.toLowerCase().includes(filter.toLowerCase()) || filter=="")).map(this.renderTableRow)}
+        {
+          projects.filter(project => ( typeof project.name  === 'string' &&
+            (filter=="" || project.name.toLowerCase().includes(filter.toLowerCase()))
+          )).map(this.renderTableRow)
+        }
```
dannyrb: Can you break this down outside of the render function so it's a bit easier to reason about?
Something like:

```js
const lowerCaseFilter = filter.toLowerCase();
const filteredProjects = projects.filter(project => 
    typeof project.name  === 'string' &&
    (filter === "" || project.name.toLowerCase().includes(lowerCaseFilter));

...

{ filteredProjects.map(this.renderTableRow) }
```

pavertomato: not sure if got this right, i've moved filtering from body generation

dannyrb: If you've tested this, then I'm happy to merge. Comment and let me know, and I'll pull the merge lever.

pavertomato: sure, i've tested it, also have tested in docker image
also i may deploy on publicly available instance

dannyrb: If the PR is coming from a branch off of `OHIF/Viewers`, you can approve this workflow and a test docker image will be published:

image

Issue #1095, opened 2019-10-25
Hotkeys not working: Z, PageUp, PageDown, =


mirnasilva:
## Issue 1 - Description:

Hotkeys not working on the Viewer page: `Z`, `PageUp`, `PageDown`.

### Steps to Reproduce

1. Open a study
2. Try to use the following hotkeys: `Z`, `PageUp`, `PageDown`
3. Checks that nothing happened.

### Actual Result

Hotkeys `Z`, `PageUp`, `PageDown` are not working, nothing happens at the viewer page.

### Expected Result

1. When user hits hotkey `Z`, the Zoom tool should be activated at the toolbar
2. User should be able to navigate on the thumbnails using hotkeys: `PageUp` and `PageDown`
2.1. When user hits hotkey `PageDown`, the next thumbnail series should be loaded in the viewport
2.2. When user hits hotkey `PageUp`, the previous thumbnail series should be loaded in the viewport

## Issue 2 - Description:

When there are multiple viewports, hotkey `=` does not reset the `inverted` status of the image in the active viewport.

In the gif below, user presses the `=` hotkey to reset all viewports before clicking on the `Reset` button

![bug-hotkey-equal](https://user-images.githubusercontent.com/13245529/67607038-82325f00-f759-11e9-81e2-ff10150a3da4.gif)

### Steps to Reproduce

1. Open a study
2. Change layout to 3 viewports
3. Invert the image by pressing hotkey `I`
4. Press hotkey `=`

### Actual Result

Hotkey `=` does not reset the `inverted` status of the image. If user clicks on the `Reset` button, the `inverted` image will be reset with success.

### Expected Result

User should be able to reset the `inverted` image (in the active viewport) by pressing hotkey `=` or clicking on `Reset` button on toolbar.

dannyrb:
### Quick Overview

- Extensions have a "CommandModule"
- A "CommandModule" adds named commands
  - Commands are used if there "context" is active

For example, the `cornerstone` extension has some command named `Rotate` that is active in the `ACTIVE_VIEWPORT::CORNERSTONE` context. This means that, if something fires the `Rotate` command, cornerstone's implementation of it will only be called/triggered if the active viewport is a cornerstone viewport.

- Hotkeys are configured in the application config
    - Default is `<root>/platform/viewer/public/config/default.js`
- Hotkeys are defined as key(s), and a command name

[...]

PR #1233, opened 2019-11-28
fix: Combined Hotkeys for special characters

rodrigolabs:
There's a bug when the user sets a special character hotkey with combined keys.

The Hotkey handler in `HotKeysPreferences.js` is generating:

`shift++` when we press `shift` + `=`, which naturally will generate the key `+`, but in that case, we must set only the key itself `+` in the settings, without the combined keys `shift`, `ctrl` or `alt`.


dannyrb:
This change doesn't appear to work on my computer. Are you testing on a `mac` or `windows`? Have we considered using `mousetrap`'s `record` extension? When I try to set `+`, `-`, `=`, or `_`, the value is an empty string.

- https://github.com/ccampbell/mousetrap/tree/master/plugins/record

See, using extensions: https://craig.is/killing/mice

It, unfortunately, is not published as an npm package. It's small, so we could either pull it in entirely, or create a package out of it.

platform/ui/src/components/userPreferencesForm/HotKeysPreferences.js
```diff
@@ -231,16 +231,31 @@ function HotKeyPreferencesRow({
     setInputValue(formatPressedKeys(hotkeys));
   }, [hotkeys]);
 
+  const isSpecialCharacter = str => {
+    return str.length === 1 && str.match(/^[^a-zA-Z0-9]+$/);
```
dannyrb: ^ - start of word
[^] - Negation (match anything not in set)
+ - Match preceding token

I didn't debug this to see what my value is here, just breaking it down for my own reference.

dannyrb:
Note: PR_CHECKS are failing
CC: @rodrigolabs

dannyrb:
Related: https://github.com/OHIF/Viewers/issues/1095

dannyrb:
image

dannyrb:
Hmmm... So `invert` and `rotate right` only fail if we've modified the settings, and not their bindings specifically, and then save. Oddly, rotate bindings now map to flips

dannyrb:
This is all a bit hairy. The `UserPreferencesForm` is managing state for all user preferences. We're passing everything down, and each hotkey field is determining if it's value is valid or not, and then we have state/props at each layer and pass back up the result.

We probably need to:
- Create a hotkeys context
- Have the Hotkey form manage validation
- Make fields as dumb as possible
- No need to force update userPreferencesModal; have it check for _any_ errors on the form to determine state
- It bubbles the save, and the `hotkeys` manager handles setting the new values

The crux of the issue was fixed, but I think we can handle this form in a much cleaner way. It's a bear to weed through it in it's current state. Might be worth leveraging the best practices set by formik here.

platform/core/src/classes/hotkeys/recordPlugin.js
```diff
@@ -111,7 +109,9 @@ export default function(Mousetrap) {
     _recordedSequence.push(_currentRecordedKeys);
     _currentRecordedKeys = [];
     _recordedCharacterKey = false;
-    _restartRecordTimer();
+    _finishRecording();
+    // comenting method below for the record to be more "real-time" w/o delay
```
igoroctaviano: Maybe add a TODO here?

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
igoroctaviano: I think we should check for effect dependencies, we should refactor state or the hook instead of disabling hook rules

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
```diff
@@ -112,6 +94,9 @@ function UserPreferencesForm({
     return Object.values(tabsError).reduce((acc, value) => acc || value);
   };
 
+  /**
+   *
```
igoroctaviano: Remove empty comment.

platform/core/src/classes/hotkeys/index.js
```diff
@@ -1,17 +1,16 @@
...
 
-// // // TODO: May need to bind these so Mousetrap = this in plugins;
-// pausePlugin(Mousetrap);
-// recordPlugin(Mousetrap);
-
-// console.log(Mousetrap);
-// console.log(Object.keys(Mousetrap));
+// Initialize record plugin
+// https://github.com/ccampbell/mousetrap/blob/master/plugins/record/mousetrap-record.js#L189-L201
+//
```
igoroctaviano: This comment is necessary? maybe adding a TODO if there's something to be done in the future?

platform/core/src/classes/hotkeys/recordPlugin.js
```diff
@@ -124,7 +124,7 @@ export default function(Mousetrap) {
    * @returns void
    */
   function _normalizeSequence(sequence) {
-    var i;
+    let i;
```
igoroctaviano: Move variable declaration inside for scope.

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
```diff
@@ -63,30 +40,35 @@ function UserPreferencesForm({
...
-      'Window Level': { windowLevelData },
-      General: { generalPreferences },
-    };
-  };
-
-  const [tabsState, setTabsState] = useState(
-    createTabsState(windowLevelData, hotkeyDefinitions, generalPreferences)
-  );
-
+  const [tabs, setTabs] = useState([
```
igoroctaviano: Looks like setTabs is not being used, this should be a component state?

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable react-hooks/exhaustive-deps */
+
 import React, { useState, useEffect } from 'react';
```
igoroctaviano: useEffect imported but never used.

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
```diff
@@ -120,16 +105,13 @@ function UserPreferencesForm({
       defaultHotKeyDefitions[commandName] = { ...values };
     });
 
-    // update local state
+    // Reset local state
     setTabsState({
       ...tabsState,
-      Hotkeys: { hotkeyDefinitions: defaultHotKeyDefitions },
+      Hotkeys: { hotkeyDefinitions: defaultHotKeyDefitions, hotkeyRecord },
```
igoroctaviano: `...otKeyDefitions ` maybe a typo here?

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
```diff
@@ -63,30 +40,35 @@ function UserPreferencesForm({
   onSave,
   onResetToDefaults,
```
igoroctaviano: onResetToDefaults is not being used inside the component, remove from here +  update proptypes.

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
igoroctaviano: In this case there are no hooks defined so this comment can be removed.

igoroctaviano:
> This is all a bit hairy. The `UserPreferencesForm` is managing state for all user preferences. We're passing everything down, and each hotkey field is determining if it's value is valid or not, and then we have state/props at each layer and pass back up the result.
> 
> We probably need to:
> 
> * Create a hotkeys context
> * Have the Hotkey form manage validation
> * Make fields as dumb as possible
> * No need to force update userPreferencesModal; have it check for _any_ errors on the form to determine state
> * It bubbles the save, and the `hotkeys` manager handles setting the new values
> 
> The crux of the issue was fixed, but I think we can handle this form in a much cleaner way. It's a bear to weed through it in it's current state. Might be worth leveraging the best practices set by formik here.

igoroctaviano:
`Uncaught TypeError: Cannot read property 'language' of undefined
    at onSave (ConnectedUserPreferencesForm.js?9106:29)
    at onSavePreferences (UserPreferencesForm.js?3563:136)
    at HTMLUnknownElement.callCallback (react-dom.development.js?5a61:336)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?5a61:385)
    at invokeGuardedCallback (react-dom.development.js?5a61:440)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js?5a61:454)
    at executeDispatch (react-dom.development.js?5a61:584)
    at executeDispatchesInOrder (react-dom.development.js?5a61:606)
    at executeDispatchesAndRelease (react-dom.development.js?5a61:713)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js?5a61:722)`

@dannyrb 
I found some issues:
- I cannot save, I'm getting errors `language of undefined` reported above. 
- When I'm switching from ctrl+s to ctrl+v I have to do it twice in order to update the form. 
- The error message for invalid combinations only appears only in onBlur.
- When i click to reset to defaults the form updates accordingly but if I close the modal and open it again the defaults are different.
- General tab on user preferences breaks the whole app.

I think we have to refactor the state of the whole user preferences flow, big part of those issues is related to complex state changes / manipulation.

galelis:
@igoroctaviano 
I will work in this PR today.

platform/core/src/classes/hotkeys/index.js
galelis: I've changed a bit the comments, It was probably added because the `MouseTrap` plugin is initialyzed by itself, so its just a headsup.

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
galelis: I've changed this and now its just a reference.

platform/ui/src/components/userPreferencesForm/UserPreferencesForm.js
galelis: Instead of removing it, I am keeping the functionality for now, added at the end of `onResetPreferences` and making sure it exists and its a function.

platform/viewer/src/components/Header/Header.js
```diff
@@ -24,6 +26,10 @@ function Header(props) {
 
   const [options, setOptions] = useState([]);
 
+  const setHotkeys = hotkeys => {
```
galelis: Somehow, when I pass the whole `hotkeysManage` instance into the modal `customProps`, the function lost the `hokeysManager` context (this)  and failed to set the hotkeys.  Any insight?

platform/core/src/utils/hotkeys/index.js
```diff
@@ -0,0 +1,19 @@
+// Only imported in environment w/ `window`
+// So we need to mock these for tests
```
igoroctaviano: Is this test comment still necessary?

platform/core/src/utils/hotkeys/recordPlugin.js
```diff
@@ -85,10 +85,8 @@ export default function(Mousetrap) {
    * @returns void
    */
   function _recordKey(key) {
-    var i;
-
     // one-off implementation of Array.indexOf, since IE6-9 don't support it
-    for (i = 0; i < _currentRecordedKeys.length; ++i) {
+    for (let i = 0; i < _currentRecordedKeys.length; ++i) {
```
igoroctaviano: 🎖

platform/ui/src/components/customForm/HotkeyField.js
```diff
@@ -0,0 +1,86 @@
...
+import PropTypes from 'prop-types';
+
+import { utils } from '@ohif/core';
+
+const { hotkeyRecord } = utils;
+
+/**
+ * Take the pressed key array and return the readable string for the keys
+ *
+ * @param {Array} [keys=[]]
```
igoroctaviano: My nitty-gritty self wants to say that it's nice to have @returns specified even when it's just void, my other self also forgets to add returns. 😆

platform/ui/src/components/customForm/HotkeyField.js
```diff
@@ -0,0 +1,86 @@
...
+        const keys = getKeys({ sequence, modifier_keys });
+
+        handleChange(keys);
+      });
+    }
+  };
+
+  return (
+    <input
+      readOnly={true}
```
igoroctaviano: Free tip ⭐️ : In React for truthy values, you can just pass readOnly prop instead of ={true}. (No need to change this code 😄 )

platform/ui/src/components/userPreferences/GeneralPreferences.js
```diff
@@ -0,0 +1,78 @@
...
+
+import i18n from '@ohif/i18n';
+
+import LanguageSwitcher from '../languageSwitcher';
+import { TabFooter } from './TabFooter';
+
+import './GeneralPreferences.styl';
+
+const languages = [
+  // TODO: list of available languages should come from i18n.options.resources
```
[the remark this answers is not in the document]
igoroctaviano: Yeah, we should have a way to retrieve this array of languages/labels from ohif/i18n. Every time this component renders, it should retrieve these values somewhere, this is the safest bet.

platform/ui/src/components/userPreferences/HotkeysPreferences.js
```diff
@@ -0,0 +1,183 @@
...
+  };
+
+  const hasErrors = Object.keys(state.errors).some(key => !!state.errors[key]);
+  const hasHotkeys = Object.keys(state.hotkeys).length;
+
+  const splitedHotkeys = [];
+  if (hasHotkeys) {
+    const arrayHotkeys = Object.entries(state.hotkeys);
+    const halfwayThrough = Math.ceil(arrayHotkeys.length / 2);
+    splitedHotkeys.push(arrayHotkeys.slice(0, halfwayThrough));
```
igoroctaviano: For blocks of code like this one which does some complicated stuff like entries -> ceil -> slice -> pushing, it's a good idea to extract into another well-named function which documents what it's doing. This helps a lot reading it in the future =)

platform/ui/src/components/userPreferences/WindowLevelPreferences.js
```diff
@@ -0,0 +1,161 @@
...
+//   };
+
+//   // tell parent to update its state
+//   useEffect(() => {
+//     onTabStateChanged(name, { windowLevelData: tabState });
+//   }, [tabState]);
+
+//   return (
+//     <table className="full-width">
+//       <thead>
```
igoroctaviano: Why this is all commented out? maybe we should add a TODO?

platform/viewer/src/components/Header/Header.js
igoroctaviano: I would avoid passing managers as props, managers are very app context-specific and should only be used inside its domain. If there's a need to pass the behaviour as a prop, use managers inside callbacks. This avoids modal to know too much. About losing this, you could explicitly bind before passing?

platform/core/src/utils/hotkeys/index.js
galelis: True, I'll remove it.

platform/ui/src/components/customForm/HotkeyField.js
galelis: Added :)

platform/ui/src/components/userPreferences/HotkeysPreferences.js
```diff
@@ -0,0 +1,183 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import { TabFooter } from './TabFooter';
+import { HotkeyField } from '../customForm';
+import { hotkeysValidators } from './hotkeysValidators';
+import { MODIFIER_KEYS, ALLOWED_KEYS } from './hotkeysConfig';
```
igoroctaviano: I think allowed_keys is not being used.

platform/ui/src/components/userPreferences/HotkeysPreferences.js
```diff
@@ -0,0 +1,183 @@
...
+ * HotkeysPreferences tab
+ * It renders all hotkeys displayed into columns/rows
+ *
+ * It stores current state and whenever it changes, component messages parent of new value (through function callback)
+ * @param {object} props component props
+ * @param {string} props.onClose
+ * @param {object} props.t
+ * @param {object} porps.hotkeyDefinitions
+ * @param {object} porps.hotkeyDefaults
+ * @param {object} porps.setHotkeys
```
igoroctaviano: porps, typo?

platform/ui/src/components/userPreferences/HotkeysPreferences.js
```diff
@@ -0,0 +1,183 @@
...
+ */
+function HotkeysPreferences({
+  onClose,
+  t,
```
igoroctaviano: Add proptypes for this component.

platform/ui/src/components/userPreferences/WindowLevelPreferences.js
```diff
@@ -0,0 +1,161 @@
+import React, { useEffect, useState } from 'react';
```
igoroctaviano: Imports not being used.

platform/ui/src/components/userPreferences/WindowLevelPreferences.js
galelis: I should finish it in another PR. I will remove the comments for now :)

platform/viewer/cypress/integration/common/OHIFUserPreferences.spec.js
```diff
@@ -386,20 +380,25 @@ describe('OHIF User Preferences', () => {
       cy.setLayout(3, 1);
       cy.waitViewportImageLoading();
 
-      // Rotate Right and Invert colors on Viewport #1
-      cy.get('body').type('RI');
-      // Check that image was rotated
-      cy.get('@viewportInfoMidTop').should('contains.text', 'R');
+      // Reset, Rotate Right and Invert colors on Viewport #1
+      cy.get('body').type(' ');
+      cy.get('body').type('r');
```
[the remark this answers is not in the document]
galelis: If you mean the argument of `type()`, it is always lowercase. Some other functions are checking for other information that is actually uppercased.

dannyrb:
Test fail, unable to build:

```
ERROR in /root/project/platform/ui/src/components/index.js
@ohif/******: Module not found: Error: Can't resolve './LanguageSwitcher' in '/root/project/platform/ui/src/components'
```

platform/ui/src/components/userPreferences/HotkeysPreferences.js
galelis: Change done.

platform/ui/src/components/userPreferences/GeneralPreferences.js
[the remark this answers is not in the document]
galelis: It does returns the values, but the values only contains the code (e.g. en, pt-BR, es ...) but not the description/language name

dannyrb:
### Prevent default behavior when focused on a hotkeys input

We may be trying to bind something that is a browser shortcut. We can still warn and prevent saving it, but should also make sure we're not accidentally triggering the shortcuts.

Example: Accidentally opening bookmarks on FF when trying to set `CTRL + B`

image

### Check against form hotkey data; not the initial value

See duplicate hotkey value for first field

image

### Prevent default application behavior for keys we allow to be bound (create new issue, follow-up in separate PR)

There are a few key combinations not on our ignore/block list that the browser still sees as hotkeys. This will happen because browser vendors implement different hotkeys, or the user may have custom ones for their browser. We should prevent all default hotkey functionality when dispatching our own.

platform/core/src/redux/reducers/preferences.js
```diff
@@ -1,12 +1,6 @@
 import cloneDeep from 'lodash.clonedeep';
 
 const defaultState = {
-  // First tab
-  hotkeyDefinitions: [
-    // commandName, label, keys
-    // [{ zoom: { label: 'Zoom', keys: ['z'] }}]
-  ],
-  // Second tab
```
dannyrb: Good call. We should be able to do this with windowLevelData soon as well

platform/ui/src/components/customForm/HotkeyField.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { utils } from '@ohif/core';
+
+const { hotkeys } = utils;
```
dannyrb: Maybe we make `hotkeys` a top level export so we don't need to be weird here.

platform/ui/src/components/customForm/HotkeyField.js
```diff
@@ -0,0 +1,91 @@
...
+  const onInputKeyDown = event => {
+    const { key = '' } = event;
+    const lowerCaseKey = key.toLowerCase();
+
+    // Prevent ESC key from propagating and closing the modal
+    if (lowerCaseKey === 'escape') {
+      event.stopPropagation();
+    }
+
+    event.stopPropagation();
```
dannyrb: If `lowerCaseKey` is escape, we run this, but we always run it?
Probably need to `stopPropagation` and `preventDefault` _always_?

platform/viewer/src/App.js
```diff
@@ -256,30 +255,27 @@ function _initExtensions(extensions, cornerstoneExtensionConfig, appConfig) {
...
-
-    if (updateStore) {
-      const { hotkeyDefinitions } = hotkeysManager;
-      const windowLevelData = {};
-      store.dispatch(
-        setUserPreferences({ windowLevelData, hotkeyDefinitions })
-      );
-    }
-  }
+  hotkeysManager.setDefaultHotKeys(appConfigHotkeys);
```
dannyrb: How do we protect against invalid configs / schemas here?
(Sanitize inputs)

platform/viewer/src/components/UserPreferences/GeneralPreferences.js
```diff
@@ -0,0 +1,83 @@
...
+
+import i18n from '@ohif/i18n';
+
+import { TabFooter, LanguageSwitcher, useSnackbarContext } from '@ohif/ui';
+import { useTranslation } from 'react-i18next';
+
+import './GeneralPreferences.styl';
+
+const languages = [
+  // TODO: list of available languages should come from i18n.options.resources
```
dannyrb: We have `i18n` here. Any reason we can't satisfy this todo stub?

galelis: So, the i18n have the list of language's codes but it does not have a map between language and its description e.g. `pt-BR <-> português`

platform/ui/src/components/customForm/HotkeyField.js
galelis: sorry, this was a left over from some tests I was doing. I'll remove it.

platform/viewer/src/components/UserPreferences/GeneralPreferences.js
dannyrb: Can we create that map for a few that we intend to support?

forEach i18n, use label as mapped friendly description, use value as i18n value?

platform/ui/src/components/customForm/HotkeyField.js
galelis: We can have as top level, but does make sense to have as top level if its just a util object?

platform/core/src/redux/reducers/preferences.js
galelis: WindowLevel is also done in another PR already, but the other PR is relying in this PR's changes as well.

platform/ui/src/components/customForm/HotkeyField.js
dannyrb: I'm not sure. I think exposing it as top level makes it easier to code split individual utils at a later date? Open to suggestion

galelis: Hotkeys are now exposed as root level. :)

platform/viewer/src/components/UserPreferences/GeneralPreferences.js
galelis: Code changes done. Now we have an exposed availableLanguages with naming on i18n that we can use.

galelis:
@dannyrb 
I have implemented the service request queue on UINotificationService, following your insights. 👍

platform/core/src/classes/HotkeysManager.js
```diff
@@ -45,38 +55,57 @@ export class HotkeysManager {
...
-
-    definitions.forEach(definition => this.registerHotkeys(definition));
+  setHotkeys(hotkeyDefinitions = []) {
+    try {
+      const definitions = this._getValidDefinitions(hotkeyDefinitions);
+
+      definitions.forEach(definition => this.registerHotkeys(definition));
+    } catch (error) {
+      const { UINotificationService } = this._servicesManager.services;
+      UINotificationService.show({
```
dannyrb: I think this is okay; may want to plan on the service _potentially_ not being available/defined

platform/core/src/services/UINotificationService/index.js
```diff
@@ -76,6 +82,11 @@ function setServiceImplementation({
   }
   if (showImplementation) {
     serviceImplementation._show = showImplementation;
+
+    while (serviceShowRequestQueue.length > 0) {
+      const showArguments = serviceShowRequestQueue.pop();
+      serviceImplementation._show(showArguments);
+    }
```
dannyrb: Let's create a separate test for this behavior? Maybe as a follow-up issue

dannyrb:
fixes https://github.com/OHIF/Viewers/issues/1439

PR #1235, opened 2019-11-29
fix: download tool fixes & improvements

### LINKED ISSUE #1217: Download/Snapshot Image tool: bugs and improvements

rodrigolabs:
This PR has a couple of fixes and improvements on the Download Tool.

**Issue 1: Minimum width and height not defined**

- When the user types an invalid value, it displays an error below the input:
_If the download form has any errors, the Download button will be disabled._

image

**Issue 2: Image is being resized while the user still typing the dimensions (width or height)**

- A `setTimeout` was added when the viewport size changes. Now it waits `1s` for the user to stop typing before performing a refresh in the image preview viewport.

viewport-refresh

**Issue 3: Empty file name**

- When the user types an empty filename, it displays an error below the input:

image

**Improvement 4: Width and Height are scaled proportionally**

- To make it easier, it was added a simple checkbox for the user to enable/disable aspect ratio.

image

---

Fixes #1217


platform/i18n/src/locales/pt-BR/UserPreferencesModal.json
```diff
@@ -1,6 +1,6 @@
 {
   "Cancel": "Cancelar",
-  "Reset to Defaults": "Restaurar Default",
+  "Reset to Defaults": "$t(Common:Reset) para Padrão",
```
dannyrb: I'm wary of mixing common words and phrases. Is there any risk of "common:reset" not making sense in this context? Is there a benefit to reusing the common definition here?

Curious if Locize spells out best practices here.

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
```diff
@@ -129,6 +142,46 @@ const ViewportDownloadForm = ({
...
 
+  useEffect(() => {
+    if (refreshViewport.current !== null) {
+      clearTimeout(refreshViewport.current);
+    }
+
+    refreshViewport.current = setTimeout(() => {
+      refreshViewport.current = null;
+      loadAndUpdateViewports();
+    }, REFRESH_VIEWPORT_TIMEOUT);
```
dannyrb: This is effectively a debounce. You could use our `useDebounce` hook? It lives in `ui/src/hooks`

dannyrb:
Some visual nitpicks. I realize these may fall outside of the exact criteria @mirnasilva laid out, but I think they're worth addressing while we're here.

image

- Textboxes should be aligned
- Group textboxes closer together to show that they're related fields

For reference, here is the legacy viewer:

image

The checkbox for "aspect ratio" instead of a toggle is fine for now, but maybe use verbiage: "maintain aspect ratio" or "lock aspect ratio" instead.

**Aspect Ratio**

This is an odd one. I would assume if I _uncheck_ aspect ratio, and set the values to:

- 500, 1000

Then recheck aspect ratio, that any changes would keep the `1:2` ratio. Instead, updates jump back to `1:1`.

**Width and Height**

I also encounter some issues when manually updating the values, but we could be limited to the native "number" input element's behavior.

My first attempt at updating was to backspace the field to empty, but it forces `0` as the minimum value. Is it possible to allow an empty field and show the minimum value error? That way I don't need to "highlight" or move the cursor in front of the `0` to set a value? Also, the fields should not "jump" as I'm updating values, regardless of the rendered image's size. See GIF:

RaLuY9CGqE

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
```diff
@@ -278,24 +352,32 @@ const ViewportDownloadForm = ({
...
           <button type="button" className="btn btn-danger" onClick={onClose}>
-            {t('Cancel')}
+            {t('Buttons:Cancel')}
           </button>
         </div>
         <div className="action-save">
-          <button onClick={downloadImage} className="btn btn-primary">
-            {t('Download')}
+          <button
+            disabled={Object.values(error).includes(true)}
```
dannyrb: Can we alias this higher up?
`hasErrors` or `isInvalid`?

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
```diff
@@ -204,19 +263,33 @@ const ViewportDownloadForm = ({
               min={minimumSize}
               max={maximumSize}
               value={width}
-              label={t('Image width (px)')}
+              label={t('imageWidth')}
               onChange={onWidthChange}
             />
+            {renderErrorHandler('width')}
```
dannyrb: This is cool beans for now. It might be worth creating a follow-up issue for form inputs that have a standard way of showing validation errors.

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
galelis: The useDebounce on our hooks is limited to update the provided state with a timeout, we cannot pass a function to be debaunced.

platform/i18n/src/locales/pt-BR/UserPreferencesModal.json
galelis: I've checked some posts online, to get an opinion about it, couldn't find any best practice related, but I would agree with you Danny, I am also wary about mixing common definitions this way. 

Its being used a lot in the i18n by the way.

galelis:
@dannyrb 
Its ready for review. 
Fixed all conflicts, add a bit of responsivite into the modal, remove some flickers once we update fields, add a better loading screen, fixed the keep aspect ratio as you commented and some other UI updates.

galelis:
@mirnasilva 
Most of the changes was UI updates, E2E tests are still working. The only new use case we have is that we can now work with the keep aspect setting it On/Off and getting the dimenstions set accordingly.

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
```diff
@@ -62,125 +76,231 @@ const ViewportDownloadForm = ({
...
+      fileType
+    );
+
+    setViewportPreview(state => ({
+      ...state,
+      src: dataUrl,
+      width: validSize(viewportElementWidth),
+      height: validSize(viewportElementHeight),
+    }));
+    /* eslint-disable-next-line react-hooks/exhaustive-deps */
```
igoroctaviano: Avoid at all costs to disable linting of hook rules, usually, this is the source of inconsistent components which calls effects constantly because of unrelated changes. If this is too big, maybe split?

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
```diff
@@ -62,125 +76,231 @@ const ViewportDownloadForm = ({
...
 
   /**
-   * @param {object} event - Input change event
-   * @param {string} dimension - "height" | "width"
-   */
+  * @param {object} event - Input change event
+  * @param {string} dimension - "height" | "width"
+  */
   const onDimensionsChange = (event, dimension) => {
+    const opositeDimension = dimension === 'height' ? 'width' : 'height';
```
igoroctaviano: A little typo? opposite  👀 (me being picky)

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
```diff
@@ -62,125 +76,231 @@ const ViewportDownloadForm = ({
...
+      dimensions.height,
     );
-  };
+
+    toggleAnnotations(showAnnotations, viewportElement);
+
+    const scaledDimensions = {
+      height: validSize(scaledHeight),
+      width: validSize(scaledWidth),
+    }
```
igoroctaviano: Not too important but maybe it's worth checking your linting config, its missing semicolon here, maybe its missing curly braces and another syntax piece throughout other code changes. It's easily fixable with VSCode autoSave too 👍

igoroctaviano:
> This PR has a couple of fixes and improvements on the Download Tool.
> 
> **Issue 1: Minimum width and height not defined**
> 
> * When the user types an invalid value, it displays an error below the input:
>   _If the download form has any errors, the Download button will be disabled._
> 
> image
> 
> **Issue 2: Image is being resized while the user still typing the dimensions (width or height)**
> 
> * A `setTimeout` was added when the viewport size changes. Now it waits `1s` for the user to stop typing before performing a refresh in the image preview viewport.
> 
> viewport-refresh
> 
> **Issue 3: Empty file name**
> 
> * When the user types an empty filename, it displays an error below the input:
> 
> image
> 
> **Improvement 4: Width and Height are scaled proportionally**
> 
> * To make it easier, it was added a simple checkbox for the user to enable/disable aspect ratio.
> 
> image
> 
> Fixes #1217

UX:
- The preview image doesn't increase/decrease in realtime when changing the dimensions incrementally using the select arrow like it used to be. The feel is that it's not changing until I cause onBlur. Maybe the debounce should only be done on keydown?

Other mentioned points were tested and LGTM. 👍

platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
galelis: Thanks for seeing that . :)

galelis:
>>>The preview image doesn't increase/decrease in realtime when changing the dimensions incrementally using the select arrow like it used to be. The feel is that it's not changing until I cause onBlur. Maybe the debounce should only be done on keydown?

We actually have a function to update the preview (`loadAndUpdateViewports()`) and the debaunce is in this function, not onBlur. We could change it: 
- make it without any debaunce, but I don't think its good idea as we render a canvas again. 
- make the timeout less than it is now, so it would update the viewport faster. 

@igor All other callouts are fixed.

igoroctaviano:
> > The preview image doesn't increase/decrease in realtime when changing the dimensions incrementally using the select arrow like it used to be. The feel is that it's not changing until I cause onBlur. Maybe the debounce should only be done on keydown?
> 
> We actually have a function to update the preview (`loadAndUpdateViewports()`) and the debaunce is in this function, not onBlur. We could change it:
> 
> * make it without any debaunce, but I don't think its good idea as we render a canvas again.
> * make the timeout less than it is now, so it would update the viewport faster.
> 
> @igor All other callouts are fixed.

Alright @galelis , i think thats fine, i was just wondering if we could use debounce only if the user is typing and no debounce if its just incrementing the selector.

dannyrb:
image

Issue #1268, opened 2019-12-09
Expose extension configuration and application configuration to extension modules


dannyrb:
RE: https://github.com/OHIF/Viewers/pull/1229

> Pass in rootConfig or applicationConfig to preInit/getModule.

Currently, only the `preRegistration` hook receives the extension's configuration as a parameter. Providing `getModuleFn`'s with the extension's configuration, and all lifecycle/modules with the application's configuration as `rootConfig` should open the doors to more configurable extensions.

PR #1279, opened 2019-12-11
feat: 🎸 Expose extension config to modules

igoroctaviano:
Currently, only the preRegistration hook receives the extension's
configuration as a parameter. Providing getModuleFn's with the
extension's configuration, and all lifecycle/modules with the
application's configuration as rootConfig should open the doors to more
configurable extensions.

Closes: #1268

### PR Checklist


dannyrb:
@igoroctaviano, do we need to update any docs along with these changes?

igoroctaviano:
> @igoroctaviano, do we need to update any docs along with these changes?

@dannyrb  I think so. But first i just wanna check if you if thats the approach we are gonna use because the app config can go through the extensions config or through the extension manager directly and then passed down to modules.

platform/core/src/extensions/ExtensionManager.js
```diff
@@ -21,15 +22,18 @@ export default class ExtensionManager {
...
+  registerExtensions(extensions, commonConfiguration = {}) {
     extensions.forEach(extension => {
       const hasConfiguration = Array.isArray(extension);
 
       if (hasConfiguration) {
         const [ohifExtension, configuration] = extension;
-        this.registerExtension(ohifExtension, configuration);
+        this.registerExtension(
+          ohifExtension,
+          merge({}, commonConfiguration, configuration)
```
dannyrb: I think, instead of merging these, we would want to keep them separate and explicit. Taking this approach leaves us open to accidental conflicts if:

- Extension author uses key: `apiKey`
- We update `@ohif/viewer` to have a new root config key of `apiKey`
- Extension author's key in configuration is accidentally blown out

`Vue.js` does an interesting thing with `vuex` where it's modules will receive `state` and `rootState`. It follows this pattern for other values as well. It allows you to grab/scope things locally, but also reference the entire "tree" or the higher up thing when needed.

We could follow a similar path where we pass some variation of: 
- `{ config, rootConfig }` or 
- `{ extConfig, appConfig }` or
- `{ config, appConfig }`

igoroctaviano: @dannyrb Thanks for the headsup. In my last commits configuration is now being passed through the extension manager though I had to instantiate extension/service managers right after the hydration/call of appConfig so we can have them available to inject. Maybe there's another better way to do it?

igoroctaviano:
> I think avoiding a deep merge of configurations is preferable here. Also, keeping them separate makes it easier to determine who is providing what and when. Please feel free to push back and discuss.

Updated. Removed appConfig from services as discussed. 🌟

PR #1290, opened 2019-12-12
ci: trial run of using custom test data source #1122

swederik:
For some reason my Git client refuses to check out the branch in https://github.com/OHIF/Viewers/pull/1122 (something weird with the naming?), so I'm closing that PR and opening this one. Sorry @dannyrb :P

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,11 +13,23 @@ version: 2.1
 ##
 orbs:
   codecov: codecov/codecov@1.0.5
-  cypress: cypress-io/cypress@1.13.0
+  cypress: cypress-io/cypress@1.26.0
+executors:
+  # Custom executor to override Cypress config
+  deploy-to-prod-executor:
+    docker:
+      - image: 'cypress/browsers:node14.15.0-chrome86-ff82'
+        environment:
+          CYPRESS_BASE_URL: https://ohif-staging.netlify.com/
+  chrome-and-pacs:
+    docker:
+      # Primary container image where all steps run.
+      - image: 'cypress/browsers:node14.15.0-chrome86-ff82'
+      - image: 'ohif/viewer-testdata:0.1-test'
 
 defaults: &defaults
   docker:
-    - image: circleci/node:12.9.1
+    - image: circleci/node:14.15.0
       environment:
         TERM: xterm # Enable colors in term
         QUICK_BUILD: true
@@ -170,7 +182,7 @@ jobs:
 
   DEPLOY_TO_DEV:
     docker:
-      - image: circleci/node:12.9.1
+      - image: circleci/node:14.15.0
         environment:
           TERM: xterm
           NETLIFY_SITE_ID: 32708787-c9b0-4634-b50f-7ca41952da77
@@ -185,7 +197,7 @@ jobs:
 
   DEPLOY_TO_STAGING:
     docker:
-      - image: circleci/node:12.9.1
+      - image: circleci/node:14.15.0
         environment:
           TERM: xterm
           NETLIFY_SITE_ID: c7502ae3-b150-493c-8422-05701e44a969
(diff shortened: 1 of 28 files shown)


dannyrb:
👢 💨 🤖

dannyrb:
@swederik, it's been a minute. What are the remaining blockers here?

swederik:
The blockers were mostly on the [dicomweb-server](https://github.com/dcmjs-org/dicomweb-server) side, but I think they are resolved now. The CI still seems to be failing because of the Python dicomweb-client. Even once we fix that, I guess it will fail because the datasets in the dicomweb-server version do not match what the Cypress tests expect, but that should be easy to fix. I don't know if the Github repo ([viewer-testdata](https://github.com/OHIF/viewer-testdata)) is the best place to put the data though (isn't there a limit?), maybe S3 would be better (or we can ask @fedorov and @pieper to store it in a Google Cloud bucket funded by IDC).

pieper:
> best place to put the data though (isn't there a limit?), maybe S3 would be better (or we can ask @fedorov and @pieper to store it in a Google Cloud bucket funded by IDC).

I definitely want to have a range of dicom test data made available for testing.  We are actively working on finding a way to make it free and easy to download for testing and demo purposes without opening up the possibility of unlimited networking (egress) charges.  Apparently this is an issue for a lot of people at NIH, so we may be getting some help from Google, but nothing concrete yet.

The other alternative I keep coming back to is [ipfs](ipfs.io).  It was rocky in the past but maybe is matured enough by now.

fedorov:
It is also possible to use GitHub releases to host data, and this is the solution that was officially recommended to me by GitHub support. There's also Dropbox.

What kind of data you need for testing? Can you select a subset from Google-hosted TCIA content and use that? How often do you update the test dataset in practice?

swederik:
@fedorov We don't update it very often, and we really don't need that much. It's going to be a small set of studies used to verify basic functionality. We can verify that most of the tools work on small studies. We might add more as we get to things like PET/CT or something. Dropbox is a good idea. I'll check the limit on Github releases, that's also a good suggestion.

fedorov:
@swederik I pinged you on Slack, I have details there about using github releases for data. Amazing, I was able to find that thread quite easily.

.circleci/config.yml
```diff
@@ -317,17 +329,24 @@ workflows:
       # E2E: PWA
       - cypress/run:
           name: 'E2E: PWA'
-          executor: cypress/browsers-chrome76
+          executor: with-chrome
```
dannyrb, line 332: This should probably be a specific version? Actually, probably a good thing to always test using latest browser/vendor in this instance.

.circleci/config.yml
```diff
@@ -317,17 +329,24 @@ workflows:
...
-          store_artifacts: false
+          record: true
+          store_artifacts: true
           working_directory: platform/viewer
-          build: npx cross-env QUICK_BUILD=true yarn run build
-          start: yarn run test:e2e:serve
+          build: |
+            # Build our client
+            npx cross-env QUICK_BUILD=true yarn run build
+          start: APP_CONFIG=config/dicomweb-server.js yarn run dev
```
dannyrb, line 349: Is this just you trying to test, because this is wonky.

.circleci/config.yml
```diff
@@ -348,13 +369,13 @@ workflows:
...
-          record: false
+          record: true
           store_artifacts: false
           working_directory: platform/viewer
-          build: npx cross-env QUICK_BUILD=true yarn run build:package
+          build: npx cross-env QUICK_BUILD=true APP_CONFIG=config/dicomweb-server.js yarn run build:package
           start: yarn run test:e2e:serve
           spec: 'cypress/integration/common/**/*,cypress/integration/script-tag/**/*'
-          wait-on: 'http://localhost:3000'
+          wait-on: 'http://localhost:3000/index.umd.js'
```
dannyrb, line 378: I'm not sure why this change is required.

swederik: it's not, I was debugging. You don't need to review anything at this time it's all garbage while I was trying to debug the test running issues, I'll squash it and remove unnecessary stuff.

.circleci/config.yml
swederik: yes

.circleci/config.yml
swederik: with-chrome is the executor, it has a specific version in it at the top of the file

Issue #1263, opened 2019-12-09
refactor ContextMenu service to allow for generic options and actions


dannyrb:
RE: https://github.com/OHIF/Viewers/pull/1229

> Make a more generic context menu with customizable options and actions when calling show.

Our `ContextMenu` service is currently hard coded as a `ContextMenu` service with cornerstone-tools options and actions. It should be reusable, and should probably work outside of the `Viewer` route and `Viewport` components.

If we scope functionality to `Viewport` components only for now, that would be a great win :+1:

Along with this change, a new `ContextMenu` service page should be added to the OHIF docs.

PR #1292, opened 2019-12-12
refactor: Remove LabelingFlow/ContextMenu services and reimplement on top of the Dialog service #1264

### LINKED ISSUE #1259: Improve / simplify UI providers usage.
Currently, the app has to wrap the generic UI provider to hydrate it will the callbacks/apps logic and then use this 'custom UI provider' in order to keep the app file usable. Also, each provider has to set its 

There are three things that need improvement:

1. Have a simpler way to have this provider hydrated instead of creating a brand new custom provider which extends/hydrates the one provided in the UI project.

2. Cleanup app usage of providers, group them by context (ui/non-ui/auth/store).

Good to have: Each provider has the same logic to set service implementation under a useEffect. We need to find ways to reuse this logic throughout all providers.

### LINKED ISSUE #1264: refactor remove IssueLabelingFlow service and reimplment on top of the Dialog service
RE: https://github.com/OHIF/Viewers/pull/1229

> Refactor LabellingFlow to DialogService with DialogContent passed from this extension. (Maybe LabellingFlow doesn't need to an independent provider).

The labeling flow and context menu behavior was, until very recently, tightly coupled to the `MeasurementsExtension`. Recent changes decoupled them, but only as much as needed to advance a separate initiative. This issue is a clean-up issue that simplifies our application and removes a service that would require maintenance.

The gist is, instead of creating an entirely new service for labeling flow, we can provide custom components to the existing `Dialog` service. Returned values from those custom components can continue to update state/ui.

This issue should be discussed before we dive in. We've just hammered out our first UI Services. Identifying how to handle situations like this isn't immediately obvious.

igoroctaviano:
Closes #1264 
Closes #1259

- ContextMenu and LabellingFlow services were removed along with its providers and custom providers. 
- The measurements panel extension is using the UIDialogService directly.
- Now instead of creating an entirely new service for labelling flow, users can provide custom components to the existing Dialog service. Returned values from those custom components can continue to update state/UI.

### PR Checklist


platform/ui/src/components/contextMenu/ContextMenu.js
```diff
@@ -0,0 +1,27 @@
...
+    </div>
+  );
+};
+
+ContextMenu.propTypes = {
+  items: PropTypes.array.isRequired,
+  onClick: PropTypes.func.isRequired,
+};
+
+export default ContextMenu;
```
dannyrb: Clean, small, and nice ^_^

platform/viewer/src/connectedComponents/ToolContextMenu.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { commandsManager } from './../App.js';
 
-import './ToolContextMenu.css';
+import { ContextMenu } from '@ohif/ui';
```
dannyrb: Do we want `ToolContextMenu` to appear on right click of annotations if the `MeasurementExtension` isn't registered? If yes, then we can keep it here. If not, then we should shift this component to the measurement's extension and pass in `CommandsManager` from a `getModuleFn`

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
```diff
@@ -46,17 +46,95 @@ export default function init({
...
+        ...props,
+      });
+    }
+  };
+
   const onRightClick = event => {
-    if (UIContextMenuService) {
-      UIContextMenuService.show({ event: event.detail });
+    if (UIDialogService) {
+      UIDialogService.dismiss({ id: 'context-menu' });
```
dannyrb: To reduce nesting, we could return early:

```js
if(!UIDialogService){
  console.warn('Unable to show tool context menu; no UI Dialog Service available.');
  return;
}
```

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
```diff
@@ -3,6 +3,9 @@ import cornerstone from 'cornerstone-core';
 import csTools from 'cornerstone-tools';
 import throttle from 'lodash.throttle';
 
+import LabellingFlow from '../../components/Labelling/LabellingFlow';
+import ToolContextMenu from '../../connectedComponents/ToolContextMenu';
```
dannyrb: Ah. Yes.
After we've completely decoupled the `MeasurementsExtension`, we'll want to pull it into it's own project, so these need to live closer.

Pretend `appExtensions/MeasurementsPanel` is a boundary we shouldn't cross for imports, unless using something like `@ohif/ui`

platform/viewer/src/appExtensions/MeasurementsPanel/index.js
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import ConnectedMeasurementTable from './ConnectedMeasurementTable.js';
 import init from './init.js';
 
+import LabellingFlow from '../../components/Labelling/LabellingFlow';
```
dannyrb: ^ Same thought here

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
igoroctaviano: Alright, its more informative too. Done! We can find a way to proxy all service calls to avoid this type of check, lets keep this in mind as well. Thanks.

platform/viewer/src/connectedComponents/ToolContextMenu.js
igoroctaviano: I think this task: https://github.com/OHIF/Viewers/issues/1263 can handle this right?

igoroctaviano:
> Holy cow. This looks so much better.
> I feel like it's easier to understand what's going on, it's less code, and I think it's more powerful.
> 
> How does the pattern feel?
> Anything in particular that's awkward about it?
> 
> Some very small changes requested (see comments), but overall this is an awesome PR. I was pretty stoked the entire time I was reading it. Great work 🎉

@dannyrb I think it's like you said, easier to maintain and very flexible. The usage complexity depends greatly upon the passed in component props and its callback dependencies... as we polish the components and its callbacks/dependencies we will simplify the service usage.

Something to keep in mind:
- We need to find a way to define error boundaries to the services so we can avoid check for services not defined. Maybe this could be something inside ServicesManager?

PR #1293, opened 2019-12-13
Chore/upgrade cypress version

dannyrb:
### PR Checklist

- [ ] Brief description of changes
- [ ] Links to any relevant issues
- [ ] Required status checks are passing
- [ ] User cases if changes impact the user's experience
- [ ] `@mention` a maintainer to request a review

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #1295, opened 2019-12-13
fix: allow empty values for dimensions

dannyrb:
We triggered two failing tests when we upgraded our version of Cypress. This PR attempts to address one of those failures. It allows for the `width` and `height` input fields to be empty.

Previously, they were `number` fields, and always had to have a value of at least `0`.

I tried not to stray too far from the task at hand, but I did consolidate a few things.

### User Cases:

- As a user, I expect to be able to clear an input field before I begin typing
  - Bad: `526` --> clear field --> `0` --> type "300" --> new value is `0300`
  - Good: `526` --> clear field --> "" --> type "300" --> new value is `300`
- As a developer, I want my tests to pass ^_^

### Notes:

- Will likely conflict with: https://github.com/OHIF/Viewers/pull/1235
  - This PR attempts to address other UX issues/bugs
- Related PR: https://github.com/OHIF/Viewers/pull/1293#issuecomment-565259817

### PR Checklist

  - The failures associated with this change appear to be resolved

### DIFF
--- a/platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
+++ b/platform/ui/src/components/content/viewportDownloadForm/ViewportDownloadForm.js
@@ -46,20 +46,20 @@ const ViewportDownloadForm = ({
 
   const [viewportElement, setViewportElement] = useState();
   const [viewportElementDimensions, setViewportElementDimensions] = useState({
-    width: minimumSize,
-    height: minimumSize,
+    width: defaultSize,
+    height: defaultSize,
   });
 
   const [downloadCanvas, setDownloadCanvas] = useState({
     ref: createRef(),
-    width: minimumSize,
-    height: minimumSize,
+    width: defaultSize,
+    height: defaultSize,
   });
 
   const [viewportPreview, setViewportPreview] = useState({
     src: null,
-    width: minimumSize,
-    height: minimumSize,
+    width: defaultSize,
+    height: defaultSize,
   });
 
   // Cornerstone's `enable/disable`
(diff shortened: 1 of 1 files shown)


dannyrb:
CC: @igoroctaviano for code review
CC: @mirnasilva for QA check

igoroctaviano:
Very good job @dannyrb ! 🌟 It's a lot cleaner as well.

I have a few points that maybe is worth fixing in another branch or this one:

- So, when the ViewportDownload first renders, the value in the width/height fields is 512 but the preview box doesn't match this value, it's very small and annotation very big, when I change to 513 or something causing a change, the preview increases properly to its correct size. Maybe we should start rendered with the correct value?

- Even if the user grows in equal pairs like 5x5 and 10x10, for values below 100 the image width is always bigger and the aspect is weird, maybe we should have a min size? to keep preview box consistent?

dannyrb:
👋 Thanks for the feedback, @igoroctaviano.

> So, when the ViewportDownload first renders, the value in the width/height fields is 512 but the preview box doesn't match this value, it's very small and annotation very big, when I change to 513 or something causing a change, the preview increases properly to its correct size. Maybe we should start rendered with the correct value?

This is a good catch. I'll patch this for this PR.

> Even if the user grows in equal pairs like 5x5 and 10x10, for values below 100 the image width is always bigger and the aspect is weird, maybe we should have a min size? to keep preview box consistent?

This fix will be bundled in a separate PR here: #1235

mirnasilva:
Hi @dannyrb , nice work! 🚀 
I just did a round of tests and we are good!
I'm not running the Visual Regression tests in this PR because we don't have any visual impact on the UX.
Moving this task to Done!
Thanks!

Issue #1242, opened 2019-12-03
Cine dialog remains on screen after return to study list


vicOnGit75:
Problem:
Cine dialog remains on screen after return to study list

Step to reproduce: 
-Clone 3rd December(git clone https://github.com/OHIF/Viewers.git)
-yarn install
-change config in default.js to       **enableStudyLazyLoad: false,**
-yarn start
-go to NAME_ONE ( actually any)
- open cine dialog
-return to study list
-see attached image

![CineRemains](https://user-images.githubusercontent.com/13784498/70083454-0f988700-1615-11ea-8f33-12235b25fd79.png)

dannyrb:
I'm able to reproduce here: https://viewer-dev.ohif.org/

This should be as simple as scoping the `DialogProvider` to the correct routes.

mirnasilva:
Hey @igoroctaviano, good job! I just did a round of tests and everything is working fine.
The issue was fixed and I'm moving it to Done!
Thanks!

PR #1301, opened 2019-12-16
fix: 🐛 Dismiss all dialogs if route change #1242

igoroctaviano:
Dismiss all dialogs if route changes. This prevents dialogs to stay open if the user goes to another page which doesn't have any dialogs available. 

This task hooks the dialog provider into application location state and observe changes to dismiss the dialogs.

Closes: #1242

User Cases
- As a user, I wanna all the dialogs to be dismissed if I'm in a different route.

### PR Checklist


igoroctaviano:
> There are a couple of situations where I could see this failing. Let me detail them out, and then we can discuss:
> 
> * The route changes when we add a second `study` to the viewing session
> * The route changes when we open a side panel
> * The route changes when we change hanging protocols
> * The route changes when we switch to a different workflow configuration
> 
> While none of the above exist today, I see them as items potentially slated for near future. Despite pushing a route change, we're still viewing the same container, but we may change some portion of the UI, fetch data, etc. The reason we would make these changes by updating the route is because we want to be able to restore the state by visiting the route directly, or by sharing the route.
> 
> A safer place to do this would be in `willUnmount` for the `viewer/` container route component. I'm struggling to think of a scenario where we'll want to persist dialogs between different top level route container components.

Thanks for the review @dannyrb . Code updated to dismiss in unmount of the viewer, makes more sense to me as well. In the future, we can even have multiple routes sharing the same toolbar state.

dannyrb:
@mirnasilva, I'm merging this but shifting to the awaiting QA so you can test, identify any changes in test cases (to verify behavior is not regressed), etc. We would hold off on prod deploy until I get your ✔️

mirnasilva:
Hi @dannyrb and @igoroctaviano , great job on this PR!

While I was testing, I found a minor bug with CINE dialog still being displayed if user drags an `SR` or `DOC` in the viewport (Issue #1321 ).
I believe this bug is not a blocker for this PR, so we should be good. :)
Thanks!

Issue #1124, opened 2019-10-29
Clicking on items in "Measurements "table only scrolls image stack once, on first click. Subsequent clicks have no effect.


ivan-aksamentov:
### Describe the Bug

Clicking on items in "Measurements "table only scrolls image stack once, on first click. Subsequent clicks have no effect.

### What steps can we follow to reproduce the bug?

1. Open, for example https://viewer.ohif.org/viewer/1.3.6.1.4.1.25403.345050719074.3824.20170125112931.11
2. Open "Measurements" side panel
3. Scroll image stack down to an arbitrary slice. Add an arbitrary measurement, for example "Length". Measurement entry should appear in the side panel
4. Scroll away from the slice where the measurement was added
5. Click on the measurement in the side panel. Image stack will jump to the slice with this measurement
6. Scroll away from this slice again
7. Click on the measurement entry in side panel. Nothing happens. 

### Expected behavior

Clicking on items in measurements table scroll to the relevant slice every time, no matter how many times stack was scrolled away.

Possibly related to #1007 #1098

PR #1306, opened 2019-12-16
fix: 🐛 #1124 Click on measurement should always move the viewport to its imageId

galelis:
### Changes
- We update redux state of `viewportSpecificData` with viewport's current `sopInstanceUid`

### User cases
As an user, I should be able to follow the steps bellow with success:
- Create any annotation supported by measurement table (e.g. length)
- Click in the measurement item in the table
- Scroll to another image after the viewport move to clicked measurement's image
- Select the same measurement item in the table
- Viewport should scroll to the measurement's image again

Closes: #1124 

### DIFF
--- a/extensions/cornerstone/package.json
+++ b/extensions/cornerstone/package.json
@@ -50,6 +50,6 @@
     "lodash.merge": "^4.6.2",
     "lodash.throttle": "^4.1.1",
     "query-string": "^6.8.3",
-    "react-cornerstone-viewport": "2.x.x"
+    "react-cornerstone-viewport": "^2.x.x"
   }
 }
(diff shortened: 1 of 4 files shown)


dannyrb:
@galelis this resolves the issue, but my computer has scrolling the stack as noticeably less snappy. We should either sit and discuss ways to accomplish without the performance hit (may require data/vm cleanup), or mark as `wontfix` for the time being.

galelis:
@dannyrb 
I've checked and I think I was able to see a bit of performance loss as well.. 
Checking the code, I realize that once we update the new image (onNewImage)
- we trigger the `setViewportSpecificData` that.. 
- triggers a change into state and.. 
- triggers an update into the OHIFCornerstoneViewport that after all checks update its state with new `sopInstanceId` and new `currentImageIdIndex` which.. 
- trigger a new update into the `cornerstoneViewport` that scrolls the viewport for the same image index it is

I did some changes locally, but couldn't find a good solution for this.

dannyrb:
We're going to take a look at alternative implementations

PR #1314, opened 2019-12-17
feat: 🎸 MeasurementService

igoroctaviano:
**Work in progress...**

### Requirements:
- [ ] Subscribe / unsubscribe to measurement changes (scoped by context ?)
- [ ] Timestamps (+ undo / redo ?)
- [ ] Measurement schema (that can be extended ?)
    - FrameOfReferenceUID, referenceSeriesUID, points[] (x, y, z), modifiedTimestamp (set internal) (int?)
- [ ] Measurement API to push or update new measurements
    - Optional initial ID / returns provided or generated id
- [ ] Maps and validate given annotations to measurement service format

### Blockers?
- Modify cornerstone-tools to add unique ids for annotations tool state

### Test Cases
- **Cornerstone Tools**: New annotation is created / updated / modified / removed -> Inform MeasurementService.
- **MeasurementsPanel Extension:**: Measurement changes -> observes / update for CRUD.

### User Cases
- As a developer, i want a new measurement service api to create and update measurements and have those changes broadcasted by context to each previously subscribed listener
- As a developer, I want to subscribe to measurement changes using a single measurement service api available through extension's module and preInit hook
- As a developer, I want to unsubscribe to measurement changes using a single measurement service api available through extension's module and preInit hook

### Usage

```js
preRegistration({ servicesManager }) {
    const { MeasurementService } = servicesManager.services;
    cornerstoneTools.addEventListener(cornerstoneTools.EVENTS.MEASUREMENT_ADDED, event => {
      const { mappedAnnotation, annotationId } = mapCornerstoneAnnotationToMeasurementServiceFormat(event.detail);
      MeasurementService.addOrUpdate({ data: { id: annotationId, ...mappedAnnotation } });
    });
}
```

```js
MeasurementService.subscribe(MeasurementService.getEvents().MEASUREMENT_ADDED, measurement => {
  const updatedMeasurement = measurement;
  /* if cs toolstate is stale */
  /* map data to cs tool state */
  /* update cs tool state */
});
```

### PR Checklist


platform/core/src/services/MeasurementService/MeasurementServiceAPI.js
```diff
@@ -0,0 +1,76 @@
...
+  constructor({ service }) {
+    this.measurements = {};
+    this.listeners = {};
+
+    if (service) {
+      service.setServiceImplementation({
+        addOrUpdate: this.addOrUpdate.bind(this),
+        subscribe: this.subscribe.bind(this),
+      });
+    }
```
dannyrb: We use `setServiceImplementation` in the UI Services because we can't know their implementation at time of instantiation. It also makes for a clearer "this is what you need to implement to fulfill this contract" as consumers seek to provide alternative implementations.

I'm unsure of if we need that separation for Measurements. I'll think on it a little more, and we can discuss this, but it may be simpler and may make more sense to provide the methods baked in. Swapping it out for a different implementation would mean matching the method signatures and using a different factory.

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Adds or update persisted measurements.
+ *
+ * @param {MeasurementSchema} schema { id, annotation }
+ */
+function _addOrUpdate({ id, annotation, broadcastEventName }) {
+  return serviceImplementation._addOrUpdate({
+    id,
+    annotation,
+    broadcastEventName,
```
dannyrb: When would a user supply a `broadcastEventName`?

platform/core/src/services/MeasurementService/MeasurementServiceAPI.js
```diff
@@ -0,0 +1,76 @@
...
+    let measurementId = id;
+    if (!measurementId) {
+      measurementId = Math.random()
+        .toString(36)
+        .substr(2, 5);
+
+      log.warn(
+        `Measurement ID not set. Using random string ID: ${measurementId}`
+      );
+    }
```
dannyrb: I think we have a `uuidv4()` utility lying around, or something similar for generating string GUIDs

platform/core/src/services/MeasurementService/MeasurementServiceAPI.js
```diff
@@ -0,0 +1,76 @@
...
+
+    const modifiedTimestamp = Math.floor(Date.now() / 1000);
+    this.measurements[measurementId] = { modifiedTimestamp, annotation };
+
+    /* Simple broadcast change */
+    if (Object.keys(this.listeners).length > 0) {
+      /* Should subscription event names be handled outside like here? Where to keep available subscription event names? */
+      this.listeners[broadcastEventName].forEach(callback => {
+        callback(this.measurements[measurementId]);
+      });
```
dannyrb: I think `MeasurementService` should probably maintain the list of events. It knows if a measurements was created/updated/deleted. It could also expose that enumeration.

dannyrb: Worth noting, today callbacks can't be removed, but we may want that ability later/soon. (think `add/removeEventListener`)

platform/core/src/services/MeasurementService/MeasurementServiceAPI.js
```diff
@@ -0,0 +1,76 @@
...
+   * @param {String} eventName
+   * @param {Function} callback
+   */
+  subscribe(eventName, callback) {
+    if (Array.isArray(this.listeners[eventName])) {
+      this.listeners[eventName].push(callback);
+    } else {
+      this.listeners[eventName] = [callback];
+    }
+  }
```
dannyrb: We can probably `warn` or `throw` if the `eventName` isn't one supported by `MeasurementService`

platform/core/src/services/MeasurementService/MeasurementServiceAPI.js
dannyrb: The `measurementId` is used as a key internally, but the broadcasted event wouldn't expose that measurementId? It might be worth including it in the returned object:

```js
{
  measurementId: '',
  measurementData: this.measurements[measurementId]
}
```

Or something similar

extensions/cornerstone/src/init.js
```diff
@@ -102,6 +102,32 @@ export default function init({ servicesManager, configuration }) {
     csTools.BrushTool,
   ];
 
+  /* Measurement Service Events */
+  cornerstone.events.addEventListener(
+    cornerstone.EVENTS.ELEMENT_ENABLED,
+    event => {
+      MeasurementService.subscribe(
+        /* Should subscription event names be handled outside like here? Where to keep available subscription event names? */
+        csTools.EVENTS.MEASUREMENT_ADDED,
```
dannyrb: This should probably be an event maintained by the MeasurementService?

```
MeasurementService.events.MEASUREMENT_CREATED
```

extensions/cornerstone/src/init.js
```diff
@@ -102,6 +102,32 @@ export default function init({ servicesManager, configuration }) {
...
+          console.log(
+            '[subscribe] Measurement added or updated',
+            updatedAnnotation
+          )
+      );
+      const element = event.detail.element;
+      element.addEventListener(csTools.EVENTS.MEASUREMENT_ADDED, event => {
+        console.log('[addOrUpdate] Adding new measurement...', event.detail);
+        MeasurementService.addOrUpdate({
+          id: 1,
```
dannyrb: If cornerstone-tools doesn't want to create the Id, we would leave this empty?

extensions/cornerstone/src/init.js
```diff
@@ -102,6 +102,32 @@ export default function init({ servicesManager, configuration }) {
...
+          annotation: event.detail,
+          /* Should subscription event names be handled outside like here? Where to keep available subscription event names? */
+          broadcastEventName: csTools.EVENTS.MEASUREMENT_ADDED,
```
dannyrb: I don't think we need to provide this. It might be worth telling the MeasurementService "who" is pushing the change (ie. cornerstone), then maybe we could ignore subscriptions/listeners for it because it should already have the most up-to-date info?)

extensions/cornerstone/src/init.js
igoroctaviano: I think we need the toolstate specific id so we can update this measurement, maybe we cannot even generate ids because the measurement state could be the same?

extensions/cornerstone/src/init.js
igoroctaviano: Yeah, that makes sense to me as well.

platform/core/src/services/MeasurementService/MeasurementServiceAPI.js
igoroctaviano: Updated.

platform/core/src/services/MeasurementService/MeasurementService.js
igoroctaviano: Updated.

platform/core/src/services/MeasurementService/MeasurementServiceAPI.js
igoroctaviano: Updated.

platform/core/src/services/MeasurementService/MeasurementServiceAPI.js
igoroctaviano: Updated.

platform/core/src/services/MeasurementService/MeasurementServiceAPI.js
igoroctaviano: Updated.

extensions/cornerstone/src/init.js
igoroctaviano: Updated.

extensions/cornerstone/src/init.js
igoroctaviano: Updated.

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,125 @@
+import log from '../../log';
+import guid from '../../utils/guid';
+
+const EVENTS = {
+  MEASUREMENT_UPDATED: 'event@measurement_updated',
+  MEASUREMENT_ADDED: 'event@measurement_added',
```
dannyrb: Bike shedding a little, but `event::measurement_added` may be slightly easier to read; if we're using a delimiter for "namespacing" / hierarchy

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,125 @@
...
+};
+
+const CONTEXTS = {
+  ALL: 'context@all',
+  VIEWER: 'context@viewer',
+  CORNERSTONE: 'context@cornerstone',
+};
```
dannyrb: These would likely be set by the providers/subscribers themselves. Unfortunately a little brittle to set these here. Fine for stubbing/poc

extensions/cornerstone/src/init.js
```diff
@@ -102,6 +102,25 @@ export default function init({ servicesManager, configuration }) {
     csTools.BrushTool,
   ];
 
+  /* Measurement Service Events */
+  cornerstone.events.addEventListener(
+    cornerstone.EVENTS.ELEMENT_ENABLED,
+    event => {
+      MeasurementService.subscribe(
+        MeasurementService.getEvents().MEASUREMENT_ADDED,
```
dannyrb: `getEvents()` because we expect something to register new events?

extensions/cornerstone/src/init.js
```diff
@@ -102,6 +102,25 @@ export default function init({ servicesManager, configuration }) {
...
+        measurement =>
+          console.log('[subscribe] Measurement added or updated', measurement)
+      );
+      event.detail.element.addEventListener(
+        csTools.EVENTS.MEASUREMENT_ADDED,
+        event => {
+          console.log('[addOrUpdate] Adding new measurement...', event.detail);
+          MeasurementService.addOrUpdate({ id: 1, data: event.detail });
```
dannyrb: This is where we need James' mapping magic before passing data. We've hit on the rough bits that need to be there.

extensions/cornerstone/src/init.js
igoroctaviano: Registering new events is possible in this branch but maybe it's not our case so I can remove it. Still, we have to find a good way to access the EVENTS constant or static prop because in this case, this passed in MeasurementService is an instance already.

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,137 @@
...
+      log.warn(`Measurement already defined. Updating measurement.`);
+      this.measurements[internalId] = measurement;
+      this._broadcastChange(internalId, EVENTS.MEASUREMENT_UPDATED);
+    } else {
+      log.warn(`Measurement added.`);
+      this.measurements[internalId] = measurement;
+      this._broadcastChange(internalId, EVENTS.MEASUREMENT_ADDED);
+    }
+
+    return measurement;
```
dannyrb: Probably okay to return _just_ the `measurementId` here. The `addOrUpdate` caller should have everything else.

extensions/cornerstone/src/init.js
```diff
@@ -106,6 +106,73 @@ export default function init({ servicesManager, configuration }) {
...
+        csTools.EVENTS.MEASUREMENT_ADDED,
+        event => {
+          console.log(
+            '[MEASUREMENT_ADDED] Adding new measurement...',
+            event.detail
+          );
+          const mappedMeasurement = mapCornerstoneAnnotationToMeasurementServiceFormat(
+            event.detail
+          );
+          MeasurementService.addOrUpdate({ data: mappedMeasurement });
```
dannyrb: Why the `data` key?

extensions/cornerstone/src/init.js
```diff
@@ -106,6 +106,73 @@ export default function init({ servicesManager, configuration }) {
...
+
+    const points = [];
+    points.push(measurementData.handles);
+
+    return {
+      imageId,
+      sopInstanceUid,
+      frameOfReferenceUID: frameOfReferenceUid,
+      referenceSeriesUID: seriesInstanceUid,
+      points, // points[] (x, y, z)
```
dannyrb: We use [x1, y1, x2, y2] when we have the sopInstancrUid.
We use [x1, y1, z1, x2, y2, z2] if we only have a frameOfReferenceUid

Also, note difference in UID casing in object.

We wouldn't use imageId here. Link to annotation's imageIdSpecificToolstate would be via `measurementId` or `sopInstanceUid`. If there's no easy way to do this today, we can expand cornerstone tool's api

dannyrb: This is where we would specify the `measurementId`. If we don't use something generated by cornerstone tools here, the alternative would be to have a way for us to `set` the ID for a given annotation (which we could do after `createOrUpdate` is called)

extensions/cornerstone/src/init.js
dannyrb: We may add a helper to the MeasurementService that allows you to grab any measurement in 3D or 2D coords. That is where we would use James' helper.

Or, if we don't want to mix the internal representation, we could store all points in xyz, but would need to key off sopInstanceUid being set to know that the original was 2D

igoroctaviano:
@dannyrb The tests were added and there's still example code to be cleaned up from init.js file. We are going to keep this branch opened to be the base for the next steps?

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,275 @@
...
+   *
+   * @return {Array} Array of objects
+   */
+  _arrayOfObjects = obj => {
+    return Object.entries(obj).map(e => ({ [e[0]]: e[1] }));
+  };
+}
+
+export default MeasurementService;
+export { EVENTS };
```
dannyrb: Top level API:

- subscribe(eventName, callback, context = 'all')
- addOrUpdate(measurement, context = 'all')
- registerEvent(eventName)
- getEvents()
- getMeasurement(id, context)
- getMeasurements(context = 'all')
- VALUE_TYPES

Separate:
- EVENTS

dannyrb: As events are emitted internally by the `MeasurementService`, I'm not sure when a newly registered event would be emitted or subscribed to as part of the `MeasurementService`'s normal operation.

The only dynamic event emitting I can think of, and we don't have this yet, would be scoping listening to one or more callers. IE. You only want to know if a measurement is modified by the VTK extension, or by any extension except for VTK. (No need to chase this down, just thinking out loud).

I'm not against a `registerEvent` and `getEvents`, but I'd want to know of some planned usage before we commit to them. If we don't have anything in mind, I'd rather we get rid of both and do a named export of EVENTS only.

dannyrb: It look like the song and dance for using this service is:

**Push:**

- Listen for library X's created/updated/deleted events
- Map measurement to `MeasurementService`'s format
- Call `MeasurementService.addOrUpdate` with a "Measurement" and a "context"
- If creating, use the returned ID so we can update/sync in the future

**Pull:**

- Subscribe to `MeasurementService` events (should we use addEventListener?)
- OnAdd (currently only log)
- OnUpdated
    - Use `measurement.source` as an alias for tool name
    - Use mapper's `toAnnotation` on known supported type?

dannyrb:
I've outlined a couple of discussion points above. Overall, I like the shape that the service is taking. The only other item that I think needs addressed is the matching/mapping logic. If possible, I'd like us to step away from hard coding it, and find a way to make it a part of the MeasurementService instead of it's own thing. Something like:

`MeasurementService.addMapping( matchCriteria, toAnnotation(), toMeasurement() );`

Where `matchCriteria` looks like:

```js
{
  valueType: POLYLINE,
  numPoints: 2
}
```

and
- `toAnnotation` takes a MeasurementService measurement and maps it the matched annotation
- `toMeasurement` takes a library annotation and maps it to the measurement service schema

Having the match criteria centralized could allow us to do match strength comparisons, or add the ability to transpose data across multiple matches in the future. It also allows us to define the criteria that comprise a measurement so we can more reliably expand that characteristics we can match on (and allow for multiple registrations for a single data type).

platform/core/src/services/MeasurementService/index.js
```diff
@@ -0,0 +1,8 @@
+import MeasurementService from './MeasurementService';
+
+export default {
+  name: 'MeasurementService',
+  create: ({ configuration = {} }) => {
+    return new MeasurementService();
+  },
+};
```
dannyrb: Ah, okay. We do this to adhere to the service pattern we created.
Almost makes me wonder if we need a simple `IService` interface that we "implement"
But I'm probably thinking too OOP

extensions/cornerstone/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.test.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  describe('_getPointsFromHandles()', () => {
+    it('converts handles to points', () => {
+      const convertedHandles = measurementServiceMappings._getPointsFromHandles(handles);
+      expect(convertedHandles).toEqual(points);
+    });
+  });
+
+  describe('_getHandlesFromPoints()', () => {
```
dannyrb: If this method is internal / not exposed, try to test it's behavior via the exposed method that uses it, not by calling it directly. This allows the implementation details to change

extensions/cornerstone/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.test.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  describe('toMeasurement()', () => {
+    it('map annotation to measurement service format', async () => {
+      const mappedAnnotation = await measurementServiceMappings.toMeasurement(eventData);
+      expect(mappedAnnotation).toEqual(measurement);
+    });
+  });
+
+  describe('_getPointsFromHandles()', () => {
```
dannyrb: Same here. The test may be important, but it should be done in terms of input/output from the method(s) that use this private one

extensions/cornerstone/src/init.js
```diff
@@ -32,9 +33,10 @@ cornerstone.metaData.addProvider(fallbackMetaDataProvider, -1);
  * @param {Object|Array} configuration.csToolsConfig
  */
 export default function init({ servicesManager, configuration }) {
```
dannyrb: Just to make init cleaner, we should shift these to two private methods:

```
_subscribeToMeasurementService()
_listenAndSyncCornerstoneMeasurements()
```

Or something similar, but with better names

extensions/cornerstone/src/init.js
```diff
@@ -107,6 +109,98 @@ export default function init({ servicesManager, configuration }) {
...
+        }
+      };
+
+      const addMeasurementServiceId = (id, eventData) => {
+        const { measurementData } = eventData;
+        Object.assign(measurementData, { _measurementServiceId: id });
+      };
+
+      event.detail.element.addEventListener(
+        csTools.EVENTS.MEASUREMENT_MODIFIED,
```
dannyrb: Are the MEASUREMENT_MODIFIED and MEASUREMENT_ADDED events really specific to the enabledElement? You'd think they are global

extensions/cornerstone/src/init.js
```diff
@@ -107,6 +109,98 @@ export default function init({ servicesManager, configuration }) {
...
+        event => {
+          console.log('[MEASUREMENT_MODIFIED] Updating measurement...', event.detail);
+          addOrUpdateMeasurement(event.detail);
```
dannyrb: If `event.detail` is the CornerstoneAnnotation, we could alias to make that more clear

dannyrb: If the listeners for these are similar, we could do:

```js
[csTools.events.MEASUREMENT_ADDED, csTools.events.MEASUREMENT_MODIFIED].forEach(csToolsEvtName => {
  event.detail.element.addEventListener(csToolsEvtName, { detail: csToolsAnnotation } => addOrUpdateMeasurement(csToolsAnnotation)); 
})
```

extensions/cornerstone/src/init.js
```diff
@@ -107,6 +109,98 @@ export default function init({ servicesManager, configuration }) {
...
+            console.log(
+              '[subscriber::MEASUREMENT_UPDATED] Measurement updated',
+              measurement
+            );
+            console.log('Mapped annotation to be saved:', annotation);
+          }
+        }
+      );
+
+      const addOrUpdateMeasurement = eventData => {
```
dannyrb: Does our `event` or `eventData` contain the tool's name/key? Note `'Length'` below

extensions/cornerstone/src/init.js
```diff
@@ -107,6 +109,98 @@ export default function init({ servicesManager, configuration }) {
...
+        try {
+          const measurementServiceId = addOrUpdate('Length', eventData);
```
dannyrb: Is the value returned  on create and update?

extensions/cornerstone/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.js
```diff
@@ -0,0 +1,130 @@
...
+        if (handles[handle].y) point.y = handles[handle].y;
+        points.push(point);
+      }
+    });
+    return points;
+  };
+
+  const _getHandlesFromPoints = points => {
+    return points
+      .map((p, i) => (i % 10 === 0 ? { start: p } : { end: p }))
```
dannyrb: I'm a little lost here

extensions/cornerstone/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.js
```diff
@@ -0,0 +1,130 @@
...
+      sopInstanceUID: sopInstanceUid,
+      frameOfReferenceUID: frameOfReferenceUid,
+      referenceSeriesUID: seriesInstanceUid,
+      label: measurementData.text,
+      description: measurementData.description,
+      unit: measurementData.unit,
+      area: measurementData.cachedStats && measurementData.cachedStats.area, /* TODO: Add concept names instead (descriptor) */
+      type: _getValueTypeFromToolType(toolType),
+      points: _getPointsFromHandles(measurementData.handles),
+    };
```
dannyrb: Are we tracking the "source" outside of the measurement? Or is that appended during the API call?

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    if (!version) {
+      log.warn('Source version not provided. Exiting early.');
+      return;
+    }
+
+    const id = guid();
+    const source = {
+      id,
```
dannyrb: Do we use this?

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,444 @@
...
+
+  /**
+   * Checks if given source is valid.
+   *
+   * @param {MeasurementSource} source Measurement source
+   * @return {boolean} Measurement source validation
+   */
+  _isValidSource(source) {
+    return source && this.sources[source.id];
+  }
```
dannyrb: Ah. Here! Perfect.

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,444 @@
...
+    };
+
+    if (Array.isArray(this.mappings[source.id])) {
+      this.mappings[source.id].push(mapping);
+    } else {
+      this.mappings[source.id] = [mapping];
+    }
+
+    log.warn(`New measurement mapping added to source '${this._getSourceInfo(source)}'.`);
+  }
```
dannyrb: Looks good :+1:

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,444 @@
...
+
+      console.warn(`Subscribing to '${eventName}'.`);
+      if (Array.isArray(this.listeners[eventName])) {
+        this.listeners[eventName].push(subscription);
+      } else {
+        this.listeners[eventName] = [subscription];
+      }
+
+      return {
+        unsubscribe: () => this._unsubscribe(eventName, listenerId),
```
dannyrb: Oh, neat. I wasn't aware this is how we made unsubscribe available.

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,444 @@
...
+      id,
+      name,
+      version,
+    };
+    source.addOrUpdate = (definition, measurement) => {
+      return this.addOrUpdate(source, definition, measurement);
+    };
+    source.getAnnotation = (definition, measurementId) => {
+      return this.getAnnotation(source, definition, measurementId);
+    };
```
dannyrb: If we have these methods on source, do we want to expose `getAnnotation` and `addOrUpdate` where we take `source` has a param, or make them private?

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,444 @@
...
+  getAnnotation(source, definition, measurementId) {
+    if (!this._isValidSource(source)) {
+      log.warn('Invalid source. Exiting early.');
+      return;
+    }
+
+    if (!definition) {
+      log.warn('No source definition provided. Exiting early.');
+      return;
+    }
```
dannyrb: Outside the scope of this PR, but might be worth allowing a "fallback" mapping that's used if there's no definition

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -0,0 +1,444 @@
...
+    );
+
+    const matchedCriteriaMapping = sourceMappingsByDefinition.find(
+      ({ matchingCriteria }) => {
+        return (
+          measurement.points &&
+          measurement.points.length === matchingCriteria.points
+        );
+      },
+    );
```
dannyrb: This is perfect for now. Curious to see how complex this gets as we contemplate adding things like... `isClosedShape`, angle between points, less than X points, more than X points, etc.

dannyrb: I had started looking at something like this: https://www.npmjs.com/package/@hapi/joi
The requirements feel similar to unit test asserts

extensions/cornerstone/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.js
igoroctaviano: appended during api call in addOrUpdate

extensions/cornerstone/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.js
igoroctaviano: This is just to get the handles back to cornerstone format instead of just array of points, we still don't have a working map for a specific tool yet, this is to use as an example of some sort of mapping we could do

extensions/cornerstone/src/init.js
igoroctaviano: In this example, the mapping functions used were generic mapping functions so they are global

platform/core/src/services/MeasurementService/index.js
igoroctaviano: The IService it's a good idea, at least it will make things like this more explicit and it will have a name/contract to be documented

extensions/cornerstone/src/init.js
igoroctaviano: Updated, thank you

extensions/cornerstone/src/init.js
igoroctaviano: Updated, definitely more descriptive, thanks.

extensions/cornerstone/src/init.js
igoroctaviano: Updated, thanks.

platform/core/src/services/MeasurementService/MeasurementService.js
igoroctaviano: Yeah, joi is really helpful asserting the scheme/data, i had used this before for mongo/node apps to validate params/body.

extensions/cornerstone/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.test.js
```diff
@@ -0,0 +1,80 @@
+import measurementServiceMappingsFactory from './measurementServiceMappingsFactory';
+
+jest.mock('cornerstone-core', () => ({
+  ...jest.requireActual('cornerstone-core'),
```
dannyrb: Ah. I've never actually seen this used. Nice.

platform/core/src/services/MeasurementService/MeasurementService.test.js
```diff
@@ -0,0 +1,391 @@
...
+
+  describe('addMapping()', () => {
+    it('adds new mapping', () => {
+      measurementService.addMapping(
+        source,
+        definition,
+        matchingCriteria,
+        toAnnotation,
+        toMeasurement
+      );
```
dannyrb: Is there a way for us to assert that the mapping was added? Do we expose mapping in any way from the `MeasurementSource`? Maybe be testing that a very simple mapping is used when retrieving a specific source's measurements?

platform/core/src/services/MeasurementService/MeasurementService.test.js
```diff
@@ -0,0 +1,391 @@
...
+    });
+
+    it('returns warning if no matching criteria provided', () => {
```
dannyrb: Semantics, but it doesn't return a warning, it logs one. AFAICT, this returns undefined?

platform/core/src/services/MeasurementService/MeasurementService.test.js
```diff
@@ -0,0 +1,391 @@
...
+      );
+      const measurementId = source.addOrUpdate(definition, annotation);
+      const mappedAnnotation = source.getAnnotation(definition, measurementId);
+
+      expect(annotation).toBe(mappedAnnotation);
+    });
+  });
+
+  describe('getMeasurements()', () => {
+    it('return all measurements', () => {
```
dannyrb: in what format?

platform/core/src/services/MeasurementService/MeasurementService.test.js
```diff
@@ -0,0 +1,391 @@
...
+
+      const measurements = measurementService.getMeasurements();
+
+      expect(measurements.length).toEqual(2);
+      expect(measurements.length).toEqual(2);
+    });
+  });
+
+  describe('getMeasurement()', () => {
+    it('return measurement with given id', () => {
```
dannyrb: in what format?

platform/core/src/services/MeasurementService/MeasurementService.test.js
```diff
@@ -0,0 +1,391 @@
...
+        definition,
+        matchingCriteria,
+        toAnnotation,
+        toMeasurement
+      );
+      const measurementId = source.addOrUpdate(definition, annotation);
+      const mappedAnnotation = source.getAnnotation(definition, measurementId);
+
+      expect(annotation).toBe(mappedAnnotation);
+    });
```
dannyrb: Do we have a "get" for when the source matches/doesn't match?

One uses matching criteria, the other should key off of source + definition because the stored value has a matching source.

platform/core/src/services/MeasurementService/MeasurementService.test.js
igoroctaviano: getMeasurement is measurement scheme, getAnnotation is source scheme

platform/core/src/services/MeasurementService/MeasurementService.test.js
igoroctaviano: Measurement scheme format, you mean i should specify?

platform/core/src/services/MeasurementService/MeasurementService.test.js
igoroctaviano: All early returns will return undefined in this case

platform/core/src/services/MeasurementService/MeasurementService.test.js
igoroctaviano: We can have a getter for mappings like getMappings our have mappings accessible through source, not sure if this is something we would need

About the test, there is one test which im adding a simple mapping to test getAnnotation functionality

platform/core/src/services/MeasurementService/MeasurementService.test.js
dannyrb: Just calling out that I sometimes use test descriptions as poor man's docs. Might be nice to be more explicit in test description, even if it's implied.

igoroctaviano:
> LGTM 👍
> 
> Need CI to pass, worth quick responses to comments, then good for merge.
> Any negative impact today to merging? I believe this mostly ships "dark" and gives us the ability to consume in next steps

No impacts, currently it only logs information about tools being added. 👍

Issue #1241, opened 2019-12-03
`2D MPR` button shows, even if the vtk extension is not registered


dannyrb:
See: https://github.com/OHIF/Viewers/issues/1240

### Currently

`@ohif/ui` has a component named `ToolbarButton`
`@ohif/viewer` has a component named `PluginSwitch` that wraps a `ToolbarButton`
`@ohif/viewer` wraps it with `ConnectedPluginSwitch`
This is used in `@ohif/viewer`'s `ToolbarRow.js`

I think it's named `PluginSwitch` because it's a "switch" to turn on/off a different "plugin". In this case, it toggles between the VTK Viewport and Cornerstone Viewport when certain criteria are met.

### Guidance

- Take a good look at `ConnectedPluginSwitch` and `PluginSwitch` to see what data it needs to function
- Identify which parts, if any, are not currently exposed to `ToolbarModule`s in extensions
- If parts are missing, we have a couple of ways to expand:
    - Expose top level redux keys (most brittle)
    - Assume access to a context provider
    - Expand required props the viewer must pass to custom toolbar button components
    - Open to other suggestions

### Acceptance Criteria
(these can be separate PRs)

- [x] ToolbarButton extension changes required to make functionality possible
- [x] Create a custom component button in the VTK extension's toolbar module
    - Remove `PluginSwitch` and `ConnectedPluginSwitch`
    - Create custom component with a better name than `PluginSwitch`
    - Preserve logic for when we show/hide the button (reconstructable datasets)
    - Preserve logic for toggling between `mpr` and `exit mpr`
- [x] Optionally, create a Toggle-able ToolbarButton as a new toolbar button type that can be used as an enumerated option in the extension/module
- [x] If ToolbarModule API was expanded, include notes in PR on how it changed
- [x] Update `VTK Extensions` readme with light documentation on `2D MPR` button

dannyrb:
If possible, clean up this bug when tackling the above issue: https://github.com/OHIF/Viewers/issues/1167

Issue #1167, opened 2019-11-12
VTK toolbar is not displayed if user clicks on 2D MPR button before image loads in viewport


mirnasilva:
### Description

If user drags a thumbnail and clicks on `2D MPR` button too fast, without waiting for the image to load on the single viewport the button `Exit 2D MPR` is displayed with the Cornerstone toolbar instead of the VTK toolbar.

This scenario is trick to reproduce manually, but it is constantly falling on the test automation.

### Steps to Reproduce

1. Load studies for patient name: Juno
2. Drag the 3rd thumbnail into the viewport
3. Before waiting the image to load, click on button `2D MPR`

**Steps to Reproduce on Test Automation:**
1. Go to file `platform/viewer/cypress/integration/pwa/OHIFExtensionVTK.spec.js`
2. Comment the command `cy.wait(1000)` in line 26.
3. On terminal, navigate to path: `Viewers/platform/viewer/`
4. Execute the test automation by running: `yarn run test:e2e`

### Actual Result

The button `Exit 2D MPR` is displayed with the Cornerstone toolbar instead of the VTK toolbar.

![Screen Shot 2019-11-12 at 17 58 37](https://user-images.githubusercontent.com/13245529/68710449-5810d780-0576-11ea-9a86-141a43e096ed.png)

### Expected Result

The button `Exit 2D MPR` should only be displayed when the VTK tools are active and the viewport layout split into 3 viewports.

dannyrb:
@galelis I don't believe this is fixed by #1241. I am still able to reproduce. You'll see an error in the console: `updateImage: image has not been loaded yet`

![image](https://user-images.githubusercontent.com/5797588/71293632-0f51f880-2344-11ea-9224-81285e0038e1.png)

The issue is, the button click updates state, but something is exiting early when we try to make the switch. Viewports show previous, but toolbar updates in response to the context change.

Not 100% sure why.

mirnasilva:
Hi @dannyrb and @galelis , great job! Thanks!
I did several tests and the bug mentioned in this issue is fixed. But I also found some others :)
The bugs found were described on #1326 .
I'm moving this issue to Done.

PR #1322, opened 2019-12-19
fix: 🐛 1241: Make Plugin switch part of ToolbarModule

galelis:
### Code Changes
- MPR buttons are now part of cornerstone toolbarModule
- `2D MPR` which has conditional behavior to show/hide has its own CustomComponent now
- `Exit 2D MPR` should only appear if `cornerstone extension` is registered 
- `2D MRP` should only appear if `vtk extension` is registered

### User cases
As an user:
- I shouldn't be able to see the `2D MPR` button if the opened displaySet into the activeViewport is not reconstructable
- I shouldn't be able to see the `2D MPR` button if `vtk extension` is not registered
- I shouldn't be able to see the `2D MPR` button if the active viewport contains a displaySet that is reconstructable
- I should be able to open a displaySet that is reconstructable, click in the `2D MPR` button and get the 2D multiplanar reconstruction done in the displaySet
- I should be able to see the `Exit 2D MPR` if I have the vtk toolbar

Closes: #1241 #1167 


extensions/cornerstone/src/commandsModule.js
```diff
@@ -348,6 +352,12 @@ const commandsModule = ({ servicesManager }) => {
       storeContexts: [],
       options: {},
     },
+    setCornerstoneLayout: {
+      commandFn: actions.setCornerstoneLayout,
+      storeContexts: ['viewports'],
+      options: {},
+      context: 'VIEWER',
+    },
```
dannyrb: This doesn't appear to need/use `storeContexts`

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -48,6 +47,9 @@ class ToolbarRow extends Component {
       activeButtons: [],
     };
 
+    // TODO: We should not need to add this here, we should be able to access extension manager throught getToolbaModule
+    this.extensionManager = extensionManager;
```
dannyrb: Can you remove this?

platform/core/src/extensions/ExtensionManager.js
```diff
@@ -117,6 +117,7 @@ export default class ExtensionManager {
       const extensionModule = getModuleFn({
         servicesManager: this._servicesManager,
         commandsManager: this._commandsManager,
+        extensionManager: this,
```
dannyrb: 👍

extensions/cornerstone/src/toolbarModule.js
```diff
@@ -34,212 +36,228 @@ const enums = {
...
+    },
+    {
+      id: '2DMPR',
+      label: '2D MPR',
+      icon: 'cube',
+      //
+      CustomComponent: VTKMPRToolbarComponent,
+      type: TOOLBAR_BUTTON_TYPES.COMMAND,
+      commandName: 'mpr2d',
+    }
```
dannyrb: Is this the only change here?

extensions/vtk/src/toolbarModule.js
```diff
@@ -5,119 +5,135 @@ const TOOLBAR_BUTTON_TYPES = {
...
+      ],
+    },
+    {
+      id: 'Exit2DMPR',
+      label: 'Exit 2D MPR',
+      icon: 'times',
+      //
+      type: TOOLBAR_BUTTON_TYPES.COMMAND,
+      commandName: 'setCornerstoneLayout',
+    }
```
dannyrb: I think the only other way to do this, would be to add this button in `cornerstone`'s extension, but set the context to `ACTIVE_VIEWPORT::VTK`. So the button would only show if VTK viewport was added; bypassing the whole "was this extension registered" conundrum

extensions/cornerstone/src/utils/setCornerstoneLayout.js
```diff
@@ -0,0 +1,22 @@
...
+const setCornerstoneLayout = () => {
+  const layout = {
+    numRows: 1,
+    numColumns: 1,
+    viewports: [{ plugin: 'cornerstone' }],
+  };
+
+  const action = setLayout(layout);
+
+  window.store.dispatch(action);
```
dannyrb: I think we're stuck doing this until we make it possible to control things like layout through a service. The only other option would be to make the store available to extensions.

I'm just hesitant to codify that.

extensions/cornerstone/src/toolbarComponents/VTKMPRToolbarComponent.js
```diff
@@ -0,0 +1,98 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
```
dannyrb: 👀

extensions/cornerstone/src/toolbarComponents/VTKMPRToolbarComponent.js
```diff
@@ -0,0 +1,98 @@
...
+import PropTypes from 'prop-types';
+import { ToolbarButton } from '@ohif/ui';
+
+let isVisible = true;
+
+const _isDisplaySetReconstructable = (studies, viewportSpecificData = {}, activeViewportIndex) => {
```
dannyrb: Where did the logic for this use to live?

platform/viewer/src/connectedComponents/ToolbarRow.js
galelis: Removed :)

extensions/cornerstone/src/commandsModule.js
galelis: Removed. 👍

extensions/vtk/src/toolbarModule.js
galelis: Done. 👍

extensions/cornerstone/src/toolbarComponents/VTKMPRToolbarComponent.js
galelis: This used to live into `PluginSwitch.js`

extensions/cornerstone/src/toolbarComponents/VTKMPRToolbarComponent.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+VTKMPRToolbarComponent.propTypes = {
+  parentContext: PropTypes.object.isRequired,
+  toolbarClickCallback: PropTypes.func.isRequired,
+  button: PropTypes.object.isRequired,
+  activeButtons: PropTypes.array.isRequired,
+  isActive: PropTypes.bool,
+  className: PropTypes.string,
+  extensionManager: PropTypes.object,
```
dannyrb: I'm okay with either of two different approaches here:

1. We do the same trick we did where the "Exit MPR" button is actually added by cornerstone, but we set it in the `VTK` button definitions for the cornerstone context

OR

2. We make this `shouldRender`, or something similar, and pass in true/false in the container ( a layer up) so the component doesn't depend directly on the `extensionManager`
2a. If we keep this, at a minimum, extensionManager should be required.

While we're touching this, can we rename it `VTKMPRToolbarButton` instead of `*Component`

galelis: I followed what we did for "Exit MPR" and removed all places where I was exposing extensionManager into getModuleFn(). 
It will make it clean.

dannyrb:
Moved to our review queue. This impacts the user's experience. Mostly through a change in toolbar button location

mirnasilva:
> * I shouldn't be able to see the `2D MPR` button if the active viewport contains a displaySet that is reconstructable

@galelis, I was checking the user cases and I got confused. I believe that in the following scenario the user should be able to see the `2D MPR` button, right?

As a user, I **should** be able to see the `2D MPR` button if the active viewport contains a displaySet that is reconstructable

mirnasilva:
Hi @dannyrb and @galelis , great job! Thanks!
I did several tests and the bug mentioned in this issue is fixed. But I also found some others 🐛😄 
The bugs found were described on #1326 .
I'm moving this PR to Done and we can tackle the other bugs in a brand new PR.

PR #1333, opened 2019-12-26
fix for correct read of LazyMode loader setting

vicOnGit75:
Problem identified :  change in configuration
window.config = {
...
    dicomWeb: [
      {
...
	    **enableStudyLazyLoad** from true to false,
still produces per series loading as if   enableStudyLazyLoad:  true.

change of comparison of sync/async mode for metadata retrieve
 !== to != fixes the issue.


swederik:
Thanks!

PR #1335, opened 2019-12-27
feat: 🎸 annotation tools (WIP)

rocammo:
#### Work in progress...

Changes done:
- [x] **Add annotations (feature)**
When drawing a new measurement with the annotation tools (i.e. RectangleRoiTool), it pop ups a modal window from which you can select the type of injury according to some categories: _malignancy_, _margins_, _density_ and _morphology_.
- [x] **Translations**
For the time being, the content of the annotation dialog has been translated into Spanish.
- [x] **Measurement Handler**
Handled EVENTS.MEASUREMENT_COMPLETED. 

Changes pending:
- [ ] **Edit annotations**
Dialog for annotations just fires once, only when the measurement is added. For instance, in case you have to edit an annotation, displaying a new entry in the context menu or clicking on the measurement may be possible solutions.
- [ ] **Data persistance**
Perhaps wait for DICOMSR to support RectangleRoi tooltype storage and attach the annotations data to it.

> This documentation has also been very useful:
[development/contributing.md](https://github.com/OHIF/Viewers/blob/master/docs/latest/development/contributing.md)<br/>[viewer/internationalization.md](https://github.com/OHIF/Viewers/blob/master/docs/latest/viewer/internationalization.md)


dannyrb:
@rocammo this is an interesting PR. What is the inspiration for the addition? Do you think this makes sense to build into the default viewer? Do you have any interest in maintaining this as an extension?

dannyrb:
HxUpgiLbZA

.webpack/webpack.base.js
```diff
@@ -81,8 +81,16 @@ module.exports = (env, argv, { SRC_DIR, DIST_DIR }) => {
...
+        'process.env.LOCIZE_PROJECTID': JSON.stringify(
+          process.env.LOCIZE_PROJECTID || ''
+        ),
+        'process.env.LOCIZE_API_KEY': JSON.stringify(
+          process.env.LOCIZE_API_KEY || ''
+        ),
+        /* Cadia */
+        'process.env.ANNOTATION_TYPE': JSON.stringify(
+          process.env.ANNOTATION_TYPE || 'MAMA'
+        ),
```
dannyrb: Would this be per instance of the OHIF Viewer, or would it make more sense to have it be contextual on a specific workflow, type of study, etc?

platform/ui/src/components/simpleDialog/SimpleDialog.js
```diff
@@ -47,6 +48,144 @@ class SimpleDialog extends Component {
     );
   };
 
+  static AnnotationMAMADialog = ({ onClose, onSubmit }) => {
```
dannyrb: These should probably have their own component files and leverage the UIDialogService:

https://docs.ohif.org/services/ui/ui-dialog-service.html

platform/viewer/src/appExtensions/MeasurementsPanel/init.js
```diff
@@ -70,6 +77,45 @@ export default function init({
...
+            .map(function(k) {
+              return value[k];
+            })
+            .join(',');
+          console.log(csv);
+          UIDialogService.dismiss({ id: dialogId });
+        },
+      },
+    });
+  };
```
dannyrb: Ah! You did use the DialogService. Nice!
This is an interesting approach. It does make sense that some measurements would require different options to edit their associated data/labels. Maybe there's a more dynamic way to achieve this?

It looks like we have a default dialog, and then we could associate certain measurement types with registered edit interfaces? We could set those associations when configuring and registering the MeasurementsPanel extension?

rocammo:
First of all, my apologies for the delay in responding. I have been assessing whether or not these additions to the project really add value to it.

As this is a specific solution for a very specific use case that is not going to have a continuity or is expected to be short, I think it does not. I am closing the pull-request for the moment, and depending on how it evolves perhaps it can be reopened.

Right now it is not possible for me to accept a request for commitment to the project. Anyway, thank you for your interest and feedback, and congratulations, since it is undoubtedly an interesting project in which I may be able to contribute in the future.

Issue #1326, opened 2019-12-24
VTK toolbar issues 


mirnasilva:
## Bug Report

The issues described here are related to #1324

### Issue 1: 2D MPR button is not displayed if user clicks on Study List and selects the same Patient

When user is viewing a Patient series that is reconstructable the `2D MPR` button will be displayed as expected. But if user decides to click on `Study List` link and then selects the same Patient, the `2d MPR` button will not be displayed anymore. If user reloads the page, the button will be displayed again.

Although I tested this in incognito mode, it looks like something related to cache. (not sure though)

**Steps to Reproduce:**
1. Search and select Patient Name: Juno
2. Drag-n-drop 3rd thumbnail into the viewport
3. Verify that button `2D MPR` is displayed as expected
4. Click on link `Study list`
5. Search and select Patient Name: Juno
6. Drag-n-drop 3rd thumbnail into the viewport
7. Verify that button `2D MPR` is **not** displayed

**Expected Result:**
1. Button `2D MPR` should always be displayed if the active viewport contains a displaySet that is reconstructable
2. If user navigates back and forth between study list page and study viewer, button `2D MPR` should always be displayed if the active viewport contains a displaySet that is reconstructable

![issue 1](https://user-images.githubusercontent.com/13245529/71385931-09207e00-25c9-11ea-8607-68dee4ba50bd.gif)

### Issue 2: User can navigate between non-existent viewports and VTK toolbar is displayed

If user exits the `2D MPR` mode and press hotkeys`ARROW RIGHT` and `LEFT`, he/she will be able to navigate between viewports that are not visible and should not exist. Once user navigates to a viewport different from the current one, the VTK toolbar is displayed but user cannot interact with it.

**Steps to Reproduce:**
1. Search and select Patient Name: Juno
2. Drag-n-drop 3rd thumbnail into the viewport
3. Click on `2D MPR`button
4. Click on `Exit 2D MPR` button
5. Press hotkey `ARROW RIGHT`
6. Verify that VTK toolbar is displayed
7. Verify that the current viewport is not set as active
8. Click on `Exit 2D MPR` button and verify that nothing happens
9. Verify that by pressing hotkeys `ARROW RIGHT` and `LEFT` user is able to navigate between viewports that are not displayed

**Expected Results:**
1. After exiting the `2D MPR` mode, user should not be able to navigate between the recently closed viewports by pressing hotkeys `ARROW RIGHT` and `LEFT`.
[...]

PR #1358, opened 2020-01-08
Remove trash data from redux storage after updates

galelis:
### Description
- After adding new viewports and removing then, some trash data is kept on state, in order to fix the issue of moving to ghost viewports by typing `ARROW LEFT`and `ARROW RIGHT` we need to clear the `viewportSpecificData` and also make sure that `activeViewportIndex` is pointing to an existent viewport. 

### User Cases
As an user: 
- I must always have a visible active viewport 
- I should see the toolbar relative to the viewport that is active

Closes: #1326 (Issues #2 and #4)


platform/core/src/redux/reducers/viewports.js
```diff
@@ -25,6 +25,40 @@ const DEFAULT_STATE = {
...
+ *  active viewport as default in case current one is not available anymore.
+ *
+ * @param {Object} action -
+ * @param {Number} action.numRows -
+ * @param {Number} action.numColumns -
+ * @param {Object} currentViewportSpecificData -
+ * @param {Number} currentActiveViewportIndex -
+ * @returns
+ */
+const handleViewportDeletions = (action, currentViewportSpecificData = {}, currentActiveViewportIndex) => {
```
dannyrb: `handle` is a good word for handling events, but this little guy is finding/determining our viewports given the number of viewports.

We should change its name, update the parameters to take the values it needs from `action` more explicitly, and update/add unit tests for the reducers that use this helper to assert that this expected behavior happens.

dannyrb:
Good work on this one Gustavo!

PR #1359, opened 2020-01-08
Fix active tool icon

galelis:
### Issue
OHIF toolbar module does not have implemented a feature to make sure the default active tool will have it's icon active in the toolbar row. 

### Description of the changes
TBD

### User cases
As an user: 
- I should be able to open a study in the viewport and see the current active tool button active in the tool bar (e.g. VTKViewport has Crosshairs tool active by default)

Closes: #1326 (Issue #3)


galelis:
I've discussed with Danny and another proposed implementation would be creating a function on each extension toolbarModule to expose the proper way to get the current active tool into the activeViewport.  

Example of a function in cornerstone extension:

```
const getActiveTool = () => {
  const toolStates = csTools.store.state.tools;
  const filterActiveTool = (tool) => {
    const isActive = tool.mode === "active";
    const isMouseButton1Tool = tool._options && tool._options.mouseButtonMask && tool._options.mouseButtonMask.includes(1);

    return isActive && isMouseButton1Tool;
  }

  return toolStates.filter(filterActiveTool)
}
```

Similarly for VTK. 
I didn't get an actual working function to get the active tool (interactorStyle) into a VTK viewport. 

Teorically, we need to get any active vtk api, look for its renderWindow and get its interactorStyle (e.g. `vtkInteractorStyleMPRCrosshairs`)  
One issue here is that the `interactorStyle` class name does not match with tool id that we have in toolbarModule, so we would need to create a map somehow or maybe James can help with a better solution.

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -88,13 +88,20 @@ class ToolbarRow extends Component {
   }
 
   componentDidUpdate(prevProps) {
+    // TODO: This does not compare the arrays correctly
```
igoroctaviano: Not sure if I get it, maybe we should add why it's not correct or maybe raise an issue to fix this? what do you think?

platform/viewer/src/connectedComponents/ToolbarRow.js
```diff
@@ -141,6 +148,24 @@ class ToolbarRow extends Component {
   }
 }
 
+function _getDefaultActiveButtons(toolbarButtons) {
+  const activeButtons = [];
+
+  const defaultActiveButton = toolbarButtons.find(button => button.options && button.options.defaultActive);
+
+  if (defaultActiveButton)
```
igoroctaviano: Avoid open conditionals.

igoroctaviano:
@galelis the default works now properly, just some tiny comments. Good job! 👍

platform/viewer/src/connectedComponents/ToolbarRow.js
galelis: Fixed

platform/viewer/src/connectedComponents/ToolbarRow.js
galelis: Arrays can not be compared just by doing: `ArrayA <conditional> ArrayB`
This was the root cause I've added the comment. 

I rechecked the issue, debugged for a bit and figured out the issue why it was not working properly if I use the correct way of campare arrays (I've created a function in this PR to do it). 

This PR is dependent on anoter PR now #1360 
We need to get this (#1360) merged first, and then merge this one because E2E tests will break.

PR #1360, opened 2020-01-09
fix: 🐛 Metadata is being mistakenly purged

galelis:
### Description
StudyMetadata is being mistakenly purged once we load a study into the viewer and blocking the 2D MPR button component to check if the current study/series is reconstructable.

### User cases
As an user: 
- I should be able to open a study twice in a row* and be able to see the 2D MPR button everytime it is required.

It should fix the issue number 1 reported on #1326 :

> Issue 1: 2D MPR button is not displayed if user clicks on Study List and selects the same Patient
> When user is viewing a Patient series that is reconstructable the 2D MPR button will be displayed as expected. But if user decides to click on Study List link and then selects the same Patient, the 2d MPR button will not be displayed anymore. If user reloads the page, the button will be displayed again.
> 
> Although I tested this in incognito mode, it looks like something related to cache. (not sure though)
> 
> **Steps to Reproduce:**
> 
> - Search and select Patient Name: Juno
> - Drag-n-drop 3rd thumbnail into the viewport
> - Verify that button 2D MPR is displayed as expected
> - Click on link Study list
> - Search and select Patient Name: Juno
> - Drag-n-drop 3rd thumbnail into the viewport
>  -Verify that button 2D MPR is not displayed 

Closes: #1326 (Issue #1)

### DIFF
--- a/platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
+++ b/platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
@@ -1,5 +1,6 @@
 import React, { useState, useEffect, useContext } from 'react';
 import { metadata, studies, utils, log } from '@ohif/core';
+import usePrevious from '../customHooks/usePrevious';
 
 import ConnectedViewer from './ConnectedViewer.js';
 import PropTypes from 'prop-types';
@@ -128,9 +129,9 @@ const _addSeriesToStudy = (studyMetadata, series) => {
 const _updateMetaDataManager = (study, studyMetadata, series) => {
   updateMetaDataManager(study, series);
 
-  const { studyInstanceUID } = study;
+  const { studyInstanceUid } = study;
 
-  if (!studyMetadataManager.get(studyInstanceUID)) {
+  if (!studyMetadataManager.get(studyInstanceUid)) {
     studyMetadataManager.add(studyMetadata);
   }
 };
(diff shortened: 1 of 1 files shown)


galelis:
@dannyrb 
Another thing I noticed, is that, new `studyInstanceUids` can change but some of the studies could still be present in this case we could only purge the non-used ones, then instead of a purge, we could do something like: 

```
const removeUnusedMetadata = () => {
    const metadatas = studyMetadataManager.all();
    metadatas.forEach(metadata => {
      if (!studyInstanceUids.includes(metadata.studyMetadataUID)) {
        studyMetadataManager.remove(metadata.studyMetadataUID);
      }
    });
  }
```

And do something similar to the `purgeCancellablePromises` to only cancel the ones that we are not going to use anymore.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -314,9 +315,15 @@ function ViewerRetrieveStudyData({
     }
   };
 
+  const prevStudyInstanceUids = usePrevious(studyInstanceUids);
+
   useEffect(() => {
-    studyMetadataManager.purge();
-    purgeCancellablePromises();
+    const hasStudyInstanceUidsChanged = !(prevStudyInstanceUids && prevStudyInstanceUids.every(e => studyInstanceUids.includes(e)));
+
```
igoroctaviano: This useffect is missing prevStudyInstanceUids and purgeCancellablePromises in its deps array.  You can add them or remove the array.

igoroctaviano:
@galelis I have tested and it works, good job! 🏅 just left one microscopic comment about useEffects.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
galelis: The hook should not depend on change of `prevStudyInstanceUids` and 
 `purgeCancellablePromises`, its only watching changes at the variable `studyInstanceUids`.

PR #1367, opened 2020-01-13
Custom Healthcare API endpoint

mukhlin:
Faster data access times for performance critical applications (i.e. primary radiologist reads) is requested by Google and it involves usage of cache that can sit transparently in front of the Cloud Healthcare API.
Currently OHIF does not allow to configure custom Cloud Healthcare API endpoint (hardcoded endpoint is https://healthcare.googleapis.com/v1beta1), so it makes impossible to use caching server.
This change adds such feature.

### DIFF
--- a/platform/viewer/src/googleCloud/api/GoogleCloudApi.js
+++ b/platform/viewer/src/googleCloud/api/GoogleCloudApi.js
@@ -15,7 +15,11 @@ class GoogleCloudApi {
   }
 
   get urlBase() {
-    return window.config.healthcareApiEndpoint || 'https://healthcare.googleapis.com/v1beta1';
+    return this.healthcareApiEndpoint || 'https://healthcare.googleapis.com/v1beta1';
+  }
+
+  set urlBase(url) {
+    this.healthcareApiEndpoint = url;
   }
 
   get urlBaseProject() {
--- a/platform/viewer/src/googleCloud/api/GoogleCloudApi.js
+++ b/platform/viewer/src/googleCloud/api/GoogleCloudApi.js
@@ -15,7 +15,7 @@ class GoogleCloudApi {
   }
 
   get urlBase() {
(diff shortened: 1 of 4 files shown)


platform/viewer/src/googleCloud/api/GoogleCloudApi.js
```diff
@@ -15,7 +15,7 @@ class GoogleCloudApi {
   }
 
   get urlBase() {
-    return `https://healthcare.googleapis.com/v1beta1`;
+    return window.config.healthcareApiEndpoint;
```
swederik: Can you add || 'https://healthcare.googleapis.com/v1beta1' here?

Thanks!

dannyrb: This expands our appContext and configuration. There may be a cleaer way to pull this value after the app's been kickstarted.

mukhlin: > Can you add || 'https://healthcare.googleapis.com/v1beta1' here?
Done.

mukhlin: > This expands our appContext and configuration. There may be a cleaer way to pull this value after the app's been kickstarted.

Probably `getServers()` in the `platform/viewer/src/customHooks/useServer.js` is the most suitable place. What do you think?

swederik:
LGTM, not sure why CI is failing though

mukhlin:
> LGTM, not sure why CI is failing though

Yesterday it failed first time. When I pushed one more change, CI passed, but my change is not related to the failed tests. I tried to run E2E tests locally and another tests failed. It looks like timeout issue when tests try to wait for some HTML elements. The problematic file is `pwa/OHIFSaveMeasurements.spec.js`.

Issue #1391, opened 2020-01-24
Segmentations Settings UI


JamesAPetts:
Ongoing work on the segmentation UI. For this issue we should address global segmentation settings.. This ticket is addressing part of [this comment](https://github.com/OHIF/Viewers/issues/1370#issuecomment-576729840):
>It would be nice to have buttons that control segment visibility (overlay, contour, both, opacity), similar to what we have in Slicer (see you already have plans for this in the figma sketch).

The per-segment settings (color/alpha/visibility) part of the request is not in the scope of this issue, and will be implemented later.

[This Figma template](https://www.figma.com/file/0CP9cmhHhDC8gQX5R9zvgo/OHIF-segmentation-side-panel?node-id=0%3A1) shows how the UI should look.

- Add a gear icon to the top right of the segmentation panel.
- Clicking a Gear icon changes the contents of the side panel to be a settings menu.
- Render segment outline: Checkbox, default Checked
  - When checked, a horizontal slider appears for Outline Opacity with values 0 - 1.0 increments of 0.05
  -  When checked a horizontal slider appears for Outline Width with values 1-5, increment 1 px.
- Render segment fill: Checkbox, default Checked
  - When checked, a horizontal slider appears for Fill Opacity with values 0 - 1.0, increments of 0.05

- Render inactive segmentations: Default off
  - When checked a horizontal slider appears for Opacity of inactive segmentations with values 0 - 1.0 increments of 0.05.

Clicking the back button should go back to the segmentation menu. When the segmentation menu is hidden and then reopened, the view should default to the segmentation menu, not the settings menu.

The UI should alter these settings which are consumed by the viewer components: https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/store/modules/segmentationModule/defaultConfiguration.js

salimkanoun:
Excellent, excellent, excellent !
Can't wait to work with you on PET/CT segmentation, when OHIF will render fused images.

fedorov:
I see #1392 has been merged, and IDC sandbox is at the latest 3.8.0 release of OHIF Viewer, but it does not work for me. 

Here are the issues:

1) maybe this is expected, but if I open TCIA QIN-HEADNECK collection, I do not see a single study that shows SEG in the Modality column; if I search for SEG in the Modality column, I get 0 studies
[...]

PR #1392, opened 2020-01-24
Segmentations Settings UI - Phase 1 #1391

igoroctaviano:
### PR Checklist

#1391


JamesAPetts:
Hey Igor,

A few comments:

The `Segment Outline` `Width` should only go from 1 to 5, right now you can make your segments funky by allowing it go higher.

In the `Render inactive segmentations` panel, when checked:

The options  `Fill Opacity` and `Outline Opacity` in this menu should only be shown when the regular `Render Fill` and `Render Outline` options are on, as otherwise they have no effect.

Also the entire `Render Inactive Segmentations` drop down should probably be hidden when both `Render Fill` and `Render Outline` are switched off, as again, then this option has no effect.

igoroctaviano:
@JamesAPetts All requested changes in the first branch were pushed here along with the removal of the unused tools you mentioned. The only thing missing is performance improvement.

JamesAPetts:
Hey @igoroctaviano , could you remove the growcut segmentation tool from the codebase? We don't want it in here for now.

Also sorry if you ended up with the wrong info, could you remove the segmentation creation tools entirely? Those 4 we will be bring back at a later date when we have STOW for SEG.

extensions/dicom-seg/src/OHIFDicomSegSopClassHandler.js
```diff
@@ -0,0 +1,68 @@
...
+      isDerived: true,
+      referencedDisplaySetUid: null, // Assigned when loaded.
+      labelmapIndex: null, // Assigned when loaded.
+      isLoaded: false,
+      seriesDate,
+      seriesTime,
+      seriesDescription,
+    };
+
+    segDisplaySet.load = function(referencedDisplaySet, studies) {
```
JamesAPetts: @dannyrb , are you ok with this hitting master? Its a bit of a funny pattern. If you guys can think of a better way to 

What we need is to have a custom load process that can be triggered on demand when you search for a display set, I can't think of a better way to do this off the top of my head. Are we happy with `load`, as I guess this is becoming cemented as api?

extensions/dicom-seg/src/commandsModule.js
```diff
@@ -0,0 +1,21 @@
+// "actions" doesn't really mean anything
+// these are basically ambigous sets of implementation(s)
+const actions = {
+  doGoofyStuff: () => {
```
JamesAPetts: I'm 110% sure this is placeholder that can be removed.

extensions/dicom-seg/src/commandsModule.js
```diff
@@ -0,0 +1,21 @@
...
+    console.log('~~ GOOFY');
+  },
+};
+
+const definitions = {
+  doGoofyStuff: {
```
JamesAPetts: And this.

extensions/dicom-seg/src/components/BrushRadius.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Range } from '@ohif/ui';
+
+const BrushRadius = ({ value, onChange }) => (
+  <div className="brush-radius">
+    <label htmlFor="brush-radius">Brush Radius</label>
+    <Range
+      value={value}
+      min={1}
```
JamesAPetts: cornerstoneTools has a min and max brush radius, so we should read it from there as a consumer may set it up differently.

```js
const {configuration} = cornerstoneTools.getModule('segmentation');
const {minRadius, maxRadius} = configuration;
```

@igoroctaviano  @dannyrb, This doesn't have to go in now, but perhaps we should should wrap get/set interactions with the cornerstoneTools store with context to make these kind of things less blurry?

We don't have to worry about this right now, as we will add the brush radius slider back in when we add creation functionality.

extensions/dicom-seg/src/components/SegmentItem.js
```diff
@@ -0,0 +1,52 @@
...
+        <button
+          className="btnAction"
+          onClick={() => console.log('Relabelling...')}
+        >
+          <span style={{ marginRight: '4px' }}>
+            <Icon name="edit" width="14px" height="14px" />
+          </span>
+          Relabel
+          </button>
+        <button
```
JamesAPetts: Can we hide the relabel and description UI for now? We can add these in when we have create, this PR should be very "read only" in style.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,466 @@
...
+  SegmentationSelect,
+} from './index';
+
+import './SegmentationPanel.css';
+import SegmentationSettings from './SegmentationSettings';
+
+const { studyMetadataManager } = utils;
+
+const segmentationModule = cornerstoneTools.getModule('segmentation');
+const DEFAULT_BRUSH_RADIUS = segmentationModule.getters.radius || 10;
```
JamesAPetts: `segmentationModule.getters` is actually an object full of functions.

This is returning undefined as you should grab the radius from the configuration:

```js
const {configuration} = cornerstoneTools.getModule('segmentation');
const DEFAULT_BRUSH_RADIUS = configuration.radius;
```
```

File: extensions/dicom-seg/src/components/SegmentationPanel.js [hunk 7]
```diff
@@ -0,0 +1,466 @@
...
+ * SegmentationPanel component
+ *
+ * @param {Object} props
+ * @param {Array} props.studies
+ * @param {Array} props.viewports - viewportSpecificData
+ * @param {number} props.activeIndex - activeViewportIndex
+ * @returns component
+ */
+const SegmentationPanel = ({ studies, viewports, activeIndex }) => {
+  /* TODO: This technically defaults to 10 if undefined (bug?) */
```
JamesAPetts: Remove this TODO when fixed issue above.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,466 @@
...
+  const [brushRadius, setBrushRadius] = useState(DEFAULT_BRUSH_RADIUS);
+  const [brushColor, setBrushColor] = useState('rgba(221, 85, 85, 1)');
```
JamesAPetts: We shouldn't hardcode this color, in the future the SEG may set the colorLUT to whatever it wants. Slap a `TODO` on it though, as we aren't using creation functionality yet.

extensions/dicom-seg/src/components/SegmentationSelect.js
```diff
@@ -0,0 +1,49 @@
...
+    styles={segmentationSelectStyles}
+    options={options}
+  />
+);
+
+const segmentationSelectStyles = {
+  control: (base, state) => ({
+    ...base,
+    cursor: 'pointer',
+    background: '#151A1F',
```
JamesAPetts: These colors should be imported from css style vars.

extensions/dicom-seg/src/loadSegmentation.js
```diff
@@ -0,0 +1,116 @@
...
+
+  if (!results) {
+    throw new Error('Fractional segmentations are not yet supported');
+  }
+
+  const { labelmapBuffer, segMetadata, segmentsOnFrame } = results;
+  const { setters } = cornerstoneTools.getModule('segmentation');
+
+  // TODO: Could define a color LUT based on colors in the SEG.
+  // TODO -> Check the index and allocate a different labelmap if one exists?
```
JamesAPetts: Remove this TODO, we do this the line below 👍 .

extensions/dicom-seg/src/components/SegmentationSelect.js
igoroctaviano: Done, thanks.

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: Done, thanks.

extensions/dicom-seg/src/loadSegmentation.js
igoroctaviano: Done, thanks.

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: Done, thanks.

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: Done, thanks.

extensions/dicom-seg/src/components/SegmentItem.js
igoroctaviano: Done, thanks.

extensions/dicom-seg/src/components/BrushRadius.js
igoroctaviano: I have added a todo for it and also added the minradius, maxradius from config. I think it's a good to have these interactions indirectly, allows us to decouple this component from cstools.

extensions/dicom-seg/README.md
```diff
@@ -0,0 +1 @@
+# @ohif/extension-dicom-html
```
swederik: typo

igoroctaviano:
> Hey @igoroctaviano , could you remove the growcut segmentation tool from the codebase (and thus `ohif-step` as a dependency? We don't want it in here for now.
> 
> Also sorry if you ended up with the wrong info, could you remove the segmentation creation tools entirely? Those 4 we will be bring back at a later date when we have STOW for SEG.



> Hey @igoroctaviano , could you remove the growcut segmentation tool from the codebase (and thus `ohif-step` as a dependency? We don't want it in here for now.
> 
> Also sorry if you ended up with the wrong info, could you remove the segmentation creation tools entirely? Those 4 we will be bring back at a later date when we have STOW for SEG.

@JamesAPetts I have commented on the toolbar access to those and removed the growcut extension entirely from the codebase, lets keep the original seg branch with some flags around so we can move this code back in.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+    };
+
+    /*
+     * These are specific to each element;
+     * Need to iterate cornerstone-tools tracked enabled elements?
+     * Then only care about the one tied to active viewport?
+     */
+    cornerstoneTools.store.state.enabledElements.forEach(enabledElement =>
+      enabledElement.addEventListener(
+        'cornersontetoolslabelmapmodified',
```
JamesAPetts: `cornersontetoolslabelmapmodified` => `cornerstonetoolslabelmapmodified`

dannyrb:
Haven't looked through this yet, but entering/exiting MPR is a odd. I no longer see load percentage, and the viewports pop into view one after the other. Exiting sometimes leaves the "Exit 2D MPR" button visible (viewport ext is not updated)

dannyrb:
CI is passing after bad annotations were removed from prod.

extensions/dicom-seg/.webpack/webpack.dev.js
```diff
@@ -0,0 +1,8 @@
+const path = require('path');
```
dannyrb: @JamesAPetts should the extension be `dicom-seg` or `dicom-segmentation`? Does it matter that it's a cornerstone flavored `dicom-seg`?

extensions/dicom-seg/.webpack/webpack.prod.js
```diff
@@ -0,0 +1,44 @@
...
+      colors: true,
+      hash: true,
+      timings: true,
+      assets: true,
+      chunks: false,
+      chunkModules: false,
+      modules: false,
+      children: false,
+      warnings: true,
+    },
```
dannyrb: We can probably set these as defaults in the parent config and avoid setting them here (`stats`)

extensions/dicom-seg/.webpack/webpack.prod.js
```diff
@@ -0,0 +1,44 @@
...
+      path: ROOT_DIR,
+      library: 'OHIFExtDicomSeg',
+      libraryTarget: 'umd',
+      libraryExport: 'default',
+      filename: pkg.main,
+    },
+    plugins: [
+      new webpack.optimize.LimitChunkCountPlugin({
+        maxChunks: 1,
+      }),
```
dannyrb: This is to ignore code splits and give us a single es5 import, right?

extensions/dicom-seg/README.md
dannyrb: ^ What Erik said

May be worth copying and updating the basic table we have in the `cornerstone` extension. We should provide some high level information about what's contained in the project. This is also used as the "info" for the npm registry.

extensions/dicom-seg/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@ohif/extension-dicom-seg",
```
dannyrb: @JamesAPetts, just want to be sure this is the name we want. This does add a sopClassHandler for "DICOM SEG" but it also contains UI and tool code that's tightly coupled to `cornerstone-tools`

extensions/dicom-seg/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "version": "0.0.1",
+  "description": "OHIF extension for rendering structured reports to HTML",
```
dannyrb: Update description

extensions/dicom-seg/src/OHIFDicomSegSopClassHandler.js
dannyrb: I guess. Before merge, I'd like to see the `sopClassHandler` docs updated to indicate that this optional/convention. This is something you, @swederik, and I should think on for our architecture talks (now that we're slowly deprecating `sopClassHandler` in extensions)

extensions/dicom-seg/src/commandsModule.js
dannyrb: Yupp. Looks like we can remove the commandsModule entirely 👍

extensions/dicom-seg/src/components/BrushColorSelector.js
```diff
@@ -0,0 +1,22 @@
...
+      className="selector-active-segment"
+      style={{ backgroundColor: defaultColor }}
+    >
+      {index}
+    </div>
+    <div className="selector-buttons">
+      <button className="db-button" onClick={onPrev}>
+        Previous
+      </button>
+      <button className="db-button" onClick={onNext}>
```
dannyrb: This is tricky. These components make the assumption that this class is set by the `viewer` project. A kind of soft dependency. We should either co-locate styles, or pull from a central/shared place like `@ohif/ui` in my opinion.

dannyrb: If you do co-locate, they should be scoped to the component, or classes should use an extension prefix to avoid conflict. IE. `.dcmseg-button`

extensions/dicom-seg/src/components/SegmentItem.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { TableListItem, Icon } from '@ohif/ui';
+
+const ColouredCircle = ({ color }) => {
```
dannyrb: ~ColouredCircle~ --> `ColorCircle` or `ColoredCircle`
@JamesAPetts 👀

extensions/dicom-seg/src/components/SegmentItem.js
```diff
@@ -0,0 +1,54 @@
...
+    itemMeta={<ColouredCircle color={color} />}
+    itemMetaClass="segment-color-section"
+    onItemClick={onClick}
+  >
+    <div>
+      <div className="segment-label" style={{ marginBottom: 4 }}>
+        {label}
+      </div>
+      {false && <div className="segment-info">{'...'}</div>}
+      {false && (
```
dannyrb: Shipping some UI dark for future edit mode?

extensions/dicom-seg/src/components/SegmentationItem.js
```diff
@@ -0,0 +1,21 @@
...
+        <div className="segmentation-meta-description">{description}</div>
+      </div>
+    </li>
+  );
+};
+
+SegmentationItem.propTypes = {
+  onClick: PropTypes.func,
+  title: PropTypes.string,
+  description: PropTypes.string,
```
dannyrb: If these are required, please mark them as such.
From a testing perspective, it creates one less test case for each required prop.

extensions/dicom-seg/src/components/SegmentationPanel.css
```diff
@@ -0,0 +1,176 @@
+.labelmap-container {
```
dannyrb: Only concern here is collision with other extensions. It may be worth considering an extension vendor prefix, shifting these to a shared library like `@ohif/ui`, or something that produces component-scoped styles that don't bleed out.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import cornerstoneTools from 'cornerstone-tools';
+import cornerstone from 'cornerstone-core';
+import moment from 'moment';
```
dannyrb: This is not a dependency in `package.json`. It works when you're symlinking and building for PWA, but it will explode if you try to use this extension as a runtime extension.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+
+import { utils } from '@ohif/core';
+import { ScrollableArea, TableList, Icon } from '@ohif/ui';
```
dannyrb: It looks like we actually don't specify `@ohif/ui` as a dependency either, haha.

dannyrb: Bundlers are weird.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+import {
+  BrushColorSelector,
+  BrushRadius,
+  SegmentationItem,
+  SegmentItem,
+  SegmentationSelect,
+} from './index';
+
+import './SegmentationPanel.css';
+import SegmentationSettings from './SegmentationSettings';
```
dannyrb: We may want to update the linter/resolve to require extensions.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+
+const { studyMetadataManager } = utils;
+
+/* TODO: wrap get/set interactions with the cornerstoneTools
+ * store with context to make these kind of things less blurry. */
+const segmentationModule = cornerstoneTools.getModule('segmentation');
```
dannyrb: I think this is fine. `getModule` doesn't return a copy, you actually have "the module", so values should stay fresh. Trying to think if this is better to grab as needed.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+const { configuration } = segmentationModule;
+const DEFAULT_BRUSH_RADIUS = configuration.radius || 10;
+
+const refreshViewport = () => {
+  cornerstone.getEnabledElements().forEach(enabledElement => {
+    cornerstone.updateImage(enabledElement.element);
```
dannyrb: Slight timing thing here to be aware of. This may blow up if the enabledElement does not yet have a displayed image.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+     * Then only care about the one tied to active viewport?
+     */
+    cornerstoneTools.store.state.enabledElements.forEach(enabledElement =>
+      enabledElement.addEventListener(
+        'cornersontetoolslabelmapmodified',
+        labelmapModifiedHandler
+      )
+    );
+
+    document.addEventListener('side-panel-change', handleSidePanelChange);
```
dannyrb: 👀 Can you explain what we're using this for? Maybe add a comment to explain. I don't 100% follow.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+        )
+      );
+
+      document.removeEventListener('side-panel-change', handleSidePanelChange);
+    };
+  });
+
+  const handleSidePanelChange = () => {
+    setShowSegSettings(false);
+  };
```
dannyrb: Is this basically an "unmount" to reset the sidepanels state when we close/open? Basically, making sure when we open, segSettings is always hidden.

extensions/dicom-seg/.webpack/webpack.prod.js
igoroctaviano: Yeah, merge all splits into one chunk to avoid http overhead, @JamesAPetts added it, maybe we can say more about this

extensions/dicom-seg/.webpack/webpack.prod.js
igoroctaviano: Its already defined in base, stats removed from dicom-seg specific webpack config.

extensions/dicom-seg/package.json
igoroctaviano: Updated. @JamesAPetts maybe you have a better description? i added one based on original pr description.

extensions/dicom-seg/src/components/SegmentItem.js
igoroctaviano: That was me, although fits as a tribute to james...

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: I have used this so if I toggle some of the panels (open/hide), when toggling back to show the settings, the settings/seg component needs to be set back to its default values since the component doesnt rerender if panel is hidden. Maybe some other approach in mind?

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: Yes. Maybe we can find another way to do this?

platform/ui/src/components/studyBrowser/ImageThumbnail.js
```diff
@@ -79,22 +79,22 @@ function ImageThumbnail(props) {
...
+  }, [canvasRef, image, image.imageId]);
 
   useEffect(() => {
     if (!image.imageId || image.imageId !== imageId) {
       purgeCancelablePromise();
       setImagePromise();
       fetchImagePromise();
     }
-  }, [imageId]);
+  }, [fetchImagePromise, image.imageId, imageId, purgeCancelablePromise, setImagePromise]);
```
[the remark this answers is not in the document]
dannyrb: I thought we kicked these fetchImage concerns outside of the UI project?

platform/core/src/utils/loadAndCacheDerivedDisplaySets.js
```diff
@@ -0,0 +1,58 @@
+import studyMetadataManager from './studyMetadataManager';
+
+export default function(referencedDisplaySet, studies) {
```
dannyrb: JSDoc? Is this appropriate as a "util"?
What is the shape of these params?

platform/core/src/studies/services/wado/studyInstanceHelpers.js
```diff
@@ -257,7 +293,7 @@ async function makeSOPInstance(server, study, instance) {
     instanceNumber: DICOMWeb.getNumber(instance['00200013']),
     imagePositionPatient: DICOMWeb.getString(instance['00200032']),
     imageOrientationPatient: DICOMWeb.getString(instance['00200037']),
-    frameOfReferenceUID: DICOMWeb.getString(instance['00200052']),
+    frameOfReferenceUid: DICOMWeb.getString(instance['00200052']),
```
dannyrb: 👀 @igoroctaviano, I can never keep track of when/why we caps UID

platform/core/src/studies/services/wado/studyInstanceHelpers.js
```diff
@@ -157,6 +158,41 @@ function getFrameIncrementPointer(element) {
   return frameIncrementPointerNames[value];
 }
 
+function getReferencedSeriesSequence(instance) {
```
dannyrb: JS Doc?
If private, can we prefix with an underscore?
May be worth kicking to an external file.

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -626,6 +692,10 @@ const makeDisplaySet = (series, instances) => {
 
   imageSet.isReconstructable = isReconstructable.value;
 
+  if (shallSort && imageSet.isReconstructable) {
```
dannyrb: shallSort sounds odd

dannyrb: We sort because we need the correct order for 2D MPR? Why not sort even if its not reconstructable?

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -186,9 +196,65 @@ export class StudyMetadata extends Metadata {
...
+    // --> Perhaps that logic should exist in the extension sop class handler and this be a dumb list.
+    // TODO -> Get x Modality by referencedSeriesInstanceUid, FoR, etc.
+  }
+
+  /**
+   * Returns a list of derived datasets in the study, filtered by the given filter.
+   * @param {object} filter An object containing search filters with
+   * optional fields for the modality, referencedSeriesInstanceUID and referencedFrameOfReferenceUID.
+   */
+  getDerivedDatasets(filter) {
```
dannyrb: List of supported filter options would be nice; either deconstructed in signature, or listed in JSDoc params

platform/core/src/classes/ImageSet.js
```diff
@@ -1,5 +1,6 @@
 import guid from '../utils/guid.js';
 import OHIFError from './OHIFError';
+import { Vector3 } from 'cornerstone-math';
```
dannyrb: Are we still doing this? Do we prefer gl-matrix? Erik mentioned a while back we may try to fade out cornerstone-math?

extensions/dicom-seg/src/components/SegmentationSettings.js
```diff
@@ -0,0 +1,157 @@
+import React, { useState, useEffect } from 'react';
+import { Range } from '@ohif/ui';
+
+import './SegmentationSettings.css';
+
+const SegmentationSettings = ({ configuration, onBack, onChange }) => {
+  const [state, setState] = useState({ ...configuration });
```
dannyrb: Oh man. What in the world is in configuration? What if they pass something we don't expect? Suddenly reactive.

extensions/dicom-seg/src/components/SegmentationSettings.js
```diff
@@ -0,0 +1,157 @@
...
+
+  useEffect(() => {
+    setState(state => ({ ...state, ...configuration }));
+  }, [configuration]);
```
dannyrb: Updates to configuration override previous local. Why this instead of just using all of new configuration? Does it provide different keys at different times?

extensions/dicom-seg/src/components/SegmentationSettings.js
```diff
@@ -0,0 +1,157 @@
...
+
+  useEffect(() => {
+    onChange(state);
+  }, [state]);
```
dannyrb: If this updates configuration in the parent, which updates local state, I'm not sure why we have local state and merge above.

extensions/dicom-seg/src/id.js
```diff
@@ -0,0 +1,3 @@
+const id = 'seg';
+
+export default id;
```
dannyrb: Probably just specify this in `index.js`?
Maybe: `com.ohif.dicom-segmentation` for lower chance of collision?

extensions/dicom-seg/src/components/SegmentationSelect.js
```diff
@@ -0,0 +1,55 @@
...
+    '&:last-of-type': {
+      borderBottomLeftRadius: 5,
+      borderBottomRightRadius: 5,
+    },
+    background: state.isSelected ? uiGrayDark : uiGrayDarker,
+    '&:hover': {
+      background: uiGrayDark,
+    },
+  }),
+};
```
dannyrb: This feels different from how we do styles across the rest of the application. What's up?

extensions/dicom-seg/src/init.js
```diff
@@ -0,0 +1,20 @@
...
+
+  tools.forEach(tool => csTools.addTool(tool));
+
+  csTools.addTool(BrushTool, {
+    name: 'BrushEraser',
+    configuration: {
+      alwaysEraseOnClick: true,
+    },
+  });
+}
```
dannyrb: @JamesAPetts this works well when we control these extensions, but I still worry about collisions here across many extensions.

extensions/dicom-seg/src/panelModule.js
```diff
@@ -0,0 +1,18 @@
+import SegmentationPanel from './components/SegmentationPanel.js';
+
+export default {
+  menuOptions: [
+    {
+      icon: 'list',
+      label: 'Segmentations',
+      target: 'example-side-panel',
```
dannyrb: no side?

extensions/vtk/src/OHIFVTKViewport.js
```diff
@@ -146,46 +143,69 @@ class OHIFVTKViewport extends Component {
...
         });
-        */
-      default:
-        imageDataObject = getImageData(stack.imageIds, displaySetInstanceUid);
 
-        return imageDataObject;
+        labelmapDataObject.getPointData().setScalars(dataArray);
+        labelmapDataObject.setDimensions(...imageDataObject.dimensions);
+        labelmapDataObject.setSpacing(
+          ...imageDataObject.vtkImageData.getSpacing()
```
dannyrb: What's in an imageDataObject?

extensions/dicom-seg/src/components/BrushColorSelector.js
igoroctaviano: Totally get it. Since we will not be using those outside this extension, I just scoped all styles within its corresponding extension component main style.

extensions/dicom-seg/src/components/SegmentItem.js
igoroctaviano: Yes, maybe we should just take it out?

extensions/dicom-seg/src/components/SegmentationItem.js
igoroctaviano: Done, thanks.

extensions/dicom-seg/src/components/SegmentationPanel.css
igoroctaviano: Done. Like commented above, since we will not be using those components outside this extension for now, I just scoped all styles within its corresponding extension component main style, this allows us to move them easily in the future.

platform/core/src/studies/services/wado/studyInstanceHelpers.js
igoroctaviano: Done! thanks.

platform/core/src/studies/services/wado/studyInstanceHelpers.js
igoroctaviano: Even in cornerstone, we can see this usage mixed up like frameOfReferenceUID and seriesInstanceUid. Outside OHIF I'm more used to see UID instead of Uid though because this is actually an abbreviation of User ID in which if written using camelcase, it would be UID, upper for each word, right? anyway, it's confusing.

platform/core/src/classes/metadata/StudyMetadata.js
igoroctaviano: @JamesAPetts this is for you my friend.

platform/core/src/classes/metadata/StudyMetadata.js
igoroctaviano: Filter options added in jsdocs 👍

platform/core/src/classes/ImageSet.js
igoroctaviano: @JamesAPetts this is for you my friend.

extensions/dicom-seg/src/components/SegmentationSettings.js
igoroctaviano: Local state is to have this component independent of the parent, this abstraction is to have it like a reusable form, given defaults, set local state and propagate current state on change. That merge is unnecessary though, updated.

extensions/dicom-seg/src/components/SegmentationSelect.js
igoroctaviano: This react-select is third-party, one way to customize it is overriding its theme, where we can use its own state like isFocused or isSelected. Since this is isolated inside its js file, any hidden trap?

extensions/dicom-seg/src/panelModule.js
igoroctaviano: Updated, thanks.

extensions/dicom-seg/src/components/SegmentationSettings.js
igoroctaviano: Not necessary to have this one since the parent will have its state updated anyways, thanks.

extensions/dicom-seg/src/components/SegmentationSettings.js
igoroctaviano: Added prop types for it, maybe we need to add validations/ value boundaries?

extensions/dicom-seg/src/id.js
igoroctaviano: This id file is being imporated+used in OHIFDicomSegSopClassHandler, as the value for the plugin property of a segDisplaySet. https://github.com/OHIF/Viewers/blob/b18d1c3b52884c33d443fc33ecb9ecae5758c9d9/extensions/dicom-seg/src/OHIFDicomSegSopClassHandler.js#L33

extensions/dicom-seg/.webpack/webpack.prod.js
dannyrb: I don't see that change pushed?

platform/ui/src/components/tableList/TableListItem.js
```diff
@@ -10,6 +10,8 @@ export class TableListItem extends Component {
     children: PropTypes.node,
     itemClass: PropTypes.string,
     itemIndex: PropTypes.number,
+    itemMeta: PropTypes.node,
+    itemMetaClass: PropTypes.string,
```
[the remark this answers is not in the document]
igoroctaviano: Not required, just added default props for those

extensions/dicom-seg/src/components/BrushColorSelector.js
dannyrb: Trying to see where these are now defined, and I can't see any pushed changes? Can you link me to the updated line?

extensions/dicom-seg/src/components/SegmentItem.js
dannyrb: Ah. I was definitely being cheeky here, haha.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+    configuration.renderFill = newConfiguration.renderFill;
+    configuration.renderOutline = newConfiguration.renderOutline;
+    configuration.shouldRenderInactiveLabelmaps = newConfiguration.shouldRenderInactiveLabelmaps;
+    configuration.fillAlpha = newConfiguration.fillAlpha;
+    configuration.outlineAlpha = newConfiguration.outlineAlpha;
+    configuration.outlineWidth = newConfiguration.outlineWidth;
+    configuration.fillAlphaInactive = newConfiguration.fillAlphaInactive;
+    configuration.outlineAlphaInactive = newConfiguration.outlineAlphaInactive;
+    refreshViewport();
+  };
```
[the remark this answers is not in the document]
igoroctaviano: configuration is readonly, theres a way to update its value with a new config object? not sure

extensions/dicom-seg/src/components/SegmentItem.js
dannyrb: Nah. It's fine. Maybe a quick comment.

extensions/dicom-seg/.webpack/webpack.prod.js
dannyrb: Nvm. Figured it out.

extensions/dicom-seg/src/components/BrushColorSelector.js
dannyrb: Nvm. I see how GitHub is filtering.

extensions/dicom-seg/src/components/SegmentationPanel.js
dannyrb: We could make sure panel component content receives access to the open/active props? Then it can react to those changes more directly?

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: Event removed, now its directly using props from viewer.

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: Event removed, now its directly using props from viewer.

platform/ui/src/components/studyBrowser/ImageThumbnail.js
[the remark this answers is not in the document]
igoroctaviano: This component is also coupled with global cornerstone, maybe we can refactor it in another ticket? i think @JamesAPetts would know more about those promises being used.

platform/core/src/utils/loadAndCacheDerivedDisplaySets.js
igoroctaviano: JSDoc added with params.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+    const studyMetadata = studyMetadataManager.get(studyInstanceUid);
+    const displaySet = studyMetadata.findDisplaySet(
+      displaySet => displaySet.displaySetInstanceUid === displaySetInstanceUid
+    );
+    return displaySet.images[0].getImageId();
+  } catch (error) {
+    console.error('Failed to retrieve image metadata');
+    return null;
+  }
+};
```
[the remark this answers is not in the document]
igoroctaviano: Sure, moved to studymetadata.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+    modality: PropTypes.string,
+    plugin: PropTypes.string,
+    seriesDate: PropTypes.string,
+    seriesDescription: PropTypes.string,
+    seriesInstanceUid: PropTypes.string,
+    seriesNumber: PropTypes.any,
+    seriesTime: PropTypes.string,
+    sopClassUids: PropTypes.arrayOf(PropTypes.string),
+    studyInstanceUid: PropTypes.string,
+  }),
```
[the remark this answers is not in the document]
igoroctaviano: Updated, some are not.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+ */
+const _setActiveLabelmap = async (
+  viewportSpecificData,
+  studies,
+  displaySet,
+  firstImageId,
+  activeLabelmapIndex
+) => {
+  if (displaySet.labelmapIndex === activeLabelmapIndex) {
+    console.warn(`${activeLabelmapIndex} is already the active labelmap`);
```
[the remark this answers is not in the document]
igoroctaviano: Sure, updated to log + info.

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -286,6 +286,9 @@ class Viewer extends Component {
               updatedState[openKey] = !updatedState[openKey];
             }
 
+            const sidePanelChange = new CustomEvent('side-panel-change');
+            document.dispatchEvent(sidePanelChange);
```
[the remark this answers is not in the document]
igoroctaviano: Removed.

platform/viewer/src/connectedComponents/ConnectedViewerMain.js
```diff
@@ -1,10 +1,10 @@
...
-import ViewerMain from "./ViewerMain";
-import { connect } from "react-redux";
+import OHIF from '@ohif/core';
+import ViewerMain from './ViewerMain';
+import { connect } from 'react-redux';
 
 const {
   setViewportSpecificData,
-  clearViewportSpecificData
+  clearViewportSpecificData,
```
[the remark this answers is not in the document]
igoroctaviano: Sets the viewport to its DEFAULT_STATE in reducer.

platform/viewer/src/components/ViewportGrid/ViewportGrid.js
```diff
@@ -30,6 +33,12 @@ const ViewportGrid = function(props) {
     return null;
   }
 
+  useEffect(() => {
+    viewportData.forEach(displaySet => {
+      loadAndCacheDerivedDisplaySets(displaySet, studies);
```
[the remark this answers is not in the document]
igoroctaviano: @JamesAPetts this is for you mate.

platform/viewer/src/components/SidePanel.css
```diff
@@ -39,6 +40,29 @@
   width: 100%;
 }
 
+.db-button {
```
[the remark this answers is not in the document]
igoroctaviano: Thats right, moved to its specific user styles.

extensions/dicom-seg/src/components/BrushRadius/BrushRadius.js
```diff
@@ -0,0 +1,37 @@
...
+BrushRadius.propTypes = {
+  value: PropTypes.number,
+  onChange: PropTypes.number,
+  min: PropTypes.number,
+  max: PropTypes.number,
+  step: PropTypes.number,
+};
+
+BrushRadius.defaultProps = {
+  value: PropTypes.number.isRequired,
```
dannyrb: Can you use this in `defaultProps`? I think you just wouldn't set the key here, and say it's required in propTypes?

platform/ui/src/components/studyBrowser/ImageThumbnail.js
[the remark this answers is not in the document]
dannyrb: Another ticket it is. Probably a good candidate for OHIF v2 redesign. Resolving.

platform/core/src/studies/services/wado/studyInstanceHelpers.js
dannyrb: WordPress folks having a similar debate: https://github.com/WordPress/gutenberg/issues/6741

extensions/dicom-seg/src/components/SegmentationSettings/SegmentationSettings.js
```diff
@@ -0,0 +1,169 @@
...
+SegmentationSettings.propTypes = {
+  configuration: PropTypes.shape({
+    renderFill: PropTypes.bool,
+    renderOutline: PropTypes.bool,
+    shouldRenderInactiveLabelmaps: PropTypes.bool,
+    fillAlpha: PropTypes.oneOfType([PropTypes.string, PropTypes.number]), /* TODO: why fillAlpha is string? */
+    outlineAlpha: PropTypes.number,
+    outlineWidth: PropTypes.number,
+    fillAlphaInactive: PropTypes.number,
+    outlineAlphaInactive: PropTypes.number,
```
dannyrb: Are each of these keys required or optional?

extensions/dicom-seg/src/components/SegmentationSettings/SegmentationSettings.js
```diff
@@ -0,0 +1,169 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Range } from '@ohif/ui';
+
+import './SegmentationSettings.css';
+
+const SegmentationSettings = ({ configuration, onBack, onChange }) => {
+  const [state, setState] = useState({ ...configuration });
```
dannyrb: Should we destructure here so we don't put anything in state by accident?

dannyrb: configuration is tied to cornerstone-tools seg module, right?
So this is our first bit of reactive state, `onChange` updates cornerstone-tools seg module.

1. Do we need to worry about anything else updating it, but we aren't notified here?
2. If parent is pulling it to pass to this component, why not hoist state to it (or move query to here?)

extensions/dicom-seg/src/id.js
dannyrb: Hmm.. I guess I don't understand the implications. Do they have to be coupled? Do we expect the created display set to have a plugin id that matches a registered extension's id? How are we keying off of this?

I know we use it to match up vtk/cornerstone viewports -- I'm not sure how it impacts seg displaysets.

extensions/dicom-seg/src/components/SegmentationSelect.js
dannyrb: Nope. Thanks for the explanation!

platform/core/src/classes/metadata/StudyMetadata.js
JamesAPetts: shallSort was already present above, always set `true`, its already in master. We could get rid of it now, it was likely placeholder.

> Why not sort even if its not reconstructable?

It only makes sense to sort by `ImagePositionPatient `for volumes, you don't want to attempt to sort an SM or an SR. The regular sort above (line 676) sorts by instance number, which is the best thing to do for an unreconstructable set of images.

One could refactor the `isReconstructable` logic into a generic "sort" that has two paths though 🤷‍♂ .

platform/core/src/classes/ImageSet.js
JamesAPetts: I'm actually indifferent to what happens here.
We currently don't have `gl-matrix` as an explicit dependency anywhere (other than bundled with `vtk.js` (they aren't peer dependencies there). And we already depend on `cornerstone-math`. Long term I would agree, but I don't think its too much of an issue at the current time.

extensions/dicom-seg/src/id.js
JamesAPetts: Right now I don't think this is actually keying off anything. We can remove it.
During implementation I thought it might be useful to know which extension made a `displaySet`. Maybe this is still useful to know for debugging purposes/ possibile collisions in the future, but its not being used right now. I don't mind removing it from the `SOPClassHandler`.

extensions/dicom-seg/src/init.js
JamesAPetts: Right now this is a problem for any extension that adds tools. I don't think its a problem at this time due to the limited scope of extensions. This won't be an issue in the future if extensions are basically pools of components, as you'll pull in the implementations you want in your mode.

At the current time when everything is self registering, I don't think this is too much of a big deal.

extensions/vtk/src/OHIFVTKViewport.js
JamesAPetts: We are just using the same nomenclature as in `react-vtkjs-viewport`. This also has nothing do with this PR, I don't know why its highlighted as changed.

extensions/dicom-seg/.webpack/webpack.dev.js
JamesAPetts: > dicom-seg or dicom-segmentation
Hmm, probably `dicom-segmentation` if we are being pedantic actually, `SEG` is just the modality whilst the P10 is a DICOM Segmentation IOD. However the other OHIF maintained extensions aren't exactly named in any kind of convention.

> Does it matter that it's a cornerstone flavored dicom-seg?
I don't get what you mean by this.

extensions/dicom-seg/.webpack/webpack.prod.js
JamesAPetts: Yeah we use this on each extension, its a bit of a hack that @swederik discovered.

extensions/dicom-seg/package.json
JamesAPetts: To be useable its going to be tied to cornerstone tools for the foresable future. If this pattern ever changes it can have a major version bump.

extensions/dicom-seg/src/components/SegmentationPanel.js
JamesAPetts: Good catch, it was probably VSCode being funky and clever when the word moment was written somewhere. Doesn't look like this was intended to be used anyway.

extensions/dicom-seg/src/components/SegmentationPanel.js
JamesAPetts: Grabbing as needed is probably better practice. In the general case modules might be added after cornerstoneTools has initialised, just this is one of the few "core" modules, so its there from the get-go.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+          for (let i = 0; i < segmentIndexes.length; i++) {
+            if (!acc.includes(segmentIndexes[i]) && segmentIndexes[i] !== 0) {
+              acc.push(segmentIndexes[i]);
+            }
+          }
+        }
+
+        return acc;
+      }, [])
+      .sort((a, b) => a - b);
```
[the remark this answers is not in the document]
JamesAPetts: This is kind of what the `metadata` field of the `labelmap3D` is for. However its optional as any data could be put in there, and the brushes don't fill it by default.

Either we make it so that the default brush tools add a default (blank object?) metadata entry when a segment is drawn, or we could add this as a helper in CST?

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+    if (!brushStackState) {
+      return 'rgba(255, 255, 255, 1)';
+    }
+
+    const colorLutTable =
+      segmentationModule.state.colorLutTables[labelmap3D.colorLUTIndex];
+    const color = colorLutTable[labelmap3D.activeSegmentIndex];
+
+    return `rgba(${color.join(',')})`;
+  };
```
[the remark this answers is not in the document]
JamesAPetts: > Do we need to worry about external things changing these SegmentationModule values and we don't catch them? Should we be emitting events?

All good questions. Its left open to the application consumer to handle this right now. I feel like only the app would be the one altering colorLUTTables. I guess dcmjs could do it when it imports colors 🤔 . Either way as you say this is likely out of scope for now.

platform/core/src/studies/services/wado/studyInstanceHelpers.js
JamesAPetts: I'm going to fix this issue for the entire app soonish. We have a source of truth which is the DICOM JSON standard.

extensions/dicom-seg/src/components/BrushRadius/BrushRadius.js
igoroctaviano: Fixed. thanks.

extensions/dicom-seg/src/components/SegmentationSettings/SegmentationSettings.js
igoroctaviano: Required, updated, thanks.

extensions/dicom-seg/src/components/SegmentationSettings/SegmentationSettings.js
igoroctaviano: Done!

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+      const { labelmapIndex, seriesDate, seriesTime } = displaySet;
+
+      /* Map to display representation */
+      const dateStr = `${seriesDate}:${seriesTime}`.split('.')[0];
+      const date = moment(dateStr, 'YYYYMMDD:HHmmss');
+      const isActiveLabelmap =
+        labelmapIndex === brushStackState.activeLabelmapIndex;
+      const displayDate = date.format('ddd, MMM Do YYYY');
+      const displayTime = date.format('h:mm:ss a');
+      const displayDescription = displaySet.seriesDescription;
```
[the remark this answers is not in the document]
igoroctaviano: @JamesAPetts do you know if we need to sanitize?

extensions/dicom-seg/src/OHIFDicomSegSopClassHandler.js
igoroctaviano: @dannyrb this is pending?

extensions/dicom-seg/src/commandsModule.js
igoroctaviano: Removed. 👍

extensions/dicom-seg/src/commandsModule.js
igoroctaviano: Removed.

extensions/dicom-seg/README.md
igoroctaviano: Update. Now it has a short description of the extension and the tool configuration section. The other sections don't apply for now. @dannyrb if there's anything else I could add, just let me know

extensions/dicom-seg/.webpack/webpack.dev.js
igoroctaviano: @dannyrb this is pending? ill update if necessary

extensions/dicom-seg/package.json
igoroctaviano: @dannyrb pending? let me know if theres something to be done

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: @dannyrb interesting, maybe we could add this as well? https://github.com/depcheck/depcheck#readme maybe this package can do the job?

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: @dannyrb missed this one, what you mean to require extensions?

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: Since this is the module, i think its fine to have it on top level inside the component. Updated.

extensions/dicom-seg/src/components/SegmentationPanel.js
igoroctaviano: @dannyrb what type of check i can do here in this case?

extensions/dicom-seg/src/components/SegmentationPanel.js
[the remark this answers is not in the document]
igoroctaviano: @dannyrb pending?

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+
+  const incrementSegment = event => {
+    event.preventDefault();
+    labelmap3D.activeSegmentIndex++;
+    setActiveSegmentColor();
+  };
+
+  const setActiveSegmentColor = () => {
+    const color = getActiveSegmentColor();
+    setBrushColor(color);
```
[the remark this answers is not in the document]
igoroctaviano: Yeah, labelmap its not reactive, the setbrushcolor was the trigger that indicated that this active segment changed. We had the selected segment state so now I'm setting it before setting brushcolor, now its clearer that this is what triggered the rerendering, thanks.

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+
+const { studyMetadataManager } = utils;
+
+/* TODO: wrap get/set interactions with the cornerstoneTools
+ * store with context to make these kind of things less blurry. */
+const segmentationModule = cornerstoneTools.getModule('segmentation');
+const { configuration } = segmentationModule;
+const DEFAULT_BRUSH_RADIUS = configuration.radius || 10;
+
+const refreshViewport = () => {
```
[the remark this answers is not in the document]
igoroctaviano: I have commented below about this refreshviewports, what type of check i should do before refreshing the viewports? + should I refresh just the active?

extensions/dicom-seg/src/components/SegmentationPanel.js
```diff
@@ -0,0 +1,473 @@
...
+
+/**
+ *
+ * @param {*} firstImageId
+ * @param {*} activeSegmentIndex
+ * @returns
+ */
+const _setActiveSegment = (firstImageId, segmentIndex, activeSegmentIndex) => {
+  if (segmentIndex === activeSegmentIndex) {
+    console.warn(`${activeSegmentIndex} is already the active segment`);
```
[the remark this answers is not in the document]
igoroctaviano: Updated.

platform/core/src/classes/metadata/StudyMetadata.js
igoroctaviano: Something to be done here? @dannyrb @JamesAPetts

extensions/dicom-seg/src/id.js
igoroctaviano: @dannyrb pending?

extensions/dicom-seg/.webpack/webpack.dev.js
dannyrb: @JamesAPetts, that this extension has a hard dependency on cornerstone-tools. I wasn't sure if the naming mattered, or if the intent is for it to _eventually_ work so long as its provided that it need by the platform.

@igoroctaviano, I'll let @JamesAPetts determine if a rename is worth the effort.

JamesAPetts: Lets go with `dicom-segmentation` :)

fedorov:
I am getting this error:

image

dannyrb:
@fedorov, is there a specific route, study, or steps that cause that error for you? I'm unable to reproduce on the latest netlify deploy preview.

fedorov:
We tried this at a call today, and @swederik, @JamesAPetts and @pieper were able to reproduce this. We were just selecting a study with SEG.

I tried again, and was able to reproduce it on the first try with this study: https://deploy-preview-1392--ohif.netlify.com/pwa/viewer/1.3.6.1.4.1.14519.5.2.1.3671.7001.133687106572018334063091507027

pieper:
I get a slightly different stack trace for the same link:

[image: image.png]

PR #1421, opened 2020-02-05
fix: Switch MetadataProvider imagePlane to imagePlaneModule

swederik:
We used to use `imagePlane` in the metadata provider, but it was switched to `imagePlaneModule` some time ago. 

Right now if we are using WADO-URI to obtain image data, the imageId which is generated does not match the imageId that the metadata provider is using for the data which was obtained from the initial WADO-RS RetrieveMetadata request. We will want to change this later somehow, so that requests for metadata are not tied to the method of ingestion.

For now this fixes some issues (e.g. clicking MPR button before a stack has loaded) when the imagePlaneModule metadata is not available for all instances yet (i.e. when you are using WADO-URI).

### DIFF
--- a/platform/core/src/classes/MetadataProvider.js
+++ b/platform/core/src/classes/MetadataProvider.js
@@ -75,7 +75,7 @@ class MetadataProvider {
 
     // If there is sufficient information, populate
     // the imagePlane object for easier use in the Viewer
-    metadata.imagePlane = this.getImagePlane(instanceMetadata);
+    metadata.imagePlaneModule = this.getImagePlane(instanceMetadata);
 
     // Add the metadata to the imageId lookup object
     this.metadataLookup.set(imageId, metadata);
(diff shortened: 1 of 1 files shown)


dannyrb:
This change appears to cause this test failure:

### Cypress Error Capture



### Viewer-dev

image

dannyrb:
There are some other test failures; re-running with some updated measurements to see if it's a persistence error or issues caused by this PR

dannyrb:
Fixed persistence issue by clearing measurements. Only issue (caught by e2e tests) appears to be that bottom overlay text

swederik:
Weird, it uses imagePlaneModule here: https://github.com/cornerstonejs/react-cornerstone-viewport/blob/a18dd08c0681302819b96a4138e048bf60a82cb1/src/ViewportOverlay/ViewportOverlay.js#L58

dannyrb:
We register this class with Cornerstone though? I think this method:

https://github.com/OHIF/Viewers/blob/9f13d03bbe0d7254759dc01fcc687bb343d06c40/platform/core/src/classes/MetadataProvider.js#L400-L410

Would find your updated property: `imagePlaneModule`
If it had higher priority than a different provider.

swederik:
There's also this: https://github.com/OHIF/Viewers/blob/cf5160fb136ab1fd9c06edcad0b57d8845b879b0/extensions/cornerstone/src/init.js#L9

Which I would have expected to allow the WADO-URI imageId to find the WADO-RS metadata.

PR #1424, opened 2020-02-06
WIP: PDF Reports

igoroctaviano:
### PR Checklist


platform/viewer/src/utils/reports/exportPDF.js
```diff
@@ -0,0 +1,105 @@
...
+
+const exportPdf = (measurementApi, timepointApi) => {
+  const currentTimepoint = timepointApi.current();
+  const { timepointId, studyInstanceUids } = currentTimepoint;
+
+  const study = studyMetadataManager.get(studyInstanceUids[0]).getData();
+
+  const report = new MeasurementReport({
+    header: {
+      trial: 'RECIST 1.1',
```
swederik: You can just make this Measurement Report ${currentDate} for now. We will maybe let the user specify it later.

platform/viewer/src/utils/reports/exportPDF.js
```diff
@@ -0,0 +1,105 @@
+import { utils } from '@ohif/core';
+import csTools from 'cornerstone-tools';
+import cornerstone from 'cornerstone-core';
+
+import MeasurementReport from './MeasurementReport';
```
swederik: Since this will pull in jspdf, maybe we can dynamically import the export functionality (somewhere, not sure it's best here), since jspdf is pretty big

platform/viewer/src/utils/reports/exportPDF.js
dannyrb: If this is based on a restored DICOM SR, it may make more sense to use the DICOM SR date.

There is some added complexity, and maybe better for a v2, but I think it's at least eventually what we would want to do.

swederik:
Closing because we won't be doing this in the current version, but maybe in the next one.

Issue #1280, opened 2019-12-11
Find and resolve two Bugs.


JiangWeian:
1.If the image header does not contain the tag of 00200032 (imagepositionpatient), the image will not be displayed. This is because not determine whether it exists when using this value. We should add judgment to the following code: Under file of '\platform\core\src\utils\isDisplaySetReconstructable.js',
function _getImagePositionPatient(instance) {
    let imagePositionPatient = instance.getTagValue('x00200032');
    // Should verify the value of imagePositionPatient is exists, if not, assign a default value
    if (!imagePositionPatient)
        imagePositionPatient = "0\\0\\0";

    return imagePositionPatient
    .split('\\')
    .map(element => Number(element));
}

2.I find that the image displayed on the thumbnail is inconsistent with the series, sometimes there are multiple thumbnails displaying the same image. This is due to in the componentDidMount() method of ImageThumbnail class, before displaying to the thumbnail when downloaded the image, it is not verified whether the image is the corresponding image of this thumbnail.
We should modify then() after fetchimage() in this way: Under file of '\platform\ui\src\components\studyBrowser\ImageThumbnail.js',
this.fetchImage(this.props.imageId)
            .then(image => {
            // Verify whether the two imageId are the same, we should download the image if they are different. Otherwise, the image of thumbnails will be inconsistent.
            if (image.imageId !== this.props.imageId) {
                cornerstone.loadAndCacheImage(this.props.imageId)
                    .then(image => {
                        cornerstone.renderToCanvas(this.canvas.current, image);
                        this.setState({
                            loading: false,
                        });
                    })
            }
            else {
                cornerstone.renderToCanvas(this.canvas.current, image);
                this.setState({
                    loading: false,
                });
            }
        })

sgala:
Hi, we have experimented on a study seen on the field the first of the two bugs commented here, and verified that the patch fixes it:
* a CT with images (DERIVED,SECONDARY,MPR) missing imagepositionpatient.
* the following exception was thrown by OHIF in the console:
```
  TypeError: "instance.getTagValue(...) is undefined"
    _getImagePositionPatient http://10.56.1.125/app.bundle.1a03b0070f6b7f042a2d.js:92886
    processSingleframe http://10.56.1.125/app.bundle.1a03b0070f6b7f042a2d.js:92813
    isDisplaySetReconstructable http://10.56.1.125/app.bundle.1a03b0070f6b7f042a2d.js:92772
[...]

PR #1426, opened 2020-02-07
Fix issues with isDisplaySetReconstructable when we have empty tags

swederik:
This addresses the issues brought up in #1280 and very slightly improves the horrible error handling we have which made this issue hard to debug.


dannyrb:
LGTM 👍 
Bonus points if you want to tack on unit tests

JamesAPetts:
LGTM also.

PR #1428, opened 2020-02-07
WIP: Lesion tracker right panel

igoroctaviano:
### PR Checklist


extensions/lesion-tracker/src/index.js
```diff
@@ -1,12 +1,4 @@
-// SEE: Async imports (code-splitting)
-// import asyncComponent from './asyncComponent.js';
-// import OHIFDicomPDFSopClassHandler from './OHIFDicomPDFSopClassHandler.js';
-
-// const ConnectedOHIFDicomPDFViewer = asyncComponent(() =>
-//   import(
-//     /* webpackChunkName: "ConnectedOHIFDicomPDFViewer" */ './ConnectedOHIFDicomPDFViewer'
-//   )
-// );
+import LesionTrackerPanel from './components/LesionTrackerPanel';
```
dannyrb: We do _eventually_ want this to be generic. Can we change this to MeasurementComparisonTable or something similar?

dannyrb: Worth noting it's technically not a panel. It's content that we ultimately end up slotting into a panel. The distinction helps note that we could throw this in a modal or some other container if the need arose.

extensions/lesion-tracker/src/index.js
```diff
@@ -20,34 +12,22 @@ export default {
...
-    //     component: MyComponent,
-    //   },
-    // ],
-    // defaultContext: ['ROUTE:VIEWER'],
-    // };
+    return {
+      menuOptions: [
+        {
+          icon: 'th-list',
+          label: 'Lesion Tracker',
```
dannyrb: This can be "Measurements"

platform/ui/src/components/measurementTable/MeasurementTable.js
```diff
@@ -25,25 +25,27 @@ class MeasurementTable extends Component {
     t: PropTypes.func,
     saveFunction: PropTypes.func,
     onSaveComplete: PropTypes.func,
+    lesionTracker: PropTypes.bool,
```
dannyrb: I would rather you duplicate this component entirely than add a bool that alters its behavior.

Either we have a generic measurement table with slots or other more common ways to extend the component for this use case, or we have an entirely different component that is composed of small, more atomic, shared generic elements.

platform/ui/src/components/measurementTable/MeasurementTable.styl
```diff
@@ -84,6 +84,106 @@
...
+                  stroke-width: 3px
+                  -webkit-transition: stroke-dashoffset 1s linear
+                  -moz-transition: stroke-dashoffset 1s linear
+                  -ms-transition: stroke-dashoffset 1s linear
+                  -o-transition: stroke-dashoffset 1s linear
+                  -webkit-transition: stroke-dashoffset 1s linear
+                  -moz-transition: stroke-dashoffset 1s linear
+                  -ms-transition: stroke-dashoffset 1s linear
+                  -o-transition: stroke-dashoffset 1s linear
+                  transition: stroke-dashoffset 1s linear
```
dannyrb: No vendor prefixes. If we have issues on those platforms, something like PostCSS should be generating vendor prefixes based on our build target on our behalf.

platform/ui/src/components/measurementTable/MeasurementTable.js
```diff
@@ -184,11 +196,48 @@ class MeasurementTable extends Component {
...
+    return timepoints.map((timepoint, index) => {
       return (
         <div key={index} className="measurementTableHeaderItem">
-          <div className="timepointLabel">{this.props.t(timepoint.key)}</div>
+          <div className="timepointLabel">{t(timepoint.key)}</div>
           <div className="timepointDate">{timepoint.date}</div>
+          {lesionTracker && timepoints.length > 1 && index === 0 && (
+            <div className="caseProgressContainer">
+              <form className="caseProgress" data-key="">
+                <div className="radialProgress">
```
dannyrb: Any reason we can't use our built-in svg icon component? If it doesn't work here, can we extract this as its own tiny component so half of the lines of the component aren't an SVG? (easier to grok semantic structure at a glance)

platform/ui/src/components/measurementTable/MeasurementTable.js
```diff
@@ -184,11 +196,48 @@ class MeasurementTable extends Component {
...
+
+    return timepoints.map((timepoint, index) => {
       return (
         <div key={index} className="measurementTableHeaderItem">
-          <div className="timepointLabel">{this.props.t(timepoint.key)}</div>
+          <div className="timepointLabel">{t(timepoint.key)}</div>
           <div className="timepointDate">{timepoint.date}</div>
+          {lesionTracker && timepoints.length > 1 && index === 0 && (
+            <div className="caseProgressContainer">
+              <form className="caseProgress" data-key="">
```
dannyrb: Why a form element if this is a number and a progress bar?

platform/viewer/src/index.js
```diff
@@ -43,6 +44,7 @@ const appProps = {
     OHIFDicomHtmlExtension,
     OHIFDicomMicroscopyExtension,
     OHIFDicomPDFExtension,
+    OHIFLesionTrackerExtension,
```
dannyrb: If you've pulled in latest, only have this in the `index-lesion-tracker.js` file, and you can test using `yarn run dev:lt`

dannyrb:
image

dannyrb:
CC: @swederik

Questions, for PDF, which of the following do we want to support?

- Preview in Viewport?
- Save (DICOM PDF, STOW)
- Download

For other actions, which ones?

- Download CSV
- Save (DICOM SR) (Current study AND/OR Comparison Study?)
  - Do we create two DICOM SRs per save?

extensions/lesion-tracker/src/index.js
igoroctaviano: Done, thanks.

extensions/lesion-tracker/src/index.js
igoroctaviano: Done, thanks.

platform/ui/src/components/measurementTable/MeasurementTable.styl
igoroctaviano: Done, thanks.

platform/viewer/src/index.js
igoroctaviano: Great, updated, thanks.

platform/ui/src/components/measurementTable/MeasurementTable.js
igoroctaviano: I'll pass in custom timepoints header and custom header for the table component, i think that would make it more generic for future use cases as well. What do you think?

swederik:
> Do we want "Comparison", which only shows one set of measurements, or should we always show "Key Timepoints" (both sets of values) and nix the tabs?

We do not need the tabs, we will only be displaying Current and Comparison for now, so there won't be any other timepoints.

> Does Targets/Non-Targets reflect the language we want?

No. We do not need these sections

> Expand/Shrink section does not work but it does nothing in Lesion Tracker as well. Should we remove the plus signs?

There was never any expand / shrink section functionality. The plus signs would enable the tool that is used to make the measurement in that section. They are not necessary, you can remove them.

> What actions should each measurement have?

TBD, but I think Relabel, Edit Description for now. Potentially Delete later, but we might let them just delete the tool on the image itself instead.

> Should clicking a measurement jump to it in both viewports, if it exists?

Yes

> For naming, did we land on "Current" and "Comparison" instead of "Baseline" and "Follow-up", where the naming is dictated by which study is in each viewport and the current HP settings?

The naming is constant. The ones that are marked Current when they open it is the Current. The other ones are the Comparison timepoint. HP settings are only used to specify where these appear in the viewport arrangement.

swederik:
Questions, for PDF, which of the following do we want to support?

> Preview in Viewport?
No

> Save (DICOM PDF, STOW)
Not yet

> Download 
Yes

For other actions, which ones?

> Download CSV
Do PDF first. When that's perfect, we can do CSV

> Save (DICOM SR) (Current study AND/OR Comparison Study?)
Yes, but save Current only.

> Do we create two DICOM SRs per save?
For now, we are only saving changes in the Current study, so it will only be one. Later on we will consider editing multiple timepoints simultaneously.

One more thing: Case Progress circle bar is not necessary for the first iteration, it's just polish which is not critical whatsoever.

extensions/lesion-tracker/src/components/MeasurementComparisonTable.js
```diff
@@ -0,0 +1,151 @@
...
+  {
+    measurementId: '120',
+    measurementNumber: '120',
+    itemNumber: 3,
+    label: '(No description)',
+    data: [{ displayText: '2.9 x 9.2' }],
+  },
+];
+
+const baselineCollections = [
```
dannyrb: current

extensions/lesion-tracker/src/components/MeasurementComparisonTable.js
```diff
@@ -0,0 +1,151 @@
...
+    measurements: measurements,
+  },
+  {
+    selectorAction: () => { },
+    groupName: 'Additional Findings',
+    measurements: additionalFindings,
+  },
+];
+
+const followupCollections = [
```
dannyrb: comparison

igoroctaviano: Done, thanks.

extensions/lesion-tracker/src/components/MeasurementComparisonTable.js
igoroctaviano: Done, thanks.

PR #1429, opened 2020-02-07
Window level tab

### LINKED ISSUE #1342: Window Level presets
### New feature
On our legacy version we had a feature to set window level presets, we want to have the same feature migrated to new react structure. 

It should be similar to what we have on Meteor version, that was a simple list of commands shortcuts, using numbers and setting the W/L accordingly with a table filled in the configuration modal.

### Legacy screenshot
Screen Shot 2020-01-02 at 17 08 47

galelis:
### Description
Window Level Preset confguration 

### User cases

Closes #1342

### DIFF
--- a/platform/core/src/utils/addUserPreferences.js
+++ b/platform/core/src/utils/addUserPreferences.js
@@ -1,0 +1,7 @@
+import redux from './../redux';
+
+const addUserPreferences = store => {
+  store.dispatch(redux.actions.setUserPreferences());
+};
+
+export default addUserPreferences;
--- a/platform/core/src/utils/index.test.js
+++ b/platform/core/src/utils/index.test.js
@@ -7,6 +7,7 @@ describe('Top level exports', () => {
       'ObjectPath',
       'absoluteUrl',
       'addServers',
+      'addUserPreferences',
       'sortBy',
       'writeScript',
       'b64toBlob',
--- a/platform/viewer/public/config/default.js
+++ b/platform/viewer/public/config/default.js
@@ -69,6 +69,52 @@ window.config = {
     },
     // ~ Cornerstone Tools
     { commandName: 'setZoomTool', label: 'Zoom', keys: ['z'] },
+    // ~ Window level presets
+    {
+      commandName: 'windowLevelPreset1',
+      label: 'W/L Preset 1',
+      keys: ['1'],
+    },
+    {
+      commandName: 'windowLevelPreset2',
+      label: 'W/L Preset 2',
+      keys: ['2'],
+    },
+    {
+      commandName: 'windowLevelPreset3',
+      label: 'W/L Preset 3',
+      keys: ['3'],
+    },
+    {
+      commandName: 'windowLevelPreset4',
+      label: 'W/L Preset 4',
+      keys: ['4'],
+    },
+    {
+      commandName: 'windowLevelPreset5',
+      label: 'W/L Preset 5',
+      keys: ['5'],
+    },
+    {
+      commandName: 'windowLevelPreset6',
+      label: 'W/L Preset 6',
+      keys: ['6'],
+    },
+    {
+      commandName: 'windowLevelPreset7',
+      label: 'W/L Preset 7',
+      keys: ['7'],
+    },
+    {
+      commandName: 'windowLevelPreset8',
+      label: 'W/L Preset 8',
+      keys: ['8'],
+    },
+    {
+      commandName: 'windowLevelPreset9',
+      label: 'W/L Preset 9',
+      keys: ['9'],
+    },
   ],
   cornerstoneExtensionConfig: {},
 };
--- a/platform/viewer/src/appExtensions/GenericViewerCommands/commandsModule.js
+++ b/platform/viewer/src/appExtensions/GenericViewerCommands/commandsModule.js
@@ -1,5 +1,5 @@
-import { redux } from "@ohif/core";
-import store from "./../../store";
+import { redux } from '@ohif/core';
+import store from './../../store';
 const { setViewportActive } = redux.actions;
 
 const actions = {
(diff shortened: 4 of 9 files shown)


dannyrb:
@galelis, playing with the demo, my only question (from the user's perspective) is if we should expose the hotkey configuration for presets, or reserve the "numbers" for them and hardcode their bindings.

platform/core/src/utils/addUserPreferences.js
```diff
@@ -0,0 +1,7 @@
+import redux from './../redux';
+
+const addUserPreferences = ({ store, preferences }) => {
+  store.dispatch(redux.actions.setUserPreferences(preferences));
```
dannyrb: This currently allows _any_ object for preferences. Should we deconstruct and make sure we're only setting for the general and window level top level keys so more state doesn't sneak in?

platform/core/src/utils/index.test.js
```diff
@@ -7,6 +7,7 @@ describe('Top level exports', () => {
       'ObjectPath',
       'absoluteUrl',
       'addServers',
+      'addUserPreferences',
```
dannyrb: Should this be "add" or "set"?

platform/core/src/utils/addUserPreferences.js
dannyrb: Is `addUserPreferences` an exposed ActionCreator? Should we expose it in a way that makes that more obvious? (Utils feels like a weird place for this)

platform/viewer/public/config/default.js
```diff
@@ -69,6 +69,52 @@ window.config = {
...
+    {
+      commandName: 'windowLevelPreset8',
+      label: 'W/L Preset 8',
+      keys: ['8'],
+    },
+    {
+      commandName: 'windowLevelPreset9',
+      label: 'W/L Preset 9',
+      keys: ['9'],
+    },
```
dannyrb: We really need a way for extensions to supply "default" hotkeys, or a way to compose configs? Our configs are getting crazy. Maybe a "hidden" or "locked" key here as a way to set these, but block redefine.

platform/viewer/src/appExtensions/GenericViewerCommands/commandsModule.js
```diff
@@ -16,24 +18,84 @@ const actions = {
...
 };
 
 const definitions = {
   // Next/Previous active viewport
   incrementActiveViewport: {
     commandFn: actions.updateActiveViewport,
-    storeContexts: ["viewports"],
-    options: { direction: 1 }
+    storeContexts: ['viewports'],
+    options: { direction: 1 },
```
dannyrb: Hmm.. We can "cache" updated hotkeys when users change them. Maybe we also allow configs to set "options" to call for hotkeys by default, and on save we can cache these updated options? That way, we're not saving the windowLevelPreset index, and can instead store the width/level as options.

platform/viewer/src/components/UserPreferences/UserPreferences.js
```diff
@@ -25,7 +27,7 @@ const tabs = [
     name: 'Window Level',
     Component: WindowLevelPreferences,
     customProps: {},
-    hidden: true,
+    hidden: false,
```
dannyrb: Where is this used?

platform/viewer/src/components/UserPreferences/WindowLevelPreferences.js
```diff
@@ -1,19 +1,115 @@
...
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import { useSelector, useDispatch } from 'react-redux';
+import { redux } from '@ohif/core';
 
-import { TabFooter } from '@ohif/ui';
+import { TabFooter, useSnackbarContext } from '@ohif/ui';
 import { useTranslation } from 'react-i18next';
 
+const { actions } = redux;
```
dannyrb: We can probably nuke the `redux` dependency if this broadcasts save/restore and hotkeys manager handles localstorage persistence? Using redux is probably fine for this PR, but I'm not sure hoisting state that high has much value

platform/viewer/src/components/UserPreferences/UserPreferences.js
galelis: It was already there, its used to display or not the tab, either the component and the header tab.

galelis: I've set false as default and remove it from configuration, as they were all false.

platform/core/src/utils/addUserPreferences.js
galelis: I've remove it.

platform/core/src/utils/index.test.js
galelis: Just remove this file, after some refactors its not necessary anymore.

galelis:
> @galelis, playing with the demo, my only question (from the user's perspective) is if we should expose the hotkey configuration for presets, or reserve the "numbers" for them and hardcode their bindings.

So, I believe its just a metter of choice, I have just followed what we have on legacy. 
I believe user could select its own hotkey, so they could have `L` to lungs.. `B` for brain.. things like this. But I am ok by hardcoding them as well.

platform/viewer/src/components/UserPreferences/WindowLevelPreferences.js
galelis: To nuke redux dependency we should update hotkeysManager so we could store hotkeys together with options (such as window level preferences). This sould be worked by this issue: https://github.com/OHIF/Viewers/issues/1447

platform/viewer/src/appExtensions/GenericViewerCommands/commandsModule.js
galelis: I agree with you Danny, I will update this issue (#1447) so we can update hotkeyManager and get this done on a different PR

platform/viewer/src/appExtensions/GenericViewerCommands/commandsModule.js
```diff
@@ -1,5 +1,8 @@
-import { redux, utils } from '@ohif/core';
+import { redux } from '@ohif/core';
 import store from './../../store';
+
+import { commandsManager } from './../../App.js';
```
dannyrb: Because this is a `commandsModule`, it should have access to the `commandsManager`. It's provided to `getCommandsModule` as an argument.

dannyrb:
Last issue for merge: https://github.com/OHIF/Viewers/pull/1429/files#r382250617

We should create a follow-up issue to discuss W/L bindings. I'm okay w/ allowing bindings to be reset, but they should either be set in the same tab (like previously) or "hard set" and moved to our block list. I'll create this issue and get @urbantrinity's take.

dannyrb:
@galelis, re-request review when updated please.

PR #1434, opened 2020-02-10
feat: Create API Layer and custom Hook to use the API 

rodrigolabs:
## Description

Currently, we have API calls on the client sprinkled around, and where/how we make requests is not always obvious.

This PR creates an API Layer to try to reduce data calls and use them in the top-level containers. The goal is to centralize all of our web requests to a single layer.

This new API layer is a generic layer that uses `axios` to manage the calls.

Also, a generic custom hook `useApi` was created to consume the new API layer. The intention with the hook is to have a UI handler separated from the API layer itself, where we'll be able to handle each call better, having its status like `isLoading`, `isError` and the `data` per call. This hook can be reusable, even extendable.

### Basic example

```javascript
import React from ‘react’;
import useApi from ‘viewers_path/src/customHooks/useApi’;

const Component = () => {
  const { state, api } = useApi();
  const fetchData = () => {
    api.GET(‘fetchUrl’);
  };
  const shouldRenderData = !state.isLoading && !state.isError;
  return (
    <>
      {state.isLoading && ‘Loading...’}
      {state.isError && state.errorMessage}
      {shouldRenderData && JSON.stringify(state.data)}
      <button onClick={fetchData}>Click to fetch</button>
    </>
  );
};

export default Component;
```


platform/viewer/package.json
```diff
@@ -88,10 +89,13 @@
...
   "devDependencies": {
+    "@testing-library/react-hooks": "^3.2.1",
     "cypress": "^3.8.0",
     "gh-pages": "2.0.1",
     "identity-obj-proxy": "3.0.x",
+    "jest-mock-axios": "^3.2.0",
     "lodash": "4.17.15",
-    "lodash.clonedeep": "4.5.0"
+    "lodash.clonedeep": "4.5.0",
+    "react-test-renderer": "^16.12.0"
```
galelis: Are we using this package?

platform/viewer/src/api/baseUrls.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  studies: 'url',
```
galelis: Any reason to add this data here? Can it be empty?

platform/viewer/package.json
rodrigolabs: Yes, because of `@testing-library/react-hooks`. It uses `react-test-renderer` but it doesn't come bundled together to allow us to install a specific version of `react-test-renderer`.

platform/viewer/src/api/baseUrls.js
rodrigolabs: This is just an initial file for us to keep in mind that we should centralize API URLs there.
I'll remove the `studies` key/value and keep this file empty for future use :)
Thanks!

rodrigolabs:
@galelis Thanks for reviewing this PR :)

> Why do we have it implemented on Viewer level?
> Maybe we could add it as an OHIF/core so we can use in other extensions, packages as well.

Initially, It will live in the viewer project folder for now, as requested by @dannyrb -- what's your opinion on that?

PR #1437, opened 2020-02-11
Feat filters persistence

ladeirarodolfo:
### PR Checklist

### DIFF
--- a/platform/core/src/utils/getContentFromArrayMap.js
+++ b/platform/core/src/utils/getContentFromArrayMap.js
@@ -1,0 +1,14 @@
+const getContentFromArrayMap = (
+  propertyKey,
+  contentArrayMap,
+  defaultContent
+) => {
+  const content =
+    propertyKey in contentArrayMap
+      ? contentArrayMap[propertyKey]
+      : defaultContent;
+
+  return content;
+};
+
+export default getContentFromArrayMap;
--- a/platform/ui/src/components/studyList/TableSearchFilter.js
+++ b/platform/ui/src/components/studyList/TableSearchFilter.js
@@ -77,10 +77,18 @@ function TableSearchFilter(props) {
 
   return translationsAreReady
     ? meta.map((field, i) => {
-        const { displayText, fieldName, inputType } = field;
+        const {
+          displayText,
+          fieldName,
+          inputType: fieldInputType,
+          clearable,
+        } = field;
         const isSortField = sortFieldName === fieldName;
         const sortIcon = isSortField ? sortIconForSortField : sortIcons[0];
-
+        let inputType = fieldInputType;
+        if (clearable && inputType === 'text') {
+          inputType = 'search';
+        }
         return (
           <th key={`${fieldName}-${i}`}>
             <label
@@ -90,15 +98,16 @@ function TableSearchFilter(props) {
               {`${displayText}`}
               <Icon name={sortIcon} style={{ fontSize: '12px' }} />
             </label>
-            {inputType === 'text' && (
+            {fieldInputType === 'text' && (
               <input
-                type="text"
+                type={inputType}
                 id={`filter-${fieldName}`}
                 className="form-control studylist-search"
                 value={values[fieldName]}
-                onChange={e => onValueChange(fieldName, e.target.value)}
+                onChange={e => onValueChange({ [fieldName]: e.target.value })}
               />
             )}
+
             {inputType === 'date-range' && (
               // https://github.com/airbnb/react-dates
               <CustomDateRangePicker
@@ -109,8 +118,10 @@ function TableSearchFilter(props) {
                 endDateId="end-date"
                 // TODO: We need a dynamic way to determine which fields values to update
                 onDatesChange={({ startDate, endDate, preset = false }) => {
-                  onValueChange('studyDateFrom', startDate);
-                  onValueChange('studyDateTo', endDate);
+                  onValueChange({
+                    studyDateFrom: startDate,
+                    studyDateTo: endDate,
+                  });
                 }}
                 focusedInput={focusedInput}
                 onFocusChange={updatedVal => setFocusedInput(updatedVal)}
--- a/platform/ui/src/hooks/useStorage.js
+++ b/platform/ui/src/hooks/useStorage.js
@@ -1,0 +1,39 @@
+import { useState, useEffect, useCallback } from 'react';
+
+export const useSessionStorage = (key, defaultValue = {}) => {
+  const valueFromStorage = window.sessionStorage.getItem(key);
+  const storageValue = valueFromStorage
+    ? JSON.parse(valueFromStorage)
+    : defaultValue;
+  const [storeItem, setStoreItem] = useState({ ...storageValue });
+
+  const saveToSessionStorage = useCallback(value => {
+    setStoreItem({ ...value });
+    window.sessionStorage.setItem(key, JSON.stringify(value));
+  }, []);
+
+  useEffect(() => {
+    saveToSessionStorage(storeItem);
+  }, []);
+
+  return [storeItem, saveToSessionStorage];
+};
+
+export const useLocalStorage = (key, defaultValue = {}) => {
+  const valueFromStorage = window.localStorage.getItem(key);
+  const storageValue = valueFromStorage
+    ? JSON.parse(valueFromStorage)
+    : defaultValue;
+  const [storeItem, setStoreItem] = useState({ ...storageValue });
+
+  const saveToLocalStorage = useCallback(value => {
+    setStoreItem({ ...value });
+    window.localStorage.setItem(key, JSON.stringify(value));
+  }, []);
+
+  useEffect(() => {
+    saveToLocalStorage(storeItem);
+  }, []);
+
+  return [storeItem, saveToLocalStorage];
+};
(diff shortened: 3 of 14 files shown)


dannyrb:
We didn't run Visual Regression Tests after @galelis's Hotkeys update, so this test likely just needs tweaked. The Percy flavored Cypress tests aren't really trying to "catch" anything, they just use Cypress to setup the app for reproducible screen shot states.

image

platform/viewer/cypress/support/commands.js
```diff
@@ -83,7 +87,11 @@ Cypress.Commands.add('isPageLoaded', (url = '/viewer/') => {
 
 Cypress.Commands.add('openStudyList', () => {
   cy.initRouteAliases();
-  cy.visit('/');
+  cy.visit('/', {
+    onBeforeLoad: win => {
+      win.sessionStorage.clear();
+    },
+  });
```
dannyrb: I'm guessing this is slightly faster than having Cypress click the `clear` button? (reduces requests by 1)

platform/viewer/src/studylist/StudyListRoute.js
```diff
@@ -172,33 +224,143 @@ function StudyListRoute(props) {
...
+        newProps
+      );
+      fetchStudies(newStudyListSessionProps);
+    },
+    [studyListSessionProps]
+  );
+
+  const memoizedList = useMemo(() => {
+    return (
+      <StudyList
```
dannyrb: Don't component only re-render if one of their props has updated? Does `useMemo` save us some CPU here? (why not wrap all component in useMemo?)

platform/viewer/src/studylist/StudyListRoute.js
```diff
@@ -172,33 +224,143 @@ function StudyListRoute(props) {
...
+      };
+
+      const newStudyListSessionProps = mergeStudyListSessionProps(
+        studyListSessionProps,
+        newProps
+      );
+      fetchStudies(newStudyListSessionProps);
+    },
+    [studyListSessionProps]
+  );
```
dannyrb: Some repeat code here for setting these. Would it make more sense to shift from individual pieces of state to a "filter + sort" object, and then handle all of the callback/session/updates in once place?

platform/viewer/src/studylist/StudyListRoute.js
```diff
@@ -78,55 +131,54 @@ function StudyListRoute(props) {
...
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [
-      debouncedFilters,
-      debouncedSort,
-      rowsPerPage,
-      pageNumber,
-      displaySize,
-      server,
-    ]
+    [server, displaySize]
```
dannyrb: Does `useCallback` net us much of a benefit here? Is the readability loss worth the performance gain?

platform/viewer/src/studylist/StudyListRoute.js
```diff
@@ -78,55 +131,54 @@ function StudyListRoute(props) {
...
+        rowsPerPage,
+        pageNumber,
+        displaySize
+      );
+
+      setStudies(response);
+      setSearchStatus({ error: null, isSearchingForStudies: false });
+    }, FETCH_DEBOUNCE_TIME),
+    [server, displaySize]
+  );
```
dannyrb: We could leverage our `useDebounce` on a more complex `filters + sort` object. When the debounce elapses, we could fetchData and store/cache the filter + sort in session storage?

platform/viewer/src/studylist/StudyListRoute.js
```diff
@@ -22,39 +29,88 @@ import UserManagerContext from '../context/UserManagerContext';
...
+}
+
+const getInitialStudyListSessionProps = () => {
+  return {
+    sort: DEFAULT_SORT,
+    filters: DEFAULT_FILTERS,
+    rowsPerPage: DEFAULT_ROWS_PER_PAGE,
+    pageNumber: DEFAULT_PAGE_NUMBER,
+  };
+};
```
dannyrb: I think this makes sense as an object, but I don't know that we need to call them "Session Props"?
When the component loads, we can hydrate from session storage, but it's still just state.

platform/viewer/src/studylist/StudyListRoute.js
```diff
@@ -22,39 +29,88 @@ import UserManagerContext from '../context/UserManagerContext';
...
+
+const mergeStudyListSessionProps = (currentProps, newProps) => {
+  return {
+    ...currentProps,
+    ...newProps,
+  };
+};
```
dannyrb: The "new props" don't really conflict with currentProps. There's no cache overlap?

platform/viewer/src/studylist/StudyListRoute.js
```diff
@@ -22,39 +29,88 @@ import UserManagerContext from '../context/UserManagerContext';
...
-    studyDescription: '',
-    //
-    patientNameOrId: '',
-    accessionOrModalityOrDescription: '',
-    //
-    allFields: '',
-  });
+  const [studyListSessionProps, setStudyListSessionProps] = useSessionStorage(
+    'studyListProps',
+    getInitialStudyListSessionProps()
```
dannyrb: I like this setup; not sure we need to call these "SessionProps" though -- threw me for a loop

platform/ui/src/hooks/useStorage.js
```diff
@@ -0,0 +1,39 @@
+import { useState, useEffect, useCallback } from 'react';
+
+export const useSessionStorage = (key, defaultValue = {}) => {
```
dannyrb: These are neat. Can we JSDOC usage? Any easy way to unit test these?

dannyrb: https://blog.logrocket.com/a-quick-guide-to-testing-react-hooks-fa584c415407/

We may also want a file per hook. I realize these are similar, but I would expect a `useStorage` hook in this file at first glance.

platform/ui/src/components/studyList/TableSearchFilter.js
```diff
@@ -77,10 +77,18 @@ function TableSearchFilter(props) {
...
+          inputType: fieldInputType,
+          clearable,
+        } = field;
         const isSortField = sortFieldName === fieldName;
         const sortIcon = isSortField ? sortIconForSortField : sortIcons[0];
-
+        let inputType = fieldInputType;
+        if (clearable && inputType === 'text') {
+          inputType = 'search';
+        }
```
dannyrb: I'm not sure I follow what's going on here.

platform/core/src/utils/getContentFromArrayMap.js
```diff
@@ -0,0 +1,14 @@
+const getContentFromArrayMap = (
```
dannyrb: JS Doc?

platform/viewer/src/studylist/StudyListRoute.js
ladeirarodolfo: ok.. changing it to a better name

platform/ui/src/components/studyList/TableSearchFilter.js
ladeirarodolfo: At user experience: add x icon to remove text on a input field
At dev level: use inputType as search when field is clearable

platform/viewer/cypress/support/commands.js
ladeirarodolfo: I would say yes.. and much more safer...

platform/viewer/src/studylist/StudyListRoute.js
ladeirarodolfo: yeah we could, but to avoid having component re-render multiple times and considering things might change simultaneously (i.e page+filter, sort+something else...) this implementation ensure unicity and scalability. We can loop through it for any change if you desire to

platform/viewer/src/studylist/StudyListRoute.js
ladeirarodolfo: useCallback brings us mainly two things:
- performance
- unicity

From my understanding, considering React`s new way of coding, this method is not less readable, is it? We can change, add comments :D, your call my friend

platform/viewer/src/studylist/StudyListRoute.js
ladeirarodolfo: yes we could by to preserve the current user experience flow its kept as its.
To do such change the effort will be higher.
Currently the flow is: child event (ex: key type on filter field), tell parent, do some stuff on parent, update child. I.e parent is totally responsible to process things and update child. With such implementation the updating of child is preserved, session always has the real data, but fetch is called only sometimes..

Totally makes sense to be the way you mentioned but, there are two problems I see:
- Integration between different child. I.e. pieces need to be in sync: When fetching list and page must be updated, ...
- It will cause some refactoring on child end.

platform/viewer/src/studylist/StudyListRoute.js
ladeirarodolfo: I guess I ddint get your point here.

platform/core/src/utils/getContentFromArrayMap.js
ladeirarodolfo: ok

sedghi:

jbocce:
This PR is being closed in favour of this new PR #3749 which targets the most current master branch. Thanks @ladeirarodolfo  for the inspiration, ideas, code and unit tests.

PR #1446, opened 2020-02-14
Creating 2 commands to activate zoom tool and also to move between di…

### LINKED ISSUE #1095: Hotkeys not working: Z, PageUp, PageDown, =

galelis:
### Description
- Create missing command for `setZoomTool`
- Implement the command function for `updateViewportDisplaySet` that will be used to move (nexty/previous) into current loaded displaySets

Closes #1095

### DIFF
--- a/extensions/cornerstone/src/commandsModule.js
+++ b/extensions/cornerstone/src/commandsModule.js
@@ -74,10 +74,6 @@ const commandsModule = ({ servicesManager }) => {
       }
       cornerstoneTools.setToolActive(toolName, { mouseButtonMask: 1 });
     },
-    updateViewportDisplaySet: ({ direction }) => {
-      // TODO
-      console.warn('updateDisplaySet: ', direction);
-    },
     clearAnnotations: ({ viewports }) => {
       const element = getEnabledElement(viewports.activeViewportIndex);
       if (!element) {
@@ -245,7 +241,7 @@ const commandsModule = ({ servicesManager }) => {
     },
     setCornerstoneLayout: () => {
       setCornerstoneLayout();
-    }
+    },
   };
 
   const definitions = {
@@ -334,23 +330,16 @@ const commandsModule = ({ servicesManager }) => {
       storeContexts: ['viewports'],
       options: {},
     },
-    // TODO: First/Last image
-    // Next/Previous series/DisplaySet
-    nextViewportDisplaySet: {
-      commandFn: actions.updateViewportDisplaySet,
-      storeContexts: [],
-      options: { direction: 1 },
-    },
-    previousViewportDisplaySet: {
-      commandFn: actions.updateViewportDisplaySet,
-      storeContexts: [],
-      options: { direction: -1 },
-    },
     // TOOLS
     setToolActive: {
       commandFn: actions.setToolActive,
       storeContexts: [],
       options: {},
+    },
+    setZoomTool: {
+      commandFn: actions.setToolActive,
+      storeContexts: [],
+      options: { toolName: 'Zoom' },
     },
     setCornerstoneLayout: {
       commandFn: actions.setCornerstoneLayout,
--- a/platform/viewer/public/config/netlify.js
+++ b/platform/viewer/public/config/netlify.js
@@ -53,12 +53,12 @@ window.config = {
     // firstImage
     // lastImage
     {
-      commandName: 'nextViewportDisplaySet',
+      commandName: 'previousViewportDisplaySet',
       label: 'Previous Series',
       keys: ['pagedown'],
     },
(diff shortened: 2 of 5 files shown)


extensions/cornerstone/src/commandsModule.js
```diff
@@ -334,24 +330,17 @@ const commandsModule = ({ servicesManager }) => {
...
-      options: { direction: -1 },
-    },
     // TOOLS
     setToolActive: {
       commandFn: actions.setToolActive,
       storeContexts: [],
       options: {},
     },
+    setZoomTool: {
+      commandFn: actions.setToolActive,
```
igoroctaviano: I think we can use the setToolActive in this case or we would go the path of having one command per tool

platform/viewer/src/appExtensions/GenericViewerCommands/commandsModule.js
```diff
@@ -16,24 +12,63 @@ const actions = {
...
+  },
+  updateViewportDisplaySet: ({ viewports, direction }) => {
+    const viewportSpecificData = { ...viewports.viewportSpecificData };
+    const activeViewport = viewportSpecificData[viewports.activeViewportIndex];
+    const studyMetadata = utils.studyMetadataManager.get(
+      activeViewport.studyInstanceUid
+    );
+
+    if (!studyMetadata) {
+      return;
```
igoroctaviano: @dannyrb we could create commands to wrap up state update logic which dispatches or a new action/reducer (redux) with a connected component? whats the model we should strive in this case?

platform/viewer/public/config/netlify.js
```diff
@@ -53,12 +53,12 @@ window.config = {
     // firstImage
     // lastImage
     {
-      commandName: 'nextViewportDisplaySet',
```
igoroctaviano: Next was previous, previous was next  🤓

galelis: Yeah. Can you see any reason for that? hehe as it was like this everywhere it make me think about it. haha

extensions/cornerstone/src/commandsModule.js
galelis: As `setZoomTool` is a hotkey command, hotkeys implementation does not support to runa command with options (yet!)

igoroctaviano:
Looks like when I click length and activate the tool and then press Z, the icon changes and the tool is activated but it's not selected in the toolbar, is length instead. Maybe we should try to activate the tool in the toolbar as well? 

Besides that, it works. Left some small comments.

extensions/cornerstone/src/commandsModule.js
dannyrb: I think we can keep this for now, but then lets create a follow-up issue where we allow hotkey definitions to include "options" that are passed to `runCommand`

galelis: I've create an issue: #1447

platform/viewer/src/appExtensions/GenericViewerCommands/commandsModule.js
dannyrb: I'm not sure I follow. Can you pseudo code an example?

Issue #1433, opened 2020-02-10
Longitudinal Viewing: Left Panel


galelis:
### Description
Create a left panel for Lesion Tracker that should receive a mock data and display it as we have on legacy lesion tracker application (screenshots taken below) we should keep the layout

This task should only cover the UI layout on top of a mock data. 

### Screenshots
<img width="308" alt="Screen Shot 2020-02-10 at 13 10 28" src="https://user-images.githubusercontent.com/2378326/74167284-cbbe9100-4c06-11ea-91b9-325e5d0b1e71.png">
<img width="303" alt="Screen Shot 2020-02-10 at 13 10 35" src="https://user-images.githubusercontent.com/2378326/74167288-cd885480-4c06-11ea-83bd-2fb6a81febeb.png">
<img width="307" alt="Screen Shot 2020-02-10 at 13 10 41" src="https://user-images.githubusercontent.com/2378326/74167289-ce20eb00-4c06-11ea-9e3e-36fb44f910cc.png">
<img width="309" alt="Screen Shot 2020-02-10 at 13 10 53" src="https://user-images.githubusercontent.com/2378326/74167291-ceb98180-4c06-11ea-832a-60d32ae94a8e.png">

sedghi:
Thank you for your past contributions and for raising this issue related to legacy versions of our OHIF Viewer. Your time and effort have been invaluable in helping us improve the product.

We wanted to update you that our development focus has shifted to OHIF Viewer v3, which is now available for testing on `viewer-dev.ohif.org`. This site is deployed from our `master` branch and incorporates new features, optimizations, and bug fixes. If you're still using `viewer.ohif.org`, it's worth noting that this is deployed from our `release` branch and may not have the latest updates. You can read more about our branching strategies [here](https://docs.ohif.org/development/getting-started#developing).

Given the move to Version 3, we are closing older issues that pertain to Versions 1 and 2, as we are no longer providing support for those versions. If you find that the issue you've raised is still relevant in Version 3, we encourage you to reopen this issue or create a new one for proper triage.

Thank you for your understanding and continued support.

PR #1452, opened 2020-02-18
#1433 - Longitudinal Viewing: Left Panel 

galelis:
### Description
- Logitudional Viewing have a left side panel to control the current and comparrison studies, its structure is similar to studyBrowser on common viewer, where we display studies and have the change of expand/colapse and see all thumbnails for each study. On Longitudional Viewer we also have 2 tabs, one showing the current compared studies and another tab showing all available studies. 

### User cases 
As an user: 
- I should be able to see a left side panel once I open a Lesion tracker viewer 
- I should be able to see 2 tabs in the left side panel, one showing current studies to be compared and another with all available studies for the current patiente 
- I should be able to expand/colapse each study and see their thumnails

NOTE: This task just cover UI part of the left side panel. 

Closes: #1452


extensions/lesion-tracker/src/LTStudyBrowser/LTStudyBrowser.js
```diff
@@ -0,0 +1,70 @@
...
+  const { Component, getProps } = tabs[currentTab];
+  const componentProps = getProps(mockData);
+
+  return (
+    <div className="LTStudyBrowser">
+      <div className="tabButtons">
+        <RoundedButtonGroup
+          value={currentTab}
+          options={getRoundedButtonsData(tabs)}
+          onValueChanged={value => setCurrentTab(value)}
```
rodrigolabs: Clicking in the active tab twice is breaking the application.

extensions/lesion-tracker/src/LTStudyBrowser/LTStudyBrowser.js
```diff
@@ -0,0 +1,70 @@
...
+const getRoundedButtonsData = tabs => {
+  return tabs.map((tabData, index) => {
+    return {
+      value: `${index}`,
+      label: tabData.name,
+    };
+  });
+};
+
+function LTStudyBrowser({}) {
```
rodrigolabs: If it doesn't receive props, can we remove the empty object?

extensions/lesion-tracker/src/LTStudyBrowser/LTStudyBrowser.js
```diff
@@ -0,0 +1,70 @@
...
+
+const findStudy = (studies, studyInstanceUid) => {
+  return studies.find(study => study.studyInstanceUid === studyInstanceUid);
+};
+
+const filterStudies = (studies, filteredStudiesUids = []) => {
+  return studies.filter(
+    study => !filteredStudiesUids.includes(study.studyInstanceUid)
+  );
+};
```
rodrigolabs: Can we move this to the `utils.js` file?

extensions/lesion-tracker/src/LTStudyBrowser/LTStudyBrowser.js
```diff
@@ -0,0 +1,70 @@
...
+
+const tabs = [
+  {
+    name: 'Viewing',
```
rodrigolabs: These tab names are definitive? I'm not sure if it would be necessary to add `i18n` here.

platform/ui/src/components/quickSwitch/StudyItem.js
```diff
@@ -26,12 +31,7 @@ export class StudiesItem extends Component {
       >
         <div className="studyItemBox">
           <div className="studyModality">
-            <div
-              className="studyModalityText"
-              style={this.getModalitiesStyle()}
-            >
-              {modalities}
-            </div>
+            <div className="studyModalityText">{modalities}</div>
```
rodrigolabs: When clicking on an opened study box, it would be nice if it closes (toggle opened/closed).

Feb-20-2020 20-27-46

galelis: I will check that with Danny, I thought about that, but then testing legacy implementation it was not collapsing, so I kept this way.

extensions/lesion-tracker/src/LTStudyBrowser/LTStudyBrowser.js
galelis: Nice catch. fixed.

extensions/lesion-tracker/src/LTStudyBrowser/LTStudyBrowser.js
galelis: Naming is not final yet I think.

platform/ui/src/components/quickSwitch/StudyItem.js
galelis: I've talked with Danny and we agreed to add expand/colapse instead of just expand. 
Code is updated.

dannyrb:
Closing this for now. We'll revisit if we need an escape hatch before August

PR #1456, opened 2020-02-20
#1447 - Adding ability of adding options into hotkeysManager

### LINKED ISSUE #1447: Allow HotkeyManager runCommands with options
### Description

Currently we only store for a hotkey configuration: 
```
{
   commandName: <string>,
   keys: <array[string]>,
   label: <string>,
}
``` 
We should be able to store `options` as well, to be used in the `runCommand()` so we can accomplish changes where we have dinamic options, e.g.: 
```
{
   commandName: 'setVOI',
   keys: ['V'],
   label: 'Set Window/Level',
   options: {
      window: 500,
      level: 100,
   }
}
``` 

### Adittional Items that must be considered: 
- Make sure window level presets data are store into hotkeys and nuke it from redux state 
- Make sure `setZoomTool` command is worked by this new changes, and we can just use the current command of `setActiveTool`
- Make sure `nextViewportDisplaySet` and `prevViewportDisplaySet` is also refactored to have a single command moving next/prev using `options`

galelis:
Closes #1447


dannyrb:
1. Failing PR checks.
2. I want to hold on this until W/L is merged so we can address changes to that code in this PR

dannyrb:
@igoroctaviano, first pass at review after https://github.com/OHIF/Viewers/pull/1429 is merged.

Issue #1407, opened 2020-01-31
Docs (Whitelabelling) - example in docs assumes React is available in the global scope


whage:
[The part about whitelabelling](https://docs.ohif.org/viewer/themeing.html#white-labeling) (changing the logo on the UI) gives a code sample that calls `React.createElement()`. `React` however is not available in the context of the global configuration object and so the browser throws an error if I add that example to my `window.config` object.
It also shows a `props` object in the same code example but there is no mention of what it is.

What is the `props` object and how is it related to the global configuration or the `React` object?
How can I get a reference to the `React` object to create a custom element in place of the logo?

I'm using the ohif/viewer:v3.3.4.8766 docker image to run the viewer.

dannyrb:
@whage, I think this is something we'll need to address with a PR. `window.config` can also be a function; we can provide the white-labeling function that returns a React element with `React`. If it's not a function, we can update `whiteLabelling` to support more values that could be used as props for a generic component (SVG resource uri, url, etc?)

whage:
Can we expect a PR with a fix anytime soon?

dannyrb:
@whage this is an Open Source project. Issues are addressed as individuals and organizations donate their time/resources, often dependent on their internal priorities. While someone will likely address this PR, I'm unsure of when.

If this is an issue you're interested in championing (or that is time-sensitive), but you're unsure of where to get started, I can provide brief technical guidance/scope so that you can draft a PR, or put you in contact with a contractor that could expedite/complete the work on your behalf.

whage:
@dannyrb , understood. I'd like to help but I don't know React at all.

whage:
@dannyrb , I drafted a quick & dirty fox, please take a look: https://github.com/OHIF/Viewers/pull/1465

ZHEL12:
加上qq群；933241376

dannyrb:
Fixed here: https://github.com/OHIF/Viewers/pull/1659

PR #1465, opened 2020-02-26
Add support for defining `whiteLabelling.logoComponent` as a function

whage:
This change is about white labeling.
The template that generates the logo on the top of the page requires a
`React Element` and gets it from the `whiteLabelling.logoComponent` part of the global
configuration object.
Since `React` is not available in the scope of the global config object a React element
can't be created on the configuration side.
These changes make it possible for the `whiteLabelling.logoComponent` configuration parameter
to be a **function** which is passed the React object.
This way it is possible to call React.createElement() on the configuration
side and replace the logo with a custom element.

A possible usage:
```
window.config = {
  ...
  whiteLabelling: {
    logoComponent: function(React) {
      return React.createElement("img", {src: "https://www.sandraboynton.com/sboynton.com.data/Components/Photos/One-Shoe-Blues-DVD-menu.gif"})
    }
  }
}
```

A possible fix for https://github.com/OHIF/Viewers/issues/1407.

@dannyrb , can you please take a look?
I never worked with React before. If this is not the right way to do it then please give me some guidance and I'll try to make the right changes.


dannyrb:
@whage, I admire your determination! Are/can React components return their type as `function`? If so, this gets a little hairy.

I think I would be okay with updating this to say that the user should supply a function that returns a React component, then we can _always_ pass `React`, and the consumer can use it if needed.

whage:
@dannyrb , I don't know anything about react components, you tell me! What do you mean by "updating this to say..." ? Do you mean to document it somewhere? Or can we enforce logoComponenet to be a function somehow?

NatiTM:
@dannyrb so the current implementation of whiteLabelling doesn't work?

whage:
@NatiTM , it doesn't, because it needs a React element, but there is no way to create one on the configuration side (unless you already have react in your main app too). Maybe this should be renamed to be a bugfix?

whage:
Why did this get assigned to me? I'm the one who opened it.

NatiTM:
Usually this is how it is done on open source projects.
You can contribute.

whage:
@NatiTM , What do you mean "you can contribute?" I'm the author of the code in this PR. We'd need *someone else* to review it I guess.

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -246,7 +246,7 @@ class Viewer extends Component {
             <UserManagerContext.Consumer>
               {userManager => (
                 <ConnectedHeader home={false} userManager={userManager}>
-                  {whiteLabelling.logoComponent}
+                  {(typeof whiteLabelling.logoComponent === 'function' ? whiteLabelling.logoComponent(React) : whiteLabelling.logoComponent)}
```
dannyrb:
```suggestion
{{ whiteLabelling.createLogoComponentFn && whiteLabelling.createLogoComponentFn(React) }}
```

dannyrb:
👋 @whage 
Sorry for the delay. When a 3rd party PR comes in, it enters a triage process and is prioritized. We only have funds for ~2 developers through our grant; white-labeling is technically outside the scope of the grant, so requests like these are addressed more slowly, and usually by me volunteering time, or one of our community partners sponsoring maintenance/development. 

I think we can update the configuration to add a new method under the whiteLabeling object:

`createLogoComponentFn(React) : ReactComponent`

It should take `React` as a parameter, and return a `ReactComponent` (or undefined/null)

So we need to:

- Update our config and docs
- Update the two files you've already touched

----

Long term, we would probably want to add a "Generic Logo Component" where you can provide an SVG, base64 encoded image, local image, text, link, etc. And then just use those as props for that component.

Then you have the ability to implement your own component, or simply configure the default one.

arantes555:
This is not stale. It is still relevant.

dannyrb:
Fixed Here: https://github.com/OHIF/Viewers/pull/1659

PR #1470, opened 2020-02-28
feat: v2 Button Component

rodrigolabs:
### PR Checklist

___

This PR has the base implementation to use Buttons components.

Different versions and variations for the buttons were created, trying to cover the buttons as much as possible according to the v2 interface.

## Preview

image

image

image
image
image
image
image


platform/ui/tailwind.config.js
```diff
@@ -253,6 +245,7 @@ module.exports = {
     },
     fontFamily: {
       sans: [
+        'Lato',
```
galelis: Is it corrent? maybe create font-lato.

platform/ui/src/components/Button/Button.jsx
```diff
@@ -26,60 +30,99 @@ const disabledClasses = {
...
+    white:
+      'bg-white text-black hover:opacity-80 active:opacity-100 focus:opacity-80',
   },
 };
 
 const sizeClasses = {
-  small: 'py-1 px-2 text-sm',
-  medium: 'py-2 px-4 text-base',
-  large: 'py-3 px-6 text-lg',
+  small: 'py-2 px-2 text-base',
```
galelis: not sure if I like the idea of having small as kind of default size using "base"..

platform/ui/src/components/Button/Button.jsx
```diff
@@ -26,60 +30,99 @@ const disabledClasses = {
...
+      {React.cloneElement(endIconProp, {
+        className: classnames('w-4 h-4 fill-current'),
+      })}
+    </div>
   );
+  const buttonElement = useRef(null);
 
-  const endIcon = endIconProp && <div className="ml-2">{endIconProp}</div>;
+  const handleOnClick = e => {
+    buttonElement.current.blur();
```
galelis: why are we dealing with Blur in a button?

platform/ui/tailwind.config.js
rodrigolabs: We're gonna be using `Lato` as default sans-serif font. That's why I've added `Lato` as a primary `sans` font. If in the future we want to change the default sans font, we can just replace `Lato` for something else, instead of having an additional CSS class for `font-lato`.

platform/ui/src/components/Button/Button.jsx
rodrigolabs: The default font size for Button is `16px`/`1rem`, equivalent to what we currently have in the theme config value for `text-lg`. Do you think it would be better to have `16px` for `text-base` instead of `text-lg`? We can change these values here:

https://github.com/OHIF/Viewers/blob/86e1faa0407817d319690b04f90b299bc90b6861/platform/ui/tailwind.config.js#L273-L284

Any idea?

platform/ui/src/components/Button/Button.jsx
rodrigolabs: When we click on a Button, it gets focused, hence style rules for `:focus` are applied. This `blur` is just to avoid these styles to be applied when we click on the Button. Here's an example without calling `blur()`:

Mar-05-2020 12-43-11

Open to ideas

platform/ui/tailwind.config.js
galelis: Make sense

galelis:
@dannyrb 
All changes look good to me. 
Can you do the final review?

platform/ui/src/components/Button/Button.jsx
```diff
@@ -26,60 +30,99 @@ const disabledClasses = {
...
-    primary: 'bg-blue-900 hover:bg-blue-800 text-white',
-    secondary: 'bg-blue-300 hover:bg-blue-500 text-black',
+    default:
+      'bg-custom-aquaBright text-white hover:opacity-80 active:opacity-100 focus:opacity-80',
+    primary:
+      'bg-custom-blue text-white hover:opacity-80 active:opacity-100 focus:opacity-80',
+    secondary:
+      'bg-custom-violetPale text-white hover:opacity-80 active:opacity-100 focus:opacity-80',
+    white:
+      'bg-white text-black hover:opacity-80 active:opacity-100 focus:opacity-80',
```
dannyrb: We should discuss the best way to maintain these classes. If the biggest difference within a variant is the color, it would be nice if we had an easier way to indicate that.

platform/ui/src/components/Button/Button.jsx
```diff
@@ -26,60 +30,99 @@ const disabledClasses = {
...
+      'text-custom-violetPale hover:bg-custom-violetPale hover:text-white active:opacity-80 focus:bg-custom-violetPale focus:text-white',
+    white:
+      'text-white hover:bg-white hover:text-black active:opacity-80 focus:bg-white focus:text-black',
   },
   outlined: {
-    default: 'border border-black text-black hover:bg-gray-200',
-    primary: 'border border-blue-900 text-blue-900 hover:bg-blue-100',
-    secondary: 'border border-blue-300 text-blue-300 hover:bg-blue-100',
+    default:
+      'border bg-trasparent border-custom-aquaBright text-custom-aquaBright hover:opacity-80 active:opacity-100 focus:opacity-80',
```
dannyrb: Is `-custom` a tailwind convention?

dannyrb: We should discuss color variant naming

platform/ui/src/components/IconButton/IconButton.jsx
```diff
@@ -0,0 +1,137 @@
+import React, { useRef } from 'react';
```
dannyrb: For naming of components, if you stack the names, they appear next to each other in the file tree. This is a Vue.js styleguide recommendation:

```js
Button
ButtonIcon
ButtonSegmented
```

platform/ui/src/components/IconButton/IconButton.jsx
```diff
@@ -0,0 +1,137 @@
...
+    white:
+      'border bg-transparent border-white text-white hover:opacity-80 active:opacity-100 focus:opacity-80',
+  },
+  contained: {
+    default:
+      'bg-custom-aquaBright text-black hover:opacity-80 active:opacity-100 focus:opacity-80',
+    primary:
+      'bg-custom-blue text-white hover:opacity-80 active:opacity-100 focus:opacity-80',
+    secondary:
+      'bg-custom-violetPale text-white hover:opacity-80 active:opacity-100 focus:opacity-80',
```
dannyrb: see repeated: hover, active, focus

platform/ui/src/components/IconButton/IconButton.jsx
```diff
@@ -0,0 +1,137 @@
...
+  children,
+  variant = defaults.variant,
+  color = defaults.color,
+  size = defaults.size,
+  rounded = defaults.rounded,
+  disabled = defaults.disabled,
+  type = defaults.type,
+  fullWidth = defaults.fullWidth,
+  className,
+  ...rest
```
dannyrb: I don't think we would expose so many props, but we can see how things shake out. IE. Maybe _all_ buttons are rounded (or not)

dannyrb: I am not a fan of `...` spread operator for props.

platform/ui/src/components/IconButton/IconButton.jsx
```diff
@@ -0,0 +1,137 @@
...
+  return (
+    <button
+      className={classnames(
+        baseClasses,
+        variantClasses[variant][color],
+        roundedClasses[rounded],
+        sizeClasses[size],
+        fullWidthClasses[fullWidth],
+        disabledClasses[disabled],
+        className
```
dannyrb: I like the idea here, but I think we can make it more semantic.

platform/ui/src/components/IconButton/IconButton.jsx
```diff
@@ -0,0 +1,137 @@
...
+      )}
+      ref={buttonElement}
+      onClick={handleOnClick}
+      type={type}
+      {...rest}
```
dannyrb: Spreading props feels antipatterny, even though it's easier.

platform/ui/src/components/IconButton/IconButton.jsx
```diff
@@ -0,0 +1,137 @@
...
+IconButton.propTypes = {
+  children: PropTypes.node,
+  size: PropTypes.oneOf(['small', 'medium', 'large']),
+  rounded: PropTypes.oneOf(['none', 'small', 'medium', 'large', 'full']),
+  variant: PropTypes.oneOf(['text', 'outlined', 'contained']),
+  color: PropTypes.oneOf(['default', 'primary', 'secondary', 'white']),
+  fullWidth: PropTypes.bool,
+  disabled: PropTypes.bool,
+  type: PropTypes.string,
+  className: PropTypes.node,
```
dannyrb: Which of these are required? None of them?

platform/ui/tailwind.config.js
```diff
@@ -385,9 +378,25 @@ module.exports = {
...
+      '50': '.50',
+      '55': '.55',
+      '60': '.60',
+      '65': '.65',
+      '70': '.70',
+      '75': '.75',
+      '80': '.80',
+      '85': '.85',
+      '90': '.90',
+      '95': '.95',
```
dannyrb: Do we think we'll need this level of flexibility?

platform/ui/tailwind.config.js
```diff
@@ -278,10 +271,10 @@ module.exports = {
...
     },
     fontSize: {
-      xs: '0.75rem',
-      sm: '0.875rem',
-      base: '1rem',
-      lg: '1.125rem',
+      xs: '0.65rem',
+      sm: '0.75rem',
+      base: '0.875rem',
+      lg: '1rem',
```
dannyrb: Were these changes made to align with Dan's design?
Better to bump these, or because these are relative to font-size of the root element, bump the font-size there?

platform/ui/tailwind.config.js
```diff
@@ -14,32 +14,24 @@ module.exports = {
...
+        navy: '#041c4a',
+        violetDark: '#2b166b',
+        violetPale: '#3a3f99',
+        blue: '#0944b3',
+        blueBright: '#348cfd',
+        aquaPale: '#7bb2ce',
+        white: '#ffffff',
+        grayBright: '#e1e1e1',
+        grayLight: '#a19fad',
+        gray: '#726f7e',
```
dannyrb: Naming is tricky. Not sure I want to have to remember _every_ single one of these. Better to use semantic meaning than the color's name as it's easier to recall. Also, I can change the `primary` color for the app, and all classes are correct. Changing "aquaBright" requires a find and replace all.

We should work with Dan to get the correct semantic names for these, and then make sure they're used consistently?

platform/ui/src/gatsby-theme-docz/tailwind.css
```diff
@@ -1,3 +1,6 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+/* IMPORT CUSTOM FONT */
+@import url('https://fonts.googleapis.com/css?family=Lato&display=swap');
```
dannyrb: Similar to current docz, let's add a docs page for font:

https://react.ohif.org/getting-started#usage

PR #1488, opened 2020-03-05
feat: ui v2 Typography Component

rodrigolabs:
_Since #1934 is not merged yet, this Typography PR will contain   the base code of #1470_

## Description

This branch was created on top of #1470 (Button).

This PR has the creation of the Typography component. It's a component that can be used to display text using the available props for customization.

There's basically a list of variants that can be used to match the expectations with the new layout.

## How to use

```javascript
import { Typography } from '@ohfi/ui';

<Typography variant="h1">
  h1. The five boxing wizards jump quickly.
</Typography>
```

## Preview

_It is just a preview. Since the image can be resized here, it may not show 100% of how it will actually be shown._

image


platform/ui/src/components/Typography/Typography.jsx
```diff
@@ -0,0 +1,142 @@
...
+        classes.variant[variant],
+        classes.color[color],
+        classes.align[align],
+        classes.gutterBottom[gutterBottom],
+        classes.paragraph[paragraph],
+        classes.noWrap[noWrap],
+        classes.display[display],
+        className
+      )}
+      {...rest}
```
dannyrb: What is rest here?

platform/ui/src/components/Typography/Typography.jsx
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
```
dannyrb: Is the goal here to duplicate React Material's Typography Component?

- https://material-ui.com/components/typography/
- https://material-ui.com/api/typography/

PR #1497, opened 2020-03-09
feat: Study List Table - v2

rodrigolabs:
## Preview
image

image

image


platform/ui/src/views/StudyList/components/StudyListTable.js
```diff
@@ -3,11 +3,11 @@ import classnames from 'classnames';
 import { Button } from '@ohif/ui';
 
 /** TODO: Icons should be imported from @ohif/ui */
-import ChevronRight from '../../../../../ui/src/assets/icons/chevron-right.svg';
-import ChevronDown from '../../../../../ui/src/assets/icons/chevron-down.svg';
-import InstancesActive from '../../../../../ui/src/assets/icons/instances-active.svg';
-import InstancesInactive from '../../../../../ui/src/assets/icons/instances-inactive.svg';
-import LaunchInfo from '../../../../../ui/src/assets/icons/launch-info.svg';
+import ChevronRight from '../../../assets/icons/chevron-right.svg';
```
dannyrb: Might be worth using a webpack alias in the Gatsby-node process

platform/ui/src/gatsby-theme-docz/tailwind.css
```diff
@@ -3,4 +3,24 @@
...
+
+/* Handle */
+.custom-scrollbar::-webkit-scrollbar-thumb {
+  @apply rounded;
+  @apply bg-custom-blue;
+}
+
+.custom-scrollbar::-webkit-scrollbar-thumb:window-inactive {
+  @apply bg-custom-blue;
+}
```
dannyrb: Can we use a vendor prefix instead of the word custom? `ohif-scrollbar` or something similar?

platform/ui/src/views/StudyList/components/StudyListTable.js
rodrigolabs: We won't need these imports here, they will be replaced for the `<Icon />` component when #1499 gets merged.

platform/ui/src/gatsby-theme-docz/tailwind.css
rodrigolabs: Sure! I'll replace that

PR #1498, opened 2020-03-09
OHIF-20: Study List Header

galelis:
## Description
Create a component with basic structure of header 

## Specs requested 
Screen Shot 2020-03-10 at 17 15 07

## Developed screenshot
Screen Shot 2020-03-10 at 17 14 23

## Missing pieces to be coded in other PR
- OHIF Logo (and text) svgs
- Settings config actions (dropdown and modals)


rodrigolabs:
I've noticed that the `cog` Icon for settings is a little cut-off:

image

Its slice box in Zeplin seems to be wrong. Just to keep in mind that we're gonna need to update that -- asking Dan to update the icon.

platform/ui/src/views/StudyList/components/Header.js
```diff
@@ -1,10 +1,46 @@
...
+function Header({ appLogo = OHIFLogo(), children, t }) {
+  const showSettingsDropdown = () => {
+    // TODO: Update once dropdown component is created
+  };
+
+  return (
+    <div className="bg-custom-navy flex flex-row justify-between p-2 text-white">
+      <div className="flex items-center">
+        <div>{appLogo}</div>
+        <div>{children}</div>
```
rodrigolabs: not sure if we need a `children`

galelis: I have added this because we will need to display some links in header, as `Back to viewer` and `Study List` or anything else. 
In this case I will leave it to the viewer to control. Thats why I kept the children.

rodrigolabs: Great!

galelis:
@dannyrb
Approved by Rodrigao. 
Cna you take a look Danny?

Issue #1130, opened 2019-10-29
Download data button


pieper:
## Request

**What feature or change would you like to see made?**

I'd like to be able to download the part10 dicom files for selected instances/series/studies. 

**Why should we prioritize this feature?**

It's useful to be able to load data in other programs, like Slicer or Osirix.

Since the user has access already, it's technically possible for them to download another way, but a button would make it easier.

salimkanoun:
Could this be made optional, defined in the config parameters of OHIF ?

Depending on the usage you want to make, in some case you might want to avoid users to be able to download a copy of the original dicom.

I know that being client side viewer, the raw dicom will always land anyway on the client computer so there will be always a way to make a copy.

Just, I would be interested to not let users be able to make a copy of source data too easily (at least not giving them the idea to do it)

pieper:
Yes, for sure it can be optional.  Or the functionality could be added via a plugin and not in the standard distribution.  But I guess we both agree this is only security by obscurity (or security by inconvenience in this case).

Perhaps another feature could be that if you click the download button it asks you to agree to some kind of terms of use, like don't share this data with anyone else.  That would make the arrangement more explicit and might better comply with institutional rules (e.g. remind people that they are only seeing the data because they signed an IRB or data use agreement and will be held responsible if they violate the terms).  We could even consider watermarking the headers somehow to track who did the download in case the data shows up someplace unexpected.

salimkanoun:
To me, the optional settings is enough.

I'm not sure the terms of use will be really usefull as someone who will want to hack the system will just have to download the raw data from the DICOMWeb APIs (even without signing the term of use)

To add a tracking security, it's better to manage that in the backend, in my project I put a reverse proxy to allow communication with DICOMWeb orthanc APIs, so I implemented my log / security in this reverse proxy so I can track each call of these APIs.

However it's a good community question, other ideas might appears...

JamesAPetts:
This is an interesting one, because more often than not we request the headers and pixeldata seperately, and the pixeldata frame-by-frame for multiframe images.

In a lot of cases it'd be easier to just hook up the button to fetch the full P10 from PACs, but this won't always be available, and duplicates download requests. We could reconstruct the dataset using dcmjs.

pieper:
Good points - and of course some people may want nifti or nrrd volumes too.

dannyrb:
[...]

Issue #11, opened 2016-03-30
Modifying filters to case insensitive


mtjvankuik:
I want to make the filters in the worklist case insensitive instead of case sensitive. My assumption was that there are three choices.

1) modify comparison server variable and filter variable.
2) capitalize input filter.
3) uncapitalize server variable.

I chose for 2. However, I get a server error after adding the toUpperCase() method.

`/**
 * Transforms an input string into a search filter for
 * the Worklist Search call
 *
 * @param filter The input string to be searched for
 * @returns {*}
 */
function getFilter(filter) {
    
    if (filter && filter.length && filter.substr(filter.length - 1) !== '*') {
        filter += ' * ' ;
    } 

    return filter.toUpperCase();
}`

server log: `Router.route('/', function () {
  this.render('Home', {
    data: function () { return Items.findOne({_id: this.params._id}); }
  });
});
        `

mtjvankuik:
Issue solved. `Transforms an input string into a search filter for
the Worklist Search call *
@param filter The input string to be searched for
@returns {*}
*/
function getFilter(filter) {

if (filter && filter.length && filter.substr(filter.length - 1) !== '*') {
filter += ' * ' ;
filter = filter.toUpperCase();
}

return filter;
}``

PR #1501, opened 2020-03-10
Feat #1130 [WIP] Adding support for downloading and zipping files

emfol:
Introducing a new extension to support downloading and zipping SOP instances from a study

### PR Checklist

### DIFF
--- a/extensions/dicom-p10-downloader/README.md
+++ b/extensions/dicom-p10-downloader/README.md
@@ -1,0 +1,1 @@
+# @ohif/extension-dicom-p10-downloader
--- a/extensions/dicom-p10-downloader/src/commandsModule.js
+++ b/extensions/dicom-p10-downloader/src/commandsModule.js
@@ -98,7 +98,7 @@ function progress(status) {
 
 function error(e) {
   if (e.message === 'Queue limit reached') {
-    OHIF.log.warn('Cannot have more than one download in progress');
+    OHIF.log.warn('A download is already in progress, please wait.');
   } else {
     OHIF.log.error(e);
   }
--- a/extensions/dicom-p10-downloader/src/commandsModule.js
+++ b/extensions/dicom-p10-downloader/src/commandsModule.js
@@ -7,7 +7,14 @@ import {
 } from './utils';
 import _downloadAndZip from './downloadAndZip';
 
+const {
+  utils: {
+    Queue,
+  },
+} = OHIF;
+
 export function getCommands(context) {
+  const queue = new Queue(1);
   const actions = {
     /**
      * @example Running this command using Commands Manager
@@ -54,16 +61,19 @@ export function getCommands(context) {
 
   const definitions = {
     downloadAndZip: {
-      commandFn: actions.downloadAndZip,
+      commandFn: queue.bindSafe(actions.downloadAndZip, error),
       storeContexts: ['servers'],
     },
     downloadAndZipSeriesOnViewports: {
-      commandFn: actions.downloadAndZipSeriesOnViewports,
+      commandFn: queue.bindSafe(actions.downloadAndZipSeriesOnViewports, error),
       storeContexts: ['servers', 'viewports'],
       options: { progress },
     },
     downloadAndZipSeriesOnActiveViewport: {
-      commandFn: actions.downloadAndZipSeriesOnActiveViewport,
+      commandFn: queue.bindSafe(
+        actions.downloadAndZipSeriesOnActiveViewport,
+        error
+      ),
       storeContexts: ['servers', 'viewports'],
       options: { progress },
     },
@@ -84,4 +94,12 @@ function progress(status) {
     'Download and Zip Progress:',
     (status.progress * 100.0).toFixed(2) + '%'
   );
+}
+
+function error(e) {
+  if (e.message === 'Queue limit reached') {
+    OHIF.log.warn('Cannot have more than one download in progress');
+  } else {
+    OHIF.log.error(e);
+  }
 }
--- a/extensions/dicom-p10-downloader/src/commandsModule.js
+++ b/extensions/dicom-p10-downloader/src/commandsModule.js
@@ -1,0 +1,85 @@
+import OHIF from '@ohif/core';
+import {
+  save,
+  getDicomWebClientFromContext,
+  getSOPInstanceReferenceFromActiveViewport,
+  getSOPInstanceReferencesFromViewports,
+} from './utils';
+import downloadAndZip from './downloadAndZip';
+
+export function getCommands(context) {
+  const actions = {
+    // Example for running this command using Commands Manager
+    // commandsManager.runCommand(
+    //   'downloadAndZip',
+    //   {
+    //     listOfUIDs: [...],
+    //     options: {
+    //       progress(status) {
+    //         console.info('Progress:', (status.progress * 100).toFixed(2) + '%');
+    //       }
+    //     }
+    //   },
+    //   'VIEWER'
+    // );
+    downloadAndZip({ servers, dicomWebClient, listOfUIDs, options }) {
+      return save(
+        downloadAndZip(
+          dicomWebClient || getDicomWebClientFromContext(context, servers),
+          listOfUIDs,
+          options
+        ),
+        listOfUIDs
+      );
+    },
+    downloadAndZipSeriesOnViewports({ servers, viewports, progress }) {
+      const dicomWebClient = getDicomWebClientFromContext(context, servers);
+      const listOfUIDs = getSOPInstanceReferencesFromViewports(viewports);
+      return save(
+        downloadAndZip(dicomWebClient, listOfUIDs, { progress }),
+        listOfUIDs
+      );
+    },
+    downloadAndZipSeriesOnActiveViewport({ servers, viewports, progress }) {
+      const dicomWebClient = getDicomWebClientFromContext(context, servers);
+      const listOfUIDs = getSOPInstanceReferenceFromActiveViewport(viewports);
+      return save(
+        downloadAndZip(dicomWebClient, listOfUIDs, { progress }),
+        listOfUIDs
+      );
+    },
+  };
+
+  const definitions = {
+    downloadAndZip: {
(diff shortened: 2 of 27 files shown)


extensions/dicom-p10-downloader/README.md
```diff
@@ -0,0 +1 @@
+# @ohif/extension-dicom-pdf
```
dannyrb: Fix name

extensions/dicom-p10-downloader/src/commandsModule.js
```diff
@@ -0,0 +1,47 @@
...
+  getDicomWebClientFromContext,
+  getSOPInstanceReferenceFromActiveViewport,
+} from './utils';
+import downloadAndZip from './downloadAndZip';
+
+export function getCommands(context) {
+  const actions = {
+    downloadAndZipSeriesOnActiveViewport({ servers, viewports, progress }) {
+      const dicomWebClient = getDicomWebClientFromContext(context, servers);
+      if (dicomWebClient) {
```
swederik: Shouldn't we just do if (!dicomWebClient) throw new Error...

Failing silently doesn't seem good here

extensions/dicom-p10-downloader/src/commandsModule.js
```diff
@@ -0,0 +1,47 @@
...
+        const reference = getSOPInstanceReferenceFromActiveViewport(viewports);
+        if (reference) {
+          downloadAndZip(dicomWebClient, reference, { progress })
+            .then(url => {
+              OHIF.log.info('Zip file successfully created:', url);
+              location.href = url;
```
swederik: Is this going to redirect the current tab? We don't want that, it might disrupt their work. Better to force a download. This is how we do it with the SaveAs tool in Cornerstone: https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/util/SaveAs.js#L34

extensions/dicom-p10-downloader/src/downloadAndZip.js
```diff
@@ -0,0 +1,190 @@
...
+  // Print tree of hierarchical references
+  OHIF.log.info('Downloading DICOM P10 files for references:');
+  OHIF.log.info(hierarchicalListUtils.print(listOfUIDs));
+
+  // Wait on created download promises
+  return Promise.all(downloads).then(results => {
+    const buffers = [];
+    // The "results" array may directly contain buffers (ArrayBuffer instances)
+    // or arrays of buffers, depending on the type of downloads initiated on the
+    // previous step (retrieveStudy, retrieveSeries or retrieveinstance). Ex:
```
swederik: This is making me think we should change dicomweb client later to always return an array for consistency.

extensions/dicom-p10-downloader/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+   * Only required property. Should be a unique value across all extensions.
+   */
+  id: 'dicom-p10-downloader',
+
+  /**
+   * LIFECYCLE HOOKS
+   */
+
+  preRegistration({ appConfig }) {
+    const dicomWebClient = getDicomWebClientFromConfig(appConfig);
```
swederik: So if the server ever changes, this will stop working? (not that we can currently change servers, just curious)

platform/viewer/src/routes/routesUtil.js
```diff
@@ -19,7 +19,9 @@ const StudyListRouting = asyncComponent(() =>
   )
 );
 const StandaloneRouting = asyncComponent(() =>
-  import(/* webpackChunkName: "ConnectedStandaloneRouting" */ '../connectedComponents/ConnectedStandaloneRouting.js')
```
swederik: Unnecessary change you can remove from this PR. We can lint all files later

platform/viewer/public/config/default.js
```diff
@@ -115,6 +115,11 @@ window.config = {
       label: 'W/L Preset 9',
       keys: ['9'],
     },
+    {
+      commandName: 'downloadAndZipSeriesOnActiveViewport',
```
swederik: We do not want this enabled by default, you can take this out of the config.

dannyrb:
> Can we use the DICOM-formatted names for variables for consistency around our codebase (StudyInstanceUID, etc..). James just made a huge PR to keep this consistent in the rest of the codebase.

This is why we need a linter.

extensions/dicom-p10-downloader/src/index.js
emfol: Actually, this is just a "default" DICOM Web Client instance built based on startup config which is only used if the used does not provide its own DICOMWebClient instance. Also, the command from extensions/dicom-p10-downloader/src/commandsModule.js file (line #11, listed above) uses the information inside the servers store to find out information about the currently active server and builds a DICOM Web Client from there.

platform/viewer/public/config/default.js
emfol: OK. I'll remove it in the following PR update.

platform/viewer/src/routes/routesUtil.js
emfol: Yeah this freaking automatic linter keeps surprising me... I'll remove it. Thanks for that.

extensions/dicom-p10-downloader/src/commandsModule.js
emfol: Nope. This is a Blob URL that directly initiates the download and leaves the browsing session as is. But this was temporary and I have already changed this in order to give the file a name.

extensions/dicom-p10-downloader/src/commandsModule.js
emfol: Great point! I'll remove this conditional as this erroneous condition is properly handled (an exception is thrown) in thee lower layer.

extensions/dicom-p10-downloader/README.md
emfol: copy and paste hahaha

platform/viewer/public/config/default.js
```diff
@@ -115,6 +115,11 @@ window.config = {
       label: 'W/L Preset 9',
       keys: ['9'],
     },
+    {
+      commandName: 'downloadAndZipSeriesOnViewports',
```
emfol: @swederik, I left it here (along with the import extension on the "platform/viewer/src/index.js" file) in order to make the feature testable by the client. My plan is to remove it after the feature is approved and before it hits the master branch. Please let me know if you don't think it's necessary.

extensions/dicom-p10-downloader/src/commandsModule.js
```diff
@@ -0,0 +1,85 @@
...
+  save,
+  getDicomWebClientFromContext,
+  getSOPInstanceReferenceFromActiveViewport,
+  getSOPInstanceReferencesFromViewports,
+} from './utils';
+import downloadAndZip from './downloadAndZip';
+
+export function getCommands(context) {
+  const actions = {
+    // Example for running this command using Commands Manager
```
JamesAPetts: You can add examples using jsdoc using the `@example` tag, like [here](https://github.com/cornerstonejs/cornerstoneTools/blob/e38e2f5c81b0f83cca5a29a03334bee287028a8a/src/store/setToolMode.js#L23).

extensions/dicom-p10-downloader/src/commandsModule.js
```diff
@@ -0,0 +1,85 @@
...
+    //       progress(status) {
+    //         console.info('Progress:', (status.progress * 100).toFixed(2) + '%');
+    //       }
+    //     }
+    //   },
+    //   'VIEWER'
+    // );
+    downloadAndZip({ servers, dicomWebClient, listOfUIDs, options }) {
+      return save(
+        downloadAndZip(
```
JamesAPetts: Can you import this as an alias? Otherwise this function is hard to follow and looks like it is recursive.

extensions/dicom-p10-downloader/src/utils.js
```diff
@@ -0,0 +1,110 @@
...
+      list,
+      validDicomUid(StudyInstanceUID),
+      validDicomUid(SeriesInstanceUID),
+      validDicomUid(SOPInstanceUID)
+    );
+  });
+  return list;
+}
+
+function save(thenable, listOfUIDs) {
```
JamesAPetts: `thenable` => `promise` ?

dannyrb:
@jamesapetts, I think we were waiting on feedback for UI and the level of polish wanted before moving forward.

I'm also hesitant to put in any restrictions. I think those could/would be done by the UI. A future performance enhancement might push the zipping to a thread.

extensions/dicom-p10-downloader/src/commandsModule.js
emfol: Nice! I'll do that! Thank you!

extensions/dicom-p10-downloader/src/commandsModule.js
emfol: I see... Yes, I can do that.

extensions/dicom-p10-downloader/src/utils.js
emfol: In this case I use "thenable" to differentiate between promise-like objects and actual Promise objects. This method is capable of handling both (actual) promises and promise-like objects (i.e., a thenable) and returns an actual Promise instance.

emfol: Oops! Indeed... In this context, promise is better! I'll change it.

emfol:
@JamesAPetts, I pushed a few changes to prevent concurrent downloads and I could no longer reproduce the issue you reported (btw, you can track download progress from messages in console which, in the future, can be easily migrated to some sort of UI). Can you please try again? If you're able to reproduce it, please send me the link to the exam you tested with and as much information as you can about the issue.

JamesAPetts:
Hey @emfol , all looks good to me.

sebastianwoinar:
@JamesAPetts what happened to your extension? Any chance, that it can still be used with OHIF Viewer > 3.8?

PR #1502, opened 2020-03-10
feat: OHIF-22 - Study List Pagination

rodrigolabs:
## Description

Create a component to be used as Study List pagination.

## Preview

image


platform/ui/src/views/StudyList/components/StudyListPagination.js
```diff
@@ -1,7 +1,71 @@
...
+  Typography,
+  ButtonGroup,
+  Button,
+  IconButton,
+} from '../../../components/';
 
-const StudyListPagination = () => {
-  return <div>{`<- PAGINATION CONTENT ->`}</div>;
+const StudyListPagination = props => {
+  const [currentPage, setCurrentPage] = useState(1);
```
galelis: Following Danny's advices, shouldn't we be receiving the currentPage and setCurentPage from props, so we can basicaly display parent state of this component.

platform/viewer/src/App_bk.js
```diff
@@ -1,319 +0,0 @@
-import React, { Component } from 'react';
```
galelis: What is this file? hehe backup?

galelis: cleanup?

rodrigolabs: This file is a copy of the current (v1) `App.js` from the Viewer application.
Removing because it was left in history but it's not needed to be here.

PR #1504, opened 2020-03-10
OHIF-1: Study List Filter

galelis:
## Description
Create a UI component to be used on studi list page and display all filters

## Required specs 
Screen Shot 2020-03-10 at 22 37 06

## Developed screenshot
Screen Shot 2020-03-10 at 22 35 32


platform/ui/src/components/Input/Input.jsx
```diff
@@ -1,50 +1,47 @@
...
 const Input = ({
   label,
   containerClassName = '',
   labelClassName = '',
   className = '',
-  transparent = defaults.transparent,
-  ...rest
+  transparent = true,
+  id,
+  ...otherProps
```
dannyrb: Similar to feedback I gave @rodrigolabs earlier, I want to try and avoid spreading props. It's handy, but in larger apps it can make things very confusing / difficult to debug.

platform/ui/src/components/Input/Input.jsx
```diff
@@ -1,50 +1,47 @@
...
+  return (
+    <div className={classnames('flex flex-col flex-1', containerClassName)}>
+      <Label className={labelClassName} text={label} for={id ? id : ''}></Label>
+      <input
+        className={classnames(
+          className,
+          baseInputClasses,
+          transparentClasses[transparent]
+        )}
+        id={id ? id : ''}
```
dannyrb: If `id` defaults to empty string, set that as the default in `defaultProps`? If `id` has to be unique, should we auto-generate a string guid?

platform/ui/src/components/Input/Input.jsx
```diff
@@ -1,50 +1,47 @@
...
 
-const transparentClasses = {
-  true: 'bg-transparent',
-  false: '',
+Input.propTypes = {
+  label: PropTypes.string,
+  containerClassName: PropTypes.string,
+  labelClassName: PropTypes.string,
+  className: PropTypes.string,
+  transparent: PropTypes.bool,
```
dannyrb: Not sure these are the props/setup we want. Probably fine for pass 1, but should discuss before cementing component API in pass 2

platform/ui/src/components/Label/Label.jsx
```diff
@@ -1,32 +1,18 @@
...
-const Label = ({ children, spacing, className, text, ...rest }) => {
-  const baseClasses = 'flex flex-1 flex-col text-gray-700 text-sm font-bold';
-
-  const getClasses = () => {
-    const classes = [];
-    classes.push(baseClasses);
-    classes.push(className);
-    return classes.join(' ');
-  };
+const Label = ({ children, className, text, ...rest }) => {
```
dannyrb: No rest spread. Should probably add an eslint rule. @rodrigolabs, is there one floating around in the react eslint rules we can steal?

platform/ui/src/components/Input/Input.jsx
rodrigolabs: I've created a new PR adding this eslint rule: https://github.com/OHIF/Viewers/pull/1509

platform/ui/src/components/Label/Label.jsx
rodrigolabs: #1509

platform/ui/src/components/Button/Button.jsx
```diff
@@ -113,13 +113,13 @@ const Button = ({
...
       className={classnames(
+        className,
         baseClasses,
         variantClasses[variant][color],
         roundedClasses[rounded],
         sizeClasses[size],
         fullWidthClasses[fullWidth],
-        disabledClasses[disabled],
-        className
+        disabledClasses[disabled]
```
dannyrb: Not sure how much it matters, but if we have a lot of these, maybe should be alphabetic.
Order mostly matters for `className` first as override, and `baseClasses`, everything else shouldn't conflict?

platform/ui/src/views/StudyList/StudyList.js
```diff
@@ -8,7 +8,7 @@ const StudyList = () => {
   return (
     <div>
       <Header />
-      <StudyListFilter />
+      <StudyListFilter onFilterChange={() => {}} />
```
dannyrb: 👋

platform/ui/src/views/StudyList/components/StudyListFilter.js
```diff
@@ -1,7 +1,103 @@
...
+  mrn,
+  studyDate,
+  description,
+  modality,
+  accession,
+  instances,
+  numOfStudies = 0,
+  sortByFilter,
+  onSortByChange,
+  onFilterChange,
```
dannyrb: Most things like this split props:

- "Meta"
- "Value"

Where meta drives how things are displayed/formatted/rendered, and value is the value we're actively modifying and communicating to the parent.

Meta might look like:

```js
filtersMeta: [
  {
    name: 'patientName',
    displayName: 'Patient Name',
    inputType: 'text',
    isSortable: true
  }
]
```

With values similar to:

```js
filterValues: {
    // note: value is using "name" to match field meta
    patientName: "",
    mrn: "",
    studyDate: "",
    // These fields might be in a separate object
    // Probably still have to be props so we can support persisting them between routes (cached filter on back navigation)
    sort: "" // empty or field name
    sortDirection: 0, // -1, 0, 1
    page: 0,
    resultsPerPage: 25,
}
```

platform/ui/src/views/StudyList/components/StudyListFilter.js
```diff
@@ -1,7 +1,103 @@
...
+  mrn: PropTypes.string,
+  studyDate: PropTypes.any,
+  description: PropTypes.string,
+  modality: PropTypes.string,
+  accession: PropTypes.string,
+  instances: PropTypes.any,
+  numOfStudies: PropTypes.number,
+  sortByFilter: PropTypes.string,
+  onSortByChange: PropTypes.func,
+  onFilterChange: PropTypes.func.isRequired,
```
dannyrb: Should have default values of empty string if not required. In general, try to hazard to required more than not-required. From a testing perspective, it's one less case to test.

platform/ui/src/components/Button/Button.jsx
galelis: it actually the order here does not metter, I did this to try to fix label size, but didn't work and I forget to revert. the

platform/ui/src/views/StudyList/components/StudyListFilter.js
galelis: Change the code to something similar to this.

platform/ui/src/views/StudyList/StudyList.js
galelis: Removed for now.

platform/ui/src/components/Input/Input.jsx
galelis: I think we can improve a lot these Input/Label component.

platform/ui/src/components/Input/Input.jsx
galelis: Removed the change for now.

PR #1515, opened 2020-03-12
feat: Data Mocking for Study List

rodrigolabs:
**OHIF-18**
**OHIF-38**

## Description

It creates a data mocking mechanism that can be used to view the Study List with different data.

## Preview

**No Studies**

image

**30 Studies**

image

**300 Studies**

image


platform/ui/src/views/StudyList/components/StudyListTable.js
```diff
@@ -9,7 +11,29 @@ import InstancesActive from '../../../assets/icons/instances-active.svg';
 import InstancesInactive from '../../../assets/icons/instances-inactive.svg';
 import LaunchInfo from '../../../assets/icons/launch-info.svg';
 
-const TableRow = () => {
+const getStudyModalities = series => {
```
dannyrb: Assume you are getting the data in the correct format. No need to compute it here.

platform/ui/src/views/StudyList/components/StudyListTable.js
```diff
@@ -9,7 +11,29 @@ import InstancesActive from '../../../assets/icons/instances-active.svg';
...
+    }
+
+    acc.push(Modality);
+    return acc;
+  }, []);
+
+  return modalities.join('/');
+};
+
+const getInstances = series => {
```
dannyrb: Same here. All data transformations from a "domain" model to a "view model" should happen before components touch data.

platform/ui/src/views/StudyList/StudyList.js
```diff
@@ -1,18 +1,55 @@
...
 import StudyListFilter from './components/StudyListFilter';
 import StudyListTable from './components/StudyListTable';
 import StudyListPagination from './components/StudyListPagination';
 
-const StudyList = () => {
+import studyListMock from '../../mocks/studyList.json';
+
+/** Values can be env vars */
+const DEFAULT_MOCKED_STUDIES_NUM = 50;
+const DEFAULT_MOCKED_STUDIES_LIMIT = 1000;
```
dannyrb: Don't bake this into the component. This would happen in docz or somewhere similar. Potentially with throw-away ui to change data. Might be worth looking at React powerplug?

dannyrb: Hmm.. Unless we are considering this top level container throwaway. Worth discussing. But baking this in makes copy-pasting general structure over slightly more difficult

platform/ui/src/views/StudyList/StudyList.js
```diff
@@ -1,18 +1,55 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+
 import Header from './components/Header';
 import StudyListFilter from './components/StudyListFilter';
 import StudyListTable from './components/StudyListTable';
 import StudyListPagination from './components/StudyListPagination';
 
-const StudyList = () => {
+import studyListMock from '../../mocks/studyList.json';
```
dannyrb: Is this checked in?

platform/ui/src/views/StudyList/components/StudyListTable.js
rodrigolabs: I have changed the JSON mock file to accomplish the correct data. Now the component is only receiving and displaying the data as it is.

platform/ui/src/views/StudyList/StudyList.js
rodrigolabs: I have changed this logic. Now the mocking data logic is only tight to Docz.

platform/ui/index.js
```diff
@@ -9,4 +11,6 @@ export {
   Typography,
 } from './src/components';
 
+export { utils };
```
dannyrb: This is okay for now, but we should have a better name than `utils`. That namespace quickly becomes a catch-all bucket.

platform/ui/src/components/Icon/getIcon.jsx
```diff
@@ -1,28 +1,32 @@
 import React from 'react';
 // Icons
-import active from './../../assets/icons/active.svg';
+import seriesActive from './../../assets/icons/series-active.svg';
+import seriesInactive from './../../assets/icons/series-inactive.svg';
```
dannyrb: We don't have to on this pass, but we should use generic names that describe these, and not application specific names.

platform/ui/src/utils/getInstances.js
```diff
@@ -0,0 +1,15 @@
...
+
+const getInstances = series => {
+  const instances = series.reduce((acc, item) => {
+    return acc + item.instances.length;
+  }, 0);
+
+  return instances;
+};
+
+export default getInstances;
```
dannyrb: If this is no longer used, can we remove it?

platform/ui/src/utils/getModalities.js
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    acc.push(Modality);
+    return acc;
+  }, []);
+
+  return modalities.join('/');
+};
+
+export default getModalities;
```
dannyrb: If this is no longer used, can we nuke it?

platform/ui/src/views/StudyList/components/StudyListTable.js
```diff
@@ -50,54 +54,65 @@ const TableRow = () => {
                   onClick={toggleRow}
                 >
                   <td className={classnames(...tdClasses)}>
-                    <ChevronIcon />
+                    <Icon name={ChevronIconName} />
+                  </td>
+                  <td className={classnames(...tdClasses)}>{PatientName}</td>
```
dannyrb: `classnames` already supports an object? why spread?

platform/ui/src/views/StudyList/components/StudyListTable.js
```diff
@@ -170,27 +185,85 @@ const TableRow = () => {
...
+  studies: PropTypes.arrayOf(
+    PropTypes.shape({
+      AccessionNumber: PropTypes.string.isRequired,
+      Modalities: PropTypes.string.isRequired,
+      Instances: PropTypes.number.isRequired,
+      PatientId: PropTypes.string.isRequired,
+      PatientName: PropTypes.string.isRequired,
+      StudyDescription: PropTypes.string.isRequired,
+      StudyDate: PropTypes.string.isRequired,
+      series: PropTypes.array.isRequired,
```
dannyrb: Need to discuss DTO naming convention with @JamesAPetts

platform/ui/src/views/StudyList/components/StudyListTable.js
```diff
@@ -50,54 +54,65 @@ const TableRow = () => {
                   onClick={toggleRow}
                 >
                   <td className={classnames(...tdClasses)}>
-                    <ChevronIcon />
+                    <Icon name={ChevronIconName} />
+                  </td>
+                  <td className={classnames(...tdClasses)}>{PatientName}</td>
+                  <td className={classnames(...tdClasses)}>{PatientId}</td>
+                  <td className={classnames(...tdClasses)}>
+                    {format(StudyDate, 'MMM-DD-YYYY')}
```
dannyrb: I think we'll force the parent to handle this formatting. You can ditch `date-fns` here.

platform/ui/src/views/StudyList/components/StudyListTable.js
```diff
@@ -170,27 +185,85 @@ const TableRow = () => {
...
-              ))}
-            </tbody>
-          </table>
+          {numOfStudies > 100 && (
+            <div className="bg-custom-blue text-center text-base py-1 rounded-b sticky top-0">
+              <p className="text-white">
+                Filter list to 100 studies or less to enable sorting
+              </p>
+            </div>
+          )}
```
dannyrb: This will probably be a render prop in finalized version. The ">100" constraint only applies to DICOMWeb server sources, so the parent may use this as a conditional target to add an "is loading" message, or a display a "banner notification".

Something we can discuss, but hesitant to hardcode "100" studies here.

platform/ui/src/views/StudyList/components/StudyListTable.js
dannyrb: Gist is, we enforce capitalization of DICOM Tag names to be consistent with DICOM naming conventions; but does that change when the value deviates from what is stored?

IE. These are "display" representations of the original values. This object is a "data transfer object" (DTO), or you can think of it as a "view model"

JamesAPetts: IMO to avoid confusing we shouldn't just change the casing of these then. Anything non-naturalized DICOM tag format should be explictly stated? I.e. `studyDateDisplay` ? Don't care too much on the naming convention but anything which is formatted to be readable etc needs to be firmly differentiated?

dannyrb: That makes sense. So consider tag names "reserved" unless they're immutable value representations?

JamesAPetts: I don't think they need to be immutable, there are scenarios where you want to "fix" the headers.

PR #1521, opened 2020-03-13
fix: UI v2 Pagination Fixes

rodrigolabs:
**OHIF-40**

## Description

- Per page `select` is now using an icon to make it be displayed with the same style in all browsers.
- Buttons padding/width is fixed.

## Preview

image

**(expected)**

image

_There's a slight fix to do in the arrow-down icon. I'll ask Dan to export the correct icon for us to be used in the `select`_.

### DIFF
--- a/platform/ui/src/views/StudyList/components/StudyListPagination.js
+++ b/platform/ui/src/views/StudyList/components/StudyListPagination.js
@@ -4,6 +4,7 @@ import {
   ButtonGroup,
   Button,
   IconButton,
+  Icon,
 } from '../../../components/';
 
 const StudyListPagination = props => {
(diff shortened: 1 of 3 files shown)


platform/ui/src/views/StudyList/components/StudyListPagination.js
```diff
@@ -39,21 +50,24 @@ const StudyListPagination = props => {
               <ButtonGroup color="primary">
                 <IconButton
                   size="small"
-                  className="border-custom-darkSlateBlue px-4"
+                  className="border-custom-darkSlateBlue px-4 py-2 text-base"
+                  style={{ padding: '3px 12px', minWidth: 0 }}
```
rodrigolabs: I'm not 100% comfortable to have inline styles. This is a specific/temporary case. Just to keep in mind that ideally, it shouldn't be done. 🙄

dannyrb: No worries. If we're forced to do this, worth mentioning to Dan. Especially if this doesn't fit the same padding increments as the rest of the design. Also, select's are hard to style

Issue #1373, opened 2020-01-14
Support of RTSTRUCT visualization


fedorov:
## Request

**What feature or change would you like to see made?**

Basic visualization of RTSTRUCT structure sets in the image plane.

**Why should we prioritize this feature?**

As we discussed and planned for the IDC MVP functionality, we thought the following should be feasible:

> Support of RTSTRUCT and SEG will be limited to in-plane visualization, not in MPR view. Visualization will be supported as semi-transparent overlay or contour. Interface will allow scrolling to the centroid of the individual structures, and toggle of the visibility for the individual structures/segments.

Submitting this to the tracker so it can be prioritized, as @swederik asked me to do.

Great example of similar capabilities implemented using OHIF: https://econtour.org/

JamesAPetts:
Hey @fedorov, I'm going to start working on this, some questions:

> Visualization will be supported as semi-transparent overlay or contour.

For the MVP, is a contour representation for RTSTRUCTs ok? In the long term it would be nice to have master formats and interoperability between different formats, but contours are the easiest route for an initial implementation.

Is there a set goal for how large an RTSTRUCT should be able to be parsed in a reasonable amount of time? The standard allows for entire studies of regions to be encoded, it'd be good to have an idea of what target is deemed reasonable.

> Interface will allow scrolling to the centroid of the individual structures, and toggle of the visibility for the individual structures/segments.

This is all easily doable.

JamesAPetts:
More thoughts:

Labels should ideally be read from the `RT ROI Observations Sequence`, however most of the useful information in here is type 3.

Do we also want `RT Related ROI Sequence (3006,0030)` information present in the UI on the first pass? Its tricky as this defines nice hierarchies with could be rendered in useful ways, but the sequence is type 3. Perhaps we should just show this information (if present) on mouse hover?

Additionally is visualisation other relevant type 3's such as:
`Anatomic Region Sequence` and the `Primary Anatomic Structure Sequence` high priority?

fedorov:
@JamesAPetts thank you for the update!

> For the MVP, is a contour representation for RTSTRUCTs ok?

Yes, absolutely.

> Is there a set goal for how large an RTSTRUCT should be able to be parsed in a reasonable amount of time? The standard allows for entire studies of regions to be encoded, it'd be good to have an idea of what target is deemed reasonable.

[...]

PR #1524, opened 2020-03-13
Support of RTSTRUCT visualization

igoroctaviano:
#1373: Basic visualization of RTSTRUCT structure sets in the image plane.

### PR Checklist


igoroctaviano:
Waiting for study with one or more struct sets. @JamesAPetts

extensions/dicom-rt/src/components/RTPanel/RTPanel.js
```diff
@@ -0,0 +1,211 @@
...
+    selectedSet: null,
+  };
+
+  const [state, setState] = useState(DEFAULT_STATE);
+  const [showSettings, setShowSettings] = useState(false);
+
+  useEffect(() => {
+    document.addEventListener('rtloaded', updateStructureSets);
+
+    return () => {
```
igoroctaviano: @dannyrb Ill need your help figuring out a way to handle this. The user can open the panel while the rtstruct sets are being loaded, if it happens, the component its not gonna react to the changes, considering the current implementation of extensions / rtload logic, any ideas?

dannyrb: In an ideal world, subscribe to `MeasurementService` in a `useEffect` hook.
Then update local state when subscriptions are fired?

For now, even though this is hacky, I think its your only option given time constraints/budget.

My recommendation: Document that this is not ideal (and articulate why), and how you would recommend improving it. Spend less than 45 minutes, but do try to capture the essence of the problem and the value of resolving it so that a future implementer isn't lost, and it's easier to prioritize in our backlog.

@JamesAPetts may weigh in differently. Just my 2 cents.

JamesAPetts: For the sake of namespacing and adding event listeners to `document`, should probably call the event `extensiondicomrtrtloaded`, so its unique enough.

igoroctaviano: Updated. thanks.

extensions/dicom-rt/src/components/LoadingIndicator/LoadingIndicator.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
```
JamesAPetts: @dannyrb Is this not a component somewhere we can just import instead of copying here?

JamesAPetts: Linked wrong line, this whole file I mean.

platform/core/src/studies/services/wado/studyInstanceHelpers.js
```diff
@@ -145,6 +145,7 @@ async function makeSOPInstance(server, study, instance) {
     wadoRoot: server.wadoRoot,
     imageRendering: server.imageRendering,
     thumbnailRendering: server.thumbnailRendering,
+    _instanceRaw: instance,
```
JamesAPetts: We no longer need this since the metadata update.

platform/ui/src/components/tableList/TableList.js
```diff
@@ -9,14 +9,21 @@ export class TableList extends Component {
     defaultItems: PropTypes.object,
     children: PropTypes.node.isRequired,
     headerTitle: PropTypes.string,
+    headless: PropTypes.bool,
+  };
+
+  static defaultProps = {
+    headless: false,
```
JamesAPetts: I'm a fan of improving the default UI components so we don't fork every little thing, you fine with this change to @ohif/ui @dannyrb?

platform/ui/src/elements/Icon/getIcon.js
```diff
@@ -85,8 +85,12 @@ import trash from './icons/trash.svg';
 import unlink from './icons/unlink.svg';
 import user from './icons/user.svg';
 import youtube from './icons/youtube.svg';
+import eye from './icons/eye.svg';
+import eyeClosed from './icons/eye-closed.svg';
 
 const ICONS = {
+  eye,
+  'eye-closed': eyeClosed,
```
JamesAPetts: @dannyrb Just a note, but not related to this PR, as this one is generic. I don't think we have a good way of adding icons via extensions?

They need to be registered to the Icon list and then be usable everywhere?

For example when adding tools to the tool bar they just reference an icon's name from this list, so you can't even inject an element at that level currently?

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
```diff
@@ -425,6 +425,56 @@ describe('OHIF Cornerstone Toolbar', () => {
     });
   });
 
+  it('checks if Eraser tool will remove the measurements added in the viewport', () => {
```
JamesAPetts: We removed the erase tool, so you can remove this test, must have came from a dodgey merge.

platform/viewer/public/config/default.js
```diff
@@ -1,120 +1,149 @@
...
-    },
-    {
-      commandName: 'windowLevelPreset9',
-      label: 'W/L Preset 9',
-      keys: ['9'],
+      // ~ REQUIRED
+      // Authorization Server URL
+      authority: 'https://accounts.google.com',
+      client_id:
+        '723928408739-k9k9r3i44j32rhu69vlnibipmmk9i57p.apps.googleusercontent.com',
```
JamesAPetts: We shouldn't commit this change.

igoroctaviano: My bad, updated. Thanks.

platform/viewer/cypress/integration/common/OHIFCornerstoneToolbar.spec.js
igoroctaviano: Removed, thanks.

extensions/dicom-rt/src/components/LoadingIndicator/LoadingIndicator.js
igoroctaviano: @JamesAPetts Although they share the same name this component is totally different than the one in the UI project. This is an animated SVG of a loading indicator that I think it's better than the icon we have today.  Currently, the Icon component doesn't work with animated SVGs. @dannyrb what do you think about having this one as the new OHIF default?

igoroctaviano:
> Some minor comments added.
> 
> The app currently breaks on the latest commit when I open the sidebar though. Let me know when its fixed so I can test the app, I've tested from ~3 commits ago and what was there is nice.
> 
> Also I can't test too well because of said sidebar issue, but it appears both of the rtstructs are being loaded automatically instead of just the most recent on https://deploy-preview-1524--ohif.netlify.com/pwa/viewer/1.3.6.1.4.1.14519.5.2.1.2193.7172.847236098565581057121195872945?

Fixed the issue in RT panel, it was introduced in the last commit. About the load, you can see that the eye will not appear for sets that are not loaded yet.

extensions/dicom-rt/src/components/LoadingIndicator/LoadingIndicator.js
JamesAPetts: @dannyrb Poking you for this, although I'm not bothered about it for this particular PR.

igoroctaviano:
> Had a play with this on IDC, and checked over the code, LGTM.
> 
> Could we just hide the relabel/description UI on each structure for now, as they don't currently do anything whilst this module is read only.
> 
> It can be merged after that imo, nice job.

@JamesAPetts 
Done. Thanks for the review.

PR #1525, opened 2020-03-13
fix: UI v2 Study List style fixes

rodrigolabs:
**OHIF-39**
**OHIF-36**

## Description

Some minor Study List style fixes.

## Preview

image

**Sticky Header + Filter**

Mar-16-2020 12-32-42

### DIFF
--- a/platform/ui/src/views/StudyList/components/StudyListFilter.js
+++ b/platform/ui/src/views/StudyList/components/StudyListFilter.js
@@ -178,8 +178,8 @@ const StudyListFilter = ({
           </div>
         </div>
       </div>
-      <div className="sticky z-10" style={{ top: 58 }}>
-        <div className="bg-custom-navyDark pt-3 pb-3 border-b-4 border-black">
+      <div className="sticky z-10 border-b-4 border-black" style={{ top: 58 }}>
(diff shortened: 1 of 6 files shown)


platform/ui/src/views/StudyList/components/StudyListFilter.js
```diff
@@ -124,87 +124,115 @@ const StudyListFilter = ({
...
+                >
+                  Studies
+                </Typography>
+              </div>
+            </div>
+          </div>
         </div>
       </div>
-    </div>
+      <div className="sticky z-10" style={{ top: 58 }}>
```
galelis: I am a bit concerned on hardcoding this style to make the sticky working. Once we componentize it and make it reusable it might break.

rodrigolabs:
> * The “Filter List” notification bar should have space between it and the list results

You're right! I fixed the border to be in the same container that has the study list filter + notification. Thanks! 😄

platform/ui/src/views/StudyList/components/StudyListFilter.js
rodrigolabs: Totally agree! Maybe when we break it into a small component, we can add props like:

```javascript
<Component sticky stickyOffset="58" />
```

What do you think?

PR #1529, opened 2020-03-16
OHIF-32: Wrapping React-select with our style requirements

galelis:
### Description 
- Wrapp React-select with our style requirements

### Preview
wraped-select


platform/ui/src/components/Select/Select.css
```diff
@@ -0,0 +1,31 @@
+.customSelect__wrapper .customSelect__control {
+  @apply bg-black border-custom-blue shadow transition duration-300 border rounded w-full text-sm text-white leading-tight;
```
rodrigolabs: I'm truly in love with the power Tailwind is giving us even to customize 3rd party libraries. It's awesome :)

platform/ui/src/components/Select/Select.jsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <ReactSelect
+      autoFocus={autoFocus}
+      className={classnames(
+        className,
+        'flex flex-col flex-1 hover:cursor-pointer mt-2 customSelect__wrapper'
+      )}
+      classNamePrefix="customSelect"
+      isDisabled={isDisabled}
+      isMulti={isMulti}
```
rodrigolabs: When `isMulti` is enabled, it's good to check how the layout should behave. We could share with Dan asking his opinion.

Temporary preview:

image

platform/ui/src/components/Select/Select.mdx
```diff
@@ -0,0 +1,20 @@
...
+name: Select
+menu: Components
+route: components/select
+---
+
+import { Playground, Props } from 'docz';
+import Select from './';
+
+# Select
+
```
rodrigolabs: Can you add here the `Import` section?

galelis: Good call. also added in some other components that was missing.

platform/ui/src/components/Select/Select.jsx
galelis: Thast a good point. 
I haven't asked before cause our use of it does not hit the mult select feature yet.

platform/ui/src/components/Select/Select.jsx
```diff
@@ -0,0 +1,63 @@
...
+  isSearchable: PropTypes.bool,
+  name: PropTypes.string,
+  onChange: PropTypes.func,
+  options: PropTypes.arrayOf(
+    PropTypes.shape({
+      value: PropTypes.string,
+      label: PropTypes.string,
+    })
+  ),
+  placeholder: PropTypes.any,
```
rodrigolabs: Isn't placeholder a string?

platform/ui/src/components/Select/Select.jsx
```diff
@@ -0,0 +1,63 @@
...
+}) => {
+  return (
+    <ReactSelect
+      autoFocus={autoFocus}
+      className={classnames(
+        className,
+        'flex flex-col flex-1 hover:cursor-pointer mt-2 customSelect__wrapper'
+      )}
+      classNamePrefix="customSelect"
+      isDisabled={isDisabled}
```
rodrigolabs: It'd be good to add a custom style like `cursor-not-allowed` when `isDisabled` is true.

platform/ui/src/components/Select/Select.jsx
```diff
@@ -0,0 +1,63 @@
...
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import ReactSelect from 'react-select';
+
+import './Select.css';
+
+const Select = ({
+  autoFocus,
+  className,
+  classNamePrefix,
```
rodrigolabs: This is not being used anywhere. Maybe it can be removed from props.

platform/ui/src/components/Select/Select.jsx
```diff
@@ -0,0 +1,63 @@
...
+  isDisabled,
+  isMulti,
+  isSearchable,
+  name,
+  onChange,
+  options,
+  placeholder,
+  noOptionsMessage,
```
rodrigolabs: I didn't understand what it does, can you give me a quick example on how to use that?

rodrigolabs:
LGTM -- Great job, dude! I just left some quick questions.

platform/ui/src/components/Select/Select.jsx
galelis: Weird way to update it, but its done :)

platform/ui/src/components/Select/Select.jsx
galelis: Make sense, I am using it to force the classes to be the same so I can update the css using tailwind.

platform/ui/src/components/Select/Select.jsx
galelis: I have added an example into the docz. also update the code, it was a bit wrong. :) 
thanks for the callout

platform/ui/src/components/Select/Select.jsx
galelis: True, hehe, updated.

platform/ui/src/components/Select/Select.jsx
rodrigolabs: Cool! Just checked, I was trying to pass a string instead of a node.

We'll just have to change its prop type definition. `noOptionsMessage` should expect a `node` instead of a `func`.

platform/ui/src/components/Select/Select.jsx
```diff
@@ -0,0 +1,61 @@
...
+  name: PropTypes.string,
+  onChange: PropTypes.func,
+  options: PropTypes.arrayOf(
+    PropTypes.shape({
+      value: PropTypes.string,
+      label: PropTypes.string,
+    })
+  ),
+  placeholder: PropTypes.string,
+  noOptionsMessage: PropTypes.func,
```
rodrigolabs:
```suggestion
  noOptionsMessage: PropTypes.node,
```

rodrigolabs:
@galelis There's a merge conflict. Also, we just need a minor change in the `noOptionsMessage` and it's good to be merged \o/

PR #1537, opened 2020-03-19
feat: ui v2 - Date Range Component

rodrigolabs:
**OHIF-33**

## Description

Create a reusable Date Range component in the v2 library.

It's using the same library being used in the current @ohif/ui version.

The component is very simplified and usage is really simple -- we just need pass three required props and `DateRange` will do everything.

## Preview

image


platform/ui/src/components/DateRange/DateRange.css
```diff
@@ -0,0 +1,107 @@
...
+
+.DateInput {
+  background: transparent;
+  @apply flex flex-1 w-auto;
+}
+
+/** INPUT FIELD COMMON STYLES **/
+
+.DateInput_input {
+  background-image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><g fill="%236b6b6b" fill-rule="evenodd"><path d="M20 20h-4v-4h4v4zm-6-10h-4v4h4v-4zm6 0h-4v4h4v-4zm-12 6h-4v4h4v-4zm6 0h-4v4h4v-4zm-6-6h-4v4h4v-4zm16-8v22h-24v-22h3v1c0 1.103.897 2 2 2s2-.897 2-2v-1h10v1c0 1.103.897 2 2 2s2-.897 2-2v-1h3zm-2 6h-20v14h20v-14zm-2-7c0-.552-.447-1-1-1s-1 .448-1 1v2c0 .552.447 1 1 1s1-.448 1-1v-2zm-14 2c0 .552-.447 1-1 1s-1-.448-1-1v-2c0-.552.447-1 1-1s1 .448 1 1v2z"/></g></svg>');
```
rodrigolabs: I ended up using this `data:image` as background-image for the DateInput because I tried to use the relative svg url and it didn't work. I tried to check the `gatsby-plugin-react-svg` configuration but it didn't work as well. To avoid spending much time in an icon that is pretty much static/won't change, I did this way.

platform/ui/package.json
```diff
@@ -29,16 +29,18 @@
...
     "gatsby-plugin-sass": "^2.1.28",
     "gatsby-theme-docz": "^2.2.0",
+    "moment": "^2.24.0",
     "node-sass": "^4.13.1",
     "prop-types": "^15.7.2",
     "react": "16.11.0",
-    "react-dom": "16.11.0"
+    "react-dates": "^21.8.0",
+    "react-dom": "16.11.0",
+    "theme-ui": "^0.2.38"
```
dannyrb: Can I get a quick run down of what each of these deps is for?

platform/ui/src/components/DateRange/DateRange.css
```diff
@@ -0,0 +1,107 @@
...
+  background-size: 14px;
```
dannyrb: Good call. Thanks for specifically calling this out. An inline comment wouldn't hurt either.

platform/ui/src/components/DateRange/DateRange.jsx
```diff
@@ -0,0 +1,159 @@
+/** REACT DATES */
+import 'react-dates/initialize';
+import 'react-dates/lib/css/_datepicker.css';
+import { DateRangePicker, isInclusivelyBeforeDay } from 'react-dates';
+import './DateRange.css';
+
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
+import moment from 'moment';
```
dannyrb: Is this a dependency of the 3rd party library, or our own? Trying to think if there's a good way to set this as a peer dependency or make it possible for the consumer to make this decision

platform/ui/src/components/DateRange/DateRange.jsx
```diff
@@ -0,0 +1,159 @@
...
+  endDate: null,
+};
+
+DateRange.propTypes = {
+  /** Start date moment object */
+  startDate: PropTypes.object, // moment date is an object
+  /** End date moment object */
+  endDate: PropTypes.object, // moment date is an object
+  /** Callback that returns on object with selected dates */
+  onChange: PropTypes.func.isRequired,
```
dannyrb: Maybe set all is required -- no defaults. Show example usage in MDX. Allows you to sidestep moment dependency. Set as devDependency.

platform/ui/src/views/StudyList/StudyList.js
```diff
@@ -55,7 +55,7 @@ const filtersMeta = [
     displayName: 'Instances',
     inputType: 'none',
     isSortable: false,
-    gridCol: 3,
+    gridCol: 2,
```
dannyrb: Still need to revisit this usage. Different ticket.

platform/ui/src/components/DateRange/DateRange.jsx
rodrigolabs: `react-dates` use Moment object as date values for `startDate` and `endDate`.

I set `startDate` and `endDate` as optional because their value can be null when the field is empty.

I'm using `moment` in our `DateRange` component to mount the custom Filter Presets (Today, Last 7 Days, Last 30 Days) and to render the customized month/year select. These fields are customized in our side, but I'm not exposing them as props of our `DateRange` to keep the wrapped component as simple as possible.

What do you think we can do here?

platform/ui/src/components/DateRange/DateRange.jsx
rodrigolabs: As react-dates uses `moment` objects for date values, I'm using `moment` as our dependency to create the custom Filter Presets (Today, Last 7 Days, Last 30 Days) and to render the customized month/year select (as described below).

platform/ui/package.json
rodrigolabs: Actually only `react-dates` was added.

`theme-ui` is a lib that allows us to customize the `docz` theme. It was already added.

PR #1542, opened 2020-03-20
feat: v2 Table Component

rodrigolabs:
**OHIF-50**
**OHIF-52**

## Preview

image


platform/ui/src/views/StudyList/components/StudyListTable.js
```diff
@@ -275,7 +239,7 @@ const StudyListTable = ({ studies, numOfStudies, filtersMeta }) => {
       <table className="w-full text-white">
```
galelis: This table are gonna stay as html table or was it missing?

rodrigolabs: Yes, this is gonna be the same as it is for now. As it's a very customized table, with expanded row, different styles, etc... we've decided to only make a reusable table for this one:

image

platform/ui/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,41 @@
...
+        className
+      )}
+      style={style}
+    >
+      {children}
+    </div>
+  );
+};
+
+Table.defaultProps = {
```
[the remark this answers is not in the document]
dannyrb: I think this is the standard/recommended approach.

platform/ui/index.js
```diff
@@ -1,22 +1,6 @@
 import * as utils from './src/utils/';
-import {
-  ModalProvider,
-  ModalConsumer,
-  useModal,
-  withModal,
-} from './src/contextProviders';
+export * from './src/components';
+export * from './src/contextProviders';
```
dannyrb: While this may be less code, it's also less clear (at least to me) what we're exposing as public when we use `*`

platform/ui/src/components/Table/Table.mdx
```diff
@@ -0,0 +1,111 @@
...
+import { Table, TableHead, TableBody, TableRow, TableCell } from '@ohif/ui';
+
+# Table
+
+Tables display sets of data.
+
+## Import
+
+```javascript
+import { Table } from '@ohif/ui';
```
dannyrb: Is it weird to use an already existing HTML tag here? Should we do something that clearly sets them apart?

platform/ui/src/components/TableCell/TableCell.mdx
```diff
@@ -0,0 +1,36 @@
...
+---
+
+import { Playground, Props, Link } from 'docz';
+import { TableCell } from '@ohif/ui';
+
+# Table Cell
+
+## Import
+
+```javascript
```
dannyrb: If these doc pages don't make sense in isolation, do this instead:

- Create a new "Molecules" section in the navigation bar
- Show how all components work together to create a table

No reason to have unhelpful doc pages. Who uses `TableCell` in isolation?

platform/ui/src/components/TableHead/TableHead.jsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <div
+      className={classnames(
+        'bg-custom-navy border-b border-custom-violetPale flex font-bold pr-2',
+        className
+      )}
+      style={style}
+    >
+      {React.cloneElement(children, {
+        isTableHead: true,
```
dannyrb: This assumes TableHead receives a children of a specific component type. Should/can we check for this?

platform/ui/src/components/TableRow/TableRow.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const TableRow = ({ children, className, isTableHead, style }) => {
+  return (
+    <div className={classnames('w-full flex', className)} style={style}>
+      {React.Children.map(children, child =>
+        React.cloneElement(child, { isTableHead })
```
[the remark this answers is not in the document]
dannyrb: If we're only supporting "text" or raw html for now, is there any need to expose `TableCell`?

platform/ui/src/gatsby-theme-docz/components/Layout/index.js
```diff
@@ -36,6 +36,7 @@ export const Layout = ({ children }) => {
             onClick={() => setSidebarOpen(false)}
           />
           <div
+            id="main-container"
```
dannyrb: Where/how is this used?

platform/ui/src/gatsby-theme-docz/theme.css
```diff
@@ -6,7 +6,11 @@
   background: #000;
 }
 
-[data-testid='main-container'] > h2 {
-  margin: 20px 0;
-  font-weight: bold;
+#main-container > h2 {
```
dannyrb: Ah. Docz theme only. That's fine.

platform/ui/index.js
rodrigolabs: It makes sense, I agree. I'll fix this file to expose all the exports explicitly.

platform/ui/src/gatsby-theme-docz/components/Layout/index.js
rodrigolabs: Only for Docz theme

platform/ui/src/components/TableCell/TableCell.mdx
rodrigolabs: The idea of having this doc page was to basically display its component props. For a full example, in the `Basic usage` section, it's linking to the Table example. 

Your suggestion is to move `Table` to the `Molecules` menu and remove the `TableCell` doc page, for example? How we could demonstrate its props in this case?

platform/ui/src/components/Table/Table.jsx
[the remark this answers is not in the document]
galelis: Hmm I have asked because I've seeing 3 ways in the code for ohif/ui right now. 
I will create a new PR to make them all using `Component.defaultprops = {}`.

platform/ui/src/components/TableRow/TableRow.jsx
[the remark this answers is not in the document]
rodrigolabs: I've created a condition to handle how the children should be rendered. If it receives a string or a custom content, it will not break the application. Also, a custom propTypes method was created in the props definition to display customized messages depending on the element type of children.

The children of a `TableRow` is normally a `TableCell`. That's what we expect. It can even receive a `div`, a `string` or custom content and it will work, but if it receives a string or an invalid element, it will throw a console warning:

image

platform/ui/src/components/Table/Table.mdx
dannyrb: `<Table>` (React Component) looks a like like `<table>`, the native HTML DOM element.
Should we rename it to something with a vendor prefix, or a two-word name?

`<OhifTable>` or `<SimpleTable>`

platform/ui/src/components/TableCell/TableCell.mdx
dannyrb: I agree that it should be documented, but if it can't be used in isolation, it may not be worth giving it it's own page. You could show how they're _all_ used together, and then show props tables for each component one after the other on the same MDX page that has the playground.

So they'd all be on the molecule's demo, even though it has smaller components, because those smaller component's can't be used outside of this very specific scenario/usage.

platform/ui/src/components/TableRow/TableRow.jsx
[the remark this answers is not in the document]
dannyrb: Our current usage only has Bold text headers and regular text inside of each cell. Is it worth creating a component for that? That's mostly what I'm getting at here.

We may eventually want something more complex, but today, a simple table just shows text tabular data with some light styling.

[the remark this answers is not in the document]
dannyrb: Basically, the type checking and display logic adds cases we won't be using out of the gate. If it's not something our current design needs, it's extra weight we need to test/support that is more likely to break.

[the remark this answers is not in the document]
rodrigolabs: So, this is basically what `TableCell` currently does:

- `colSpan` prop to define its size. (I just added)
- It adds a blue `border-right` to separate the cell (if it's in the TableBody tree).
- `align` prop, to define its content alignment.
- base classes (Tailwind) that set a default padding and word-break.

Even being simple, I think that's enough to have an exposed `TableCell` component. And also to keep consistency in the use of our table components, even if they are simple, to avoid things like:

```javascript
<TableRow>
  <div className="tailwind-classes">...</div>
  <div className="tailwind-classes">...</div>
  <div className="tailwind-classes">...</div>
</TableRow>
```
About the type checking and display logic, I can definitely remove them, as `TableRow` will naturally/always expect a `TableCell` or a `<div>`, but never a string.

I'm sorry for this long text 😅 we can discuss offline if you want.

platform/ui/src/components/TableCell/TableCell.mdx
rodrigolabs: Fixed! All props are placed together in the Table page.

platform/ui/src/components/Table/Table.mdx
rodrigolabs: I think it's ok to have a `Table`, it's capitalized, implying that it's a React component instead of a pure HTML element. This is also how many other component libraries do. What do you think we can do here?

@galelis what's your opinion?

dannyrb: If this is common in other libraries, that's cool with me.

platform/ui/src/components/Table/Table.mdx
```diff
@@ -0,0 +1,124 @@
...
+import { Table, TableBody, TableCell, TableHead, TableRow } from '@ohif/ui';
+```
+
+## Basic usage
+
+<Playground>
+  <div className="p-4">
+    <Table>
+      <TableHead>
+        <TableRow>
```
dannyrb: `[0, 1, 2, 3, 4, 5].map(x => ());` would condense the size of this when the example code is displayed on page. Lots of scrolling to get past:

image

PR #1543, opened 2020-03-20
fix: 🐛 #1124 - Clicking on items in "Measurements" table only scrolls image stack once, on first click. Subsequent clicks have no effect.

ckedar:
### PR Checklist

- [ ] Brief description of changes
- [ ] Links to any relevant issues
- [ ] Required status checks are passing
- [ ] User cases if changes impact the user's experience
- [ ] `@mention` a maintainer to request a review

Set `SOPInstanceUID` in redux state to null on change of image.
Add null guard for `SOPInstanceUID` in cDU of `OHIFCornerstoneViewport`
Update `currentImageIdIndex` in local state in `OHIFCornerstoneViewport` on change of image.

Fix for https://github.com/OHIF/Viewers/issues/1124 - Clicking on items in "Measurements" table only scrolls image stack once, on first click. Subsequent clicks have no effect.

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #1544, opened 2020-03-21
avoid-wasteful-renders

ckedar:
### PR Checklist

### DIFF
--- a/platform/core/src/redux/reducers/viewports.js
+++ b/platform/core/src/redux/reducers/viewports.js
@@ -1,5 +1,5 @@
 import cloneDeep from 'lodash.clonedeep';
-import merge from 'lodash.merge';
+import produce from 'immer';
 
 import {
   CLEAR_VIEWPORT,
@@ -93,12 +93,13 @@ const viewports = (state = DEFAULT_STATE, action) => {
      * @return {Object} New state.
      */
     case SET_VIEWPORT_ACTIVE: {
-      const activeViewportIndex = getActiveViewportIndex(
-        state.numRows,
-        state.numColumns,
-        action.viewportIndex
-      );
-      return { ...state, activeViewportIndex };
+      return produce(state, draftState => {
+        draftState.activeViewportIndex = getActiveViewportIndex(
+          draftState.numRows,
+          draftState.numColumns,
+          action.viewportIndex
+        );
+      });
     }
 
     /**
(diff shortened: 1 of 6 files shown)


ckedar:
SET_VIEWPORT_ACTIVE action is called frequently when scrolling through images using up/down hotkey or cine dialog.
Even when it does not really change the active viewport, code handling this action ends up creating new state object, resultig in wasteful render of many components.

Same is the case with SET_VIEWPORT action, which will be dispatched more frequently with
my PR https://github.com/OHIF/Viewers/pull/1543 for fixing https://github.com/OHIF/Viewers/issues/1124.

This PR avoids wasteful change of redux state when these actions do not really change any data in state.

Also includes small changes in other components to avoid re-render when there is not change in props, by memoizing the computed values passed down to children and avoidig un-necessary use of arrow function.

platform/core/src/redux/reducers/viewports.js
```diff
@@ -93,12 +93,13 @@ const viewports = (state = DEFAULT_STATE, action) => {
...
-        action.viewportIndex
-      );
-      return { ...state, activeViewportIndex };
+      return produce(state, draftState => {
+        draftState.activeViewportIndex = getActiveViewportIndex(
+          draftState.numRows,
+          draftState.numColumns,
+          action.viewportIndex
+        );
+      });
```
dannyrb: Oh wow. That's nice :clap:

platform/viewer/src/connectedComponents/ViewerMain.js
```diff
@@ -140,9 +144,7 @@ class ViewerMain extends Component {
 
   render() {
     const { viewportSpecificData } = this.props;
-    const viewportData = viewportSpecificData
-      ? Object.values(viewportSpecificData)
-      : [];
+    const viewportData = values(viewportSpecificData);
```
dannyrb: Advantage here is the cache we get from `memoize`? We're using `lodash/values` because `Object.values` doesn't work with this pattern?

platform/viewer/src/components/ViewportGrid/ConnectedViewportGrid.js
```diff
@@ -2,15 +2,20 @@ import ViewportGrid from './ViewportGrid.js';
 import { MODULE_TYPES } from '@ohif/core';
 import { connect } from 'react-redux';
 import { extensionManager } from './../../App.js';
+import memoize from 'lodash/memoize';
 
-const mapStateToProps = state => {
+const getAvailableViewportModules = memoize(viewportModules => {
```
dannyrb: Good point. If the viewportModules is the same value, so to would be the result of `getAvailableViewportModules`

platform/viewer/src/connectedComponents/ViewerMain.js
ckedar: We can use `Object.values` as well. Will have have to use null guard, though.
i.e. 
```
const values = memoize(Object.values);
......
const viewportData = values(viewportSpecificData || {} );
```

PR #1553, opened 2020-03-24
Refactor Filters for Text, Select and DateRange

galelis:
**OHIF-46
OHIF-47
OHIF-48**

### Descrioption
Separate into small components the filters for Text, Select and DateRage

### Screenshot
NOTE: Nothing should change on current layout. 
Screen Shot 2020-03-25 at 05 26 56


dannyrb:
If possible, modality should be:

- Multi-select
- Values should be all caps
- Checkboxes in expanded dropdown to add to list
- When collapsed, show values not as pills, but as text
- When collapsed, if value exceeds container, truncate with `...`

platform/ui/index.js
```diff
@@ -10,6 +10,10 @@ export {
   Button,
   ButtonGroup,
   DateRange,
+  FilterDateRange,
+  FilterSelect,
+  FilterText,
```
dannyrb: These are not necessarily limited to use with filtering. Maybe:

`InputText`
`InputSelect`
`InputDateRange`

platform/ui/index.js
```diff
@@ -10,6 +10,10 @@ export {
...
+  FilterWrapper,
```
dannyrb: Not sure what this does yet. Will come back if naming feels off.

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { DateRange, FilterWrapper } from '@ohif/ui';
+
+const FilterDateRange = ({
+  label,
+  isSortable,
+  isBeingSorted,
+  sortDirection,
```
dannyrb: Should this sort stuff be composable? Part of some HOC magic?

dannyrb: I think we can cut this down to:

`isSortable` and `sortDirection`
It's being sorted if `isSortable` is true and `sortDirection` is `ascending` or `descending` (or whatever their corresponding values are)

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
```diff
@@ -0,0 +1,66 @@
...
+  onLabelClick,
+  inputValue,
+  inputProps,
```
dannyrb: This implies a leaky abstraction. Can we just expose the props directly that we want to pass through so the user doesn't have to look up docs for `DateRange`?

dannyrb: That would means these would be renamed to `value`. With `inputProps` expanded based on the props we want to pass through.

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
```diff
@@ -0,0 +1,66 @@
...
+import PropTypes from 'prop-types';
+
+import { DateRange, FilterWrapper } from '@ohif/ui';
+
+const FilterDateRange = ({
+  label,
+  isSortable,
+  isBeingSorted,
+  sortDirection,
+  onLabelClick,
```
dannyrb: Should this be changed to `onSortUpdate` or something similar?

dannyrb: Maybe `onLabelClick` does make more sense 🤔 
LMK your thoughts.

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
```diff
@@ -0,0 +1,66 @@
...
+  isBeingSorted: PropTypes.bool,
+  sortDirection: PropTypes.number,
+  onLabelClick: PropTypes.func,
+  value: PropTypes.shape({
+    startDate: PropTypes.string,
+    endDate: PropTypes.string,
+  }),
+  inputProps: PropTypes.object,
+  onChange: PropTypes.func,
+};
```
dannyrb: When possible, default to `required` with no default state. It's less variations we have to support and account for. For example, `isSortable` should simply be required. As should label, and a few others.

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
```diff
@@ -0,0 +1,66 @@
...
+  label: PropTypes.string,
+  isSortable: PropTypes.bool,
+  isBeingSorted: PropTypes.bool,
+  sortDirection: PropTypes.number,
+  onLabelClick: PropTypes.func,
+  value: PropTypes.shape({
+    startDate: PropTypes.string,
+    endDate: PropTypes.string,
+  }),
+  inputProps: PropTypes.object,
```
dannyrb: This is not specific enough. I know we're going to make these more explicit, but almost always want `shape` if we're going to support an object.

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+FilterDateRange.propTypes = {
+  label: PropTypes.string,
+  isSortable: PropTypes.bool,
+  isBeingSorted: PropTypes.bool,
+  sortDirection: PropTypes.number,
+  onLabelClick: PropTypes.func,
+  value: PropTypes.shape({
+    startDate: PropTypes.string,
+    endDate: PropTypes.string,
```
dannyrb: Are these string? Do they have to be in a specific format? Inline comment to clarify?

platform/ui/index.js
galelis: Liked it. I'll update.

platform/ui/src/components/FilterSelect/FilterSelect.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+const FilterSelect = ({
+  label,
+  isSortable,
+  isBeingSorted,
+  sortDirection,
+  onLabelClick,
+  inputValue,
+  inputProps,
+  onChange,
```
dannyrb: Same notes on props for this component.

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
galelis: I have added LabelClick because used can do anything he wanted in case any click on label accours.

platform/ui/src/components/FilterSelect/FilterSelect.jsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <FilterWrapper
+      label={label}
+      isSortable={isSortable}
+      isBeingSorted={isBeingSorted}
+      sortDirection={sortDirection}
+      onLabelClick={onLabelClick}
+    >
+      <Select
+        {...inputProps}
```
dannyrb: Never spread props 🙏

platform/ui/src/components/FilterSelect/FilterSelect.jsx
```diff
@@ -0,0 +1,66 @@
...
+  onChange: () => {},
+};
+
+FilterSelect.propTypes = {
+  label: PropTypes.string,
+  isSortable: PropTypes.bool,
+  isBeingSorted: PropTypes.bool,
+  sortDirection: PropTypes.number,
+  onLabelClick: PropTypes.func,
+  value: PropTypes.oneOfType([PropTypes.array, PropTypes.object]),
```
dannyrb: array of? shape?

platform/ui/src/components/FilterWrapper/FilterWrapper.jsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
```
dannyrb: Maybe renamed "FormGroup"
Similar semantics of Bootstrap where you wrap label + input + extra
Not necessarily limited to filtering

platform/ui/src/components/FilterWrapper/FilterWrapper.jsx
```diff
@@ -0,0 +1,67 @@
...
+  sortDirection: 0,
+  onLabelClick: () => {},
+  className: '',
+};
+
+FilterWrapper.propTypes = {
+  label: PropTypes.string,
+  isSortable: PropTypes.bool,
+  isBeingSorted: PropTypes.bool,
+  sortDirection: PropTypes.number,
```
dannyrb: Maybe "one of" if we only support three values

platform/ui/src/components/FilterWrapper/FilterWrapper.jsx
```diff
@@ -0,0 +1,67 @@
...
+    <label className={classnames(baseLabelClassName, className)}>
+      <span
+        className={spanClassName}
+        onClick={onLabelClick}
+        onKeyDown={onLabelClick}
+      >
+        {label}
+        {isSortable && <Icon {...iconProps} />}
+      </span>
+      <span>{children && children}</span>
```
dannyrb: `children && children` would have the same effect as `children`?

dannyrb: But why inside a `<span>`?

platform/ui/src/components/FilterWrapper/FilterWrapper.jsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+FilterWrapper.propTypes = {
+  label: PropTypes.string,
+  isSortable: PropTypes.bool,
+  isBeingSorted: PropTypes.bool,
+  sortDirection: PropTypes.number,
+  onLabelClick: PropTypes.func,
+  className: PropTypes.string,
+  children: PropTypes.node,
```
dannyrb: 100% should be required.

platform/ui/src/components/FilterWrapper/FilterWrapper.jsx
```diff
@@ -0,0 +1,67 @@
...
+  isBeingSorted,
+  sortDirection,
+  onLabelClick,
+  className,
+  children,
+}) => {
+  const iconProps = {
+    name: isBeingSorted ? sortIconMap[sortDirection] : 'sorting',
+    className: classnames(
+      'mx-2 w-2',
```
dannyrb: In general, the "outer element" of a React Component should not have margin.
The consumer of the component can set that, and this limits the places where this component can be used (with limited benefit)

In Viewer "view", this would be set on the outside of each element, and that would be cool beans

platform/ui/src/components/FilterWrapper/FilterWrapper.jsx
```diff
@@ -0,0 +1,67 @@
...
+      isBeingSorted ? 'text-custom-aquaBright' : 'text-custom-blue'
```
dannyrb: CC: @rodrigolabs, have we not addressed colors yet? Any blockers?

platform/ui/src/components/Select/Select.jsx
```diff
@@ -55,7 +52,7 @@ Select.propTypes = {
   ),
   placeholder: PropTypes.string,
   noOptionsMessage: PropTypes.func,
-  value: PropTypes.string,
+  value: PropTypes.oneOfType([PropTypes.array, PropTypes.object]),
```
dannyrb: Same refactoring energy here

platform/ui/src/views/StudyList/StudyList.js
```diff
@@ -11,55 +11,57 @@ const filtersMeta = [
...
-    inputType: 'text',
+    inputType: 'Text',
     isSortable: true,
     gridCol: 4,
   },
   {
     name: 'instances',
     displayName: 'Instances',
-    inputType: 'none',
+    inputType: 'None',
```
dannyrb: Make sure PropTypes reflects available options for `inputType`

platform/ui/src/views/StudyList/StudyList.js
```diff
@@ -11,55 +11,57 @@ const filtersMeta = [
...
-      { value: 'mr', label: 'mr' },
-      { value: 'sr', label: 'sr' },
-    ],
+    inputType: 'Select',
+    inputProps: {
+      options: [
+        { value: 'seg', label: 'seg' },
+        { value: 'ct', label: 'ct' },
+        { value: 'mr', label: 'mr' },
+        { value: 'sr', label: 'sr' },
```
dannyrb: Caps

platform/ui/src/components/FilterText/FilterText.jsx
```diff
@@ -0,0 +1,60 @@
...
+  isSortable: PropTypes.bool,
+  isBeingSorted: PropTypes.bool,
+  sortDirection: PropTypes.number,
+  onLabelClick: PropTypes.func,
+  value: PropTypes.string,
+  inputProps: PropTypes.object,
+  onChange: PropTypes.func,
+};
+
+export default FilterText;
```
dannyrb: Same notes as other components

platform/ui/src/views/StudyList/components/StudyListFilter.js
```diff
@@ -241,19 +143,24 @@ const StudyListFilter = ({
...
-                          isSortable && numOfStudies <= 100 && numOfStudies > 0
-                        }
-                        isBeingSorted={sortBy === name}
-                        sortDirection={sortDirection}
-                        onLabelClick={() => handleFilterLabelClick(name)}
-                        inputType={inputType}
-                      >
-                        {renderInput(inputType, name, { selectOptions })}
-                      </FilterLabel>
+                      {FilterInputComponent && (
```
dannyrb: Just mark as required?

platform/ui/src/views/StudyList/components/StudyListFilter.js
```diff
@@ -275,12 +182,32 @@ const StudyListFilter = ({
...
+  numOfStudies: 90,
+};
+
 StudyListFilter.propTypes = {
   filtersMeta: PropTypes.arrayOf(
     PropTypes.shape({
       name: PropTypes.string,
       dsplayName: PropTypes.string,
-      inputType: PropTypes.oneOf(['text', 'select', 'date-range', 'none']),
+      inputType: PropTypes.oneOf(['Text', 'Select', 'DateRange', 'None']),
```
dannyrb: Ah. Perfect. Thanks!

platform/ui/src/views/StudyList/components/StudyListFilter.js
```diff
@@ -275,12 +182,32 @@ const StudyListFilter = ({
...
+    description: '',
+    modality: [],
+    accession: '',
+    sortBy: '',
+    sortDirection: 0,
+    page: 0,
+    resultsPerPage: 25,
+  },
+  numOfStudies: 90,
+};
```
dannyrb: On the fence about defaults here. Less pressing as this is higher up; might be worth forcing above layer to provide values so its easier to see at a glance where everything is coming from.

We should have a dedicated page for just `StudyListFilter`

platform/ui/src/views/StudyList/components/StudyListFilter.js
galelis: This is for cases like "Instances" where we don't have an input but we have the label and sorting. 
Shoul we treat that in a different way? Like creating a component for that expecifically?

platform/ui/src/components/FilterWrapper/FilterWrapper.jsx
galelis: This is actually the iconProps.className that is used only inside the "FilterWrapper".  
I agree with you and I think we are fine on this one.

platform/ui/src/views/StudyList/components/StudyListFilter.js
dannyrb: Good question. Is the "just a label" a component? Maybe I skipped over where/how you're rendering it.

galelis: As it could have sorting on it, I have used the same stucture but I don't add an input for it.

platform/ui/src/components/FilterSelect/FilterSelect.jsx
galelis: Updated :)

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
galelis: PropTypes updated. Comment added.

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
galelis: I've removed it from the input components. Passing only the needed props

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
galelis: Done. make sense.

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
galelis: Updated :)

platform/ui/src/components/FilterDateRange/FilterDateRange.jsx
galelis: I've updated the code to work only with the required properties on each input component.

platform/ui/src/components/FilterSelect/FilterSelect.jsx
galelis: Updated the code.

platform/ui/src/components/FilterWrapper/FilterWrapper.jsx
rodrigolabs: I created a specific ticket for color naming.

platform/ui/src/components/InputDateRange/InputDateRange.jsx
```diff
@@ -0,0 +1,56 @@
...
+      isSortable={isSortable}
+      sortDirection={sortDirection}
+      onLabelClick={onLabelClick}
+    >
+      <div className="relative">
+        <DateRange
+          startDate={startDate}
+          endDate={endDate}
+          onChange={({ startDate, endDate }) => {
+            if (onChange) {
```
rodrigolabs: As `onChange` is required, maybe we don't need this condition here.

platform/ui/src/components/InputDateRange/InputDateRange.jsx
```diff
@@ -0,0 +1,56 @@
...
+  );
+};
+
+InputDateRange.defaultProps = {
+  value: {},
+};
+
+InputDateRange.propTypes = {
+  label: PropTypes.string.isRequired,
+  isSortable: PropTypes.bool.isRequired,
```
rodrigolabs: `isSortable` is really necessary? it's only used to define if the sort icon should be displayed or not. Maybe we can use the `sortDirection` itself to define it.

- If `sortDirection` has any value, the field is automatically sortable, right?
- `sortDirection` can be "not required" -- field is not sortable if `sortDirection` is empty or  has`none` as value.

What do you think? Open to ideas.

platform/ui/src/components/InputSelect/InputSelect.jsx
```diff
@@ -0,0 +1,84 @@
...
+      isSortable={isSortable}
+      sortDirection={sortDirection}
+      onLabelClick={onLabelClick}
+    >
+      <Select
+        placeholder={placeholder}
+        className="mt-2"
+        options={options}
+        value={value}
+        isMulti={true}
```
rodrigolabs: `<Select />` will always be `multi`?

platform/ui/src/components/InputText/InputText.jsx
```diff
@@ -0,0 +1,48 @@
...
+  );
+};
+
+InputText.defaultProps = {
+  value: '',
+};
+
+InputText.propTypes = {
+  label: PropTypes.string.isRequired,
+  isSortable: PropTypes.bool.isRequired,
```
rodrigolabs: Same thing about `isSortable`

platform/ui/src/components/Select/Select.jsx
```diff
@@ -1,61 +1,115 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import ReactSelect from 'react-select';
+import ReactSelect, { components } from 'react-select';
 
 import './Select.css';
 
+const MultiValueContainer = ({ data }) => {
+  return <span>{`${data.label}, `}</span>;
```
rodrigolabs: Is it possible to have a comma separating the values only if there are two or more values?
To avoid this:

image

platform/ui/src/components/Select/Select.jsx
```diff
@@ -1,61 +1,115 @@
...
+const Option = props => {
+  return (
+    <div>
+      <components.Option {...props}>
+        <input
+          type="checkbox"
+          checked={props.isSelected}
+          className="w-6 h-6"
+          onChange={e => null}
+        />{' '}
```
rodrigolabs: Was this "blank space" `{' '}` intentional?

rodrigolabs: Checkbox + Text is not aligned:

dannyrb: I think autofix inserts the blank space to preserve line breaks that would not persist with current formatting rules.

platform/ui/src/components/Select/Select.jsx
dannyrb: Array.split(', ') works well for this; then just display the string

platform/ui/src/components/InputSelect/InputSelect.jsx
dannyrb: I'm okay with it being locked down and potentially renaming this `MultiSelect`

Again, less options is better, unless we need that level of configuration. It's less to test, and these are UI states we won't be able to verify the accuracy of as they're not encountered in the application.

If we find ourselves needing a non-multi-select, then we can revisit the API here?

platform/ui/src/components/Select/Select.jsx
```diff
@@ -1,61 +1,115 @@
...
+    ? { MultiValueContainer, Option, MultiValue }
+    : {};
+
   return (
     <ReactSelect
-      autoFocus={autoFocus}
       className={classnames(
         className,
-        'flex flex-col flex-1 mt-2 customSelect__wrapper'
+        'flex flex-col flex-1 customSelect__wrapper'
```
rodrigolabs: Maybe we could define a `min-width` for this component, here's how it's being displayed in the docz page for Select:

image

I know it depends on the parent width, but to avoid break lines like this, we could just set a `min-width`.

What do you think?

platform/ui/src/components/InputDateRange/InputDateRange.jsx
dannyrb: I'm okay with either option, so long as its documented.

I'm trying to think about a parent setting all as sortable or not. Does it make more sense to set this per input, or to clear sort direction? What would you expect as a consumer of this library?

platform/ui/src/views/StudyList/components/StudyListFilter.js
```diff
@@ -108,61 +59,43 @@ const StudyListFilter = ({
...
+    const FilterInputComponent = getFieldInputComponent(inputType);
+
+    return (
+      <FilterInputComponent
+        key={name}
+        label={displayName}
+        isSortable={_isSortable}
+        sortDirection={_sortDirection}
+        onLabelClick={onLabelClick}
+        {...inputProps}
```
rodrigolabs: `inputProps` is currently used for the Select options, right? As we should avoid spreading props, isn't it better to explicitly pass this value to an `options` prop?

platform/ui/src/components/InputSelect/InputSelect.jsx
rodrigolabs: Perfect! I agree. Let's keep as it is for now!

platform/ui/src/components/InputDateRange/InputDateRange.jsx
galelis: I have 2 thoughts here. 
>Maybe we can use the sortDirection itself to define it.

`isSortable + sortDirection` could be a single `sorting` with values as `['none', 'ascending', 'descending', 'hidden']` but the way its now, is more intuitive to understand the direction and if we show or not. 
> Does it make more sense to set this per input, or to clear sort direction?

platform/ui/src/components/InputSelect/InputSelect.jsx
galelis: I thoguht the same as Danny, as he is always suggesting to lower the options if we can, as our input field is a multiSelect for now, I decided to move with it locking this option. 

But I definetely liked the idea of renaming to InputMultiSelect.

platform/ui/src/components/Select/Select.jsx
galelis: I will try to work with that. 
Select allows me to override some of its internal components, I was overriding only the MultiSelectValue which only allows me to touch 1 by 1 of the selected items. 

I will work on this to avoid displaying the coma on the last item.

platform/ui/src/components/Select/Select.jsx
galelis: I haven't touch the react-select css for that. I can do it thought. 
Ideally it works like this so it can expand and always show all selected itens. 
I don't think setting a min with will solve this. cause it will depends on the size of the options as well. 

We would need to think on how to deal with bigger inputs, as we did for the InputMultiSelect, when we wrap the content and add ellipses. But this was an especific scenario.

platform/ui/src/views/StudyList/components/StudyListFilter.js
galelis: The code will be a little bit bigger now, but I removed this spreading porps.

platform/ui/src/components/Select/Select.jsx
galelis: Fixed alingment and also the extra `{' '}`.

platform/ui/src/components/Select/Select.jsx
galelis: Fixed.

platform/ui/src/components/Select/Select.mdx
```diff
@@ -22,10 +22,10 @@ import { Select } from '@ohif/ui';
     <div className="w-20">
```
rodrigolabs: About the Select `min-width`, let's change this width and then we're good!

```suggestion
    <div className="w-56">
```

platform/ui/src/components/Select/Select.jsx
rodrigolabs: Ok, let's only add a bigger width in the doc page as suggested

rodrigolabs:
Only two changes:

- Remove the condition for `onChange` in `InputDateRange.jsx`
- Increase the width of the `div` wrapping the Select component in the Select doc page

And then we're good to merge!

Awesome work! 👏🏻

galelis:
@dannyrb @rodrigolabs 
All requested changes done. 
- Renamed InputSelect to InputMultiSelect
- Removed the condition for `onChange` in `InputDateRange.jsx`
- Updated Select docz page adding bigger width on select

PR #1556, opened 2020-03-24
feat: UI v2 EmptyStudies Component

rodrigolabs:
**OHIF-51**

Creates an EmptyStudies component to be used in the StudyList when the study list is empty.

## Preview

image


rodrigolabs:
After an offline discussion, we decided to merge this PR and create another PR for organizing better the Menu structure.

Issue #1522, opened 2020-03-13
Support for rendering different measurements in different viewports


JamesAPetts:
RE: https://github.com/OHIF/Viewers/issues/1373#issuecomment-598722322

For IDC we need support for displaying the same scan in two viewports and displaying different RTSTRUCTs, SEGs and cornerstone planar measurements in each.

> In Slicer there is a pop up menu with a check box per viewport, but it might also make sense to have a drag and drop option.

![](https://user-images.githubusercontent.com/126077/76625367-75968380-650d-11ea-9ec8-cfa224139dea.png)

What is the best way to do this in OHIF in an extensible way?

dannyrb:
@swederik and I have recently discussed something similar. Currently, most tools use the default global `imageIdSpecificToolStateManager`. We could create/destroy them per viewport, and designate their usage for data like this (restored / 3rd party).

Tool's are also fairly flexible. In the render loop, we can examine ToolState or use a callback to make determinations on what should be rendered.

Taking a step back, we may want to discuss how we want to structure this with the MeasurementService, and if we expect this to impact vtk viewports.

sedghi:
@fedorov @igoroctaviano Do you want this to be open?

fedorov:
This is v2, and I think this can already be done at least to some degree.

PR #1558, opened 2020-03-25
Load measurement in active viewport.

ckedar:
Possible solution for request https://github.com/OHIF/Viewers/issues/1522 ?


dannyrb:
This does not address #1522, but is a bugfix for a different outstanding issue. Thanks for the contribution @ckedar!

PR #1559, opened 2020-03-25
fix: Add support for single entries in SequenceOfUltrasoundRegions. M…

swederik:
…ore than one region will need lower-level changes

### PR Checklist

### DIFF
--- a/platform/core/src/utils/metadataProvider/getPixelSpacingInformation.js
+++ b/platform/core/src/utils/metadataProvider/getPixelSpacingInformation.js
@@ -19,45 +19,53 @@ export default function getPixelSpacingInformation(instance) {
     '1.2.840.10008.5.1.4.1.1.12.1.1', //	Enhanced XA Image Storage
     '1.2.840.10008.5.1.4.1.1.12.2', //	X-Ray Radiofluoroscopic Image Storage
     '1.2.840.10008.5.1.4.1.1.12.2.1', //	Enhanced XRF Image Storage
-    '1.2.840.10008.5.1.4.1.1.12.3' // X-Ray Angiographic Bi-plane Image Storage	Retired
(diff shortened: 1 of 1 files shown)


platform/core/src/utils/metadataProvider/getPixelSpacingInformation.js
```diff
@@ -67,24 +75,56 @@ export default function getPixelSpacingInformation(instance) {
...
     return {
       PixelSpacing: CorrectedImagerPixelSpacing,
       isProjection,
     };
+  } else if (
+    SequenceOfUltrasoundRegions &&
+    typeof SequenceOfUltrasoundRegions === 'object'
+  ) {
+    const { PhysicalDeltaX, PhysicalDeltaY } = SequenceOfUltrasoundRegions;
+    let USPixelSpacing = [PhysicalDeltaX * 10, PhysicalDeltaY * 10];
```
dannyrb:
```suggestion
    const USPixelSpacing = [PhysicalDeltaX * 10, PhysicalDeltaY * 10];
```

platform/core/src/utils/metadataProvider/getPixelSpacingInformation.js
```diff
@@ -67,24 +75,56 @@ export default function getPixelSpacingInformation(instance) {
...
+        'EstimatedRadiographicMagnificationFactor was not present. Unable to correct ImagerPixelSpacing.'
+      );
     }
 
     return {
       PixelSpacing: CorrectedImagerPixelSpacing,
       isProjection,
     };
+  } else if (
+    SequenceOfUltrasoundRegions &&
```
JamesAPetts: It might be a bit cleaner to check if `SequenceOfUltrasoundRegions` exists, then do a nested if which just checks Array.isArray, and else. Rather than two `else if`s that check the same thing at the start of their conditional.

PR #1570, opened 2020-03-27
OHIF-44: Navigation bar component

galelis:
### Description
Create a navigation bar component to wrap a header with all basic style needed. 

### Preview
- Basic usage: 
Screen Shot 2020-03-27 at 13 20 42

- NavBar applied on viewer
Screen Shot 2020-03-27 at 13 20 28


platform/ui/src/components/NavBar/NavBar.mdx
```diff
@@ -0,0 +1,32 @@
+---
+name: NavBar
+menu: Components
+route: components/navBar
+---
+
+import { Playground, Props } from 'docz';
+import Button from './../Button';
```
rodrigolabs: Let's import these components from `@ohif/ui` instead of its relative path.

`Button` is not being used? Remove if not.

galelis: True. Did this way because we were doing this way. hehe 

Just updated all docz pages that were not importing ohif/ui components not from @ohif/ui

rodrigolabs: \o/ \o/ \o/ Thanks!

platform/ui/src/components/Button/Button.mdx
```diff
@@ -5,9 +5,7 @@ route: components/button
 ---
 
 import { Playground, Props } from 'docz';
-import Button from './';
-import IconButton from '../IconButton';
-import Icon from './../Icon';
+import { Button, IconButton, Icon } from '@ohif/ui';
```
dannyrb: Hmmmmmmmmmmm.... I think this is okay.
It feels weird to me, but if you and @rodrigolabs are in agreement then I'm cool with it.
Just make sure you don't push anything to a top level export when it shouldn't be to satisfy this.

rodrigolabs: Yes, that's only for the exposed components. "Local" components should never be exposed.

platform/ui/src/views/StudyList/components/Header.js
```diff
@@ -12,7 +12,7 @@ function Header({ appLogo = OHIFLogo(), children, t }) {
   };
 
   return (
-    <div className="bg-custom-navy flex flex-row justify-between px-3 py-1 text-white sticky top-0 z-10 border-b-4 border-black">
+    <NavBar className="justify-between border-b-4 border-black">
```
dannyrb: A tiny bit torn. @rodrigolabs, should border be added inside our outside of component? are there any scenarios where you wouldn't want border?

I do agree with `justify-between` being external.

platform/ui/src/views/StudyList/components/Header.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { withTranslation } from 'react-i18next';
 
-import { IconButton, Icon } from '@ohif/ui';
+import { IconButton, Icon, NavBar } from '@ohif/ui';
```
dannyrb: Can we demo the Header component in a category? I want to be able to see it in something other than the generic "Components" category.

PR #1575, opened 2020-03-30
Reduce bundle size

ckedar:
Reduces main app bundle size by 63% (size goes down from 9.83MB to 3.66MB), by splitting it into chunks that can be loaded asynchronously.
50% of current app.js bundle size is taken up by two big components - `cornerstoe-wado-image-loader` and `react-vtkjs-viewport`. 
Both are not necessary for rendering initial Study List screen.
In fact `react-vtkjs-viewport` is not required until we enter into `2D MPR`.

By splitting the bundle into chunks and delaying their loading as much as possible, initial load time can improve.


swederik:
You are definitely my favourite new contributor 🥇

JamesAPetts:
This is awesome, main PWA feels a lot snappier! Awesome job.

Currently `/local` and the json launch are broken, although these are likely easy to fix, they can be found under these urls:

[Local](https://deploy-preview-1575--ohif.netlify.com/pwa/local)

[JSON Launch](https://deploy-preview-1575--ohif.netlify.com/pwa/viewer?url=https://ohif-viewer.s3.eu-central-1.amazonaws.com/JSON/testJSON.json) (this just takes a JSON file and loads a study based on its JSON structure).

Both just seem to need dynamic loading of `cornerstone-wado-image-loader` added.

Sure @dannyrb  will have some other comments on the PR.

Thanks again for the awesome contribution!

platform/viewer/src/utils/initWebWorkers.test.js
```diff
@@ -20,4 +20,4 @@ describe('initWebWorkers', () => {
       cornerstoneWADOImageLoader.webWorkerManager.initialize
     ).toHaveBeenCalledTimes(1);
   });
-});
+}); */
```
dannyrb: Hmm.. I guess we should/would mock `cornerstoneWADOImageLoader` here

platform/viewer/src/utils/cornerstoneWADOImageLoader.js
```diff
@@ -0,0 +1,57 @@
+import OHIF from '@ohif/core';
```
dannyrb: Maybe we should move all of these `getSomeExternalDependency` methods to a folder so they're easier to locate? I think we'll eventually want to make this split possible/easier with dependency injection instead of using a utility like this to isolate the import?

platform/viewer/src/routes/ViewerRouting.js
```diff
@@ -29,10 +30,21 @@ function ViewerRouting({ match: routeMatch, location: routeLocation }) {
...
+  const [
+    cornerstoneWADOImageLoaderInitialized,
+    setCornerstoneWADOImageLoaderInitialized,
+  ] = useState(false);
+
+  React.useEffect(() => {
+    initCornerstoneWADOImageLoader().then(() =>
+      setCornerstoneWADOImageLoaderInitialized(true)
+    );
+  }, []);
```
dannyrb: 👏 👏 👏

platform/viewer/src/lib/localFileLoaders/fileLoaderService.js
```diff
@@ -53,7 +53,8 @@ class FileLoaderService extends FileLoader {
     return result;
   }
 
-  addFile(file) {
+  async addFile(file) {
+    const cornerstoneWADOImageLoader = await getCornerstoneWADOImageLoader();
```
dannyrb: Do these cause a timing issue with the init that takes place in `ViewerRouting`? The assumption is that the init has completed, right?

platform/viewer/src/config.js
```diff
@@ -15,31 +15,33 @@ window.info = {
...
-cornerstoneWADOImageLoader.configure({
-  beforeSend: function(xhr) {
-    const headers = OHIF.DICOMWeb.getAuthorizationHeader();
-
-    if (headers.Authorization) {
-      xhr.setRequestHeader('Authorization', headers.Authorization);
-    }
-  },
-});
+// getCornerstoneWADOImageLoader().then(cornerstoneWADOImageLoader => {
```
dannyrb: You can delete all of these comments. It was odd from the outset that we initialized in this file.

platform/viewer/src/utils/cornerstoneWADOImageLoader.js
```diff
@@ -0,0 +1,57 @@
...
+import cornerstone from 'cornerstone-core';
+import dicomParser from 'dicom-parser';
+
+export function getCornerstoneWADOImageLoader() {
+  return import(
+    /* webpackChunkName: "CornerstoneWADOImageLoader" */ 'cornerstone-wado-image-loader'
+  );
+}
```
dannyrb, line 9: Hmm.. Maybe `getCornerstoneWADOImageLoader` should call `initCornerstoneWADOImageLoader` if it hasn't been initialized yet? Then we don't need to export or worry about when/what calls `initCornerstoneWADOImageLoader`?

platform/viewer/src/App.js
```diff
@@ -140,7 +140,7 @@ class App extends Component {
      */
     _initHotkeys(appConfigHotkeys);
     _initServers(servers);
-    initWebWorkers();
+    // initWebWorkers();
```
dannyrb, line 143: You can delete this.

platform/core/src/utils/cornerstoneWADOImageLoader.js
```diff
@@ -0,0 +1,5 @@
+export function getCornerstoneWADOImageLoader() {
+  return import(
+    /* webpackChunkName: "CornerstoneWADOImageLoader" */ 'cornerstone-wado-image-loader'
+  );
+}
```
dannyrb, line 5: Hmm.. This is tricky. Maybe instead of having `cornerstone-wado-image-loader` as a dependency in `@ohif/core`, we update `dicomLoaderService.js`, studyInstanceHelpers.js`, and `studyLoadingListener.js` to take an image loader in their constructors. Then we don't need to know when/if it has already been initialized?

platform/viewer/src/utils/initWebWorkers.test.js
ckedar: The test it self got removed as part of clean-up.

platform/viewer/src/config.js
ckedar: done

platform/viewer/src/App.js
ckedar: done

platform/viewer/src/lib/localFileLoaders/fileLoaderService.js
ckedar: > Do these cause a timing issue with the init that takes place in ViewerRouting? 

No. Looks like this does not get called from `ViewerRouting` at all.

There are call paths to this from `ViewerLocalFileData` and  `StudyListRoute`.
`ViewerLocalFileData` is now being shown only after initialising cornerstoneWADOImageLoader.

Call path from `StudyListRoute` is from unused `onDrop` method.

platform/core/src/utils/cornerstoneWADOImageLoader.js
ckedar: Passing it through constructor will require changes in multiple places across other packages, at least for `DicomLoaderService`. 

One not-so-clean alternative is to use to from global as `initCornerstoneWADOImageLoader` is setting up `window.cornerstoneWADOImageLoader`.

ckedar:
Updated the PR to fix the broken functionalities (Local and JSON Launch) and remove the dead code.

dannyrb:
@ckedar, thanks again for your work here 🎉 
If you have any other insights to share, we're all ears.

Also more than happy to donate a few hours of my own time if there's anything I can assist you with. You've been incredibly helpful with all of these PRs.

ckedar:
@dannyrb, @swederik Thank you for the appreciation.
Really glad that I could make some meaningful contribution to an open source project. 
Kind of first one for me.

swederik:
This is where we are at now after your changes (just uncomment this line https://github.com/OHIF/Viewers/blob/master/platform/viewer/.webpack/webpack.pwa.js#L50 for anyone that wants to try it themselves):

Screenshot 2020-04-03 at 13 47 12

Some other stuff we could try to dynamically load:
- dcmjs (though maybe we are using it to normalize the QIDO response on the study list, I think. Maybe it could still be part of the second chunk to load and not part of the initial bundle)
- Cornerstone core, cornerstone tools, dicomParser,  - there's no real reason why these need to be part of the first bundle. They are probably dependencies of react-cornerstone-viewport and maybe one of the platform libraries. Maybe we can find a way to push them to a later bundle as well

PR #1576, opened 2020-03-30
Show series in Study before their metadata is retrieved

swederik:
This PR changes the behaviour for showing display sets in the Studies panel while the study is loading. Essentially right now, only the first series is shown while the first image is retrieved. Once the first image is available, the remainder of the display sets appear. 

The updated behaviour is that we now show some basic information about the series that we have from the QIDO-RS Search for Series request while we are waiting for the metadata and images to arrive for the other series. Once they arrive, display sets are created and the images appear.

This makes a big difference when you have either large images (in which case the initial-image blocking make it appear very strangely to the user) or a slow PACS. In some cases, you aren't sure if the whole case is loaded because it may take a few seconds for the other series to appear.

With this change, the user knows the minimum number of series they should be expecting, and therefore they know when the case has loaded. Note: we may want to set some loading state for these cards, but that can happen after the redesign.

GIF shows before and after. It's a minimal change if you have a fast PACS.

Mar-30-2020 18-29-46-2

### PR Checklist


platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -303,13 +316,25 @@ export class StudyMetadata extends Metadata {
...
-      );
-      return true;
+    if (!this.containsSeries(series)) {
+      return false;
     }
-    return false;
+
+    const displaySets = this._createDisplaySetsForSeries(sopClassHandlerModules, series)
+
+    this._displaySets.forEach((displaySet, idx) => {
```
ckedar: Possible issue here: deleting current item in forward iteration loop will skip next item from iteration.

Way out could be reverse iteration using `for` loop or `filter`

swederik: Yeah normally I would use filter but the array has writable: false. Good catch!

platform/core/src/DICOMSR/utils/findMostRecentStructuredReport.js
```diff
@@ -37,6 +37,10 @@ const isStructuredReportSeries = series => {
   ];
 
   const firstInstance = series.getFirstInstance();
+  if (!firstInstance) {
+    return false;
+  }
```
dannyrb: Is this a timing issue caused by these changes? Are we checking "too early" and this short-circuits a would-be exception? Do we eventually re-check to grab the appropriate SR?

If this is used somewhere that's reactive, would the check be more appropriate one layer up?

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
```diff
@@ -1,4 +1,5 @@
 import { api } from 'dicomweb-client';
+import * as dcmjs from 'dcmjs';
```
dannyrb: I thought we fixed this so you no longer need to import as `*`?

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
```diff
@@ -101,15 +102,18 @@ export default class RetrieveMetadataLoaderAsync extends RetrieveMetadataLoader
 
   async preLoad() {
     const preLoaders = this.getPreLoaders();
-    const result = await this.runLoaders(preLoaders);
+    const seriesData = await this.runLoaders(preLoaders);
```
dannyrb: This is better than "result", but what is "seriesData"?

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
```diff
@@ -118,24 +122,47 @@ export default class RetrieveMetadataLoaderAsync extends RetrieveMetadataLoader
...
   async posLoad(loadData) {
     const { server } = this;
 
-    const { sopInstances, asyncLoader } = loadData;
+    const { sopInstances, asyncLoader, seriesData } = loadData;
 
     const study = await createStudyFromSOPInstanceList(server, sopInstances);
 
+    // TODO: Should this be in a helper
+    const { naturalizeDataset } = dcmjs.data.DicomMetaDictionary;
```
dannyrb: Can we import just this part a bit higher up instead of destructuring here?

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
swederik: Oh cool, I'm not sure I had the latest version on this branch

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -281,13 +293,14 @@ export class StudyMetadata extends Metadata {
...
-          displaySets
-        )
-    );
+      series => {
+         const displaySetsForSeries = this._createDisplaySetsForSeries(
+           sopClassHandlerModules,
+           series,
+         );
+
+         displaySets.push(...displaySetsForSeries);
```
dannyrb: I like this better than passing the array and pending in the fn. Less side-effecty 👍

platform/core/src/DICOMSR/utils/findMostRecentStructuredReport.js
swederik: I'm not actually sure. I had this in as a fix at one point but it might not be necessary any longer.

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
swederik: It's the result of the QIDO-RS Search For Series request, but that's too long to write.

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -318,6 +345,9 @@ export class StudyMetadata extends Metadata {
    */
   setDisplaySets(displaySets) {
     if (Array.isArray(displaySets) && displaySets.length > 0) {
+      // TODO: This is weird, can we just switch it to writable: true?
+      this._displaySets.splice(0);
```
dannyrb: I think this is part of a "data cleanup" conversation with James. We need to discuss how/when/where we should be able to access and mutate data. If we're doing this anyway, I don't think there's any harm in `writable: true`. I'm not sure what assumptions we would be breaking, unless this impacts reactivity (For example, in vuejs, you have to use `splice` for arrays or the reactivity isn't triggered)

platform/core/src/redux/reducers/viewports.js
```diff
@@ -170,6 +170,10 @@ const viewports = (state = DEFAULT_STATE, action) => {
         draftState.viewportSpecificData[action.viewportIndex] =
           draftState.viewportSpecificData[action.viewportIndex] || {};
 
+        if (!action.viewportSpecificData) {
+          return;
```
dannyrb: Why would this be undefined?

platform/core/src/classes/metadata/StudyMetadata.js
swederik: I left it as writable: false for now. We can come back to this later. I think it's just to prevent developers from mutating that stuff directly. We want the class to handle it internally.

platform/core/src/DICOMSR/utils/findMostRecentStructuredReport.js
swederik: Ah it's just to prevent this from failing if a series has no instances yet. I moved it one level up.

platform/core/src/studies/services/wado/retrieveMetadataLoaderAsync.js
swederik: Added a comment.

PR #1579, opened 2020-03-30
feat: StudyListTable, ExpandedRow, API definitions

rodrigolabs:
**OHIF-53**, **OHIF-54**, **OHIF-55**
___

## Description

Lots of things in this PR. 

### Exposing new components

- `StudyListExpandedRow` - Used to pass via props to the StudyListTable the content that will be displayed when the method `onClickRow` is called.

- `StudyListPagination` - Used to manage the pagination for the StudyListTable.

- `StudyListTable` - The study list table component that will basically receive its content via props.

### API Definition and component's implementation enhancement

- `StudyListTable`
- `StudyListPagination`

### New Docz examples

- `StudyListTable`
- `StudyListPagination`

### Misc

Refactor in the `StudyList` doc page to show a real example on how to assemble components to have a complete StudyList view.

## Preview

Nothing has changed visually.


platform/ui/src/components/StudyListPagination/StudyListPagination.js
```diff
@@ -1,18 +1,18 @@
...
-  IconButton,
-  Icon,
-} from '../../../components/';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button, ButtonGroup, Icon, IconButton, Typography } from '@ohif/ui';
 
-const StudyListPagination = props => {
-  const [currentPage, setCurrentPage] = useState(1);
+const StudyListPagination = ({ paginationData }) => {
```
galelis, line 5: We could pass only the paginationData data, as we are not using it.

platform/ui/src/components/StudyListTable/StudyListTableRow.js
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const StudyListTableRow = (props) => {
+  const { tableData } = props;
+  const { row, expandedContent, onClickRow, isExpanded } = tableData;
+
+  const tdClasses = [
```
galelis, line 9: We have 3 different td and only one is using this classes, maybe we should just add this in the jsx as the other ones.

platform/ui/src/components/StudyListTable/StudyListTableRow.js
```diff
@@ -0,0 +1,83 @@
...
+          </div>
+        </td>
+      </tr>
+    </>
+  );
+};
+
+StudyListTableRow.propTypes = {
+  tableData: PropTypes.shape({
+    row: PropTypes.object.isRequired,
```
galelis, line 76: This property row seems to be an object that we can expand its propType.

platform/ui/src/views/StudyList/StudyList.js
```diff
@@ -83,14 +95,174 @@ const filtersValues = {
...
+    const isExpanded = expandedRows.some((k) => k === rowKey);
+    const {
+      AccessionNumber,
+      Modalities,
+      Instances,
+      StudyDescription,
+      PatientId,
+      PatientName,
+      StudyDate,
+      series,
```
dannyrb, line 114: Only issue is if any of these are mapped values for the "view", we probably don't want them using these normalized names and naming convention. We may need to break from the rules the data services follow here. Fine for now. Just making a note as this could cause confusion when we integrate.

PR #1580, opened 2020-03-30
OHIF-45: StudyListFilter refactor

galelis:
**OHIF-45**

### Description
Create a component that will receive a set of filterMeta information to build them all as InputFiled components. 

### Preview
Screen Shot 2020-03-31 at 18 10 07

### Demo page
https://deploy-preview-1580--ohif.netlify.com/ui/components/InputGroup


platform/ui/src/components/DateRange/DateRange.jsx
```diff
@@ -60,35 +76,30 @@ const DateRange = props => {
...
 
+    const handleMonthChange = event => {
+      onMonthSelect(month, event.target.value);
+    };
+
+    const handleYearChange = event => {
+      onYearSelect(month, event.target.value);
+    };
+
+    const handleOnBlur = () => {};
```
rodrigolabs, line 93: why it's calling an empty function?

platform/ui/src/components/InputGroup/InputGroup.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+import {
+  InputText,
+  InputDateRange,
+  InputMultiSelect,
+  InputLabelWrapper,
+} from '@ohif/ui';
+
+const InputGroup = ({
+  filtersMeta,
```
rodrigolabs, line 13: Sharing some thoughts:

- About the `filter` prefix on `filtersValues`, `filtersMeta`... - Currently, we use `InputGroup` in the filter, but it is not necessarily tied to a filter. Maybe we can use a more generic name for these props.

- setFilterValues, setFilterSorting - I'm not sure if this is the best approach (passing the hook itself to the component). We can avoid passing it and avoid extra logic in the component if we pass something like a `onValueChange` and `onSortingChange`, and the application will be responsible for handling the logic when something changes.

What do you think? If you have any question, we can discuss offline.

platform/ui/src/components/InputLabelWrapper/InputLabelWrapper.jsx
```diff
@@ -24,9 +24,11 @@ const InputLabelWrapper = ({
   return (
     <label className={classnames(baseLabelClassName, className)}>
       <span
+        role="button"
```
rodrigolabs, line 27: +1

platform/ui/src/views/StudyList/components/StudyListFilter.js
```diff
@@ -1,138 +1,28 @@
...
+}) => {
+  const { sortBy, sortDirection } = filterValues;
+  const filterSorting = { sortBy, sortDirection };
+  const setFilterSorting = sortingValues => {
+    setFilterValues({
+      ...filterValues,
+      ...sortingValues,
+    });
   };
+  const isSortingEnable = numOfStudies > 0 && numOfStudies <= 100;
```
rodrigolabs, line 22: Maybe it's a good time to define a prop that enables/disable the sorting, and let the application to deal with this, instead of hard-coding this logic here. What do you think?

platform/ui/src/components/DateRange/DateRange.jsx
galelis: Our eslint has a rule that requires us to add an onBlur on <select> components. 
As we don't do anything onBlur I just create the empty function. 

We could change the the eslint rules.

platform/ui/src/views/StudyList/components/StudyListFilter.js
galelis: Ideally this StudyListFilter will do not exist in our future implementation and this rule will become part of the studyList page.

platform/ui/src/components/InputGroup/InputGroup.jsx
galelis: Both changes make sense, I will work in this.

galelis: Code updated. 
Nice callout, thanks 👍

Issue #1536, opened 2020-03-18
Improving first display time


salimkanoun:
Hi all,

I come back again on the image rendering time.
We made a HUGE improvement since the first time I used OHIF when OHIF was querying the full metadata at the study level and where Orthanc DicomWeb was affected by performance issue that have been solved with the recent dicomweb 1.1 release.
On my server, the first image rendering time using Orthanc dropped from 50 seconds to 5 seconds, so this is a very significant performance improvement.

However it might still have a small performance improvement in OHIF side, that could reduce these 5 seconds before the first rendering.
OHIF start by making a /series call to fetch series details, for this one the answer is extremely fast  (50ms on my server).
And then for each series OHIF call the /metadata API that calls for all instance metadata. 
Here there is a difference with Orthanc and DCM4CHE, Orthanc do not store all metadata in his internal database, it read the original source file which is I/O consuming in the hard disk, thus in my server (HDD) the /metadata of a CT serie (600 slices) is around 5 seconds.
There is not too much we can do about it (except retrieve the metadata at the instance level, slice by slice for progressive rendering but I guess you have some reason you didn't choose the instance level)

The point I think OHIF can be a bit faster for the first rendering is : why OHIF have to wait for the first serie's metadata answer to show the series thumbnail series in the left column ?
The study/UID/series API already returns all series descriptions, series numbers, number of slices.
I guess it is enough to display the all thumbnails series before calling the first /metadata API that is much more time consuming.

So the order might be
/series apis
get 1 thumbnail of each series 
=> display all series thumbnail in the left column
then call the /metadata API of the first serie (and you already have the waiting wheel on the thumbnail to show that there are some data being fetched).

All of this is to reduce the first rendering, for now OHIF display an almost black page (except OHIF menu bar) during 5 seconds. During these 5 seconds the user do not know if there is a process pending so the app look like it have crashed (while it's not).

These 5 seconds probably do not exist with DCM4CHE because DCM4CHE duplicate a lot of data in it's internal database.
However the slowness of Orthanc might highlight some area of optimization that could benefit for all backends.

Best regards,

Salim

dannyrb:
@salimkanoun, I recommend trying out the new setting exposed in #1581. @swederik has also recently completed a change that allows thumbnails to show sooner. Anything faster (for thumbnails) will be difficult to justify until our redesign takes in later this year.

[...]

PR #1581, opened 2020-03-31
Limit concurrent series metadata requests

ckedar:
Further enhancements to series loading, to be able to show first meaningful display at the earliest.

### Issue
Concurrent loading of series metadata fills the request queue with metadata requests. 
Requests for thumbnail and viewport image get queued up after all the metadata requests.
As browser limits concurrent requests to a domain to 4, the thumbnail image and viewport image fetch requests get blocked.
Ultimately, thumbnail and viewport image is displayed only after all (actually, all - 3) metadata requests are done.

This could be a long wait,  if number of series are more and especially if server is slow to responding metadata requests. (https://github.com/OHIF/Viewers/issues/1536)

### Solutions
Solution implemented is to limit the max concurrent series metadata requests to 3 so that it leaves room for more important image requests to be served.

For slow servers, like mentioned in https://github.com/OHIF/Viewers/issues/1536, one might want to further restrict the max concurrent metadata requests to just 1 so as to not send the server in spin.
A config property `maxConcurrentMetadataRequests` can be specified app config for his purpose.

### DIFF
--- a/platform/core/src/classes/metadata/StudyMetadata.js
+++ b/platform/core/src/classes/metadata/StudyMetadata.js
@@ -322,11 +322,13 @@ export class StudyMetadata extends Metadata {
 
     const displaySets = this._createDisplaySetsForSeries(sopClassHandlerModules, series)
 
-    this._displaySets.forEach((displaySet, idx) => {
+    // Note: filtering in place because this._displaySets has writable: false
+    for (let i = this._displaySets.length - 1; i >= 0; i--) {
+      const displaySet = this._displaySets[i];
       if (displaySet.SeriesInstanceUID === series.getSeriesInstanceUID()) {
-        this._displaySets.splice(idx, 1);
+        this._displaySets.splice(i, 1);
       }
-    })
+    }
 
     displaySets.forEach(displaySet => {
       this.addDisplaySet(displaySet);
--- a/platform/viewer/public/config/default.js
+++ b/platform/viewer/public/config/default.js
@@ -117,4 +117,12 @@ window.config = {
     },
   ],
   cornerstoneExtensionConfig: {},
+  // Following property limits number of simultaneous series metadata requests.
+  // For http/1.x-only servers, set this to 5 or less to improve
+  //  on first meaningful display in viewer
+  // If the server is particularly slow to respond to series metadata
+  //  requests as it extracts the metadata from raw files everytime,
+  //  try setting this to even lower value
+  // Leave it undefined for no limit, sutiable for HTTP/2 enabled servers
+  // maxConcurrentMetadataRequests: 5,
 };
--- a/platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
+++ b/platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
@@ -153,25 +153,6 @@ const _updateStudyDisplaySets = (study, studyMetadata) => {
 const _sortStudyDisplaySet = (study, studyMetadata) => {
   studyMetadata.sortDisplaySets(study.displaySets);
 };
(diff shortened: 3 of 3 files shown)


swederik:
That's very interesting idea...

> As browser limits concurrent requests to a domain to 4, the thumbnail image and viewport image fetch requests get blocked.

I think this is usually 6, but this limit is not really a problem anymore with HTTP2 as I understand it. I do agree that the more open requests means that each individual request can often take long to return, though.

dannyrb:
I'm hesitant to accept this. I think a better solution is `stream prioritization` in `http/2`

Read: https://hpbn.co/http2/#stream-prioritization

It seems cloudflare has a header you can use to set the request's priority: https://blog.cloudflare.com/better-http-2-prioritization-for-a-faster-web/

I'd hope there is a generic way to do this. If not, we'd have to enforce prioritization on the client instead.

ckedar:
If we expect that we will alwasy be talking to HTTP/2 capabale server, then it would be fine to implement a solution using HTTP/2 feature.
But if significant chunk of our users (or their service providers) are expected to run HTTP/1.x-only servers then we may have to find solution that works in HTTP/1.x as well.

Based on quick read of the cloudfare blog article linked above, my understanding is that the described feature is a server side feature:

> ability to configure the prioritization is also exposed to Cloudflare Workers

It will be usable only for users hosting their server on Cloudfare. (paid users)
It will have to programmed or configured.
Even there, not sure if it will be right to de-prioritize all series metadata request and prioritize all image requests.
My guess is that image requests might be large in number compared to series metadata requests, not all truely high priority as many of them are for pre-fetching.
Essentially, priority of a request is context dependant and needs be assigned by client.

In HTTP/2, all requests are by default high priority.
Request priority can be changed through `options` provided to to `fetch`.
At present, both `cornerstoneWADOImageLoader` and `dicomweb-client` use `XMLHttpRequest`.
So far, I could not find a way to specify a priority on `XMLHttpRequest`.
Both these libraries, then, need to switch to `fetch` and also allow priority to be specified in `load...` methods.

I think solution in HTTP/2 domain will be a bigger change and also may not address all users.

Disclaimer:
I must say that I have very limited and shallow understanding of HTTP/2, coming from skimming through a few articles in last couple of hours. Please take all I said above with a pinch of salt.

dannyrb:
@ckedar, thank you for the thoughtful write-up! We generally recommend enabling `gzip` compression and `http/2` when advising "easy performance wins". You are right though, many could be stuck in an `http/1.x` world.

I also did a little more digging. I was aware of `http/2` prioritization, but I hadn't read or heard of a way to influence it from the client. It looks like, today, this is mostly set by the browser vendor. You can use ["priority hints" when loading resources](https://developers.google.com/web/fundamentals/performance/resource-prioritization). We do this already for preloading/precaching the files our PWA build process outputs. The `fetch` spec calls out that `priority` is `null` or defined by the [User-Agent (browser)](https://fetch.spec.whatwg.org/#concept-request-priority).

Cloudflare appears to offer the ability to influence streams by exposing priority as an option via headers. I don't know that there is a standard way to leverage this.

I think all of that is mostly right. More than happy to hear contradicting information.

-----

Sorry for the goose chase. Back to this issue. I think max number of series metadata requests makes sense. These can be long-running and tie up the server. Especially for Orthanc series metadata requests.

I think this is a good value add that we does the trick until we (if ever) reevaluate how we want to centralize and prioritize requests. My only ask is that we add the key + default value to `default.js` with a comment that indicates what the field does. At least this way it's minimally documented.

ckedar:
> add the key + default value to default.js with a comment that indicates what the field does

Done.
I have kept the default as unlimited.

PR #1586, opened 2020-03-31
OHIF-44: Header docz page examples

galelis:
**OHIF-44**

### Description
Improve NavBar example page with OHIF header 

### Preview
Screen Shot 2020-03-31 at 18 13 48

### Demo page
https://deploy-preview-1586--ohif.netlify.com/ui/components/NavBar


platform/ui/src/components/NavBar/NavBar.mdx
```diff
@@ -25,6 +25,41 @@ import { NavBar } from '@ohif/ui';
   </NavBar>
 </Playground>
 
+
```
dannyrb, line 28: You can remove the changes in this file, and then I think we're good for merge.
The `Study List` view captures these, and we discussed ditching `header.js`, so I'm all good.

PR #1610, opened 2020-04-06
OHIF-76: SidePanel component

galelis:
### Description
Create basic structure for Viewer page
Create a SidePanel component to be used on viewer

### Preview 
Apr-07-2020 17-13-06

## Demo page
https://deploy-preview-1610--ohif.netlify.com/ui/views/viewer

### DIFF
--- a/platform/ui/src/assets/icons/icon-nav-studies.svg
+++ b/platform/ui/src/assets/icons/icon-nav-studies.svg
@@ -1,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="25" height="25" viewBox="0 0 25 25">
+    <g fill="none" fill-rule="evenodd" stroke="#348CFD" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" transform="translate(1 1)">
+        <path d="M6 17H1c-.552 0-1-.448-1-1V1c0-.552.448-1 1-1h15c.552 0 1 .448 1 1v5"/>
+        <rect width="17" height="17" x="6" y="6" rx="1"/>
+    </g>
+</svg>
--- a/platform/ui/src/components/Icon/getIcon.jsx
+++ b/platform/ui/src/components/Icon/getIcon.jsx
@@ -6,7 +6,7 @@ import calendar from './../../assets/icons/calendar.svg';
 import cancel from './../../assets/icons/cancel.svg';
 import chevronDown from './../../assets/icons/chevron-down.svg';
 import chevronRight from './../../assets/icons/chevron-right.svg';
-import iconNavStudies from './../../assets/icons/icon-nav-studies.svg';
+import groupLayers from './../../assets/icons/group-layers.svg';
 import infoLink from './../../assets/icons/info-link.svg';
 import launchArrow from './../../assets/icons/launch-arrow.svg';
 import launchInfo from './../../assets/icons/launch-info.svg';
@@ -14,8 +14,8 @@ import listBullets from './../../assets/icons/list-bullets.svg';
 import logoOhifSmall from './../../assets/icons/logo-ohif-small.svg';
 import magnifier from './../../assets/icons/magnifier.svg';
 import notificationwarningDiamond from './../../assets/icons/notificationwarning-diamond.svg';
-import panelLeft from './../../assets/icons/panel-left.svg';
-import panelRight from './../../assets/icons/panel-right.svg';
+import pushLeft from './../../assets/icons/push-left.svg';
+import pushRight from './../../assets/icons/push-right.svg';
 import settings from './../../assets/icons/settings.svg';
 import sorting from './../../assets/icons/sorting.svg';
 import sortingActiveDown from './../../assets/icons/sorting-active-down.svg';
@@ -27,7 +27,7 @@ const ICONS = {
   cancel: cancel,
   'chevron-down': chevronDown,
   'chevron-right': chevronRight,
-  'icon-nav-studies': iconNavStudies,
+  'group-layers': groupLayers,
   'info-link': infoLink,
   'launch-arrow': launchArrow,
   'launch-info': launchInfo,
@@ -35,8 +35,8 @@ const ICONS = {
   'logo-ohif-small': logoOhifSmall,
   magnifier: magnifier,
   'notificationwarning-diamond': notificationwarningDiamond,
-  'panel-left': panelLeft,
-  'panel-right': panelRight,
+  'push-left': pushLeft,
+  'push-right': pushRight,
   settings: settings,
   'sorting-active-down': sortingActiveDown,
   'sorting-active-up': sortingActiveUp,
--- a/platform/ui/src/components/Icon/getIcon.jsx
+++ b/platform/ui/src/components/Icon/getIcon.jsx
@@ -1,18 +1,21 @@
 import React from 'react';
 // Icons
+
 import arrowDown from './../../assets/icons/arrow-down.svg';
-import seriesActive from './../../assets/icons/series-active.svg';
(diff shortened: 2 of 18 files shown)


platform/ui/src/components/SidePanel/SidePanel.jsx
```diff
@@ -0,0 +1,131 @@
+import React, { useState, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import { Button, Icon } from '@ohif/ui';
+
+const borderSize = 4;
+const expandedWidth = 320;
+const collapsedWidth = 60;
```
rodrigolabs, line 9: Maybe these values can be a non-required prop? And then you set the default values in the `defaultProps`.

platform/ui/src/components/SidePanel/SidePanel.jsx
```diff
@@ -0,0 +1,131 @@
...
+};
+
+const styleMap = {
+  open: {
+    left: { marginLeft: '0px' },
+    right: { marginRight: '0px' },
+  },
+  closed: {
+    left: { marginLeft: `-${expandedWidth - collapsedWidth - borderSize}px` },
+    right: { marginRight: `-${expandedWidth - collapsedWidth - borderSize}px` },
```
rodrigolabs, line 23: `-${expandedWidth - collapsedWidth - borderSize}px` can be a variable for a better readability, it's being used twice.

platform/ui/src/components/SidePanel/SidePanel.jsx
```diff
@@ -0,0 +1,131 @@
...
+  },
+};
+
+const baseClassName =
```
rodrigolabs, line 27: Some components we adopt the naming for base classes for `const baseClasses = ''`, what do you think? We can later go through all components to standardize them as well.

platform/ui/src/components/SidePanel/SidePanel.jsx
galelis: I think we can leave it to the pass 2.

platform/ui/src/components/SidePanel/SidePanel.jsx
galelis: Updated.

platform/ui/src/components/SidePanel/SidePanel.jsx
galelis: Updated.

platform/ui/src/components/SidePanel/SidePanel.jsx
```diff
@@ -0,0 +1,132 @@
...
+const openIconName = {
+  left: 'panel-left',
+  right: 'panel-right',
+};
+
+const SidePanel = ({
+  side,
+  className,
+  children,
+  defaultIsOpen,
```
rodrigolabs, line 51: `defaultIsOpen` I think is not a good name, we can rename it to `isOpen` or just `open`. It's clearer to have a prop like this than with a `default` prefix. I saw you maybe did it because there's a `useState` using `isOpen`. Maybe you can rename the prop and rename to defaultIsOpen when getting the variable here.

```suggestion
  isOpen: defaultIsOpen,
```

What do you think?

platform/ui/src/views/Viewer/Viewer.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
rodrigolabs, line 1: Same as `StudyList.js`, I don't think we will need this component to be exposed, or even to exist.

platform/ui/src/components/SidePanel/SidePanel.jsx
galelis: I thought about that and changed to its current name, in my idea, once we set it to isOpen or Open, gives the user the idea that the parent is controling.. and right now, its not. thats why I changed it to defaultIsOpen. but we can rename it, just don't think open or isOpen good either

platform/ui/src/views/Viewer/Viewer.js
galelis: I agree, but before we have in on '@ohif/viewer' I am totally fine by having it here, so we can test it out of the docz.

platform/ui/src/components/SidePanel/SidePanel.jsx
rodrigolabs: Hmm, It's true! Sorry, I hadn't noticed. We can leave it like this and think later if the open/closed state will be controlled in the component.

platform/ui/src/views/Viewer/Viewer.js
rodrigolabs: I'm ok with that. My only concern is that we will have to go through it later to delete those files that are left behind. Perhaps they will not be many. Let's think about it later.

platform/ui/src/assets/icons/icon-nav-studies.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="25" height="25" viewBox="0 0 25 25">
```
dannyrb, line 1: @rodrigolabs & @galelis, having trouble remembering. Do we want width/height cleared from icons? The only thing that's important is `viewBox`, right? Having these removed would make it easier to scale w/ styles

dannyrb: Having trouble remembering if these matter for aspect ratio, or if you try to fix that by using a grid system when creating the SVG

platform/ui/src/components/Icon/getIcon.jsx
```diff
@@ -1,37 +1,42 @@
...
+import iconNavStudies from './../../assets/icons/icon-nav-studies.svg';
 import infoLink from './../../assets/icons/info-link.svg';
 import launchArrow from './../../assets/icons/launch-arrow.svg';
 import launchInfo from './../../assets/icons/launch-info.svg';
+import listBullets from './../../assets/icons/list-bullets.svg';
 import logoOhifSmall from './../../assets/icons/logo-ohif-small.svg';
 import magnifier from './../../assets/icons/magnifier.svg';
 import notificationwarningDiamond from './../../assets/icons/notificationwarning-diamond.svg';
+import panelLeft from './../../assets/icons/panel-left.svg';
+import panelRight from './../../assets/icons/panel-right.svg';
```
dannyrb, line 18: These icons could be used for something other than panel. Maybe `pushLeft` and `pushRight`

platform/ui/src/components/Icon/getIcon.jsx
```diff
@@ -1,37 +1,42 @@
...
 // Icons
+
 import arrowDown from './../../assets/icons/arrow-down.svg';
-import seriesActive from './../../assets/icons/series-active.svg';
-import seriesInactive from './../../assets/icons/series-inactive.svg';
 import calendar from './../../assets/icons/calendar.svg';
 import cancel from './../../assets/icons/cancel.svg';
 import chevronDown from './../../assets/icons/chevron-down.svg';
 import chevronRight from './../../assets/icons/chevron-right.svg';
+import iconNavStudies from './../../assets/icons/icon-nav-studies.svg';
```
dannyrb, line 9: Similarly, I'm not sure the naming should isolate this to studies

platform/ui/src/components/Icon/getIcon.jsx
galelis: I liked your idea. just updated .

platform/ui/src/components/Icon/getIcon.jsx
galelis: Cool, just update the naming as well.

PR #1612, opened 2020-04-07
feat: Toolbar Header

rodrigolabs:
## Description

This PR creates an example of how to build a Header with Toolbar using basically NavBar, Icons, Buttons, etc..

## Preview

image

image

## Deploy Preview

https://deploy-preview-1612--ohif.netlify.com/


package.json
```diff
@@ -35,6 +35,7 @@
     "test:e2e:dist": "lerna run test:e2e:dist --stream",
     "test:e2e:serve": "lerna run test:e2e:serve --stream",
     "see-changed": "lerna changed",
+    "docs:preview": "lerna run docs:preview --stream",
```
rodrigolabs, line 38: Created this script to run the documentation from the application root folder.

platform/ui/src/components/IconButton/IconButton.jsx
```diff
@@ -87,15 +87,16 @@ const IconButton = ({
   disabled = defaults.disabled,
   type = defaults.type,
   fullWidth = defaults.fullWidth,
+  onClick,
```
rodrigolabs, line 90: Exposing `onClick` prop.

platform/ui/tailwind.config.js
```diff
@@ -435,9 +436,10 @@ module.exports = {
     },
     padding: (theme) => theme('spacing'),
     placeholderColor: (theme) => theme('colors'),
-    stroke: {
+    stroke: (theme) => ({
+      ...theme('colors'),
```
rodrigolabs, line 440: cc @galelis here's the setting that will allow us to use `stroke-primary-main` to colorize SVG strokes. With that, we will be able to set different colors for fill and stroke. :)

platform/ui/src/gatsby-theme-docz/tailwind.css
```diff
@@ -22,3 +22,14 @@
 .ohif-scrollbar::-webkit-scrollbar-thumb:window-inactive {
   @apply bg-primary-main;
 }
+
+/* TOOLTIP WORKAROUND FOR ARROW UP */
+.tooltip::before {
```
galelis, line 27: Is this going to be used on all tooltips? maybe worth adding a custom class for this scenario.

platform/ui/src/components/Icon/getIcon.jsx
```diff
@@ -21,11 +22,20 @@ import sorting from './../../assets/icons/sorting.svg';
 import sortingActiveDown from './../../assets/icons/sorting-active-down.svg';
 import sortingActiveUp from './../../assets/icons/sorting-active-up.svg';
 
+/** Tools */
+import toolZoom from './../../assets/icons/tool-zoom.svg';
+import toolCapture from './../../assets/icons/tool-capture.svg';
+import toolLayout from './../../assets/icons/tool-layout.svg';
+import toolMore from './../../assets/icons/tool-more-menu.svg';
+import toolMove from './../../assets/icons/tool-move.svg';
+import toolWindow from './../../assets/icons/tool-window-level.svg';
```
dannyrb, line 31: I can't think of better names at the moment. I think the `tool` prefix works for now 👍

platform/ui/src/components/IconButton/IconButton.jsx
```diff
@@ -87,15 +87,16 @@ const IconButton = ({
...
   ...rest
 }) => {
   const buttonElement = useRef(null);
 
   const handleOnClick = (e) => {
     buttonElement.current.blur();
-    if (rest.onClick) {
-      rest.onClick(e);
+    if (onClick) {
+      onClick(e);
```
dannyrb, line 99: I sometimes see a default `no operation` function for things like this, so we don't have to litter in checks:

```onClick: () => {}```

File: platform/ui/src/gatsby-theme-docz/tailwind.css [hunk 7]
```diff
@@ -22,3 +22,14 @@
...
+  @apply bg-primary-dark absolute z-10;
+  content: '';
+  width: 14px;
+  height: 1px;
+  top: -1px;
+  left: 50%;
+  transform: translateX(-50%);
+}
```
dannyrb, line 35: This is the gatsby theme or our theme?

rodrigolabs: yeah, some very specific custom styles are here. Better split?

dannyrb: Eventually. Fine to punt for now.

platform/ui/src/components/IconButton/IconButton.jsx
rodrigolabs: I've fixed the `defaultProps` definition for this component and created a default method for `onClick`. Thanks!

platform/ui/src/gatsby-theme-docz/tailwind.css
rodrigolabs: Yes! I was only thinking if we have in the future tooltip position like `up`, `left`, `bottom`...
But for now, we have only this kind of Tooltip.

platform/ui/src/gatsby-theme-docz/tailwind.css
rodrigolabs: I've created a separate CSS to use when we need to create custom classes for the components.

PR #1622, opened 2020-04-09
fix: Revert "Reduce bundle size"

dannyrb:
Reverts OHIF/Viewers#1575


dannyrb:
Temporary revert until auth header fix is in place

Issue #1458, opened 2020-02-21
Segmentation Settings UI - Phase 2


JamesAPetts:
Second part of #1370.

Now #1392 is complete, we need two more features from the original ticket:

## Overview:
- There needs to be a hide/show button for each segment on the segmentation.
- Clicking on the left of the panel should jump to the ticket.

Example: Eye icon on the right of the panel to toggle visibility. clicking on left panel containing color and number will jump to segment:
![Screenshot 2020-03-17 at 11 15 07](https://user-images.githubusercontent.com/25818497/76851350-06889a00-6841-11ea-8f90-88911e477a8e.png)

## Implementation Details:

- There should be UI to hide and show each segment within a segmentation.
This should hook up to the following cornerstoneTools segmentation module endpoints:

```js
setters.toggleSegmentVisibility(
  elementOrEnabledElementUID,
  segmentIndex,
  labelmapIndex
);

isSegmentVisible(
  elementOrEnabledElementUID,
  segmentIndex,
  labelmapIndex
)
```

- Clicking on the left of this panel should jump to the centre of the segment. For cornerstone: this can be done by the following procedure:

For a given segment Index:
- In the relevant `labelmap3D` Check if the segment is on each `labelmap2D` using the `segmentsOnLabelmap2D` array, store a list of all the indicies of these `labelmap2D` objects.
- Given the list of labelmap2D indicies, get the minimum and the maximum index and take the average.
- The index in the array closest to this average is the slice to jump to.

vtkjs specific logic shall be a seperate ticket:

JamesAPetts:
On hold until further discussion about timeline for new OHIF UI.

JamesAPetts:
@pieper @fedorov Is the desired functionality here correct?

fedorov:
@JamesAPetts yes, this looks fine to me. Sorry for the delay, I missed the mention.

fedorov:
The screencast below demonstrates some of the issues with the current interface:

It is basically not clear whether segmentations are available for a given study. Segmentations list shows as empty, and looks like one needs to load imaging series that has segmentations first before segmentations list is populated. And even after that there seems to be a delay in that list being populated. See screencast here:

https://www.screencast.com/t/g6wzeltT

[...]

PR #1643, opened 2020-04-16
Segmentation Settings UI - Phase 2 #1458: Add segment part 2 (jump to frame and visibility toggle)

igoroctaviano:
#1458
 
- hide/show button for each segment on the segmentation.
- Clicking on the left of the panel should jump to the ticket.

### PR Checklist


JamesAPetts:
Hey, the jump to segment works really well on [this PET example]( https://deploy-preview-1643--ohif.netlify.app/pwa/viewer/1.3.6.1.4.1.14519.5.2.1.2744.7002.373729467545468642229382466905).

But on [this prostate example](https://deploy-preview-1643--ohif.netlify.app/pwa/viewer/1.3.6.1.4.1.14519.5.2.1.7311.5101.170561193612723093192571245493) it doesn't seem to work, scroll away from the segments and try to jump to them. Is it having an issue because its out of plane? Cornerstone is 2D so I doubt this is the case, could you look into this?

Also the second consequtive time you click on a segment on the sidepanel it doesn't jump to the segment. Follow these steps:
-> click on a segment in the side panel, the cornerstone window will jump to this segment.
-> scroll away so the segment is out of view.
-> click on the segment in the side panel again, the segment doesn't jump.

We want to always jump, even if the segment is "active", as you may want to recentre it.

extensions/dicom-rt/src/index.js
```diff
@@ -51,8 +48,9 @@ export default {
               if (study && study.series) {
                 for (let j = 0; j < study.series.length; j++) {
                   const series = study.series[j];
-                  console.log(studies);
-                  if (['RTSTRUCT', 'RTPLAN', 'RTDOSE'].includes(series.Modality)) {
+                  if (
```
JamesAPetts, line 51: I realise this was from a previous PR, but this only caught my eye just now. We should only show the side panel if we have an RTSTRUCT, for now. We don't have any support for the other two types, maybe just add a comment "Could be expanded to contain `RTPLAN` and `RTDOSE` information in the future".

extensions/dicom-segmentation/src/components/SegmentationPanel/SegmentationPanel.js
```diff
@@ -28,13 +28,19 @@ const refreshViewport = () => {
 /**
  * SegmentationPanel component
  *
- * @param {Object} props
- * @param {Array} props.studies
```
JamesAPetts, line 32: `studies` is still being passed in?

extensions/dicom-segmentation/src/components/SegmentationPanel/SegmentationPanel.js
```diff
@@ -108,126 +129,201 @@ const SegmentationPanel = ({ studies, viewports, activeIndex, isOpen }) => {
...
+
+          const validIndexList = [];
+          labelmap3D.labelmaps2D.forEach((labelMap2D, index) => {
+            if (labelMap2D.segmentsOnLabelmap.includes(segmentNumber)) {
+              validIndexList.push(index);
+            }
+          });
+          const avg = array => array.reduce((a, b) => a + b) / array.length;
+          const average = avg(validIndexList);
+          const closest = validIndexList.reduce((prev, curr) => {
```
JamesAPetts, line 245: Nice.

igoroctaviano:
> Hey, the jump to segment works really well on [this PET example](https://deploy-preview-1643--ohif.netlify.app/pwa/viewer/1.3.6.1.4.1.14519.5.2.1.2744.7002.373729467545468642229382466905).
> 
> But on [this prostate example](https://deploy-preview-1643--ohif.netlify.app/pwa/viewer/1.3.6.1.4.1.14519.5.2.1.7311.5101.170561193612723093192571245493) it doesn't seem to work, scroll away from the segments and try to jump to them. Is it having an issue because its out of plane? Cornerstone is 2D so I doubt this is the case, could you look into this?
> 
> Also the second consequtive time you click on a segment on the sidepanel it doesn't jump to the segment. Follow these steps:
> -> click on a segment in the side panel, the cornerstone window will jump to this segment.
> -> scroll away so the segment is out of view.
> -> click on the segment in the side panel again, the segment doesn't jump.
> 
> We want to always jump, even if the segment is "active", as you may want to recentre it.
> 
> Some other minor comments in-line.

@JamesAPetts About the consecutive time, you click the seg to jump after moving the stack, it's because of a known bug in OHIF, it happens in RT and measurements as well, the viewport specific data is set but if the user moves the stack, the state continues the same so it doesn't cause a rerender in the second click. I had a solution merged in react-cornerstone-viewport that would call a callback after the user stops moving the stack (debouncing) so we could update the state. I'll check with @dannyrb about it.

About the prostate example, the segmentsOnLabelmap2D are all the same which causes the closest number to be 9 for all segments. Maybe something wrong with the way I'm getting the frameindex?

Checking other comments now.

JamesAPetts:
> About the prostate example, the segmentsOnLabelmap2D are all the same which causes the closest number to be 9 for all segments. Maybe something wrong with the way I'm getting the frameindex?

Ahh interesting, this probably means inside the SEG they are encoding each frame for each segment even if its empty. This is slightly annoying, checking each frame for occupancy would be slow, which is why we do the `segmentsOnLabelmap2D` thing in the first place. I liked this approach of checking which frames are encoded, as It'll also make the VTK integration easier as we just need to find the central frame, and then search only that frame for voxels to set the crosshairs for all viewports.

TLDR: This is a "poorly encoded", but completely valid SEG.

Check if this affects any of the IDC MVP datasets, if not a better way of finding the correct slice may be out of scope for this ticket. We could do one search through each frame that claims it contains the segment and then cache the central frame that actually does, but this could be very slow for large CTs if the data is encoded similarly to that prostate SEG.

cc @pieper @fedorov for opinions on the above.

pieper:
❤️ the progress on this!

Regarding empty frames, couldn't we just record in some kind of meta-metadata while decoding the SEG?

fedorov:
> TLDR: This is a "poorly encoded", but completely valid SEG.

I wouldn't call it "poorly encoded". Keeping empty frames has its advantages, since there are systems that cannot deal with skipped empty frames. Also, when a SEG with empty frames is converted to a NIfTI or like, it will have the geometry which is matching the geometry of the image, which is really important for those AI researchers that ignore origin/spacing/orientation information (and there are such AI researchers).

TLDR: OHIF Viewer should be robust enough to handle this kind of SEGs.

fedorov:
As aside, I was not aware of existence of SEGs in the prostateX collection - I am glad you came across it!

fedorov:
> Hey, the jump to segment works really well on [this PET example](https://deploy-preview-1643--ohif.netlify.app/pwa/viewer/1.3.6.1.4.1.14519.5.2.1.2744.7002.373729467545468642229382466905).

I tried and failed to figure out how the jump to segment works on that PET example ... :-\

Can you elaborate?

igoroctaviano:
> this PET example

@fedorov you can try this flow again, I just pushed the fix, a check was missing in jumpToImage command. This specific PET just screamed the need for it.

fedorov:
@igoroctaviano thank you, it works for me now - very nice!

fedorov:
@igoroctaviano what is the source of the SEG for the ProstateX collection?

According to the information on the TCIA page for that collection, the only modality it has is MR (which is consistent with what I thought, and that's why I was surprised when you mentioned SEG in the context of that collection).

JamesAPetts:
> ❤️ the progress on this!
> 
> Regarding empty frames, couldn't we just record in some kind of meta-metadata while decoding the SEG?

Hey, yeah we could do that, currently dcmjs just looks which frames are present and flags these as occupied, not if they are empty. We can iterate through each frame and check its actually occupied by a single positive bit, and only add these ones to the list.

JamesAPetts:
> According to the information on the TCIA page for that collection, the only modality it has is MR (which is consistent with what I thought, and that's why I was surprised when you mentioned SEG in the context of that collection).

I asked Erik about this, he doesn't remember when he uploaded these, but thinks the SEGs were made later and uploaded as tests, i.e. they aren't on TCIA.

JamesAPetts:
I've talked to Igor and he's going to add this:

> Regarding empty frames, couldn't we just record in some kind of meta-metadata while decoding the SEG?

To dcmjs.

PR #1647, opened 2020-04-17
fix: `showStudyList` config

dannyrb:
w/ `showStudyList` `false`

image

w/ `showStudyList` `true` or `undefined`

image

### DIFF
--- a/platform/viewer/src/components/Header/Header.js
+++ b/platform/viewer/src/components/Header/Header.js
@@ -1,17 +1,13 @@
 import React, { useState, useEffect } from 'react';
 import { Link, withRouter } from 'react-router-dom';
 import { withTranslation } from 'react-i18next';
-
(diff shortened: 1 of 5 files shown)


platform/viewer/src/components/Header/Header.js
```diff
@@ -105,7 +101,11 @@ function Header(props) {
 }
 
 Header.propTypes = {
-  home: PropTypes.bool.isRequired,
+  // Study list, /
+  linkText: PropTypes.string,
+  linkPath: PropTypes.string,
+  useLargeLogo: PropTypes.bool,
```
dannyrb, line 107: `home` prop to props that are more flexible and less context specific.
Use of app config is determined one layer higher. Helps for a dumber component.

JamesAPetts:
LGTM.

PR #1674, opened 2020-04-24
feat: New @ohif/viewer project structure + routes

rodrigolabs:
**OHIF-116**

## Description

This PR creates a "new" `@ohif/viewer` project structure and adds basic route navigation for Study List `/` and Viewer `/viewer/:id`.

More cleaning in the project can be done later, such as `package.json` and/or webpack config. The purpose with this PR was just to have a fresh `viewer` to start the new implementation.

## Routing

There's a simple configuration in `.env` to define which route should be used `BrowserRouter` or `HashRouter`.


platform/viewer/.env
```diff
@@ -6,5 +6,6 @@
 # Be careful not to commit anything sensitive to source control.
 #
 
-PUBLIC_URL=/
-APP_CONFIG=config/default.js
+REACT_APP_PUBLIC_URL=/
+REACT_APP_APP_CONFIG=config/default.js
+REACT_APP_USE_HASH_ROUTER=false
```
dannyrb, line 11: We're not using CRA, so you shouldn't need to prefix these env vars with `REACT_APP_`

platform/viewer/README 2.md
```diff
@@ -0,0 +1,68 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
dannyrb, line 1: We will for sure have to eject. All or most of the tooling should still be what we're targeting, unless there's something in particular that jumps out at you? This may be as simple as keeping the `.webpack` folder, but simplifying the build commands

platform/viewer/src/serviceWorker.js
```diff
@@ -0,0 +1,141 @@
+// This optional code is used to register a service worker.
+// register() is not called by default.
+
+// This lets the app load faster on subsequent visits in production, and gives
```
dannyrb, line 4: We'll likely just use the Workbox service worker plugin for webpack

rodrigolabs:
@dannyrb @galelis 

As already discussed, I've reverted this branch to work on the same structure we currently have in `platform/viewer` doing just clean up to have a fresh project to start implementing the next steps.

A new PR Review is needed. Thanks :)

PR #1781, opened 2020-06-11
Add Desktop Electron client

swederik:
This is a first pass at an Electron client for OHIF. It's a desktop application that includes [dicomweb-server](https://github.com/dcmjs-org/dicomweb-server) and [pouchdb](https://github.com/pouchdb/pouchdb) to store/provide data to the viewer.

There's a ton of small stuff left to do:
- [ ] Use hash router so we start on the correct page
- [ ] Fix WADO-RS for dicomweb-server
- [ ] Add production build steps to CI
- [x] Make sure all of our icons and other assets are showing up properly
- [ ] Simplify development usage, since the current approach is pretty weird
- [ ] Fix the ugly broken upload box
- [ ] Add a way for users to delete studies

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,6 +122,143 @@ jobs:
             docker push ohif/viewer:PR_BUILD-$CIRCLE_BUILD_NUM
 
   ###
+  # Workflow: PR_OPTIONAL_ELECTRON_BUILD
+  # Borrowed from Rick Herrick from XNAT (https://bitbucket.org/xnatdev/xnat-desktop-client/src/master/.circleci/config.yml)
+  ###
+  ELECTRON_BUILD_WIN_AND_LINUX:
+    environment:
+      TERM: xterm # Enable colors in term
+      QUICK_BUILD: true
+      SKIP_SERVICE_WORKER: true
+      PUBLIC_URL: ./
+      DEBUG: electron-builder
+    docker:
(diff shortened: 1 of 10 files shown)


swederik:
Tagging @jeremypatrickdahan - in case you know all the production build stuff and would be willing to help out, it would be awesome to have Windows, Mac, and Linux versions build and added to Github releases by CircleCI.

jeremypatrickdahan:
Would be glad to give a hand ! I am not super familiar with CircleCI, but I will give it a try

dannyrb:
Realistically, I won't have time to look at this until after August. If someone else wants to take a crack, please feel free.

> Use hash router so we start on the correct page

Can use a env var to swap router implementation at build time. This is what we're doing in the redesign branch.

> Add production build steps to CI

Some good examples of this in the wild. Worth noting that you'll need to build on your target platform. I think CircleCI now offers all platforms? Previously, you had to use multiple services as they had different offerings:

- https://circleci.com/build-environments/windows/

Pluralsight has a really good guide + example app for v2.0 w/ Electron
Egghead.io has an information dense guide for v3.0
Current major version is 9.x, but majority of info holds true.

If you have specific questions, I can poke my head in.

pieper:
Could also consider using github actions for this.  

I'm also interesting in seeing this move forward, so thanks for digging in. 👍

Issue #1638, opened 2020-04-14
Regression: WADOURI requires image level metadata


knopkem:
### Describe the Bug

Since the big metadata overhaul beginning from V3.7.8 fetching images via WADOURI will fail with cornerstone reporting failure to decode the image. Previously the metadata query for WADOURI required only Study, Series and SOPInstanceUIDs to work. Now several image level attributes need to be returned as well, e.g. pixelspacing, rows, cols, bits allocated, bits stored, highbit etc.
This is especially annoying when using OHIF with a dicomweb to dimse proxy as C-FIND won't be able to query those image level attributes from PACS, thus it's necessary to fetch the data first and to parse it.

### What steps can we follow to reproduce the bug?

* Install Orthanc with dicomweb plugin
* Import some data then test with default settings and latest OHIF viewer. 
** Test viewing: Viewer should render the images correctly
* Now change the [dicomweb plugin](https://book.orthanc-server.com/plugins/dicomweb.html) config to return only the most basic attributes:
** SeriesMetadata: MainDicomTags 
* Restart the server
* Test viewing: Viewer fails to render the image

Alternative approach: Use the [dicomweb-proxy](https://github.com/knopkem/dicomweb-proxy) and update the viewer version (currently at V3.7,7) in the public folder 

I tried to find the issue myself but the code changes for the big overhaul are immense and I need to better understand the code first.

dannyrb:
Do you have any recommendations around how to proceed or address this issue?

knopkem:
Sorry, I did not have the time to look at this any further, but will try to look into this asap.

knopkem:
I could find the issue, please see my PR here https://github.com/OHIF/Viewers/pull/1785

sedghi:
I will document the absolute minimum metadata to run OHIF soon, you can track it here https://github.com/OHIF/Viewers/issues/3093

The additional regrettablyis not supported at present and is not included in our current plans. However, we would be delighted to review your pull request if you would like to contribute.

PR #1785, opened 2020-06-12
MetadataProvider should return null if metadata values are not set

knopkem:
The fix is rather simple: it adds 2 checks if the requested metadata for a module is valid.
Explanation why this is needed: See https://github.com/OHIF/Viewers/issues/1638
Explanation why this fix works: In version 3.7.7 and before the MetadataProvider returned null when the metadata for a DICOM module (e.g. ImagePlaneModule) was not available. This allowed cornerstone to try the next available metadata provider. Since cornerstone in the case of wadouri has all the data it needs (complete p10 file) it then uses it's fallback internal parser to gather IMAGE level information like pixel spacing, rows and cols etc. When using wadouri it should not be needed to provide this information in the metadata call.


JamesAPetts:
Hey @knopkem , could you give me a description of your use case? We wanted to to take control over the metadata usage in OHIF, and the `cornerstoneWADOImageLoader` metadataProvider-like groups are there just for compatability. If using DICOMWeb, the headers should contain this information, and if using the JSON launch, we also read the metadata directly from the P10 file. We wish to drop reliance on the inbuilt metadata providers with time, which is I'm hesistent to accept this change without understanding the issue the current paradigm prevents.

Additionally, if you really wish to force the CWIL metadata loaders to be used as priority, I would just change the priority order of the loaders instead.

knopkem:
Hi @JamesAPetts, I see your point, so here is my usecase:
I'm maintaining a [DICOMWEB to DIMSE proxy](https://github.com/knopkem/dicomweb-proxy) allowing DICOMWEB clients (e.g. viewers) to connect to any regular/legacy PACS (that do not speak DICOMWEB), which I believe can be very useful. It currently comes with the OHIF viewer embedded but is not limited to it (so e.g. it should be possible to run the latest OHIF master against it)
For a DIMSE-proxy QIDO to C-FIND and WADOURI to C-GET/Move are perfect matches as they are almost identical in use. However in the OHIF workflow no matter if wadouri or wadors are configured the metadata call happens when switching from studyview to viewer. I understand that for the metadata API a server is supposed to return the full metadata without bulkdata to the caller which is a problem for any proxy as it would need to fetch all files of the series/study first before answering (which would probably lead to timeouts on large datasets). Before the metdata overhaul (v.3.7.8) it was possible to return only very minimal information (uids) and the viewer would still work. In fact, in the wadorui configuration the requesting the metadata is unnecessary as it will fetch the complete file anyway which contains everything it needs. I know that for wadors (using bulkdata) the viewer needs this information at some point. 
So ideally OHIF would use different workflows for wadouri vs wadors but that would be a bigger change and I don't think this is easy to maintain either.
Anyway, coming back to my PR. I must agree that I'm not a big fan of the change either as it feels a bit uncomplete, changing priority of parsers might be cleaner. IMHO a better fix would be to improve the check in cornerstone.js directly when fetching metadata for a specific type (e.g. ImagePlaneModule). At that point it should verify that the metadata provided by the provider is actually sufficient for rendering instead of failing later. I could prepare a PR for that too if this has any chance of getting merged. I understand that fading out the internal parser would at one point reintroduce the issue (but this would buy some time). Let me know. Cheers.

JamesAPetts:
Hey @ knopkem , thanks so much for the in depth response! The context certainly helps.

TLDR: The main issue I had with this PR here is you check if `BitsAllocated`, `Rows`, `Columns`, likely just because these are the fields you need in your case. However to be pragmatic, if these aren't here these legacy groups are kind of useless. I'm happy merging this unless anyone has any large objections.

The rest are just thoughts/discussion:

> At that point it should verify that the metadata provided by the provider is actually sufficient for rendering instead of failing later.

Bingo, the thing I don't like about rejections of groups like this is you can't know a priori if a certain the selection of data returned by one provider is "more useful" than another.

This is something I've thought about before, and a more granular `metadataProvider` seems like a really good solution to this, the backwards compatability (mainly with `cornerstoneWADOImageLoader` (CWIL)) is the kicker though. With OHIF you can do something like `cornerstone.metaData.get("SOPInstanceUID", imageId)` and you can just pull the field you want. If you had multiple providers like this, you could fallback through them all for individual fields. The (honestly arbitrary) CWIL-style groups are a bit awkward. It'd be cool if we could request groups of fields and they would all fall back, but for now its the instance or one "top-level" field or sequence.


> I understand that for the metadata API a server is supposed to return the full metadata 

This is the crux of the issue that made us flip out how we serve the metadata. We got a load of questions here like "Why doesn't 3D work when I use the JSON launch", and so we needed to make an API layer where we can satisfy a minimum contract, and everything should work from there (secondary, also the fact that the `imageId` is a rather strange concept, as often the image and the metadata are obtained seperately, so having the `imageId` tied to the URI of the pixeldata can be less useful).

The main reasons I really wanted to understand the issue, is if it "works slightly" again in some cases where the metadata isn't present, this "minimum metadata requirement to use OHIF" gets a bit more unclear.

TLDR2: Thanks for the interesting discussion :), I'm not really opposed to merging this. Long term we should update cornerstone and other libraries to have more granular providers, but this isn't anywhere on the roadmap right now.

JamesAPetts:
I've poked people who care, lets merge this.

knopkem:
Hi @JamesAPetts , thanks for the long answer, much appreciated that you took the time for it. 

> The main issue I had with this PR here is you check if BitsAllocated, Rows, Columns, likely just because these are the fields you need in your case
Yes I know, this looks a bit arbitrary, but in fact I deliberately limited the check to values that are critical for the viewer to actually render anything, this could be taken further by including all level 1 attributes of a module but then there are conditional (1C) attributes shouldn't those be included as well and soon you end up with quite some complex logic all in imho the wrong place. So this is only making sure to try another parser if those values are undefined.

> if we could request groups of fields and they would all fall back, but for now its the instance or one "top-level" field or sequence.
I agree with you on the group issue, which basically comes back to the problem that we have tags with different types so returning all or nothing doesn't make sense here and a tag level granularity would definitely help.

> We got a load of questions here like "Why doesn't 3D work when I use the JSON launch"
I must agree that I had similar questions (e.g. "what is the minimum metadata needed to get wadors running" and then later "ok it works, but MPR is not there"). For 3D/MPR my best guess is that it needs orientation vectors (IOP and IPP) but this becomes try and error. Here adding better logging would be really helpful (e.g. missing Image Orientation Patient vector - disabling MPR feature).

PR #1803, opened 2020-06-23
Feat/ui v2 dropdown

rodrigolabs:
**OHIF-182**

## Description

Creates a `Dropdown` component in the `ui` library and use it for the Settings menu.

This PR also implements a basic Modal for the `Learn More` (search info) and the `About` and `Preferences` in the settings.

## Preview

image

image

### PR Checklist


extensions/default/src/ViewerLayout/Header.jsx
```diff
@@ -3,10 +3,28 @@ import PropTypes from 'prop-types';
...
+import { NavBar, Svg, Icon, IconButton, Dropdown, useModal } from '@ohif/ui';
 
 function Header({ children }) {
   const history = useHistory();
+  const { show } = useModal();
+
+  const showAboutModal = () => {
+    const modalComponent = () => <div>About modal</div>;
+    show({
+      title: 'About',
```
igoroctaviano, line 15: I'm not sure but translations will be added in separate tasks?

extensions/default/src/ViewerLayout/Header.jsx
```diff
@@ -42,16 +60,29 @@ function Header({ children }) {
...
+          <Dropdown
+            titleElement={
+              <IconButton
+                variant="text"
+                color="inherit"
+                size="initial"
+                className="text-primary-active"
+                onClick={() => {}}
+              >
+                <React.Fragment>
```
igoroctaviano, line 72: This fragment is necessary?

extensions/default/src/ViewerLayout/Header.jsx
```diff
@@ -42,16 +60,29 @@ function Header({ children }) {
...
-            </React.Fragment>
-          </IconButton>
+          <Dropdown
+            titleElement={
+              <IconButton
+                variant="text"
+                color="inherit"
+                size="initial"
+                className="text-primary-active"
+                onClick={() => {}}
```
igoroctaviano, line 70: Remove onClick?

extensions/default/src/ViewerLayout/Header.jsx
```diff
@@ -42,16 +60,29 @@ function Header({ children }) {
...
+              >
+                <React.Fragment>
+                  <Icon name="settings" /> <Icon name="chevron-down" />
+                </React.Fragment>
+              </IconButton>
+            }
+            list={[
+              { title: 'About', icon: 'info', onClick: () => showAboutModal() },
```
igoroctaviano, line 78: Tip of the day 😄  i think you know about it: We should avoid passing arrow funcs directly to UI elements whenever possible, I know this is just an initial step towards final implementation but i think we can train ourselves to define them in the nearest component or parent as a handler and use the handler instead. This help us avoid rerenderings since the arrow func will be attached to the component instance.

platform/viewer/src/components/PreferencesDropdown/PreferencesDropdown.jsx
```diff
@@ -0,0 +1,51 @@
...
+      content: modalComponent,
+      title: 'Preferences',
+    });
+  };
+
+  return (
+    <Dropdown
+      titleElement={
+        <IconButton
+          variant="text"
```
igoroctaviano, line 28: I think this code is duplicated in the Header, maybe we should use the preferences modal instead?

platform/viewer/src/routes/WorkList/WorkList.jsx
```diff
@@ -364,17 +365,7 @@ function WorkList({ history, data: studies, dataSource }) {
           <span className="mr-3 text-lg text-common-light">
             FOR INVESTIGATIONAL USE ONLY
           </span>
```
igoroctaviano, line 367: If the navbar is the same for two views, what do you think about a layout with the navbar included? so every new page would use the same layout and set the content of the navbar if needed, like longitudinal-workflow where the only thing that changed was the content in the middle of the navbar. This way we can reuse the investigational + navbar + preferencesdropdown code everywhere.

platform/ui/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,107 @@
...
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import { Icon, Typography } from '@ohif/ui';
+
+const Dropdown = ({ titleElement, title, list }) => {
+  const [open, setOpen] = useState(false);
+  const element = useRef(null);
+
+  const renderTitleElement = () => {
```
igoroctaviano, line 11: titleElement is kinda ambiguous to me. If this is an element socket, where its located? lets document this prop and how to use it.

platform/ui/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+  };
+
+  const renderList = () => {
+    const itemsAmount = list.length;
+
+    return (
+      <div
+        className={classnames(
+          'absolute origin-top-right transition duration-300 transform top-100 right-0 mt-2 z-10 bg-primary-dark border border-secondary-main rounded shadow',
```
igoroctaviano, line 40: To avoid a huge list of classes I'm breaking them into separate names variables in another project like:
```
const flex = 'flex flex-direction.... all stuff related to flex';
const border = 'border border-top. shadow.. all stuff related to border';
const font = 'font font-whatever etc...';
const theme = 'bg-color or something';
```
And then use it like: classNames(font, theme, whatever...)

Let me know what you think about this approach, I know it a lot of variables but I find it easier to reason about what style i have to update.

platform/ui/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,107 @@
...
+              'flex px-4 py-2 cursor-pointer items-center transition duration-300 hover:bg-secondary-main',
+              {
+                'border-b border-secondary-main': itemsAmount !== idx + 1,
+              }
+            )}
+            onClick={() => {
+              setOpen(false);
+
+              if (item.onClick) {
+                item.onClick();
```
igoroctaviano, line 60: You can define a ```const noop = () => {}``` in this jsx file outside the component and use it as default in proptypes? this removes the need to use these types of ifs. (Danny's tip)

extensions/default/src/ViewerLayout/Header.jsx
rodrigolabs: We're not adding it yet, but I added it anyway to avoid untranslated texts throughout the application 👍

extensions/default/src/ViewerLayout/Header.jsx
rodrigolabs: not really 😊  thanks

platform/ui/src/components/Dropdown/Dropdown.jsx
rodrigolabs: Good catch! As this is a nested prop of `list`, which is an array, I would have to create a workaround to deal with its default props values. I ended up using an es6 approach (destructuring assignment), I think it's clean and simple. If I had to define a `defaultProps`, another approach would be to split the "list item" into a component to define its default values.

What do you think?

extensions/default/src/ViewerLayout/Header.jsx
rodrigolabs: Are you mentioning to have methods like `showAboutModal` centralized in the top-parent component and then pass to the `Header` as props?

https://github.com/OHIF/Viewers/blob/140c562519ed9173d53d3fea06cb08ca867b1ea3/extensions/default/src/ViewerLayout/index.jsx#L82

rodrigolabs: Are you mentioning to have methods like `showAboutModal` centralized in the top-parent component and then pass to the Header as props? Here for example:

https://github.com/OHIF/Viewers/blob/140c562519ed9173d53d3fea06cb08ca867b1ea3/extensions/default/src/ViewerLayout/index.jsx#L82

We could even maybe think in a better way to split/reuse this Header component since it's being used in `extensions/default` and `platform/viewer`. I'll take a look at it tomorrow :D

platform/viewer/src/components/PreferencesDropdown/PreferencesDropdown.jsx
rodrigolabs: Yeah, that's what I was talking in that comment: https://github.com/OHIF/Viewers/pull/1803#discussion_r444632898
Let's discuss offline about it

platform/ui/src/components/Dropdown/Dropdown.jsx
rodrigolabs: agree! I kept this approach from the old Dropdown component, but I ended up changing a little the component implementation, avoiding ambiguity by removing the title/titleElement. The children itself is the "title" and it can be customizable. A Dropdown Icon `˅` will be visible by default to have a basic dropdown style if the title is a single string. It can be hidden through the new prop `showDropdownIcon`. Let me know your opinion on this new approach

extensions/default/src/ViewerLayout/Header.jsx
igoroctaviano: @rodrigolabs I mean talking about js itself and React, arrow functions is basically a function declaration so every render this will be a different point in memory that could possibly cause unnecessary renders. Moving the declaration to the nearest parent or any component so we have a function instance for each component instance, this way we are not redeclaring functions in each rerender.

platform/ui/src/components/Dropdown/Dropdown.jsx
igoroctaviano: I think its fine to just extract the list item into its own component declaration inside this dropdown component and then use defaultProps or destructuring assignment if you wish. Using composing pieces that have semantic names is a lot easier to reason :)

rodrigolabs: I've split the list item into a separated component to set its `defaultProps` instead of destructuring assignment, as we're doing in other components

platform/ui/src/components/Dropdown/Dropdown.jsx
rodrigolabs: This is an interesting approach. We're currently trying to group class names by subject. I think this is something we can definitely discuss but for the future, to avoid inconsistency by having different approaches for now.

extensions/default/src/ViewerLayout/Header.jsx
rodrigolabs: oh gotcha, you're right! for this scenario, we can use `useCallback`.

platform/ui/src/assets/icons/unlink.svg
```diff
@@ -0,0 +1,11 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  aria-labelledby="unlink"
+  viewBox="0 0 512 512"
+  width="1em"
+  height="1em"
```
dannyrb, line 6: No width/height

platform/ui/src/assets/icons/link.svg
```diff
@@ -0,0 +1,11 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  viewBox="0 0 32 32"
+  aria-labelledby="title"
+  width="1em"
+  height="1em"
```
dannyrb, line 6: no width/height

platform/ui/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  DropdownItem.defaultProps = {
+    icon: '',
+    onClick: () => {},
+  };
+
+  DropdownItem.propTypes = {
+    title: PropTypes.string.isRequired,
+    icon: PropTypes.string,
```
dannyrb, line 42: No icon by default?

platform/ui/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,128 @@
...
+    onClick: PropTypes.func,
```
dannyrb, line 43: Pointless without click? Should be required?

platform/ui/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,128 @@
...
+    index: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
```
dannyrb, line 44: If string, maybe identifier/key instead of index?

platform/ui/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,128 @@
...
+
+import { Icon, Typography } from '@ohif/ui';
+
+const Dropdown = ({ children, showDropdownIcon, list }) => {
+  const [open, setOpen] = useState(false);
+  const element = useRef(null);
+
+  const DropdownItem = ({ title, icon, onClick, index }) => {
+    const itemsAmount = list.length;
+    const isLastItem = itemsAmount === index + 1;
```
dannyrb, line 13: Maybe require int index if we're going to `+ 1`?

dannyrb: Why do we need parent to provide index at all if iterator can apply key?

platform/ui/src/components/InputNumber/InputNumber.jsx
```diff
@@ -0,0 +1,50 @@
...
+InputNumber.propTypes = {
+  label: PropTypes.string.isRequired,
+  isSortable: PropTypes.bool,
+  sortDirection: PropTypes.oneOf(['ascending', 'descending', 'none']),
+  onLabelClick: PropTypes.func,
+  value: PropTypes.number,
+  onChange: PropTypes.func.isRequired,
+};
+
+export default InputNumber;
```
dannyrb, line 50: Not sure if this needs sortable props/func if it's never used as a list filter?

platform/ui/src/components/Modal/Modal.jsx
```diff
@@ -41,20 +54,32 @@ const Modal = ({
 
   return (
     <ReactModal
-      className={classNames(className)}
+      className={classNames(
```
dannyrb, line 57: Why does the user need a way to pass className?

platform/viewer/src/App.jsx
```diff
@@ -1,12 +1,13 @@
...
 import React from 'react';
 import PropTypes from 'prop-types';
+import i18n from '@ohif/i18n';
+import { I18nextProvider } from 'react-i18next';
 import { BrowserRouter, HashRouter } from 'react-router-dom';
 import {
   DialogProvider,
   Modal,
   ModalProvider,
-  Notification,
```
dannyrb, line 9: 👀 Why?

platform/ui/src/components/Dropdown/Dropdown.jsx
rodrigolabs: yes, it actually should be required even to avoid all those defaultProps/destructuring assignment for onClick. Thanks for remembering 🤦  -- thanks

platform/ui/src/components/Dropdown/Dropdown.jsx
rodrigolabs: I was using it just to handle a "border: none" for the last-child, but after research I saw that Tailwind gives us this power by using `last:border-b-0`. Thanks, Tailwind!

platform/ui/src/components/Dropdown/Dropdown.jsx
rodrigolabs: removed this logic due to the tailwind class

platform/viewer/src/App.jsx
rodrigolabs: linter was highlighting as an unused variable

platform/viewer/src/routes/WorkList/WorkList.jsx
rodrigolabs: had an offline conversation with @dannyrb and he said that we'll keep it this way for now and take care of it later.

platform/ui/src/components/InputNumber/InputNumber.jsx
rodrigolabs: `InputNumber` was actually nuked, we're now using the `Input` component for it.

platform/ui/src/components/Modal/Modal.jsx
rodrigolabs: It was removed here: d7cd2f9e511c2cad94725c47ad2b68fea5e3624c for this PR: #1799

PR #1805, opened 2020-06-24
OHIF-176: v2 - Render non-tracked measurements as a dashed lines.

JamesAPetts:
This PR adds dashed lines for non-tracked measurements.

### DIFF
--- a/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
+++ b/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
@@ -219,7 +219,6 @@ function TrackedCornerstoneViewport({
     SliceThickness,
   } = displaySet.images[0];
 
-  console.log('TRACKED SERIES');
   console.log();
 
   if (trackedSeries.includes(SeriesInstanceUID) !== isTracked) {
--- a/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
+++ b/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import cornerstone from 'cornerstone-core';
+import cornerstoneTools from 'cornerstone-tools';
 import CornerstoneViewport from 'react-cornerstone-viewport';
 import OHIF, { DicomMetadataStore } from '@ohif/core';
 import {
@@ -12,6 +13,18 @@ import {
 import debounce from 'lodash.debounce';
 import throttle from 'lodash.throttle';
 import { useTrackedMeasurements } from './../getContextModule';
+
+// TODO -> Get this list from the list of tracked measurements.
+const {
+  ArrowAnnotateTool,
+  BidirectionalTool,
+  EllipticalRoiTool,
+  LengthTool,
+} = cornerstoneTools;
+
+const BaseAnnotationTool = cornerstoneTools.importInternal(
+  'base/BaseAnnotationTool'
+);
 
 // const cine = viewportSpecificData.cine;
 
@@ -27,6 +40,7 @@ function TrackedCornerstoneViewport({
   viewportIndex,
 }) {
   const [trackedMeasurements] = useTrackedMeasurements();
+
   const [
     { activeViewportIndex, viewports },
     dispatchViewportGrid,
@@ -34,6 +48,8 @@ function TrackedCornerstoneViewport({
   // viewportIndex, onSubmit
   const [viewportDialogState, viewportDialogApi] = useViewportDialog();
   const [viewportData, setViewportData] = useState(null);
+  const [element, setElement] = useState(null);
+  const [isTracked, setIsTracked] = useState(false);
   // TODO: Still needed? Better way than import `OHIF` and destructure?
   // Why is this managed by `core`?
   useEffect(() => {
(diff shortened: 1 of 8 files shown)


extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -131,6 +219,13 @@ function TrackedCornerstoneViewport({
     SliceThickness,
   } = displaySet.images[0];
 
+  console.log('TRACKED SERIES');
```
dannyrb, line 222:
```suggestion
```

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -131,6 +219,12 @@ function TrackedCornerstoneViewport({
     SliceThickness,
   } = displaySet.images[0];
 
+  console.log();
```
dannyrb, line 222:
```suggestion
```

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -42,6 +58,77 @@ function TrackedCornerstoneViewport({
...
 
+  useEffect(() => {
+    if (!element) {
+      return;
+    }
+    const allTools = cornerstoneTools.store.state.tools;
+    const toolsForElement = allTools.filter(tool => tool.element === element);
+
+    toolsForElement.forEach(tool => {
+      debugger;
```
dannyrb, line 69:
```suggestion
```

dannyrb: Remove debugger

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -42,6 +58,76 @@ function TrackedCornerstoneViewport({
...
+    const toolsForElement = allTools.filter(
+      tool => tool.element === targetElement
+    );
+
+    toolsForElement.forEach(tool => {
+      if (
+        tool instanceof ArrowAnnotateTool ||
+        tool instanceof BidirectionalTool ||
+        tool instanceof EllipticalRoiTool ||
+        tool instanceof LengthTool
```
dannyrb, line 106: Hah

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -42,6 +58,76 @@ function TrackedCornerstoneViewport({
...
+      ) {
+        const configuration = tool.configuration;
+
+        configuration.renderDashed = !isTracked;
+
+        tool.configuration = configuration;
```
dannyrb, line 79: How are you setting dashed for measurements that will always be temp? (angle)

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -42,6 +58,76 @@ function TrackedCornerstoneViewport({
...
+      }
+    });
+
+    const enabledElement = cornerstone.getEnabledElement(element);
+
+    if (enabledElement.image) {
+      cornerstone.updateImage(element);
+    }
+  }, [isTracked]);
```
dannyrb, line 88: This runs initially, and each time this prop is updated.
Because we set local state, that will trigger a re-render. I think it'd be possible to do this by only setting this initially, and then running the same logic each time we get a `subscription` event from the tracking service (without using internal state to trigger)

We can keep for now. Small refactor item that sits outside of the sprint?

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -42,6 +58,76 @@ function TrackedCornerstoneViewport({
...
+
+  const onElementEnabled = evt => {
+    const eventData = evt.detail;
+    const targetElement = eventData.element;
+
+    debugger;
```
dannyrb, line 94:
```suggestion
```

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -42,6 +58,75 @@ function TrackedCornerstoneViewport({
...
+
+        configuration.renderDashed = !isTracked;
+
+        tool.configuration = configuration;
+      } else if (tool instanceof BaseAnnotationTool) {
+        const configuration = tool.configuration;
+
+        configuration.renderDashed = true;
+
+        tool.configuration = configuration;
```
dannyrb, line 117: Ah, okay. Here is where you set the config for all annotation tools as the default config.

dannyrb: Wish we had a way to know when/if a tool is "added" to an element so we could make this more fault tolerant. But ignore for now

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
JamesAPetts: Yeah, it'd be good if we could check types of tools that are registered to the measurement service. I think we could actually once the measurement service definitions are updated 🤔

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
JamesAPetts: > Small refactor item that sits outside of the sprint?

Will make :+1:.

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
JamesAPetts: Yeah, this will at least always work, but we could do with better helpers in CST for a bunch of this more granular stuff.

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
JamesAPetts: They are set once on element enabled.

PR #1806, opened 2020-06-24
OHIF-151: Replace the "..." icon in more menu with active tool in menu.

igoroctaviano:
If a tool in the more menu is activated the "..." icon of the more menu should be replaced with the active tool.
If an action in the toolbar menu is clicked the more menu should close and the icon should remain as it is ("..." or active tool from the more menu).

### PR Checklist

### DIFF
--- a/extensions/default/src/ViewerLayout/ToolbarButtonNestedMenu.jsx
+++ b/extensions/default/src/ViewerLayout/ToolbarButtonNestedMenu.jsx
@@ -20,6 +20,8 @@ function NestedMenu({ children, label, icon }) {
     };
   }, [isOpen]);
 
+  const isActive = isOpen || children.props.children.some(c => c.props.isActive);
+
   return (
     <ToolbarButton
       id="NestedMenu"
@@ -27,7 +29,7 @@ function NestedMenu({ children, label, icon }) {
       icon={icon}
       onClick={toggleNestedMenu}
       dropdownContent={isOpen && children}
-      isActive={isOpen}
+      isActive={isActive}
       type="primary"
     />
   );
--- a/extensions/default/src/ViewerLayout/ToolbarButtonNestedMenu.jsx
+++ b/extensions/default/src/ViewerLayout/ToolbarButtonNestedMenu.jsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { ToolbarButton } from '@ohif/ui';
 
-function NestedMenu({ children, label, icon }) {
+function NestedMenu({ children, label, icon, isActive }) {
   const [isOpen, setIsOpen] = useState(false);
 
   const toggleNestedMenu = () => setIsOpen(!isOpen);
@@ -20,8 +20,6 @@ function NestedMenu({ children, label, icon }) {
     };
   }, [isOpen]);
 
-  const isActive = isOpen || children.props.children.some(c => c.props.isActive);
-
   return (
     <ToolbarButton
       id="NestedMenu"
@@ -29,7 +27,7 @@ function NestedMenu({ children, label, icon }) {
       icon={icon}
       onClick={toggleNestedMenu}
       dropdownContent={isOpen && children}
-      isActive={isActive}
+      isActive={isActive || isOpen}
       type="primary"
     />
   );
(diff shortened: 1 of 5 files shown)


extensions/default/src/ViewerLayout/ToolbarButtonNestedMenu.jsx
```diff
@@ -2,32 +2,31 @@ import React, { useEffect, useState } from 'react';
...
-      }
+  const toggleNestedMenu = () => setIsOpen(!isOpen);
+
+  const closeNestedMenu = () => {
+    if (isOpen) {
+      setIsOpen(false);
     }
+  };
+
+  useEffect(() => {
```
dannyrb, line 16: Much cleaner. Thanks

extensions/default/src/ViewerLayout/index.jsx
```diff
@@ -55,16 +55,22 @@ function ViewerLayout({
     };
   };
 
+  const defaultTool = { icon: 'tool-more-menu', label: 'More' };
   const [toolbars, setToolbars] = useState({ primary: [], secondary: [] });
+  const [activeTool, setActiveTool] = useState(defaultTool);
+  const onSecondaryClickHandler = () => setActiveTool(defaultTool);
```
dannyrb, line 61: The secondary toolbar shouldn't impact which icon is shown for the nested/more menu?
Only buttons _in_ the nested menu that are tools that can be set?

platform/core/src/services/ToolBarService/ToolBarService.js
```diff
@@ -154,12 +154,15 @@ export default class ToolBarService {
...
+      if (props && props.onClick) {
+        props.onClick(evt, btn, btnSection, props);
+      }
     };
 
     return {
       id,
       Component: component || buttonType.defaultComponent,
-      componentProps: Object.assign({}, props, { onClick }), //
+      componentProps: Object.assign({}, btn.props, props, { onClick }), //
```
dannyrb, line 165: This is tricky. I wouldn't just override the `onClick` you intend to set in the `viewer` project for all buttons. It might make more sense to wrap the buttons `onClick`, if it has one? Kind of chaining them, so you don't squash some user defined functionality by accident.

Each toolbar module button definition can optionally provide a `clickHandler`. We have some examples of these in: `extensions/default/src/getToolbarModule`

When our UI binds `onClick` for a button component (here), it does:

buttonType.clickHandler -->
btn.props.onClick -->
btn.props.clickHandler

When you call `getButtonSection`, you're passing in `props: { onClick: fn }` and adding your onClick to the bottom of that chain of callbacks (from 3x to 4x).

buttonType.clickHandler -->
btn.props.onClick -->
btn.props.clickHandler -->
props.onClick (from `viewer`)

------

Just thinking out loud.

extensions/default/src/ViewerLayout/index.jsx
igoroctaviano: I did this to get the nested menu back to its original icon, I'll take a better look and understand the existent code, I'm missing a good part of it. Thanks.

extensions/default/src/ViewerLayout/index.jsx
```diff
@@ -86,11 +92,10 @@ function ViewerLayout({
 
             if (!isNested) {
               const { id, Component, componentProps } = toolDef;
-
               return <Component key={id} id={id} {...componentProps} />;
             } else {
               return (
-                <NestedMenu>
+                <NestedMenu icon={activeTool.icon} label={activeTool.label}>
```
dannyrb, line 98: I think these props need to be exposed, but I only "radioGroup" tools in this menu would update the icon, if one is set? You might be able to determine this by. This is super tricky to "get right" in a reusable way that doesn't introduce coupling. We may need to do a "good enough" solution here, and create a follow-up issue.

You could bind your `setActiveTool` setState functions to:

- Tools that are in nested menus + of type `ohif.radioGroup`; when clicked, setActiveTool
- Tools that are not in a nested menu + of type `ohif.radioGroup`; when clicked, setActiveTool(defaulTool)

Long-term, you'd almost have to have the ToolbarService pub/sub tool config changes, or find a different/better way to expose toolbar "state" like this, and then just watch it.

igoroctaviano: @dannyrb updated, lets discuss this new approach.

PR #1808, opened 2020-06-25
v2 - Feat/ohif 156

JamesAPetts:
One way cornerstoneTools measurements => measurements service.

Do we need to implement the toAnnotation functions at the current time? They are unused for now, and likely not a priority until we have multiple sources of annotations (cornerstoneTools, vtk).

The originally proposed pattern of having each measurement fit to one SCOORD type doesn't gel well with SR, so we'll need to discuss @dannyrb.

I removed mathematical calculations from the UI, as these can be cashed in the measurement service representation, and shouldn't be calculated every render.

### DIFF
--- a/extensions/cornerstone/src/init.js
+++ b/extensions/cornerstone/src/init.js
@@ -90,6 +90,7 @@ export default function init({ servicesManager, configuration }) {
 
   /* Add extension tools configuration here. */
   const internalToolsConfig = {
+    /* TODO ArrowAnnotate input
     ArrowAnnotate: {
       configuration: {
         getTextCallback: (callback, eventDetails) =>
@@ -98,6 +99,7 @@ export default function init({ servicesManager, configuration }) {
           callInputDialog(data, eventDetails, callback),
       },
     },
+    */
   };
 
   /* Abstract tools configuration using extension configuration. */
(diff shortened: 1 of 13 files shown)


extensions/cornerstone/src/init.js
```diff
@@ -90,6 +90,7 @@ export default function init({ servicesManager, configuration }) {
 
   /* Add extension tools configuration here. */
   const internalToolsConfig = {
+    /* TODO ArrowAnnotate input
```
dannyrb, line 93: 👍  Thanks for commenting out so default prompt works

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/index.js
```diff
@@ -169,14 +171,22 @@ function _mapMeasurementToDisplay(measurement, index) {
  * @param {*} points
  * @param {*} pixelSpacing
  */
-function _getDisplayText(points, pixelSpacing, seriesNumber, instanceNumber) {
+function _getDisplayText(
```
dannyrb, line 174: Will be interesting to see how this works for generic groups of SCOORDs / measurements

extensions/cornerstone/src/init.js
JamesAPetts: :+1: We have an issue for the other prompt somewhere in JIRA.

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/index.js
JamesAPetts: In a later refactor, I would actually want the _getDisplayText function to be part of the defition of the measurement-service level tool schema.

PR #1811, opened 2020-06-25
[OHIF-177] - First pass at SR STOW with Arrow Annotate, Ellipse, Length and Bidirectional.

JamesAPetts:
When you click the Create Report button, we should serialize all tracked measurements to DICOM SR and STOW to PACS.

I also have local download as that was the first test, currently bound to export, but with no fancy UI yet.

No UI for successful post yet. After posting refresh the page after a few seconds and you can view the SR with the SR viewport.

### DIFF
--- a/extensions/default/src/DicomWebDataSource/index.js
+++ b/extensions/default/src/DicomWebDataSource/index.js
@@ -12,8 +12,13 @@ import getImageId from './utils/getImageId';
 import * as dcmjs from 'dcmjs';
 import { retrieveStudyMetadata } from './retrieveStudyMetadata.js';
 
-const { naturalizeDataset } = dcmjs.data.DicomMetaDictionary;
+const { DicomMetaDictionary, DicomDict } = dcmjs.data;
+
+const { naturalizeDataset, denaturalizeDataset } = DicomMetaDictionary;
 const { urlUtil } = utils;
+
+const VERSION_NAME = 'OHIF-0.1';
+const SR_TRANSFER_SYNTAX_UID = '1.2.840.10008.1.2.1';
 
 /**
  *
@@ -126,6 +131,32 @@ function createDicomWebApi(dicomWebConfig) {
             });
           });
         },
+      },
+    },
+    store: {
+      dicom: async dataset => {
+        const meta = {
+          FileMetaInformationVersion:
+            dataset._meta.FileMetaInformationVersion.Value,
+          MediaStorageSOPClassUID: dataset.SOPClassUID,
+          MediaStorageSOPInstanceUID: dataset.SOPInstanceUID,
+          TransferSyntaxUID: SR_TRANSFER_SYNTAX_UID,
+          ImplementationClassUID: DicomMetaDictionary.uid(),
+          ImplementationVersionName: VERSION_NAME,
+        };
+
+        const denaturalized = denaturalizeDataset(meta);
+        const dicomDict = new DicomDict(denaturalized);
+
+        dicomDict.dict = denaturalizeDataset(dataset);
+
+        const part10Buffer = dicomDict.write();
+
+        const options = {
+          datasets: [part10Buffer],
+        };
+
+        await wadoDicomWebClient.storeInstances(options);
       },
     },
     retrieveSeriesMetadata: async ({ StudyInstanceUID } = {}) => {
--- a/extensions/default/src/Panels/WrappedPanelStudyBrowser.jsx
+++ b/extensions/default/src/Panels/WrappedPanelStudyBrowser.jsx
@@ -20,7 +20,7 @@ function WrappedPanelStudyBrowser({
 }) {
   // TODO: This should be made available a different way; route should have
   // already determined our datasource
-  const dataSource = extensionManager.getDataSources('dicomweb')[0];
+  const dataSource = extensionManager.getDataSources()[0];
   const _getStudiesForPatientByStudyInstanceUID = getStudiesForPatientByStudyInstanceUID.bind(
     null,
     dataSource
--- a/extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/index.js
+++ b/extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/index.js
@@ -125,7 +125,6 @@ function PanelMeasurementTableTracking({ servicesManager, extensionManager }) {
         trackedSeries.includes(m.referenceSeriesUID)
     );
 
-    console.log(servicesManager);
 
     const dataSources = extensionManager.getDataSources();
     // TODO -> Eventually deal with multiple dataSources.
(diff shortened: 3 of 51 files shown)


extensions/default/src/Panels/WrappedPanelStudyBrowser.jsx
```diff
@@ -20,7 +20,7 @@ function WrappedPanelStudyBrowser({
 }) {
   // TODO: This should be made available a different way; route should have
   // already determined our datasource
-  const dataSource = extensionManager.getDataSources('dicomweb')[0];
+  const dataSource = extensionManager.getDataSources()[0];
```
dannyrb, line 23: Didn't you add some magic way to get the "default" data source? Or doesn't the mode route publish/identify the selected data source in some way?

JamesAPetts: If you don't give it a tag you get the default. We could add an explicit proxy for the same function also, just saw this explictly assumption and nixxed it as I walked by.

platform/core/src/DataSources/IWebApiDataSource.js
```diff
@@ -44,13 +45,20 @@ function create({
...
+      throw new Error(
+        'store.dicom(naturalizedDicom, StudyInstanceUID) not implemented for dataSource.'
+      );
+    },
+  };
+
   return {
     query: query || defaultQuery,
     retrieve: retrieve || defaultRetrieve,
+    store: store || defaultStore,
```
dannyrb, line 59: We really need to discuss the shape of this API pre big release.
This follows the current pattern. Data sources are weird.

extensions/default/src/Panels/WrappedPanelStudyBrowser.jsx
dannyrb: So it returns a list of "active data sources", with the first being the default? Providing a name filters, but still gives a list?

A proxy would be neat. If there's enough here for a follow-up issue, great. If not, just resolve this comment thread.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -128,6 +133,32 @@ function createDicomWebApi(dicomWebConfig) {
...
+
+        const part10Buffer = dicomDict.write();
+
+        const options = {
+          datasets: [part10Buffer],
+        };
+
+        await wadoDicomWebClient.storeInstances(options);
+      },
+    },
```
dannyrb, line 161: You make it look so easy 😎

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/index.js
```diff
@@ -102,6 +105,37 @@ function PanelMeasurementTableTracking({ servicesManager, commandsManager }) {
...
+  const onCreateReportClick = () => {
+    const measurements = MeasurementService.getMeasurements();
+    const trackedMeasurements = measurements.filter(
+      m =>
+        trackedStudy === m.referenceStudyUID &&
+        trackedSeries.includes(m.referenceSeriesUID)
+    );
+
+    console.log(servicesManager);
+    console.log(extensionManager);
```
dannyrb, line 129:
```suggestion
```

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/index.js
```diff
@@ -102,6 +105,36 @@ function PanelMeasurementTableTracking({ servicesManager, commandsManager }) {
...
+
+  const onCreateReportClick = () => {
+    const measurements = MeasurementService.getMeasurements();
+    const trackedMeasurements = measurements.filter(
+      m =>
+        trackedStudy === m.referenceStudyUID &&
+        trackedSeries.includes(m.referenceSeriesUID)
+    );
+
+    console.log(servicesManager);
```
dannyrb, line 128:
```suggestion
```

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/index.js
```diff
@@ -102,6 +105,35 @@ function PanelMeasurementTableTracking({ servicesManager, commandsManager }) {
 
   const activeMeasurementItem = 0;
 
+  const onExportClick = () => {
```
dannyrb, line 108: `onExportClick` is the name of the prop. For it to read smoother you can name this `localDownloadStructuredReport`

So the prop binding reads:

`onExportClick={localDownloadStructredReport}`

dannyrb: If this doesn't depend on reactive values, you don't need to define this function in the component. Re-renders will recreate it. You can shift it outside of the component in the same file, or a separate file.

platform/core/src/DICOMSR/dataExchange.js
```diff
@@ -78,4 +82,112 @@ const storeMeasurements = async (measurementData, filter, server) => {
   }
 };
 
-export { retrieveMeasurements, storeMeasurements };
+const downloadReport = measurementData => {
```
dannyrb, line 85: JS Doc or description of what `measurementData` is?

platform/core/src/DICOMSR/dataExchange.js
```diff
@@ -78,4 +82,112 @@ const storeMeasurements = async (measurementData, filter, server) => {
...
+
+  const report = MeasurementReport.generateReport(
+    filteredToolState,
+    cornerstone.metaData
+  );
+
+  return report.dataset;
+};
+
+const storeMeasurements = async (measurementData, dataSource) => {
```
dannyrb, line 106: Same here. If this takes a `dataSource`, can we call out the required props in JSDoc?

`@param {func} dataSource.store.dicom - does stuff`

platform/core/src/DICOMSR/dataExchange.js
```diff
@@ -78,4 +82,112 @@ const storeMeasurements = async (measurementData, filter, server) => {
...
+    }
+
+    return {
+      message: 'Measurements saved successfully',
+    };
+  } catch (error) {
+    log.error(
+      `[DICOMSR] Error while saving the measurements: ${error.message}`
+    );
+    throw new Error('Error while saving the measurements.');
```
dannyrb, line 133: If you haven't already, separate issue calling out UI notification for success/fail.
Outside of sprint, blocked by design/user/func requirements

extensions/default/src/Panels/WrappedPanelStudyBrowser.jsx
JamesAPetts: It returns the active data source, the datasource might have multiple servers. In our example it returns the list of dicomweb datasources. We have the [0] everywhere throughout the code so we could build with multiple in mind ^^.

dannyrb: Ah! Thanks for the refresher.

platform/core/src/DataSources/IWebApiDataSource.js
JamesAPetts: Agreed. I at least chose not to call it stow as that is DICOMWeb specific.

platform/core/src/DICOMSR/dataExchange.js
dannyrb: https://precisionmetrics.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=OHIF&modal=detail&selectedIssue=OHIF-177

dannyrb:
Merging for now. Follow-up on anything you feel is relevant.

PR #1812, opened 2020-06-25
OHIF-152: Create window level dropdown menu

igoroctaviano:
### PR Checklist

### DIFF
--- a/platform/ui/src/components/IconButton/IconButton.jsx
+++ b/platform/ui/src/components/IconButton/IconButton.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
 const baseClasses =
-  'text-center items-center justify-center outline-none transition duration-300 ease-in-out font-bold focus:outline-none';
+  'text-center items-center justify-center outline-none font-bold focus:outline-none';
 
 const roundedClasses = {
   none: '',
@@ -113,7 +113,7 @@ const IconButton = ({
 };
 
 IconButton.defaultProps = {
-  onClick: () => {},
+  onClick: () => { },
   color: 'default',
   disabled: false,
   fullWidth: false,
--- a/platform/ui/src/components/Tooltip/Tooltip.jsx
+++ b/platform/ui/src/components/Tooltip/Tooltip.jsx
@@ -24,7 +24,7 @@ const arrowPositionStyle = {
   },
 };
 
-const Tooltip = ({ content, isSticky, position, tight, children }) => {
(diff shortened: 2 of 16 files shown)


platform/ui/src/components/Tooltip/Tooltip.jsx
```diff
@@ -39,7 +39,7 @@ const Tooltip = ({ content, isSticky, position, tight, children }) => {
     }
   };
 
-  const isOpen = isSticky || isActive;
+  const isOpen = (isSticky || isActive) && !isDisabled;
```
dannyrb, line 42: Interesting. I'm guessing we have to do it this way because the `Tooltip` wraps it's children? Works for me, but can we do an inline comment above the prop?

```js
/** prevents tooltip from rendering despite hover/active/sticky */
isDisabled: false,
```

dannyrb:
Generating deploy preview to test behavior now

platform/ui/src/components/Tooltip/Tooltip.jsx
```diff
@@ -84,10 +84,12 @@ Tooltip.defaultProps = {
   tight: false,
   isSticky: false,
   position: 'bottom',
+  isDisabled: false
 };
 
 Tooltip.propTypes = {
   // Allow null
```
dannyrb, line 91:
```suggestion
/** prevents tooltip from rendering despite hover/active/sticky */
```

platform/ui/src/components/IconButton/IconButton.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
 const baseClasses =
-  'text-center items-center justify-center outline-none transition duration-300 ease-in-out font-bold focus:outline-none';
+  'text-center items-center justify-center outline-none font-bold focus:outline-none';
```
rodrigolabs, line 6: I'm curious about why you removed the transition from the IconButton @igoroctaviano

igoroctaviano: @rodrigolabs well remembered! so, at first at thought it was making the buttons feel a little leggy in the transitions but then i had trouble matching the transition of the button and the new  arrow below the button. Lets discuss about these transitions, do you feel them smooth enough?

rodrigolabs: I think it's ok to keep the transition, since other things like the Button itself have the same transition, to keep the consistency. I can revert it in my PR. And then later we can try to do this minor adjustment in the arrow below the button. Is that ok for you?

PR #1814, opened 2020-06-25
[OHIF-154] Dannyrb/ohif 154/restore command functionality for toolbar

dannyrb:
See most recent commit for next steps.


extensions/cornerstone/src/commandsModule.js
```diff
@@ -66,7 +68,7 @@ const commandsModule = ({ servicesManager }) => {
       }
     },
     invertViewport: () => {
-      const activeViewportIndex = 0;
+      const { activeViewportIndex } = ViewportGridService.getState();
```
dannyrb, line 71: New way to fetch activeViewportIndex from exposed service

extensions/cornerstone/src/init.js
```diff
@@ -54,6 +56,23 @@ export default function init({ servicesManager, configuration }) {
...
+    const { viewportIndex, enabledElement } = evt.detail;
+
+    console.log(evt, viewportIndex, enabledElement);
+    setEnabledElement(viewportIndex, enabledElement);
+  };
+
+  document.addEventListener(
+    'ohif-cornerstone-enabled-element-event',
+    handleOhifCornerstoneEnabledElementEvent
+  );
```
dannyrb, line 75: The 'ohif-cornerstone-enabled-element-event' is a way for extensions that extend the Cornerstone Viewport in this extension to notify it that a `cornerstone Enabled Element` event occurred, so we can track its reference, so commands continue to function (this is a bit nuts, and I've left some comments regarding clean-up)

PR #1821, opened 2020-06-26
[OHIF-173]: Jump between (potentially) tracked measurements in tracking viewport

JamesAPetts:
OHIF-173 - Action bar for tracking viewport: If tracked, jump between tracked measurements. If not tracked jumped between potentially trackable measurements.


platform/ui/src/components/Viewport/Viewport.jsx
```diff
@@ -2,12 +2,17 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { ViewportActionBar, Notification, Button } from '@ohif/ui';
 
-const Viewport = ({ viewportIndex, onSeriesChange, studyData, children }) => {
+const Viewport = ({
```
dannyrb, line 5: I think this is an unused component -- or only used for the docz "view" to replicate the UI we're targeting

platform/ui/src/components/ViewportActionBar/ViewportActionBar.jsx
```diff
@@ -14,7 +14,7 @@ const ViewportActionBar = ({
   studyData,
   showNavArrows,
   showPatientInfo,
-  onSeriesChange,
+  onNavigationClick,
```
dannyrb, line 17: I agree. This really should be an empty container where you can optionally render all these bits as children

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -327,3 +353,59 @@ async function _getViewportData(dataSource, displaySet) {
...
+    }
+  } else if (direction === 'right') {
+    measurementIndex++;
+
+    if (measurementIndex === measurementCount) {
+      measurementIndex = 0;
+    }
+  }
+
+  console.log(measurementIndex);
```
dannyrb, line 404:
```suggestion
```

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -217,15 +221,37 @@ function TrackedCornerstoneViewport({
...
     setIsTracked(!isTracked);
   }
 
+  /**
+   * If the series is tracked, jumps through tracked measurements in this series.
+   * if the series is not tracked, jumps through potentially trackable measurements
+   * (those tracked by the MeasurementService.)
+   * @param {string} direction 'right' or 'left' navigation through the measurements.
+   */
+  function switchMeasurement(direction) {
```
dannyrb, line 234: See clarification comments:
- https://precisionmetrics.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=OHIF&modal=detail&selectedIssue=OHIF-173&assignee=5ea2fa311f32260c130c86b5

PR #1822, opened 2020-06-26
[OHIF-192] - Viewport letter only if using multiple viewport layout.

JamesAPetts:
Viewport letter only if using multiple viewport layout.

Lettered, left to right, top to bottom: A, B, C, D.

Duplicate lettering not in place yet as still need confirmation from OHIF-164.


extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -225,33 +225,45 @@ function TrackedCornerstoneViewport({
...
-            patientName: PatientName ? OHIF.utils.formatPN(PatientName.Alphabetic) : '',
+            patientName: PatientName
+              ? OHIF.utils.formatPN(PatientName.Alphabetic)
+              : '',
             patientSex: PatientSex || '',
             patientAge: PatientAge || '',
             MRN: PatientID || '',
             thickness: `${SliceThickness}mm`,
-            spacing: PixelSpacing && PixelSpacing.length ? `${PixelSpacing[0].toFixed(2)}mm x ${PixelSpacing[1].toFixed(2)}mm` : '',
+            spacing:
```
dannyrb, line 261: While you're in here, care to fix up `thickness`?

image

dannyrb: It may be worth faking some large values to make sure that component uses `ellipsis` and "title" tags for large string values

JamesAPetts: This is a merge quirk and not part of this PR. I'll rebase the actual commit and merge.

PR #1823, opened 2020-06-26
fix: sort number of instances | format date in study panel

rodrigolabs:
**OHIF-155, OHIF-159**

## Description

- Should not be able to sort by number of instances
- Display Study Date in DD-Mon-YYYY format in study panel

### PR Checklist


platform/ui/src/utils/getFormattedDate.js
```diff
@@ -0,0 +1,12 @@
+import moment from 'moment';
```
rodrigolabs, line 1: Since `@ohif/ui` already have `momentjs` as a dependency and to avoid having the same dependency around other projects, I ended up creating a new `utils` that lives under the UI project. LMK what you guys think about it

extensions/default/src/Panels/PanelStudyBrowser.jsx
```diff
@@ -1,6 +1,8 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { StudyBrowser, useImageViewer } from '@ohif/ui';
+import { StudyBrowser, useImageViewer, utils } from '@ohif/ui';
```
igoroctaviano, line 3: Since we are using getFormattedDate inside the UI package and some extensions may not use ohif/ui, I think this util should not live there. Can we just:
```var todayDate = new Date().toISOString().slice(0,10);``` or maybe move this util to ohif/core/utils?

dannyrb: I agree that it shouldn't live in `@ohif/ui`.
`@ohif/core/utils` is better

I'd rather we not use `toISOString()` --> date-fns and moment are more robust, and JavaScript dates are notoriously awful.

rodrigolabs: Thanks guys! this util was moved to `@ohif/core`, still using moment.

PR #1827, opened 2020-06-29
OHIF-127-196-191: Hotkeys config and defaults

igoroctaviano:
Delivers:
- OHIF-127
- OHIF-196
- OHIF-191

### PR Checklist


platform/core/src/defaults/hotkeyBindings.js
```diff
@@ -0,0 +1,95 @@
...
+    label: 'Previous Series',
+    keys: ['pagedown'],
+  },
+  {
+    commandName: 'nextViewportDisplaySet',
+    label: 'Next Series',
+    keys: ['pageup'],
+  },
+  // ~ Cornerstone Tools
+  { commandName: 'setZoomTool', label: 'Zoom', keys: ['z'] },
```
dannyrb, line 48: Let's remove this one

platform/core/src/defaults/hotkeyBindings.js
```diff
@@ -0,0 +1,95 @@
...
+  {
+    commandName: 'windowLevelPreset1',
+    label: 'W/L Preset 1',
+    keys: ['1'],
+  },
+  {
+    commandName: 'windowLevelPreset2',
+    label: 'W/L Preset 2',
+    keys: ['2'],
+  },
```
dannyrb, line 59: These do not have default values/options. Can we hardcode options for the values we have set in the wwwc dropdown?

platform/core/src/defaults/hotkeyBindings.js
```diff
@@ -0,0 +1,95 @@
+export default [
```
dannyrb, line 1: Interesting. I guess these are all technically agnostic of any extension. Putting them in core as defaults works well. Good idea :+1:

platform/core/src/defaults/hotkeyBindings.js
```diff
@@ -0,0 +1,95 @@
...
+    commandName: 'windowLevelPreset8',
+    label: 'W/L Preset 8',
+    keys: ['8'],
+  },
+  {
+    commandName: 'windowLevelPreset9',
+    label: 'W/L Preset 9',
+    keys: ['9'],
+  },
+];
```
dannyrb, line 95: Can you poke on each of these? If it doesn't work, or doesn't have the prescribed outcome, let's create a summary/follow-up issue regarding which we should fix/remove/do later

platform/viewer/src/routes/Mode/Mode.jsx
```diff
@@ -83,6 +85,8 @@ export default function ModeRoute({
     route,
     servicesManager,
     extensionManager,
+    hotkeysManager,
+    isHotkeysLoaded
```
dannyrb, line 89: Is anything keying off of this? Shouldn't modes just expect hotkeys to be loaded?

platform/viewer/src/routes/Mode/Mode.jsx
```diff
@@ -15,7 +15,9 @@ export default function ModeRoute({
   dataSourceName,
   extensionManager,
   servicesManager,
+  hotkeysManager
 }) {
+  const isHotkeysLoaded = useHotkeys(hotkeysManager, mode.hotkeys);
```
dannyrb, line 20: Easier to just drop in `useEffect` that inits/clears? Return value probably not needed

platform/viewer/src/hooks/useHotkeys.js
```diff
@@ -0,0 +1,31 @@
...
+    console.debug('[hotkeys] Setting up hotkeys...');
+    manager.setDefaultHotKeys(defaultHotkeys || hotkeys);
+    manager.setHotkeys(hotkeys);
+    setIsLoaded(true);
+
+    return () => {
+      console.debug('[hotkeys] Removing hotkeys...');
+      manager.destroy();
+    };
+  }, [manager, hotkeys, defaultHotkeys]);
```
dannyrb, line 26: I don't know that we need react to monitor the manager or the mode config. If someone wants to change these, they would probably interact with the hotkeysManager directly? Or we'll have somewhere in the pipe where userPreferences influence this?

Then on destroy, this will clean up whatever the active/current values are in the manager, regardless of how they've changes or current state.

platform/viewer/src/routes/Mode/Mode.jsx
igoroctaviano: Totally fine removing the custom hook and using it directly =)

platform/viewer/src/routes/Mode/Mode.jsx
igoroctaviano: Nope, removing hook, thanks

platform/core/src/defaults/hotkeyBindings.js
igoroctaviano: @dannyrb since hotkeysmanager uses the commandName as the key for the map of bindings, we cant just use one command and then pass the options in each binding, maybe i should update the manager to use a different property as key? currently each commandname has its own values/options like windowlevel1, windowlevel2, etc..

dannyrb: It's okay if we have to tackle it that way for now. If there's an easy way to add the functionality from this stale/closed PR, we could also do that. It was drafted to address this problem, but was not merged:

https://github.com/OHIF/Viewers/pull/1456/files

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -39,36 +42,36 @@ export default [
...
-          commandName: 'windowLevelPreset4',
-          commandOptions: {},
+        4: {
+          commandName: 'setWindowLevel',
+          commandOptions: windowLevelPresets[4],
         },
-        'windowLevelPreset5': {
-          commandName: 'windowLevelPreset5',
-          commandOptions: {},
+        5: {
```
dannyrb, line 61: Does this work? I though objects couldn't have `number` keys?

igoroctaviano: It will always convert to a string, the code works

PR #1833, opened 2020-06-30
OHIF-197

JamesAPetts:
Create an SR displaySet which is appended to the sessions displaySets on SR export.
The SR is found and correctly indexed on both soft navigation (back to study list and into new mode instance with study), and hard navigation (refresh).


extensions/default/src/Panels/PanelStudyBrowser.jsx
```diff
@@ -28,6 +28,8 @@ function PanelStudyBrowser({
   const [displaySets, setDisplaySets] = useState([]);
   const [thumbnailImageSrcMap, setThumbnailImageSrcMap] = useState({});
 
+  console.log(DisplaySetService);
+
```
dannyrb, line 32:
```suggestion
```

extensions/default/src/Panels/PanelStudyBrowser.jsx
```diff
@@ -28,6 +28,7 @@ function PanelStudyBrowser({
   const [displaySets, setDisplaySets] = useState([]);
   const [thumbnailImageSrcMap, setThumbnailImageSrcMap] = useState({});
 
+  console.log(DisplaySetService);
```
dannyrb, line 31:
```suggestion
```

platform/core/src/DICOMSR/dataExchange.js
```diff
@@ -119,7 +80,11 @@ const generateReport = measurementData => {
  * that you wish to serialize.
  * @param {object} dataSource The dataSource that you wish to use to persist the data.
  */
-const storeMeasurements = async (measurementData, dataSource) => {
+const storeMeasurements = async (
+  measurementData,
+  dataSource,
+  displaySetService
```
dannyrb, line 86: This method feels weird to me. DICOMSR/dataExchange, but it's coordinating behavior between a `WebApiDataSource` and has a sideEffect on `displaySetService`?

Where else is this used? It's already async/promise. I think we should either return something more useful, await that result, and then do follow-up stuff: notification, update displaySetService, etc.

Or `displaySetService` parameter should be swapped out for an onSuccess callback that receives that naturalizedReport

platform/core/src/DataSources/IWebApiDataSource.js
```diff
@@ -59,6 +60,7 @@ function create({
     store: store || defaultStore,
     getImageIdsForDisplaySet,
     retrieveSeriesMetadata,
+    deleteStudyMetadataPromise,
```
dannyrb, line 63: This does need to be exposed :+1:
I wonder if this version has issues where navigating back to the list after loading a route still shows requests as in-flight.

platform/core/src/DICOMSR/dataExchange.js
JamesAPetts: I agree with the onSuccess, I guess you wouldn't always want to make displaySets locally 👍

JamesAPetts: The `deleteStudyMetadataPromise` should be part of the main function call though, otherwise you will get caching issues as you _will_ have new data on your server. Its also like this on main branch.

JamesAPetts: Updated.

platform/core/src/DataSources/IWebApiDataSource.js
JamesAPetts: Lets look into this at some point.

PR #1834, opened 2020-06-30
OHIF-198 - Jump to thumbnail when a new displaySet is added

JamesAPetts:
WIP Jump to thumbnail when SR saved.

on displaySet Creation:
- jumps to the right tab
- opens the study in the browser if not open.
- scrolls to the thumbnail element.


platform/ui/src/components/ThumbnailNoImage/ThumbnailNoImage.jsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classnames from 'classnames';
```
dannyrb, line 3: You ended up not using this

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.jsx
```diff
@@ -34,6 +34,7 @@ function PanelStudyBrowserTracking({
   const [studyDisplayList, setStudyDisplayList] = useState([]);
   const [displaySets, setDisplaySets] = useState([]);
   const [thumbnailImageSrcMap, setThumbnailImageSrcMap] = useState({});
+  const [jumpToDisplaySet, setJumpToDisplaySet] = useState(null);
```
dannyrb, line 37: I don't think we need this

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.jsx
```diff
@@ -138,16 +139,23 @@ function PanelStudyBrowserTracking({
...
         newDisplaySets.forEach(async dSet => {
+          const displaySetInstanceUID = dSet.displaySetInstanceUID;
+
           const newImageSrcEntry = {};
           const displaySet = DisplaySetService.getDisplaySetByUID(
-            dSet.displaySetInstanceUID
+            displaySetInstanceUID
           );
+
+          if (dSet.madeInClient) {
```
dannyrb, line 149: I'm not sure I want to expand `displaySet`'s props to support this specifically. If there are other benefits, we can talk about keeping it?

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.jsx
```diff
@@ -401,3 +453,24 @@ function _createStudyBrowserTabs(
 
   return tabs;
 }
+
+function _findTabAndStudyOfDisplaySet(displaySetInstanceUID, tabs) {
```
dannyrb, line 458: You have `StudyInstanceUID` at time of save. Depending on how we order thumbnails, the newly created SR will either be the first or last SR in the list of thumbnails.

I think we can key most of this off either the current subscription, or identify a new platform level event that makes sense for what's occurring, and then accomplish all of this. Unfortunately, for scrollIntoView, each thumbnail may need a `ref`

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.jsx
JamesAPetts: Changed, now on event.

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.jsx
JamesAPetts: Kept generic for now as discussed.

PR #1846, opened 2020-07-01
[OHIF-200] - Add callouts to all DICOM SR read only measurements.

JamesAPetts:
- For now the measurments show the first 2 derived measurements from the SR.
- We have a small lib to convert the DICOM CodeMeanings to renderables, far from a "complete" solution at this point.
- Render SCOORD POINTs as arrows to retain a consistent look-and-feel with the cornerstone annotations.


extensions/dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -26,12 +26,16 @@ const CodeNameCodeSequenceValues = {
   ImageLibraryGroup: '126200',
   TrackingUniqueIdentifier: '112040',
   TrackingIdentifier: '112039',
+  Finding: '121071',
```
dannyrb, line 29: I wish this were easier to read/parse

extensions/dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -26,12 +26,16 @@ const CodeNameCodeSequenceValues = {
...
 };
 
 const RELATIONSHIP_TYPE = {
   INFERRED_FROM: 'INFERRED FROM',
 };
 
+const CORNERSTONE_CODING_SCHEME_DESIGNATOR = 'CST4';
+const CORNERSTONE_FREETEXT_CODE_VALUE = 'CORNERSTONEFREETEXT';
```
dannyrb, line 37: Almost need this in `@ohif/core` or a separate library that manages/describes our custom definitions/magic

extensions/dicom-sr/src/getSopClassHandlerModule.js
JamesAPetts: The point here is I'm making a mapping for this.

We could make this a part of the naturalization process of dcmjs, but someone would have to put in the work of naturalizing all these.

extensions/dicom-sr/src/getSopClassHandlerModule.js
JamesAPetts: Yeah we need all this doc'd for `dcmjs`.

extensions/dicom-sr/src/tools/DICOMSRDisplayTool.js
```diff
@@ -1,4 +1,5 @@
 import { importInternal, getToolState, toolColors } from 'cornerstone-tools';
+import { pixelToCanvas } from 'cornerstone-core';
```
dannyrb, line 2: All of this makes sense. Makes me wonder if there's a nice way to add this to our monorepo that publishes this tool as a separate package. `@ohif/cstool-dicomsr`

JamesAPetts: I don't think the tool has much value without the viewport, as it needs something to set up what is tracked per element.

PR #1849, opened 2020-07-02
[OHIF-235] Lazy processing of DICOM SR for viewport.

JamesAPetts:
Defer processing of SR until it is loaded in the viewport, to decrease initial viewer load times if hundreds of SRs present.

### DIFF
--- a/extensions/dicom-sr/src/OHIFCornerstoneSRViewport.js
+++ b/extensions/dicom-sr/src/OHIFCornerstoneSRViewport.js
@@ -106,6 +106,14 @@ function OHIFCornerstoneSRViewport({
     setElement(targetElement);
   };
 
+  useEffect(() => {
+    debugger;
+    if (!displaySet.isLoaded) {
+      displaySet.load();
+      setIsHydrated(displaySet.isHydrated);
+    }
+  }, [displaySet]);
+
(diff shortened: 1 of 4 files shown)


extensions/dicom-sr/src/OHIFCornerstoneSRViewport.js
```diff
@@ -106,6 +106,14 @@ function OHIFCornerstoneSRViewport({
     setElement(targetElement);
   };
 
+  useEffect(() => {
+    debugger;
```
dannyrb, line 110:
```suggestion
```

extensions/dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -93,22 +89,36 @@ function _getDisplaySetsFromSeries(
...
     '4'
   );
 
-  if (isRehydratable(displaySet, mappings)) {
-    displaySet.isLocked = false;
-    displaySet.isHydrated = false;
-  } else {
-    displaySet.isLocked = true;
-  }
+  displaySet.isHydrated = false;
```
dannyrb, line 120: Hydrated check is interesting. If we cache displaySets, do we need to clear this or displaySets on mode exit?

PR #1854, opened 2020-07-02
[OHIF-248] - Update cornerstoneTools version to get handle rendering improvements.

JamesAPetts:


extensions/cornerstone/src/OHIFCornerstoneViewport.js
```diff
@@ -102,6 +102,10 @@ class OHIFCornerstoneViewport extends Component {
   }
 
   getViewportData = async displaySet => {
+    if (!displaySet) {
```
JamesAPetts, line 105: Add these lines in to prevent errors in the consoles when the displaySet had yet to be set.

extensions/cornerstone/src/init.js
```diff
@@ -144,8 +145,12 @@ export default function init({ servicesManager, configuration }) {
     ) {
       if (props.configuration) {
         parsedProps.configuration.drawHandlesOnHover = true;
+        parsedProps.configuration.hideHandlesIfMoving = true;
       } else {
-        parsedProps.configuration = { drawHandlesOnHover: true };
+        parsedProps.configuration = {
```
dannyrb, line 150: Kind of weird that we don't just do `Object.assign`

extensions/cornerstone/src/OHIFCornerstoneViewport.js
dannyrb: When would we pass the `OHIFCornerstoneViewport` a null/undefined displaySet? And why wouldn't this line return before this check?

https://github.com/OHIF/Viewers/blob/b748aaba1dcbe83cf04172489c2f8959dba3907d/extensions/cornerstone/src/OHIFCornerstoneViewport.js#L135

extensions/cornerstone/src/OHIFCornerstoneViewport.js
```diff
@@ -139,9 +143,11 @@ class OHIFCornerstoneViewport extends Component {
...
 
     this.getViewportData(displaySet).then(viewportData => {
-      this.setState({
-        viewportData,
-      });
+      if (viewportData) {
+        this.setState({
+          viewportData,
+        });
+      }
```
dannyrb, line 150: Same here. These feels like protecting against something that shouldn't happen? If it happens, that means we're hiding a mistake we made somewhere else? What am I missing 🤔

JamesAPetts: @igoroctaviano  realised It was being hit sometimes, so I put in these guards. We should check this more vigorously later.

Now you mention it, its strange as we aren't even using this viewport, we are using the measurement tracking ones. This did fix the error messages though, so strange its being hit?

extensions/cornerstone/src/OHIFCornerstoneViewport.js
JamesAPetts: See other comment.

extensions/cornerstone/src/OHIFCornerstoneViewport.js
JamesAPetts: Lets try and review this seperate issue some point today.

extensions/cornerstone/src/init.js
JamesAPetts: If it doesn't exist, nothing to assign to, you need to declare the object first.

PR #1859, opened 2020-07-02
OHIF-252: "Recent" and "All" tabs should be disabled if there are no studies on their category

rodrigolabs:
Jul-02-2020 20-57-50

### DIFF
--- a/platform/ui/src/components/Button/Button.jsx
+++ b/platform/ui/src/components/Button/Button.jsx
@@ -85,6 +85,8 @@ const Button = ({
   startIcon: startIconProp,
   endIcon: endIconProp,
   className,
+  onClick,
+  /** TODO: All possible props should be explicitly defined -- avoid spreading props  */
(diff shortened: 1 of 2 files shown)


dannyrb:
We may want a more obvious visual style that doesn't require _hover_ here, but I'll leave that to Dan. Making both _gray_ may not convey that they're disabled without a reference state? I've also never come across a button bar w/ disabled options.

rodrigolabs:
> We may want a more obvious visual style that doesn't require _hover_ here, but I'll leave that to Dan. Making both _gray_ may not convey that they're disabled without a reference state? I've also never come across a button bar w/ disabled options.

Yes, totally agree! As we didn't have specification on it, and this tab originally has no hover state (it maybe should have too), I avoided spending time doing that. We would probably need to do a minor refactor in the `Button` component to accomplish that. -- Thanks! :)

PR #1862, opened 2020-07-03
[OHIF-261 & OHIF-262]

JamesAPetts:
- Add lifecycle hooks.
- Remove all measurements on mode switch.
- cs extension listens to this and cleans its toolState.
- SR’s hydratable status is reset.
- Viewport grid reset so we don’t get flash of other displaySet before its changed to new displaySet data.


extensions/cornerstone/src/init.js
```diff
@@ -324,6 +326,22 @@ const _connectToolsToMeasurementService = measurementService => {
       }
     }
 
+    const { MEASUREMENT_REMOVED } = measurementService.EVENTS;
+
+    measurementService.subscribe(MEASUREMENT_REMOVED, ({ id }) => {
+      //
```
dannyrb, line 332:
```suggestion
```

extensions/cornerstone/src/init.js
```diff
@@ -325,6 +327,21 @@ const _connectToolsToMeasurementService = measurementService => {
       }
     }
 
+    const { MEASUREMENT_REMOVED } = measurementService.EVENTS;
+
+    measurementService.subscribe(MEASUREMENT_REMOVED, ({ id }) => {
+
+      const toolState = cornerstoneTools.globalImageIdSpecificToolStateManager.saveToolState();
+    });
```
dannyrb, line 335: ☝️ Was ist das?

PR #1873, opened 2020-07-06
fix: OHIF-260 - "Clear Filters" button should not be visible on first load, when no filters are set

rodrigolabs:
- Fixes the "Clear Filters" button on first load.
- Fixes the pagination select value when clearing filters (clear was not reflecting).

### DIFF
--- a/platform/ui/src/components/StudyListPagination/StudyListPagination.js
+++ b/platform/ui/src/components/StudyListPagination/StudyListPagination.js
@@ -20,7 +20,7 @@ const StudyListPagination = ({
           <div className="flex items-center">
             <div className="relative mr-3">
               <select
-                defaultValue={perPage}
+                value={perPage}
                 className="block appearance-none w-full bg-transparent border border-common-active text-white text-base px-2 pr-4 rounded leading-tight focus:outline-none h-8"
                 onChange={e => onChangePerPage(e.target.value)}
                 onBlur={() => {}}
--- a/platform/viewer/src/routes/WorkList/WorkList.jsx
+++ b/platform/viewer/src/routes/WorkList/WorkList.jsx
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
 import moment from 'moment';
 import qs from 'query-string';
-import { isEqual } from 'lodash';
+import isEqual from 'lodash.isequal';
 //
 import filtersMeta from './filtersMeta.js';
 import { useAppConfig } from '@state';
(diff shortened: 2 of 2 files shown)


platform/viewer/src/routes/WorkList/WorkList.jsx
```diff
@@ -315,6 +317,7 @@ function WorkList({ history, data: studies, isLoadingData, dataSource }) {
                   disabled={false}
                   endIcon={<Icon name="launch-arrow" />} // launch-arrow | launch-info
                   className={classnames('font-bold', { 'ml-2': !isFirst })}
+                  onClick={() => {}}
```
rodrigolabs, line 320: (fix console error) -- Adding an empty `onClick` because `Button` expects this as a required prop.

platform/viewer/src/routes/WorkList/WorkList.jsx
```diff
@@ -1,9 +1,10 @@
-import React, { Suspense, useState, useEffect } from 'react';
+import React, { useState, useEffect } from 'react';
```
rodrigolabs, line 1: Removed unused `Suspense`

platform/viewer/src/routes/WorkList/WorkList.jsx
```diff
@@ -1,9 +1,10 @@
...
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
 import moment from 'moment';
 import qs from 'query-string';
+import { isEqual } from 'lodash';
```
dannyrb, line 7: `loadash` is a devDependency, not a `dependency`

platform/viewer/src/routes/WorkList/WorkList.jsx
```diff
@@ -181,7 +183,7 @@ function WorkList({ history, data: studies, isLoadingData, dataSource }) {
 
   const isFiltering = (filterValues, defaultFilterValues) => {
     return Object.keys(defaultFilterValues).some(name => {
-      return filterValues[name] !== defaultFilterValues[name];
+      return !isEqual(filterValues[name], defaultFilterValues[name]);
```
dannyrb, line 186: We know our filter values. I think we can do strict equality comparison for everything except the date range object and the modalities array? For those, we can use the key value to indicated we need to compare differently?

platform/ui/src/components/StudyListPagination/StudyListPagination.js
```diff
@@ -20,7 +20,7 @@ const StudyListPagination = ({
           <div className="flex items-center">
             <div className="relative mr-3">
               <select
-                defaultValue={perPage}
+                value={perPage}
```
dannyrb, line 23: Just switches from default to controlled?

platform/viewer/src/routes/WorkList/WorkList.jsx
swederik: I just switch to using `!isEqual(filterValues, defaultFilterValues)` since I don't see any reason why this isn't easier.

platform/viewer/src/routes/WorkList/WorkList.jsx
swederik: Switched to lodash.isequal which is in dependencies

platform/ui/src/components/StudyListPagination/StudyListPagination.js
swederik: yes

PR #1874, opened 2020-07-06
OHIF-287: Clear all measurements should not clear SR Viewport measurements

igoroctaviano:
### PR Checklist

### DIFF
--- a/extensions/cornerstone/src/init.js
+++ b/extensions/cornerstone/src/init.js
@@ -5,7 +5,7 @@ import cornerstone from 'cornerstone-core';
 import csTools from 'cornerstone-tools';
 import merge from 'lodash.merge';
 import initCornerstoneTools from './initCornerstoneTools.js';
-import cornerstoneTools from 'cornerstone-tools';
+import cornerstoneTools, { globalImageIdSpecificToolStateManager } from 'cornerstone-tools';
 import initWADOImageLoader from './initWADOImageLoader.js';
 import measurementServiceMappingsFactory from './utils/measurementServiceMappings/measurementServiceMappingsFactory';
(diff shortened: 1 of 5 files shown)


igoroctaviano:
@dannyrb maybe theres a much better way to do it?

extensions/cornerstone/src/init.js
```diff
@@ -360,7 +352,18 @@ const _connectMeasurementServiceToTools = (
   const sourceId = measurementSource.id;
 
   measurementService.subscribe(MEASUREMENTS_CLEARED, () => {
-    cornerstoneTools.globalImageIdSpecificToolStateManager.restoreToolState({});
+    const DICOM_SR_DISPLAY_TOOL = 'DICOMSRDisplayTool';
+    const toolState = globalImageIdSpecificToolStateManager.saveToolState();
```
JamesAPetts, line 356: We need to do this a different way, the cornerstone extension should have no knowledge of the SR extension, and this should be true for extensions in general. 

The way we need to do this is by only removing the tools associated with this extension.
So that would be measurements where the source is cornerstoneTools?

@dannyrb I guess the way to do this is actually to pass the list of measurements directly from the MeasurementService on the MEASUREMENTS_CLEARED event, allow the extension to itterate and remove the _measruements_, and then have the MeasurementService clean up its own measurements after?

dannyrb: @JamesAPetts that sounds like the right approach. Trinity has since updated toolbar requirements to remove the "Clear Measurements" button. So let's use this PR to either remove the "Clear all" approach, or add this comment and comment it out or disable it.

PR #1875, opened 2020-07-06
fix: OHIF-215 - Patient information should truncate w/ ellipsis instead of expanding box

rodrigolabs:
## Description

Information text will be truncated and the max-width will depend on the num of viewport cols. When the mouse hovers over the text, it will display its complete information.

image

image

### DIFF
--- a/platform/ui/src/components/ViewportActionBar/ViewportActionBar.jsx
+++ b/platform/ui/src/components/ViewportActionBar/ViewportActionBar.jsx
@@ -6,7 +6,7 @@ import useOnClickOutside from '../../utils/useOnClickOutside';
 
 const classes = {
   infoHeader: 'text-base text-primary-light',
-  infoText: 'text-base text-white',
+  infoText: 'text-base text-white max-w-24 truncate',
   firstRow: 'flex flex-col',
   row: 'flex flex-col ml-4',
 };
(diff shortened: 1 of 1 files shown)


platform/ui/src/components/ViewportActionBar/ViewportActionBar.jsx
```diff
@@ -224,23 +219,36 @@ ViewportActionBar.propTypes = {
...
   spacing,
   scanner,
+  numCols,
   isOpen,
 }) {
+  const classes = {
+    infoHeader: 'text-base text-primary-light',
+    infoText: classnames(
+      'text-base text-white truncate',
+      numCols === 1 ? 'max-w-48' : 'max-w-16'
```
dannyrb, line 246: I _really_ don't want this thing's width to be based on the reactive viewportGridService's number of columns state.
This should either be a fixed width regardless of the number of columns, or we should find a way for it to determine it's width relative to the viewportPane it's contained inside.

platform/ui/src/components/ViewportActionBar/ViewportActionBar.jsx
```diff
@@ -224,23 +219,36 @@ ViewportActionBar.propTypes = {
...
+  numCols: 1,
 };
 
 function PatientInfo({
   patientName,
   patientSex,
   patientAge,
   MRN,
-  thickness,
+  thickness = 'N/A',
```
dannyrb, line 236: Was this called out as a requirement? If this is a prop, can we set the default value in `defaults` instead of here?

swederik:
Just made it fixed width and removed all the weird numCols stuff.

PR #1876, opened 2020-07-06
fix: OHIF-255 Fixes Thumbnail height for active item

rodrigolabs:
## Preview

Jul-06-2020 18-40-25

### DIFF
--- a/platform/ui/src/components/Thumbnail/Thumbnail.jsx
+++ b/platform/ui/src/components/Thumbnail/Thumbnail.jsx
@@ -50,11 +50,14 @@ const Thumbnail = ({
       <div ref={drag}>
         <div
           className={classnames(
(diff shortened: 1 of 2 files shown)


platform/ui/src/components/Thumbnail/Thumbnail.jsx
```diff
@@ -50,11 +50,14 @@ const Thumbnail = ({
...
         <div
           className={classnames(
-            'flex flex-1 items-center justify-center rounded-md bg-black text-base text-white overflow-hidden mb-2 min-h-32',
+            'flex flex-1 items-center justify-center rounded-md bg-black text-base text-white overflow-hidden min-h-32',
             isActive
               ? 'border-2 border-primary-light'
               : 'border border-secondary-light group-focus:border-blue-300 hover:border-blue-300'
           )}
+          style={{
+            margin: isActive ? '0' : '1px',
```
rodrigolabs, line 59: The same little trick we did in `ViewportPane`

Issue #1797, opened 2020-06-19
Looping error related to SEG loading


fedorov:
### Describe the Bug

while trying to load this study: https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc-tcia/dicomStores/idc-tcia/study/1.3.6.1.4.1.9328.50.1.216116555221814778114703363464001196508, I get looping error below:

![2020-06-19_16-31-31 (1)](https://user-images.githubusercontent.com/313942/85177911-2322dc80-b24b-11ea-8c83-685315e9d3e0.gif)

This may well be a problem with the dataset, but probably this is not the expected way to handle it in OHIF?

### What steps can we follow to reproduce the bug?

Open URL above, or download the study from TCIA as shown below:

![image](https://user-images.githubusercontent.com/313942/85177974-4baad680-b24b-11ea-8dec-e17d8d0ac1e1.png)

![image](https://user-images.githubusercontent.com/313942/85178103-aa705000-b24b-11ea-8adc-c233add93b9d.png)

JamesAPetts:
For what reason was this moved back to in progress? Does the issue still persist after  #1877?

fedorov:
@JamesAPetts no, the issue does not persist. I moved it back into "in progress", since it was not closed, and I was not sure about the true status. I will close it now.

JamesAPetts:
Awesome, thanks for the quick response.

Issue #1856, opened 2020-07-02
RTSTRUCT load: "RangeError: Offset is outside the bounds of the DataView"


fedorov:
### Describe the Bug

There is looping error "RangeError: Offset is outside the bounds of the DataView" lading RTSTRUCT, see here:

https://dev-viewer.canceridc.dev/projects/idc-dev-etl/locations/us/datasets/pre-mvp-temp/dicomStores/cross-collection-temp/study/1.3.6.1.4.1.14519.5.2.1.1706.4009.421312468402475769388366275116

This is subject `TCGA-CV-6433` from the `TCGA-HNSC` collection.

fedorov:
It is interesting that this happens for some other but not all of the subjects from that collection.

It works fine for this one: https://dev-viewer.canceridc.dev/projects/idc-dev-etl/locations/us/datasets/pre-mvp-temp/dicomStores/cross-collection-temp/study/1.3.6.1.4.1.14519.5.2.1.1706.4009.279236404491517275639164431409

fedorov:
Also note when this happens, it is impossible to download the study (or maybe it is a separate regression).

@pieper any ideas?

![image](https://user-images.githubusercontent.com/313942/86605142-4dc3a380-bf74-11ea-8a73-db64ef3c1511.png)

fedorov:
Or is this due to the proxy?

pieper:
I haven't looked at this in detail.  Do we need it fixed before the Wednesday meeting?

fedorov:
Considering the earlier experience, I don't think this is feasible or desirable (since other things can get broken, and I will not know what to expect to work during the demo).

I am just testing things as I prepare, and populate issues that I think should be fixed, if possible, sometime after the Wednesday meeting.

fedorov:
Adding @wlongabaugh to the thread, since it is not clear if the issue is actually relevant to the download functionality.

wlongabaugh:
[...]

PR #1877, opened 2020-07-07
fix: 🐛 Fix RT Panel hide/show and Fix looping load errors

### LINKED ISSUE #1768: Add reset filter option to RTstruct and ANON functionality
### Describe the Bug

When working with the OHIF dev tier, the RTstruct functionality doesn't seem to reset the filter selections once the top-level icon has been selected more than once.

### What steps can we follow to reproduce the bug?

1. load the OHIF viewer from the OHIF dev tier
2. open RTstruct panel
2. select a filter option e.g., CouchSurface
3. select the hide-all option
4. select the display-all filter
5. the filter option is still hidden

The filter option selected in step 3 should be reset to displayed in step 6.

A similar issue can be found within the ANON filter options under the RTstruct functionality. 

@JamesAPetts @swederik

JamesAPetts:
Closes #1768 and #1797


fedorov:
@JamesAPetts is this also fixing #1856?

PR #1881, opened 2020-07-07
fix: OHIF-219 - Viewport Labels should be sequential

rodrigolabs:
**Part 1**

Viewports with matching displaysets currently use the same letter label:

- Viewport 1: DisplaySet 1 --> A
- Viewport 2: DisplaySet 1 --> A
- Viewport 3: DisplaySet 2 --> B

Regardless of the displayset, lettering should be sequential:

- Viewport 1: DisplaySet 1 --> A
- Viewport 2: DisplaySet 1 --> B
- Viewport 3: DisplaySet 2 --> C

**Part 2**

Each thumbnail should list each viewport label they are displayed in. For example, if DisplaySet 1 is the displayset in all 9 viewports in a 3x3 grid, it's label should be: A, B, C, D, E, F, G, H, I

## Previews

A provisional design was created for when the series is being shown in more than 3 viewports.

Series in 3 viewports:

image

Series in more than 3 viewports (icon with a tooltip):

image

Tracked Series in more than 3 viewports:

image

Viewport:

image

### DIFF
--- a/extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.jsx
+++ b/extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.jsx
@@ -23,7 +23,7 @@ function PanelStudyBrowserTracking({
   // Tabs --> Studies --> DisplaySets --> Thumbnails
   const [{ StudyInstanceUIDs }, dispatchImageViewer] = useImageViewer();
   const [
-    { activeViewportIndex, viewports },
+    { activeViewportIndex, viewports, numCols, numRows },
     viewportGridService,
   ] = useViewportGrid();
   const [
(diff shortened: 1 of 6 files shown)


JamesAPetts:
Tested, looks good to me, thanks.

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.jsx
```diff
@@ -355,21 +359,22 @@ function _mapDisplaySets(
...
     const componentType = _getComponentType(ds.Modality);
-    const firstViewportIndexWithMatchingDisplaySetUid = viewports.findIndex(
-      vp => vp.displaySetInstanceUID === ds.displaySetInstanceUID
-    );
-
-    const viewportIdentificator =
-      viewports.length > 1
-        ? _viewportLabels[firstViewportIndexWithMatchingDisplaySetUid]
-        : '';
+    const viewportIdentificator = isSingleViewport
```
swederik, line 370: Can we stop using viewportIdentificator and use viewportIdentifier instead? Identificator is not a word

PR #1883, opened 2020-07-08
[OHIF-295] - Esc button should stop the creation of a measurement or annotation

JamesAPetts:
Does what it says on the tin. Prevents the measurement being added to the measurement service and cancels it.

### DIFF
--- a/platform/core/src/defaults/hotkeyBindings.js
+++ b/platform/core/src/defaults/hotkeyBindings.js
@@ -1,4 +1,4 @@
-import windowLevelPresets from "./windowLevelPresets";
+import windowLevelPresets from './windowLevelPresets';
 
 /*
  * Supported Keys: https://craig.is/killing/mice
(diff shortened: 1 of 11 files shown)


platform/core/src/defaults/hotkeyBindings.js
```diff
@@ -19,6 +19,11 @@ export default [
   { commandName: 'rotateViewportCW', label: 'Rotate Right', keys: ['r'] },
   { commandName: 'rotateViewportCCW', label: 'Rotate Left', keys: ['l'] },
   { commandName: 'invertViewport', label: 'Invert', keys: ['i'] },
+  {
+    commandName: 'cancelMeasurement',
+    label: 'Cancel Cornerstone Measurement',
```
dannyrb, line 24: This is displayed in the hotkeys panel. Might be odd to specifically name `cornerstone`. The idea is that the command name is library agnostic, and each viewport can optionally implement the command

PR #1886, opened 2020-07-08
fix: OHIF-307 - Click outside patient info should close tooltip

rodrigolabs:
## Description

Any interaction outside of an expanded patient info tooltip should close the Tooltip.

## Preview

Jul-08-2020 13-47-08


platform/ui/src/utils/clickOutsideHandler.js
```diff
@@ -1,7 +1,8 @@
-export default element => {
+export default (element, callback = () => {}) => {
   const handleClickOutside = event => {
     if (element.current && !element.current.contains(event.target)) {
       element.current.blur();
+      callback();
```
dannyrb, line 5: `onClickOutside` and assume it's set?
Or create a separate `blueHandlerListener` if we're trying to use this for both things?

platform/ui/src/components/Thumbnail/Thumbnail.jsx
```diff
@@ -35,7 +35,7 @@ const Thumbnail = ({
 
   return (
     <div
-      onFocus={() => blurHandlerListener(thumbnailElement)}
+      onFocus={() => clickOutsideHandler(thumbnailElement)}
```
dannyrb, line 38: Ah, this is why. Yeah. Just duplicate the method, and give it a different name.
I'd rather it be explicit what the method does.

platform/ui/src/components/ViewportActionBar/ViewportActionBar.jsx
```diff
@@ -188,6 +198,7 @@ const ViewportActionBar = ({
       )}
       <div className="flex ml-4 mr-2" onClick={onPatientInfoClick}>
         <PatientInfo
+          showPatientInfoRef={showPatientInfoRef}
```
dannyrb, line 201: It's a little weird that we have to prop-drill this ref. Can we wrap this component in a div or something else that we set `ref` on so we don't have to drill?

rodrigolabs: Yes, I agree! I've removed the prop from tooltip and wrapped the component into a div 👍

platform/ui/src/utils/clickOutsideHandler.js
rodrigolabs: They were split into two separate methods 👍  Even though they seem to do the same thing, one is related to blur, and the other to click outside an element. Thanks!

platform/ui/src/components/Thumbnail/Thumbnail.jsx
```diff
@@ -31,12 +30,8 @@ const Thumbnail = ({
     },
   });
 
-  const thumbnailElement = useRef(null);
-
   return (
     <div
-      onFocus={() => blurHandlerListener(thumbnailElement)}
```
rodrigolabs, line 38: Not needed anymore. `blurHandlerListener` was removed too.

PR #1888, opened 2020-07-09
[OHIF-278] Jumping to a measurement should set it to "active" in cornerstone-tools, highlighting it as green

JamesAPetts:
- Jumping to a measurement sets it active
- Other annotations are set inactive, so that you don't get multiple active measurements on a single slice by clicking on multiple measurements in the panel.


extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/setCornerstoneMeasurementActive.js
```diff
@@ -0,0 +1,33 @@
...
+const { globalImageIdSpecificToolStateManager } = cornerstoneTools;
+
+export default function setMeasurementActive(measurement) {
+  const { id } = measurement;
+
+  const toolState = globalImageIdSpecificToolStateManager.saveToolState();
+
+  Object.keys(toolState).forEach(imageId => {
+    const imageIdSpecificToolState = toolState[imageId];
+    debugger;
```
swederik, line 13: you should remove this

extensions/measurement-tracking/src/_shared/setCornerstoneMeasurementActive.js
```diff
@@ -0,0 +1,32 @@
+import cornerstone from 'cornerstone-core';
```
dannyrb, line 1: This would be a weird command, but it could be a command in the Cornerstone extension.

Could also add it to `lib` exports in cornerstone-tools, even though it's a bit of workaround

PR #1889, opened 2020-07-09
[OHIF 284] - JumpToMeasurement + prevent layout change on scroll Prevent layout change on scroll

JamesAPetts:


extensions/cornerstone/src/init.js
```diff
@@ -220,45 +224,45 @@ export default function init({ servicesManager, configuration }) {
   csTools.setToolEnabled('Overlay', {});
 }
 
-const _initMeasurementService = measurementService => {
+const _initMeasurementService = (MeasurementService, DisplaySetService) => {
```
dannyrb, line 227: Only weird bit is that this doesn't have it's own lexical scope. Does it conflict with the `MeasurementService` we're pulling out of the `servicesManager.services`?

extensions/cornerstone/src/init.js
```diff
@@ -269,12 +273,16 @@ const _initMeasurementService = measurementService => {
   return csToolsVer4MeasurementSource;
 };
 
-const _connectToolsToMeasurementService = measurementService => {
+const _connectToolsToMeasurementService = (
+  MeasurementService,
```
dannyrb, line 277: I assume the same thing here. I definitely prefer this casing. If there's no conflict let's keep it.
(Can't see the full file in the `diff` view, so likely no conflict)

platform/ui/src/contextProviders/ViewportGridProvider.jsx
```diff
@@ -27,14 +27,10 @@ export function ViewportGridProvider({ children, service }) {
         return { ...state, ...{ activeViewportIndex: action.payload } };
       }
       case 'SET_DISPLAYSET_FOR_VIEWPORT': {
-        const {
-          viewportIndex,
-          displaySetInstanceUID,
-          imageIndex,
-        } = action.payload;
+        const { viewportIndex, displaySetInstanceUID } = action.payload;
```
dannyrb, line 30: I believe this becomes an API change that needs to be reflected in the `ViewportGridService` in @ohif/core

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/index.js
```diff
@@ -162,27 +162,7 @@ function PanelMeasurementTableTracking({ servicesManager, extensionManager }) {
...
-    const imageIndex = displaySet.images
-      .map(i => i.SOPInstanceUID)
-      .indexOf(SOPInstanceUID);
-
-    viewportGridService.setDisplaysetForViewport({
-      viewportIndex: viewportGrid.activeViewportIndex,
-      displaySetInstanceUID: displaySet.displaySetInstanceUID,
-      imageIndex,
-    });
+    MeasurementService.jumpToMeasurement(viewportGrid.activeViewportIndex, id);
```
dannyrb, line 165: This does look fancier.

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -498,6 +500,40 @@ class MeasurementService {
...
+
+    const hasListeners = Object.keys(this.listeners).length > 0;
+    const hasCallbacks = Array.isArray(this.listeners[eventName]);
+
+    if (hasListeners && hasCallbacks) {
+      this.listeners[eventName].forEach(listener => {
+        listener.callback({ viewportIndex, measurement });
+      });
+    }
+  }
```
dannyrb, line 523: Can't we re-use the pub/sub "broadcast" here? (instead of iterating listeners)

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -68,6 +70,64 @@ function TrackedCornerstoneViewport({
     };
   }, []);
 
+  useEffect(() => {
+    const { unsubscribe } = MeasurementService.subscribe(
+      MeasurementService.EVENTS.JUMP_TO_MEASUREMENT,
+      ({ viewportIndex: jumpToMeasurementViewportIndex, measurement }) => {
+        // check if the correct viewport index.
+        // TODO -> Jump to measurement implementation.
```
dannyrb, line 78: I assume this TODO stub is done?

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -68,6 +70,64 @@ function TrackedCornerstoneViewport({
...
+
+        if (viewportIndex !== jumpToMeasurementViewportIndex) {
+          // Event for a different viewport.
+          return;
+        }
+
+        if (
+          measurement.displaySetInstanceUID !== displaySet.displaySetInstanceUID
```
dannyrb, line 86: Wouldn't this update the set displaySet? Or are we having the ViewportGridService reconcile this bit, and then on `create/propChange` we check the jump cache to see if we have one pending?

If it's the latter, where's that jump cache check?

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -68,6 +70,64 @@ function TrackedCornerstoneViewport({
...
+          MeasurementService,
+          DisplaySetService
+        );
+      }
+    );
+
+    // Check if there is a queued jumpToMeasurement event
+    const measurementIdToJumpTo = MeasurementService.getJumpToMeasurement(
+      viewportIndex
+    );
```
dannyrb, line 105: Ah. Here is the check, and it's in the useEffect w/ `displaySet` as a change check. That makes sense. Can we pull this out to make it super easy to read?

```js
useEffect(() => {
  const unsubcribeFromJumpToMeasurementEvents = subscribeToJumpToMeasurementEvents();
  
  checkForCachedJumpToMeasurementEvents();

 return () => {
   unsubcribeFromJumpToMeasurementEvents();
 }
}, [element, displaySet]);
```

dannyrb: When component's start to get large, it's a PITA to understand what's going on (in my opinion). My hope is, that at  glance, I can see the intent, and then dig into behavior as needed.

The actual logic can live further down in this file, or in separate files that are imported (which also makes them easier to test, if we ever go that route.

platform/viewer/src/components/ViewportGrid.jsx
```diff
@@ -43,30 +48,56 @@ function ViewerViewportGrid(props) {
...
-    if (!displaySets || !displaySets.length) {
-      const displaySet = DisplaySetService.activeDisplaySets[0];
-      displaySetInstanceUID = displaySet.displaySetInstanceUID;
-    } else {
-      const displaySet = displaySets[0];
-      displaySetInstanceUID = displaySet.displaySetInstanceUID;
-    }
+  useEffect(() => {
+    const { unsubscribe } = MeasurementService.subscribe(
+      MeasurementService.EVENTS.JUMP_TO_MEASUREMENT,
```
dannyrb, line 53: I'm torn between this happening:

1. In the `ViewportGrid` component
2. In the `ViewportGridService`
3. In the `Viewport` extension module component

I'm leaning toward number 3, then mixed on 1 & 2.

For 3, we already have the subscribe, and the logic's in one place, and the viewport could handle any wonkiness if it needs alternative behavior.

For 1, it's another responsibility the Viewer project has, when we're trying to make it _as minimal as possible_ as it becomes part of the un-reusable glue if you want to provide an alternative `Viewer` project.

For 2, it's more reusable for those that want to create their own `Viewer` project, but creates a more direct service to service dep.

extensions/dicom-sr/src/utils/getToolStateToCornerstoneMeasurementSchema.js
```diff
@@ -174,6 +245,15 @@ function ArrowAnnotate(measurementData, imageId, _getValueTypeFromToolType) {
...
 
+  const displaySets = DisplaySetService.getDisplaySetsForSeries(
+    SeriesInstanceUID
+  );
+  const displaySet = displaySets.find(ds => {
+    return (
+      ds.images && ds.images.some(i => i.SOPInstanceUID === SOPInstanceUID)
+    );
+  });
+
```
dannyrb, line 256: This bit appears to be reused and may be worth pulling out

extensions/cornerstone/src/utils/measurementServiceMappings/Length.js
```diff
@@ -57,12 +61,22 @@ const Length = {
...
     } = getSOPInstanceAttributes(element);
 
+    const displaySets = DisplaySetService.getDisplaySetsForSeries(
+      SeriesInstanceUID
+    );
+    const displaySet = displaySets.find(ds => {
+      return (
+        ds.images && ds.images.some(i => i.SOPInstanceUID === SOPInstanceUID)
+      );
+    });
```
dannyrb, line 71: We sure do this a lot

dannyrb: `DisplaySetService.getDisplaySetForImage(SOPInstanceUID, SeriesInstanceUID?)`

Where second option, if provided, is more efficient?

extensions/cornerstone/src/utils/measurementServiceMappings/EllipticalRoi.js
```diff
@@ -58,6 +71,7 @@ const EllipticalRoi = {
       FrameOfReferenceUID,
       referenceSeriesUID: SeriesInstanceUID,
       referenceStudyUID: StudyInstanceUID,
+      displaySetInstanceUID: displaySet.displaySetInstanceUID,
```
dannyrb, line 74: Right. Because MeasurementService now tracks this for the sanity of all involved.
While it's serializable, it'd be nonsense to persist this anywhere and try to re-load it.

extensions/cornerstone/src/utils/measurementServiceMappings/ArrowAnnotate.js
```diff
@@ -26,6 +30,15 @@ const ArrowAnnotate = {
...
     } = getSOPInstanceAttributes(element);
 
+    const displaySets = DisplaySetService.getDisplaySetsForSeries(
+      SeriesInstanceUID
+    );
+    const displaySet = displaySets.find(ds => {
+      return (
+        ds.images && ds.images.some(i => i.SOPInstanceUID === SOPInstanceUID)
+      );
+    });
```
dannyrb, line 40: Yeah, this is everywhere, haha

extensions/cornerstone/src/init.js
JamesAPetts: Yeah I changed it throughout in this file, I'm just trying to slowly bring consistency to the app where I see differences.

extensions/cornerstone/src/init.js
JamesAPetts: It doesn't conflict, this is a private function outside of the scope of the main init call.

platform/ui/src/contextProviders/ViewportGridProvider.jsx
JamesAPetts: Yes, that is the point of the PR, we removed imageIndex throughout, and it affects the `ViewportGridProvider` and thus the `ViewportGridService`, unless I'm missing something?

platform/core/src/services/MeasurementService/MeasurementService.js
JamesAPetts: The pub/sub broadcast broadcasts the same thing for every other event. I'd be happy to factor it out into a different function, but it can't be the same one.

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
JamesAPetts: Yup, good catch.

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
JamesAPetts: The `ViewportGrid` calls the `ViewportGridService` and changes the layout.
In the meantime if we have a different viewport that can't process the jumpToMeasurement event, it exits early here. The viewport that is replaced later will then consume it.

platform/viewer/src/components/ViewportGrid.jsx
JamesAPetts: I'm against 2, as that links the two services together expliclty.

3 we have to write for _every_ viewport, and it will be the same behavior. The Measurement references a displaySet, so if that displaySet is not present, it will always be changed/

Hence why I went with 1, happy to discuss.

extensions/dicom-sr/src/utils/getToolStateToCornerstoneMeasurementSchema.js
JamesAPetts: Done in next commit.

extensions/cornerstone/src/utils/measurementServiceMappings/EllipticalRoi.js
JamesAPetts: It is not. Its for session use only. New displaySetInstanceUIDs are created on refresh anyway.

platform/ui/src/contextProviders/ViewportGridProvider.jsx
dannyrb: `platform/core/src/srvices/ViewportGridService/ViewportGridService.js`
--> `_setDisplaysetForViewport`

Defines the interface for the service. I didn't see it updated in this file. It indicates `viewportIndex` as an expected property for the method.

platform/viewer/src/components/ViewportGrid.jsx
dannyrb: I think I'm fine with it. I'm just trying to figure out _what_ things like this are, and how we document it as a requirement of the "glue" project; something that's not solved by the reusable part of the platform.

platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -70,6 +71,7 @@ class MeasurementService {
     this.mappings = {};
     this.measurements = {};
     this.listeners = {};
+    this._jumpToMeasurmentCache = {};
```
swederik, line 74: That is not how you spell Measurement

platform/core/src/DICOMSR/dataExchange.js
```diff
@@ -71,6 +71,8 @@ const generateReport = (measurementData, additionalFindingTypes) => {
     additionalFindingTypes
   );
 
+  debugger;
```
swederik, line 74: Missed a debugger

PR #1894, opened 2020-07-10
feat: OHIF-331 - About Modal

rodrigolabs:
image


platform/viewer/src/components/PreferencesDropdown/AboutModal.jsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
```
dannyrb, line 1: The goal of the docz project is to be able to preview all UI without  having to touch the `viewer` project. We should have created the smaller reusable components in here, and then assembled to create an example in docz.

No need to do it now, but something to try and enforce moving forward.

PR #1895, opened 2020-07-13
OHIF-332: Users should be able to see all available hotkeys and language settings in one place

igoroctaviano:
image
image

This design is a reference design, and not necessarily indicative of which hotkeys should be set, what their labels should be, or what keys they should be bound to. 

The modal's title should be "User Preferences"

Should support a "disabled" state that:

Disables all "shortcut fields

Disables the "Save" and "Restore Defaults" buttons

The modal should consist of two sections

General

Contains a dropdown to change the active i18n language for the application

Hotkeys

Contains a two column list of functions and shortcuts

Function names are right aligned

"Setting" a shotcut key combination should reuse the logic from the previous major version of OHIF

Function display names and default shortcut bindings should use the mode configuration's values

The mode's hotkey configuration may require updates

Action buttons

Restore defaults should set all "shortcut" fields to the mode's default values

Cancel (and the model's X to dismiss) should clear the form's temporary values and cause no changes

Save should apply the form's values as the new bindings/settings

### PR Checklist


igoroctaviano:
> * Window level presets aren't in the designs, the designs are on the MGH Jira, could you update your PR to match these please?
> * The "Function" header should be right aligned, it is currently center aligned. Check the designs.
> * Note these new shortcut requirements from OHIF-62:
>   
>   * Home: Skip to first image of series in active viewport
>   * End: Skip to last image of series in active viewport
> 
> If you turn off "disabled" to test:
> 
> * Editing a value correctly shows a message that there is a clash, but you can still save it, saving the settings should be disabled in this case. Currently you can save it, and then both say the hotkey when you re-open the menu, but only the most recently changed one works.
> * The preferences don't persist when you switch from one study to another by navigating the study list. Even if we don't have persistence for now, the settings should persist within a single viewer session if you navigate to the same mode with another study? Do we not care about this as we are disabling it by default for now? cc @swederik  @dannyrb .

@JamesAPetts fixed the disabled state, removed the presets. I didn't touch the persistence inside the mode, I'll wait for their answers since this functionality is currently disabled although it works.

dannyrb:
Modals:

- Viewer route: About Modal is empty
- List route: User Preferences Modal is empty

Design:

I don't think @JamesAPetts realized that your UI was generated from the hotkeys config. The window level preset hotkeys should still be hotkeys, they just shouldn't be user editable or visible in the panel (for the user). I think in an old PR we created a "display name" field for hotkeys. If that's null or undefined, we could filter them out when displaying the fields.

igoroctaviano:
> Modals:
> 
> * Viewer route: About Modal is empty
> * List route: User Preferences Modal is empty
> 
> Design:
> 
> I don't think @JamesAPetts realized that your UI was generated from the hotkeys config. The window level preset hotkeys should still be hotkeys, they just shouldn't be user editable or visible in the panel (for the user). I think in an old PR we created a "display name" field for hotkeys. If that's null or undefined, we could filter them out when displaying the fields.

@dannyrb I have added the requested changes. Also, I have extracted the header component so we can reuse it, we had 2 dropdowns with the same logic before.

dannyrb:
image

No hotkey settings are visible from list page

PR #1903, opened 2020-07-15
[v2-main] Remove all legacy code

JamesAPetts:
Remove all legacy code from V2-main.

There is a lot of code which is outdated. This will:
- Reduce confusion for developers who try to look and repurpose code built for the old architecture.
- Make v2 easier to document.

When/if features are reintroduced, we can look at the old implementation from master/github history if it is relevant.

Comments on specific files.

platform/core/src/classes/metadata
- We only use the metadataprovider and DicomMetadataStore
platform/core/src/classes/StudyLoadingListener.js
- DataSources are now external
platform/core/src/classes/StudyMetadataSource.js
- DataSources are now external
platform/core/src/classes/StudyPrefetcher.js
- DataSources are now external
platform/core/src/classes/OHIFError.js
- Only used by metadata classes now not in use.
platform/core/src/classes/TypeSafeCollection.js
- Only used by metadata classes now not in us.
platform/core/src/hanging-protocols
- Unused as of yet, needs to be reimplemented as a service.
platform/core/src/utils/studyMetadataManager.js
- Replaced with StudyMetadataStore
platform/core/src/DICOMSR/parseMeasurementsData.js
- Replaced by measurements service.
platform/core/src/measurements
- Replaced measurements API with measurements service. No timepoint API in use.
platform/core/src/utils/loadAndCacheDerivedDisplaySets.js
- Need to implement derived dataSets differently through the displaySetService.
platform/core/src/lib/parsingUtils.js
- Unused now we are using naturalized DICOM JSON everywhere.
platform/core/src/studies
- Replaced with the DICOMWeb DataSource from the default extension.
platform/core/src/redux
- No longer using redux. Using React hooks + context.
platform/core/src/utils
- A lot of the files in here are outdated.
platform/core/src/ui
- A bunch of stuff here isn’t used
platform/core/src/lib
- Unused functions that appear to look like old lesion tracker code.
platform/viewer/src/googleCloud
- googleCloud adapter not yet implemented.
platform/viewer/src/lib
- Contains outdated local file loaders and pdf loaders.
platform/viewer/src/utils
- This contains initWebWorkers for cornerstoneWADOImageLoader. It is unused. Are we actually setting web workers somewhere or is it just using defaults? If we aren’t we should move this to the cornerstone extension anyway.
platform/viewer/src/config.js
- We aren’t using this anymore. Need to reimplement the errorHandler back into v2-main via appInit.js in the future.


dannyrb:
@JamesAPetts, the only thing that concerns me are the conflicts. I'm not sure why we'd have any if these are for dead code.

JamesAPetts:
Don't merge this yet, tested locally and the auto merge, although it has no conflicts, is broken.

JamesAPetts:
Updated actually, was an easy fix, some unused code was pointing to removed redux code.

PR #1920, opened 2020-07-24
OHIF 294+245 - Arrrow annotation updates.

JamesAPetts:
- Persist ArrowAnnotations if no label is given.
- Correctly show measurement tracking notifications after arrow annotate measurement instead of getting conflicting popups.


JamesAPetts:
#1921

PR #1947, opened 2020-08-03
[OHIF-313] - Sync measurement panel label updates to cornerstoneTools

JamesAPetts:
This PR adds a handler for cornerstoneTools that updates annotation labels upon an update to the measurement service annotation representation.

I had to add a field to `MeasurementService.update`'s emitted event data called `notYetUpdatedAtSource`, to prevent circular updates if the measurement service fires an update to listeners when cornerstone was the source of the update itself.

Edit: This also solved OHIF-271 & OHIF-275


JamesAPetts:
Included in #1955.

PR #1953, opened 2020-08-04
[OHIF-302] Series number for OHIF SR

JamesAPetts:
SRs are now saved with series number 4701+:

The new SR will always have a `SeriesNumber` 1 higher than the most recent in the study, if other SRs exist.

1st SR: 4701
2nd SR: 4702
3rd SR 4703 (even if SR 4701 was deleted from the study at some point)
...


extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/index.js
```diff
@@ -11,6 +11,7 @@ import { DicomMetadataStore, DICOMSR, utils } from '@ohif/core';
 import { useDebounce } from '@hooks';
 import ActionButtons from './ActionButtons';
 import { useTrackedMeasurements } from '../../getContextModule';
+import setCornerstoneMeasurementActive from '../../_shared/setCornerstoneMeasurementActive';
```
swederik, line 14: This seems unnecessary?

swederik: Turns out this is easy to fix myself via github UI

PR #1928, opened 2020-07-27
OHIF-193: Context menu integration

igoroctaviano:
### PR Checklist

- [ ] Brief description of changes
- [ ] Links to any relevant issues
- [ ] Required status checks are passing
- [ ] User cases if changes impact the user's experience
- [ ] `@mention` a maintainer to request a review

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #1830, opened 2020-06-29
feat: v2 context menu - OHIF-193

rodrigolabs:
**OHIF-193**

### Description

- Created a new component `ContextMenu` to the UI library `@ohif/ui`.
- Created a `MeasurementsContextMenu` component based on the `ContextMenu` to be used when `UIDialogService` is called.
- Basic listener implementation (right-click) in the cornerstone library to show the context menu -- most of the implementation was preserved from the current `master`.
- Basic action implementation. The final implementation for `delete`, `set label`, and `set/edit description` will be done later.

### Preview

Jun-29-2020 17-05-11

### PR Checklist

- [ ] Brief description of changes
- [ ] Links to any relevant issues
- [ ] Required status checks are passing
- [ ] User cases if changes impact the user's experience
- [ ] `@mention` a maintainer to request a review

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #1955, opened 2020-08-04
[OHIF-193 + OHIF-313]

JamesAPetts:
This PR contains the contents of #1830 #1928 and #1947 all bundled on top of the recent workflow changes, along with the rest of the features needed to complete OHIF-193.

### DIFF
--- a/extensions/cornerstone/src/init.js
+++ b/extensions/cornerstone/src/init.js
@@ -27,7 +27,7 @@ const TOOL_TYPES_WITH_CONTEXT_MENU = [
   'RectangleRoi',
 ];
 
-const _refreshViewport = () =>
+const _refreshViewports = () =>
   cs.getEnabledElements().forEach(({ element }) => cs.updateImage(element));
 
 /**
@@ -97,11 +97,7 @@ export default function init({
         onDelete: item => {
           const { tool: measurementData, toolType } = item.value;
           measurementServiceSource.remove(measurementData.id);
-          commandsManager.runCommand('removeToolState', {
-            element: event.detail.element,
-            toolType,
-            tool: measurementData,
-          });
+          _refreshViewports();
           CONTEXT_MENU_OPEN = false;
         },
         onClose: () => {
@@ -223,6 +219,18 @@ export default function init({
         : data.label
       : '';
 
+    const onSubmitHandler = ({ action, value }) => {
+      switch (action.id) {
+        case 'save':
+          callback(value.label, action.id);
+          break;
+        case 'cancel':
+          callback('', action.id);
+          break;
+      }
+      UIDialogService.dismiss({ id: dialogId });
+    };
+
     if (UIDialogService) {
       UIDialogService.create({
         id: dialogId,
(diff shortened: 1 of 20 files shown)


platform/core/src/services/MeasurementService/MeasurementService.js
```diff
@@ -610,30 +624,20 @@ class MeasurementService {
   /**
    * Broadcasts measurement changes.
    *
-   * @param {string} eventName The event name
-   * @param {MeasurementSource} source The measurement source
-   * @param {string} measurement The measurement id
+   * @param {string} eventName The event name.
+   * @param {MeasurementSource} eventData.source The measurement source.
+   * @param {string} eventData.measurement The measurement.
```
swederik, line 629: string? Probably not?

platform/ui/src/contextProviders/DialogProvider.jsx
```diff
@@ -302,3 +314,27 @@ DialogProvider.propTypes = {
...
+     * Alert if clicked on outside of element
+     */
+    function handleClickOutside(event) {
+      if (wrapperRef.current && !wrapperRef.current.contains(event.target)) {
+        props.onClickOutside();
+      }
+    }
+
+    // Bind the event listener
+    document.addEventListener('mousedown', handleClickOutside);
```
swederik, line 332: what about touch devices? We should probably add touchstart as well

swederik: Can you close dialogs via Esc?

platform/ui/src/components/ContextMenuMeasurements/ContextMenuMeasurements.jsx
```diff
@@ -0,0 +1,44 @@
...
+      label: 'Delete measurement',
+      actionType: 'Delete',
+      action: item => {
+        onDelete(item);
+        onClose();
+      },
+      value: {}
+    },
+    {
+      label: 'Relabel',
```
swederik, line 22: In previous versions, it said "Relabel" if the measurement already had a label, and "Add Label" otherwise. If we don't have that behaviour, I think it should Add Label by default.

extensions/cornerstone/src/init.js
```diff
@@ -1,40 +1,239 @@
...
 
-  const callInputDialog = (data, event, callback) => {
+  /* Measurement Service */
+  const measurementServiceSource = _connectToolsToMeasurementService(
+    MeasurementService,
+    DisplaySetService
+  );
+
+  const onRightClick = event => {
+    if (!UIDialogService) {
```
swederik, line 57: Does / can / should this ever happen? Couldn't we bail out at the top of the parent function?

extensions/cornerstone/src/init.js
```diff
@@ -1,40 +1,239 @@
...
+      onClickOutside: () => {
+        UIDialogService.dismiss({ id: 'context-menu' });
+        CONTEXT_MENU_OPEN = false;
+      },
+      contentProps: {
+        onGetMenuItems,
+        eventData: event.detail,
+        onDelete: item => {
+          const { tool: measurementData, toolType } = item.value;
+          measurementServiceSource.remove(measurementData.id);
```
swederik, line 99: do we need this AND commandsManager.runCommand('removeToolState') which happens immediately afterwards?

platform/ui/src/contextProviders/DialogProvider.jsx
JamesAPetts: Good shout on touchstart, yeah esc works on the dialogs.

extensions/cornerstone/src/init.js
JamesAPetts: fixed.

PR #1961, opened 2020-08-05
OHIF-323: Cine Dialog

igoroctaviano:
- Toggable cine tool.
- Hotkey: 'c' to toggle cine.
- CinePlayer component.
- Update to viewport grid service to support cine state.

image
image

### PR Checklist


igoroctaviano:
@JamesAPetts @evren217 Note: the design of the fps bar is slightly different here, the remaining half of the bar should have a different color but this is kinda tricky to do with native input, not sure if this is worth the time.

swederik:
> * The design of the fps bar is slightly different here, the remaining half of the bar should have a different color but this is kinda tricky to do with native input, not sure if this is worth the time.

This is fine as is, it looks good enough.

extensions/default/src/commandsModule.js
```diff
@@ -1,10 +1,21 @@
...
   const actions = {
     clearMeasurements: () => {
       MeasurementService.clear();
     },
+    toggleCine: () => {
+      const { viewports, isCineEnabled } = ViewportGridService.getState();
+      ViewportGridService.setIsCineEnabled(!isCineEnabled);
+      ToolBarService.setButton('Cine', { props: { isActive: !isCineEnabled } });
+      viewports.forEach((vp, index) => {
+        ViewportGridService.setCineForViewport({
```
JamesAPetts, line 13: We need to not polute the viewport grid service with methods specific to a viewport type.
The viewport grid service should control nothing but the viewport grid layout.

For this we should use a specific service or context (see the tracking context as an example), inside the measurement-tracking extension, that the viewport can listen for to set its UI.

cc @dannyrb @swederik you agree?

JamesAPetts: Thinking about this, we want the CINE to be reusable, so we should do this inside the cornerstone extension, and the  measurement tracking viewport should just subscribe to that service (so that the cornerstone extension's viewport could just do the same).

@dannyrb I guess we need to listen to services/context cross extension, we need to think about dependency management more concretely at some point? 🤔

igoroctaviano: @JamesAPetts https://precisionmetrics.atlassian.net/browse/OHIF-323 I decided to go with the approach in the card description. @dannyrb @JamesAPetts can we document the viewport grid service/all services and its final purpose? sometimes I found it hard to keep up with all design decisions made previously in OHIF and the atomicity of each service.

igoroctaviano: I know you guys have been thinking a lot about these services together, maybe we can set up a call to get me up to speed with design decisions?

platform/viewer/src/components/ViewportGrid.jsx
```diff
@@ -194,14 +194,15 @@ function ViewerViewportGrid(props) {
...
-            className={classNames('h-full w-full flex flex-col align-center', {
+            className={classNames('h-full w-full flex flex-col', {
               'pointer-events-none': !isActive,
             })}
           >
             <ViewportComponent
               displaySet={displaySet}
               viewportIndex={viewportIndex}
               dataSource={dataSource}
+              commandsManager={commandsManager}
```
dannyrb, line 205: I'm pretty sure the `ViewportCompnent` is the component exported/registered by the extension. Each extension module receives the `commandsManager` and `servicesManager`. If we need to pass them, we should do it _in the extension_ instead of making it an app concern that impacts _all_ ViewportComponents, as they may not all use/need them.

platform/viewer/src/appInit.js
```diff
@@ -55,7 +56,8 @@ function appInit(appConfigOrFunc, defaultExtensions) {
     DisplaySetService,
     ToolBarSerivce,
     ViewportGridService,
-    HangingProtocolService
+    HangingProtocolService,
+    CineService
```
dannyrb, line 60: So many services. I think this makes sense for our current pattern, but worth noting @JamesAPetts, how do we stop services from ballooning? What other patterns/strategies can we lean on?

platform/ui/src/assets/icons/old-play.svg
```diff
@@ -0,0 +1,8 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  viewBox="0 0 448 512"
+  aria-labelledby="title"
```
dannyrb, line 4: Missing "title" tag

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -189,14 +189,17 @@ export default [
       type: 'primary',
     },
   },
-  // TODO: Toggle
   {
     id: 'Cine',
-    type: 'ohif.action',
+    type: 'ohif.toggle',
+    config: {
+      groupName: 'primaryTool',
```
dannyrb, line 196: I think we're using this to identify which tools in a group only allow for the selection of a single tool within the group. Toggling CINE is more of a separate on/off the shows/hides UI controls.

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -408,6 +419,7 @@ function _getCornerstoneStack(displaySet, dataSource) {
   return stack;
 }
 
+// TODO -> disabled double click for now: onDoubleClick={_onDoubleClick}
```
dannyrb, line 422: True. Some functionality should probably be disabled if _isPlaying_

modes/longitudinal/src/toolbarButtons.js
igoroctaviano: > I think we're using this to identify which tools in a group only allow for the selection of a single tool within the group. Toggling CINE is more of a separate on/off the shows/hides UI controls.

I see. Maybe the name is misleading then? I thought that what you described would be a radio or something

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
igoroctaviano: This TODO was added previously, any impact? I just moved it a couple of lines

platform/ui/src/assets/icons/old-play.svg
igoroctaviano: Updated, thanks.

platform/viewer/src/components/ViewportGrid.jsx
igoroctaviano: Updated, thanks.

extensions/default/src/commandsModule.js
```diff
@@ -1,10 +1,17 @@
...
+  const { MeasurementService, ViewportGridService, ToolBarService, CineService } = servicesManager.services;
 
   const actions = {
     clearMeasurements: () => {
       MeasurementService.clear();
     },
+    toggleCine: () => {
+      const { viewports } = ViewportGridService.getState();
+      const { isCineEnabled } = CineService.getState();
+      CineService.setIsCineEnabled(!isCineEnabled);
```
dannyrb, line 11: `CineService.setIsEnabled()` or `CineService.enable()` / `CineService.disable()`

extensions/default/src/commandsModule.js
```diff
@@ -1,10 +1,17 @@
...
+      ToolBarService.setButton('Cine', { props: { isActive: !isCineEnabled } });
```
dannyrb, line 12: 🤔

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -47,19 +44,19 @@ function TrackedCornerstoneViewport({
...
   const {
     ToolBarService,
     DisplaySetService,
     MeasurementService,
   } = servicesManager.services;
   const [trackedMeasurements] = useTrackedMeasurements();
-  const [{ activeViewportIndex, viewports }] = useViewportGrid();
-  // viewportIndex, onSubmit
+  const [{ activeViewportIndex, viewports }, viewportGridService] = useViewportGrid();
+  const [{ isCineEnabled, cines }, cineService] = useCine();
```
dannyrb, line 56: cine state is hoisted?

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -71,6 +68,10 @@ function TrackedCornerstoneViewport({
     };
   }, []);
 
+  useEffect(() => {
+    cineService.setCine({ id: viewportIndex });
```
dannyrb, line 72: How does this work if I drop in a new series to this viewport? Will it continue to use the previous series settings?

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -298,7 +299,9 @@ function TrackedCornerstoneViewport({
     );
   }
 
-  // TODO -> disabled double click for now: onDoubleClick={_onDoubleClick}
+  const cine = cines[viewportIndex];
+  const isPlaying = cine && cine.isPlaying || false;
+  const frameRate = cine && cine.frameRate || 24;
```
dannyrb, line 304: does `cine.frameRate` ever not have a value?
When would `cine` be undefined?

PR #1976, opened 2020-08-07
[OHIF-326] Export as CSV.

JamesAPetts:
- Export as CSV on export click.
- Remove drop down on export for now.

Still awaiting feedback on the CSV columns format, do not merge yet.

### DIFF
--- a/extensions/cornerstone/src/utils/measurementServiceMappings/EllipticalRoi.js
+++ b/extensions/cornerstone/src/utils/measurementServiceMappings/EllipticalRoi.js
@@ -33,7 +33,9 @@ const EllipticalRoi = {
       SeriesInstanceUID
     );
 
-    const { start, end } = measurementData.handles;
+    const { cachedStats, handles } = measurementData;
+
+    const { start, end } = handles;
 
     const halfXLength = Math.abs(start.x - end.x) / 2;
     const halfYLength = Math.abs(start.y - end.y) / 2;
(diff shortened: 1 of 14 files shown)


extensions/cornerstone/src/utils/measurementServiceMappings/EllipticalRoi.js
```diff
@@ -33,7 +33,9 @@ const EllipticalRoi = {
       SeriesInstanceUID
     );
 
-    const { start, end } = measurementData.handles;
+    const { cachedStats, handles } = measurementData;
```
dannyrb, line 36: It's silly, but I don't know that we've fixed the "cachedStats" are sometimes out of date issue (based on timing for event and throttled caching). If it's fixed, great. If not, we can fix, or calculate from the handles.

extensions/measurement-tracking/src/_shared/downloadCSVReport.js
```diff
@@ -0,0 +1,164 @@
+import { DicomMetadataStore } from '@ohif/core';
+
+export default function downloadCSVReport(
```
dannyrb, line 3: Is this from the old lesion-tracker code? Can we shift to `@ohif/core`?

dannyrb:
What are the remaining blockers/unknowns here? If it's columns/format, we can posit the question on the issue, but merge this if the "we generated a CSV!" portion is met. An issue to clarify that format is non-blocking, and can be updated later (or never).

Issue #1957, opened 2020-08-05
[IDC] Hitting the MPR button launches way too many requests simultaneously


JamesAPetts:
We are launching all requests at the same time, which is severely hurting the IDC platform's auto-scaling:
https://github.com/OHIF/react-vtkjs-viewport/blob/master/src/lib/loadImageData.js#L19

We should instead use a request pool manager to only send a small number of requests (~5 at a time).

PR #1984, opened 2020-08-10
feat: 🎸 Update react-vtkjs-viewport usage to use requestPool

JamesAPetts:
RE: #1957  https://github.com/OHIF/react-vtkjs-viewport/pull/103

Switched to using the `cornerstone-tools` `requestPoolManager`, so that we don't send 1000 requests at the same time to the PACS on `imageDataObject.loadImage()`.

This is a breaking change, as we the `imageDataObject` no-longer has `insertPixelDataPromises`, instead the consumer may subscribe to an `imageDataObject`'s o`nPixelDataInserted` and/or `onAllPixelDataInserted `hooks. All the relavent examples have been updated to reflect this. As we are still pre-1.0, I have avoided a major version bump.

### DIFF
--- a/extensions/vtk/src/OHIFVTKViewport.js
+++ b/extensions/vtk/src/OHIFVTKViewport.js
@@ -49,7 +49,7 @@ class OHIFVTKViewport extends Component {
   state = {
     volumes: null,
     paintFilterLabelMapImageData: null,
-    paintFilterBackgroundImageData: null
+    paintFilterBackgroundImageData: null,
   };
 
   static propTypes = {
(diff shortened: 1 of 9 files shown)


dannyrb:
@JamesAPetts, do you need a review?

JamesAPetts:
@dannyrb Sure, bear in mind this is just an update to the new API introduced in https://github.com/OHIF/react-vtkjs-viewport/pull/103 to function in OHIF, but feel free to take a look.

extensions/vtk/src/OHIFVTKViewport.js
```diff
@@ -358,36 +360,38 @@ class OHIFVTKViewport extends Component {
   }
 
   loadProgressively(imageDataObject) {
+    debugger;
```
swederik, line 363: bad @JamesAPetts

swederik:
LGTM now that the debugger is gone

PR #1985, opened 2020-08-10
OHIF-329: Increase fetch limit when paginating more than 101 studies

igoroctaviano:
- OHIF-329: Study list never fetches results after `101` when paging
- Update data source wrapper to ignore page number changes to avoid unnecessary requests.

### PR Checklist


igoroctaviano:
@dannyrb I have used limit in this case to avoid page drifting caused by changing the offset. If we have to use offset, let's discuss how the UI/UX should be after changing the offset.

dannyrb:
User testing:

- When I load the page, it changes the URL to set a "limit". If there are no applied filters/paging, there should be no change to the URL.
    - `https://deploy-preview-1985--ohif.netlify.app/pwa/?limit=101`
- If I go to a page that exceeds that limit, the URL is updated, and the new study records load as expected. Clicking clear preserves the new limit.
    - `https://deploy-preview-1985--ohif.netlify.app/pwa/?limit=201&pageNumber=5`

I think "limit" can/should be hidden from the user. It's an implementation detail that doesn't need to be preserved in the URL as it isn't needed to determine the querying behavior. Similarly, it's probably not the "limit" that should be changed, but rather the offset. We're querying for X new records, we're only holding up to 101 in memory. For large PACS, if we keep bumping the limit, it will eventually get high enough that the query time slows significantly.

igoroctaviano:
> User testing:
> 
> * When I load the page, it changes the URL to set a "limit". If there are no applied filters/paging, there should be no change to the URL.
>   
>   * `https://deploy-preview-1985--ohif.netlify.app/pwa/?limit=101`
> * If I go to a page that exceeds that limit, the URL is updated, and the new study records load as expected. Clicking clear preserves the new limit.
>   
>   * `https://deploy-preview-1985--ohif.netlify.app/pwa/?limit=201&pageNumber=5`
> 
> I think "limit" can/should be hidden from the user. It's an implementation detail that doesn't need to be preserved in the URL as it isn't needed to determine the querying behavior. Similarly, it's probably not the "limit" that should be changed, but rather the offset. We're querying for X new records, we're only holding up to 101 in memory. For large PACS, if we keep bumping the limit, it will eventually get high enough that the query time slows significantly.

Yep, that's the offset problem I commented upfront. To keep doing client-side pagination and changing offset, we have to define how the UI will behave in this case since the offset will drift the data (because of new number of pages) after the new load. And if the user wants to go back, we have to keep track of the page number. @dannyrb do you have any suggestion? maybe something you have previously thought about

dannyrb:
> Yep, that's the offset problem I commented upfront. To keep doing client-side pagination and changing offset, we have to define how the UI will behave in this case since the offset will drift the data (because of new number of pages) after the new load. And if the user wants to go back, we have to keep track of the page number. @dannyrb do you have any suggestion? maybe something you have previously thought about

The UI and the client's URL just wouldn't care about the offset or the limit. The data source knows when:

- page number * results per page is greater than (internal query limit)
    - Request: next set of results (same query, offset of "internal query limit minus 1")
    - UI: the data to show is take "results per page" from local data, offset by: "page number" * "results per page" % "internal query limit"

With a reminder that... Paging while an request is in-flight should either be prevented or cancel the previous promise. And when a new promise is in-flight, wait to show the data (otherwise we see some timing issues / data-flicker)

Hopefully that makes sense. If not, we can review tomorrow.

igoroctaviano:
@dannyrb updated.

igoroctaviano:
> **Steps to reproduce:**
> 
> 1. Load the app
> 2. "Next page" until page 6
> 3. Note: there are 6 pages, with the last page having less than 25 items
> 4. Load this page: https://deploy-preview-1985--ohif.netlify.app/pwa/?pageNumber=6
> 5. Navigate back to page 1 using pagination arrows
> 6. Navigate to the furthers page using pagination arrows
> 7. Note: despite starting on page 6, our new "last page" is page 5.

@dannyrb the total was page based, now I updated it to use the biggest one.

platform/ui/src/components/StudyListPagination/StudyListPagination.js
```diff
@@ -29,25 +29,17 @@ const StudyListPagination = ({
       <div className="container m-auto relative px-8">
         <div className="flex justify-between">
           <div className="flex items-center">
-            <div className="relative mr-3">
-              <select
-                value={perPage}
-                className="block appearance-none w-full bg-transparent border border-common-active text-white text-base px-2 pr-4 rounded leading-tight focus:outline-none h-8"
-                onChange={e => onChangePerPage(e.target.value)}
```
dannyrb: May need to update the `onChange` -- double check.
This was a merge conflict PR that may not have been resolved correctly.

Issue #1986, opened 2020-08-11
404 handling in MPR and other modes


pieper:
### Describe the Bug

It's currently possible for the app to hang or present a black screen when a request gives 404 or otherwise fails.

### What steps can we follow to reproduce the bug?

MPR mode makes many frame requests and the server may not be able to handle them.

I was able to get this behavior by opening several tabs with different studies and entering 2D MPR mode on each one.  Eventually the IDC throttle proxy stopped responding and I got a black screen on one of the tabs.

I also went into MPR mode on Kheops with a CT and the view hung at 4% with the console full of 404 messages.

### Suggested behavior

It would be best to report back to the user that there is trouble communicating with the server, perhaps suggesting that they reload and/or try again later.

pieper:
Related to #1957, which is about avoiding too many requests, but this one is about graceful handling of failures that might happen anyway.

pieper:
@JamesAPetts we can discuss on Thursday with @fedorov.   Probably this means adding some UI to report errors from the request pool.

JamesAPetts:
Soungs good, yeah #1957 should help somewhat, but yeah a clean path for error handling without breaking the app in MPR would be nice. Look forward to discussing.

PR #1992, opened 2020-08-14
fix: 🐛 Fail gracefully on an MPR load error

JamesAPetts:
RE: #1986.

Related: https://github.com/OHIF/react-vtkjs-viewport/pull/107

- Display an error notification when one or more images fail to return from the server.
- Gracefully fail and complete loading when all promises are processed, regardless if they were successful or not. This allows us to still view an image which had few problematic requests, e.g. a single missing/malformed frame.

If many requests fail it will only show the error for the first one, to reduce spam.

Example below. I just disconnected my internet here so there is no error message, but if your server returns an error message it will be shown in the notfication:

FWIW with #1984 in place, I couldn't reproduce "real" errors on the IDC proxy with multiple tabs, which is good. I have tested faked error messages work correctly.


extensions/vtk/src/OHIFVTKViewport.js
```diff
@@ -66,6 +68,7 @@ class OHIFVTKViewport extends Component {
     viewportIndex: PropTypes.number,
     children: PropTypes.node,
     onScroll: PropTypes.func,
+    servicesManager: PropTypes.object,
```
dannyrb, line 71: Required? In general, prefer `required` as default unless there's a good reason to mark as optional. Optional means you have one more edge case to guard/test.

JamesAPetts:
The `servicesManager` is just being passed into the wrapper so that we can use the `UINotificationService` on errors, not being passed to the 3rd party lib. I'll update the branch and merge then.

Issue #1994, opened 2020-08-14
Sort series within the study 


fedorov:
## Request

Series list in the left panel should be sorted, since otherwise it appears the individual series are shown in a random, non-reproducible order. It would make sense to show by series number, and if multiple identical series numbers are encountered - sort those by date/time.

**Why should we prioritize this feature?**

* this should improve usability for the user
* current sorting order was flagged by some of the IDC developers as confusing

PR #2010, opened 2020-08-24
[IDC-1994] Sort series list by SeriesNumber, and sort by same SeriesNumber by date/time.

JamesAPetts:
RE: #1994 

Update:
- Fix regression so that we still have low priority modalities which are sorted after.
- Switch to a clever insert rather than a sort, as sorting by SeriesNumber + SeriesDate/SeriesTime on each displaySet creation was too slow.

### DIFF
--- a/platform/core/src/classes/metadata/StudyMetadata.js
+++ b/platform/core/src/classes/metadata/StudyMetadata.js
@@ -311,7 +311,7 @@ export class StudyMetadata extends Metadata {
       displaySetsForSeries.forEach(ds => this._insertDisplaySet(ds));
     });
 
-    return displaySets;
+    return this._displaySets;
   }
 
   /**
@@ -343,19 +343,6 @@ export class StudyMetadata extends Metadata {
     });
 
     return true;
-  }
-
-  /**
-   * Set display sets
-   * @param {Array} displaySets Array of display sets (ImageSet[])
-   */
-  setDisplaySets(displaySets) {
-    if (Array.isArray(displaySets) && displaySets.length > 0) {
-      // TODO: This is weird, can we just switch it to writable: true?
-      this._displaySets.splice(0);
-
-      displaySets.forEach(displaySet => this.addDisplaySet(displaySet));
-    }
   }
 
   /**
@@ -473,6 +460,7 @@ export class StudyMetadata extends Metadata {
     }
 
     this._displaySets.splice(insertIndex, 0, displaySet);
+    this.displaySets = this._displaySets;
   }
 
   /**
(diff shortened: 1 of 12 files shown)


platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -365,18 +368,36 @@ const _mapStudiesToThumbnails = function(studies) {
...
+
+      if (SeriesDate) {
+        if (SeriesTime) {
+          seriesDateTime = `${SeriesDate}${SeriesTime}`;
+        } else {
+          seriesDateTime = `${SeriesDate}$`;
+        }
+      }
+
+      if (!displaySet.hasOwnProperty('SeriesDate')) {
```
JamesAPetts, line 397: Happy for a better suggestion about how to do this @dannyrb .

The sorting is expensive and this is called many times on load whilst each displaySet is processed. The placeholder objects don't have a SeriesDate, and all displaySets do once processed. DisplaySet creation can be asynchronus so its a little sticky.

JamesAPetts: This is now outdated.

JamesAPetts:
@dannyrb ready for re-review.

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -393,6 +386,95 @@ export class StudyMetadata extends Metadata {
...
+    const displaySets = this._displaySets;
+    let insertIndex = displaySets.length;
+    let firstIndexWithSameSeriesNumber;
+
+    // If low priority, start search from next low priority.
+    if (isLowPriorityModality(displaySet.Modality)) {
+      let startingIndex;
+
+      // Find where the first low priority displaySet is.
+      for (let i = 0; i < displaySets.length; i++) {
```
dannyrb, line 407: optimization, so not blocking: we can cache this?

dannyrb: Or we can maintain these as separate lists internally, and then provide "displaysets" by joining the two

platform/core/src/classes/metadata/StudyMetadata.js
```diff
@@ -372,7 +365,7 @@ export class StudyMetadata extends Metadata {
    */
   addDisplaySet(displaySet) {
     if (displaySet instanceof ImageSet || displaySet.sopClassModule) {
-      this._displaySets.push(displaySet);
+      this._insertDisplaySet(displaySet);
```
dannyrb, line 368: Might be nice to pull this bit out.

JamesAPetts: What do you mean? pull this out where?

platform/core/src/classes/metadata/StudyMetadata.js
JamesAPetts: Good points, we can keep further optimisations in mind if speed becomes an issue (maybe with multiple hundred SRs?), but I think we've already spent enough IDC time sorting the list.

JamesAPetts:
> Does it purr like a kitten now?
It does purr like a kitten meow.

PR #2015, opened 2020-08-26
Feat/nuke global tool sync

dannyrb:
- Disables global tool sync, in favor of managing active tool manually (in the toolbar service)
- Removes 2nd toolbar row
- Fixes various small style issues with CINE player and Toolbar
- Makes ToolbarService stateful; now tracks active primary tool, toggles, and other state
- Provides a mechanism for extensions to listen for Toolbar state changes; and notify/trigger toolbar state changes
- Adds untracked variants of measurement tools to SR Viewport
- Various small fixes required due to tools sync disable
- Fixes paging and filtering issues with study list


extensions/cornerstone/src/init.js
```diff
@@ -200,93 +233,14 @@ export default function init({
...
-
   const { csToolsConfig } = configuration;
   const metadataProvider = OHIF.cornerstone.metadataProvider;
 
   cs.metaData.addProvider(metadataProvider.get.bind(metadataProvider), 9999);
 
   // ~~
   const defaultCsToolsConfig = csToolsConfig || {
-    globalToolSyncEnabled: true,
+    globalToolSyncEnabled: false, // hold on to your pants!
```
JamesAPetts, line 243: Can you remove the comment now your pants are still attached?

extensions/cornerstone/src/utils/getTools.js
```diff
@@ -0,0 +1,36 @@
+import csTools from 'cornerstone-tools';
+
+const toolsGroupedByType = {
```
JamesAPetts, line 3: An aside, but I like all this refactoring, `init` was getting super messy 👍 .

extensions/dicom-sr/src/OHIFCornerstoneSRViewport.js
```diff
@@ -64,6 +73,8 @@ function OHIFCornerstoneSRViewport({
   let trackedMeasurements;
   let sendTrackedMeasurementsEvent;
 
+  // TODO: this is a hook that fails if we register/de-register
```
JamesAPetts, line 76: Is this still a problem?

JamesAPetts: Oh I guess not for now.

extensions/dicom-sr/src/OHIFCornerstoneSRViewport.js
```diff
@@ -81,25 +92,89 @@ function OHIFCornerstoneSRViewport({
...
+  let isLocked = trackedMeasurements?.context?.trackedSeries?.length > 0;
+  useEffect(() => {
+    isLocked = trackedMeasurements?.context?.trackedSeries?.length > 0;
+  }, [trackedMeasurements]);
+
+  function _getToolAlias() {
+    const primaryToolId = ToolBarService.state.primaryToolId;
+    let toolAlias = primaryToolId;
+
+    switch (primaryToolId) {
```
JamesAPetts, line 105: I see why you've done this, it feels nasty but I can't think of a better idea. You could add some meta to the tool's configuration and key off that, but that's not much different.

extensions/dicom-sr/src/OHIFCornerstoneSRViewport.js
```diff
@@ -81,25 +92,89 @@ function OHIFCornerstoneSRViewport({
...
-
+    // ~~ MAGIC
+    cornerstoneTools.addToolForElement(targetElement, DICOMSRDisplayTool);
     cornerstoneTools.setToolEnabledForElement(
       targetElement,
       TOOL_NAMES.DICOM_SR_DISPLAY_TOOL
     );
 
+    // ~~ Variants
+    cornerstoneTools.addToolForElement(
```
JamesAPetts, line 136: We need the angle and the rectangle here, they currently are still messed up like before.

platform/core/src/services/ToolBarService/ToolBarService.js
```diff
@@ -17,24 +20,78 @@ export default class ToolBarService {
        */
     };
 
+    // TODO: Do we need to track per context? Or do we allow for a mixed
```
JamesAPetts, line 23: That's a good point, you mean if you have 2 viewports that have different tools, yeah I think we'd need to. That can be for the future though.

platform/core/src/services/ToolBarService/ToolBarService.js
```diff
@@ -17,24 +20,78 @@ export default class ToolBarService {
...
+
     Object.assign(this, pubSubServiceInterface);
   }
 
   init(extensionManager) {
     this.extensionManager = extensionManager;
   }
 
+  /**
+   *
```
JamesAPetts, line 43: I feel like this is obscure enough at first glance that fleshing this jsdoc out would help.

JamesAPetts (commented):
Mostly fine, the angle and rectangle tools need to be updated for SR.

PR #2016, opened 2020-08-27
enable docker image to be run on Google Cloud Run

nbyl:
We have an installation of the viewer in a docker image that sees not much traffic. As the project uses GCP, we opted to use Cloud Run to run the container only when the viewer is used. This PR enables the docker image to be compatible to the runtime contract (https://cloud.google.com/run/docs/reference/container-contract) by using the PORT variable as the configuration on what port to expose the service.


swederik:
Cool, that's really interesting. Change seems fine to me. I wonder if people will realize that they need to expose the new port (e.g. https://github.com/OHIF/Viewers/blob/master/Dockerfile#L62) if they set one though.

nbyl:
@swederik Unfortunately the tests failed. I'm not sure if this is related to the change. Maybe you could trigger a rebuild.

JamesAPetts:
Triggered a rebuld incase the error was unrelated.

nbyl:
@JamesAPetts @swederik After the rebuild, at least the automatic checks were successfull. If have triggered an update from the current master. If this works out, you can merge this one.

swederik:
Thanks! Sorry for the delay.

Issue #1532, opened 2020-03-18
Query params to filter/promote multiple series instances


ladeirarodolfo:
> :hand: Many people requests features. Tell us why yours is important to the
> community. How does it add value? Why _this feature_?
>
> Is your request very specific to your needs? Consider
> [contributing it](https://docs.ohif.org/contributing.html) yourself! Or reach
> out to a community member that offers
> [consulting services](https://docs.ohif.org/help.html#paid--commercial).

## Request

**What feature or change would you like to see made?**
Today, at viewer page, there is the possibility to filter or promote series based on seriesInstanceUID query param. But this works only for only one series UIDs. 
Would be useful if user could filter/promote one or any.
...

**Why should we prioritize this feature?**

...

### Acceptance criteria
- User must be capable to filter/promote series based on query param **seriesInstanceUID**
- If many seriesInstanceUID it must be separated by comma. I.e
`viewer/1.3.6.1.4.1.25403.345050719074.3824.20170126085406.1/?seriesInstanceUID=1.3.6.1.4.1.25403.345050719074.3824.20170126085406.2,2.25.374532628115939034791691984300357809891`
- If one or many seriesInstanceUID were not filtered/promoted a user message must pop up

### Technical details
- App can filter or promote series for a given series into query param
- To control  between filtering or promoting there is a flag **filterQueryParam** on app config file. False means promote is ON and true means filter is ON.
- It will filter/promote based on url query param named as **seriesInstanceUID** and only if there is a given series into that study, otherwise a message will be prompt to user.

### PR
There is already a PR going on which might help for this possible new feature
https://github.com/OHIF/Viewers/pull/1533

fedorov:
I was testing this feature with this URL:

https://idc-sandbox-000.firebaseapp.com/projects/canceridc-data/locations/us/datasets/idc/dicomStores/idc/study/1.3.6.1.4.1.14519.5.2.1.6279.6001.131383203689189807643685075952/?seriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.6279.6001.214252223927572015414741039150

I still see all of the series in the study loaded (I tried both upper and lower case for the parameter).

Did I misunderstand the syntax, or is it not yet implemented?

swederik:
[...]

PR #2026, opened 2020-09-03
IDC-1532 multiple series search google

JamesAPetts:
RE: #1532 

Google Healthcare (and potentialy other platforms), only supports a single query parameter, so you can't filter on multiple `SeriesInstanceUIDs`.

This PR splits up the calls if using the google cloud adapter (I.e. if `enableGoogleCloudAdapter` is `true` in your app config).


platform/core/src/studies/retrieveStudiesMetadata.js
```diff
@@ -11,20 +11,28 @@ import { retrieveStudyMetadata } from './retrieveStudyMetadata';
  * @param {Array} studyInstanceUIDs The UIDs of the Studies to be retrieved
  * @param {Object} [filters] - Object containing filters to be applied on retrieve metadata process
  * @param {string} [filter.seriesInstanceUID] - series instance uid to filter results against
+ * @param {boolean} [seperateSeriesInstanceUIDFilters = false] - If true, split filtered metadata calls into multiple calls,
```
swederik, line 14: Typo in separate

platform/core/src/studies/retrieveStudyMetadata.js
```diff
@@ -33,18 +40,75 @@ export function retrieveStudyMetadata(server, StudyInstanceUID, filters) {
...
+    filters.seriesInstanceUID &&
+    seperateSeriesInstanceUIDFilters
+  ) {
+    promise = _seperateSeriesRequestToAggregatePromise(
+      server,
+      StudyInstanceUID,
+      filters
+    );
+  } else {
+    promise = new Promise((resolve, reject) => {
```
swederik, line 56: what is the point of this new promise? isn't it the same as promise = RetrieveMetadata(server, StudyInstanceUID, filters)?

platform/core/src/studies/retrieveStudyMetadata.js
```diff
@@ -33,18 +40,75 @@ export function retrieveStudyMetadata(server, StudyInstanceUID, filters) {
...
+  }
 
   // Store the promise in cache
   StudyMetaDataPromises.set(StudyInstanceUID, promise);
 
   return promise;
 }
 
+/**
+ * Splits up seriesInstanceUID filters to multiple calls for platforms
```
swederik, line 70: can we capitalize SeriesInstanceUID?

platform/viewer/public/config/idc.js
```diff
@@ -7,9 +7,12 @@ window.config = function(props) {
     enableGoogleCloudAdapter: true,
     enableGoogleCloudAdapterUI: false,
     showStudyList: true,
+    enableStudyLazyLoad: false,
```
swederik, line 10: Why change this file? and why false?

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
```diff
@@ -340,6 +341,10 @@ function ViewerRetrieveStudyData({
         }
       }
 
+      if (appConfig.enableGoogleCloudAdapter) {
+        retrieveParams.push(true); // Seperate SeriesInstanceUID filter calls.
```
swederik, line 345: This is really weird. Let's hope we never reorganize the arguments to the function.

swederik (approved):
Seems fine if you fix the spelling of separate

platform/core/src/studies/retrieveStudyMetadata.js
JamesAPetts: I'm on the fence about this, `seriesInstanceUID` is the actual parameter passed to dicomweb client, that is the parameter this splits up. So I think it makes sense to have the jsdoc reflect this.

platform/viewer/public/config/idc.js
JamesAPetts: reverted.

platform/viewer/src/connectedComponents/ViewerRetrieveStudyData.js
JamesAPetts: I agree, I'd rather have a deconstructed object inside `retrieveStudiesMetadata`, but it'd change what is actually top level API in `@ohif/core`.

platform/core/src/studies/retrieveStudyMetadata.js
JamesAPetts: This is just what was in the previous code. Hmm you are right, weird. Removed and tested, all works fine.

Issue #2006, opened 2020-08-24
[IDC] Support reporting of problematic studies - Part 2 - Post report to IDC Discourse.


JamesAPetts:
Follow up to #1939

Post report of problematic studies to IDC Discourse.

- Be able to automatically post a user defined issues/message to the IDC Discourse.
- Autogenerate screenshots of the application.

fedorov:
So what would be the workflow for posting to discourse? At what point the user would need to log in? Did you think the process to be fully automatic, or OHIF would pre-populate the post for the user and let them click "submit"?

A simpler approach would be to pre-generate an email text, or send email to support@canceridc.dev. Emails sent to that address will show up on discourse, but in a private category, so we can triage it. Another benefit is that even users who are not registered with discourse can post to that address.

JamesAPetts:
For the MVP we will just have a mailto button:
- This should copy in the debug table to the body if possible.

fedorov:
Looks like the following two might fulfill many of the Discourse integration needs!
* https://meta.discourse.org/t/compose-a-new-pre-filled-personal-message-via-url/35984
* https://meta.discourse.org/t/compose-a-new-pre-filled-topic-via-url/28074

Maybe just adding a pre-populated Discourse link side by side with the mailto might be sufficient?

JamesAPetts:
Sure thing, we'll need to do this in the IDC fork though, as its rather use case specific. Once we have closed the other major open tickets we can update the fork and work on it in there.

JamesAPetts:
The last discourse bit has been pushed to post MVP on the fork.

igoroctaviano:
@fedorov regarding this issue...
- Using widgets like https://widgetbot.io/ we can create a very limited **_message-only_** panel like:
![image](https://user-images.githubusercontent.com/13886933/110715755-99ee7400-81e4-11eb-970e-a2f4314d2a86.png)
- To use advanced features like attachments, we'll need to Implement a server-side bot (using https://discord.js.org/#/) or a proxy since Discord doesn't allow connecting to the API as a bot from a browser. We may do it using Bearer tokens but it will require us to implement its oauth2 flow which I think may not worth the trouble since mailto would cover the requirements + user wouldnt need to authenticate.

pieper:
Does this support Discourse or only Discord?

igoroctaviano:
> Does this support Discourse or only Discord?

[...]

Issue #2023, opened 2020-09-01
About box reports incorrect version number


pieper:
### Describe the Bug

The about box has contradictory information.  As shown below, the version number is old but this should be a build of the current master branch.

![image](https://user-images.githubusercontent.com/126077/91900388-9e056c00-ec6c-11ea-8c08-b8afac6a1976.png)

### What steps can we follow to reproduce the bug?

Open viewer.ohif.org (or other build) and go to Options->About.

My guess is that the version is based on when the build directory was first initialized and doesn't get updated on new builds.

fedorov:
Discussed and agreed to query idcGitVersion.txt that is available on IDC deployment and show that in About. This will be done in fork, so @pieper will create an issue under fork to do this?

pieper:
After a quick look at the webpack configuration for OHIF I think I'm not the best person to do this, or at least it would take me longer.

I believe all we need to do is add this plugin: https://www.npmjs.com/package/git-revision-webpack-plugin

and then make sure it is available to use in the dialogs.  I think this would be valuable for any ohif deployment.

@JamesAPetts or @dannyrb could you help?

pieper:
Mailto tag is fixed.

About box is more tricky and I'm seeing what the OHIF team can suggest.

fedorov:
From earlier conversation with @dannyrb:

![image](https://user-images.githubusercontent.com/313942/101032167-d0628600-3547-11eb-9591-d93952bb5d76.png)

dannyrb:
Oof. Not a great comment for me to make before a long stint of vacation. Blocking off some time today to whip up a quick something.

fedorov:
Thanks @dannyrb! The intent was not to put you on the spot, but to pull together all the pieces of information related to this issue. We would most definitely appreciate your help though!

fedorov:
I have to admit I am not sure what is the best way to test this!

igoroctaviano:
> I have to admit I am not sure what is the best way to test this!

1. build the viewer (generate dist)
2. Update version
3. build the viewer again (should update the last dist)
4. check the version if its updated (before the version wasn't being updated)

 the version was based on when the build directory was first initialized and doesn't get updated on new builds

fedorov:
This was deployed to IDC production, and versions reported appear to be accurate.

PR #2027, opened 2020-09-03
[IDC-2006] - optional mailTo for debugging extension.

JamesAPetts:
RE: #2006

This PR adds an optional `mailTo` e-mail which can be set as configuration of the `@ohif/extension-debugging` component as follows:

`[OHIFDebuggingExtension, { mailTo: 'support@canceridc.dev' }],`

If set in the configuration, the button appears at the top of the debug dialog:

Screenshot 2020-09-03 at 14 28 41

The resulting e-mail body is populated with the tables data, and given an appropriate subject.

Screenshot 2020-09-03 at 14 28 00

We have no real designs for the modal, do you want more UI work done on it @fedorov? I just did the bare minimum functional requirements for now.

The debug extension (with this dialog, the download functions, etc) are not default extensions, we'll have to enable it in the IDC fork.


fedorov:
@JamesAPetts this looks great for the initial functionality, definitely sufficient for the IDC MVP!

What will be the interface to get to the debugging dialog?

Also note the related issue of version not being reported correctly - see #2023.

Issue #1672, opened 2020-04-24
Highlight segment when click to jump


JamesAPetts:
When we click on a segment in the SEG side-panel and jump to it (possible contour from the RTSTRUCT panel also?), it can be unclear which segment it is if there are many many segments.

It would be nice if this segment flashed for a brief moment so that the user could easily see what segment was jumped to.

fedorov:
This should also help when segmentation is very tiny, and does not have very good contrast:

![image](https://user-images.githubusercontent.com/313942/90694054-f50b4a00-e245-11ea-94c8-077f37eedf75.png)

PR #2034, opened 2020-09-08
[IDC-1672] Highlight SEG segment/ RT structure when click to jump.

JamesAPetts:
RE: #1672 

I've added temporary crosshairs that stay until you scroll/move from the segment, as discussed on a previous call.

I added highlighting for the RT contours, which we might as well keep as well, but without the crosshairs it really wasn't clear sometimes what was being highlighted.

I'd like to add highlighting for the DICOM SEG segments, but this actually isn't straight forward with the way our labelmaps currently work. We currently have each segment set at 100% opacity, and a global opacity slider, so we can't highlight one segment for a frame without changing the way our current opacity slider works. I think we can do this in a hacky way, but won't do this unless it is priority @fedorov .

We already had good indication in the MPR view with the crosshairs, so that is left as is.

Examples:

### DIFF
--- a/extensions/dicom-rt/src/tools/RTStructDisplayTool.js
+++ b/extensions/dicom-rt/src/tools/RTStructDisplayTool.js
@@ -96,9 +96,12 @@ export default class RTStructDisplayTool extends BaseTool {
         crossHairCenter.x /= points.length;
         crossHairCenter.y /= points.length;
 
-        options.fillStyle = color = `rgba(${colorArray[0]},${colorArray[1]},${
-          colorArray[2]
-        },${highlightOpacity})`;
+        // TODO: Disabling hightlight for now, it'd be good to bring it back
+        // when we have a good way of doing this for SEG.
+
+        // options.fillStyle = color = `rgba(${colorArray[0]},${colorArray[1]},${
+        //   colorArray[2]
+        // },${highlightOpacity})`;
 
         // Draw highlight lines.
 
--- a/extensions/dicom-rt/src/tools/RTStructDisplayTool.js
+++ b/extensions/dicom-rt/src/tools/RTStructDisplayTool.js
@@ -1,11 +1,12 @@
 import { importInternal, getToolState, toolColors } from 'cornerstone-tools';
-
+import cornerstone from 'cornerstone-core';
 import TOOL_NAMES from '../utils/toolNames';
 
 // Cornerstone 3rd party dev kit imports
 const draw = importInternal('drawing/draw');
 const drawCircle = importInternal('drawing/drawCircle');
 const drawJoinedLines = importInternal('drawing/drawJoinedLines');
+const drawLine = importInternal('drawing/drawLine');
 const getNewContext = importInternal('drawing/getNewContext');
 const BaseTool = importInternal('base/BaseTool');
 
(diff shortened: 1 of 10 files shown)


fedorov:
The demo looks great - thank you @JamesAPetts! 👍

JamesAPetts:
Awesome, glad you are happy, we'll get this after the technical review :+1:.

fedorov:
> I added highlighting for the RT contours, which we might as well keep as well, but without the crosshairs it really wasn't clear sometimes what was being highlighted.

Forgot to mention - maybe we should turn off this feature. I think it is best if we keep the user experience consistent between RT and SEG, and this will add to the current divergence which we already have. If you prefer to keep it, I am fine with that, but otherwise let's turn it off.

JamesAPetts:
> Forgot to mention - maybe we should turn off this feature. I think it is best if we keep the user experience consistent between RT and SEG, and this will add to the current divergence which we already have. If you prefer to keep it, I am fine with that, but otherwise let's turn it off.

Yeah I originally started with it but it wasn't clear enough, which is why I went for crosshairs. Long term I think its beneficial, so long as we can have it for segments too, as sometimes the crosshairs aren't super clear if there are overlapping segments in the RTSTRUCT, or in the case of SEG, where the segment is an arc, not a blob (the crosshairs just point to the "center of mass" within the plane. We could also improve the latter, but its probably fine in 99% of cases for the MVP.

I'll disable it for consistency for now 👍 .

extensions/dicom-rt/src/tools/RTStructDisplayTool.js
```diff
@@ -75,27 +81,52 @@ export default class RTStructDisplayTool extends BaseTool {
         colorArray[2]
       },${opacity})`;
 
+      // TODO
```
swederik, line 84: todo what?

extensions/dicom-rt/src/utils/drawCanvasCrosshairs.js
```diff
@@ -0,0 +1,56 @@
...
+
+export default function _drawCanvasCrosshairs(eventData, center, options) {
+  const context = getNewContext(eventData.canvasContext.canvas);
+  const { element } = eventData;
+
+  const centerCanvas = cornerstone.pixelToCanvas(element, center);
+
+  const { clientWidth: width, clientHeight: height } = element;
+
+  const offset = 10;
```
swederik, line 17: what is the offset for?

extensions/dicom-segmentation/src/utils/drawCanvasCrosshairs.js
```diff
@@ -0,0 +1,56 @@
+import cornerstone from 'cornerstone-core';
+import cornerstoneTools from 'cornerstone-tools';
+
+const { importInternal } = cornerstoneTools;
+const draw = importInternal('drawing/draw');
+const drawLine = importInternal('drawing/drawLine');
+const getNewContext = importInternal('drawing/getNewContext');
+
+export default function _drawCanvasCrosshairs(eventData, center, options) {
```
swederik, line 9: I'm having deja vu here. I guess I see why this function is in both extensions though.

JamesAPetts: Yeah it felt bad, but the extensions should be independent. We should keep this in mind when thinking about v3 and sharing generic code snippets.

extensions/dicom-rt/src/utils/drawCanvasCrosshairs.js
JamesAPetts: That is for drawing the crosshairs, i.e. the bit of the crosshair that "isn't drawn". It could be configurable, but its hardcoded for now.

extensions/dicom-rt/src/tools/RTStructDisplayTool.js
JamesAPetts: TODO: remove this pointless comment.

Issue #2052, opened 2020-09-21
Problem launching the installed PWA app


cronus17:
> **Before Creating an issue**
>
> - Are you running the latest version?
> - Are you reporting to the correct repository?
> - Did you search existing issues?

### Describe the Bug
I found this issue on both Chrome and Brave browser in the macOS Catalina 10.15.5 (19F101).
I installed your demo web app (https://viewer.ohif.org/) as an PWA app by clicking the "+" button next to the address bar on the top of the browser.
The app is successfully installed and it automatically opens the web app in the separate window.

The problem happens whenever you reopen the web app after closing the window. (when you open the web app from your desktop e.g, using Spotlight Search on mac (command + space)).
  
The error message is shown as "Sorry, this page does not exist."

I think this is because PWA tries to access https://viewer.ohif.org/index.html and the app doesn't have the valid route for the index.html.

### What steps can we follow to reproduce the bug?

1. Open Spotlight Search bar by entering command + space
2. Enter OHIF VIewer

```js
Sorry, this page does not exist.
Go back to the Study List
```

cronus17:
It is caused by the start_url property in the /platform/viewer/public/manifest.json.
This can be fixed by changing "start_url": "./index.html" to "start_url": "./".

Let me pull request for this.

cronus17:
#2070 merged

PR #2070, opened 2020-09-30
PWA start url

cronus17:
Issue: https://github.com/OHIF/Viewers/issues/2052

This changes the value of the start_url property in the manifest.json from "./index.html" to "./"

When the installed PWA app is launched, the url set in the start_url in the manifest.json is initially opened.
But the problem is ./index.html, which is currently set for the value of the property, has no matching routes, the following error is shown.
image

After changing the value to "./" this shows the main page.
 
image

### DIFF
--- a/platform/viewer/public/manifest.json
+++ b/platform/viewer/public/manifest.json
@@ -52,7 +52,7 @@
       "type": "image/png"
     }
   ],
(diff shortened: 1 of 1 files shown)


JamesAPetts:
Good catch.

Issue #2098, opened 2020-10-08
[IDC] Get in new rotatable crosshairs from react-vtkjs-viewport


JamesAPetts:
After the next round of updates to rvtkjsv, we should pull these crosshairs into OHIF.

PR #2101, opened 2020-10-09
Bring in rotatable crosshairs.

JamesAPetts:
RE: #2098 

Bring in recent crosshair updates from react-vtkjs-viewport.


salimkanoun:
Hi, amazing feature ! 

I wonder if the orientation isn't inverted in the preview,

see the screen capture, in the axial view I got the costal axis witch is oriented foward and down.

But the red line that should symbolize by axial view in the saggital view is oriented foward and up

image

JamesAPetts:
Ahh, good catch, I think you are right.

JamesAPetts:
I've moved the react-vtkjs-viewport to a brainscan which makes this much more obvious.

The issue was actually just a trivial error due to SVG and the webgl coordinates being inverted with respect to each other, I've fixed it locally, I'll ping you when its up if you want to have another look @salimkanoun  :)

salimkanoun:
thx @JamesAPetts , I will give a look, that's a very important feature as the MPR wasn't really usable without these crosshairs (you get often lost without crosshairs)

JamesAPetts:
Should be up now after netlify rebuilds (or you can pull locally and check).

salimkanoun:
Great ! Much better.

One question, the cross hair seems to be relative to the image, when you zoom In / Out you have the interaction dot moving out the field of view, and possible land on the center of the screen (which is disturbing). 

Did you considered fixing the interation dot at a fix % of the viewport axis (ex 25 % - 75%) this way the dots will stay accessible no matter of the applied zoom and will never cover the center of the image ? 

image
image

JamesAPetts:
Good points, for now we need to merge this, but feel free to make a follow up issue with feedback.

PR #2126, opened 2020-10-20
Update crosshairs UI and fix bug where you could scroll offscreen.

JamesAPetts:
Update to react-vtkjs-viewport 0.14.1 to get this patch: https://github.com/OHIF/react-vtkjs-viewport/pull/121

Remove now unneeded rotate tool.

### DIFF
--- a/extensions/vtk/package.json
+++ b/extensions/vtk/package.json
@@ -50,7 +50,7 @@
   "dependencies": {
     "@babel/runtime": "^7.5.5",
     "lodash.throttle": "^4.1.1",
-    "react-vtkjs-viewport": "^0.14.0"
+    "react-vtkjs-viewport": "^0.14.1"
   },
   "devDependencies": {
     "@ohif/core": "^2.10.7",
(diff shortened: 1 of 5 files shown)


salimkanoun:
very nice,
Small question, the zoom is available in right click, this is nice.
How can we pan the zoomed image ? Do you plan a mouse shortcut for this ?

JamesAPetts:
Hey @salimkanoun , pan is on middle-click-drag!

salimkanoun:
OK I have only the scroll wheel in my mouse, it can be clicked (physically) but it's seems not to be interpreted as a middle click button.
Could that mouse shortcut could be parametrizable In can you don't have middle click button ?

salimkanoun:
hum maybe just an issue with my mouse, Will investigate this, sorry

JamesAPetts:
> hum maybe just an issue with my mouse, Will investigate this, sorry

No problem, no need to apologise. I've not encountered this before, but its very possible vtk.js doesn't respond to all hardware.

Issue #2132, opened 2020-10-23
Some tags in DICOM Tag Browser have VR type "xs"


JamesAPetts:
In the [dictionary used to build the tag browser](https://raw.githubusercontent.com/dcmjs-org/dcmjs/master/src/dictionary.js) some fields have VR of xs, e.g. `SmallestImagePixelValue`. It seems some of these fields can be multiple VRs: https://dicom.innolitics.com/ciods/cr-image/general-image/00880200/00280106

`dcmjs` seems to cast these to other types somewhere:
https://github.com/dcmjs-org/dcmjs/blob/a091a51c4506f53fa07e3f0e932aff8ec29c9c96/src/ValueRepresentation.js#L211-L213

![image (5)](https://user-images.githubusercontent.com/25818497/97022091-9915b800-154b-11eb-97ef-ced484828cca.png)

We should see what the actual VR of these fields are of the actual DICOM here:
Study on IDC: 1.3.6.1.4.1.14519.5.2.1.7695.1700.186130315530827043916665138479

And investigate if logic can be written to replace the dictionary tags with the correct ones based on the dataset. The VR will actually be stored on the DICOM, so maybe we need to read it explictly when naturalizing the data using the dictionary.

PR #2143, opened 2020-11-02
Some tags in DICOM Tag Browser have VR type "xs"

igoroctaviano:
#2132: Add a fallback conditional to use the original dataset VR when tag is statically defined as XS by the dcmjs dictionary.
- This PR also removes proptype errors from DICOMTagBrowser and ErrorBoundary.


Punzo:
tested locally works, and looks good.

LGTM +1

Punzo (approved):

sscotti:
I don't know if this is relevant to the report, but I work with Orthanc quite a bit, and it appears the that xs VR type and also the lt VR type is somewhat non-standard.  There was / is a bug in Orthanc whereby is doesn't return the values for those tags in the DICOMWeb reqeusts on that platform.  It might be fixed shortly:

e.g.

(0028,3010) SQ (Sequence with explicit length #=1)      # 8222, 1 VOILUTSequence
  (fffe,e000) na (Item with explicit length #=2)          # 8214, 1 Item
    (0028,3002) xs 4096\0\12                                #   6, 3 LUTDescriptor
    (0028,3006) lt 0002\0002\0002\0002\0002\0002\0002\0002\0002\0002\0002\0002\0002... # 8192, 1 LUTData
  (fffe,e00d) na (ItemDelimitationItem for re-encoding)   #   0, 0 ItemDelimitationItem
(fffe,e0dd) na (SequenceDelimitationItem for re-encod.) #   0, 0 SequenceDelimitationItem

is the dcmdump from the original instance.

DICOMWeb on Orthanc was return NULL or EMPTY for those.

That usually happens when WindowCenter and WindowWidth are empty and there is a LUT specified instead.

PR #2148, opened 2020-11-04
Added German translations

matthiasg:
This pull request is in relation to #618 and adds some German translations.


matthiasg:
Fixed missing exports

PR #2149, opened 2020-11-04
Added de (German) translations

matthiasg:
Added translations for German usage (only de, not de-DE,de-AT, etc but they shouldn't differ much anyway)


matthiasg:
Related to  #618

matthiasg:
Is anything not ok with this ?

swederik:
Hi @matthiasg, thanks for the PR! There's nothing wrong with it that I can see.

Our original idea was that we could allow translators to use the Locize.com service to add their translations via crowdsourced links. Here is the link for German: https://www.locize.app/register?invitation=Zw7HarcoKIU9EwvVwjgZX7FaryZFvsHEHlG1qONEGbd8pmHbnFQbxwoi01GF7e9v

Then what happens is that we pull the data from Locize as the source of truth to update the repository. The assumption was that translations wouldn't really be contributed via Github, but we seem to be getting them (this PR, there's another Russian one too), so maybe we should rethink the approach, or test sync from repo to Locize, maybe I'll look at that with your branch.

swederik:
Just merged this via another branch.

Thanks for the contribution!

Issue #2109, opened 2020-10-14
Internet dependency


matthiasg:
Even though most users would have access to the internet and the google domain is pretty safe from a security standpoint. When trying to version the code for compliance, I would rather have the option to integrate the external reference. This would also allow for intranet deployments without internet access (which many radiologist use)

https://github.com/OHIF/Viewers/blob/352407c71ae93946e9ebad41446d6086cfbc237b/platform/viewer/public/init-service-worker.js#L6

swederik:
I completely agree that there should be no internet dependency. We also
need to remove polyfill.io usage.

matthiasg:
The 'stale' bot is quite quick :) I added a pull request to address it now if anybody is interested. 

I am still trying to deepen my understanding of the project and as such and as always would welcome any feedback.

salimkanoun:
Hi, just in case, this commit has a side effect for users who will serve OHIF through Apache as the distribution now contains .mjs file that are emited with text/plain MIME Type (instead of javascript). 
This has been fixed recently in apache 2.5.1 (https://github.com/apache/httpd/blob/trunk/CHANGES) (jun 2022) but many linux distro are not updated yet

sedghi:
@matthiasg @salimkanoun Do you know if this still is a problem?

sedghi:
@matthiasg @salimkanoun bump

PR #2151, opened 2020-11-05
No internet dependency (addressing #2109)

matthiasg:
As was mentioned in #2109, the build should have the least amount (preferably none) of outside dependencies which require an internet connection or any other dynamically updated code from outside sources that is not part of the distribution.

Primary Reason:

Handling of SOUPs ([Software of Unknown Provenance](https://www.johner-institute.com/articles/software-iec-62304/soup-and-ots/)) is integral to getting certifications and having potential software changes from the outside makes it much harder.

Of course one can argue that `workbox` is version locked on what is in essence a CDN, but depending on the auditing process this could lead to unnecessary complexity in proving how this is ascertained (there are ways of course).

Secondary Reason:

Also a standalone viewing library for medical images should not be required to have a dependency on a working Internet connection as many internal medical networks are disconnected from that and would at the very least require exemptions to be put into place just for the sake of using OHIF.


matthiasg:
I did realize in this pull request I also included a new .gitignore rule to exclude *.js.map files which are generated during the build process. This could potentially be done in a separate pull request if so desired.

matthiasg:
It is just waiting for the pull request to be merged :)

matthiasg:
Nobody ? ;) 

We appreciate the amount of work that goes into reviewing even smaller PR as this. Is there anything we can contribute to ensure more resources are available ?

swederik:
Thanks! So sorry for the ridiculously long reply to this @matthiasg!

PR #2161, opened 2020-11-09
added MAX_CONCURRENCY limit to prevent crashes on bigger machines

matthiasg:
As each webworker costs RAM and machines with more cores are available (like AMD puts out or XEON) this can and does easily crash the processes.

Added a limit of six as a variable, but it could also be defined in a way overridable during the build process.

PS: In general, many more places in the code would benefit from BUILD time variables.

### DIFF
--- a/platform/viewer/src/utils/initWebWorkers.js
+++ b/platform/viewer/src/utils/initWebWorkers.js
@@ -2,9 +2,11 @@ import cornerstoneWADOImageLoader from 'cornerstone-wado-image-loader';
 
 let initialized = false;
 
+const MAX_CONCURRENCY = 6;
+
 export default function initWebWorkers() {
   const config = {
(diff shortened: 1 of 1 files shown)


matthiasg:
This change is quite important for high core count machines, is there anything we can improve to allow for merging ?

swederik:
It looks fine to me, but I have no high-core machines to test it on so I'll have to trust you. I can easily imagine that you could hit memory limits if you started up 20 webworkers for decoding.

matthiasg:
@swederik thanks ! Great to hear about more manpower.

We are in the process of integrating OHIF into our own platform and in the process trying to fix any bugs or extend to make it more general. But we obviously don't want to step into anybodys toes as we don't know you or your long term design goals yet. 

We do have experience in DICOM viewing, due to our previous PACS and DICOM Viewer product lines, but have extended into a more general direction and think that putting effort into OHIF/Viewer ourselves makes more sense.

Issue #2117, opened 2020-10-15
RTSTRUCT tab is shown when not applicable, does not load when opened


fedorov:
If I open this study, it shows the dose image, and somehow RTSTRUCT tab button is populated. When I click that button, I get a spinner and nothing loads.

Study: 1.3.6.1.4.1.14519.5.2.1.1706.4009.149676576527243140820654927789
Series: 1.3.6.1.4.1.14519.5.2.1.1706.4009.104748144873675756696310742946

![image](https://user-images.githubusercontent.com/313942/96172386-5bce7c00-0ef4-11eb-8dc9-e5492b9f8aa3.png)

fedorov:
Seems to be reproducible with any RTDOSE? 

![image](https://user-images.githubusercontent.com/313942/98252284-ba14da80-1f47-11eb-8b7c-cc9efd86260e.png)

igoroctaviano:
@fedorov I have opened the PR to dynamically disable the RT panel if there are no rtstructs. 

About this study uid you gave: 1.3.6.1.4.1.14519.5.2.1.1706.4009.149676576527243140820654927789 
Its not showing because the only series which have RTs its modality is CT and not RTStruct so now at least the button will be disabled for those. The other series is RTDOSE but they dont have any RTs.

The same happens here: http://localhost:3000/projects/chc-tcia/locations/us-central1/datasets/hnscc-3dct-rt/dicomStores/hnscc-3dct-rt/study/1.3.6.1.4.1.14519.5.2.1.2193.7172.113038498776356011546245623307
- This one again, CT modality with RTStructs and other series with modality RTDOSE and empty RTs.

fedorov:
That's very weird, I need to look into that... I will assign this to myself to hopefully not forget about it...

fedorov:
> Its not showing because the only series which have RTs its modality is CT and not RTStruct so now at least the button will be disabled for those. 

@igoroctaviano I looked, and I do not understand your point.

In that study referenced above, RTSTRUCT series is definitely RTSTRUCT. Can you explain your point?

igoroctaviano:
> > Its not showing because the only series which have RTs its modality is CT and not RTStruct so now at least the button will be disabled for those.
> 
> @igoroctaviano I looked, and I do not understand your point.
> 
> In that study referenced above, RTSTRUCT series is definitely RTSTRUCT. Can you explain your point?

[...]

PR #2168, opened 2020-11-13
RTSTRUCT tab is shown when not applicable, does not load when opened 

igoroctaviano:
Issue: #2117
- The rt struct event was resolving with the wrong active viewport data. I have updated the codebase so the event grabs the latest/correct active viewport.

http://localhost:3000/projects/chc-tcia/locations/us-central1/datasets/hnscc-3dct-rt/dicomStores/hnscc-3dct-rt/study/1.3.6.1.4.1.14519.5.2.1.2193.7172.113038498776356011546245623307

- RTDOSE series should have empty rt struct panel.
- Loading should only appear of series have rtstructs.
- RTStructs should be displayed in the panel.


Punzo (approved):
:+1:

Issue #2164, opened 2020-11-11
Support for overlapping SEG segments


fedorov:
I didn't realize it until now that overlapping segments are not supported. Example below shows an example of such study: https://idc-sandbox-000.firebaseapp.com/projects/idc-tcia/locations/us-central1/datasets/tcia-idc-datareviewcoordination/dicomStores/issue11-lung1-20201015-review/study/1.3.6.1.4.1.32722.99.99.203715003805996641695765332389135385095

![image](https://user-images.githubusercontent.com/313942/98854737-f1dcc000-2428-11eb-92df-0aee2ed3985f.png)

Specifically, look for the GTV-1 in RTSTRUCT and Neoplasm in SEG, which should match up. 

Same region below is shown in Slicer, SEG is in faded green outline.

![image](https://user-images.githubusercontent.com/313942/98854848-1c2e7d80-2429-11eb-9577-6a5820025245.png)

As a short term solution, until this is fixed, I think it would be good to give some kind of a warning alerting the user when overlapping segments are present.

pieper:
I'm not sure I completely understand the issue.  Is discussing this with a demo at the meeting next Thursday soon enough?

fedorov:
Not to imply I want it to be fixed by Thursday, but I do want the issue to be understood, so let me try again.

For the study here: https://idc-sandbox-000.firebaseapp.com/projects/idc-tcia/locations/us-central1/datasets/tcia-idc-datareviewcoordination/dicomStores/issue11-lung1-20201015-review/study/1.3.6.1.4.1.32722.99.99.203715003805996641695765332389135385095

There is segmentation of tumor in RTSTRUCT and the same tumor segmentation in SEG.

## OHIF Viewer

* RTSTRUCT tumor contour

![image](https://user-images.githubusercontent.com/313942/98858243-4c2c4f80-242e-11eb-96d4-d95f43dca772.png)

* SEG tumor contour

![image](https://user-images.githubusercontent.com/313942/98858330-6e25d200-242e-11eb-8878-6777e83cc218.png)


## 3D Slicer

* RTSTRUCT tumor contour

![image](https://user-images.githubusercontent.com/313942/98858383-81d13880-242e-11eb-99f0-dfb5d0eb8a83.png)


* SEG tumor contour

![image](https://user-images.githubusercontent.com/313942/98858403-8b5aa080-242e-11eb-8641-59265f9cf61b.png)

[...]

PR #2185, opened 2020-11-25
Re #2164: load overlapping segmentation WIP

Punzo:
WIP: this is a work in progress PR

based on the logic implemented in https://github.com/dcmjs-org/dcmjs/pull/155 as described in point (2) in https://github.com/OHIF/Viewers/issues/2164#issuecomment-732240664.

dcmjs can return an array of LabelMaps if segments are overlapping. In this PR we load the labelMaps in the segmentation display and we define as active the first one https://github.com/OHIF/Viewers/compare/master...Punzo:IDC2164?expand=1#diff-0835569e2ca4d01ae2d66254833e8ab800b668a0cdce905e21c127e05c775f9bR105

Still to do:

- the UI does not allow to change the active label Map. It would be necessary to have in the UI, widget to see all the loaded labelMaps and pick the active one, so that one can change between:

image
image

- at the moment, when using MPR mode (vtk.js) only the active one will be passed from cornerstone to vtk.

- memory consumption can be large and we should add a warning.

### DIFF
--- a/extensions/cornerstone/package.json
+++ b/extensions/cornerstone/package.json
@@ -36,7 +36,7 @@
     "cornerstone-math": "^0.1.8",
     "cornerstone-tools": "^4.20.1",
     "cornerstone-wado-image-loader": "^3.1.0",
-    "dcmjs": "0.16.5",
+    "dcmjs": "0.17.0",
     "dicom-parser": "^1.8.3",
     "hammerjs": "^2.0.8",
     "prop-types": "^15.6.2",
--- a/extensions/dicom-html/package.json
+++ b/extensions/dicom-html/package.json
@@ -29,7 +29,7 @@
   },
   "peerDependencies": {
     "@ohif/core": "^0.50.0",
-    "dcmjs": "0.16.5",
+    "dcmjs": "0.17.0",
     "prop-types": "^15.6.2",
     "react": "^16.8.6",
     "react-dom": "^16.8.6"
--- a/extensions/dicom-rt/package.json
+++ b/extensions/dicom-rt/package.json
@@ -31,7 +31,7 @@
     "@ohif/core": "^0.50.0",
     "cornerstone-core": "^2.2.8",
     "cornerstone-tools": "^4.20.1",
-    "dcmjs": "0.16.5",
+    "dcmjs": "0.17.0",
     "gl-matrix": "^3.3.0",
     "prop-types": "^15.6.2",
     "react": "^16.8.6",
--- a/extensions/dicom-segmentation/package.json
+++ b/extensions/dicom-segmentation/package.json
@@ -31,7 +31,7 @@
     "@ohif/core": "^0.50.0",
     "cornerstone-core": "^2.2.8",
     "cornerstone-tools": "^4.20.1",
-    "dcmjs": "0.16.5",
+    "dcmjs": "0.17.0",
     "prop-types": "^15.6.2",
     "react": "^16.8.6",
     "react-dom": "^16.8.6"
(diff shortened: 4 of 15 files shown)


Punzo:
@igoroctaviano for point 1 I was thinking to have in the UI something like this:
image

to pick the active labelMap of teh segmentation. Please let me know what do you think about it.

fedorov:
Can we skip the selector of the active label map, by default show segments from all of the label maps in the viewer, and show the combined list of all segments from all label maps under the segmentation selector?

I do not like the concept of the labelmap selector available to the user.

fedorov:
Sorry, closed by accident!

Punzo:
> Sorry, closed by accident!

no problem.

> by default show segments from all of the label maps in the viewer, and show the combined list of all segments from all label maps under the segmentation selector?

A) for doing this properly, we would need cornerstone/vtk/OIHF UI to handle multiple active labelMaps (for a given segmentation item). This would need major changes in the OIHF UI and cornerstonetools rendering (e.g. https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/eventListeners/internals/renderSegmentationFill.js) and how the segmentation is copied from cornerstone to vtk. 

B) with the current infrastructure is actually not possible without having bugs in the UI. An example: if you have two labelMaps as in screenshots (https://github.com/OHIF/Viewers/pull/2185#issue-527522637), we can set the first labelMap (segments 1,2,3,6) as active and the second labelMap (segments 4 and 5) as inactive. Then, we could force to show anyway all the segments controls for both the label Maps as:

image

but the UI controls for segments 4 and 5 will not work. Moreover, in vtk things will be broken, because only the the active label map will be copied in vtk, and therefore only the segments of the active labelmap will be visualized. 

@igoroctaviano if you have any idea (better than https://github.com/OHIF/Viewers/pull/2185#issuecomment-733782643), please feel free to comment.

P.S.: note there is a substantial difference as active and non active LabelMap are rendered in cornerstoneTools and this reflects on the OIHF UI: 
image

fedorov:
Thanks for the explanation Davide! I think we will need to brainstorm this at the call on Thursday, and get feedback from @pieper before proceeding with the implementation.

Punzo:
> Thanks for the explanation Davide! I think we will need to brainstorm this at the call on Thursday, and get feedback from @pieper before proceeding with the implementation.

Fine for me :+1:

Punzo:
we can just dynamically swap active label map when the user select the segment

Punzo:
> we can just dynamically swap active label map when the user select the segment **(solutionCornerstone)**

@fedorov @pieper this will be a perfect solution for cornerstone viewport, but I actually forgot that there is an additional issue with VTK (my fault I forgot to talk about it during the meeting): VTK currrently take only the active label map from cornerstone.

We can update the Cornerstone-VTK connection to pass all the labelmaps (not only the active one) to VTK **(solutionVTK)**, but VTK has 3 views and duplicates the volume for each view. This will bring memory issues for sure for large datasets.

So our plan would be to use **(solutionCornerstone)** for cornerstone, and then **(solutionVTK)** for VTK. But in the case there will be memory issue for VTK (based on estimation of number of label maps and size of the volumes < 1-1.5 Gb), we would disable the VTK viewport. What do you think about it?

pieper:
I'd say let's not worry about the triple memory issue, since that will go away once we switch to the off-screen MPR mode.

We do already have some series that cannot be rendered in MPR mode (e.g. the 3000 slice one) so we need to have graceful handling of out-of-memory scenarios anyway.

igoroctaviano:
@Punzo https://github.com/Punzo/Viewers/pull/1 here is the PR to dynamically update the label maps on segment click.

Punzo:
> @Punzo [Punzo#1](https://github.com/Punzo/Viewers/pull/1) here is the PR to dynamically update the label maps on segment click.

tested https://github.com/Punzo/Viewers/pull/1 everything working, great! thanks. I'll merge everything now and I'll update the IDC fork as well.

PR #2204, opened 2020-12-09
About box reports incorrect version number

igoroctaviano:
- Use runtime value configuration to get pkg version.
- Issue: #2023


Punzo:
tested locally on version 4.7.2  

image
image

looks good

Issue #2192, opened 2020-11-30
Buffer allocation failure trying to do MPR for a large series


fedorov:
Here's a large series that takes very very long time to download, and causes MRP failure:

https://viewer.imaging.datacommons.cancer.gov/viewer/1.3.6.1.4.1.14519.5.2.1.1706.4009.243406849348906151042089406192?seriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.1706.4009.304054433668574201838767925472

![image](https://user-images.githubusercontent.com/313942/100664709-09231500-3326-11eb-8abc-137326a0015b.png)

Is this the expected behavior for this specific dataset? When the series is too large, is it possible to disable MPR button?

pieper:
yes, it's 3000 images, so it makes sense to me that it's too large.

We may not be able to know the exact allocation limit in advance, but we could certainly detect this error and give a more helpful message.  Ideally offering an alternative viewing option (e.g. download to slicer or similar).

I would bump this one to IDC:priorirty.

fedorov:
Here's a legacy converted enhanced MF version of the same series, just in case: https://idc-sandbox-000.firebaseapp.com/projects/idc-tcia/locations/us/datasets/afshin_results_00_idc_tcia_mvp_wave0/dicomStores/MULTIFRAME/study/1.3.6.1.4.1.14519.5.2.1.1706.4009.243406849348906151042089406192?seriesInstanceUID=1.3.6.1.4.1.43046.3.1.9.5060456482850075494393251866013736567025

pieper:
Here's the same study in the sandbox in case that helps:

https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc-tcia/dicomStores/idc-tcia/study/1.3.6.1.4.1.14519.5.2.1.1706.4009.243406849348906151042089406192

pieper:
Related to #2223

igoroctaviano:
@pieper @fedorov This PR adds a more helpful notification: https://github.com/OHIF/Viewers/pull/2246.

...also it adds a new option to the notification service which allows us to define an action (button), as suggested by Steve, we can use this action option to define a new button that will trigger something like: `download to slicer`.  If that's the case, we should align that's the correct behavior and add the action to that PR.

pieper:
[...]

PR #2246, opened 2021-01-20
feat(IDC-2192): Notify buffer allocation failure trying to do MPR for a large series

igoroctaviano:
Issue: #2192 

- Update the snack bar to have a new option to define an action.
- Add notification of exceeded buffer limit if buffer related errors.
- Add return button to exit mpr after failure.

image

### DIFF
--- a/extensions/vtk/src/OHIFVTKViewport.js
+++ b/extensions/vtk/src/OHIFVTKViewport.js
@@ -363,7 +363,7 @@ class OHIFVTKViewport extends Component {
       const { UINotificationService } = this.props.servicesManager.services;
       if (this.props.viewportIndex === 0) {
         const message = error.message.includes('buffer')
-          ? 'Buffer allocation limit exceeded'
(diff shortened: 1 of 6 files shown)


Punzo (approved):

Punzo (commented):
Instead of trying to load the dataset and catch the error, would not be simpler to just check the dimensions of the volume and segmentation (you can get them from the vtkImageData https://kitware.github.io/vtk-js/api/Common_DataModel_ImageData.html#getDimensions-setDimensions-array-i-j-k-setDimensions-i-j-k) and calculate the relative memory needed? and if > than a limit return immediately? Or we could even disable the MPR button in these cases?

Punzo:
uhm it is true that an user can modify the maximum ram available in a tab, (e.g. in google-chorme by setting the max_old_space_size variable). So picking a memory theshold as I suggested in https://github.com/OHIF/Viewers/pull/2246#pullrequestreview-573098502 is not straightforward.

PR #2250, opened 2021-01-22
fixes #2249

### LINKED ISSUE #2249: Unable to display study list after anonymizing using Google Healthcare API
## Bug Report

Unable to display study list after anonymizing using Google Healthcare API

### Describe the Bug

_When using the [de-id functionality](https://cloud.google.com/healthcare/docs/how-tos/dicom-deidentify) on a DICOM store in the Google Healthcare API, certain tags that are used in the QIDO-RS DICOMweb query are removed from the DICOM metadata. This includes the StudyDate. The OHIF query for study list relies directly on StudyDate when no parameters are entered and uses a default value, which yields no results._

### What steps can we follow to reproduce the bug?

1. Create a de-id store using the Google Healthcare API using the DEIDENTIFY_TAG_CONTENTS profile.
2. Deploy OHIF and associate with the Dicom Store as described [here](https://docs.ohif.org/connecting-to-image-archives/google-cloud-healthcare.html)
3. Notice the study list shows nothing.

jasonklotzer:
Fixes issue #2249 

@swederik

### DIFF
--- a/platform/viewer/src/studylist/StudyListRoute.js
+++ b/platform/viewer/src/studylist/StudyListRoute.js
@@ -336,12 +336,6 @@ async function getStudyList(
   } = filters;
   const sortFieldName = sort.fieldName || 'PatientName';
   const sortDirection = sort.direction || 'desc';
-  const studyDateFrom =
-    filters.studyDateFrom ||
-    moment()
-      .subtract(25000, 'days')
-      .toDate();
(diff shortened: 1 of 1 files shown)


jasonklotzer:
Looks like the Netlify deployment failed due to a timeout.

swederik:
I just restarted the netlify build, not sure why it would have timed out. We'll see if it happens again

Issue #2258, opened 2021-01-28
Replace dropdown with slider for series instance selector in DICOM Tag viewer


fedorov:
Slider is a lot more useful for quickly examining differences across instances. It is also what is implemented in Slicer, and proved to be helpful.

![image](https://user-images.githubusercontent.com/313942/106175734-760b2d80-6164-11eb-9f00-44485e062d15.png)

PR #2260, opened 2021-01-28
IDC-2258: Replace instance dropdown to slider for dicom tag browser

igoroctaviano:
Slider is a lot more useful for quickly examining differences across instances. It is also what is implemented in Slicer, and proved to be helpful.

#2258 

image


Punzo (approved):
LGTM!

Issue #2251, opened 2021-01-25
Black screen opening series not available in the DICOM Store


fedorov:
This IDC series does not load in sandbox:

https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc-tcia/dicomStores/idc-tcia/study/1.3.46.670589.11.17169.5.0.12048.2015091016112459008?seriesInstanceUID=1.3.46.670589.11.17169.5.0.6180.2015091017050303723

storage bucket: https://console.cloud.google.com/storage/browser/idc-tcia-pdmr-bl0293-f563/dicom/1.3.46.670589.11.17169.5.0.12048.2015091016112459008

It's a mouse image, but dciodvfy does not detect any major issues. 

![image](https://user-images.githubusercontent.com/313942/105647861-61821900-5e76-11eb-9e88-9455dfc8767e.png)

If this is because it is not in the DICOM Store, the message to the user should be better than a black screen.

![image](https://user-images.githubusercontent.com/313942/105647814-17009c80-5e76-11eb-959a-9ba714629526.png)

cc @bcli4d @wlongabaugh

fedorov:
Correction: the valid URL for the series in question is this https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc_tcia_dev/dicomStores/idc_tcia/study/1.3.46.670589.11.17169.5.0.12048.2015091016112459008.

So the issue is just poor communication to the user about the error - it should not be black screen.

pieper:
I don't get a black screen.  Can you try again?

![image](https://user-images.githubusercontent.com/126077/105730297-2e826880-5efc-11eb-8f24-75f16e566175.png)

fedorov:
@pieper the black screen will appear for the URL in the first issue post, since the series is not available in that store.

pieper:
I see, I thought when you said "correction" you meant that was the URL to replicate the issue.  Yes, the error should be caught and a reasonable message should be given (something like a 404 screen saying content cannot be found).

PR #2267, opened 2021-02-03
IDC-2251: Add error page and not found pages if failed to retrieve study data

igoroctaviano:
Issue: #2251 

The ViewerRetrieveStudyData component was not rendering an error page if an error occurred, rendering a black screen without any information. Now if any not found error occurs it will show a not found page otherwise it will show a message that it failed to load the study data.

Screen Shot 2021-02-03 at 13 59 51


Punzo (approved):
LGTM

PR #2283, opened 2021-02-25
feat(IDC-2223): add new log service to be used by debugger extension

igoroctaviano:
Screen Shot 2021-02-25 at 18 26 34


igoroctaviano:
@Punzo if possible, test the updated parts of the app by forcing errors in the code and checking the debug dialog. thank you in advance.

fedorov:
Very nice, thanks!

Punzo:
@igoroctaviano tetsed very nice. I like it a lot. I have two comments/questons:

1. should we report all the  OIHF errors (either printed in console.error/throw new Error) in the LoggerService.error?
2. should we can catch error from third party library and print them in the LoggerService.error too? 

@igoroctaviano @fedorov what do you think about (1) and (2).

If we need to report all the error, in order to automatize it, instead of going along the code and change all the console.error etc... into LoggerService.error, is there any way to catch the error printed on console and copy and paste in the LoggerService? essentially something like the qInstallMsgHandler in Qt fro desktop application (e.g. [link](https://stackoverflow.com/questions/4954140/how-to-redirect-qdebug-qwarning-qcritical-etc-output) ) . I see that there is an [eventListener(https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror) , @igoroctaviano  is this a good option?

igoroctaviano:
> @igoroctaviano tested, very nice! I like it a lot. I have two comments/questons:
> 
> 1. should we report all the  OIHF errors (either printed in console.error/throw new Error) in the LoggerService.error?
> 2. should we can catch error from third party libraries and print them in the LoggerService.error too?
> 
> @igoroctaviano @fedorov what do you think about (1) and (2)?
> 
> If we need to report all the errors, in order to automatize it, instead of going along the code and change manually all the console.error etc... into LoggerService.error, is there any way to catch the error printed on console and copy and paste in the LoggerService? essentially something like the qInstallMsgHandler in Qt for desktop application (e.g. [link](https://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler)) . I see that there is an [eventListener](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror), @igoroctaviano would this be a good option?

Thanks for the review.

@Punzo For point 2, we have the error boundary which is basically a layer of error handling/logging but it's not always possible to just rely on a generic layer without losing some error context, that's why we have some catches inside extensions that notify the user. BTW I would love to hear how we could implement this in the browser using the current architecture.

For point 1. I have added a flag to control errors being displayed in the browser's console, we just need to decide. The majority of catches in the app were just displaying notification and the console log was being done inside the notification which I manage to move to the logger service.

Punzo:
> @Punzo For point 2, we have the error boundary which is basically a layer of error handling/logging but it's not always possible to just rely on a generic layer without losing some error context, that's why we have some catches inside extensions that notify the user. BTW I would love to hear how we could implement this in the browser using the current architecture. 

ok thanks for the explanation, we can return on this later on.

> For point 1. I have added a flag to control errors being displayed in the browser's console, we just need to decide. The majority of catches in the app were just displaying notification and the console log was being done inside the notification which I manage to move to the logger service.

what I mean is the opposite. I see in the code that there are still many error printed directly to console (e.g.: https://github.com/OHIF/Viewers/blob/IDC-2223-update/extensions/vtk/src/OHIFVTKViewport.js#L362). These will not get catched by the new  LoggerService, are they? I guess the easier way is to just replace the console.error calls with  LoggerService.error calls. 

However, my point is that I was wandering if we can have a stream of the errors like https://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler. In this way along the code you don't have to replace manually console.error with LoggerService.error calls, but you automatically stream the errors into another output.

Punzo:
Of course, if you have already added everywhere LoggerService (e.g. https://github.com/OHIF/Viewers/blob/IDC-2223-update/extensions/vtk/src/OHIFVTKViewport.js#L371 is realtiove to https://github.com/OHIF/Viewers/blob/IDC-2223-update/extensions/vtk/src/OHIFVTKViewport.js#L362)). For me it is fine and good to go.

Punzo (approved):
LGTM!

igoroctaviano:
> > @Punzo For point 2, we have the error boundary which is basically a layer of error handling/logging but it's not always possible to just rely on a generic layer without losing some error context, that's why we have some catches inside extensions that notify the user. BTW I would love to hear how we could implement this in the browser using the current architecture.
> 
> ok thanks for the explanation, we can return on this later on.
> 
> > For point 1. I have added a flag to control errors being displayed in the browser's console, we just need to decide. The majority of catches in the app were just displaying notification and the console log was being done inside the notification which I manage to move to the logger service.
> 
> what I mean is the opposite. I see in the code that there are still many error printed directly to console (e.g.: https://github.com/OHIF/Viewers/blob/IDC-2223-update/extensions/vtk/src/OHIFVTKViewport.js#L362). These will not get catched by the new LoggerService, are they? I guess the easier way is to just replace the console.error calls with LoggerService.error calls.
> 
> Of course, if you have already added everywhere LoggerService.error for each console.error (e.g. https://github.com/OHIF/Viewers/blob/IDC-2223-update/extensions/vtk/src/OHIFVTKViewport.js#L371 is realtiove to https://github.com/OHIF/Viewers/blob/IDC-2223-update/extensions/vtk/src/OHIFVTKViewport.js#L362)), for me it is good to go.
> 
> However, my point is that I was wandering if we can have a stream of the errors like https://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler. In this way along the code you don't have to replace manually console.error with LoggerService.error calls, but you automatically stream the errors into another output (and we should be able to catch errors also from third party libraries in this way). But again, this is out of scope of this PR and we can talk about it later on.

Got it, that makes sense, replacing all would be a more delicate change because of service imports. I could change the javascript console prototype for now if needed to call the logger service, this way we dont have to replace the old ones. lets see what @fedorov thinks and ill update the PR.

fedorov:
@igoroctaviano thank you for this feature, it looks good to me!

I defer to you guys about the depth of error reporting, since it depends on what you would need to understand what happened on the user side. Prior to this PR, we had nothing in terms of error information.

igoroctaviano:
@Punzo now I'm using the onerror event for other unhandled errors but it won't catch those inside catch blocks, this way i had to leave the LoggerService calls close to the notifications.

Punzo:
> @Punzo now I'm using the onerror event for other unhandled errors but it won't catch those inside catch blocks, this way i had to leave the LoggerService calls close to the notifications.

ok, thanks! we can merge this and @fedorov can test. For @fedorov now the logger should catch the majority of the errors, but some errors from third-party libraries will still not be catched (am I right Igor?). E.g.: the current error Cannot read property 'ReferencedSegmentNumber' of undefined from dcmjs
image

Punzo (approved):

Punzo:
> > @Punzo now I'm using the onerror event for other unhandled errors but it won't catch those inside catch blocks, this way i had to leave the LoggerService calls close to the notifications.
> 
> ok, thanks! I can still get this error: the current error Cannot read property 'ReferencedSegmentNumber' of undefined from dcmjs
> image
> 
> but the logger is not catching it. Should the logger catch it?

@igoroctaviano fixed this in last commit https://github.com/OHIF/Viewers/pull/2283/commits/80d8870df9c377308a15c770cc71841f160db028 . @igoroctaviano please merge, thanks!

Issue #2288, opened 2021-03-03
RTStruct tab not visible on the IDC viewer production tier


madelyngreyes:
### Describe the Bug

On the IDC portal when selecting a study with RTStruct data visible the tab to view RTStruct data does not appear.

### What steps can we follow to reproduce the bug?

1. open the explore page on the IDC
2. open a study with RTStruct data
3. No RTStruct tab is visible on the right hand side 

<img width="1429" alt="Screen Shot 2021-03-03 at 10 38 51 AM" src="https://user-images.githubusercontent.com/16824894/109832842-c0be1080-7c0e-11eb-919f-2de1bdc5b547.png">

igoroctaviano:
@pieper @fedorov looks like this is related to https://github.com/dcmjs-org/dcmjs/issues/177
This ticket can be revisited after a new deployment. I'll comment here when it's available in the sandbox.

igoroctaviano:
@Punzo can you please review this one? https://github.com/OHIF/Viewers/pull/2307

fedorov:
Confirmed working in 4.9.7 in test tier.

PR #2307, opened 2021-03-11
IDC-2288: Update toolbar tabs after derived displayset loaded

igoroctaviano:
- The ToolbarRow calls the toolbar module isDisabled function to get the state for an extension's panel tab. Now the isDisabled function depends on async data to be interpreted and used to decide whether the tab is displayed (e.g. check if this study has rt structs).
- Update ToolbarRow component so that it keeps track of derived display sets being loaded.
- Resolves issue #2288 

## Improvements
- Implement a pubsub utility class that studyMetadata could extend to trigger DERIVED_DISPLAY_SETS_CHANGED events that parts of the app could subscribe to update its content.


Punzo (approved):
LGTM!

Issue #2360, opened 2021-04-14
TypeError: windowCenter.toFixed is not a function


xeonfeng:
### Describe the Bug

When using lastest version Viewer, opening a study,and drag the mouse with defalut level icon by pressed and holding the left mouse button, the error occurred!

![image](https://user-images.githubusercontent.com/11417207/114640553-f3c1ec80-9d02-11eb-9ec6-dbadd20a9f7f.png)
![image](https://user-images.githubusercontent.com/11417207/114640617-1522d880-9d03-11eb-8634-9f3eb8a192cc.png)

igoroctaviano:
Fixed in the latest version!

Issue #2297, opened 2021-03-09
windowWidth.toFixed is not a function in viewerMain  [failed to load image]


brikjr:
i have some issues of app in my local host
**TypeError: windowWidth.toFixed is not a function**
and am not changed anything of code just upgrade node.js
![enter image description here](https://i.ibb.co/JFJ0m5p/Screenshot-297.png)

alexiusll:
I have the same question

msibal6:
I have the same problem. Did y'all get this problem by running `yarn run dev`? I worked around it by giving APP_CONFIG some value other than default.j, but I would also like some explanation on what is happening here.

donghakang:
Did anyone find the solution or explanation for this?

donghakang:
According to issue #2138 there is a work around.

I did what @gccgomes and @bianyueyan suggested. I am pretty sure it is not a fix but it is a workaround for CT images

brikjr:
> I have the same problem. Did y'all get this problem by running `yarn run dev`? I worked around it by giving APP_CONFIG some value other than default.j, but I would also like some explanation on what is happening here.

i think the problem come from upgrading some functions on cornerstone library , problem show in CT ,  CT/PT images

donghakang:
> > I have the same problem. Did y'all get this problem by running `yarn run dev`? I worked around it by giving APP_CONFIG some value other than default.j, but I would also like some explanation on what is happening here.
> 
> i think the problem come from upgrading some functions on cornerstone library , problem show in CT , CT/PT images

How about other image files like DICOM? Did you get them to work?

brikjr:
just SM images work 
after changing `var wwwc = "W: ".concat(windowWidth.toFixed(0), " L: ").concat(windowCenter.toFixed(0));
into
var wwwc = "W: ".concat(Number(windowWidth).toFixed(0), " L: ").concat(Number(windowCenter).toFixed(0));var wwwc = "W: ".concat(windowWidth.toFixed(0), " L: ").concat(windowCenter.toFixed(0));
into
var wwwc = "W: ".concat(Number(windowWidth).toFixed(0), " L: ").concat(Number(windowCenter).toFixed(0));`

in the ./node_modules/react-cornerstone-viewport/dist/index.umd.js.

brikjr:
> > > I have the same problem. Did y'all get this problem by running `yarn run dev`? I worked around it by giving APP_CONFIG some value other than default.j, but I would also like some explanation on what is happening here.
[...]

Issue #2138, opened 2020-10-30
OHIF error when opening studies


marcopin:
I'm using lastest ohif viewer with docker (ohif/viewer:lastest), and I have configured to read my local images (dcm4chee). But when I try to open more of my studies I get this message:

Something went wrong in ViewerMain
ViewerMain: windowWidth.toFixed is not a function

<img width="1204" alt="Captura de Pantalla 2020-10-29 a la(s) 8 49 11 p  m" src="https://user-images.githubusercontent.com/3383909/97647291-3ed39600-1a28-11eb-865d-e1051e99ba0e.png">

And no images is shown.

marcopin:
Additional: Stack Trace:

in ViewportOverlay
    in div
    in div
    in CornerstoneViewport
    in ConnectFunction
    in OHIFCornerstoneViewport
    in Suspense
    in OHIFCornerstoneViewport
    in ExtendedOHIFCornerstoneViewport
    in div
    in ViewportPane
    in div
    in ViewportGrid
    in ConnectFunction
    in div
    in ViewerMain
    in ConnectFunction
    in ErrorBoundary
    in OHIFErrorBoundary
    in ErrorBoundaryDialog
    in div
    in div
    in Viewer
    in WrappedComponent
    in ConnectFunction
    in ViewerRetrieveStudyData
    in ConnectFunction
    in ViewerRouting
    in AsyncComponent
    in ErrorBoundary
    in OHIFErrorBoundary
    in Transition
    in CSSTransition
    in Route
    in OHIFStandaloneViewer
    in ConnectFunction
    in withRouter(Connect(OHIFStandaloneViewer))
    in Unknown
    in Unknown
    in ModalProvider
    in DialogProvider
    in SnackbarProvider
    in Router
    in BrowserRouter
    in I18nextProvider
    in AppProvider
    in Provider
    in ErrorBoundary
    in OHIFErrorBoundary
    in App

marcopin:
Try to open this dicom study from any pacs.

[https://drive.google.com/file/d/1GGzKjVCB7t3ItrzeR96SB7p_f1hb_esL/view?usp=sharing](url)

marcopin:
Study failing (one from others):

[https://drive.google.com/file/d/1GGzKjVCB7t3ItrzeR96SB7p_f1hb_esL/view?usp=sharing](url)

marcopin:
Study failing (one from some others):

[...]

PR #2363, opened 2021-04-15
fix(MetadataProvider): Update metadata provider to resolve issues after dicomweb update/new return value

igoroctaviano:
The MetadataProvider should parse string values from the dataset to valid numbers to avoid parsing issues inside libraries that consume the provider like cornerstone.

- #2360
- #2297
- #2138


Punzo (approved):
LGTM!

Issue #2366, opened 2021-04-15
Incorrect LUT descriptor when applying Palette Color


tkenda:
In the code part:

[https://github.com/OHIF/Viewers/blob/master/platform/core/src/utils/metadataProvider/fetchPaletteColorLookupTableData.js#L114](https://github.com/OHIF/Viewers/blob/master/platform/core/src/utils/metadataProvider/fetchPaletteColorLookupTableData.js#L114)

```
function _getPaletteColor(server, paletteColorLookupTableData, lutDescriptor) {
  const numLutEntries = lutDescriptor[0];
  const bits = lutDescriptor[2];

  const arrayBufferToPaletteColorLUT = arraybuffer => {
    const byteArray = bits === 16 ?
      new Uint16Array(arraybuffer) :
      new Uint8Array(arraybuffer);
    const lut = [];

    for (let i = 0; i < numLutEntries; i++) {
      lut[i] = byteArray[i];
    }

    return lut;
};
```

Following http://dicom.nema.org/medical/Dicom/2018d/output/chtml/part03/sect_C.7.6.3.html#sect_C.7.6.3.1.5, LUT Entries shall be 0 if the table entries value is 2^16. 

If numLutEntries is 0, there is no for loop and return empty lut array.

Solution:

**const numLutEntries = lutDescriptor[0] ? lutDescriptor[0] : 65536;**

If is necessary, I can provide you with some DICOM file examples.

PR #2395, opened 2021-05-03
Update fetchPaletteColorLookupTableData.js

idria:
Opened issue https://github.com/OHIF/Viewers/issues/2366. 

Following http://dicom.nema.org/medical/Dicom/2018d/output/chtml/part03/sect_C.7.6.3.html#sect_C.7.6.3.1.5, LUT Entries shall be 0 if the table entries value is 2^16.

This is often used by Philips Ultrasound systems. In case of an LUT Entries value of 0, must be converted to 2^16.

If numLutEntries is 0, there is no for loop and return empty lut array.

Solution:

const numLutEntries = lutDescriptor[0] ? lutDescriptor[0] : 65536;


swederik:
Thanks!

Issue #2259, opened 2021-01-28
Mechanism to communicate to the users inconsistencies within a single series


fedorov:
Example series that motivate some discussion what should be done:

*  https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc_tcia_dev/dicomStores/idc_tcia/study/1.3.6.1.4.1.14519.5.2.1.8162.4009.498881901170332065029863238265 - first instance is a localizer
* https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc_tcia_dev/dicomStores/idc_tcia/study/1.3.6.1.4.1.14519.5.2.1.8162.4009.498881901170332065029863238265?seriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.8162.4009.329297441360267384643622758743-  one series has 2 subvolumes with different slice and pixel spacing

At the moment, OHIF Viewer does not have any indication about those inconsistencies, and presents the slices as if they belong to a single continuous volume.

![image](https://user-images.githubusercontent.com/313942/106286951-39970a80-6214-11eb-893a-d83ccc7272d5.png)

Punzo:
have a little icon on the series to have a dialog overlay text box that reports any issue in the data (different spacing, etc....).
Tell the user also if you can have MPR button or not and why.

Punzo:
PR in review https://github.com/OHIF/Viewers/pull/2331

fedorov:
@Punzo the series in this study

![image](https://user-images.githubusercontent.com/313942/113185008-e6e1cb00-9223-11eb-9c4f-89cd360af456.png)

https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc/dicomStores/v2/study/1.3.6.1.4.1.14519.5.2.1.7695.1700.320115820687123759708360318530

is indicated as having inconsistent spacing, but I do not see those inconsistencies. Maybe a rounding issue, or something else?

![image](https://user-images.githubusercontent.com/313942/113185067-fc56f500-9223-11eb-8c3d-c028b3557612.png)

Punzo:
ok, I will double check the spacing check and the data. I would also expect a tolerance issue (at the moment there is a 10% tolerance).

fedorov:
[...]

PR #2397, opened 2021-05-07
Re #2259: UI refinements for warnings

Punzo:
IDC issue: https://github.com/OHIF/Viewers/issues/2259

- [x] highlight with a larger border the thumbnails images of the current active series (i.e. series in the current active viewport)
- [x]  add the warning icon also on the active viewport as an overlay

IDC Test data:
http://localhost:3000/projects/idc-dev-etl/locations/us-central1/datasets/idc/dicomStores/v2/study/1.3.6.1.4.1.14519.5.2.1.7695.1700.320115820687123759708360318530

http://localhost:3000/projects/idc-tcia/locations/us-central1/datasets/tcia-idc-datareviewcoordination/dicomStores/issue11-lung1-20201015-review/study/1.3.6.1.4.1.32722.99.99.203715003805996641695765332389135385095


Punzo:
@igoroctaviano this is ready for review, for testing you can use http://localhost:3000/projects/idc-dev-etl/locations/us-central1/datasets/idc/dicomStores/v2/study/1.3.6.1.4.1.14519.5.2.1.7695.1700.320115820687123759708360318530 

e.g.: 

Screenshot-20210510112506-1744x1336
Screenshot-20210510112449-1738x1335

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
```diff
@@ -0,0 +1,236 @@
+import { PureComponent } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
+import cornerstone from 'cornerstone-core';
+import './WarningsViewportOverlay.css';
+import moment from 'moment';
+import classNames from 'classnames';
+import { Icon } from '@ohif/ui/src/elements/Icon';
+import { Tooltip } from '@ohif/ui/src/components/tooltip';
+import { OverlayTrigger } from '@ohif/ui/src/components/overlayTrigger';
```
igoroctaviano, line 10: This component doenst fit well in utils, move it to a components folder under cornerstone extension.

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
```diff
@@ -0,0 +1,236 @@
...
+
+function isValidNumber(value) {
+  return typeof value === 'number' && !isNaN(value);
+}
+
+function formatNumberPrecision(number, precision) {
```
igoroctaviano, line 16: All these formatting and checking functions fits well in utils though.

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
```diff
@@ -0,0 +1,236 @@
...
+      lossyImageCompressionRatio,
+      2
+    );
+    return compressionMethod + compressionRatio + ' : 1';
+  }
+
+  return 'Lossless / Uncompressed';
+}
+
+class WarningsViewportOverlay extends PureComponent {
```
igoroctaviano, line 88: Since this will actually be just the same viewport overlay + warnings, I think we should rename it to just OHIFCornerstoneViewportOverlay or something.

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
```diff
@@ -0,0 +1,236 @@
...
+    const frameRate = formatNumberPrecision(1000 / frameTime, 1);
+    const compression = getCompression(imageId);
+    const wwwc = `W: ${
+      windowWidth.toFixed ? windowWidth.toFixed(0) : windowWidth
+    } L: ${windowWidth.toFixed ? windowCenter.toFixed(0) : windowCenter}`;
+    const imageDimensions = `${columns} x ${rows}`;
+
+    const { imageIndex, stackSize } = this.props;
+
+    const warningListOn = warningsList && warningsList.length !== 0 ? true : false;
```
igoroctaviano, line 144: Not sure how big this change will be but it's too late to change warningsList to just warnings or inconsistencyWarnings for clarity?

extensions/dicom-rt/src/getSourceDisplaySet.js
```diff
@@ -1,38 +1,15 @@
-export default function getSourceDisplaySet(studies, rtStructDisplaySet) {
-  const referencedDisplaySet = _getReferencedDisplaySet(
+import { getReferencedDisplaySet } from '../../../platform/core/src/classes/metadata/StudyMetadata.js';
```
igoroctaviano, line 1: We shouldnt have imports like this. You have to export it properly in OHIF core then use import { getRefernecedDisplaySet } from '@ohif/core'

extensions/dicom-segmentation/src/getOHIFDicomSegSopClassHandler.js
```diff
@@ -127,6 +128,8 @@ export default function getSopClassHandlerModule({ servicesManager }) {
             []
           );
         }
+
+        refreshViewports();
```
igoroctaviano, line 132: Why? a get function shouldn't generate any side effects like refreshing cornerstone viewports

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -557,7 +578,25 @@ const _mapStudiesToThumbnails = function(studies) {
 
       const hasWarnings = _checkForSeriesInconsistencesWarnings(displaySet, studies);
 
+      let active = false;
+      if (displaySet.Modality !== 'SEG' && displaySet.Modality !== 'RTSTRUCT') {
```
igoroctaviano, line 582: We sometimes have RTDOSE and in the future, we could possibly support new modalities... I'm ok for now to have these hardcoded checks but we may want to add a TODO to improve it. 

Another thing, can you extract this whole block of code into its own function + add jsdocs?

platform/viewer/src/connectedComponents/Viewer.js
```diff
@@ -197,18 +197,38 @@ class Viewer extends Component {
...
+    const {
+      studies,
+      isStudyLoaded,
+      activeViewportIndex,
+      viewports
+    } = this.props;
+
+    const activeViewport = viewports[activeViewportIndex];
+    const activeDisplaySetInstanceUID =
+      activeViewport ? activeViewport.displaySetInstanceUID : undefined;
```
igoroctaviano, line 219: This change is really necessary? I would try to avoid changes to viewer.js lifecycle hooks. This function is kinda delicate and could potentially decrease the viewer performance or introduce bugs.

If this is necessary for it to work, can you describe why? maybe we can find another alternatives

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
```diff
@@ -0,0 +1,236 @@
...
+* @param {string} strFormat
+*/
+function formatDICOMTime(time, strFormat = 'HH:mm:ss') {
+  return moment(time, 'HH:mm:ss').format(strFormat);
+}
+
+/**
+ * Formats a patient name for display purposes
+ */
+function formatPN(name) {
```
igoroctaviano, line 51: I think you can extract all these utils to a utils file in utils folder and move the component to a components folder

igoroctaviano (changes requested):
Well done sir, let's do another iteration on Monday :)

One thing though, the icon is ginormous, lets decrease it a bit.
Screen Shot 2021-05-14 at 14 38 17
Screen Shot 2021-05-14 at 14 40 11

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
Punzo: sure np, changed to inconsistencyWarnings

extensions/dicom-segmentation/src/getOHIFDicomSegSopClassHandler.js
Punzo: when opening http://localhost:3000/projects/idc-tcia/locations/us-central1/datasets/tcia-idc-datareviewcoordination/dicomStores/issue11-lung1-20201015-review/study/1.3.6.1.4.1.32722.99.99.203715003805996641695765332389135385095 could happen that the seg was not rendered since loadAndCacheDerivedDisplaySets() just load the seg (and all the react segmentation menu elements are still hidden, so they do not refresh the viewport). But you are right the refreshViewports should not be in the laod function of the displaySet, so I moved the refreshViewports into loadAndCacheDerivedDisplaySets function.

platform/viewer/src/connectedComponents/Viewer.js
Punzo: this is necessary since when chaning active viewport or displayset in the current active viewport, we have to recalculate the 'active' property for the thumbnail (i.e. if the border of the thumbnail images has to be highlithed). 

I run perfomance profiles and and I didn't see any change in the perfomance.

I also measured _mapStudiesToThumbnails  
1) the first run (at loading time): some series ~20 millisec, all the other < 1 microsec to run
2) successive runs (i.e. when clicking to change active displayset) : all series < 1 microsec to run

Notice that _mapStudiesToThumbnails essentially runs two functions: 
A) _checkForSeriesInconsistencesWarnings
B) _isDisplaySetActive 

(B) is the one that we care to rerun to compute the 'active' property
(A) Is computed only once (the first time at loading time), because we cache the values at the first time.

So I don't think we will have perfomances issues, I tested also on several other datasets and I did not find issues.

platform/core/src/utils/loadAndCacheDerivedDisplaySets.js
```diff
@@ -108,6 +116,7 @@ async function loadAndCacheDerivedDisplaySets(referencedDisplaySet, studies, log
 
     try {
       await recentDisplaySet.load(referencedDisplaySet, studies);
+      refreshViewports();
```
igoroctaviano, line 119: So the thing is... 

The core package should contain business logic and we should strive as much as possible to align with the current architecture and avoid as much as possible to reference specific libraries here. 

Cornerstone code should live in cornerstone extension, and if this is specific to segmentation, we can trigger an event after loading and caching the derived display set, and then cs extension could refresh by subscribing to this event.

platform/viewer/src/connectedComponents/Viewer.js
Punzo: done see https://github.com/OHIF/Viewers/pull/2397/files#diff-4587d3942badf11b333cbd33f64c30d642840db46b853e146a5a514a92f0176eR548-R593

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
Punzo: done

extensions/dicom-rt/src/getSourceDisplaySet.js
Punzo: done

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
Punzo: done

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
Punzo: done

extensions/cornerstone/src/utils/WarningsViewportOverlay.js
Punzo: done

igoroctaviano (changes requested):
All good, except one point regarding the refreshViewports call. Let me know if you have any questions about it. 

Thanks for updating.

platform/core/src/utils/loadAndCacheDerivedDisplaySets.js
Punzo: I see your point. We actually have already a `extensiondicomsegmentationsegloaded` event everytime a seg is loaded. Thie event is used in the segmentation menu and a `refreshSegmentations` slot method is called. 

The problem is that when you just open the study, the segmentation menu is closed and therefore refreshSegmentations method is never called.

Punzo: at the moment there are some issues realted to the `refreshViewports()` (essentially `cornerstone.updateImage`). 

For segmentation
1) `getSourceDisplaySet` is called in many points, and `getSourceDisplaySet` calls for `setActiveLabelmap` (unless there is the specific flag `activateLabelMap` set to false). `setActiveLabelmap` actually also calls refreshViewports.
2) `getSourceDisplaySet` is used in the core. In `setViewportData` in ViewerMain.js and `mapDispatchToProps` in ConnectedStudyBrowser. And finally we have also the `load` in loadAndCacheDerivedDisplaySets (which needs refresh and triggered this discussion).
3) In extension, in the segmentaionPanel both  refreshViewports() and setActiveLabelmap are called in several palces.

In general cornerstone extension specific code is used both in the core and extensions, i.e.:
image

Punzo:
> All good, except one point regarding the refreshViewports call. Let me know if you have any questions about it.
> 
> Thanks for updating.

Addressed in https://github.com/OHIF/Viewers/pull/2397/commits/f7aff6de985a46683f6100bd91df483225748f9e and re-tested. Merging this.

Issue #2374, opened 2021-04-20
exponential backoff and retry after 500 error


pieper:
As discussed in IDC slack. the proxy server is auto-scaling and sometimes returns 500 errors.  We should take this as an indication that the server needs time to scale and we should not re-issue the request right away.

Might use a package like this: https://www.npmjs.com/package/retry

pieper:
See related discussion for python here: https://github.com/MGHComputationalPathology/dicomweb-client/pull/50

igoroctaviano:
@sedghi do we have exponential backoff or retry capabilities in v3?

Issue #2170, opened 2020-11-16
OHIF keeps undefinetly retrying request after a 400 Bad Request DicomWeb answer


salimkanoun:
Hi there,

I have a series that produce a "Bad Request" answer in my DicomWeb Orthanc Back End (a special dicom file from a manufacturer that is probably not displayable).

This File is asked by OHIF when loading image (see screen capture) but this dicom is not an image so it can't be loaded.
The problem is : OHIF keep continue to reask this URI in an infinit loop, making overload in the server side.
As the HTTP answer is 400 Bad Request, it doesn't make sense to repeat the request without any substantial modification, that's a client error that souldn't be retried identically.

![Capture du 2020-11-16 11-17-53](https://user-images.githubusercontent.com/25229822/99241280-eeb54b80-27fd-11eb-9a0a-1a0167a6efa4.png)

I also put in attachement the responsible dicom (a settings file of a manufacturer that does not contains images but appreas in the series list)

[3589c0cc-4b40-4b71-94be-768e0bdbbc80.dcm.zip](https://github.com/OHIF/Viewers/files/5545612/3589c0cc-4b40-4b71-94be-768e0bdbbc80.dcm.zip)

Best regards,

Salim

sedghi:
@salimkanoun Can you please re-verify this issue if still happens in master?

salimkanoun:
We are OK now, no infinite queries anymore

PR #2408, opened 2021-05-13
IDC-2374: Exponential backoff and retry after 500 error

igoroctaviano:
- DICOMWebClient issue: https://github.com/dcmjs-org/dicomweb-client/pull/36
- Issue: #2374
- Add configurable retry request hook for xhr 
- Use hook for all dicomweb-client instances
- Library used: https://github.com/tim-kos/node-retry


salimkanoun:
Hi,

Is this fix able to use this issue : https://github.com/OHIF/Viewers/issues/2170
I see in the code there is a condition of retry in status code 429 or >500

Are client error code 4XX now removed from retrying ? 

Best regards,

Salim

igoroctaviano:
> Hi,
> 
> Is this fix able to use this issue : #2170
> I see in the code there is a condition of retry in status code 429 or >500
> 
> Are client error code 4XX now removed from retrying ?
> 
> Best regards,
> 
> Salim

I think it's unrelated. dicomweb-client doesn't provide any built-in retry functionality, this PR just introduced it via request hooks which is a new change in dicomweb-client library to allow decoration of its xhr requests. 

These two status codes will be retried by default now but it's up to you to change them in your fork. I'm planning to update this PR to add a better way to config the retry via the app config file.

Punzo:
@igoroctaviano testing this and it looks like it does not find the study list. i.e. starting default server:
image

or also using google config, by opening a study in the url:
image

looks like some undefined variables, could you please have a look? thanks!

Punzo:
@igoroctaviano  now works if using the default config (i.e. server.dcmjs.org  server).

However using google config (i.e. google server), I get 401 errors
image (1)
image

google links in current master works

Issue #2348, opened 2021-04-01
Expected behavior for studies that contain a single SR should be refined


fedorov:
This study contains a single SR series, but the SR series does not show up in the list on the left.

https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc/dicomStores/v2/study/1.3.6.1.4.1.5962.1.1.0.0.0.1563690660.7173.1309

![image](https://user-images.githubusercontent.com/313942/113352620-85485c00-930a-11eb-9dfd-b7e33855e226.png)

igoroctaviano:
@fedorov Currently, the sorting series function had only two supported SR sop classes ((Enhanced SR Storage and Grayscale Softcopy Presentation State Storage) in its array: https://github.com/OHIF/Viewers/blob/2f443fa2efa223ac0fcdf238eee834c6a4ee3fc1/platform/core/src/DICOMSR/utils/findMostRecentStructuredReport.js#L41.
- I tried adding the one in this series (Acquisition Context SR Storage) but now I'm getting errors in dcmjs: Error: This package can currently only interpret DICOM SR TID 1500.

fedorov:
> Currently, the sorting series function had only two supported SR sop classes ((Enhanced SR Storage and Grayscale Softcopy Presentation State Storage) in its array

I think that should be revised. TID 1500 instances can be saved as Comprehensive SR storage, so that is not a good assumption to have.

> **This package can currently only interpret DICOM SR TID 1500.**

I would also think that display of the SR content as text should be supported for any SR, independent of the template. Do you agree? I think the current constraints are too limiting.

igoroctaviano:
> > Currently, the sorting series function had only two supported SR sop classes ((Enhanced SR Storage and Grayscale Softcopy Presentation State Storage) in its array
> 
> I think that should be revised. TID 1500 instances can be saved as Comprehensive SR storage, so that is not a good assumption to have.
> 
> > **This package can currently only interpret DICOM SR TID 1500.**
> 
> I would also think that display of the SR content as text should be supported for any SR, independent of the template. Do you agree? I think the current constraints are too limiting.

@fedorov This dataset contains ContentTemplateSequence.TemplateIdentifier === "8101" and the dcmjs cornerstone adapter can currently only interpret DICOM SR TID 1500. We should update dcmjs cornerstone adapter to interpret 8101?

fedorov:
[...]

PR #2419, opened 2021-05-24
fix(dicom-html): Add parsed dicom meta info section

igoroctaviano:
Resolve parsing issues spotted in #2348.

image

DICOM meta info is currently attached to a naturalized dicom object before is sent to the parsing function which expects only naturalized objects. 

This PR naturalizes the meta object and includes a separate section to display dicom meta info.


Punzo (approved):
LGTM!

PR #2431, opened 2021-06-01
OHIF v3 - Support for OpenID Connect

swederik:
### PR Checklist


extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -4,15 +4,41 @@ import dicomParser from 'dicom-parser';
...
+    },
+  };
+
+  if (!initialized) {
+    cornerstoneWADOImageLoader.webWorkerManager.initialize(config);
+    initialized = true;
+  }
+}
+
+export default function initWADOImageLoader(UserAuthenticationService) {
```
sedghi, line 28: Maybe we expose webworker configs here as well? We discussed `convertFloat32ToInt`

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/getStudiesForPatientByStudyInstanceUID.js
```diff
@@ -16,7 +18,7 @@ async function getStudiesForPatientByStudyInstanceUID(
   // Naturalized DICOM JSON when returning. It makes things like this awkward (mrn)
   if (getStudyResult && getStudyResult.length && getStudyResult[0].mrn) {
     return dataSource.query.studies.search({
-      patientId: getStudyResult[0].mrn,
+      patientId: getStudyResult[0]['00100020'],
```
sedghi, line 21: The check is on `.mrn`, but we are getting it from `00100020`

platform/core/src/extensions/ExtensionManager.js
```diff
@@ -277,13 +277,16 @@ export default class ExtensionManager {
   };
 
   _initDataSourcesModule(extensionModule, extensionId, dataSources = []) {
+    const { UserAuthenticationService } = this._servicesManager.services;
```
sedghi, line 280: minor comment: does userAuthenticationService needs reset or clear upon exit as we do for these
https://bit.ly/3p9zAgt

platform/viewer/src/utils/initWebWorkers.js
```diff
@@ -0,0 +1,22 @@
+import cornerstoneWADOImageLoader from 'cornerstone-wado-image-loader';
+
+let initialized = false;
+
+export default function initWebWorkers() {
+  const config = {
```
sedghi, line 6: I might be missing something, but why do we have 3 separate `initiwebWorkers` for CornerstoneWADOImageLoader (two of them are new)

- here
- extensions/cornerstone/src/init.js
- platform/viewer/src/config.js

sedghi (commented):
Minor comments, looks great otherwise

platform/core/src/extensions/ExtensionManager.js
swederik: I tried this but then navigating between pages (i.e. viewer and worklist) requires the user to log in again.

igoroctaviano:
@swederik @sedghi I think I'm too late but just want to point out that I have used https://github.com/bjerkio/oidc-react (which is based on oidc-client.js that we used in OHIF) for the [SLIM](https://github.com/MGHComputationalPathology/slim) project which simplified/encapsulated a lot of OIDC related codebase.

extensions/cornerstone/src/initWADOImageLoader.js
swederik: I'll do this in another PR

swederik:
Thanks @igoroctaviano - I just removed https://github.com/maxmantz/redux-oidc because we don't use redux for anything else and it wasn't too hard to ditch the dependency. The handful of components for OIDC don't seem like they will be too difficult to maintain. The reason I didn't want to add another library is because this way we can implement login flows other than OpenID Connect using the same UserAuthenticationService (e.g. when people are embedding the app and have their own auth system). That might not matter as much for SLIM though.

Issue #2430, opened 2021-05-31
Download Complete Study via Download Dialog


Christian-e-S:
OHIF is a perfect frontend to the Google Healthcare API, and they both make up a complete PACS solution that is super easy to deploy.

Still, we see that many clinicians (e.g., clinical advisors in clinical studies) want to use their preferred and super-feature-rich (e.g., with fusion capability) locally installed DICOM viewers. 

A common workflow would be:
* Clinical advisor uses OHIF to log in the Web-based PACS (e.g., backed-up by Google Healthcare API)
* Clinical advisor can search/filter contents via OHIF
* Investigate the images online with OHIF capabilities
* When required, the user downloads the entire study of interest to his local machine (maybe even download multiple studies at once!) and switches to his local viewer.
* Over time, OHIF will get more and more features, and users will use their local viewers less and less.

The missing download capability is preventing to get the foot into the doors.

There is a related issue https://github.com/OHIF/Viewers/issues/1130# where a debugging extension has been developed that can be triggered with a shortcut. That is already great, but seems to be targeted on developers and not on end-users. There is no UI and any feedback is only visible in the developer console.

Augmenting the existing solution mentioned above with a simple download dialog, or at least a notification (e.g., "download started") would lift the feature to be acceptable by a standard medical user.

pieper:
Agreed, this would be a nice general purpose features.  Note that for the IDC use case we only wanted this for debugging since we don't want to encourage downloads so restricted the scope to be consistent with our funding.  But this shouldn't prevent others from implementing this, ideally as an optional feature.

For reference here is the command that can be pasted in the developer console to initiate a download:
```
ohif.app.commandsManager.runCommand("downloadAndZip", {listOfUIDs: [window.location.href.split("/").pop()]})
```

Christian-e-S:
> Agreed, this would be a nice general purpose features. Note that for the IDC use case we only wanted this for debugging since we don't want to encourage downloads so restricted the scope to be consistent with our funding. But this shouldn't prevent others from implementing this, ideally as an optional feature.
> 
> For reference here is the command that can be pasted in the developer console to initiate a download:
> 
> ```
> ohif.app.commandsManager.runCommand("downloadAndZip", {listOfUIDs: [window.location.href.split("/").pop()]})
> ```

@pieper:
I have created a pull request for a dedicated study download module.

Compared to the debugging extension, I had to choose a different phase because of the following reasons:

[...]

PR #2454, opened 2021-06-23
Added an extension for downloading an entire study

Christian-e-S:
related issue: https://github.com/OHIF/Viewers/issues/2430#issue-907754649

* User can trigger the download of a complete study, via toolbar button
* User gets detailed status information in a modal view 
* * Individual steps (downloading, zipping, saving)
* * a pull request for the dicomweb-client client has been issued. With this version the user sees also the progress of the download in terms of downloaded bytes (see https://github.com/dcmjs-org/dicomweb-client/pull/44#issue-676285054).

  


igoroctaviano:
Hey @Christian-e-S, great job here :) 🎖️ 

So, I just quickly tested the feature and it seems to be working fine. We have some of this logic in another extension (debugging) but I think it its currently broken, maybe we are better off accepting this separate extension and dumping the old download logic from debugging (what do you think @swederik?)

I have some comments regarding the modal though:
Screen Shot 2021-07-07 at 08 42 11

- When clicking to open the modal, it downloads right away but I think we should add a 'download' button there to avoid clicking the button in the toolbar by mistake and starting a huge downloading request in the background
- The size was not updated after downloading 
- There's no visible loading indicator, I thought it was stuck with a blank status: ...
- You can leverage the notification service to display notifications e.g. 'Study downloaded successfully' or something 
## Extras
- What about adding options in the modal to download display set (series) based or instance-based (we can tackle this in another branch)
- Another extra case that is really not important right now (just thought about it): Adding an option to disable and enable a save button in the viewport overlay so users can download the active viewport

igoroctaviano (changes requested):
Great job 🎖️  I just left a comment, let's discuss it first 😄

Christian-e-S:
Hey @igoroctaviano!

Thank you very much for your review!

Please find my comment below:

> Hey @Christian-e-S, great job here :) 🎖️
> 
> So, I just quickly tested the feature and it seems to be working fine. We have some of this logic in another extension (debugging) but I think it its currently broken, maybe we are better off accepting this separate extension and dumping the old download logic from debugging (what do you think @swederik?)
> 
> I have some comments regarding the modal though:
> Screen Shot 2021-07-07 at 08 42 11
> 
> * I think toolbar button should not be enabled in master by default (@pieper @swederik do we have liability concerns?)

I agree, this makes absolutely sense!

> * When clicking to open the modal, it downloads right away but I think we should add a 'download' button there to avoid clicking the button in the toolbar by mistake and starting a huge downloading request in the background

We had it like this, but users complained about the additional click. In some use cases, users use the online viewer just for previewing images, and then download the series to be able to use the specific features of the tool that they are used to, so downloading becomes their major use case and they want to have it efficient.

> * The size was not updated after downloading

This works only with a recent commit of the dicomweb-client https://github.com/dcmjs-org/dicomweb-client/commit/c292726a421cd84e6a34aec4e72fbbad3c87621c that was created from my pull request to the dicomweb-client repository. This is required to get progress information in terms of bytes downloaded.

Should we wait until this gets into the next release of dicomweb-client?

image


> * There's no visible loading indicator, I thought it was stuck with a blank status: ...

See comment above. With the latest master branch of the dicomweb-client you can see the transferred bytes progressing, during the download, which gives a good indication:

image

image



> * You can leverage the notification service to display notifications e.g. 'Study downloaded successfully' or something

Agreed

> 
> ## Extras
> * What about adding options in the modal to download display set (series) based or instance-based (we can tackle this in another branch)
> * Another extra case that is really not important right now (just thought about it): Adding an option to disable and enable a save button in the viewport overlay so users can download the active viewport

I like these ideas.

sedghi:
@igoroctaviano what is the status on this?

igoroctaviano:
@sedghi codewise LGTM but we need to check if we want to merge it (I know that there are some implications to having an easy download button in a viewer).

sedghi:

threehappypenguins:
> @Christian-e-S , do you have this extension to v3? rgds and great job.

Did you ever try this with v3? If so, how did you go about it? I don't even know how to get started with adding the extension.

sedghi:
@threehappypenguins No we haven't done it in v3

sedghi:

PR #2465, opened 2021-07-05
Extensibility of OHIF with config-point service

wayfarer3130:
This is just a preliminary discussion point for the extensibility service.  Expect this to take a fair bit more development before considering including this.


platform/ui/src/components/StudyListPagination/StudyListPagination.js
```diff
@@ -2,27 +2,28 @@ import React, { useState } from 'react';
...
   };
 
-  const ranges = [
-    { value: '25', label: '25' },
-    { value: '50', label: '50' },
-    { value: '100', label: '100' },
-  ];
-  const [selectedRange, setSelectedRange] = useState(ranges.find(r => r.value === perPage));
-  const onSelectedRange = (selectedRange) => {
+  extensibility = extensibility || StudyListPaginationLevel;
```
wayfarer3130, line 21: @swederik As you requested, I've changed the parameter to be a passed in parameter, rather than accessing this.  I think that actually works really well, because the parameter doesn't depend at all on it being from the extensibility service, but is really just a bag of configuration items that can be whatever the parent chooses to pass in.  The only use of the actual service here is to register a default version which comes from the extensibility service.

wayfarer3130:
@swederik 
I've minimized the changes in the pull request quite a bit.  Could you take a look and see if you think this is small enough to include or if there are still things I should do?  I can also send you a link to the config-point library pages if you want to look at that side as well.

Issue #1215, opened 2019-11-22
Add DICOM SR Support for additional/common annotations


dannyrb:
I think on first pass we should support:

- Length (done)
- Bidirectional (WIP)
- arrow annotation (…is this doable, could be a single point + a descriptive label?)
- Probe (control point)
- Ellipse
- Rectangle

@pieper has kindly provided some guidance on how to approach these mappings in a discussion with @galelis, @JamesAPetts, and @dannyrb 

@galelis, if you transition off this issue, please provide insights/next-steps in a follow-up comment.

dclunie:
There are 2D (image) relative coordinates in DICOM SR, and 3D (frame of reference e.g., patient, volume, slide).

All of Graphic Types specified for those that are co-planar with a referenced image orientation should be supported (3D types that are not in the same plane as the image would require volume "awareness", which may be out of scope for now):

2D: POINT, MULTIPOINT, POLYLINE, CIRCLE, ELLIPSE
3D: POINT, MULTIPOINT, POLYLINE, POLYGON, ELLIPSE

See:

http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.18.6.html
http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.18.9.html

In DICOM SR TID 1500 (which is the template to focus on), see examples of use in these templates:

http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1410
http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1411
http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_320

Examples at:

http://dicom.nema.org/medical/dicom/current/output/chtml/part17/sect_RRR.5.html

David

PS. Note that "generic" SR annotation rendering (as opposed to authoring) code may theoretically "ignore" what template is defined to be or appears to be used, and just search the tree for any SCOORD or SCOORD3D content item, and then explore the local neighborhood for children that are images, and parents that describe its context (i.e., any associated description, identification, measurements, etc.). That approach is robust and can support all sorts of SR templates and applications from the generic TID 1500 to Mammography CAD, as well as private templates.

dclunie:
Agree "arrows" should be single point.

JamesAPetts:
Thanks the input David, I thought as much for the arrow.

[...]

PR #2469, opened 2021-07-08
Add DICOM SR Support for additional/common annotations #1215

igoroctaviano:


fedorov:
We have some of this implemented in v2 and will definitely need it in v3 for IDC.

PR #2470, opened 2021-07-11
Added the Korean (kr) language support

WoonchanCho:
Hi team,

Thank you for all your efforts for implementing this create viewer.

I want to add the the Korean locales to the viewer.

If you prefer letting me to work on the Locize, you can create the space for the Korean language and let me know too.

Thank you in advance for your support.


sedghi (changes requested):
I'm not familiar with the Korean language, but in Locize `ko-KR` seems to be the only option. Is this the same as `kr` that you have used?

We use Locize to maintain the translations. You can read more [here](https://v3-docs.ohif.org/platform/internationalization/).

Can you change your PR according to `ko-KR`, since all files will get overwritten by Locize ... 



image

platform/viewer/.webpack/webpack.pwa.js
```diff
@@ -41,6 +41,13 @@ module.exports = (env, argv) => {
       path: DIST_DIR,
       filename: isProdBuild ? '[name].bundle.[chunkhash].js' : '[name].js',
       publicPath: PUBLIC_URL, // Used by HtmlWebPackPlugin for asset prefix
+      devtoolModuleFilenameTemplate: function(info) {
```
sedghi, line 44: why?

sedghi:

sedghi:

Issue #2482, opened 2021-07-23
Open SEG and RTSTRUCT side panel automatically


pieper:
## Request

When loading or dragging in a seg or rtstruct the change in the image may be subtle and people may not realize anything has happened (real users have reported confusion on this).  Currently one needs to know that opening the right side panel will show the list of segments/structures.

The request would be to either open the panel whenever it gets populated, or at least make it visually change, perhaps with a little badge telling the number of segments/structures that are available.  The badge option may be preferable is people don't want the side panel re-opening every time and they need to repeatedly close it to see the images.  Also if it's possible that there are both structs and segments in the same study then it would be more clear that there is data in one or both and avoid ambiguity over which one to open.
...

**Why should we prioritize this feature?**

Current behavior has been shown to be confusing.
...

pieper:
@igoroctaviano is this something you could have a look at?

igoroctaviano:
> @igoroctaviano is this something you could have a look at?

Sure, I'll do that asap. The optional badge in the panel tab makes sense.

PR #2491, opened 2021-07-29
OHIF-2482: Add badge numbers for seg and rtstruct

igoroctaviano:
#2482 

little badge telling the number of segments/structures that are available. The badge option may be preferable is people don't want the side panel re-opening every time and they need to repeatedly close it to see the images. Also if it's possible that there are both structs and segments in the same study then it would be more clear that there is data in one or both and avoid ambiguity over which one to open.

Screen Shot 2021-07-28 at 22 16 05
Screen Shot 2021-07-28 at 22 15 17


Punzo (approved):
LGTM!

Issue #2493, opened 2021-07-29
Annotation dialog is not getting autofocus (for v2 React.js)


PoyangLiu:
### Describe the Bug

When using the annotation tool, the annotation name dialog pops up but the input field is not automatically focused as expected.  This is the same bug as #183 (but for React.js v2)

### What steps can we follow to reproduce the bug?

1. Open some file in the viewer
2. Select the annotation tool
3. Drag and drop over the image

Notice that the dialog appears, but the input field doesn't receive the focus

### Propose Fix

In [platform/ui/src/components/simpleDialog/SimpleDialog.js line 40](https://github.com/OHIF/Viewers/blob/master/platform/ui/src/components/simpleDialog/SimpleDialog.js#L40), add the property `autofocus`
```js
      <div className="InputDialog">
        <SimpleDialog
          headerTitle={title}
          onClose={onClose}
          onConfirm={onSubmitHandler}
        >
          <TextInput
            autoFocus <----- add this
            type="text"
            value={value}
            onChange={event => setValue(event.target.value)}
            label={label}
          />
        </SimpleDialog>
      </div>
```

sedghi:
Please review the latest code in the master branch. I am confident that this issue has been resolved. If it persists, kindly reopen the issue with updated details. 

Try viewer-dev.ohif.org instead of viewer.ohif.org
Our `viewer.ohif.org` is deployed from `release` branch while `viewer-dev.ohif.org` is our `master ` branch
Read more about branch explanations here https://docs.ohif.org/development/getting-started#developing

PR #2494, opened 2021-07-29
fix: add autoFocus proeprty to TextInput control

PoyangLiu:
@swederik, @dannyrb 
This PR addresses #2493, where annotation tool dialog text input doesn't get autofocus when the dialog is shown.
The fix adds `autoFocus` attribute to the `TextInput` component, which then gets passed down to the input element.


sedghi (approved):
This looks good to me for merge. ~~Can you rebase and push so that the netlify triggers again? (previously there was a problem with your netlify deploy which is now fixed)~~

PR #2499, opened 2021-08-03
Static WADO display changes

wayfarer3130:
Contains a set of changes to allow using the static wado output.  Mostly this is a configuration definition, plus a bunch of small changes to allow null series description and set the frame number correctly in a couple of spots.


wayfarer3130:
@swederik These changes are needed to display some of the studies.  They are on top of the config point changes, which are assuming to be merged first.

wayfarer3130:
Hi @sedghi - can you look over the changes here?  I will start annotating them as there are a number of changes in this branch now.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -15,6 +15,7 @@ import {
   retrieveStudyMetadata,
   deleteStudyMetadataPromise,
 } from './retrieveStudyMetadata.js';
+import StaticWadoClient from './utils/StaticWadoClient.js';
```
wayfarer3130, line 18: That static wado stuff includes a bunch of search functionality to allow sub-selecting the returned studies results.  It may well be something we want to have supported when queries are done against servers which are less than robust in terms of search functionality.  Alternatively, it could simply be left out for now.

extensions/default/src/DicomWebDataSource/utils/getWADORSImageId.js
```diff
@@ -6,7 +6,7 @@ function buildInstanceWadoRsUri(instance, config) {
 function buildInstanceFrameWadoRsUri(instance, config, frame) {
   const baseWadoRsUri = buildInstanceWadoRsUri(instance, config);
 
-  frame = frame != null || 1;
+  frame = frame || 1;
```
wayfarer3130, line 9: This had been returning "true" for the frame number, which meant that frames couldn't be navigated to.

extensions/default/src/Panels/PanelStudyBrowser.jsx
```diff
@@ -165,11 +165,11 @@ function PanelStudyBrowser({
     );
     const updatedExpandedStudyInstanceUIDs = shouldCollapseStudy
       ? // eslint-disable-next-line prettier/prettier
-        [
```
wayfarer3130, line 168: My understanding is that the prettier plugin is supposed to be run - the spacing changes are caused by that.

extensions/default/src/Panels/PanelStudyBrowser.jsx
```diff
@@ -247,7 +247,7 @@ function _mapDisplaySets(displaySets, thumbnailImageSrcMap) {
 
     array.push({
       displaySetInstanceUID: ds.displaySetInstanceUID,
-      description: ds.SeriesDescription,
+      description: ds.SeriesDescription || 'Unspecified',
```
wayfarer3130, line 250: Series description is sometimes null, and that was causing an exception rendering the thumbnails.

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/getStudiesForPatientByStudyInstanceUID.js
```diff
@@ -20,6 +20,9 @@ async function getStudiesForPatientByStudyInstanceUID(
     return dataSource.query.studies.search({
       patientId: getStudyResult[0].mrn,
     });
+  } else {
```
wayfarer3130, line 23: mrn is sometimes null for anonymized data - I agree it is type 1, but some sample data has it that way, so support it.

platform/ui/src/assets/styles/styles.css
```diff
@@ -1,6 +1,6 @@
 /* CUSTOM OHIF SCROLLBAR */
 .ohif-scrollbar {
-  scrollbar-color: #0944b3 transparent;
+  scrollbar-color: #3E5967 transparent;
```
wayfarer3130, line 3: These are the requested changes for the scrollbar colours.

platform/ui/src/components/StudyListPagination/StudyListPagination.js
```diff
@@ -2,27 +2,28 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Button, ButtonGroup, Typography, Select } from '../';
 import { useTranslation } from 'react-i18next';
+import ConfigPoint from 'config-point';
 
-const StudyListPagination = ({
+export function StudyListPagination({
```
wayfarer3130, line 7: Shows a simple config-point extension for this item.  That allows customizing the pagination items.

platform/viewer/public/config/aws_static.js
```diff
@@ -0,0 +1,159 @@
+window.config = {
```
wayfarer3130, line 1: This is a copy of the default.js configuration.  Generally, this monolithic configuration isn't a great idea, particularly mixing in data sources etc.  Need to figure out some better options for this longer term.

platform/viewer/src/routes/DataSourceWrapper.jsx
```diff
@@ -49,6 +49,7 @@ function DataSourceWrapper(props) {
     total: 0,
     resultsPerPage: 25,
     pageNumber: 1,
+    location: 'PERFORM_INITIAL_QUERY',
```
wayfarer3130, line 52: The changes in this file prevent endlessly querying the studies when no results are found, and cause the actual query to be done whenever the query parameters change.  That means the query is done correctly initially, as the currently querying results are specified in "location".  There should be a query update prevention when there is an outstanding query, and the parameters change before it comes back.

platform/viewer/src/routes/WorkList/StudyListConfig.jsx
```diff
@@ -0,0 +1,218 @@
+import React, { useState, useEffect } from 'react';
```
wayfarer3130, line 1: Added a new file to specify the study list portion of the worklist.  It is configurable by modifying the config-point stuff to do things like adding or removing coluumns, defining new columns etc.

platform/viewer/src/routes/WorkList/WorkList.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState, useEffect } from 'react';
```
wayfarer3130, line 1: Changes here are to split out the study list portion of this file, and then use that study list configuration to actually generate the study list part.
Longer term, fully configuring the child elements of the study list would also be a good idea.  That would allow alternative querying strategies such as querying FHIR ImagingDocument instead, or splitting it up into patient/study lists rather than study/series lists (or perhaps both)

.gitignore
```diff
@@ -23,6 +23,7 @@ package-lock.json
 yarn-error.log
 .DS_Store
 .env
+*.code-workspace
```
sedghi, line 26: Why do we need this?

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -64,7 +66,7 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
 
   // TODO -> Two clients sucks, but its better than 1000.
   // TODO -> We'll need to merge auth later.
-  const qidoDicomWebClient = new api.DICOMwebClient(qidoConfig);
+  const qidoDicomWebClient = new StaticWadoClient(qidoConfig);
```
sedghi, line 69: I understand `StaticWadoClient` extends the `api.DICOMWebClient` and returns early if no staticWado config; however, I would still prefer (for readability) simpler if statement that chooses the client based on `staticWado` like 

`const qidoDicomWebClient = staticWado ? new StaticWadoClient(...) : api.DICOMwebClient(...)`

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.js
```diff
@@ -0,0 +1,53 @@
...
+    this.staticWado = qidoConfig.staticWado;
+  }
+
+  async searchForStudies(options) {
+    if (!this.staticWado) return super.searchForStudies(options);
+
+    let searchResult = await super.searchForStudies(options);
+    const { queryParams } = options;
+    if (!queryParams) return searchResult;
+    // console.log('Query params', queryParams);
```
sedghi, line 30: delete

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.js
```diff
@@ -0,0 +1,53 @@
...
+    const filtered = searchResult.filter(study => {
+      for (const key of Object.keys(StaticWadoClient.filterKeys)) {
+        if (!this.filterItem(key, queryParams, study)) return false;
+      }
+      return true;
+    });
+    // console.log("Searching - filtered response has", filtered.length, "items in it:", filtered);
```
sedghi, line 37: delete

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.js
```diff
@@ -0,0 +1,53 @@
...
+
+  filterItem(key, queryParams, study) {
+    const altKey = StaticWadoClient.filterKeys[key] || key;
+    if (!queryParams) return true;
+    const testValue = queryParams[key] || queryParams[altKey];
+    if (!testValue) return true;
+    const valueElem = study[key] || study[altKey];
+    if (!valueElem) {
+      return false;
+    }
```
sedghi, line 49: maybe we stick to the same if () return .... (same line) in the same file?

extensions/default/src/DicomWebDataSource/utils/getWADORSImageId.js
sedghi: So, multiframe was broken?
Does `frame` return `null` or `undefined`? This line returns 1 when frame is both `undefined` OR `null`.

extensions/default/src/Panels/PanelStudyBrowser.jsx
sedghi: Can you do so for `extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.jsx` as well?

sedghi: Reading the next comment (inside `getSopClassHandlerModule` I don't understand why we need `Unspecified` again here. Wouldn't `makeDisplaySet` handles this?

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/getStudiesForPatientByStudyInstanceUID.js
sedghi: We don't need an `else` here

```js
if (getStudyResult && getStudyResult.length && getStudyResult[0].mrn) {
    return dataSource.query.studies.search({
      patientId: getStudyResult[0].mrn,
   });
}
 
console.log('No mrn found for', getStudyResult);
return getStudyResult;

}
```

sedghi: Also it shouldn't console.log `getStudyResult` as it contains whole study, maybe just UID?

sedghi: Also we have another one here `extensions/default/src/Panels/getStudiesForPatientByStudyInstanceUID.js`

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
```diff
@@ -257,7 +257,7 @@ function TrackedCornerstoneViewport(props) {
           isLocked: false,
           isRehydratable: false,
           studyDate: formatDate(SeriesDate), // TODO: This is series date. Is that ok?
-          currentSeries: SeriesNumber,
+          currentSeries: SeriesNumber || 0,  // TODO - switch entire currentSeries to be UID based or actual position based
```
sedghi, line 260: Can't we do this at the displaySet creation level?

package.json
```diff
@@ -19,16 +19,18 @@
...
     "build:demo": "lerna run build:viewer:demo --stream",
     "dev": "lerna run dev:viewer --stream",
     "dev:project": ".scripts/dev.sh",
     "dev:orthanc": "lerna run dev:orthanc --stream",
+    "dev:dcm4chee": "lerna run dev:dcm4chee --stream",
+    "dev:static": "lerna run dev:static --stream",
     "orthanc:up": "docker-compose -f .docker/Nginx-Orthanc/docker-compose.yml up",
-    "start": "yarn run dev",
-    "test": "yarn run test:unit",
+    "start": "yarn run dev",    "test": "yarn run test:unit",
```
sedghi, line 33: new line

platform/ui/src/components/StudyBrowser/StudyBrowser.jsx
```diff
@@ -105,7 +105,7 @@ const StudyBrowser = ({
           })}
         </ButtonGroup>
       </div>
-      <div className="flex flex-col flex-1 overflow-auto invisible-scrollbar">
+      <div className="flex flex-col flex-1 overflow-auto ohif-scrollbar">
```
sedghi, line 108: I'm not sure if this is intended by design... I'll ask

platform/viewer/src/routes/DataSourceWrapper.jsx
sedghi: I need more explanation for changes on this file to review it. I'll send a message

extensions/default/src/DicomWebDataSource/utils/getWADORSImageId.js
wayfarer3130: Yes, the multiframe was always requesting  .../frames/true  which seems to work for some DICOMweb implementations, but  not the static-wado one.

extensions/default/src/Panels/PanelStudyBrowser.jsx
wayfarer3130: Maybe it should, but I was seeing exceptions getting thrown.  Maybe it should be in getSopClassHandlerModule to put it in a single place, but I would need a bit of help figuring out exactly where.

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/getStudiesForPatientByStudyInstanceUID.js
wayfarer3130: I know there is a second one - I didn't tweak that one as the behaviour is different enough I wasn't quite sure how to.

platform/viewer/src/routes/DataSourceWrapper.jsx
wayfarer3130: Provided offline

sedghi:
Minor comments, looks great otherwise

.gitignore
wayfarer3130: Only needed if you happen to store a VS Code workspace in the directory, which I tend to do  regularly.  Could be private, but I don't expect it will be that uncommon.

extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.js
wayfarer3130: Done at the display set creation level now.

package.json
wayfarer3130: Fixed

extensions/default/src/getSopClassHandlerModule.js
```diff
@@ -20,9 +20,9 @@ const makeDisplaySet = instances => {
     SeriesTime: instance.SeriesTime,
     SeriesInstanceUID: instance.SeriesInstanceUID,
     StudyInstanceUID: instance.StudyInstanceUID,
-    SeriesNumber: instance.SeriesNumber,
+    SeriesNumber: instance.SeriesNumber || 0,
     FrameRate: instance.FrameTime,
-    SeriesDescription: instance.SeriesDescription,
+    SeriesDescription: instance.SeriesDescription || 'Unspecified',
```
swederik, line 25: probably needs a translation string

extensions/default/src/getSopClassHandlerModule.js
```diff
@@ -20,9 +20,9 @@ const makeDisplaySet = instances => {
     SeriesTime: instance.SeriesTime,
     SeriesInstanceUID: instance.SeriesInstanceUID,
     StudyInstanceUID: instance.StudyInstanceUID,
-    SeriesNumber: instance.SeriesNumber,
+    SeriesNumber: instance.SeriesNumber || 0,
```
swederik, line 23: can't it just be left undefined?

swederik (approved):

sedghi (approved):

extensions/default/src/getSopClassHandlerModule.js
wayfarer3130: No, undefined throws an exception if one accesses something by position.

extensions/default/src/getSopClassHandlerModule.js
wayfarer3130: What about just using an empty string?  That is what is supposed to come back?  I'm going to change it to empty.

platform/viewer/cypress/integration/study-list/OHIFStudyList.spec.js
```diff
@@ -21,31 +21,42 @@ describe('OHIF Study List', function() {
       //cy.get('@modalities')
     });
 
-    it('searches Patient Name with exact string', function() {
+    it('Displays several studies initially', function () {
```
wayfarer3130, line 24: There wasn't an initial check previously to ensure that the base state was correct.  This check will fail in previous versions on the length > 1

platform/viewer/cypress/integration/study-list/OHIFStudyList.spec.js
```diff
@@ -21,31 +21,42 @@ describe('OHIF Study List', function() {
...
+        expect($list).to.contain('832040');
+      });
+    });
+
+    it('searches Patient Name with exact string', function () {
       cy.get('@PatientName').type('Juno');
       //Wait result list to be displayed
       cy.waitStudyList();
-      cy.get('@searchResult').should($list => {
+      cy.get('@searchResult2').should($list => {
```
wayfarer3130, line 37: The correct value to read is searchResult2, which returns the correct list of tr items being found

platform/viewer/cypress/integration/study-list/OHIFStudyList.spec.js
```diff
@@ -21,31 +21,42 @@ describe('OHIF Study List', function() {
...
         expect($list.length).to.be.eq(1);
         expect($list).to.contain('Juno');
       });
     });
 
-    it('searches MRN with exact string', function() {
+    /* TODO - re-enable this once the test server is fixed to allow searching by mrn
```
wayfarer3130, line 43: The MRN query is commented out because querying by the hex code doesn't work in the test server setup.
This ALSO means that the actual results come back incorrectly.  Hmm, let me add a tweak to this to allow it to work.

platform/viewer/cypress/integration/study-list/OHIFStudyList.spec.js
sedghi: nice

platform/viewer/cypress/integration/study-list/OHIFStudyList.spec.js
sedghi: great

platform/viewer/cypress/integration/study-list/OHIFStudyList.spec.js
sedghi: Makes sense

PR #2566, opened 2021-09-27
Added the static wado client access to OHIF v2

wayfarer3130:
Run against the static-wado for local access.
@swederik - who do you want looking at v2 changes?


wayfarer3130:
@igoroctaviano - here is the first of a bunch of small changes.  I believe the failing checks are race condition ones.

PR #2505, opened 2021-08-05
Expose the Cobb's Angle tool from cornerstone.

dbejarano32:
### PR Checklist

- [ Expose the Cobb's Angle tool from cornerstone.] Brief description of changes
- [ https://github.com/OHIF/Viewers/issues/2399 ] Links to any relevant issues
- [ ] Required status checks are passing
- [ ] `@` a maintainer to request a review

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #2579, opened 2021-10-07
feat(OHIFv2): Added an extension to enable the cobb-angle tool

wayfarer3130:
This is an extension that adds the cobb angle tool from cornerstone into the OHIF v2 version.
@igoroctaviano  - this is for consideration for inclusion int he master branch (OHIF v2)


sedghi (changes requested):
I don't think this should be a separate extension, `extension/corenerstone` is already adding a lot of tools. So we just need to ad `CobbAngleTool` to it

sedghi:
Someone has already done what I asked here https://github.com/OHIF/Viewers/pull/2505
So if that is ok we close this PR for now

wayfarer3130:
Was done as part of 2505 as requested

PR #2597, opened 2021-10-26
Feat/v2 video

wayfarer3130:
This feature change for OHIF v2 (master) adds a video extension to display videos.  The URL for the video must be available as raw video on the DICOMweb /pixeldata endpoint as straight video, and that resource, depending on the browser and video plugin, may need to be cacheable.
To support that, the thumbnails list can now fetch a DICOMweb thumbnail image, and is scrollable.


sedghi (changes requested):
Looks good to me, please see my comments

wayfarer3130:
Decided to only merge this into the v3 OHIF codebase, as there are significant changes to the display set handling in V2.

PR #2600, opened 2021-11-02
Feat/v3 hanging protocol MxN

wayfarer3130:
There were a number of defects in actually creating usable hanging
protocols.  This feature fixes a couple of them, and adds two hanging
protocols in the default theme.json5 file.

@sedghi - I think you are the person knowing something about hanging protocols.  I added a few extra options to the hanging protocols, including the ability to have them site-configured on a per-url basis (eg to allow different users to get different protocols).  They are also somewhat easier to configure than they used to be.  Should I have used stages to manage the hanging protocol setup for this?  Where should I document the changes?

### DIFF
--- a/platform/core/src/services/HangingProtocolService/HPMatcher.js
+++ b/platform/core/src/services/HangingProtocolService/HPMatcher.js
@@ -21,7 +21,7 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
   let score = 0;
 
   rules.forEach(rule => {
-    const attribute = rule.attribute;
+    const { attribute } = rule;
 
     // Do not use the custom attribute from the metadataInstance since it is subject to change
     if (customAttributeRetrievalCallbacks.hasOwnProperty(attribute)) {
@@ -36,7 +36,13 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
 
     // Create a single attribute object to be validated, since metadataInstance is an
     // instance of Metadata (StudyMetadata, SeriesMetadata or InstanceMetadata)
-    const attributeValue = metadataInstance[attribute];
+    let attributeValue = metadataInstance[attribute];
+    if (attributeValue === undefined) {
+      if (attribute === 'NumberOfStudyRelatedSeries') {
+        attributeValue = metadataInstance.series?.length;
+      }
+      // Add other computable values such as modalities in study
+    }
     const attributeMap = {
       [attribute]: attributeValue,
     };
@@ -50,6 +56,7 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
     }
 
     if (!errorMessages) {
+
       // If no errorMessages were returned, then validation passed.
 
       // Add the rule's weight to the total score
--- a/platform/core/src/services/HangingProtocolService/HangingProtocolService.js
+++ b/platform/core/src/services/HangingProtocolService/HangingProtocolService.js
@@ -2,6 +2,8 @@ import cloneDeep from 'lodash.clonedeep';
 import pubSubServiceInterface from '../_shared/pubSubServiceInterface';
 import sortBy from '../../utils/sortBy.js';
 import ProtocolEngine from './ProtocolEngine';
+import ThemeProtocols from './themeProtocolProvider';
+import ConfigPoint from 'config-point';
 
 const EVENTS = {
   STAGE_CHANGE: 'event::hanging_protocol_stage_change',
@@ -35,11 +37,21 @@ class HangingProtocolService {
     Object.assign(this, pubSubServiceInterface);
   }
 
+  listenThemeProtocols(point) {
+
+    this.addProtocols(point.protocols);
+  }
+
   reset() {
     this.studies = [];
     this.protocols = [];
     this.hpAlreadyApplied = [];
     this.matchDetails = [];
+
+    if (!this._listenThemeProtocols) {
+      this._listenThemeProtocols = this.listenThemeProtocols.bind(this);
+    }
+    ConfigPoint.addLoadListener(ThemeProtocols, this._listenThemeProtocols);
     // this.ProtocolEngine.reset()
   }
 
@@ -54,6 +66,24 @@ class HangingProtocolService {
   addProtocols(protocols) {
     protocols.forEach(protocol => {
       if (this.protocols.indexOf(protocol) === -1) {
+        protocol.id = protocol.id || protocol.name;
+        // Automatically compute some number of attributes if they
+        // aren't present.  Makes defining new HPs easier.
+        protocol.name = protocol.name || protocol.id;
+        const { stages } = protocol;
+
+        // Generate viewports automatically as required.
+        stages.forEach(stage => {
+          if (!stage.viewports) {
+            const viewport = stage.viewport || {};
+            stage.viewports = [];
+            const { rows, columns } = stage.viewportStructure.properties;
+
+            for (let i = 0; i < rows * columns; i++) {
+              stage.viewports.push(viewport);
+            }
+          }
+        });
         this.protocols.push(protocol);
       }
     });
@@ -213,9 +243,10 @@ class HangingProtocolService {
     // for each of the viewport matching procedures
 
     // Loop through each viewport
+    const alreadyMatched = [];
     stageModel.viewports.forEach((viewport, viewportIndex) => {
       this.hpAlreadyApplied.push(false);
-      const details = this._matchImages(viewport);
+      const details = this._matchImages(viewport, alreadyMatched);
 
       let currentMatch = details.bestMatch;
 
@@ -227,6 +258,7 @@ class HangingProtocolService {
       // Viewport Settings
       //
       // protocol defined callback
+      viewport.viewportSettings = viewport.viewportSettings || [];
       const protocolCallbacks = viewport.viewportSettings.filter(
         setting => setting.type === VIEWPORT_SETTING_TYPES.PROPS
       );
@@ -286,13 +318,13 @@ class HangingProtocolService {
   }
 
   // Match images given a list of Studies and a Viewport's image matching reqs
-  _matchImages(viewport) {
+  _matchImages(viewport, existingMatches) {
     console.log('ProtocolEngine::matchImages');
 
-    // TODO: matching is applied on study and series level, instance
-    // level matching needs to be added in future
+    // TODO: matching is applied on study and series level,
+    // instance level matching needs to be added in future
 
-    const { studyMatchingRules, seriesMatchingRules } = viewport;
+    const { studyMatchingRules = [], seriesMatchingRules = [] } = viewport;
 
     const matchingScores = [];
     let highestStudyMatchingScore = 0;
@@ -314,7 +346,12 @@ class HangingProtocolService {
 
       highestStudyMatchingScore = studyMatchDetails.score;
 
+      // TODO - use display sets instead of series, as that will use the
+      // correct re-grouped information.
       study.series.forEach(aSeries => {
+        if (existingMatches && existingMatches.indexOf(aSeries) != -1) {
+          return;
+        }
         const seriesMatchDetails = this.ProtocolEngine.findMatch(
           aSeries,
           seriesMatchingRules
@@ -357,6 +394,7 @@ class HangingProtocolService {
           SeriesInstanceUID: aSeries.SeriesInstanceUID,
           matchingScore: totalMatchScore,
           matchDetails: matchDetails,
+          displaySet: aSeries,
           sortingInfo: {
             score: totalMatchScore,
             study: study.StudyInstanceUID,
@@ -387,7 +425,7 @@ class HangingProtocolService {
     );
 
     const bestMatch = matchingScores[0];
-
+    if (bestMatch?.displaySet) existingMatches.push(bestMatch.displaySet);
     console.log('ProtocolEngine::matchImages bestMatch', bestMatch);
 
     return {
--- a/platform/core/src/services/HangingProtocolService/themeProtocolProvider.js
+++ b/platform/core/src/services/HangingProtocolService/themeProtocolProvider.js
@@ -1,0 +1,116 @@
+import ConfigPoint from 'config-point';
+
+const { ThemeProtocols } = ConfigPoint.register({
+  ThemeProtocols: {
+    configBase: {
+      protocols: [
+        {
+          id: '1x2',
+          locked: true,
+          hasUpdatedPriorsInformation: false,
+          name: '1x2',
+          createdDate: '2021-11-01T18:32:42.849Z',
+          modifiedDate: '2021-11-01T18:32:42.849Z',
+          availableTo: {},
+          editableBy: {},
+          protocolMatchingRules: [
+            {
+              id: 'NumberOfStudyRelatedSeries>1',
+              weight: 10,
+              attribute: 'NumberOfStudyRelatedSeries',
+              constraint: {
+                greaterThan: {
+                  value: 1,
+                },
+              },
+              required: true,
+            },
+          ],
+          stages: [
+            {
+              name: 'OneByTwo',
+              viewportStructure: {
+                type: 'grid',
+                properties: {
+                  rows: 1,
+                  columns: 2,
+                },
+              },
+              viewport: {},
+            },
+          ],
+          numberOfPriorsReferenced: 0,
+        },
+        {
+          id: '2x2',
+          locked: true,
+          hasUpdatedPriorsInformation: false,
+          name: '2x2',
+          createdDate: '2021-11-01T18:32:42.849Z',
+          modifiedDate: '2021-11-01T18:32:42.849Z',
+          availableTo: {},
+          editableBy: {},
+          protocolMatchingRules: [
+            {
+              id: 'NumberOfStudyRelatedSeries>2',
+              weight: 20,
+              attribute: 'NumberOfStudyRelatedSeries',
+              constraint: {
+                greaterThan: {
+                  value: 2,
+                },
+              },
+              required: true,
+            },
+          ],
+          stages: [
+            {
+              name: 'TwoByTwo',
+              viewportStructure: {
+                type: 'grid',
+                properties: {
+                  rows: 2,
+                  columns: 2,
+                },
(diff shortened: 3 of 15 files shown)


platform/core/src/services/HangingProtocolService/HPMatcher.js
```diff
@@ -36,7 +36,13 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
 
     // Create a single attribute object to be validated, since metadataInstance is an
     // instance of Metadata (StudyMetadata, SeriesMetadata or InstanceMetadata)
-    const attributeValue = metadataInstance[attribute];
+    let attributeValue = metadataInstance[attribute];
```
wayfarer3130, line 39: The study level information is very sparse, because it doesn't actually start with a study query, so generate some of the study level information automatically.

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -35,11 +36,21 @@ class HangingProtocolService {
...
+    this.addProtocols(point.protocols);
+  }
+
   reset() {
     this.studies = [];
     this.protocols = [];
     this.hpAlreadyApplied = [];
     this.matchDetails = [];
+
+    if (!this._listenThemeProtocols) {
```
wayfarer3130, line 50: Load any protocols specified by the theme properties, which allows setting up custom sets of hanging protocols for specific uses or users - that is, as a user configurable item rather than hard coded in the extensions modules.

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -54,6 +65,21 @@ class HangingProtocolService {
   addProtocols(protocols) {
     protocols.forEach(protocol => {
       if (this.protocols.indexOf(protocol) === -1) {
+        protocol.id = protocol.id || protocol.name;
```
wayfarer3130, line 68: This is just a bunch of code to allow easier configuration of the hanging protocol by defining fewer bits of it.

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -314,7 +342,12 @@ class HangingProtocolService {
 
       highestStudyMatchingScore = studyMatchDetails.score;
 
+      // TODO - use display sets instead of series, as that will use the
+      // correct re-grouped information.
       study.series.forEach(aSeries => {
+        if (existingMatches && existingMatches.indexOf(aSeries) != -1) {
```
wayfarer3130, line 348: Allows matching consecutive series, without re-displaying the same one multiple times.  Need to figure out a good way to allow re-matching an existing instance.  I was thinking of adding viewport rules that will find a previous series again.

sedghi:
I will take a look into this, maybe let's wait until my review and then you add to documentations?

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -314,7 +342,12 @@ class HangingProtocolService {
 
       highestStudyMatchingScore = studyMatchDetails.score;
 
+      // TODO - use display sets instead of series, as that will use the
```
wayfarer3130, line 345: Changing the matching here would allow interaction with the display sets, which means that proper regrouping rules could be defined, and then directly used by the hanging protocols.  That ultimately makes for much easier to define HPs because you don't have to specify both the regrouping rules as well as the hanging rules.  HOwever, compatibility with the existing DICOM HPs should be maintained, so need to think about how to do this properly.

platform/viewer/src/appInit.js
```diff
@@ -17,6 +17,7 @@ import {
   errorHandler
   // utils,
 } from '@ohif/core';
+import ConfigPoint from 'config-point';
```
sedghi, line 20: Taking a quick look, I saw you are using config-point, is this intended or mistakenly put here? since we haven't reviewed the config-point additions before

platform/viewer/public/theme/theme.json5
```diff
@@ -0,0 +1,80 @@
+{
+  ThemeProtocols: {
```
wayfarer3130, line 2: It is possible to define additional hanging protocols for different uses at a site by defining more themes and using theme=MoreHangingProtocols

platform/viewer/public/theme/theme.json5
```diff
@@ -0,0 +1,80 @@
...
+    protocols: [
+      {
+        id: '1x2',
```
wayfarer3130, line 5: Note how this is somewhat easier to define - it has a bunch of stuff defaulted or not required.

platform/viewer/src/appInit.js
```diff
@@ -49,6 +50,14 @@ function appInit(appConfigOrFunc, defaultExtensions) {
     appConfig,
   });
 
+  if (window) {
```
wayfarer3130, line 53: This stuff loads the theme configuration values at startup, defaulting to the theme.json5 is not otherwise specified.

platform/viewer/src/appInit.js
wayfarer3130: It is intended to be used - it allows a site to reasonably define application specific extensions.

sedghi: I don't think this the best approach. This PR currently adds the `config-points` dependencies (which is an external 3rd party package dependency on your Github) to a lot of core code. First of all, I think if we are going ahead with `config-point` which we were going to discuss, and it should have its own PR; for transparency and maintainability, it should be an OHIF package (something like platform/config-point). The same is true for the static-wado stuff which we need to move to @cornerstone (which I think we have talked about it previously).

Hanging protocols are somehow flexible right now, and they can be set for specific user (for instance fetching from a server user-specific HP's), and they can be added in the ModeEnter to the HangingProtocolService (`HangingProtocolService.setHangingProtocol()`), so extensions are not the only place providing the protocols for matching, Mode's can manipulate the HP (and that is the intension of Modes, PT/CT mode: 3x3 layout, Prostate MRI 2x2 layout etc.). What is config-point doing for enhancement of HP stuff?

CC @swederik

wayfarer3130: I'm fine with config-point having it's own PR, but it can't be an OHIF package since it isn't OHIF specific, but is intended to be used in other contexts (in fact I'm already using it elsewhere).  
modes are defined in extensions, so whether the additional HPs are in the mode or are in the base extension, they are still defined elsewhere, and that means modifying code in order to provide them.  The design for ConfigPoint is that it is an easily included configuration system that allows having several variations on the configuration, without needing to include services that are not commonly available, such as the DICOM Hanging Protocol service - which exists in a couple of places but is very cumbersome to use and will mostly require sites adding custom code in order to support it.  config-point just needs static files added, so it meets the same sort of requirements for deployment that you are going to have for OHIF anyways.  The files can be gzipped on disk and served already compressed, they can be edited in any plain text editor and redeployed without rebuilding any of the software.  That is being used already for specific configurations deployed to different teams within a healthcare setting (FU Berlin specifically, different deployments for each department, and for each species being treated).
In terms of use with modes, I actually agree there - the modes should have names/settings for specific use cases, but should also be site definable - not require code level customizations to setup.  If a site wants the PT/CT mode to auto-open a fusion viewport, and then two other viewports, well, that becomes configurable at the site, as long as the underlying code has it supported.

wayfarer3130: I'm fine with config-point having it's own PR, but it can't be an OHIF package since it isn't OHIF specific, but is intended to be used in other contexts (in fact I'm already using it elsewhere).
modes are defined in extensions, so whether the additional HPs are in the mode or are in the base extension, they are still defined elsewhere, and that means modifying code in order to provide them. The design for ConfigPoint is that it is an easily included configuration system that allows having several variations on the configuration, without needing to include services that are not commonly available, such as the DICOM Hanging Protocol service - which exists in a couple of places but is very cumbersome to use and will mostly require sites adding custom code in order to support it. config-point just needs static files added, so it meets the same sort of requirements for deployment that you are going to have for OHIF anyways. The files can be gzipped on disk and served already compressed, they can be edited in any plain text editor and redeployed without rebuilding any of the software. That is being used already for specific configurations deployed to different teams within a healthcare setting (different deployments for each department, and for each species being treated).
In terms of use with modes, I actually agree there - the modes should have names/settings for specific use cases, but should also be site definable - not require code level customizations to setup. If a site wants the PT/CT mode to auto-open a fusion viewport, and then two other viewports, well, that becomes configurable at the site, as long as the underlying code has it supported.

extensions/cornerstone/src/commandsModule.js
```diff
@@ -291,32 +291,6 @@ const commandsModule = ({ servicesManager, commandsManager }) => {
         });
       }
     },
-    updateTableWithNewMeasurementData({
```
sedghi, line 294: why is this removed?

platform/core/src/services/HangingProtocolService/HPMatcher.js
```diff
@@ -7,7 +7,7 @@ import validate from './lib/validator';
  * @param  {Array} rules Array of MatchingRules instances (StudyMatchingRule|SeriesMatchingRule|ImageMatchingRule) for the match
  * @return {Object}      Matching Object with score and details (which rule passed or failed)
  */
-const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
+const match = (metadataInstance, rules, customAttributeRetrievalCallbacks, existing) => {
```
sedghi, line 10: can you update the documentation of this function too?

platform/core/src/services/HangingProtocolService/HPMatcher.js
```diff
@@ -36,7 +36,13 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
 
     // Create a single attribute object to be validated, since metadataInstance is an
     // instance of Metadata (StudyMetadata, SeriesMetadata or InstanceMetadata)
-    const attributeValue = metadataInstance[attribute];
+    let attributeValue = metadataInstance[attribute];
+    if (attributeValue === undefined) {
+      if (attribute == 'NumberOfStudyRelatedSeries') {
```
sedghi, line 41: === instead of ==

platform/docs/docs/configuration/configuration.md
```diff
@@ -59,6 +73,38 @@ window.config = {
...
+parameters, for example:
+```
+https://ohif.hospital.org/?theme=mgHP&theme=euroKeyboard
+```
+to load two hypothetical theme settings files mgHP and euroKeyboard to add
+mammographic hanging protocols and European keyboard settings.
+
+A site can add such settings by creating custom files in:
+```
+.../build/theme/mgHP.json5
```
sedghi, line 90: can we be more specific about where these themes should be located? where is `build` located? how to experiment with it in the `development`

platform/docs/docs/configuration/configuration.md
```diff
@@ -59,6 +73,38 @@ window.config = {
...
+.../build/theme/euroKeyboard.json5
+```
+A site might build such different themes to support various user preferences
+or site differences between users, such as themes to support specific clinics
+or differences in user groups such as left on right mammography viewing versus
+right on left mammography viewing.
+
+The decision to use the JSON5 perser for this was primarily aimed at allowing
```
sedghi, line 98: parser

platform/docs/docs/configuration/theme-configuration.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Config-Point
+The configuration schema is based on the [config-point](https://github.com/OHIF/config-point)
+library.  This library is design to allow developers to creation configuration
+points for their code by declaring static values, which can be modified externally.
+See [config-point-service](../platform/services/config-point-service.md) for
+internal details and development documentation.
+
+## Theme Files
+Theme files are simple json5 definition files, placed into the deployment
```
sedghi, line 13: same here, the location of the theme files are not clear to me

platform/docs/docs/configuration/theme-configuration.md
```diff
@@ -0,0 +1,238 @@
...
+  },
+}
+```
+as an example that modifies the hanging protocols, both by adding and updating
+elements.  Note how this is a deep modification to a value.  The intent is
+to allow modifying configuration values which are heavily nested and/or list
+based.
+
+## Working with Objects
+The object tree is matched by the simple attribute name.  The attribute tree
```
sedghi, line 53: I understadn what this section is, but I needed to read twice to figure out that we are explaining how config-point works here (first I thought this is continuation of theme and hp stuff)

platform/docs/docs/configuration/theme-configuration.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Immediate operations
+Immediate operations perform the action immediately, and can thus be
+extended further.  They support the additional keys:
+* position to modify something at a given position
+
+The immediate operations are:
+* replace, insert, delete
+
+For example, supposing there was a configuraiton point ModalityList, then
```
sedghi, line 125: typo in configurations

platform/docs/docs/configuration/theme-configuration.md
```diff
@@ -0,0 +1,238 @@
...
+// Delete an item with value or id 'MR'
+ModalityList: {
+  'MR': {configOperation: 'delete'},
+
+// Insert before the item with id 'MG'
+ModalityList: {
+  'MG': {configOperation: 'insert', value: 'MGTomo'},
+```
+
+### Getter Opreations
```
sedghi, line 145: typo in operations

platform/docs/docs/configuration/theme-configuration.md
```diff
@@ -0,0 +1,238 @@
...
+```
+The MG protocol doesn't initially exist, so this would add a new hanging
+protocol, which would be defined in the normal hanging protocol definition.
+
+## Query List
+One of the suggested areas for customization is the columns in the query table.
+TODO
+
+## Demographics Overlay
+Another recommended change is to configure the demographcis overlay using
```
sedghi, line 233: typo demographics

platform/viewer/public/theme/theme.json5
```diff
@@ -0,0 +1,2 @@
+{
```
sedghi, line 1: is this intentionally empty? can we have an actual hp here so that the demo can showcase the `confi-point` stuff?

platform/core/src/services/HangingProtocolService/themeProtocolProvider.js
```diff
@@ -0,0 +1,116 @@
+import ConfigPoint from 'config-point';
+
+const { ThemeProtocols } = ConfigPoint.register({
```
sedghi, line 3: why are we specifying hangingProtocols inside the `core`?

platform/viewer/src/appInit.js
```diff
@@ -49,6 +50,14 @@ function appInit(appConfigOrFunc, defaultExtensions) {
...
+    // Make the config point accessible in the window
+    window.ConfigPoint = ConfigPoint;
```
sedghi, line 55: is this window assignment intentional?

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -54,6 +65,21 @@ class HangingProtocolService {
   addProtocols(protocols) {
     protocols.forEach(protocol => {
       if (this.protocols.indexOf(protocol) === -1) {
+        protocol.id = protocol.id || protocol.name;
+        protocol.name = protocol.name || protocol.id;
+        const { stages } = protocol;
+
+        stages.forEach(stage => {
```
sedghi, line 72: I know what you are doing here, but maybe put some comment of why for people reading this

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -227,6 +254,7 @@ class HangingProtocolService {
       // Viewport Settings
       //
       // protocol defined callback
+      viewport.viewportSettings = viewport.viewportSettings || { filter: () => [] };
```
sedghi, line 257: default viewportSettings can be just an array (no need to define `filter` as Arra.filter exist)

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
sedghi: what do you mean by "interaction with display sets"?

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
sedghi: what if it is fusion? pt-axial, ct-axial, fusion-axial (both pt-axial, ct-axial)

sedghi: we are actually moving to `displaySets` in future (because of fusion stuff), so I would suggest to stick with the series matching by rules only

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -357,6 +390,7 @@ class HangingProtocolService {
           SeriesInstanceUID: aSeries.SeriesInstanceUID,
           matchingScore: totalMatchScore,
           matchDetails: matchDetails,
+          displaySet: aSeries,
```
sedghi, line 393: I'm not sure if `displaySet` makes sense here, since this matching function should only care about metadata matching and sorting them. Why do we need `displaySet` here?

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -387,7 +421,7 @@ class HangingProtocolService {
     );
 
     const bestMatch = matchingScores[0];
-
+    if (bestMatch?.displaySet) existingMatches.push(bestMatch.displaySet);
```
sedghi, line 424: If for any reason we end up using the existingMatches (which I explained why we shouldn't above), this push to array should happen outside of `match` function

platform/viewer/src/appInit.js
sedghi: we can resolve this i guess after the new commits

wayfarer3130:
This will be revisited when the cornerstone 3d changes are integrated.

PR #2605, opened 2021-11-05
Fix/v3 scrollbars

wayfarer3130:
Just enable the CSS/HTML class to enable showing the scrollbars on the StudyBrowser thumbnail list

### DIFF
--- a/platform/ui/src/assets/styles/styles.css
+++ b/platform/ui/src/assets/styles/styles.css
@@ -1,6 +1,11 @@
 /* CUSTOM OHIF SCROLLBAR */
 .ohif-scrollbar {
   scrollbar-color: #3E5967 transparent;
+  scrollbar-gutter: stable;
+}
+
+.ohif-scrollbar:hover {
+  overflow-y: auto;
 }
 
 .ohif-scrollbar::-webkit-scrollbar {
--- a/platform/ui/src/components/MeasurementTable/MeasurementTable.jsx
+++ b/platform/ui/src/components/MeasurementTable/MeasurementTable.jsx
@@ -15,7 +15,7 @@ const MeasurementTable = ({ data, title, amount, onClick, onEdit }) => {
         </span>
         <span className="text-base font-bold text-white">{amount}</span>
       </div>
-      <div className="overflow-x-hidden overflow-y-auto ohif-scrollbar max-h-112">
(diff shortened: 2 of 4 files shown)


sedghi:
I deliberately removed it as it was not in the initial design, based on the discussion if you can auto-hide it when mouse is out of the study browser panel, and it only appears if the mouse enters, I don't see why not to merge.

wayfarer3130:
> I deliberately removed it as it was not in the initial design, based on the discussion if you can auto-hide it when mouse is out of the study browser panel, and it only appears if the mouse enters, I don't see why not to merge.

The issue is the scroll bar wasn't appearing at all, whether or not the mouse was in the area, so there wasn't a way to scroll many thumbnails.   If you have a CSS rule that shows only when hovering, I'd be fine with that, but that isn't how it works right now.

platform/ui/src/assets/styles/styles.css
```diff
@@ -1,6 +1,11 @@
 /* CUSTOM OHIF SCROLLBAR */
 .ohif-scrollbar {
   scrollbar-color: #3E5967 transparent;
+  overflow-y: hidden;
```
wayfarer3130, line 4: As requested, make this hidden unless hovering.
@sedghi - I did this for both the measurements and the thumbnail area, do you want me to do that just for thumbnails?

sedghi (changes requested):
The UI has an undesired width change (see video below), I believe you should use [scrollbar-gutter ](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter)

https://user-images.githubusercontent.com/7490180/147146475-56c0563d-a30a-4304-8d7d-c1d842d8fcbf.mp4

)

wayfarer3130:
@sedghi  -added the scrollbar gutter as requested.  Also, I moved the scrollbar to the thumbnail level, which keeps the header stable.

sedghi:
@wayfarer3130 great it works. One of the tests is failing which is for snapshot modal which has a auto-hide scrollbar right now as well.

wayfarer3130:
@sedghi  - I fixed the failing test (well, at least locally, still waiting for a CI build).  The problem is a defect in WebKit that doesn't allow scrollIntoView to execute if the overflow is hidden.  I resolved it by making the scrollbars always visible for ohif-scrollbar, and then set them to hide for the thumbnail view.

Issue #2612, opened 2021-11-17
UI lags (jittery) when navigating studies with a large number of series


ngutman:
### Describe the Bug

For studies with a large number of series there's a noticeable UI lag when navigating between series (by clicking on the thumbnail or clicking on a measurement)

We started noticing issues when we loaded full tomosynthesis (3D Mammograms)  studies to OHIF. Such studies contain a relatively large number of series (~12-18) containing large frames (~2000x2500px). After some investigation, it seems it's related to the study browser thumbnails being rendered on every display set change.

Here's a screenshot of the call-tree after doing performance instrumentation on Firefox DevTools:  

![image](https://user-images.githubusercontent.com/1540134/142286172-62bd05ea-da0e-43d0-b2dc-3d4ea30b582c.png)

Diving into the JS Flame Chart shows that the expensive [renderToCanvas](https://github.com/cornerstonejs/cornerstone/blob/master/src/rendering/renderToCanvas.js) function is called many times from [Thumbnail](https://github.com/OHIF/Viewers/blob/master/platform/ui/src/components/studyBrowser/ImageThumbnail.js#L90)

<img width="1919" alt="image" src="https://user-images.githubusercontent.com/1540134/142286072-7ce8f342-126f-48fd-a670-cce7f623f56a.png">

### What steps can we follow to reproduce the bug?

1. Load a study with a lot of series of large frames and navigate

I'll open a PR with our fix in a bit.

Cheers!

sedghi:
Please review the latest code in the master branch. I am confident that this issue has been resolved. If it persists, kindly reopen the issue with updated details. 

Try viewer-dev.ohif.org instead of viewer.ohif.org
Our `viewer.ohif.org` is deployed from `release` branch while `viewer-dev.ohif.org` is our `master ` branch
Read more about branch explanations here https://docs.ohif.org/development/getting-started#developing

PR #2613, opened 2021-11-17
OHIF-2612: Fix ImageThumbnail unnecessary renderToCanvas calls

ngutman:
- Issue https://github.com/OHIF/Viewers/issues/2612
- Don't call renderToCanvas if ImageThumbnail already loaded (to avoid unnecessary renders on every DOM change)


platform/ui/src/components/studyBrowser/ImageThumbnail.js
```diff
@@ -86,7 +86,7 @@ function ImageThumbnail(props) {
   }, [purgeCancelablePromise]);
 
   useEffect(() => {
-    if (image.imageId) {
+    if (image.imageId && isLoading) {
```
sedghi, line 89: shouldn't we add `isLoading` in the dependency array of the `useEffect` too?

sedghi (commented):
Looks good to me, please see my comment

sedghi:

sedghi:

PR #2614, opened 2021-11-17
fix(OHIFv3):Overlays were broken.

wayfarer3130:
Adds the ability to display overlays in OHIF v3.
This relies on 3 components getting changed: OHIF, dicomweb-client and cornerstoneTools.
The change here adds a retrieveBulkData method to any metadata with a BulkDataURI, returning a promise when the data is available.
The dicomweb-client change allows fetching both multipart and singlepart bulkdata, required to fetch data from compliant DICOMweb servers, as well as handle non-compliant servers which return multipart encoded data.
The cornerstoneTools adds the ability to fetch the overlay asynchronously inside OverlayTool and render it from a binary DICOM or JSON InlineBinary result to the format required for OverlayTool.
The changes are backwards compatible - that is, if none of them are used, then there isn't any affect on the client.


extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -252,9 +252,38 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
...
+       */
+      const addRetrieveBulkData = instance => {
+        const naturalized = naturalizeDataset(instance);
+        Object.keys(naturalized).forEach(key => {
+          const value = naturalized[key];
+          if (value && value.BulkDataURI && !value.Value) {
+            // Provide a method to fetch this
+            value.retrieveBulkData = () => {
+              const options = {
+                multipart: false,
```
wayfarer3130, line 269: This is a single bulkdata retrieve, so set multipart to false

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -252,9 +252,38 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
...
+                BulkDataURI: value.BulkDataURI,
+                StudyInstanceUID: naturalized.StudyInstanceUID,
```
wayfarer3130, line 271: The study UID is required if the BulkDataURI is relative, which is true for the back end I'm testing against.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -252,9 +252,38 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
         sortFunction
       );
 
+      /**
+       * naturalizes the dataset, and adds a retrieve bulkdata method
+       * to any values containing BulkDataURI.
+       * @param {*} instance
+       * @returns naturalized dataset, with retrieveBulkData methods
+       */
+      const addRetrieveBulkData = instance => {
```
wayfarer3130, line 261: @swederik and @sedghi  Could you take a look at how this works?  There is an async function that gets added which gets called when the actual value is needed.  That should support both Overlay and Palette colour data, as well as anything else that happens to get added as BulkDataURI.  It isn't perfect since it requires the user of the data to know whether to call it or not, but since it is the user who knows whether they need that data, I couldn't see a better location for tracking it.

platform/core/src/classes/MetadataProvider.js
```diff
@@ -1,8 +1,6 @@
 import queryString from 'query-string';
 import dicomParser from 'dicom-parser';
 import getPixelSpacingInformation from '../utils/metadataProvider/getPixelSpacingInformation';
-import fetchPaletteColorLookupTableData from '../utils/metadataProvider/fetchPaletteColorLookupTableData';
```
wayfarer3130, line 4: Should also remove fetchpaletteColor at some point, but for now, just remove the commented out calls.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -252,9 +252,38 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
...
+              const options = {
+                multipart: false,
+                BulkDataURI: value.BulkDataURI,
+                StudyInstanceUID: naturalized.StudyInstanceUID,
+              };
+              return qidoDicomWebClient.retrieveBulkData(options).then(val => {
+                const ret = val && val[0] || undefined;
+                value.Value = ret;
+                return ret;
+              });
```
wayfarer3130, line 277: @chafey  - here is the corresponding change in OHIFv3 - the retrieveBulkData call gets made and returns a promise to return an array buffer.  Fairly simple to implement, and is completely backwards compatible with OHIF v2, and any other cornerstone clients.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -252,9 +252,38 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
...
+       * naturalizes the dataset, and adds a retrieve bulkdata method
+       * to any values containing BulkDataURI.
+       * @param {*} instance
+       * @returns naturalized dataset, with retrieveBulkData methods
+       */
+      const addRetrieveBulkData = instance => {
+        const naturalized = naturalizeDataset(instance);
+        Object.keys(naturalized).forEach(key => {
+          const value = naturalized[key];
+          if (value && value.BulkDataURI && !value.Value) {
```
sedghi, line 265: what are you checking for `!valu.Value`? are you checking for null/undefined, can it be 0?

sedghi: mabye we add a comment above this if statement to say what is the branch here

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -252,9 +252,38 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
...
+            // Provide a method to fetch this
+            value.retrieveBulkData = () => {
+              const options = {
+                multipart: false,
+                BulkDataURI: value.BulkDataURI,
+                StudyInstanceUID: naturalized.StudyInstanceUID,
+              };
+              return qidoDicomWebClient.retrieveBulkData(options).then(val => {
+                const ret = val && val[0] || undefined;
```
sedghi, line 274: shouldn't we check `val.length` too?

extensions/default/src/DicomWebDataSource/index.js
sedghi: Similarly for this, this is a useful comment as a code comment

wayfarer3130: Added the comment.

extensions/default/src/DicomWebDataSource/index.js
wayfarer3130: value.Value can only be a binary byte array for pixel data, so, no, 0 isn't an option.  The reason this is checked here is that
the value might already have been read, in which case it isn't necessary to read it a second time.

wayfarer3130: Updated with a comment.

extensions/default/src/DicomWebDataSource/index.js
wayfarer3130: No, the API contract is a bit odd, it returns things in an array, or else returns undefined if there isn't anything, so there is
always at least 1 element or else the entire return is null.

sedghi (approved):

extensions/default/src/DicomWebDataSource/index.js
sedghi: @wayfarer3130  Something that just came to my mind is that how many fields have BulDataURI, since wee are checking for all the dicom keys and all the instances, and it might be slow ....

sedghi:
@wayfarer3130 Can you check my last comment on the performance concern?

extensions/default/src/DicomWebDataSource/index.js
wayfarer3130: It is a lot faster than always fetching all the data, and I suspect is faster than parsing the data once (which can take a noticeable amount of time).   
It is also currently only done on the top level tags.  It would be possible to have a set of tag values to check, but given how overlays are encoded, I suspect that would be slower because there isn't a good way to iterate over just those.
I don't think this is a significant performance concern right now.

sedghi: I see, thanks.

sedghi (changes requested):
Thanks for your contribution, please see my comments.

PR #2627, opened 2021-12-02
fix: don't retry failed requests while prefetching

slak44:
The StudyPrefetcher adds all the non-cached image ids to cornerstone.imageLoadPoolManager. If something happens with the network (eg user loses internet, dicom-web server fails), then the prefetch requests for each image will fail. Then, on the next stack scroll, all of these failed requests will be resent (multiple times even; prefetchImageIds seems to be called several times per scroll). For large-ish studies, thousands of requests make the browser tab freeze.

The fix is to stop prefetching images that failed with non-429 status codes. If the failure is temporary, the images will load correctly once the user scrolls to them manually.


sedghi:

sedghi:

PR #2630, opened 2021-12-03
fix: don't retry thumbnail requests after errors

slak44:
When the request for a thumbnail image fails, it is retried immediately, which can cause a loop of infinite failed requests.

This checks the existing error state before starting a new request, and hides the loader if an error did occur.

### DIFF
--- a/platform/ui/src/components/studyBrowser/ImageThumbnail.js
+++ b/platform/ui/src/components/studyBrowser/ImageThumbnail.js
@@ -93,12 +93,13 @@ function ImageThumbnail(props) {
   }, [canvasRef, image, image.imageId]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


salimkanoun:
Nice! this should fix my issue i discribed some times ago here : #2170

sedghi (commented):
I think we shouldn't absolutely avoid re-fetching if there is an error. Wouldn't re-fetching resolve the error in some situations? maybe filter some error codes? (CC @swederik )

sedghi:
we should see if there is still issue and whether this resolves it

salimkanoun:
No I don't see infinite request anymore in v3 but the HP don't start if there are failing series/uid/metadata request (even if some series are available) (get stuck in the OHIF spinner) (reported here :https://github.com/OHIF/Viewers/issues/3674)

PR #2639, opened 2021-12-22
Fix/v3 remove unused measurement code

wayfarer3130:
The updateTableWithNewMeasurementData is obsolete code from OHIFv2 and should be removed.  Checked that measurements still work and that there are no references anywhere to this code.

### DIFF
--- a/extensions/cornerstone/src/commandsModule.js
+++ b/extensions/cornerstone/src/commandsModule.js
@@ -291,32 +291,6 @@ const commandsModule = ({ servicesManager, commandsManager }) => {
         });
       }
     },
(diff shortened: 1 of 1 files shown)


extensions/cornerstone/src/commandsModule.js
```diff
@@ -291,32 +291,6 @@ const commandsModule = ({ servicesManager, commandsManager }) => {
         });
       }
     },
-    updateTableWithNewMeasurementData({
```
wayfarer3130, line 294: @sedghi  - this is all dead code

PR #2641, opened 2021-12-22
feat(OHIFv3):Add theme/config-point capability and docs.

wayfarer3130:
This change has just the config-point stuff in it, separated out from any use of it.
@sedghi as I said, I was going to split the stuff out.

### DIFF
--- a/extensions/cornerstone/src/commandsModule.js
+++ b/extensions/cornerstone/src/commandsModule.js
@@ -186,7 +186,7 @@ const commandsModule = ({ servicesManager, commandsManager }) => {
         }
 
         const viewportInfo = getEnabledElement(i);
-        const hasCornerstoneContext =
+        const hasCornerstoneContext = viewportInfo.context &&
           viewportInfo.context === 'ACTIVE_VIEWPORT::CORNERSTONE';
 
         if (hasCornerstoneContext) {
--- a/platform/docs/docs/configuration/theme-configuration.md
+++ b/platform/docs/docs/configuration/theme-configuration.md
@@ -1,0 +1,287 @@
+# Theme Configuration
+When adding new theme extendible configuration items, please document
+them here.  See [Theme Configuration with Config Point](#configPoint) on how to modify certain types of
+configuration values using the config-point defintions.
+
+## Hanging Protocols
+It is possible to customize the available hanging protocols by defining them
+in a new theme file OR by defining the hanging protocols in a custom mode.
+If done correctly, the mode defined hanging protocols are automatically
+applied when using a particular view mode, allowing for further customization
+by theme files.
+
+The default hanging protocols for the cornerstone mode is defined in
+themeProtocolProvider.js,  in a configuration point named
+`ThemeProtocols.protocols`  To define a new
+hanging protocol, it is possible to simply extend the protocols list with
+a new definition, for example, in the file mgHP.json5, the following definition
+would add a new hanging protocol:
+```js
+{
+  ThemeProtocols: {
+    protocols: {
+      MG:         {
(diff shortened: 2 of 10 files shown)


platform/viewer/src/appInit.js
```diff
@@ -49,6 +50,12 @@ function appInit(appConfigOrFunc, defaultExtensions) {
     appConfig,
   });
 
+  if (window) {
```
sedghi, line 53: why checking `window`?

platform/docs/docs/configuration/theme-configuration.md
```diff
@@ -0,0 +1,287 @@
...
+The default hanging protocols for the cornerstone mode is defined in
+themeProtocolProvider.js,  in a configuration point named
+`ThemeProtocols.protocols`  To define a new
+hanging protocol, it is possible to simply extend the protocols list with
+a new definition, for example, in the file mgHP.json5, the following definition
+would add a new hanging protocol:
+```js
+{
+  ThemeProtocols: {
+    protocols: {
```
sedghi, line 22: Why this is an object instead of an array? I believe protocols should be an array of prtocols. Is this how config-points work internaly?

```
{
 ThemeProtocols: {
   protocols:[
     {
       id: 'MG'
       ....
     },
     {
       id:'prostate',
       ...
     }
   ]
 }
}
```

platform/docs/docs/configuration/theme-configuration.md
```diff
@@ -0,0 +1,287 @@
+# Theme Configuration
```
sedghi, line 1: Is `theme` a config-point concept? since `theme` in OHIF might mean appearance ... 
CC: @swederik

wayfarer3130: The theme settings are mostly around appearance, so I called it theme.  If you want another name, thats fine with me, just let me know what it is.  I think of them as themes, even if it is somewhat more than just appearance, it still tends to be related to how things look and interact to the user.

platform/docs/docs/configuration/theme-configuration.md
wayfarer3130: protocols is an array as originally declared, but there is an object based syntax that can be used to extend the array.  If you could do something like:
[ insertAtEnd: newValue ]
then it would work, but because that syntax is kind of ugly, the alternate syntax here works.  In fact, I have a syntax
defined that allows adding them as array elements, but it is kind of long/ugly:
[ {configOperation: 'insert', value: { ... MG protocol definition } }, ... ]

platform/viewer/src/appInit.js
wayfarer3130: I could just avoid that, and have the check inside the ConfigPoint load itself - it is just that it is only relevant if you are actually in a browser and have a URL that you can fetch the data from.  If you want, I will move the check into config point itself.

extensions/cornerstone/src/commandsModule.js
```diff
@@ -186,7 +186,7 @@ const commandsModule = ({ servicesManager, commandsManager }) => {
         }
 
         const viewportInfo = getEnabledElement(i);
-        const hasCornerstoneContext =
+        const hasCornerstoneContext = viewportInfo.context &&
```
wayfarer3130, line 189: This was just an annoying error message that occasionally popped up due to load timing.

wayfarer3130:
@swederik - would you have a chance to look at this sometime?  Would it help if I scheduled 50 minutes to go over it with you and Alireza, and anyone else you think should look at it?
FYI: @sedghi

sedghi (approved):

PR #2651, opened 2021-12-29
fix: remove temporary link element after csv download

pwespi:
- the `_createAndDownloadFile` method in `extensions/measurement-tracking/src/_shared/downloadCSVReport.js` creates a temporary link element to download the data
- currently this temporary element is not removed after download, which leads to many `<a>` elements being appended to the body when performing multiple downloads
- with this PR the temporary element is removed after download


swederik:
Good catch! Thanks!

PR #2658, opened 2022-01-12
Feat/v2 annotations

Punzo:
Hi @igoroctaviano, just updated the annotation branch for OHIF-v2. Ref https://github.com/OHIF/Viewers/issues/1215

NOTE: it does still not work for study 1.3.6.1.4.1.14519.5.2.1.7310.5101.860473186348887719777907797922 (SCOORD3D points). The study has multiple data series and 1 SR series. The problem is that in the SR series, we miss any information regarding _ContentSequence_ (which contain _ReferencedSOPSequence_). In general, we derive to which series and frame the annotation is located from _ReferencedSOPSequence_. 

For example, for one point we have:
image

The frame could be computed by comparing the annotation SCOORD3D coordinates with the ImagePatientPosition of the frames, but I don't think we can derive univocally anyhow which one is the referenced series (i.e. I could have two source series, with the same frameOfReferenceUID, that have each a frame with the same z coordinate between ImagePositionPatient and the point scoord3d coordinates).

I will double check with Andrey, Steve and Markus, but I believe the dataset for this study is not complete.


igoroctaviano:
@Punzo thanks for taking a look at this🥇  
Let's wait for their reply then we can move forward with the review.

Punzo:
@igoroctaviano this is ready for testing and review. Please have a look when you have time!

platform/core/src/DICOMSR/SCOORD3D/utils/getRenderableData.js
```diff
@@ -1,9 +1,24 @@
...
-const getRenderableData = (GraphicType, GraphicData, ValueType) => {
+const getRenderableData = (
+  GraphicType,
+  GraphicData,
+  ValueType,
+  imageMetadata
+) => {
   let renderableData;
 
+  // Question for Steve and Andrey, with study 1.3.6.1.4.1.14519.5.2.1.6279.6001.100063870746088919758706456900
```
igoroctaviano: Can we remove this comment for Steve/Andrey and move it to PR comments instead of code?

platform/core/src/DICOMSR/SCOORD3D/utils/getRenderableData.js
```diff
@@ -1,9 +1,24 @@
...
+  // (the one with ~1000 polygons), looks like the SCOORD3D coordinates are given respect to the
+  // center of the image, however in the dicom refers that the coordinates are given respect to
```
igoroctaviano: I saw these comments in other parts of the code and I think moving forward we can stick with this style:

```
/** 
 * Some random text...
 * Some random text...
 */
```
 
 It looks better / follows the same pattern as jsDocs.

igoroctaviano (approved):
LGTM.

Screen Shot 2022-01-31 at 14 23 39
Screen Shot 2022-01-31 at 14 25 34
Screen Shot 2022-01-31 at 14 24 36
Screen Shot 2022-01-31 at 14 23 58
Screen Shot 2022-01-31 at 14 23 39

igoroctaviano:
@Punzo before merging, have you faced any performance issues with this branch?

platform/core/src/DICOMSR/SCOORD3D/utils/getRenderableData.js
Punzo: oh yes this has to be removed, thanks for catching it!

Punzo: uhm this commet was already removed in latest commit

platform/core/src/DICOMSR/SCOORD3D/utils/getRenderableData.js
Punzo: fixed all

Punzo:
> @Punzo before merging, have you faced any performance issues with this branch?

@igoroctaviano 
potentially we have 2 perfomance bottleneck:

1 ) https://github.com/OHIF/Viewers/pull/2658/files#diff-71437d32409e6f0c2c6d7bc53623f31d89e5897a92a78c1c03bd5fb8baf1eb58R67

this can be an issue only if we have thousands of coords for which we didn't find the referenced image from the metadata (and at the same time  a huge amount of images, i.e. at least tens of thousands of images in all series).  

Still the number of operations per iteration is not too large and also in our user cases only the annotations of one study were missing the ReferencedSOPSequence (so we are making this calculcations only for one case). We could use the SliceLocation instead of calculating this https://github.com/OHIF/Viewers/pull/2658/files#diff-71437d32409e6f0c2c6d7bc53623f31d89e5897a92a78c1c03bd5fb8baf1eb58R76-R88 , but the SliceLocation is optional and many times a wrong value is provided.

2) Here we have a mtrix inversion https://github.com/OHIF/Viewers/pull/2658/files#diff-c4af53e9621d9280ba06c716f9211263d1f97b4d90fa8358996c149ef5b27131R44, but it is done only once (at loading) per annotation, so it does not impact the perfomances. 

On the other hand I noticied this:

image

the loading bar in the thumbnail stops at less than 100% for some series. Can you double-check if you can reproduce it?

igoroctaviano:
> > @Punzo before merging, have you faced any performance issues with this branch?
> 
> @igoroctaviano potentially we have 2 perfomance bottleneck:
> 
> 1. https://github.com/OHIF/Viewers/pull/2658/files#diff-71437d32409e6f0c2c6d7bc53623f31d89e5897a92a78c1c03bd5fb8baf1eb58R67
> 
> this can be an issue only if we have thousands of coords for which we didn't find the referenced image from the metadata (and at the same time a huge amount of images, i.e. at least tens of thousands of images in all series).
> 
> Still the number of operations per iteration is not too large and also in our user cases only the annotations of one study were missing the ReferencedSOPSequence (so we are making this calculcations only for one case). We could use the SliceLocation instead of calculating this https://github.com/OHIF/Viewers/pull/2658/files#diff-71437d32409e6f0c2c6d7bc53623f31d89e5897a92a78c1c03bd5fb8baf1eb58R76-R88 , but the SliceLocation is optional and many times a wrong value is provided.
> 
> 2. Here we have a matrix inversion https://github.com/OHIF/Viewers/pull/2658/files#diff-c4af53e9621d9280ba06c716f9211263d1f97b4d90fa8358996c149ef5b27131R44, but it is done only once (at loading) per annotation, so it does not impact the perfomances.
> 
> On the other hand I noticed this:
> 
> image
> 
> the loading bar in the thumbnail stops at less than 100% for some series. Can you double-check if you can reproduce it?

This issue seems to be unrelated / but I have changed the code to initialize the tools only once.

igoroctaviano (approved):
LGTM. @Punzo can you give it another try + merge?

If possible, can you play with all datasets to check that they are working and also test with other OHIF datasets to check that they are playing just fine after these changes?

Thank you in advance.

Punzo:
> LGTM. @Punzo can you give it another try + merge?
> 
> If possible, can you play with all datasets to check that they are working and also test with other OHIF datasets to check that they are playing just fine after these changes?
> 
> Thank you in advance.

ok the issue with loading bar seems to happen only when pointing to my local DCM4CHEE. So it is unrealted to this.

I retested the user cases with annotations and I tested also against several studies (SEG, RT, etc...). LGTM.

Merging!

PR #2660, opened 2022-01-14
Config based mode and extension registration

JamesAPetts:
Configuration based registration for OHIF.

- Modes and extensions now just have a single default export.
- CLI will soon be added for installing extensions and modes, but at time of this PR:
-- extensions and mode must be installed via yarn and then added to platform/viewer/pluginConfig.json.

pluginConfig.json is processed prior to the webpack build to inject code which will automatically hook up the extensions before the webpack build.


sedghi (approved):
Looks great

PR #2664, opened 2022-01-17
feat: Add the list command to print extensions and modes

sedghi:
image

### DIFF
--- a/platform/cli/src/index.js
+++ b/platform/cli/src/index.js
@@ -13,6 +13,7 @@ import {
   removeExtension,
   addMode,
   removeMode,
+  listPlugins,
 } from './commands/index.js';
 
 const currentDirectory = process.cwd();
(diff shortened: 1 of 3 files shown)


platform/cli/src/index.js
```diff
@@ -94,4 +95,12 @@ program
     removeMode(packageName);
   });
 
+program
+  .command('list')
+  .description('List Added Extensions and Modes')
+  .action(() => {
+    const configPath = path.join(__dirname, '../../viewer/pluginConfig.json');
```
JamesAPetts, line 102: Hmm, I think this is fine for now, but I think eventually the cli tool will be used in the viewer directory? I'm currently using all the other commands in the viewer directory like:

```node ../cli/src/index.js add-mode @ohif-test/mode-clock```

So maybe it should just check locally, and log an error if the user is not in an `@ohif/viewer` repository?

JamesAPetts (approved):
Minor comment/discussion point, otherwise looks good.

platform/cli/src/index.js
sedghi: Added

PR #2683, opened 2022-01-24
fix(OHIFv2):Fix a dependency issue on windows caused by node-gyp

wayfarer3130:
The node-gyp version used by the docz installed for OHIFv2 is very old
and is incompatible with some recent python updates and other packaging
issues.  Just force a node-gyp newer version to be used.  No functional
changes.


wayfarer3130:
Going to just let the e2e commit cover this one as it seems to take a bunch of overlapping files.

PR #2691, opened 2022-01-28
Feat/v3 video

wayfarer3130:
Support video DICOM objects.  Requires the DICOMweb service to be able to stream the video data directly on the .../pixeldata endpoint (or whatever the BulkDataURI is).


platform/core/src/services/DisplaySetService/DisplaySetService.js
```diff
@@ -151,13 +176,15 @@ export default class DisplaySetService {
     }
   };
 
-  makeDisplaySetForInstances(instances, settings) {
+  makeDisplaySetForInstances(instancesSrc, settings) {
+    let instances = instancesSrc;
     const instance = instances[0];
```
wayfarer3130, line 181: @swederik and @sedghi  - please note this change in makeDisplaySetForInstances.  The change modifies the handling by continuing with adding display sets until all instances are used up, or there are no handlers left, and grouping all the display sets into a single set that gets returned.  There is an integration test that checks this, as well as the appropriate test data to validate it.

PR #2696, opened 2022-01-31
Ohif cli

sedghi:
### PR Checklist


extensions/cornerstone/src/id.js
```diff
@@ -1,3 +1,6 @@
-const id = 'org.ohif.cornerstone';
+import packageJson from '../package.json';
```
JamesAPetts, line 1: Will this cause us any issues?

The ugly thing here is that we are including the package json of all extensions/modes in source now. But I get that you don't want to have the ID defined twice.

extensions/default/src/ViewerLayout/index.jsx
```diff
@@ -89,12 +88,13 @@ function ViewerLayout({
   rightPanels,
   viewports,
   ViewportGridComp,
+  useNavigate,
```
JamesAPetts, line 91: What's the benefit of passing this down as a prop?

extensions/dicom-sr/src/index.js
```diff
@@ -23,18 +23,21 @@ const OHIFCornerstoneSRViewport = props => {
...
   /**
-   * Only required property. Should be a unique value across all extensions.
+   * Only two required properties. Should be a unique value across all extensions.
    */
   id,
+  version,
   dependencies: [
     // TODO -> This isn't used anywhere yet, but we do have a hard dependency, and need to check for these in the future.
+    // TODO #2 -> When ohif-cli installs this, we should check if these dependencies are installed and install them if not
+    //         -> Added these to peer dependencies in preparation for this.
```
JamesAPetts, line 33: Good shout about adding these to the peer deps.

modes/longitudinal/src/index.js
```diff
@@ -1,38 +1,54 @@
...
 };
 
 const dicomvideo = {
-  sopClassHandler: 'org.ohif.dicom-video.sopClassHandlerModule.dicom-video',
-  viewport: 'org.ohif.dicom-video.viewportModule.dicom-video',
-}
+  sopClassHandler:
+    '@ohif/extension-dicom-video.sopClassHandlerModule.dicom-video',
+  viewport: '@ohif/extension-dicom-video.viewportModule.dicom-video',
+};
```
JamesAPetts, line 28: I know this wasn't changed in this PR, but its just come to my attention. Why are we adding dicom video and pdf to the longitudinal mode when these weren't in the original spec?

We can make a kitchen sink mode to demo everything, but the longitudinal mode has a very specific purpose/is a great of example of an actual mode rather than a swiss army knife. I'm not sure we should bloat it with new features? It shouldn't be the default mode to throw all shinies in?

cc @swederik

platform/viewer/src/routes/Mode/Mode.jsx
```diff
@@ -109,7 +111,10 @@ export default function ModeRoute({
   // For each extension, look up their context modules
   // TODO: move to extension manager.
   let contextModules = [];
-  extensions.forEach(extensionId => {
+
+  // TODO - JAMES -> Extensions now objects not arrays -> Fix.
```
JamesAPetts, line 115: We've done this now, can you remove this comment?

JamesAPetts (approved):
Some minor comments/questions.

Tested extension/mode creation locally and it seems like the files outputted from the templates are now fine, contain the correct keywords, etc.

modes/longitudinal/src/index.js
sedghi: that makes sense to me, i'm removing it

PR #2703, opened 2022-02-08
Fix retrieveSeriesMetadata

JamesAPetts:
We already had a defined pattern for retrieving, yet for some reason we were using an additional tacked on method to datasources. This fixes that.


sedghi (approved):
Looks great

PR #2708, opened 2022-02-14
refactor: 💡 Bump lodash and immer to avoid vulnerabilities

igoroctaviano:
Bump lodash and immer to avoid vulnerabilities.

# Highest vulnerability risk according to yarn audit.
image (1)
image


sedghi (approved):
Great, we need to bump all other things too at some point

Issue #2426, opened 2021-05-28
Problem viewer displaying US modality


sylvainb31:
Version : "@ohif/viewer@4.5.28"

Problem : 

I am using the ohif viewer to display images for several hospitals.
There is one hospital in which the ultrasound images (US modality) don't display correctly in the ohif viewer. I have tried changing the transfer syntax of the images but it still doesn't work.

Here is the display of the ultrasound :

![image](https://user-images.githubusercontent.com/37896804/120010298-44548700-bfdd-11eb-8d11-604013698b1d.png)

When trying to display an other ultrasound for an other hospital, it displays correctly :

![image](https://user-images.githubusercontent.com/37896804/120010525-88e02280-bfdd-11eb-8ea3-5cdef33ceaa3.png)

I thought the problem came from the images .dcm, but I tried displaying them on a viewer online and it displays correctly : 

![image](https://user-images.githubusercontent.com/37896804/120010814-e07e8e00-bfdd-11eb-9c3b-3d86abaafc19.png)

I've tried comparing the dataset of .dcm of both exams, but I can't find what's causing this.

Here is the result of the study query of the viewer for the ultrasound not displaying correctly :
studies/1.2.826.0.1.3680043.8.1275.9882141517/series
[metadatastudy.txt](https://github.com/OHIF/Viewers/files/6562064/metadatastudy.txt)

Here is the result of the series query of the viewer for the ultrasound not displaying correctly :
studies/1.2.826.0.1.3680043.8.1275.9882141517/series/1.2.410.200001.1.11801.447477310.2.20210427.1134649616.946.1/metadata
[metadataserie.txt](https://github.com/OHIF/Viewers/files/6562066/metadataserie.txt)

Here is the dataset of one of the image of the ultrasound not displaying correctly :
[Ultrasound dataset not displaying correctly.xlsx](https://github.com/OHIF/Viewers/files/6562008/Ultrasound.dataset.not.displaying.correctly.xlsx)

Here is the result of the study query of the viewer for the ultrasound displaying correctly :
studies/1.2.250.1.59.0.999.20210526.2117302/series
[metadatastudy.txt](https://github.com/OHIF/Viewers/files/6562136/metadatastudy.txt)

Here is the result of the series query of the viewer for the ultrasound not displaying correctly :
studies/1.2.250.1.59.0.999.20210526.2117302/series/1.2.840.113619.2.481.84108090782.1622017378.387/metadata
[...]

Issue #2420, opened 2021-05-26
Problem loading the US image. Image overwriting


pilacoepouco:
I have a problem loading images in US mode, all of them are overwritten. I'm accessing via URL.

![image](https://user-images.githubusercontent.com/53221609/119730769-7aaecc80-be4c-11eb-9259-7f73b4d1a3ec.png)

![image](https://user-images.githubusercontent.com/53221609/119731125-ee50d980-be4c-11eb-84a4-3029c9c7e101.png)

Any idea what it might be?

pilacoepouco:
I found the solution by adding the TAG PlanarConfiguration in the JSON of the instance.

**Tag**                 
 (0028,0006) 
 **VR**
  US   
 **Name** 
  Planar Configuration

shlomgad:
@pilacoepouco Can you please explain how exactly are you adding this tag to the JSON ?
Is it: "PlanarConfiguration" : "US"  ?

ed-morais:
Hello, @pilacoepouco!
How can I access the JSON of the instance?

Shuun26:
Thanks @pilacoepouco, but how do I find this JSON instance?

pilacoepouco:
@Shuun26  and @ed-morais  I use a DICOMWeb server which returns a JSON. Following the documentation reference : https://v3-docs.ohif.org/configuration/dataSources/dicom-json

Shuun26:
> @Shuun26 and @ed-morais I use a DICOMWeb server which returns a JSON. Following the documentation reference : https://v3-docs.ohif.org/configuration/dataSources/dicom-json

Thank you so much! 
Looking for some info in https://dicom.innolitics.com/ciods/rt-dose/image-pixel/00280006. Discovereded something interesting, the planar configuration solves the error because in some studies the tag Samples per Pixel has a value greater than 1, and, for some reason (I could be wrong, so please tell me if I'm ignoring something) the OHIF Viewer it's ignoring the Planar Configuration, because in the metadata doesn't appers the PlanarConfiguration tag.

Psd: So sorry for my bad redaction in English.

![image](https://user-images.githubusercontent.com/68860132/153613506-825c3db8-51af-47c5-91ac-96e4fd590eed.png)

RoshniPal:
Hello @pilacoepouco ,
I did add the PlanarConfiguration tag in the instance JSON and the structure looks like below.
{
	"studies": [{
		"StudyInstanceUID": "1.2.8",
[...]

PR #2720, opened 2022-02-23
Remove validNumber for photometricInterpretation key

Pierre-Baptiste:
I've had issues visualizing WADO DICOM files that had the `imagePixelModule` tag for the Image Loader.

It used to work on the master branch and while diving in, I realized the problem came from trying to parse a string value to a number one.

According to the [DICOM standards](https://dicom.innolitics.com/ciods/ct-image/image-pixel/00280004), `photometricInterpretation` should be a string (ex. "RGB", "HSV"...). In V3, a validation on numbers was added to all keys of the metadata object, including `photometricInterpretation` which returns undefined if it does not parse to a number. 

This PR suggests to remove the `validNumber` check for that key.

looks like #2420 , https://github.com/OHIF/Viewers/issues/2426#issuecomment-887435505

Before:
Screenshot 2022-02-23 at 17 50 59

After:
Screenshot 2022-02-23 at 17 50 23

### PR Checklist

### DIFF
--- a/platform/core/src/classes/MetadataProvider.js
+++ b/platform/core/src/classes/MetadataProvider.js
@@ -244,9 +244,7 @@ class MetadataProvider {
       case WADO_IMAGE_LOADER_TAGS.IMAGE_PIXEL_MODULE:
         metadata = {
           samplesPerPixel: validNumber(instance.SamplesPerPixel),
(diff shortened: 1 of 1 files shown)


swederik:
🤦 

That definitely seems wrong indeed. Thanks for the fix!

PR #2723, opened 2022-02-25
Feat/v3 themeable overlay

wayfarer3130:
This change redefines the viewport overlay as a customizeable viewport overlay that can be site changes without recompiling OHIF.  There is documentation provided to explain how to make the changes, as well as an example file that is suitable for
testing that demonstrates inclusions of patient and study demogrpahics in the general overlay.


platform/ui/src/components/ViewportOverlay/index.js
```diff
@@ -0,0 +1,117 @@
...
+        },
+      ],
+      bottomLeft: [
+      ],
+      bottomRight: [
+      ],
+      itemGenerator: props => {
+        const { item } = props;
+        const { title, value: valueFunc, condition, contents } = item;
+        if (condition && !condition(props)) return null;
```
igoroctaviano, line 97: The condition here might be a string, how you're handling the safe function which is a string?

igoroctaviano (changes requested):
Left a question regarding safe functions.

platform/ui/src/components/ViewportOverlay/index.js
wayfarer3130: The point of the safe function is that you can define the function using a string, so it can come from a format like JSON, but it gets auto-converted into a function that is safe for execution because the context is tightly controlled, only allowing values from the single argument of hte function to be used.

igoroctaviano: Do I have to change something on my end? I'm getting: condition is not a function when using these changes.

wayfarer3130: You need to do a yarn install probably - then check that node_modules/config-point/package.json has version 0.4.1 in it.

platform/ui/src/components/ViewportOverlay/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Generates a viewport overlay given a configuration element.
+ * The element typically comes from config-point, and contains
+ * things like the topleft/topRight/bottomleft/bottomRight definitions.
+ * @param {*} config is a configuration object that defines four lists of elements,
+ * one topLeft, topRight, bottomLeft, bottomRight contents.
+ * @returns React Viewport overlay function configured to use config for display
+ */
+const generateFromConfig = (config) => {
```
JamesAPetts, line 15: I understand the premise here, but I have the following comments:

- `@ohif/ui` is supposed to be a presentational component library (although it does admitly fail this on some components, which need to be improved). By using the `ConfigPoint` to generate components, we might be opening the door to effectively making them stateful, although I don't think that is quite done in this PR, but the API surface sure is a lot larger than before.
- I fail to see why we can't just achieve this logic with Higher Order Components (HOC's) instead, and leave the UI library as plain JSX driven by props:
  - Have a `configurationService` at the top level passed into `extensions` and `modes`  via the `extension`'s `preRegistration` hook, and the `mode`'s `onModeEnter`.
  - Improve the UI components in platform UI to take configuration as one of its props, which is used to populate the UI element with E.g. fields, other child components, etc.
  - The extension can create a configuration object by merging the extensions object with the set configuration.
  - The extension can then create a HOC which wraps the UI component, and passes this configuration in as props (e.g. similar to how the  `wrappedMeasurementPanel` component in `extensions/default/src/getPanelModule.js` passes services to the panel module).
  
For all the examples in this PR, right now the extension would do the heavy lifting of configuring its components with HOCs (since the configs used here are hardcoded in the extensiosn), and a configurationService wouldn't actually be needed at all. But I understand how configuration from the app level in future would be important also.

These are all high level thoughts and discussions, its very possible I missed something fundemental which makes this approach better.

JamesAPetts (commented):
Just some high level comments, I think it would be good for @swederik , @sedghi, you and I to discuss how we tackle this configuration problem going forwards.

extensions/cornerstone/src/ViewportOverlay.js
```diff
@@ -1,95 +1,57 @@
...
-  const bottomLeft = 'bottom-viewport left-viewport';
-  const overlay = 'absolute pointer-events-none';
-
-  const isZoomActive = activeTools.includes('Zoom');
-  const isWwwcActive = activeTools.includes('Wwwc');
-
-  if (!imageId) {
-    return null;
+import classnames from 'classnames'
+import ConfigPoint from 'config-point'
```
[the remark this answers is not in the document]
wayfarer3130: Yes, thats what this code does as it is written.  It also allows runtime configurable extensions to it, for example, by adding the ?theme=viewportOverlay parameter one could load additional overlay extensions.   Extensions don't have to depend on config-point to generate overlays - there are multiple options here:
1. Inherit an extension of the default overlay, and extend it with specific extensions using config-point
2. Use the base default overlay, without any extensions by just using the default component (no config-point involved in the extension side of this)
3. Write a custom overlay, without config point involved at all, just using it directly.
4. Write an extension that modifies another (earlier) extension's overlay by using config-point to customize the earlier extensions view.

Neither of 2 or 3 depend on config-point except indirectly.  Again, what I'm trying to do is decouple things that sites or extensions may want to configure from the extensions themselves in a way that doesn't require re-writing extensions or inheritting from an extension in a way that tightly couples the change into the actual implementation of the extension.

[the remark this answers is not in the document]
JamesAPetts: I see, thanks for this. The URL params is an interesting one, we can pass this to the mode, but it would be a bit clunky to somehow pass this to extensions, although possible.

> Again, what I'm trying to do is decouple things that sites or extensions may want to configure from the extensions themselves in a way that doesn't require re-writing extensions or inheritting from an extension in a way that tightly couples the change into the actual implementation of the extension.

I 100% get the motivation, where I'm wondering if we can improve is mostly on the formalization of this. The configuration being centrally managed and passed down to modes and extensions might keep things more logically seperated. Right now its unclear at what layer config-point should be used, as it looks like the configuration can be defined on any level. I may be wrong.

P.S. I think the problem you're trying to solve might be solvable (at least partially) by us thinking harder about how flexible the UI components in @ohift/ui themselves are, making the data flow of the configuration required by an extension/mode/application more unidirectional. Happy to discuss this tomorrow.

platform/ui/src/components/ViewportOverlay/index.js
JamesAPetts: > Obviously I didn't quite hit the mark the first time

FWIW I'm not suggesting that in any of these comments actually, I'm not sure of the answer, just thinking out loud here so we all come to the best place for the OSS lib :) . 

> but I'm going to try to update the PR a bit to show exactly this idea.
Sounds good, there may be entire swathes of problems that this approach solves that I'm just not thinking of yet.

PR #2727, opened 2022-03-01
fix(ohif-cli): Update generated files

matthis-d:
Hi!
I'm still working on an extension, and I think I faced other issues.

Generated webpack, babel and jest configuration assume that extension or mode are created in extensions directory, which is not the case with the CLI. This commit updates the paths accordingly. 
Base webpack requires having a tailwind config in the directory, which is not mandatory. It is now used only if the file exists.
Also, `yarn build` on an extension removes directory sources and put generated code in place. This commit fixes that.

Please tell me if I misunderstood where extensions are supposed to be generated: it is generated at the root of the application when I run it locally

### PR Checklist

- [x] `@mention` a maintainer to request a review: @sedghi 

### DIFF
--- a/.webpack/rules/cssToJavaScript.js
+++ b/.webpack/rules/cssToJavaScript.js
@@ -1,11 +1,14 @@
 const autoprefixer = require('autoprefixer');
 const path = require('path');
+const fs = require('fs');
 const tailwindcss = require('tailwindcss');
 const tailwindConfigPath = path.resolve(
   '../../platform/viewer/tailwind.config.js'
 );
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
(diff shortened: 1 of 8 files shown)


sedghi:
Extensions are not meant to be inside OHIF Viewer base. They can be created anywhere in your disk. You can then `link` them with the link option in the ohif-cli. 

Not sure, if you have seen this or not but might give some context to this upcoming feature https://github.com/NA-MIC/ProjectWeek/tree/master/PW36_2022_Virtual/Projects/OHIFModeGallery

matthis-d:
This is what I thought about when the prompt asked about creating a git directory.

I guess we have to change generated config files not to reference internal files like babel.config.js or webpack.base.js then.

PR #2730, opened 2022-03-04
feat(OHIFv3):Add PDF viewer

wayfarer3130:
Support viewing PDF documents.  Has a basic e2e test that checks the basic view capability is working.

### DIFF
--- a/extensions/dicom-pdf/package.json
+++ b/extensions/dicom-pdf/package.json
@@ -1,0 +1,49 @@
+{
+  "name": "@ohif/extension-dicom-pdf",
+  "version": "0.0.1",
+  "description": "OHIF extension for PDF display",
+  "author": "OHIF",
+  "license": "MIT",
+  "repository": "OHIF/Viewers",
+  "main": "dist/index.umd.js",
+  "module": "src/index.js",
+  "engines": {
+    "node": ">=10",
+    "npm": ">=6",
+    "yarn": ">=1.16.0"
+  },
+  "files": [
+    "dist",
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "dev": "cross-env NODE_ENV=development webpack --config .webpack/webpack.dev.js --watch --debug --output-pathinfo",
+    "dev:cornerstone": "yarn run dev",
+    "build": "cross-env NODE_ENV=production webpack --config .webpack/webpack.prod.js",
+    "build:package": "yarn run build",
+    "start": "yarn run dev",
+    "test:unit": "jest --watchAll",
+    "test:unit:ci": "jest --ci --runInBand --collectCoverage --passWithNoTests"
+  },
+  "peerDependencies": {
+    "@ohif/core": "^0.50.0",
+    "@ohif/ui": "^0.50.0",
+    "cornerstone-core": "^2.6.0",
+    "cornerstone-math": "^0.1.9",
+    "cornerstone-tools": "6.0.2",
+    "cornerstone-wado-image-loader": "4.0.4",
+    "dcmjs": "0.16.1",
+    "dicom-parser": "^1.8.9",
+    "hammerjs": "^2.0.8",
+    "prop-types": "^15.6.2",
+    "react": "^17.0.2",
+    "react-cornerstone-viewport": "4.1.2"
+  },
+  "dependencies": {
+    "@babel/runtime": "7.7.6",
+    "classnames": "^2.2.6"
+  }
+}
--- a/extensions/dicom-pdf/src/getSopClassHandlerModule.js
+++ b/extensions/dicom-pdf/src/getSopClassHandlerModule.js
@@ -1,0 +1,66 @@
+import { SOPClassHandlerName, SOPClassHandlerId } from './id';
+import { utils, classes } from '@ohif/core';
+
+const { ImageSet } = classes;
+
+const SOP_CLASS_UIDS = {
+  ENCAPSULATED_PDF: '1.2.840.10008.5.1.4.1.1.104.1',
+};
+
+const sopClassUids = Object.values(SOP_CLASS_UIDS);
+
+
+const _getDisplaySetsFromSeries = (instances, servicesManager, extensionManager) => {
+  console.log('Called getDisplaySetsFromSeries', instances);
+  const dataSource = extensionManager.getActiveDataSource()[0];
+  return instances
+    .map(instance => {
+      const { Modality, SOPInstanceUID, EncapsulatedDocument } = instance;
+      const { SeriesDescription, MIMETypeOfEncapsulatedDocument, } = instance;
+      const { SeriesNumber, SeriesDate, SeriesInstanceUID, StudyInstanceUID, NumberOfFrames } = instance;
+      console.log('Display PDF of type', MIMETypeOfEncapsulatedDocument, EncapsulatedDocument);
+      const pdfUrl = dataSource.retrieve.directURL({ instance, tag: 'EncapsulatedDocument', defaultType: MIMETypeOfEncapsulatedDocument });
+      const displaySet = {
+        //plugin: id,
+        Modality,
+        displaySetInstanceUID: utils.guid(),
+        SeriesDescription,
+        SeriesNumber,
+        SeriesDate,
+        SOPInstanceUID,
+        SeriesInstanceUID,
+        StudyInstanceUID,
+        SOPClassHandlerId,
+        referencedImages: null,
+        measurements: null,
+        pdfUrl,
+        others: [instance],
+        thumbnailSrc: dataSource.retrieve.directURL({ instance, defaultPath: "/thumbnail", defaultType: "image/jpeg", tag: "Absent" }),
+        isDerivedDisplaySet: true,
+        isLoaded: false,
+        sopClassUids,
+        numImageFrames: NumberOfFrames,
+        instance,
+      };
+      return displaySet;
+    });
+};
+
+export default function getSopClassHandlerModule({ servicesManager, extensionManager }) {
+  const getDisplaySetsFromSeries = instances => {
+    return _getDisplaySetsFromSeries(
+      instances,
+      servicesManager,
+      extensionManager
+    );
+  };
+
+  console.log("Registering", SOPClassHandlerName, sopClassUids);
+  return [
+    {
+      name: SOPClassHandlerName,
+      sopClassUids,
+      getDisplaySetsFromSeries,
+    },
+  ];
+}
--- a/extensions/dicom-pdf/src/id.js
+++ b/extensions/dicom-pdf/src/id.js
@@ -1,0 +1,7 @@
+const id = 'org.ohif.dicom-pdf';
+
+export default id;
+
+const SOPClassHandlerName = 'dicom-pdf';
+const SOPClassHandlerId = `${id}.sopClassHandlerModule.${SOPClassHandlerName}`;
+export { SOPClassHandlerName, SOPClassHandlerId };
--- a/extensions/dicom-pdf/src/index.js
+++ b/extensions/dicom-pdf/src/index.js
@@ -1,0 +1,93 @@
+import React from 'react';
+import getSopClassHandlerModule from './getSopClassHandlerModule';
+import id from './id.js';
+
+const Component = React.lazy(() => {
+  return import(
+    /* webpackPrefetch: true */ './viewports/OHIFCornerstonePdfViewport'
+  );
+});
+
+const OHIFCornerstonePdfViewport = props => {
+  return (
+    <React.Suspense fallback={<div>Loading...</div>}>
+      <Component {...props} />
+    </React.Suspense>
+  );
+};
+
+/**
+ *
+ */
+export default {
+  /**
+   * Only required property. Should be a unique value across all extensions.
+   */
+  id,
+  dependencies: [
+    // TODO -> This isn't used anywhere yet, but we do have a hard dependency, and need to check for these in the future.
+    // OHIF-229
+    {
+      id: 'org.ohif.cornerstone',
+      version: '3.0.0',
+    },
+    {
+      id: 'org.ohif.measurement-tracking',
+      version: '^0.0.1',
+    },
+  ],
+
+  preRegistration({ servicesManager, configuration = {} }) {
+    // No-op for now
+  },
+
+  /**
+   *
+   *
+   * @param {object} [configuration={}]
+   * @param {object|array} [configuration.csToolsConfig] - Passed directly to `initCornerstoneTools`
+   */
+  getViewportModule({ servicesManager, extensionManager }) {
+    const ExtendedOHIFCornerstonePdfViewport = props => {
+      return (
+        <OHIFCornerstonePdfViewport
+          servicesManager={servicesManager}
+          extensionManager={extensionManager}
+          {...props}
+        />
+      );
+    };
+
+    return [{ name: 'dicom-pdf', component: ExtendedOHIFCornerstonePdfViewport }];
+  },
+  getCommandsModule({ servicesManager }) {
+    return {
+      definitions: {
+        setToolActive: {
+          commandFn: ({ toolName, element }) => {
+            if (!toolName) {
+              console.warn('No toolname provided to setToolActive command');
+            }
+
+            // Set same tool or alt tool
+            const toolAlias = _getToolAlias(toolName);
+
+            cornerstoneTools.setToolActiveForElement(element, toolAlias, {
+              mouseButtonMask: 1,
+            });
+          },
+          storeContexts: [],
+          options: {},
+        },
+      },
+      defaultContext: 'ACTIVE_VIEWPORT::PDF',
+    };
+  },
+  getSopClassHandlerModule,
+};
+
+function _getToolAlias(toolName) {
+  let toolAlias = toolName;
+
+  return toolAlias;
+}
--- a/extensions/dicom-pdf/src/tools/modules/dicomVideoModule.js
+++ b/extensions/dicom-pdf/src/tools/modules/dicomVideoModule.js
@@ -1,0 +1,61 @@
+import cornerstone from 'cornerstone-core';
+
+const state = {
+  TrackingUniqueIdentifier: null,
+  trackingIdentifiersByEnabledElementUUID: {},
+};
+
+function setTrackingUniqueIdentifiersForElement(
+  element,
+  trackingUniqueIdentifiers,
+  activeIndex = 0
+) {
+  const enabledElement = cornerstone.getEnabledElement(element);
+  const { uuid } = enabledElement;
+
+  state.trackingIdentifiersByEnabledElementUUID[uuid] = {
+    trackingUniqueIdentifiers,
+    activeIndex,
+  };
+}
+
+function setActiveTrackingUniqueIdentifierForElement(
+  element,
+  TrackingUniqueIdentifier
+) {
+  const enabledElement = cornerstone.getEnabledElement(element);
+  const { uuid } = enabledElement;
+
+  const trackingIdentifiersForElement =
+    state.trackingIdentifiersByEnabledElementUUID[uuid];
+
+  if (trackingIdentifiersForElement) {
+    const activeIndex = trackingIdentifiersForElement.trackingUniqueIdentifiers.findIndex(
+      tuid => tuid === TrackingUniqueIdentifier
+    );
+
+    trackingIdentifiersForElement.activeIndex = activeIndex;
+  }
+}
+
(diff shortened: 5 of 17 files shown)


wayfarer3130:
@swederik and @sedghi  - Fairly simple PDF viewer.

extensions/dicom-pdf/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@ohif/extension-dicom-pdf",
```
sedghi, line 2: There is already a published package for `@ohif/extension-dicom-pdf` for v2, this is actually true for other extensions as well. We are not publishing to npm until v3 is on master? what happens to the old v2 extensions then? @swederik 
It think we should do better versioning of these extensions as well. A major version release of OHIF should reset everyone to the version base number ...

extensions/dicom-pdf/src/getSopClassHandlerModule.js
```diff
@@ -0,0 +1,66 @@
...
+
+const SOP_CLASS_UIDS = {
+  ENCAPSULATED_PDF: '1.2.840.10008.5.1.4.1.1.104.1',
+};
+
+const sopClassUids = Object.values(SOP_CLASS_UIDS);
+
+
+const _getDisplaySetsFromSeries = (instances, servicesManager, extensionManager) => {
+  console.log('Called getDisplaySetsFromSeries', instances);
```
sedghi, line 14: do we need this log? are we logging other sop class handlers?

extensions/dicom-pdf/src/getSopClassHandlerModule.js
```diff
@@ -0,0 +1,66 @@
...
+  const dataSource = extensionManager.getActiveDataSource()[0];
+  return instances
+    .map(instance => {
+      const { Modality, SOPInstanceUID, EncapsulatedDocument } = instance;
+      const { SeriesDescription, MIMETypeOfEncapsulatedDocument, } = instance;
+      const { SeriesNumber, SeriesDate, SeriesInstanceUID, StudyInstanceUID, NumberOfFrames } = instance;
+      console.log('Display PDF of type', MIMETypeOfEncapsulatedDocument, EncapsulatedDocument);
+      const pdfUrl = dataSource.retrieve.directURL({ instance, tag: 'EncapsulatedDocument', defaultType: MIMETypeOfEncapsulatedDocument });
```
sedghi, line 22: do we have directURL implementation for other dataSources? or we just throw error here if the dataSource is not dicomWebDataSource?

extensions/dicom-pdf/src/getSopClassHandlerModule.js
```diff
@@ -0,0 +1,66 @@
...
+export default function getSopClassHandlerModule({ servicesManager, extensionManager }) {
+  const getDisplaySetsFromSeries = instances => {
+    return _getDisplaySetsFromSeries(
+      instances,
+      servicesManager,
+      extensionManager
+    );
+  };
+
+  console.log("Registering", SOPClassHandlerName, sopClassUids);
```
sedghi, line 58: do we have logs for other sopHandlers too? if not let's remove this then

extensions/dicom-pdf/src/id.js
```diff
@@ -0,0 +1,7 @@
+const id = 'org.ohif.dicom-pdf';
+
+export default id;
+
+const SOPClassHandlerName = 'dicom-pdf';
+const SOPClassHandlerId = `${id}.sopClassHandlerModule.${SOPClassHandlerName}`;
```
sedghi, line 6: we don't need to insert extension scope in modules, ExtensionManager will handle it internally.

extensions/dicom-pdf/src/index.js
```diff
@@ -0,0 +1,93 @@
...
+        />
+      );
+    };
+
+    return [{ name: 'dicom-pdf', component: ExtendedOHIFCornerstonePdfViewport }];
+  },
+  getCommandsModule({ servicesManager }) {
+    return {
+      definitions: {
+        setToolActive: {
```
sedghi, line 66: What tools can be set to Active on pdf viewer? is it going to be any tool at all? if yes, can it not be setToolActive by the Cornerstone extension commandsManager? (since it is actually a cornerstoneTool's tool)

extensions/dicom-pdf/src/tools/modules/dicomVideoModule.js
```diff
@@ -0,0 +1,61 @@
+import cornerstone from 'cornerstone-core';
```
sedghi, line 1: This file is called dicomVideoModule. What does it do? Is it different than the dicom-video extension? why we are calling it module? (since we have extension's modules and that might be not ideal naming)

extensions/dicom-pdf/src/viewports/OHIFCornerstonePdfViewport.js
```diff
@@ -0,0 +1,44 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react';
```
sedghi, line 1: Do you use `useCallback` and others at all here?

modes/longitudinal/src/index.js
```diff
@@ -108,9 +117,15 @@ export default function mode({ modeConfiguration }) {
       'org.ohif.measurement-tracking',
       'org.ohif.dicom-sr',
       'org.ohif.dicom-video',
+      'org.ohif.dicom-pdf',
     ],
     hangingProtocols: [ohif.hangingProtocols],
-    sopClassHandlers: [dicomvideo.sopClassHandler, ohif.sopClassHandler, dicomsr.sopClassHandler,],
+    sopClassHandlers: [
+      dicomvideo.sopClassHandler,
```
sedghi, line 124: can we sort this? maybe starting with `ohif.` one just for readability

sedghi (commented):
Thanks, please see my comments on couple of files.

modes/longitudinal/src/index.js
wayfarer3130: No, you can't sort it, the order is important because the first one that grabs an object takes it out of the available list.

extensions/dicom-pdf/src/index.js
wayfarer3130: Nothing can be set active, but the wrapper expects to be able to call this.

modes/longitudinal/src/index.js
sedghi: So if a mode say `[ohif.sopClassHandler, dicomvideo.sopClassHandler]` instead of `[dicomvideo.sopClassHandler, ohif.sopClassHandler]` it breaks? I don't understand, can you maybe point to the source of the problem?

extensions/dicom-pdf/src/tools/modules/dicomVideoModule.js
wayfarer3130: Deleted it

modes/longitudinal/src/index.js
wayfarer3130: Yes, it breaks.  The problem is that ohif.sopClassHandler handles some of the same objects as dicomvideo.sopClassHandler, because it doesn't know about the available transfer syntax UID, so it will grab them first and add them to display sets.  Then, dicomvideo never gets called because there isn't anything for it to handle.

extensions/dicom-pdf/src/getSopClassHandlerModule.js
wayfarer3130: Working on updating this one - needs a change to dicomweb-client

extensions/dicom-pdf/src/id.js
wayfarer3130: Leaving as is because the full ID's are used for referencing things sometimes.

modes/longitudinal/src/index.js
wayfarer3130: As requested, changed to make each item declare the requested ordering, so that the mode doesn't need to know about the ordering of extensions inside itself.  In fact, the mode no longer has to have it's own sop class handler list, as there is a generic one it can inherit from.

extensions/dicom-pdf/package.json
wayfarer3130: If we need to have different versioning, then probably using @ohif3/... would be better, as then you can still release v2 ones, or else the major version of the release could be 3.x, and the rest could be the individual version stuff.

extensions/dicom-pdf/src/pdfConfig.js
```diff
@@ -0,0 +1,23 @@
+import id, { ViewportModuleId, SOPClassHandlerId } from './id.js';
+import ConfigPoint from 'config-point';
+
+// Register the handlers with the default mode settings
+const { modeConfig } = ConfigPoint.register({
+  modeConfig: {
```
wayfarer3130, line 6: @swederik and @sedghi This file shows how the an extension can register itself with the default mode settings.  Registering that way also registers with any mode that inherits from default.
The idea is that it defines the viewports object that gets injected into the viewports list, and then is accessible directly inside the modes instance.  This does NOT prevent the mode from customizing itself to exclude defaults, or to handle them in some other way.
This also injects the get sop class handler configuration - not the sop class handler itself, but just the configuration,
although it would be possible to inject the actual handler as well, which might be easier than loading it via a service.

modes/longitudinal/src/index.js
```diff
@@ -13,25 +14,31 @@ const tracked = {
...
-const dicomsr = {
-  sopClassHandler: 'org.ohif.dicom-sr.sopClassHandlerModule.dicom-sr',
-  viewport: 'org.ohif.dicom-sr.viewportModule.dicom-sr',
-};
 
-const dicomvideo = {
-  sopClassHandler: 'org.ohif.dicom-video.sopClassHandlerModule.dicom-video',
-  viewport: 'org.ohif.dicom-video.viewportModule.dicom-video',
-}
+const { longitudinalMode } = ConfigPoint.register({
```
wayfarer3130, line 18: @swederik and @sedghi  - note how the mode itself is configured with config point, which was necessary in order to allow injecting viewports and sop class handlers.  This means that the mode is no longer a global object, and can be added/created dynamically by choosing what code to import.    This is the other piece that I wanted Erik to see.
Other than these two items, I think the rest can be reviewed by other people.

extensions/dicom-pdf/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "version": "0.0.1",
```
wayfarer3130, line 3: @swederik Talking with Alireza, we discussed just changing all of the extensions to 3.x.x versions.  Will have a separate PR for that.

platform/viewer/src/routes/Mode/modeConfig.js
```diff
@@ -0,0 +1,61 @@
...
+          namespace: tracked.viewport,
+          displaySetsToDisplay: [ohif.sopClassHandler],
+        },
+        {
+          namespace: dicomsr.viewport,
+          displaySetsToDisplay: [dicomsr.sopClassHandler],
+        },
+      ],
+    },
+    sopClassHandlers: {
```
wayfarer3130, line 41: Add documentation on this one as to what this means

extensions/dicom-pdf/src/pdfConfig.js
JamesAPetts: Hmm, interesting.

My wonder is what if I want to use just the sop class handler, but not the viewport at all from the extension (I want to use my own 100% custom viewport for this use case).

Currently with the way modes work I would just create a mode config that uses the SOPClassHandler and then my own viewport from a different extension.

I see how defaults could be useful, but I really don't want a user to have to define a ConfigPoint configuration if they aren't doing something complex (i.e. just to override defaults), so it would be great if we could still just define a mode as is (just as a dumb object that works via composition).

Then again, it would be really cool if you could add an extension and easily set up default behaviour...

Could you comment on how we approach these points in this model?

JamesAPetts: Another thought is that a limitation of this approach is that mode schema is leaking into extension schema.

If mode schema changes, extensions now all have to change. Currently, extensions only have to oblige to comply to the extension contract (getPanelModule, getSopClassHandlerModule, etc), and the mode schema is decoupled so that it can change how it consumes these extensions freely without everyone's extensions changing.

This will be important in the OHIF vision of a large scale ecosystem, whereby many if not all of the best extensions are third party.

modes/longitudinal/src/index.js
```diff
@@ -13,25 +14,31 @@ const tracked = {
...
+      // Can customize sopClassHandlers here if desired
+    },
+  },
+});
 
-export default function mode({ modeConfiguration }) {
-  return {
-    // TODO: We're using this as a route segment
-    // We should not be.
+const { longitudinalMode: updatedLongitudinalMode } = ConfigPoint.register({
```
JamesAPetts, line 36: Doesn't appear this is used yet, but what was the intention? Can we remove this if its not used or is there more to come in this PR?

JamesAPetts (commented):
Some comments for discussion.

One overarching comment, not directly related to this PR, is I'm not sure if we should even be contributing these back to the core repo. We should probably keep the framework as light as possible (only cornerstone + default extension?), and everything else should be an extension that lives in its own repo.

I get its a mono-repo, but I envision there being 100s of extensions, so maybe we should tackle this one early... food for thought.

wayfarer3130:
> Some comments for discussion.
> 
> One overarching comment, not directly related to this PR, is I'm not sure if we should even be contributing these back to the core repo. We should probably keep the framework as light as possible (only cornerstone + default extension?), and everything else should be an extension that lives in its own repo.
> 
> I get its a mono-repo, but I envision there being 100s of extensions, so maybe we should tackle this one early... food for thought.

Great point, and one that I'm specifically trying to address with config-point.  The design of the config-point is to decouple what is getting configured from the configuration that gets puts there, while still allowing the default/basic configuration to be included.  Whether or not pdf and video are included in core I don't really care either way, but I don't want to have to have n^2 modes where each one needs to know/have deep understanding of the modes in order to get them working. 
For example, suppose we have a mode without PDF and video, then do we need the same mode + video, the same mode + pdf, the same mode with video + pdf?  That gets ugly already, with just 2 extensions.  Sure, some modes will be orthogonal, like not having a cornerstone 2d mode and a cornerstone 3d mode combined, but even with just that, it would add an extra 4 modes.  What config point allows is to:
1. Have a basic mode (longitudinal for example)
2. Add in the default providers
3. Just load the desired extensions
4. Extensions register themselves with config-point as to what they provide for things like getSopClassHandler, viewport etc
5. Modes simply get extended and are have the additional functionality mixed in.

Then, you can have individual theme configs that simply say  add video and add pdf, or do neither.  

That vision isn't quite realized in this commit - I didn't want to go quite that far initially, but in static-wado, it is realized that far as you can load third party deployment mechanisms as well as third party custom services (eg custom image conversions) by defining them in the static-wado configuration file.    Adding a new module adds both default configuration to multiple areas and default code.

modes/longitudinal/src/index.js
wayfarer3130: Added a comment inline about why a second register occurs, and removed the actual second variable, which isn't needed.

wayfarer3130: > Another thought is that a limitation of this approach is that mode schema is leaking into extension schema.
The mode schema is already coupled to how it consumes the extensions, as they are explicitly listed in the mode itself with the viewport being linked by name to the sop class handler, and the list of sop class handlers being listed - I would argue that config point actually reduces the coupling because the extension only needs to understand the part of the mode schema that it is enhancing.  The extension does specify that it modifies the mode by adding the schema and the viewport, but that is all it specifies, for example, the pdf extension only lists the viewport and sop class handlers sections:
modeConfig: {
    panels: {
      viewports: {
        pdf: {
          id,
          namespace: ViewportModuleId,
          displaySetsToDisplay: [SOPClassHandlerId],
        },
      },
    },
    sopClassHandlers: {
      pdf: { id: SOPClassHandlerId, priority: 5000 },
    },

Now, suppose we have another mode that doesn't want any default handling and uses a completely new schema for extensions.  The new mode is just defined the old way, and explicitly lists all the extensions it wants included in whatever schema format it wants.  Yes, it is now tightly coupled to the extensions it uses, but it still works just fine.  
Or alternatively, suppose a new global schema is defined for new modes - it can use config point to define the new schema, and extensions that get updated can plugin to the new schema, OR the new schema can specifically list the new default extensions it wants to include (or any combination of the two).  The extensions supporting the new schema can now be used with both the old and new schemas by just listing both configurations.  Config point is designed to support shared configuration, but it isn't required anywhere, just that it makes the changes a bit more standard and more decoupled when it does get used.  If one were clever, it would be possible to write extensions that work on both OHIF v2 and OHIF v3 by registering different functions into config-point.

JamesAPetts: > The new mode is just defined the old way, and explicitly lists all the extensions it wants included in whatever schema format it wants. 

Thanks for this point especially, I think this is something I didn't realise. I'll mull over your thoughts.

JamesAPetts:
> Great point, and one....

All good points, I think all this really depends on how much modes want to be shared and used. Additionally modes are currently static because the examples ones were designed to be simple, but as you say they may need to be configurable by the client/some external resource, so its good you are thinking about this problem.

wayfarer3130:
@sedghi and @swederik  - I've removed the config point mode configuration from this.  I've left the update to config-point 0.4.3, as the earlier config-point version needs some updating and is committed to v3-stable already.

extensions/dicom-pdf/src/id.js
```diff
@@ -0,0 +1,9 @@
+const Name = 'dicom-pdf';
+const id = `org.ohif.${Name}`;
+
+export default id;
+
+const SOPClassHandlerId = `${id}.sopClassHandlerModule.${Name}`;
```
sedghi, line 6: we don't need to have the module name here. It is handled inside the ExtentionManager

sedghi: same for viewportModuleId, the module itself doesn't need an Id, the components it provides should have a name. `deafult` extension has a lot of these you can take a look

extensions/dicom-pdf/src/id.js
```diff
@@ -0,0 +1,9 @@
...
+const ViewportModuleId = `${id}.viewportModule.${Name}`;
```
sedghi, line 7: I don't think the viewportModuleId is used anywhere either. I searched an there is none. dicom-video has this also which should be deleted too

extensions/dicom-video/src/viewports/OHIFCornerstoneVideoViewport.js
```diff
@@ -1,51 +1,36 @@
...
-  servicesManager,
-  extensionManager,
 }) {
-  const {
-    DisplaySetService,
-    MeasurementService,
-    ToolBarService,
-  } = servicesManager.services;
   const { videoUrl } = displaySet;
   const mimeType = "video/mp4";
```
sedghi, line 8: maybe move this outside of the function and make it all caps since it is constant

extensions/dicom-pdf/src/id.js
sedghi: You can resolve this

platform/viewer/src/index.js
```diff
@@ -34,11 +36,21 @@ const appProps = {
...
-
-/** Render */
-ReactDOM.render(app, document.getElementById('root'));
+const { defaultTheme="theme" } = appProps.config;
+/**
+ * Load the default theme settings,
+ * and then render the app in the then block so that themes get
+ * loaded early enough to modify the initial render.
+ */
+loadSearchConfigPoint(defaultTheme, '/theme', 'theme').then(() => {
```
sedghi, line 49: tagging @JamesAPetts for core

extensions/dicom-video/src/viewports/OHIFCornerstoneVideoViewport.js
wayfarer3130: Rather leave it where it is since it is supposed to be a computed field, just that the code to computed it isn't yet added.

platform/viewer/src/index.js
wayfarer3130: This is only changing the location of existing code to allow upgrading the version, but config-point is already in v3-stable.

JamesAPetts: I know @swederik dind't like the name theme?

So loadSearchConfigPoint comes from config-point, I thought we were removing it as a dependency of this PR? I get it was already in v3 stable, but aren't we now using it?

Now we decided v32-stable is the home of `config-point` for now, can we now remove `config-point` entirely from v3-stable?

sedghi: @wayfarer3130 This is the only comment left, can we please address this and merge this outstanding branch?

PR #2731, opened 2022-03-07
Fix palette colour display in OHIF v3

wayfarer3130:
The palette colour fetch was broken in OHIF v3.  This is working now, being decoding in OHIFv3 into the format required by cornerstone.


platform/core/src/utils/metadataProvider/fetchPaletteColorLookupTableData.js
```diff
@@ -4,162 +4,69 @@ import str2ab from '../str2ab';
...
-  });
-}
-
 /**
- * Simple cache schema for retrieved color palettes.
+ * Gets the palette color data for the specified tag - red/green/blue,
+ * either from the given UID or from the tag itself.
+ * Returns an array if the data is immediately available, or a promise
+ * which resolves to the data if the data needs to be loaded.
+ * Returns undefined if the palette isn't specified.
```
wayfarer3130, line 12: Add docs here.

PR #2756, opened 2022-03-28
feat(OHIFv32):Extend and explain the config-point configuration.

wayfarer3130:
Add helper methods to make it more obvious what developers are doing when adding new configuration points.
@swederik @JamesAPetts @sedghi 
I created this PR to address some of the issues in our earlier meeting.


platform/docs/docs/platform/services/config-point-service.md
```diff
@@ -89,6 +94,44 @@ data, defaulting to the base behaviour of the application, allowing it to
 be exposed later to enhancements in a declarative fashion.  The remaining
 sections below simply expand on the idea with some more advanced concepts.
 
+## Why isn't Config-Point a general service?
```
wayfarer3130, line 97: This addresses the comments about config-point being provided as the same type of service as the other services.  The argument is basically that it is too late to be provided at that point, but please read the description and see if it makes sense.

PR #2763, opened 2022-03-30
feat(OHIFv32):Add the configurable site settings menu back in

wayfarer3130:
This adds a configurable site settings menu, like OHIF v2 had, to be able to select the site settings from a list.  Unlike v2, this is added as a themeable configuration value.


wayfarer3130:
@JamesAPetts - this is another example of how to add configuration at runtime, in this case to allow the site findings to be configured.  It is live at one site in OHIF v2, and is for use as a demo at another site.
I still need to add e2e tests, documentation and fix the save/restore functionality.

PR #2766, opened 2022-04-01
feat(OHIFv32):Adding a 2x2 mode

wayfarer3130:
This makes a number of changes to enable themeable hanging protocols.  That allows for custom definitions independently of the mode, when the mode wants to allow that, OR it allows the existing hanging protocols to be defined.

### DIFF
--- a/extensions/cornerstone/src/initWADOImageLoader.js
+++ b/extensions/cornerstone/src/initWADOImageLoader.js
@@ -6,8 +6,9 @@ import { errorHandler } from '@ohif/core';
 let initialized = false;
 
 function initWebWorkers() {
+  // TODO - increase the limit of 3 web worker once the bug in the decompression is fixed
   const config = {
-    maxWebWorkers: Math.max(navigator.hardwareConcurrency - 1, 1),
+    maxWebWorkers: Math.min(Math.max(navigator.hardwareConcurrency - 1, 1),3),
     startWebWorkersOnDemand: true,
     taskConfiguration: {
       decodeTask: {
--- a/extensions/default/src/getHangingProtocolModule.js
+++ b/extensions/default/src/getHangingProtocolModule.js
@@ -1,188 +1,313 @@
-const hangingProtocolName = 'petCT';
+import ConfigPoint from 'config-point';
 
-const petCTProtocol = {
-  id: 'PET/CT',
-  locked: true,
-  hasUpdatedPriorsInformation: false,
-  name: 'PET/CT',
-  createdDate: '2021-02-23T18:32:42.849Z',
-  modifiedDate: '2021-02-23T18:32:42.849Z',
-  availableTo: {},
-  editableBy: {},
-  protocolMatchingRules: [
-    {
-      id: 'wauZK2QNEfDPwcAQo',
-      weight: 1,
-      attribute: 'StudyInstanceUID',
-      constraint: {
-        equals: {
-          value: '1.3.6.1.4.1.25403.345050719074.3824.20170125112931.11',
-        },
+const hangingProtocolName = 'default';
+
+/**
+ * Define some hanging protocols that can be activated by setting the priority
+ * See hangingProtocols.json5 for examples.
+ */
+const hangingProtocolConfig = ConfigPoint.createConfiguration("hangingProtocolConfig", {
+  protocols: {
+    configOperation: 'sort',
+    sortKey: 'priority',
+    value: [
+      {
+        id: '1x2',
+        // Note the priority is null, so this will get excluded unless the priority is set
+        priority: null,
+        locked: true,
+        hasUpdatedPriorsInformation: false,
+        name: '1x2',
+        createdDate: '2021-11-01T18:32:42.849Z',
+        modifiedDate: '2021-11-01T18:32:42.849Z',
+        availableTo: {},
+        editableBy: {},
+        protocolMatchingRules: [
+          {
+            id: 'NumberOfStudyRelatedSeries>1',
+            weight: 10,
+            attribute: 'NumberOfStudyRelatedSeries',
+            constraint: {
+              greaterThan: {
+                value: 1,
+              },
+            },
+            required: true,
+          },
+        ],
+        stages: [
+          {
+            name: 'OneByTwo',
+            viewportStructure: {
+              type: 'grid',
+              properties: {
+                rows: 1,
+                columns: 2,
+              },
+            },
+            viewport: {},
+          },
+        ],
+        numberOfPriorsReferenced: 0,
       },
-      required: true,
-    },
-  ],
-  stages: [
-    {
-      id: 'hYbmMy3b7pz7GLiaT',
-      name: 'oneByTwo',
-      viewportStructure: {
-        type: 'grid',
-        properties: {
-          rows: 2,
-          columns: 2,
-        },
+      {
+        id: '2x2',
+        priority: null,
+        locked: true,
+        hasUpdatedPriorsInformation: false,
+        name: '2x2',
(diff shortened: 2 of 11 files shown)


platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -54,6 +66,24 @@ class HangingProtocolService {
...
     protocols.forEach(protocol => {
       if (this.protocols.indexOf(protocol) === -1) {
+        protocol.id = protocol.id || protocol.name;
+        // Automatically compute some number of attributes if they
+        // aren't present.  Makes defining new HPs easier.
+        protocol.name = protocol.name || protocol.id;
+        const { stages } = protocol;
+
+        // Generate viewports automatically as required.
+        stages.forEach(stage => {
```
wayfarer3130, line 76: This is just helper code to make it easier to define hanging protocols - it could all be done manually

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -213,9 +243,10 @@ class HangingProtocolService {
     // for each of the viewport matching procedures
 
     // Loop through each viewport
+    const alreadyMatched = [];
```
wayfarer3130, line 246: @ladeirarodolfo This is the change that makes it possible to define hanging protocols that match subsequent images in a series.  The problem was that there wasn't a good way to say "match the SECOND display set" that matches.  So it is a fairly important fix to enable 2x2 layouts to actually work effectively.

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -314,7 +346,12 @@ class HangingProtocolService {
 
       highestStudyMatchingScore = studyMatchDetails.score;
 
+      // TODO - use display sets instead of series, as that will use the
+      // correct re-grouped information.
```
wayfarer3130, line 350: This is probably another thing to fix - it is quite important to use display sets, otherwise you will never see anything except the first split of a series.

platform/core/src/services/HangingProtocolService/lib/validator.js
```diff
@@ -36,4 +36,12 @@ validate.validators.endsWith = function (value, options, key) {
   }
 };
 
+validate.validators.greaterThan = function (value, options, key) {
```
wayfarer3130, line 39: Allows checking for NumberOfSeries >=1

platform/core/src/services/HangingProtocolService/themeProtocolProvider.js
```diff
@@ -0,0 +1,116 @@
...
+          name: '1x2',
+          createdDate: '2021-11-01T18:32:42.849Z',
+          modifiedDate: '2021-11-01T18:32:42.849Z',
+          availableTo: {},
+          editableBy: {},
+          protocolMatchingRules: [
+            {
+              id: 'NumberOfStudyRelatedSeries>1',
+              weight: 10,
+              attribute: 'NumberOfStudyRelatedSeries',
```
wayfarer3130, line 20: Note that this SHOULD be NumberOfDisplaySets rather than number of study related series, as a series like a DICOM SR will get a box for it, but that shouldn't actually cause the rule to be activated.  But that is hard to arrange for because
the display set calculations aren't necessarily completed at this point.

platform/core/src/services/HangingProtocolService/themeProtocolProvider.js
```diff
@@ -0,0 +1,116 @@
...
+                properties: {
+                  rows: 2,
+                  columns: 2,
+                },
+              },
+            },
+          ],
+          numberOfPriorsReferenced: 0,
+        },
+        {
```
wayfarer3130, line 80: The 2x3 mode should probably be disabled or included in a separate file, as many people don't want it.

platform/viewer/public/theme/theme.json5
```diff
@@ -1,3 +1,38 @@
 {
-  // Currently empty because nothing is configurable yet with themes
+  // This is an example hanging protocol definition.  See theme-configuration.md
+  // for more details on custom configurations:
+  ThemeProtocols: {
```
wayfarer3130, line 4: We should remove this for now, as it just replicated other stuff.

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -6,8 +6,9 @@ import { errorHandler } from '@ohif/core';
 let initialized = false;
 
 function initWebWorkers() {
+  // TODO - increase the limit of 3 web worker once the bug in the decompression is fixed
   const config = {
-    maxWebWorkers: Math.max(navigator.hardwareConcurrency - 1, 1),
+    maxWebWorkers: Math.min(Math.max(navigator.hardwareConcurrency - 1, 1),3),
```
wayfarer3130, line 11: This is relevant right now because it crashes the browser in 2x2 mode otherwise.

extensions/default/src/getHangingProtocolModule.js
```diff
@@ -1,188 +1,313 @@
...
-                  windowWidth: 500,
-                  windowCenter: 500,
-                },
+const hangingProtocolName = 'default';
+
+/**
+ * Define some hanging protocols that can be activated by setting the priority
+ * See hangingProtocols.json5 for examples.
+ */
+const hangingProtocolConfig = ConfigPoint.createConfiguration("hangingProtocolConfig", {
```
wayfarer3130, line 9: This defines four different hanging protocols that can be activated - see the hangingProtocols.json5 file for an example.

platform/core/src/services/HangingProtocolService/HPMatcher.js
```diff
@@ -36,7 +36,13 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
 
     // Create a single attribute object to be validated, since metadataInstance is an
     // instance of Metadata (StudyMetadata, SeriesMetadata or InstanceMetadata)
-    const attributeValue = metadataInstance[attribute];
+    let attributeValue = metadataInstance[attribute];
```
wayfarer3130, line 39: This bit just reduces the number of repeated fields required to define hanging protocols.

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -54,6 +54,24 @@ class HangingProtocolService {
   addProtocols(protocols) {
     protocols.forEach(protocol => {
       if (this.protocols.indexOf(protocol) === -1) {
+        protocol.id = protocol.id || protocol.name;
```
wayfarer3130, line 57: More reduction in how many fields are required.

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
wayfarer3130: @sedghi  - can we get this file into your updated hanging protocol versions?  It just makes it easier to define HPs

wayfarer3130:
@sedghi and @JamesAPetts 
The changes here are quite a bit smaller now, being much closer to the previously defined hanging protocols.  There are really only 3 things done here:
1. Define some new hanging protocols
2. Make it easier to declare HPs by repeating/calculating values when possible
3. Allow a matching rule to match the "next" instance

I'd like to merge this change into OHIFv32 tomorrow, and I'm hoping that Alireza will use some of it as a base for some of the new Cornerstone3d hanging protocols, but if not, I will just merge those changes into OHIFv32 and manage the merge between those.

sedghi:
@wayfarer3130 sure, I will look into this after I have the new engine setup.

PR #2767, opened 2022-04-02
Fix/v3 dicomweb client prefixs

kzgrzendek:
## Dicomweb requests fixes

- Refactoring the usage of the dicomweb client to keep one client for all the dicomweb-related operations (still keeping the staticWado client appart)
- All relevant details are mentioned here -> https://community.ohif.org/t/qido-wado-endpoints-handling-in-ohif-v3/24/8
- Updated config file templates and docs
- Also updated the base node image in the dockerfile, the one from the v3/stable was causing issues in the image build process
- Slight changes in the DicomWeb Datasource config (needs to specify qido, wado and stow prefixes in the config file)
- @swederik this is what we talk last month :) don't hesitate to tell me if you need me to modify anything


kzgrzendek:
> @wayfarer3130 can you please review this after you come back from vacation?
> 
> @kzgrzendek why we are changing qidoRoot to qidoPrefixURL? (and others)

Hi @sedghi! The goal is to adapt to the way the dicomweb-client we use is working -> https://github.com/dcmjs-org/dicomweb-client/blob/447dfe040b454037ed86d55a1ad35304dbb5041e/src/api.js#L62

It expects a base URL,  then different prefixs for the dicomweb endpoint you want to consume. That way, we use the same dicom web client for QIDO, WADO and STOW. 

I've detailed my actions in a series of (too)  long posts in the OHIF Community forum -> https://community.ohif.org/t/qido-wado-endpoints-handling-in-ohif-v3/24/8

PS : I've updated the MR with the laste version of the ohif-v3 branch to resolve conflicts. I've also fixed the Docker that wasn't building at all.

wayfarer3130:
For a single web service, the DICOMweb definition in part 18 specifies:
The Studies Resource enables a user agent to store, retrieve, update, and search an origin server for DICOM Studies, Series, and Instances, along with their /metadata, /rendered, and /thumbnail variants; as well as Frames and Bulkdata.

That actually means that if the studies query response returns a response, you are supposed to be able to fetch data etc from the same path on it.  Now, you can simply say that they are three unrelated services, but the standard's compliant method of doing that is to have the study response include the Tag RetrieveURL included in the response.  Note that this tag MAY be included for series level queries and be different for different series.  It is also allowed to return the RetrieveAE field, and key off the AE name for the next retrieve level.  You can also have some proprietary method to determine the next level, such as having fixed AE names (aka data source names), or the custom prefixes you suggested.

It would be very useful instead of having the logic to determine the paths be embedded in the dicomweb client, to have that determined by the data source provider, and to simply define multiple data sources in the case that they reference different URLs.  That way, one could define a WADO URI retriever, a WADO RS one on several paths, and some sort of method to determine which data source is used for a given level/mapping.  That is, unfortunately, harder to implement.  I think we can probably get by with fewer changes by having some of the values be optional.  Let me try to make some suggestions in the files themselves.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -37,21 +37,23 @@ const metadataProvider = classes.MetadataProvider;
...
 /**
  *
- * @param {string} name - Data source name
- * @param {string} wadoUriRoot - Legacy? (potentially unused/replaced)
- * @param {string} qidoRoot - Base URL to use for QIDO requests
- * @param {string} wadoRoot - Base URL to use for WADO requests
- * @param {boolean} qidoSupportsIncludeField - Whether QIDO supports the "Include" option to request additional fields in response
- * @param {string} imageRengering - wadors | ? (unsure of where/how this is used)
- * @param {string} thumbnailRendering - wadors | ? (unsure of where/how this is used)
+ * @param {string} baseUrl - The root URL for DICOMWeb calls
```
wayfarer3130, line 40: Have the names matche the dicomweb client definition - that is, it should take url as the base url parameter, and ONLY add wadoURLPrefix etc if specified.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -37,21 +37,23 @@ const metadataProvider = classes.MetadataProvider;
...
+ * @param {bool} enableStudyLazyLoad - Request series meta async instead of blocking
+ * @param {bool} supportsWildcard - Whether the server supports wildcards calls (i.e. DCM4CHEE)
  * @param {bool} supportsReject - Whether the server supports reject calls (i.e. DCM4CHEE)
- * @param {bool} lazyLoadStudy - "enableStudyLazyLoad"; Request series meta async instead of blocking
+ * @param {bool} staticWado - Wether the DICOMWeb client is a static wado one fot test purposes
  * @param {string|bool} singlepart - indicates of the retrieves can fetch singlepart.  Options are bulkdata, video, image or boolean true
  */
 function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
   const {
-    qidoRoot,
```
wayfarer3130, line 53: Just extract the ...options from the parameter, so that you can pass any additional options in as required.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -60,31 +62,19 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
...
-    singlepart,
-    headers: UserAuthenticationService.getAuthorizationHeader(),
-    errorInterceptor: errorHandler.getHTTPErrorHandler(),
-  };
-
-  const wadoConfig = {
-    url: wadoRoot,
-    staticWado,
-    singlepart,
+  const dicomWebClientConfig = {
```
wayfarer3130, line 65: Just pass in options directly, or else use 
const wadoConfig = {
...options,
<any custom over-rides}
}

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -220,16 +207,16 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
           (hasAccept ? '' : (hasQuery ? '&' : '?') + `accept=${defaultType}`);
         if (BulkDataURI.indexOf('http') === 0) return acceptUri;
         if (BulkDataURI.indexOf('/') === 0) {
-          return wadoRoot + acceptUri;
+          return `${baseUrl}/${wadoURLPrefix}/${acceptUri}`;
```
wayfarer3130, line 210: Create a function on the data access object which gets the URL for a given query type, defaulting to the base URL.  That way there is just a single query generator.

wayfarer3130: You could put the getter function on the config object, so that it can be used for doing things like generating the wadors URI's.  Something like
if( !config.getUrl ) {
  config.getUrl = function(typeId, ...additions)
to get the URL.  It should handle query parameters defined in the base URL or in the suffixes as well....   typeId would be 'wado' or 'qido' etc.

extensions/default/src/DicomWebDataSource/utils/getImageId.js
```diff
@@ -13,7 +13,7 @@ function buildInstanceWadoUrl(config, instance) {
 
   const paramString = params.join('&');
 
-  return `${config.wadoUriRoot}?${paramString}`;
+  return `${config.baseUrl}/${config.wadoURLPrefix}?${paramString}`;
```
wayfarer3130, line 16: Use the function added to the config to get the query parameters/setup, eg make it:
config.getUrl('wado', `?{paramString}`)

platform/core/src/utils/addServer.test.js
```diff
@@ -6,8 +6,9 @@ describe('addServers', () => {
       {
         name: 'DCM4CHEE',
         wadoUriRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/wado',
-        qidoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
-        wadoRoot: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
+        qidoUrlPrefix: 'https://server.dcmjs.org/dcm4chee-arc/aets/DCM4CHEE/rs',
```
wayfarer3130, line 9: Should all be replaced with just uri.  Then, if necessary, specific versions can include the prefix definitions as needed.

wayfarer3130 (changes requested):
This should be more consistent with the dicomweb client if that is the direction we want to take - that is, use url and then only define the prefixes as necessary.  Encapsulate the logic to do all that.

sedghi:
@kzgrzendek Thanks for this PR, any one free you have to work on this?

kzgrzendek:
> @kzgrzendek Thanks for this PR, any one free you have to work on this?

Hello @sedghi, not at the moment. I can have a look at it on my free time, but atm I'm very busy on other topics, and unfortunately not allowed to do it on my working hours for legal reasons. If you have people available for doing the requested changes, I'd be glad to assist! If not, I'll try to have a look at it in the following days/week but can't give any estimations unfortunately

PR #2773, opened 2022-04-06
feat measurement visibility

ladeirarodolfo:
## Includes
- Visibility icon at measurement panel
- Visibility functionality
- Hotkeys assigned for toggling visibility: m for current active measurement, shift+m for all measurements

Screen Shot 2022-04-12 at 17 10 24

### PR Checklist

### DIFF
--- a/extensions/cornerstone/src/OHIFCornerstoneViewport.js
+++ b/extensions/cornerstone/src/OHIFCornerstoneViewport.js
@@ -236,9 +236,14 @@ function OHIFCornerstoneViewport({
         isOverlayVisible={true}
         loadingIndicatorComponent={ViewportLoadingIndicator}
         viewportOverlayComponent={props => {
+          const { imageIndex } = props;
+          const instances = displaySet.images || displaySet.other;
+          const image = instances && imageIndex > 0 &&
+            imageIndex <= instances.length && instances[imageIndex - 1];
+          const viewportProps = { ...props, cine, isPlaying, frameRate, image };
           return (
             <ViewportOverlay
-              {...props}
+              {...viewportProps}
               activeTools={ToolBarService.getActiveTools()}
             />
           );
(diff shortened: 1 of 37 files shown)


wayfarer3130:
Generally the changes are mostly ok.  A couple of things that I think would make it a bit more useful:
1. Remove the icon from the table to toggle visibility.  Having too many icons will be a problem.  If we can figure out how to make right-click in the table bring up the context menu, then that would be useful.
2. If the user clicks the measurement for any reason, make it visible
3. Have a global hide-measurements command, acting on all objects (similarly a global show all)

I think the user experience is better that way.  We can talk tomorrow about how it works.

ladeirarodolfo:
> Generally the changes are mostly ok. A couple of things that I think would make it a bit more useful:
> 
> 1. Remove the icon from the table to toggle visibility.  Having too many icons will be a problem.  If we can figure out how to make right-click in the table bring up the context menu, then that would be useful.
> 2. If the user clicks the measurement for any reason, make it visible
> 3. Have a global hide-measurements command, acting on all objects (similarly a global show all)
> 
> I think the user experience is better that way. We can talk tomorrow about how it works.

1. We definitely can make it, shall we? I dont disagree but as a user I would prefer to have the icon already displayed instead of having to do two actions to hide it (one to open menu other to hide/show)
2. Measurement you mean on viewport? or on the table? If first one is the case: This is really doable, but, the solution should go under image library, otherwise it will be a workaround. Shall we? I can not estimate the effort yet. What about second one: what should happen if user clicks on measurement item (on table)?
3. Alright, and will there be a icon for that? Where to place it.

ladeirarodolfo:
@wayfarer3130  code is updated with what we discussed internally. Can you take a look please?

sedghi:
Do you need this? If so can you please rebase on v3-stable?

sedghi:
will revisit this here 
https://github.com/OHIF/Viewers/issues/4095

PR #2791, opened 2022-05-02
JSON server configuration launch

dsmith1547:
### PR Checklist


extensions/default/src/DicomWebProxyDataSource/index.js
```diff
@@ -0,0 +1,74 @@
+import { IWebApiDataSource } from '@ohif/core';
+import { createDicomWebApi } from '../DicomWebDataSource/index';
+
+function createDicomWebProxyApi(
```
sedghi, line 4: Can you add couple of lines in the documentation of this function and also explain the schema for which the data source is being used for? what does the url look like as an exmaple etc.

sedghi: For instance, it seems like it needs `servers` info to be present at the url

extensions/default/src/DicomWebProxyDataSource/index.js
```diff
@@ -0,0 +1,74 @@
...
+        const response = await fetch(url);
+        let data = await response.json();
+
+        // there seem to be a couple of variations of the case for this parameter
+        const queryStudyInstanceUIDs =
+          query.get('studyInstanceUIDs') || query.get('studyInstanceUids');
+        if (!queryStudyInstanceUIDs) {
+          console.error(`No studyInstanceUids in request for '${name}'`);
+        }
+        if (data.servers && queryStudyInstanceUIDs) {
```
sedghi, line 29: Shouldn't we throw here if `data.servers` don't exist? The rest of implementation relies on it

sedghi (changes requested):
Great addition, had some comments. Tagging @JamesAPetts for review as well

sedghi:
@dsmith1547 Also forgot to mention we need something similar to the docs here https://v3-docs.ohif.org/configuration/dataSources/dicom-json for explaining the proxy and what it expects at the url etc.

extensions/default/src/DicomWebProxyDataSource/index.js
dsmith1547: I added a short comment to the class, but the (added) markdown documentation is probably a better way to see this

JamesAPetts (commented):
This functionality looks fine. But since this is for a particular niche usecase, I don't think it be should pushed to the default OHIF extension. If it is designed to be an open source contribution, that platforms other than Kheops may use, it should have its own repo, which is an extension, which houses the datasource.

Anything that is not on the ohif roadmap: https://ohif.org/roadmap/, should ideally in an extension rather than in the core repo.

If we need to change the cornerstone extension to accept optional headers, that feels like something that _should_ go core.

sedghi:
Looking back at this, I like this to be part of the OHIF upstream. Now that we are building the Google/Azure/AWS adapters uses this almost entirely. @dsmith1547 Can you please do that? thanks

sedghi (approved):
Looks great thanks

platform/viewer/public/config/default.js
```diff
@@ -24,6 +24,8 @@ window.config = {
     prefetch: 25,
   },
   // filterQueryParam: false,
+  // enable imageLoaderAcceptHeader configuration to force a reconfiguration of WADOImageLoader to support preflight request
```
jbocce, line 27: Do we still want these commented lines included?

extensions/default/src/DicomWebProxyDataSource/index.js
```diff
@@ -0,0 +1,84 @@
...
+ */
+function createDicomWebProxyApi(
+  dicomWebProxyConfig,
+  UserAuthenticationService
+) {
+  const { name } = dicomWebProxyConfig;
+  let dicomWebDelegate = undefined;
+
+  const implementation = {
+    initialize: async ({ params, query, url }) => {
```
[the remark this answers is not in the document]
dsmith1547: I think you're right - I likely "borrowed" this part of the impl from the DicomJSONDataSource, but it looks like it must always get it from the query param

platform/viewer/public/config/default.js
dsmith1547: nope

extensions/default/src/DicomWebProxyDataSource/index.js
```diff
@@ -0,0 +1,84 @@
...
+      let studyInstanceUIDs = [];
+
+      // there seem to be a couple of variations of the case for this parameter
+      const queryStudyInstanceUIDs =
+        query.get('studyInstanceUIDs') || query.get('studyInstanceUids');
+      if (!queryStudyInstanceUIDs) {
+        throw new Error(`No studyInstanceUids in request for '${name}'`);
+      }
+
+      let url = query.get('url');
```
jbocce, line 30: This should be `const` I think.

dsmith1547: Indeed, this no longer gets modified.

jbocce (approved):
Thanks for the updates!

PR #2792, opened 2022-05-04
fix: Crash/hang issues and a minor log issue

wayfarer3130:
Fixing some crash/hang issues in CS3D version of OHIF.


extensions/cornerstone-3d/src/initWADOImageLoader.js
```diff
@@ -13,7 +13,7 @@ function initWebWorkers(appConfig) {
   const config = {
     maxWebWorkers: Math.min(
       Math.max(navigator.hardwareConcurrency - 1, 1),
-      appConfig.maxNumberOfWebWorkers
+      appConfig.maxNumberOfWebWorkers || 3
```
wayfarer3130, line 16: This prevents a windows crash.

platform/ui/src/components/StudyBrowser/StudyBrowser.tsx
```diff
@@ -140,7 +140,10 @@ StudyBrowser.propTypes = {
               imageSrc: PropTypes.string,
               imageAltText: PropTypes.string,
               seriesDate: PropTypes.string,
-              seriesNumber: PropTypes.string,
+              seriesNumber: PropTypes.oneOfType([
```
wayfarer3130, line 143: This allows viewing more studies without having long stack traces.

platform/viewer/src/components/ViewportGrid.tsx
```diff
@@ -363,6 +363,8 @@ function _getViewportComponent(displaySets, viewportComponents) {
       return component;
     }
   }
+  console.warn("Didn't find a viewport component handling", SOPClassHandlerId);
```
wayfarer3130, line 366: This prevents crashing when viewing a study with DICOM SR in it.

extensions/cornerstone-3d/src/initWADOImageLoader.js
sedghi: This always return the first value if `appConfig.maxNumberOfWebWorkers` is not 0, which is not in the default.js

Maybe I'm missing the point here

wayfarer3130: It returns:
This defaults maxNumberOfWebWorkers to 3, and then returns the smaller of:
hardwareConcurrency -1, maxNumberOfWebworkers
So in a default situation where max number of webworkers isnt set, and hardware concurrency is high, it will return 3.
On a phone/tablet, it will probably return hardware concurrency (probably 1 or 2).
This prevents a browser crash on Chrome on windows.

sedghi: but if it is set and it is let's say 10, 10||3 is 10, but don't you want it to be 3?

wayfarer3130:
No, if they set it specifically to 10 then it can stay as 10.

platform/ui/src/components/StudyBrowser/StudyBrowser.tsx
sedghi: This is a comment in general: can't we just put these valuable comments (that are in Github pull request) as part of the code comment? If they need to be explained for a reviewer, a code reader benefits from the comment as well. You don't need to add the comment for this PR, but talking generally.

platform/viewer/src/components/ViewportGrid.tsx
sedghi: Same for this comment

Issue #2807, opened 2022-05-19
Preset Windowing Bone and Brain inverted in v3-stable


salimkanoun:
In v3 stable there is an error in the preset of windowing: 

Currently we have 
Bone 80/40
Brain 2500/480

It should be
Brain 80/40
Bone 2500/480

PR #2808, opened 2022-05-19
Fix : Window level value for Bone and Brain

Celian-abd:
Here is a fix related to this issue : https://github.com/OHIF/Viewers/issues/2807

The preset of Window level for bone and brain is wrong : 
Before : 

image

After : 

image

I only reversed the value between Bone and Brain.


sedghi (approved):
Great catch

PR #2811, opened 2022-05-20
fix: various bugs with regard to tracking workflow

sedghi:
- Fix UI/UX discrepancies in the measurement panel between v3-stable and cornerstone3D
- Fix ViewportPane layout different bw v3-stable and cornerstone3D
- Remove using cursors by default
- Fix ArrowAnnotate SRDisplayTool
- Add support for legacy cornerstone SR display and hydration


extensions/cornerstone-dicom-sr/src/utils/addMeasurement.ts
```diff
@@ -92,6 +105,36 @@ function _getRenderableData(GraphicType, GraphicData, imageId) {
 
         renderableData.push(worldPos);
       }
+
+      // In Cornerstone-Legacy only one point is saved in the SR and not the start and
```
JamesAPetts, line 109: This is the same in Cornerstone3D, no? As in we only store one point.

JamesAPetts (approved):
LGTM, one comment on your comment.

extensions/cornerstone-3d/src/init.js
```diff
@@ -36,6 +37,9 @@ export default async function init({
   await cs3DInit();
   initCornerstoneTools();
 
+  // Don't use cursors in viewports
+  Settings.getRuntimeSettings().set('useCursors', false);
```
JamesAPetts, line 41: Shouldn't we set this based on extension configuration?

...aside: are we even using extension configuration. How do we use it with the current ohif-cli dev flow?

extensions/cornerstone-3d/src/utils/measurementServiceMappings/EllipticalROI.js
```diff
@@ -199,29 +199,27 @@ function getDisplayText(mappedAnnotations, displaySet) {
...
-  //     displayText.push(
-  //       `mean: ${roundedMean} <small>${unit}</small> (S:${SeriesNumber})`
-  //     );
-  //   }
-  // });
+  mappedAnnotations.forEach(mappedAnnotation => {
+    const { mean, unit, max, SeriesNumber } = mappedAnnotation;
+
+    if (mean && max) {
+      // const roundedMean = utils.roundNumber(mean, 2);
```
JamesAPetts, line 209: Can we get rid of the unused comments?

extensions/cornerstone-dicom-sr/src/tools/DICOMSRDisplayTool.ts
```diff
@@ -270,11 +274,38 @@ export default class DICOMSRDisplayTool extends AnnotationTool {
...
+    referencedImageId,
     options
   ) {
-    let canvasCoordinates;
+    const canvasCoordinates = [];
     renderableData.map((data, index) => {
-      canvasCoordinates = data.map(p => viewport.worldToCanvas(p));
+      const point = data[0];
+      // This gives us one point for arrow
+      canvasCoordinates.push(viewport.worldToCanvas(point));
```
JamesAPetts, line 284: I get why we are doing all of this, but it does seem a bit weird. What about when we start storing probe. I guess we render them both as arrows? Is there any harm? cc @swederik

extensions/default/src/ViewerLayout/index.tsx
```diff
@@ -207,7 +202,7 @@ function ViewerLayout({
       </Header>
       <div
         className="flex flex-row items-stretch w-full overflow-hidden flex-nowrap"
-        style={{ height: 'calc(100vh - 60px' }}
+        style={{ height: 'calc(100vh - 52px' }}
```
JamesAPetts, line 205: 👁️ What's the reason for this change? Won't this affect regression testing?

extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/_hydrateStructuredReport.js
```diff
@@ -119,6 +120,10 @@ export default function _hydrateStructuredReport(
     }
   }
 
+  // When Hydrating the structure report we need to cleanup the cornerstone state
+  // since it will result to a duplicated states
+  cornerstone3DTools.annotation.state.removeAllAnnotations();
```
JamesAPetts, line 125: Is this true? Where we doing this before? What if I want to fuse two SRs together? Didn't we used to have a workflow which was like "choose if you want to add these measurements to the current read, or start a new set"? cc @swederik

JamesAPetts (commented):
Mostly good, some questions for @swederik

PR #2812, opened 2022-05-23
fix: Jump presets cs3d

sedghi:
- Add JumpPreset to OHIF for hangingProtocol
- Some remaining comments from previous PR


JamesAPetts (approved):
LGTM

PR #2814, opened 2022-05-26
feat: volume API and TMTV mode

sedghi:
- Adds petSUV panel which shows the metadata for the matched PT and also makes them editable
- Adds metadata invalidation and force re-loading the data when the metadata is changed
- Adds feature to be able to set tool configurations on the tool
- Adds automatic slabThickness calculation for the volume viewport
- Fix UI discrepancy between v3-stable and tmtv mode


extensions/cornerstone-3d/src/services/ViewportService/Cornerstone3DViewportService.ts
```diff
@@ -489,6 +502,35 @@ class Cornerstone3DViewportService implements IViewportService {
...
+
+    // if there is a slabThickness set as a number then use it
+    if (
+      displaySetOptions.slabThickness &&
+      typeof displaySetOptions.slabThickness === 'number'
+    ) {
+      return displaySetOptions.slabThickness;
+    }
+
+    if (displaySetOptions.slabThickness === 'auto') {
```
JamesAPetts, line 519: Shouldn't it be something like "fullBody", "fullVolume" or "max"? "auto" is a bit vague.

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/index.tsx
```diff
@@ -3,9 +3,17 @@ import PropTypes from 'prop-types';
 //
 import PanelStudyBrowserTracking from './PanelStudyBrowserTracking';
 import getImageSrcFromImageId from './getImageSrcFromImageId';
-import getStudiesForPatientByStudyInstanceUID from './getStudiesForPatientByStudyInstanceUID';
 import requestDisplaySetCreationForStudy from './requestDisplaySetCreationForStudy';
 
+function _getStudyForPatientUtility(extensionManager) {
+  const utilityModule = extensionManager.getModuleEntry(
```
JamesAPetts, line 9: The function from the utility module is so small that it feels weird to me to have this dependency. No change needed for now, but I'm starting to be wary of all of the inter-extension dependencies. You now have to use the default extension in your app for this tiny utility. But I get the measurement-tracking extension is highly coupled with the mode, and the mode uses a default so... eh. But it still feels bad and we shouldn't try to make this commonplace.

extensions/tmtv/src/getHangingProtocolModule.js
```diff
@@ -244,6 +244,9 @@ const ptCT = {
           displaySets: [
             {
               id: 'ctDisplaySet',
+              options: {
+                blendMode: 'MIP',
```
JamesAPetts, line 248: I'm confused, didn't we say the crosshairs should just set the volume to `MIP` when they set a slab of any size other than the min MPR thickness? I don't think we should have to specify MIP on the hanging protocol level for every single viewport which may be effected by the crosshairs tool at some point?

extensions/tmtv/src/utils/getStudiesForPatientByStudyInstanceUID.js
```diff
@@ -21,9 +21,6 @@ async function getStudiesForPatientByStudyInstanceUID(
       patientId: getStudyResult[0].mrn,
     });
   }
-  console.log('No mrn found for', getStudyResult);
-  // The original study we KNOW belongs to the same set, so just return it
-  return getStudyResult;
```
JamesAPetts, line 26: What's the reason for this change?

extensions/tmtv/src/index.tsx
```diff
@@ -1,5 +1,6 @@
 import { id } from './id';
 import getHangingProtocolModule from './getHangingProtocolModule';
+import getPanelModule from './getPanelModule';
```
JamesAPetts, line 3: I wonder if this should just be a generic pet-suv module rather than named the `tmtv` extension? The hanging protocol and side panel could be re-used in other modes, no?

JamesAPetts (commented):
My review is limited as I cant get the actual mode to work. I get blank viewports and the following react error: `The prop `viewportData` is marked as required in `CornerstoneImageScrollbar`, but its value is `null`.`

This happens with both the current published cornerstone3D versions, and the linked versions from your [most recent PR](https://github.com/cornerstonejs/cornerstone3D-beta/pull/113).

Otherwise, some comments in line.

extensions/tmtv/src/index.tsx
salimkanoun: Yes I agree, pet-suv will be more understood and get a better audiance than tmtv.

extensions/cornerstone-3d/src/Viewport/Overlays/ViewportOverlay.tsx
```diff
@@ -42,6 +42,11 @@ function CornerstoneViewportOverlay({
         upper
       );
 
+      // Don't set VOI if already set
+      if (voi.windowCenter !== null || voi.windowWidth !== null) {
```
JamesAPetts, line 46: Surely this needs to be an `&&`? I know its an edge case and shouldn't be hit, but if one of them is not set we should set the voi right?

extensions/cornerstone-3d/src/commandsModule.js
```diff
@@ -304,11 +322,146 @@ const commandsModule = ({ servicesManager }) => {
...
+      ]);
+
+      return segmentationId;
+    },
+    async addSegmentationRepresentationToToolGroup({
+      segmentationId,
+      toolGroupId,
+      options,
+    }) {
+      const { representationType } = options;
```
JamesAPetts, line 363: If representationType is mandatory, should it really be in `options`? Shouldn't it be a top level argument after `segmentationId` and `toolGroupId`?

extensions/cornerstone-3d/src/commandsModule.js
```diff
@@ -304,11 +322,146 @@ const commandsModule = ({ servicesManager }) => {
...
+        segmentations = SegmentationService.getSegmentations();
+      }
+
+      const labelmapVolumes = segmentations.map(segmentation => {
+        return cornerstone.cache.getVolume(segmentation.id);
+      });
+
+      return labelmapVolumes;
+    },
+    getSegmentationReport: ({ segmentations }) => {
```
JamesAPetts, line 387: This seems extremely specific to whatever the tmtv mode wants and seems like its a bit strange at the cornerstone extension level. Can this be hoisted to the tmtv extension?

extensions/cornerstone-3d/src/commandsModule.js
```diff
@@ -304,11 +322,146 @@ const commandsModule = ({ servicesManager }) => {
...
+          PatientName: instance.PatientName.Alphabetic,
+          StudyInstanceUID: instance.StudyInstanceUID,
+          SeriesInstanceUID: instance.SeriesInstanceUID,
+          StudyDate: instance.StudyDate,
+        };
+      }
+
+      return report;
+    },
+    exportRTReportForAnnotations: ({ annotations }) => {
```
JamesAPetts, line 462: Again this seems at the wrong layer, especially because its a very specific output for the RTSTRUCT in the tmtv case.

extensions/cornerstone-3d/src/initMeasurementService.js
```diff
@@ -212,17 +212,19 @@ const connectMeasurementServiceToTools = (
...
-
-      if (sourceAnnotation) {
-        sourceAnnotation.label = label;
-        if (sourceAnnotation.hasOwnProperty('text')) {
+      const sourceAnnotation = annotation.state.getAnnotation(uid);
+      const data = sourceAnnotation.data;
+      if (data) {
+        data.label = label;
+        if (data.hasOwnProperty('text')) {
           // Deal with the weird case of ArrowAnnotate.
```
JamesAPetts, line 222: Maybe we should really just make the arrow annotate tool change the label of its annotation rather than add a proprietary `text` field. We seem to have to do this over and over.

extensions/cornerstone-3d/src/utils/dicomRTAnnotationExport/RTStructureSet/RTSSReport.js
```diff
@@ -0,0 +1,259 @@
...
+import dcmjs from 'dcmjs';
+import { DicomMetadataStore } from '@ohif/core';
+
+const { DicomMetaDictionary } = dcmjs.data;
+
+export default class RTSSReport {
+  constructor() {}
+
+  /**
+   * Convert Cornerstone tool annotation data to RTSSReport report object
```
JamesAPetts, line 11: > Convert Cornerstone tool annotation data

Only if its formatted in a specific way, I think this is actually "convert handles to RTSTRUCT". E.g. Freehand poly won't work.

As this is super limited right now and doesn't really make what people expect from a full RTSTRUCT (e.g. a set of multiple ROIContours containing many 2D Contours), this should all be moved to the TMTV extension?

extensions/cornerstone-3d/src/utils/dicomRTAnnotationExport/RTStructureSet/RTSSReport.js
```diff
@@ -0,0 +1,259 @@
...
+
+  const { studyInstanceUID } = metadataProvider.get(
+    'generalSeriesModule',
+    imageId
+  );
+
+  const patientModule = getPatientModule(imageId, metadataProvider);
+  const rtSeriesModule = getRTSeriesModule(imageId, metadataProvider);
+
+  return {
```
JamesAPetts, line 111: Doesn't have to be right now, but this seems like it should all go down to dcmjs at some point in the future as a DICOM RTSTRUCT derivation.

extensions/cornerstone-3d/src/utils/dicomRTAnnotationExport/RTStructureSet/RTSSReport.js
```diff
@@ -0,0 +1,259 @@
...
+}
+
+function getStructureSetModule(toolData, index, metadataProvider) {
+  const { FrameOfReferenceUID } = toolData.metadata;
+
+  return {
+    ROINumber: index + 1,
+    ROIName: `Todo: name ${index + 1}`,
+    ROIDescription: `Todo: description ${index + 1}`,
+    ROIGenerationAlgorithm: 'Todo: algorithm',
```
JamesAPetts, line 247: This is type 2, so just leave it blank rather than putting a dummy value.

extensions/cornerstone-3d/src/utils/dicomRTAnnotationExport/RTStructureSet/measurements/AnnotationToPointData.js
```diff
@@ -0,0 +1,58 @@
...
+
+  static convert(annotation, index, metadataProvider) {
+    validateAnnotation(annotation);
+
+    const { toolName } = annotation.metadata;
+    const toolClass = AnnotationToPointData.TOOL_NAMES[toolName];
+
+    if (!toolClass) {
+      throw new Error(
+        `Unknown tool type: ${toolName}, cannot convert to RTSSReport`
```
JamesAPetts, line 24: In general I think this whole thing is honestly a bit too generic. For most annotations, SR makes a lot more sense. RTSTRUCT is going to be extremely specific (points, open/closed 2D/3D contours).

extensions/default/src/commandsModule.js
```diff
@@ -2,9 +2,17 @@ const commandsModule = ({ servicesManager, commandsManager }) => {
   const {
     MeasurementService,
     HangingProtocolService,
+    UINotificationService,
   } = servicesManager.services;
 
   const actions = {
+    displayNotification: ({ text }) => {
+      UINotificationService.show({
+        title: 'RectangleROI Threshold Tip',
```
JamesAPetts, line 11: Why is this in the default extension? This is specifically for the RectangleROI. An OHIF mode using default might not even be using Cornerstone.

extensions/tmtv/src/Panels/PanelROIThresholdSegmentation.tsx
```diff
@@ -0,0 +1,456 @@
...
+  const [labelmapLoading, setLabelmapLoading] = useState(false);
+  const [selectedSegmentationId, setSelectedSegmentationId] = useState(null);
+  const [segmentations, setSegmentations] = useState(() =>
+    SegmentationService.getSegmentations()
+  );
+  const [config, setConfig] = useState({
+    strategy: 'roiStat',
+    minValue: 0,
+    maxValue: 100,
+    weight: 0.41,
```
JamesAPetts, line 36: Where do these magic numbers come from?

extensions/tmtv/src/Panels/PanelROIThresholdSegmentation.tsx
```diff
@@ -0,0 +1,456 @@
...
+        // temp: swap button types until colors are updated
+        { id: 'cancel', text: 'Cancel', type: 'primary' },
+        { id: 'save', text: 'Save', type: 'secondary' },
+      ],
+      onSubmit: onSubmitHandler,
+    },
+  });
+}
+
+function ROIThresholdConfiguration({ config, setConfig, runCommand }) {
```
JamesAPetts, line 353: I feel like it would be good to break up some of these components into multiple files, there is a lot of JSX here.

extensions/tmtv/src/init.js
```diff
@@ -0,0 +1,42 @@
...
+  commandsManager,
+  configuration,
+}) {
+  const {
+    MeasurementService,
+    DisplaySetService,
+    Cornerstone3DViewportService,
+  } = servicesManager.services;
+
+  const { RectangleROIStartEndThreshold } = measurementServiceMappingsFactory(
```
JamesAPetts, line 22: So the mapping is added here, but the tool is added in the cornerstone extension? Since they are for the TMTV mode, they should both really be added in this extension?

extensions/tmtv/src/utils/calculateSUVPeak.ts
```diff
@@ -0,0 +1,152 @@
...
+  );
+
+  const direction = labelmapImageData
+    .getDirection()
+    .slice(0, 3) as Types.Point3;
+  /**
+   * 2. Find the bottom and top of the great circle for the second sphere (1cc sphere)
+   * diameter of 12mm = 1.2cm ~ = 1cc sphere
+   */
+  const diameter = 12;
```
JamesAPetts, line 113: Can you make this a constant outside this file and actually calculate it correctly once at run time? i.e. 2 * ³⎷(3 * Volume / (4 π))

extensions/tmtv/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.js
```diff
@@ -0,0 +1,48 @@
...
+   * @param {Measurement} measurement The measurement instance
+   * @param {string} definition The source definition
+   * @return {Object} Cornerstone annotation data
+   */
+
+  const _getValueTypeFromToolType = toolType => {
+    const { SUV_PEAK } = MeasurementService.VALUE_TYPES;
+
+    // TODO -> I get why this was attemped, but its not nearly flexible enough.
+    // A single measurement may have an ellipse + a bidirectional measurement, for instances.
```
JamesAPetts, line 20: for instance.

extensions/tmtv/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.js
```diff
@@ -0,0 +1,48 @@
...
+    // You can't define a bidirectional tool as a single type..
+    const TOOL_TYPE_TO_VALUE_TYPE = {
+      ptSUVPeak: SUV_PEAK,
+    };
+
+    return TOOL_TYPE_TO_VALUE_TYPE[toolType];
```
JamesAPetts, line 26: What exactly is this function supposed to be doing? Right now it returns SUV_PEAK if the toolType is ptSUVPeak, so why all the fluff?

extensions/tmtv/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
```diff
@@ -0,0 +1,69 @@
...
+  const instance = cornerstone.metaData.get('instance', imageId);
+
+  return {
+    SOPInstanceUID: instance.SOPInstanceUID,
+    SeriesInstanceUID: instance.SeriesInstanceUID,
+    StudyInstanceUID: instance.StudyInstanceUID,
+  };
+}
+
+// function getUIDFromScene(Cornerstone3DViewportService) {
```
JamesAPetts, line 51: We don't have scenes anymore so surely you can just fully nuke this.

extensions/tmtv/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
```diff
@@ -0,0 +1,69 @@
...
+ * inside the scene and then it checks each viewport to find the one that has
+ * acquisition plane view, and uses the currentImageId of the viewport to
+ * query the metadata and get UIDs.
+ * @param {string} imageId The image id of the referenced image
+ * @param {string} sceneUID The scene UID of the measurement tool
+ * @returns
+ */
+export default function getSOPInstanceAttributes(
+  imageId,
+  Cornerstone3DViewportService,
```
JamesAPetts, line 16: Wouldn't it be a more natural api to pass in the servicesManager, and then attempt to get this service (throwing if it does not exist)? (I know this isn't used yet, but we could at least change the API ahead of time, right now this is a mandatory argument even though it isn't used).

extensions/tmtv/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
```diff
@@ -0,0 +1,69 @@
+import * as cornerstone from '@cornerstonejs/core';
```
JamesAPetts, line 1: Shouldn't we just expose the `metaData` object through the cornrerstone extensions utilities? Its going to be used pretty commonly I expect?

JamesAPetts (commented):
A bunch of comments in line, but lets discuss tomorrow as well.

extensions/cornerstone-3d/src/commandsModule.js
sedghi: getCSVReport

extensions/cornerstone-3d/src/utils/measurementServiceMappings/measurementServiceMappingsFactory.test.js
```diff
@@ -1,102 +0,0 @@
-import measurementServiceMappingsFactory from './measurementServiceMappingsFactory';
```
JamesAPetts, line 1: What's the rationalle for deleting this test?

extensions/tmtv/src/Panels/PanelROIThresholdSegmentation/PanelROIThresholdSegmentation.tsx
```diff
@@ -114,6 +133,28 @@ export default function PanelRoiThresholdSegmentation({
...
+    const removed = SegmentationService.EVENTS.SEGMENTATION_REMOVED;
+
+    const { unsubscribe } = SegmentationService.subscribe(removed, () => {
+      const segmentations = SegmentationService.getSegmentations();
+      setSegmentations(segmentations);
+
+      if (segmentations.length > 0) {
+        setSelectedSegmentationId(segmentations[0].id);
+        handleTMTVCalculation();
+        return;
```
JamesAPetts, line 146: I feel like `if`...`else` would be easier to read than a return in this block.

modes/tmtv/README.md
```diff
@@ -0,0 +1,78 @@
...
+Finally, the results can be saved in the CSV format. The RectangleROI annotations
+can also be extracted as a dicom RT Structure Set and saved as a DICOM file.
+
+
+## Video Tutorial
+
+Below you can see a video tutorial on how to use the TMTV workflow mode.
+
+
+https://user-images.githubusercontent.com/7490180/171065443-35369fba-e955-48ac-94da-d262e0fccb6b.mp4
```
JamesAPetts, line 77: Just checking, are we sure this is a permalink that won't die? Maybe we should host it on ohif.org or something? cc @swederik .

PR #2815, opened 2022-05-27
fix(OHIFv3):Allow multiple study UIDs in the path.

wayfarer3130:
The wrong call was being made to fetch the study instance uids, so it wasn't possible to fetch two or more instances at once.


wayfarer3130:
@sedghi - can you take a quick look, very simple fix.

sedghi:
I tried it like this 

`http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2403.334240657131972136850343327463&StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2403.459769504433903221904322299373`

and it showed me 4 studies instead of two

image

How should i try it?

wayfarer3130:
@sedghi - this is working again.  There was a bug in the panel study browser causing the studies to be loaded twice.

PR #2816, opened 2022-05-27
Feat/dynamic mode configuration

wayfarer3130:
This commit adds the ability to load modes via configuration, but doesn't actually load any modes that way yet.


package.json
```diff
@@ -56,8 +56,9 @@
   },
   "dependencies": {
     "@babel/runtime": "7.16.3",
-    "core-js": "^3.2.1",
-    "@kitware/vtk.js": "^24.0.0"
+    "@kitware/vtk.js": "^24.0.0",
+    "config-point": "^0.4.5",
```
wayfarer3130, line 60: config-point as a dependency is required in order to load the sub-modes correctly, but isn't used anywhere unless you actually have dynamic loading enabled.

platform/viewer/.webpack/writePluginImportsFile.js
```diff
@@ -42,7 +42,11 @@ function getFormattedImportBlock(importLines) {
 }
 
 function getFormattedWindowBlock(addToWindowLines) {
-  let content = `window.extensions = [];\nwindow.modes = [];\n\n`;
+  let content = "const extensions = [];\n" +
+    "const modes = [];\n" +
+    "const modesFactory = [];\n" +
```
wayfarer3130, line 47: Add a modesFactory section that creates modes via an async function.

platform/viewer/.webpack/writePluginImportsFile.js
```diff
@@ -51,21 +55,35 @@ function getFormattedWindowBlock(addToWindowLines) {
   return content;
 }
 
+function getRuntimeLoadModesExtensions() {
```
wayfarer3130, line 58: And actually generate the exported module that generates the mode names.

platform/viewer/pluginConfig.json
```diff
@@ -25,9 +25,10 @@
       "version": "3.0.1"
     }
   ],
-  "modes": [
+  "modes": [],
```
wayfarer3130, line 28: I will revert this entire file before merging, but I wanted to show how it gets used.

PR #2831, opened 2022-06-28
fix: various type issues and error handlings

wayfarer3130:
Fix a number of warnings and add some error messages on extension/mode load failure.

Also tweak the config files to make them consistent with other branches.

### DIFF
--- a/modes/longitudinal/src/index.js
+++ b/modes/longitudinal/src/index.js
@@ -3,6 +3,10 @@ import toolbarButtons from './toolbarButtons.js';
 import { id } from './id.js';
 import initToolGroups from './initToolGroups.js';
 
+// Allow this mode by excluding non-imaging modalities such as SR, SEG
+// Also, SM is not a simple imaging modalities, so exclude it.
+const NON_IMAGE_MODALITIES = ['SM', 'ECG', 'SR', 'SEG'];
+
 const ohif = {
   layout: '@ohif/extension-default.layoutTemplateModule.viewerLayout',
   sopClassHandler: '@ohif/extension-default.sopClassHandlerModule.stack',
@@ -120,15 +124,12 @@ function modeFactory() {
       series: [],
     },
 
-    // Allow this mode by excluding non-imaging modalities such as SR, SEG
-    // Also, SM is not a simple imaging modalities, so exclude it.
-    NON_IMAGE_MODALITIES: ['SM', 'ECG', 'SR', 'SEG'],
-    isValidMode: function ({ modalities }) {
+    isValidMode: function({ modalities }) {
       const modalities_list = modalities.split('\\');
 
       // Exclude non-image modalities
-      return !!modalities_list.filter(modality =>
-        this.NON_IMAGE_MODALITIES.indexOf(modality) == -1
+      return !!modalities_list.filter(
+        modality => NON_IMAGE_MODALITIES.indexOf(modality) === -1
       ).length;
     },
     routes: [
--- a/modes/longitudinal/src/index.js
+++ b/modes/longitudinal/src/index.js
@@ -44,7 +44,7 @@ const extensionDependencies = {
   '@ohif/extension-dicom-video': '^3.0.1',
 };
 
-function modeFactory({ modeConfiguration }) {
+function modeFactory() {
   return {
     // TODO: We're using this as a route segment
     // We should not be.
@@ -119,11 +119,17 @@ function modeFactory({ modeConfiguration }) {
       study: [],
       series: [],
     },
-    isValidMode: ({ modalities }) => {
+
+    // Allow this mode by excluding non-imaging modalities such as SR, SEG
+    // Also, SM is not a simple imaging modalities, so exclude it.
+    NON_IMAGE_MODALITIES: ['SM', 'ECG', 'SR', 'SEG'],
+    isValidMode: function ({ modalities }) {
       const modalities_list = modalities.split('\\');
 
-      // Slide Microscopy modality not supported by basic mode yet
-      return !modalities_list.includes('SM');
+      // Exclude non-image modalities
+      return !!modalities_list.filter(modality =>
+        this.NON_IMAGE_MODALITIES.indexOf(modality) == -1
+      ).length;
     },
     routes: [
       {
@@ -131,7 +137,7 @@ function modeFactory({ modeConfiguration }) {
         /*init: ({ servicesManager, extensionManager }) => {
           //defaultViewerRouteInit
         },*/
-        layoutTemplate: ({ location, servicesManager }) => {
+        layoutTemplate: () => {
           return {
             id: ohif.layout,
             props: {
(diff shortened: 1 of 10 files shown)


modes/longitudinal/src/index.js
```diff
@@ -123,15 +123,15 @@ function modeFactory({ modeConfiguration }) {
       const modalities_list = modalities.split('\\');
 
       // Slide Microscopy modality not supported by basic mode yet
-      return !modalities_list.includes('SM');
+      return !!modalities_list.filter(modality => modality != 'SM' && modality !== 'ECG').length;
```
wayfarer3130, line 126: This fixes some display issues when viewing a study that is viewable, but the mode failed to load because the test was incorrect.  Also remove ECG from a default view mode, as that just causes problems.

platform/viewer/public/config/dicomweb_relative.js
```diff
@@ -0,0 +1,160 @@
+window.config = {
```
wayfarer3130, line 1: This is just a generic config that accesses a dicomweb relative directory - it is added for consistency with other builds.  Note that it will get removed eventually when it gets moved into a custom mode handler, but for now it is important to allow this one to exist.

platform/viewer/public/config/local_static.js
```diff
@@ -12,9 +12,9 @@ window.config = {
       sourceName: 'dicomweb',
       configuration: {
         name: 'DCM4CHEE',
-        wadoUriRoot: '/dicomweb',
-        qidoRoot: '/dicomweb',
-        wadoRoot: '/dicomweb',
+        wadoUriRoot: 'http://localhost:5000/dicomweb',
```
wayfarer3130, line 15: And the localhost:5000 prevents reloads/serving up a fairly large dicomweb directory within the webpack config.

platform/viewer/src/components/ViewportGrid.tsx
```diff
@@ -369,13 +369,18 @@ function _getViewportComponent(displaySets, viewportComponents) {
   const SOPClassHandlerId = displaySets[0].SOPClassHandlerId;
 
   for (let i = 0; i < viewportComponents.length; i++) {
+    if (!viewportComponents[i]) throw new Error("viewport components not defined");
```
wayfarer3130, line 372: This set of changes makes it a lot easier to debug what is going on for new modes - they often don't quite get loaded correctly.

platform/viewer/src/service-worker.js
```diff
@@ -10,7 +10,7 @@ workbox.core.clientsClaim();
 
 // Cache static assets that aren't precached
 workbox.routing.registerRoute(
-  /\.(?:js|css)$/,
+  /\.(?:js|css|json5)$/,
```
wayfarer3130, line 13: Prevents a bug if json5 configuration files are used.  Not really hit right now in this branch, but generally useful to include.

modes/longitudinal/src/index.js
sedghi: We don't have any ECG in the viewer (public OHIF repo), and I'm pretty sure that will raise more questions that where is the ECG data and where is the viewer that can load them

platform/ui/src/components/StudyBrowser/StudyBrowser.tsx
```diff
@@ -140,7 +140,7 @@ StudyBrowser.propTypes = {
               imageSrc: PropTypes.string,
               imageAltText: PropTypes.string,
               seriesDate: PropTypes.string,
-              seriesNumber: PropTypes.string,
+              seriesNumber: PropTypes.any,
```
sedghi, line 143: How can `seriesNumber` be any? isn't it `number` or `string`? `any` means it can be an object or a function too which is wrong

platform/ui/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -98,7 +98,7 @@ Thumbnail.propTypes = {
   }),
   imageAltText: PropTypes.string,
   description: PropTypes.string.isRequired,
-  seriesNumber: PropTypes.string.isRequired,
+  seriesNumber: PropTypes.any.isRequired,
```
sedghi, line 101: same here

platform/ui/src/components/ThumbnailList/ThumbnailList.tsx
```diff
@@ -110,7 +110,7 @@ ThumbnailList.propTypes = {
       imageSrc: PropTypes.string,
       imageAltText: PropTypes.string,
       seriesDate: PropTypes.string,
-      seriesNumber: PropTypes.string,
+      seriesNumber: PropTypes.any,
```
sedghi, line 113: same here

platform/ui/src/components/ThumbnailTracked/ThumbnailTracked.tsx
```diff
@@ -137,7 +137,7 @@ ThumbnailTracked.propTypes = {
   imageSrc: PropTypes.string,
   imageAltText: PropTypes.string,
   description: PropTypes.string.isRequired,
-  seriesNumber: PropTypes.string.isRequired,
+  seriesNumber: PropTypes.any.isRequired,
```
sedghi, line 140: same here

platform/viewer/public/config/local_static.js
sedghi: Can you add this comment to the source code? this is a `js` file and we can add comments there

platform/viewer/src/components/ViewportGrid.tsx
sedghi: To be consistent with the rest of the codebase, can we have brackets for the if statements?

sedghi (changes requested):
Great error handling stuff, please see my comments

modes/longitudinal/src/index.js
wayfarer3130: There doesn't have to be - the issue is that these are not viewable by the viewer right now, so it shouldn't attempt to load them.  However, I can revert this change, or else make a list of disallowed modalities.

wayfarer3130: Changed to use NON_IMAGE_MODALITIES list and .indexOf, so it is clear what is going on.

modes/longitudinal/src/index.js
```diff
@@ -119,19 +119,25 @@ function modeFactory({ modeConfiguration }) {
       study: [],
       series: [],
     },
-    isValidMode: ({ modalities }) => {
+
+    // Allow this mode by excluding non-imaging modalities such as SR, SEG
+    // Also, SM is not a simple imaging modalities, so exclude it.
+    NON_IMAGE_MODALITIES: ['SM', 'ECG', 'SR', 'SEG'],
```
wayfarer3130, line 125: This makes it obvious that it is simply not running for non-imaging modalities.

platform/ui/src/Types.ts
```diff
@@ -0,0 +1,24 @@
+import PropTypes from 'prop-types';
```
wayfarer3130, line 1: Make it more clear what the various types are, so they could be re-used.

sedghi (approved):

PR #2837, opened 2022-07-06
feat: make hanging protocol work on displaySets instead of series

wayfarer3130:
Fixes numerous issues causing real hanging protocols to not be definable or not work as expected, or work intermittently.

### DIFF
--- a/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
+++ b/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
@@ -39,6 +39,8 @@ class CornerstoneViewportService implements IViewportService {
   renderingEngine: Types.IRenderingEngine | null;
   viewportsInfo: Map<number, ViewportInfo>;
   viewportGridResizeObserver: ResizeObserver | null;
+  // TODO - get the right type here.
+  hangingProtocolService: object;
 
   /**
    * Service-specific
@@ -59,7 +61,7 @@ class CornerstoneViewportService implements IViewportService {
     this.listeners = {};
     this.EVENTS = EVENTS;
     const { HangingProtocolService } = servicesManager.services;
-    this.HangingProtocolService = HangingProtocolService;
+    this.hangingProtocolService = HangingProtocolService;
     Object.assign(this, pubSubServiceInterface);
     //
   }
@@ -265,8 +267,9 @@ class CornerstoneViewportService implements IViewportService {
     viewportInfo: ViewportInfo
   ) {
     const displaySetOptions = viewportInfo.getDisplaySetOptions();
-
+    const viewportOptions = viewportInfo.getViewportOptions();
     const { imageIds, initialImageIndex } = viewportData;
+    const { hangingProtocolService } = this;
 
     let initialImageIndexToUse = initialImageIndex;
 
@@ -352,7 +355,7 @@ class CornerstoneViewportService implements IViewportService {
     // (This call may or may not create sub-requests for series metadata)
     const volumeInputArray = [];
     const displaySetOptionsArray = viewportInfo.getDisplaySetOptions();
-    const { HangingProtocolService } = this;
+    const { hangingProtocolService } = this;
 
     for (let i = 0; i < viewportData.imageIds.length; i++) {
       const imageIds = viewportData.imageIds[i];
@@ -384,11 +387,11 @@ class CornerstoneViewportService implements IViewportService {
     }
 
     if (
-      HangingProtocolService.hasCustomImageLoadStrategy() &&
-      !HangingProtocolService.customImageLoadPerformed
+      hangingProtocolService.hasCustomImageLoadStrategy() &&
+      !hangingProtocolService.customImageLoadPerformed
     ) {
       // delegate the volume loading to the hanging protocol service if it has a custom image load strategy
-      return HangingProtocolService.runImageLoadStrategy({
+      return hangingProtocolService.runImageLoadStrategy({
         viewportId: viewport.id,
         volumeInputArray,
       });
@@ -552,7 +555,7 @@ class CornerstoneViewportService implements IViewportService {
   ): {
     viewportOptions: ViewportOptions;
     displaySetOptions: DisplaySetOptions[];
-  } {
+    } {
     const viewportIndex = viewportInfo.getViewportIndex();
 
     // Creating a temporary viewportInfo to handle defaults
--- a/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
+++ b/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
@@ -168,12 +168,14 @@ class CornerstoneViewportService implements IViewportService {
     const viewportInfo = this.viewportsInfo.get(viewportIndex);
     viewportInfo.setRenderingEngineId(renderingEngine.id);
 
-    const { viewportOptions, displaySetOptions } =
-      this._getViewportAndDisplaySetOptions(
-        publicViewportOptions,
-        publicDisplaySetOptions,
-        viewportInfo
-      );
+    const {
+      viewportOptions,
+      displaySetOptions,
+    } = this._getViewportAndDisplaySetOptions(
+      publicViewportOptions,
+      publicDisplaySetOptions,
+      viewportInfo
+    );
 
     viewportInfo.setViewportOptions(viewportOptions);
     viewportInfo.setDisplaySetOptions(displaySetOptions);
@@ -267,9 +269,7 @@ class CornerstoneViewportService implements IViewportService {
     viewportInfo: ViewportInfo
   ) {
     const displaySetOptions = viewportInfo.getDisplaySetOptions();
-    const viewportOptions = viewportInfo.getViewportOptions();
     const { imageIds, initialImageIndex } = viewportData;
-    const { hangingProtocolService } = this;
 
     let initialImageIndexToUse = initialImageIndex;
 
@@ -418,8 +418,9 @@ class CornerstoneViewportService implements IViewportService {
     ) {
       const { index, preset } = initialImageOptions;
 
-      const { numberOfSlices } =
-        csUtils.getImageSliceDataForVolumeViewport(viewport);
+      const { numberOfSlices } = csUtils.getImageSliceDataForVolumeViewport(
+        viewport
+      );
 
       const imageIndex = this._getInitialImageIndex(
         numberOfSlices,
@@ -555,7 +556,7 @@ class CornerstoneViewportService implements IViewportService {
   ): {
     viewportOptions: ViewportOptions;
     displaySetOptions: DisplaySetOptions[];
-    } {
+  } {
     const viewportIndex = viewportInfo.getViewportIndex();
 
     // Creating a temporary viewportInfo to handle defaults
--- a/extensions/default/src/DicomWebDataSource/index.js
+++ b/extensions/default/src/DicomWebDataSource/index.js
@@ -473,12 +473,12 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
 
       DicomMetadataStore.addSeriesMetadata(seriesSummaryMetadata, madeInClient);
 
-      const numberOfSeries = seriesPromises.length;
-      seriesPromises.forEach(async (seriesPromise, index) => {
-        const instances = await seriesPromise;
-        storeInstances(instances);
-        if (index === numberOfSeries - 1) setSuccessFlag();
-      });
+      const seriesDeliveredPromises = seriesPromises.map(
+        promise => promise.then(instances => {
+          storeInstances(instances);
+        }));
+      await Promise.all(seriesDeliveredPromises);
+      setSuccessFlag();
     },
     deleteStudyMetadataPromise,
     getImageIdsForDisplaySet(displaySet) {
--- a/extensions/default/src/DicomWebDataSource/index.js
+++ b/extensions/default/src/DicomWebDataSource/index.js
@@ -473,10 +473,11 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
 
       DicomMetadataStore.addSeriesMetadata(seriesSummaryMetadata, madeInClient);
 
-      const seriesDeliveredPromises = seriesPromises.map(
-        promise => promise.then(instances => {
+      const seriesDeliveredPromises = seriesPromises.map(promise =>
+        promise.then(instances => {
           storeInstances(instances);
-        }));
+        })
+      );
       await Promise.all(seriesDeliveredPromises);
       setSuccessFlag();
     },
--- a/extensions/tmtv/src/commandsModule.js
+++ b/extensions/tmtv/src/commandsModule.js
@@ -41,9 +41,9 @@ const commandsModule = ({
   }
 
   function _getMatchedViewportsToolGroupIds() {
-    const [matchedViewports] = HangingProtocolService.getState();
+    const { viewportMatchDetails } = HangingProtocolService.getMatchDetails();
     const toolGroupIds = [];
-    matchedViewports.forEach(({ viewportOptions }) => {
+    viewportMatchDetails.forEach(({ viewportOptions }) => {
       const { toolGroupId } = viewportOptions;
       if (toolGroupIds.indexOf(toolGroupId) === -1) {
         toolGroupIds.push(toolGroupId);
@@ -54,33 +54,30 @@ const commandsModule = ({
   }
 
   const actions = {
-    getMatchingPTDisplaySet: () => {
+    getMatchingPTDisplaySet: ({ viewportMatchDetails }) => {
       // Todo: this is assuming that the hanging protocol has successfully matched
       // the correct PT. For future, we should have a way to filter out the PTs
       // that are in the viewer layout (but then we have the problem of the attenuation
       // corrected PT vs the non-attenuation correct PT)
-      const matches = HangingProtocolService.getDisplaySetsMatchDetails();
-
-      const matchedSeriesInstanceUIDs = Array.from(matches.values()).map(
-        ({ SeriesInstanceUID }) => SeriesInstanceUID
-      );
 
       let ptDisplaySet = null;
-      for (const SeriesInstanceUID of matchedSeriesInstanceUIDs) {
-        const displaySets = DisplaySetService.getDisplaySetsForSeries(
-          SeriesInstanceUID
+      for (const matched of viewportMatchDetails) {
+        const { displaySetsInfo } = matched;
+        const displaySets = displaySetsInfo.map(({ displaySetInstanceUID }) =>
+          DisplaySetService.getDisplaySetByUID(displaySetInstanceUID)
         );
 
         if (!displaySets || displaySets.length === 0) {
           continue;
         }
 
-        const displaySet = displaySets[0];
-        if (displaySet.Modality !== 'PT') {
-          continue;
-        }
+        ptDisplaySet = displaySets.find(
+          displaySet => displaySet.Modality === 'PT'
+        );
 
-        ptDisplaySet = displaySet;
+        if (ptDisplaySet) {
+          break;
+        }
       }
 
       return ptDisplaySet;
@@ -121,7 +118,10 @@ const commandsModule = ({
     createNewLabelmapFromPT: async () => {
       // Create a segmentation of the same resolution as the source data
       // using volumeLoader.createAndCacheDerivedVolume.
-      const ptDisplaySet = actions.getMatchingPTDisplaySet();
+      const { viewportMatchDetails } = HangingProtocolService.getMatchDetails();
+      const ptDisplaySet = actions.getMatchingPTDisplaySet({
+        viewportMatchDetails,
+      });
 
       if (!ptDisplaySet) {
         UINotificationService.error('No matching PT display set found');
@@ -565,8 +565,11 @@ const commandsModule = ({
     },
     setFusionPTColormap: ({ toolGroupId, colormap }) => {
       const toolGroup = ToolGroupService.getToolGroup(toolGroupId);
+      const { viewportMatchDetails } = HangingProtocolService.getMatchDetails();
 
-      const ptDisplaySet = actions.getMatchingPTDisplaySet();
+      const ptDisplaySet = actions.getMatchingPTDisplaySet({
+        viewportMatchDetails,
+      });
 
       if (!ptDisplaySet) {
         return;
--- a/platform/core/src/services/DicomMetadataStore/DicomMetadataStore.js
+++ b/platform/core/src/services/DicomMetadataStore/DicomMetadataStore.js
@@ -211,7 +211,14 @@ const BaseImplementation = {
     let study = _getStudy(StudyInstanceUID);
     if (!study) {
       study = createStudyMetadata(StudyInstanceUID);
+      // Will typically be undefined with a compliant DICOMweb server, reset later
       study.StudyDescription = seriesSummaryMetadata[0].StudyDescription;
+      seriesSummaryMetadata.forEach(item => {
+        if (study.ModalitiesInStudy.indexOf(item.Modality) == -1) {
+          study.ModalitiesInStudy.push(item.Modality);
+        }
+      });
+      study.NumberOfStudyRelatedSeries = seriesSummaryMetadata.length;
       _model.studies.push(study);
     }
 
--- a/platform/core/src/services/DicomMetadataStore/DicomMetadataStore.js
+++ b/platform/core/src/services/DicomMetadataStore/DicomMetadataStore.js
@@ -214,7 +214,7 @@ const BaseImplementation = {
       // Will typically be undefined with a compliant DICOMweb server, reset later
       study.StudyDescription = seriesSummaryMetadata[0].StudyDescription;
       seriesSummaryMetadata.forEach(item => {
-        if (study.ModalitiesInStudy.indexOf(item.Modality) == -1) {
+        if (study.ModalitiesInStudy.indexOf(item.Modality) === -1) {
           study.ModalitiesInStudy.push(item.Modality);
         }
       });
--- a/platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
+++ b/platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
@@ -4,6 +4,7 @@ function createStudyMetadata(StudyInstanceUID) {
   return {
     StudyInstanceUID,
     StudyDescription: '',
+    ModalitiesInStudy: [],
     isLoaded: false,
     series: [],
     /**
@@ -13,6 +14,9 @@ function createStudyMetadata(StudyInstanceUID) {
      */
     addInstanceToSeries: function(instance) {
       const { SeriesInstanceUID } = instance;
+      if (!this.StudyDescription) {
+        this.StudyDescription = instance.StudyDescription;
+      }
       const existingSeries = this.series.find(
         s => s.SeriesInstanceUID === SeriesInstanceUID
       );
@@ -22,16 +26,22 @@ function createStudyMetadata(StudyInstanceUID) {
       } else {
         const series = createSeriesMetadata([instance]);
         this.series.push(series);
+        const { Modality } = series;
+        if (this.ModalitiesInStudy.indexof(Modality) === -1) {
+          this.ModalitiesInStudy.push(Modality);
+        }
       }
     },
     /**
      *
      * @param {object[]} instances
      * @param {string} instances[].SeriesInstanceUID
+     * @param {string} instances[].StudyDescription
      * @returns {bool} true if series were added; false if series already exist
      */
     addInstancesToSeries: function(instances) {
       const { SeriesInstanceUID } = instances[0];
+      if (!this.StudyDescription) this.StudyDescription = instances[0].StudyDescription;
       const existingSeries = this.series.find(
         s => s.SeriesInstanceUID === SeriesInstanceUID
       );
--- a/platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
+++ b/platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
@@ -41,7 +41,9 @@ function createStudyMetadata(StudyInstanceUID) {
      */
     addInstancesToSeries: function(instances) {
       const { SeriesInstanceUID } = instances[0];
-      if (!this.StudyDescription) this.StudyDescription = instances[0].StudyDescription;
+      if (!this.StudyDescription) {
+        this.StudyDescription = instances[0].StudyDescription;
+      }
       const existingSeries = this.series.find(
         s => s.SeriesInstanceUID === SeriesInstanceUID
       );
--- a/platform/core/src/services/HangingProtocolService/HPMatcher.js
+++ b/platform/core/src/services/HangingProtocolService/HPMatcher.js
@@ -4,10 +4,18 @@ import validate from './lib/validator';
  * Match a Metadata instance against rules using Validate.js for validation.
  * @param  {InstanceMetadata} metadataInstance Metadata instance object
  * @param  {Array} rules Array of MatchingRules instances (StudyMatchingRule|SeriesMatchingRule|ImageMatchingRule) for the match
+ * @param {object} options is an object containing additional information
+ * @param {object[]} options.studies is a list of all the studies
+ * @param {object[]} options.displaySets is a list of the display sets
  * @return {Object}      Matching Object with score and details (which rule passed or failed)
  */
-const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
-  const options = {
+const match = (
+  metadataInstance,
+  rules,
+  customAttributeRetrievalCallbacks,
+  options
+) => {
+  const validateOptions = {
     format: 'grouped',
   };
 
@@ -16,18 +24,32 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
     failed: [],
   };
 
+  const readValues = {};
+
   let requiredFailed = false;
   let score = 0;
 
   rules.forEach(rule => {
-    const attribute = rule.attribute;
-
+    const { attribute } = rule;
     // Do not use the custom attribute from the metadataInstance since it is subject to change
     if (customAttributeRetrievalCallbacks.hasOwnProperty(attribute)) {
-      const customAttribute = customAttributeRetrievalCallbacks[attribute];
-      metadataInstance[attribute] = customAttribute.callback(metadataInstance);
+      readValues[attribute] = customAttributeRetrievalCallbacks[
+        attribute
+      ].callback(metadataInstance, options);
+    } else {
+      readValues[attribute] =
+        metadataInstance[attribute] ??
+        ((metadataInstance.images || metadataInstance.others || [])[0] || {})[
+        attribute
+        ];
     }
 
+    console.log(
+      'Test',
+      attribute,
+      readValues[attribute],
+      JSON.stringify(rule.constraint)
+    );
     // Format the constraint as required by Validate.js
     const testConstraint = {
       [attribute]: rule.constraint,
@@ -35,13 +57,7 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
 
     // Create a single attribute object to be validated, since metadataInstance is an
     // instance of Metadata (StudyMetadata, SeriesMetadata or InstanceMetadata)
-    let attributeValue = metadataInstance[attribute];
-    if (attributeValue === undefined) {
-      if (attribute === 'NumberOfStudyRelatedSeries') {
-        attributeValue = metadataInstance.series?.length;
-      }
-      // Add other computable values such as modalities in study
-    }
+    let attributeValue = readValues[attribute];
     const attributeMap = {
       [attribute]: attributeValue,
     };
@@ -49,7 +65,7 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
     // Use Validate.js to evaluate the constraints on the specified metadataInstance
     let errorMessages;
     try {
-      errorMessages = validate(attributeMap, testConstraint, [options]);
+      errorMessages = validate(attributeMap, testConstraint, [validateOptions]);
     } catch (e) {
       errorMessages = ['Something went wrong during validation.', e];
     }
@@ -58,8 +74,7 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
       // If no errorMessages were returned, then validation passed.
 
       // Add the rule's weight to the total score
-      score += parseInt(rule.weight, 10);
-
+      score += parseInt(rule.weight || 1, 10);
       // Log that this rule passed in the matching details object
       details.passed.push({
         rule,
--- a/platform/core/src/services/HangingProtocolService/HPMatcher.js
+++ b/platform/core/src/services/HangingProtocolService/HPMatcher.js
@@ -40,7 +40,7 @@ const match = (
       readValues[attribute] =
         metadataInstance[attribute] ??
         ((metadataInstance.images || metadataInstance.others || [])[0] || {})[
-        attribute
+          attribute
         ];
     }
 
--- a/platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
+++ b/platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
@@ -1,6 +1,8 @@
 import pubSubServiceInterface from '../_shared/pubSubServiceInterface';
-import sortBy from '../../utils/sortBy.js';
+import sortBy from '../../utils/sortBy';
 import ProtocolEngine from './ProtocolEngine';
+import StudyMetadata from '../DicomMetadataStore/StudyMetadata';
+import DisplaySet from '../DisplaySetService/DisplaySet';
 
 const EVENTS = {
   STAGE_CHANGE: 'event::hanging_protocol_stage_change',
@@ -10,10 +12,64 @@ const EVENTS = {
 };
 
 class HangingProtocolService {
+  studies: StudyMetadata[];
+  protocols: object[];
+  protocol: object;
+  stage: number;
+  _commandsManager: object;
+  protocolEngine: ProtocolEngine;
+  matchDetails: object[];
+  hpAlreadyApplied: boolean[] = [];
+  customViewportSettings = [];
+  displaySets: DisplaySet[] = [];
+  activeStudy: object;
+  debugLogging: false;
+
+  customAttributeRetrievalCallbacks = {
+    NumberOfStudyRelatedSeries: {
+      name: 'The number of series in the study',
+      callback: metadata =>
+        metadata.NumberOfStudyRelatedSeries ?? metadata.series?.length,
+    },
+    NumberOfSeriesRelatedInstances: {
+      name: 'The number of instances in the display set',
+      callback: metadata => metadata.numImageFrames,
+    },
+    ModalitiesInStudy: {
+      name: 'Gets the array of the modalities for the series',
+      callback: metadata =>
+        metadata.ModalitiesInStudy ??
+        (metadata.series || []).reduce((prev, curr) => {
+          const { Modality } = curr;
+          if (Modality && prev.indexOf(Modality) == -1) prev.push(Modality);
+          return prev;
+        }, []),
+    },
+  };
+  listeners = {};
+  registeredImageLoadStrategies = {};
+  activeImageLoadStrategyName = null;
+  customImageLoadPerformed = false;
+
+  /**
+   * displaySetMatchDetails = <displaySetId, match>
+   * DisplaySetId is the id defined in the hangingProtocol
+   * match is an object that contains information about
+   *
+   * {
+   *   SeriesInstanceUID,
+   *   StudyInstanceUID,
+   *   matchDetails,
+   *   matchingScore,
+   *   sortingInfo
+   * }
+   */
+  displaySetMatchDetails = new Map();
+
   constructor(commandsManager) {
     this._commandsManager = commandsManager;
     this.protocols = [];
-    this.ProtocolEngine = undefined;
+    this.protocolEngine = undefined;
     this.protocol = undefined;
     this.stage = undefined;
     /**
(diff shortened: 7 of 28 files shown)


extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -417,7 +417,7 @@ OHIFCornerstoneViewport.propTypes = {
   displaySets: PropTypes.array.isRequired,
   dataSource: PropTypes.object.isRequired,
   viewportOptions: PropTypes.object,
-  displaySetOptions: PropTypes.arrayOf(PropTypes.object),
+  displaySetOptions: PropTypes.arrayOf(PropTypes.any),
```
wayfarer3130, line 420: Just reduces warnings sometimes.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -293,7 +294,13 @@ class CornerstoneViewportService implements IViewportService {
 
     viewport.setStack(imageIds, initialImageIndexToUse).then(() => {
       viewport.setProperties(properties);
+      const { initialScale } = viewportOptions;
```
wayfarer3130, line 297: This allows setting the initial scaling parameters - there seems to be other parameters of this type, but I wasn't sure if this was how it should get added/specified.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -473,12 +473,12 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
 
       DicomMetadataStore.addSeriesMetadata(seriesSummaryMetadata, madeInClient);
 
-      const numberOfSeries = seriesPromises.length;
```
wayfarer3130, line 476: The code here returns before the series actually get loaded, breaking the contract on the call.  That made it impossible to know when the series load finished, which is the point required to actually determine the hanging protocol.

extensions/tmtv/src/commandsModule.js
```diff
@@ -60,9 +60,8 @@ const commandsModule = ({
       // that are in the viewer layout (but then we have the problem of the attenuation
       // corrected PT vs the non-attenuation correct PT)
       const matches = HangingProtocolService.getDisplaySetsMatchDetails();
-
       const matchedSeriesInstanceUIDs = Array.from(matches.values()).map(
-        ({ SeriesInstanceUID }) => SeriesInstanceUID
+        match => (match?.SeriesInstanceUID || '')
```
wayfarer3130, line 64: Just fix an exception if the match comes up empty - this is a permitted situation, although it probably should cause a different HP to be applied, but there isn't a good way to do that here.

modes/longitudinal/src/index.js
```diff
@@ -123,15 +123,15 @@ function modeFactory({ modeConfiguration }) {
       const modalities_list = modalities.split('\\');
 
       // Slide Microscopy modality not supported by basic mode yet
-      return !modalities_list.includes('SM');
+      return !!modalities_list.filter(modality => modality != 'SM' && modality !== 'ECG').length;
```
wayfarer3130, line 126: Fixing the test condition to work correctly - it allows applying the mode as long as it contains at least one other modality.

platform/core/src/services/DicomMetadataStore/DicomMetadataStore.js
```diff
@@ -211,7 +211,15 @@ const BaseImplementation = {
     let study = _getStudy(StudyInstanceUID);
     if (!study) {
       study = createStudyMetadata(StudyInstanceUID);
+      // Will typically be undefined with a compliant DICOMweb server, reset later
```
wayfarer3130, line 214: The study description is not a required field at the series level, so it is possible this will come up blank against a valid DICOMweb service, which means it will fail to apply the tmtv hanging protocol

platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
```diff
@@ -13,6 +13,7 @@ function createStudyMetadata(StudyInstanceUID) {
      */
     addInstanceToSeries: function(instance) {
       const { SeriesInstanceUID } = instance;
+      if (!this.StudyDescription) this.StudyDescription = instance.StudyDescription;
```
wayfarer3130, line 16: Recover the study description when teh first instance gets loaded - this instance is required to have the study description.

platform/core/src/services/HangingProtocolService/HPMatcher.js
```diff
@@ -71,6 +71,9 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
       // mark that a required Rule has failed
       if (rule.required) {
         requiredFailed = true;
+        // console.log("Failed required readValues", attribute, readValues, rule, errorMessages)
```
wayfarer3130, line 74: The console methods are really helpful when figuring out why an HP isn't applying, so I've left them commented

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -295,6 +304,7 @@ class HangingProtocolService {
   _updateViewports() {
     // Make sure we have an active protocol with a non-empty array of display sets
     if (!this._getNumProtocolStages()) {
+      console.log("No protocol stages - nothing to display");
```
wayfarer3130, line 307: Just add messages so there is some idea why an HP failed to run.

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -339,9 +351,19 @@ class HangingProtocolService {
...
     // only contains the information of the id of the displaySet to be matched
     // based on some rules
+    const matchedUids = {};
+    this.customAttributeRetrievalCallbacks.alreadyMatched = (metadata) => {
+      return !!matchedUids[metadata.uid];
+    }
+
     stageModel.displaySets.forEach(displaySet => {
-      const { bestMatch } = this._matchImages(displaySet);
+      const { bestMatch, matchingScores } = this._matchImages(displaySet);
```
wayfarer3130, line 360: The matchingScores list is already returned, and gets used to allow a hanging protocol to specify a best match other than the 0 match, or to specify the zero match a second time, with alternate parameters.

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -339,9 +351,19 @@ class HangingProtocolService {
...
       this.displaySetMatchDetails.set(displaySet.id, bestMatch);
+      if (bestMatch) {
+        const { displaySetInstanceUID } = bestMatch;
+        matchedUids[displaySetInstanceUID] = true;
+        bestMatch.matchingScores = matchingScores;
```
wayfarer3130, line 365: The bestMatch also has the list of other matches, in case the viewport species an alternative view.

platform/core/src/services/HangingProtocolService/ProtocolEngine.js
```diff
@@ -10,8 +10,16 @@ export default class ProtocolEngine {
...
   }
 
-  run(studyMetaData) {
+  /** Evaluate the hanging protocol matches on the given:
+   * @param props.studies is a list of studies to compare against (for priors evaluation)
+   * @param props.study is the current metadata for the study to display.
+   *     This object may be modified at the top level ONLY
+   * @param props.displaySets are the list of display sets which can be modified.
+   */
+  run({ studies, studyMetaData, displaySets }) {
```
wayfarer3130, line 19: Pass the values as a set of parameters

platform/core/src/services/HangingProtocolService/lib/validator.js
```diff
@@ -1,20 +1,23 @@
 import validate from 'validate.js';
 
 validate.validators.equals = function (value, options, key, attributes) {
-  if (options && value !== options.value) {
-    return key + 'must equal ' + options.value;
+  const testValue = options?.value ?? options;
```
wayfarer3130, line 4: Use the raw options value to make specifying the comparison easier - for example:
equals: 5
rather than:
equals: { value: 5 },

platform/ui/src/components/ViewportActionBar/ViewportActionBar.tsx
```diff
@@ -336,7 +336,10 @@ ViewportActionBar.propTypes = {
     isTracked: PropTypes.bool.isRequired,
     isRehydratable: PropTypes.bool.isRequired,
     studyDate: PropTypes.string.isRequired,
-    currentSeries: PropTypes.string.isRequired,
+    currentSeries: PropTypes.oneOfType([
```
wayfarer3130, line 339: Should I extract a PropTypes class that is string | number?

platform/viewer/pluginConfig.json
```diff
@@ -27,12 +27,17 @@
...
       "version": "3.0.0"
+
     }
   ],
   "modes": [
     {
       "packageName": "@ohif/mode-longitudinal",
       "version": "3.0.0"
+    },
+    {
```
wayfarer3130, line 38: I will undo this change before commit, but it is required to test the HP changes.

platform/viewer/public/config/local_static.js
```diff
@@ -12,9 +13,9 @@ window.config = {
       sourceName: 'dicomweb',
       configuration: {
         name: 'DCM4CHEE',
-        wadoUriRoot: '/dicomweb',
-        qidoRoot: '/dicomweb',
-        wadoRoot: '/dicomweb',
+        wadoUriRoot: 'http://localhost:5000/dicomweb',
```
wayfarer3130, line 16: Having the yarn start serve up the dicomweb file results in a number of issues, so I'm separating it out.  This is only for testing right now.

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,20 +38,41 @@ async function defaultRouteInit({
...
 
-  const { unsubscribe: seriesAddedUnsubscribe } = DicomMetadataStore.subscribe(
-    DicomMetadataStore.EVENTS.SERIES_ADDED,
-    ({ StudyInstanceUID, madeInClient }) => {
-      const studyMetadata = DicomMetadataStore.getStudy(StudyInstanceUID);
-      if (!madeInClient) {
-        HangingProtocolService.run(studyMetadata);
-      }
+  const { unsubscribe: seriesAddedUnsubscribe } = DisplaySetService.subscribe(
+    DisplaySetService.EVENTS.DISPLAY_SETS_CHANGED,
```
wayfarer3130, line 42: The trigger for hanging protocol has to be when the display sets are computed, because a viewport has a display set in it, not a series.  That allows leaving the display set logic in the extensions, which otherwise would need to be moved into the hanging protocol itself, making each one very fragile.

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,20 +38,41 @@ async function defaultRouteInit({
...
+      const studyMetaData = studies[0];
+      HangingProtocolService.run({ studies, studyMetaData, displaySets });
     }
   );
   unsubscriptions.push(seriesAddedUnsubscribe);
 
+  // No point trying to fire partial retrieves, as it causes havoc
+  // when some series are not displayable, and causes the wrong series
+  // to be fetched.
+  DisplaySetService.holdChangeEvents();
```
wayfarer3130, line 67: Hold off on the display set change notification until everything is read in.

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -43,7 +43,17 @@ class HangingProtocolService {
     this.hpAlreadyApplied = [];
     this.studies = [];
     this.customViewportSettings = [];
-    this.customAttributeRetrievalCallbacks = {};
+    this.customAttributeRetrievalCallbacks = {
+      NumberOfStudyRelatedSeries: (metadata) => metadata.NumberOfStudyRelatedSeries ?? metadata.series?.length,
```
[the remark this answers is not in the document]
swederik: For logging I think we should keep the original structure of customAttributeRetrievalCallbacks, which was an object with a name and a callback function: https://github.com/OHIF/Viewers/blob/2f443fa2efa223ac0fcdf238eee834c6a4ee3fc1/platform/core/src/hanging-protocols/customAttributes.js#L24

(in the past we had a UI which would display the name of the custom attribute, which is why it was there)

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -43,7 +43,17 @@ class HangingProtocolService {
     this.hpAlreadyApplied = [];
     this.studies = [];
     this.customViewportSettings = [];
-    this.customAttributeRetrievalCallbacks = {};
+    this.customAttributeRetrievalCallbacks = {
```
wayfarer3130, line 46: This should be an object [id]: {callback: function, name: display name}

platform/core/src/services/HangingProtocolService/HangingProtocolService.js
```diff
@@ -142,14 +150,38 @@ class HangingProtocolService {
...
-   * @param options
+   * It is possible to add more than one setting to a given id, in which case all are called.
+   * @param id The ID used to refer to the setting (e.g. 'displayCADMarkers')
+   * @param name The name of the setting to be displayed (e.g. 'Display CAD Markers')
    * @param callback A function to be run after a viewport is rendered with a series
+   * @param options
    */
-  addCustomViewportSetting(...params) {
-    this.customViewportSettings.push(...params);
+  addCustomViewportSetting(id, name, callback, options) {
```
wayfarer3130, line 159: This should be stored as an object, same as above.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -39,6 +39,8 @@ class CornerstoneViewportService implements IViewportService {
   renderingEngine: Types.IRenderingEngine | null;
   viewportsInfo: Map<number, ViewportInfo>;
   viewportGridResizeObserver: ResizeObserver | null;
+  // TODO - get the right type here.
```
wayfarer3130, line 42: I've made HangingProtocolService typescript now, but haven't created an interface yet for it.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -166,6 +168,17 @@ class CornerstoneViewportService implements IViewportService {
     const viewportInfo = this.viewportsInfo.get(viewportIndex);
     viewportInfo.setRenderingEngineId(renderingEngine.id);
 
+    const priorViewportOptions = viewportInfo.getViewportOptions();
+    const viewport = renderingEngine.getViewport(viewportInfo.getViewportId());
+    if (viewport) {
```
wayfarer3130, line 173: This stuff just remembers the data - if no applier is found, then nothing new happens, but it at least allows remembering the viewport data.

platform/viewer/pluginConfig.json
```diff
@@ -27,12 +27,17 @@
...
+      "packageName": "@ohif/mode-tmtv",
+      "version": "3.0.0"
```
sedghi, line 40: remove this please, not ready yet

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -39,6 +39,8 @@ class CornerstoneViewportService implements IViewportService {
   renderingEngine: Types.IRenderingEngine | null;
   viewportsInfo: Map<number, ViewportInfo>;
   viewportGridResizeObserver: ResizeObserver | null;
+  // TODO - get the right type here.
+  hangingProtocolService: object;
```
sedghi, line 43: I'm not sure if we should use `any` or `object`, I haven't seen object anywhere. @ladeirarodolfo

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -292,6 +295,7 @@ class CornerstoneViewportService implements IViewportService {
     }
 
     viewport.setStack(imageIds, initialImageIndexToUse).then(() => {
+      hangingProtocolService.applyCustomViewportSettings(viewportOptions, viewport, properties);
```
sedghi, line 298: General comment (thinking out loud)

First of all we should be consistent between settings and options. If it is viewportOptions, then it should be `applyCustomViewportOptions`

I understand what you are doing here. However, I don't understand why we are adding `viewportSettings/Options` through the `HangingProtocolService` and then calling the `apply...`. Can't we just those callbacks here, instead of adding and then applying?

`customViewportOptions` was originally added to have simple key-value pairs, but I also understand the value of the functions, which makes `Options` more general. I think `customViewportOptions` should remain as simple key-value pairs accessible throughout the code and maybe we add a new `customViewportCallbacks` that get executed? but can we serialize that in JSON?

sedghi: I also don't understand why `properties` is being passed to the apply callback?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -554,6 +563,7 @@ class CornerstoneViewportService implements IViewportService {
       viewportInfo.getViewportId()
     );
 
+    this.hangingProtocolService.applyCustomViewportStore(publicViewportOptions, priorViewportOptions, viewport);
```
sedghi, line 566: I don't like the idea of adding a store for the HangingProtocolService, it gets out of hand immediately e.g,. currently displaying extensionA's viewport, then drag and drop a displaySet which will cause extensionB's viewport to mount. You can achieve the store functionality by an external service which will listen to enabledElement or so.

platform/core/src/services/DicomMetadataStore/DicomMetadataStore.js
```diff
@@ -211,7 +211,15 @@ const BaseImplementation = {
...
       study.StudyDescription = seriesSummaryMetadata[0].StudyDescription;
+      study.ModalitiesInStudy = [];
```
sedghi, line 216: First: `Modalities` is simple and representative enough, since it is accessed in the study (`study.Modalities`)

Second: You should add it to `createStudyMetadata` too

platform/core/src/services/DicomMetadataStore/DicomMetadataStore.js
```diff
@@ -211,7 +211,15 @@ const BaseImplementation = {
...
+      seriesSummaryMetadata.forEach(item => {
+        if (study.ModalitiesInStudy.indexOf(item.Modality) == -1) {
+          study.ModalitiesInStudy.push(item.Modality);
+        }
+      });
+      study.NumberOfStudyRelatedSeries = seriesSummaryMetadata.length;
```
sedghi, line 222: same here, it should be defined in the `create...`

Also isn't `study.series.length` the same length as `NumberOfStudyRelatedSereis`? why do you need to define a new one?

platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
sedghi: use {} for the if statements, same as through out the code

platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
```diff
@@ -28,10 +29,12 @@ function createStudyMetadata(StudyInstanceUID) {
      *
      * @param {object[]} instances
      * @param {string} instances[].SeriesInstanceUID
+     * @param {string} instances[].StudyDescription
      * @returns {bool} true if series were added; false if series already exist
      */
     addInstancesToSeries: function(instances) {
       const { SeriesInstanceUID } = instances[0];
+      if (!this.StudyDescription) this.StudyDescription = instances[0].StudyDescription;
```
sedghi, line 37: {}

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -10,6 +10,57 @@ const EVENTS = {
...
+  protocol: object;
+  stage: number;
+  _commandsManager: object;
+  ProtocolEngine: object;
+  matchDetails: object[];
+  hpAlreadyApplied: boolean[] = [];
+  customViewportSettings = [];
+  displaySets = [];
+  customAttributeRetrievalCallbacks = {
+    NumberOfStudyRelatedSeries: {
```
sedghi, line 24: Why we are hardcoding these here? These should be added by the mode/extension if it is of mode/extension interest.

sedghi: The same for `NumberOfSeriesRelatedInstances` and `ModalitiesInStudy`. It should not be part of core service, the API to add exists for extensions to add (if they want)

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -99,7 +127,11 @@ class HangingProtocolService {
 
     // if there is no pre-defined protocol
     if (!protocol || protocol.id === undefined) {
-      const matchedProtocol = this.ProtocolEngine.run(metaData);
+      const matchedProtocol = this.ProtocolEngine.run({
+        studies: this.studies,
+        studyMetaData: studies[0],
```
sedghi, line 132: There was a reason that we copied the studyMetadata with Object.assign, please revert/include the copy instead of direct reference

sedghi: If you are passing both `studies` and `studyMetadata` (Metadata instead of MetaData - to be consistent with cs3d onwards), then can we name `studyMetadata` as `activeStudyMetadata`?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -99,7 +127,11 @@ class HangingProtocolService {
...
+        displaySets,
```
sedghi, line 133: also, if this is active study displaysets, it should be renamed

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -130,26 +162,61 @@ class HangingProtocolService {
...
+  public applySingleCustomViewportSetting(id, value, viewport, ...args) {
+    let callCount = 0;
+    for (const setting of this.customViewportSettings) {
+      // Can have pure storage settings as well.
+      if (setting.id !== id || !setting.callback) continue;
+      setting.callback(id, value, viewport, ...args);
+      callCount += 1;
+    }
+    if (callCount === 0) console.warn("No custom setting found for", id);
   }
```
sedghi, line 220: See my general comment for these in the `CornerstoneViewportService`

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -159,18 +226,21 @@ class HangingProtocolService {
...
    * @param attributeName The name of the attribute to be displayed (e.g. 'Timepoint Type')
    * @param callback The function used to calculate the attribute value from the other attributes at its level (e.g. study/series/image)
-   */
-  addCustomAttribute(attributeId, attributeName, callback) {
+   * @param options to add to the "this" object for the custom attribute retriever
+  */
+  public addCustomAttribute(attributeId, attributeName, callback, options) {
     this.customAttributeRetrievalCallbacks[attributeId] = {
+      ...options,
+      id: attributeId,
```
sedghi, line 234: why do you need `id` in the value object? it exist as the key, no?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -352,17 +429,22 @@ class HangingProtocolService {
       // DisplaySets for the viewport, Note: this is not the actual displaySet,
       // but it is a info to locate the displaySet from the displaySetService
       let displaySetsInfo = [];
-      viewport.displaySets.forEach(({ id, options: displaySetOptions }) => {
-        const viewportDisplaySet = this.displaySetMatchDetails.get(id);
+      viewport.displaySets.forEach(({ id, displaySetIndex = 0, options: displaySetOptions }) => {
```
sedghi, line 432: can you please document why `displaySetIndex` is destructured here? I don't recall havving that in the HP

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -352,17 +429,22 @@ class HangingProtocolService {
...
-          const { SeriesInstanceUID } = viewportDisplaySet;
+          const { SeriesInstanceUID, displaySetInstanceUID } = viewportDisplaySet;
 
           const displaySetInfo = {
             SeriesInstanceUID,
+            displaySetInstanceUID,
             displaySetOptions,
           };
 
+          // console.log("Choose displaySetInstanceUID", displaySetInstanceUID, SeriesInstanceUID);
```
sedghi, line 447: remove

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -382,50 +464,52 @@ class HangingProtocolService {
...
-          seriesMatchDetails.requiredFailed === true ||
-          seriesMatchDetails.score < highestSeriesMatchingScore
-        ) {
+        if (seriesMatchDetails.requiredFailed === true) {
+          // console.log("Display set required failed", displaySet, seriesMatchingRules);
           return;
         }
 
-        highestSeriesMatchingScore = seriesMatchDetails.score;
+        // console.log("Found displaySet for rules", displaySet);
```
sedghi, line 511: so many commented console.log in this file

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -450,21 +534,27 @@ class HangingProtocolService {
...
           },
         };
 
+        // console.log("Adding display set", displaySet, imageDetails);
         matchingScores.push(imageDetails);
       });
     });
 
+    if (matchingScores.length === 0) {
+      console.log("No match found");
```
sedghi, line 555: This is so weird in the console as there is no context. Maybe `HangingProtocolService:: No Match found for the specified rules`?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -382,50 +464,52 @@ class HangingProtocolService {
...
-    const { studyMatchingRules, seriesMatchingRules } = displaySet;
+    const { studyMatchingRules = [], seriesMatchingRules, findAll = false } = displaySetRules;
 
     const matchingScores = [];
     let highestStudyMatchingScore = 0;
     let highestSeriesMatchingScore = 0;
 
+    console.log('ProtocolEngine::matchImages', studyMatchingRules, seriesMatchingRules);
     this.studies.forEach(study => {
+      // TODO - only pass in display sets matching the top level study being tested
```
sedghi, line 481: This is an easy todo to fix, `this.displaySets.filter(ds => ds.StudyInstanceUID..........)

platform/core/src/services/HangingProtocolService/ProtocolEngine.js
```diff
@@ -62,33 +70,35 @@ export default class ProtocolEngine {
...
    * Finds the best protocols from Protocol Store, matching each protocol matching rules
    * with the given study. The best protocol are ordered by score and returned in an array
    * @param  {Object} study StudyMetadata instance object
+   * @param {object} extraMatchData to match on.
    * @return {Array}       Array of match objects or an empty array if no match was found
    *                       Each match object has the score of the matching and the matched
    *                       protocol
    */
-  findMatchByStudy(study) {
+  findMatchByStudy(study, extraMatchData) {
```
sedghi, line 101: I would rename `extraMatchData` to `options` or something similar, there is no need to have `Match` in the name since the method name is `findMatchByStudy`

platform/core/src/services/HangingProtocolService/HPMatcher.js
sedghi: So many commented console.log in this PR, please remove them all

platform/core/src/services/HangingProtocolService/HPMatcher.js
```diff
@@ -16,32 +19,32 @@ const match = (metadataInstance, rules, customAttributeRetrievalCallbacks) => {
...
   };
 
+  const readValues = {};
+
   let requiredFailed = false;
   let score = 0;
 
   rules.forEach(rule => {
-    const attribute = rule.attribute;
+    const { attribute, source } = rule;
```
sedghi, line 28: what is `source` in the rule? we did not have this IIRC

platform/viewer/src/components/ViewportGrid.tsx
```diff
@@ -42,18 +42,23 @@ function ViewerViewportGrid(props) {
           continue;
         }
 
+        // console.log("Applying HP matchDetails", i, matchDetails[i]);
```
sedghi, line 45: commented console.log (I haven't commented on all of these, but there are at least 20 unnecessary line additions)

platform/viewer/src/components/ViewportGrid.tsx
```diff
@@ -42,18 +42,23 @@ function ViewerViewportGrid(props) {
...
         const displaySetUIDsToHangOptions = [];
-        displaySetsInfo.forEach(({ SeriesInstanceUID, displaySetOptions }) => {
-          const matchingDisplaySet = availableDisplaySets.find(ds => {
-            return ds.SeriesInstanceUID === SeriesInstanceUID;
-          });
-
+        displaySetsInfo.forEach(({ SeriesInstanceUID, displaySetInstanceUID, displaySetOptions }) => {
+          const matchingDisplaySet =
+            availableDisplaySets.find(ds => ds.displaySetInstanceUID == displaySetInstanceUID) ||
+            availableDisplaySets.find(ds => ds.SeriesInstanceUID === SeriesInstanceUID);
```
sedghi, line 56: This is an unnecessary long operation, can you break it for readability into `matchingDisplaySetByUID` and `matchingDisplaySetBySeries` maybe?

platform/viewer/src/components/ViewportGrid.tsx
```diff
@@ -42,18 +42,23 @@ function ViewerViewportGrid(props) {
...
+
+          // console.log("Displaying", matchingDisplaySet.displaySetInstanceUID);
+          if (matchingDisplaySet.displaySetInstanceUID !== displaySetInstanceUID) {
```
sedghi, line 59: So I guess this happens when the availalbeDisplaySets cannot find the DSUID, and it is matched by the SeriesUID, right? But, if we have used DS to match in the HP, there is always a garantee that it exists (since we matched and sorted them out).

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,20 +38,41 @@ async function defaultRouteInit({
...
-    DicomMetadataStore.EVENTS.SERIES_ADDED,
-    ({ StudyInstanceUID, madeInClient }) => {
-      const studyMetadata = DicomMetadataStore.getStudy(StudyInstanceUID);
-      if (!madeInClient) {
-        HangingProtocolService.run(studyMetadata);
-      }
+  const { unsubscribe: seriesAddedUnsubscribe } = DisplaySetService.subscribe(
+    DisplaySetService.EVENTS.DISPLAY_SETS_CHANGED,
+    displaySets => {
+      if (!displaySets || !displaySets.length) return;
```
sedghi, line 44: {}

platform/viewer/src/routes/Mode/Mode.tsx
sedghi: This makes sense. 
Bring back the `madeInClient` stuff since you don't want this to run when a new displaySet has been created by the user (e.g., export SR). The `madeInClient` make sure this is `.run` only happens when it is coming from the server

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,20 +38,41 @@ async function defaultRouteInit({
...
+      // The assumption is that the display set at position 0 is the first
```
sedghi, line 45: Bring back the `madeInClient` stuff, since you don't want this subscription to run on SR export (client side generated DS)

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,20 +38,41 @@ async function defaultRouteInit({
...
+      HangingProtocolService.run({ studies, studyMetaData, displaySets });
     }
   );
   unsubscriptions.push(seriesAddedUnsubscribe);
 
+  // No point trying to fire partial retrieves, as it causes havoc
+  // when some series are not displayable, and causes the wrong series
+  // to be fetched.
+  DisplaySetService.holdChangeEvents();
+  const allRetrieves = [];
```
sedghi, line 68: Can you make it a `map` instead?

```
const allRetrieves = studyInstanceUIDs.forEach((StudyInstanceUID) => {
  return dataSource.retrieve.series.metadata({ StudyInstanceUID })
})
```

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,20 +38,41 @@ async function defaultRouteInit({
...
   studyInstanceUIDs.forEach(StudyInstanceUID => {
-    dataSource.retrieve.series.metadata({ StudyInstanceUID });
+    const item = dataSource.retrieve.series.metadata({ StudyInstanceUID });
+    allRetrieves.push(item);
   });
+  Promise.all(allRetrieves).then(() => {
```
sedghi, line 73: This is already an async function, so you can simply await here

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,20 +38,41 @@ async function defaultRouteInit({
...
+        }, []);
+      const studyMetaData = studies[0];
+      HangingProtocolService.run({ studies, studyMetaData, displaySets });
     }
   );
   unsubscriptions.push(seriesAddedUnsubscribe);
 
+  // No point trying to fire partial retrieves, as it causes havoc
+  // when some series are not displayable, and causes the wrong series
+  // to be fetched.
```
sedghi, line 66: I'm not sure If I understand why we need to hold off here? Aren't we matching on the displaySets of the current study? Do you need other studies' metadata to load first? Can you please add more comments?

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,20 +38,41 @@ async function defaultRouteInit({
...
+      const studies = displaySets.
+        reduce((prev, curr) => {
+          const { StudyInstanceUID } = curr;
+          if (!studyMap[StudyInstanceUID]) {
+            const study = DicomMetadataStore.getStudy(StudyInstanceUID);
+            studyMap[StudyInstanceUID] = study;
+            prev.push(study);
+          }
+          return prev;
+        }, []);
```
sedghi, line 57: So we are creating studies object here again and again, when DS are being added? Why there is no garantee that `.getStudy` gives us the study metadata? since later you just use the first study `const studyMetadata = studies[0]` 

Also here we are running the HP engine on EVERY displaySet creation, is that right? Does that mean that every time there is a new ranking (since DS are in the process of creation as we match).

sedghi (changes requested):
Please see my comments

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
wayfarer3130: Changed the name consistently to applyCustomViewportOptions
The reason one adds them through the HangingProtocolService is that they are defined in the hanging protocol service as part of hte protocol itself - that is, in 
export type PublicViewportOptions = {
  customViewportOptions?: Record<string, unknown>;
};
it already defines the key/value pairs.  However, it doesn't define what to do with them, and if they can be handled by the extension/viewport directly, then they should be direct options - you didn't like adding direct options to CS3D, so one needs the ability to apply the custom options in some fashion.  That pretty much requires a setter.  Now, it would be possible to define that on the cornerstone service itself, but it makes sense to put them with the spot they are defined, thus on the hanging protocol service.  It doesn't matter what type of viewport it is - if the viewport is a completely new one, then one just needs to write new apply code for it.
It doesn't make sense to add generic custom viewport options for all hanging protocols, so I think this design is in line with the general design.

wayfarer3130: properties is just being passed because I thought it might be important to have access to.  I'm not currently using it, but I wasn't sure if it would be required.  There is no guarantee as to what gets passed beyond the viewport

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
wayfarer3130: No actually, you can't achieve it because it gets called on both the old and the new settings values.  If you want to have access to it, then you need to build it in.  Since the hanging protocol service owns the viewport settings, that is the appropriate place to build it in.  If they don't get used, then the calls are just no-ops, but if you need it, I don't see a good way of getting access other than constant forking the OHIF code.

The general pattern I'm trying to make in changes is to provide hooks into various places to allow changes to be made, without otherwise making deep changes, so that it becomes possible to inject certain types of changes.  Yes, if the underlying code is re-implemented, and the hook isn't applicable, then that change doesn't apply any longer (eg if the viewport changes such that setPan/setZoom no longer apply), but that is the nature of hooks, that they depend on exactly the spot they get added to.

platform/core/src/services/DicomMetadataStore/DicomMetadataStore.js
wayfarer3130: ModalitiesInStudy in the DICOM name, and this is just recording the value when a study query isn't directly performed.  However it is good to add it to createStudyMetadata as well.

platform/core/src/services/DicomMetadataStore/DicomMetadataStore.js
wayfarer3130: Because NumberOfStudyRelatedSeries is the formal DICOM name, and the metadata is supposed to mirror the DICOM names, it is just that the particular loading strategy chosen doesn't allow getting the value.

platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
wayfarer3130: I really dislike making the code less readable for the sake of formal style rules, but whatever.  A single line keeping the code shorter makes it easier to read because it is obvious that line is a single unit.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: They are added because they are failings of the metadata loader service to be able to load some of the standard values.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -57,40 +87,38 @@ class HangingProtocolService {
...
 
-  run(studyMetaData, protocol) {
-    if (!this.studies.includes(studyMetaData)) {
-      this.studies.push(studyMetaData);
-    }
-    // copy here so we don't mutate it
-    const metaData = Object.assign({}, studyMetaData);
+  public run({ studies, displaySets }, protocol) {
+    this.studies = [...studies];
+    this.displaySets = displaySets;
```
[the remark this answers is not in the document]
wayfarer3130: It is all display sets for all studies.  One needs to sub-select them to get just the active ones for a particular study.  The HP service doesn't yet work with multi-study hanging protocols.  I'd rather not have more changes right now to deal with multi-study support.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: The reason for the object copy was that the object was modified internally.  That is no longer happening as the matcher creates a new referenced object containing just the values being matched, so it isn't necessary to have an extra copy.  If you want an extra copy of that, then the display sets should also be fully duplicated everywhere they get accessed.  It also wasn't a fully hierarchical copy, so the old code incorrectly modified the series data.  It would be really nice to have a readonly attribute on the value.

platform/viewer/src/routes/Mode/Mode.tsx
wayfarer3130: There was existing code that built up the list of studies in the HangingProtocolService itself, assuming that the run method got called multiple times, as each series got added.  However, it was broken because the order of the calls HAD to match the order of the series being matched or else it would fail to display the right series (eg the correct match).  That allowed the Pet SUV table to work as expected IF the basic match worked, but would fail to display anything if the first series found was incorrect.    So here, what I'm trying to do is make it work all at once, with a single DisplaySets changed event.  The change event only gets called once per study creation, right at the end - everything else is held until that point (see the holdChangeEvents code in DisplaySetService).   Then, since the run method needs the list of studies, the list is created externally and passed in - since we don't want to depend on the metadata service.
I'm trying to make this not have race conditions while minimizing how many changes are made.

platform/viewer/src/routes/Mode/Mode.tsx
wayfarer3130: Because if the order of the series arrival doesn't match the display set matches, then you never see anything.

platform/viewer/src/routes/Mode/Mode.tsx
wayfarer3130: Actually one only wants the rendering to occur the first time things arrive, which was what was previously causing problems.    The problem with doing it multiple times is that it messes up the display if it runs a subsequent time.  There are quite a few issues with that, but I don't think I can solve all of them.  The made in client stuff only applies to the added event, so what I will do for now is not fire the changed event if it is made int he client, assuming that is what is desired.

wayfarer3130: There is already a comment in the display set service indicating it isn't quite right - what should be done is to have an event that fires once the study metadata is loaded, and fires once, then have the added/removed/changed events, which are on:
1.  When a new display set is added
2. When a display set is removed
3. When a display set is changed, but NOT when it is added or removed
I've changed the display set to not call the changed if made in client is set, under the assumption that it already knows about that, but it seems like a bad idea.

platform/viewer/src/routes/Mode/Mode.tsx
wayfarer3130: Why?  I don't mind doing it, but would like to know why you think that is a good idea.  Objects in Javascript are in fact maps, so I don't see why it is necessary or desirable.

platform/viewer/src/routes/Mode/Mode.tsx
wayfarer3130: It is an async function, but it doesn't actually need to be, since the result of the function is a set of unsubscriptions, and the behaviour being added isn't actually used as part of the result, it is just something that needs to happen once the subscription is completed.

wayfarer3130: Removed the async function part of it - this shouldn't have to block the rest of the loading.

platform/viewer/src/components/ViewportGrid.tsx
wayfarer3130: Yes, will delete that code.

platform/viewer/src/routes/Mode/Mode.tsx
sedghi: with forEach and push you are basically doing what is implemented via map which is transformation of an array as we are doing here

platform/viewer/src/routes/Mode/Mode.tsx
sedghi: It is async, since this can be overwritten by a mode that has async operations for `init`

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
wayfarer3130: They are fairly different - any includes  number, string, function, whereas object means a JavaScript object value, so I'd prefer to be more descriptive here since we actually know that it is an object and NOT just any old value.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: It is all the display sets associated with the studies.  There might be a concept of inactive display sets, but I'm not quite sure what it means - it is NOT active in terms of what is actually being shown right now, because that hasn't yet been determined.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: The HangingProtocolService owns the application of the hanging protocol, and the custom viewport settings were previously owned by the HPS, so I would claim they still should be owned.
The store service is part of the HPS because it describes how to apply newly displayed data in the context of an existing hanging protocol.

platform/core/src/services/HangingProtocolService/HPMatcher.js
wayfarer3130: TODO - add documentation on the source attribute, or remove it.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
swederik: I don't think we use inactive display sets at the moment. It was originally envisioned to be there for cases when you switched modes within a session. Some display sets would be created for one mode but not relevant for another. We have previously discussed nuking it though since it's not being used and I don't think we need it.

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,19 +42,45 @@ async function defaultRouteInit({
...
-    dataSource.retrieve.series.metadata({ StudyInstanceUID });
+  // The hanging protocol matching service is fairly expensive to run multiple
+  // times, and doesn't allow partial matches to be made (it will simply fail
+  // to display anything if a required match failes), so hold off the matches
+  // here until the entire study is ready.
+  DisplaySetService.holdChangeEvents();
+  const allRetrieves = studyInstanceUIDs.map(StudyInstanceUID =>
+    dataSource.retrieve.series.metadata({ StudyInstanceUID })
+  );
+  Promise.all(allRetrieves).then(() => {
```
swederik, line 82: Error handling? Promise.allSettled?

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -38,19 +42,45 @@ async function defaultRouteInit({
 
   unsubscriptions.push(instanceAddedUnsubscribe);
 
-  const { unsubscribe: seriesAddedUnsubscribe } = DicomMetadataStore.subscribe(
```
wayfarer3130, line 41: @sedghi  The old workflow here was buggy.  The problem is that the series retrieve ordering is non-deterministic on SOME back ends.  As long as the series come in in the ordering required by the hanging protocol, the study will hang correctly.  However, if the series comes in a different order, then the match can fail, and it won't retry that failure, so you don't get to see anything.

sedghi (approved):
This is a great addition, great job

PR #2838, opened 2022-07-07
feat(OHIF):Allow modes and extensions to be added after compile time.

wayfarer3130:
This is just the changes required to allow the modes and extensions to be loaded dynamically.
It also includes automatically loading the public/* directory of the linked files, and fixes some build issues on the cli generated output (required to allow the dynamic load to work).


platform/viewer/.webpack/webpack.pwa.js
```diff
@@ -108,13 +111,14 @@ module.exports = (env, argv) => {
...
-      // new FaviconsWebpackPlugin( path.join(PUBLIC_DIR, 'assets', 'icons-512.png')),
+      // Generate a service worker for fast local loads
       new InjectManifest({
         swDest: 'sw.js',
         swSrc: path.join(SRC_DIR, 'service-worker.js'),
         // Increase the limit to 4mb:
-        maximumFileSizeToCacheInBytes: 4 * 1024 * 1024,
+        maximumFileSizeToCacheInBytes: 5 * 1024 * 1024,
+        // Need to exclude the theme as it is updated independently
+        exclude: [/theme/],
```
wayfarer3130, line 121: This is required so that themes get reloaded successfully, which is needed for the dynamic imports.

platform/viewer/public/config/dicomweb_relative.js
```diff
@@ -0,0 +1,160 @@
+window.config = {
```
wayfarer3130, line 1: This is the old local_static.js configuration, needed for any site that uses a single domain for both dicomweb and OHIF, but is otherwise just a copy.

platform/viewer/public/config/local_static.js
```diff
@@ -12,9 +13,9 @@ window.config = {
       sourceName: 'dicomweb',
       configuration: {
         name: 'DCM4CHEE',
-        wadoUriRoot: '/dicomweb',
-        qidoRoot: '/dicomweb',
-        wadoRoot: '/dicomweb',
+        wadoUriRoot: 'http://localhost:5000/dicomweb',
```
wayfarer3130, line 16: Need to use a different port for testing the local static server, as doing it through deploy is quite slow.

platform/viewer/src/index.js
```diff
@@ -15,19 +15,21 @@ import ReactDOM from 'react-dom';
...
-/**
- * Combine our appConfiguration with installed extensions and modes.
- * In the future appConfiguration may contain modes added at runtime.
- *  */
-const appProps = {
-  config: window ? window.config : {},
-  defaultExtensions: window.extensions,
-  defaultModes: window.modes,
-};
+loadDynamicImports().then(() => {
```
wayfarer3130, line 20: This code change here is the ONLY active running code that is changed in this PR, and it simply calls a function to load any dynamic imports.   How that function does the load isn't currently defined, but it would be possible to merge some of the code from OHIFModesAndExtensions into core OHIF to make this possible.

PR #2840, opened 2022-07-13
Logout for v3

Michael-Andersen:
Added logout route to OpenIdConnectRoutes and add a button for logout to the menu.


platform/ui/src/assets/icons/power-off.svg
```diff
@@ -0,0 +1,11 @@
+<svg
```
sedghi, line 1: Where is this ICON coming from? We need to be careful about the licensing to be added to the file if necessary.

sedghi (commented):
Thanks for the contribution, please see my comment

platform/ui/src/assets/icons/power-off.svg
Michael-Andersen: I just reused the icon from the v2 logout: https://github.com/OHIF/Viewers/blob/master/platform/ui/src/elements/Icon/icons/power-off.svg

If there's another preferred source of icon, I'm happy to swap this one out.

sedghi:
So I tried it out locally with our QA server (on Google Healthcare), and got a blank screen after redirect. Opening the console I saw the following error 

image

Michael-Andersen:
> So I tried it out locally with our QA server (on Google Healthcare), and got a blank screen after redirect. Opening the console I saw the following error
> 
> image

What does your oidc config look like? The end session endpoint is retrieved from the metadata endpoint on your auth server but the metadata endpoints can be configured explicitly if it's inaccessible (maybe CORS issue?) . See: https://stackoverflow.com/questions/50844515/signoutredirect-of-oidc-client-js-against-auth0-returns-no-end-session-endpoint and the oidc client docs: https://github.com/IdentityModel/oidc-client-js/wiki

My oidc provider (keycloak server) doesn't have this issue but it still works if I explicitly add the metadata property to the config (see below) and it definitely uses these values because if I leave out the token_endpoint then auth fails completely and if I leave off the token_endpoint then I recreate the error you were getting. 

```
metadata: {
        issuer: `https://<keycloak-server>/auth/realms/dcm4che`,
        authorization_endpoint: `https://<keycloak-server>/auth/realms/dcm4che/protocol/openid-connect/auth`,
        token_endpoint: `https://<keycloak-server>/auth/realms/dcm4che/protocol/openid-connect/token`,
        userinfo_endpoint: `https://<keycloak-server>/auth/realms/dcm4che/protocol/openid-connect/userinfo`,
        jwks_uri: 'https://<keycloak-server>/auth/realms/dcm4che/protocol/openid-connect/certs',
      }
```

Let me know if adding this property solves the issue for you.

Michael-Andersen:
By mistake I left out the end_session_endpoint part of my config sample. I posted the config for recreating your issue...this is what I use to provide the metadata and not get an error:

```
metadata: {
        issuer: `https://<keycloak-server>/auth/realms/dcm4che`,
        authorization_endpoint: `https://<keycloak-server>/auth/realms/dcm4che/protocol/openid-connect/auth`,
        token_endpoint: `https://<keycloak-server>/auth/realms/dcm4che/protocol/openid-connect/token`,
        userinfo_endpoint: `https://<keycloak-server>/auth/realms/dcm4che/protocol/openid-connect/userinfo`,
        jwks_uri: 'https://<keycloak-server>/auth/realms/dcm4che/protocol/openid-connect/certs',
        end_session_endpoint: 'https://qipcm-pacs-dev.uhn.ca:8843/auth/realms/dcm4che/protocol/openid-connect/logout'
      }
```

Also I have rebased this branch on top of the latest v3-stable.

Michael-Andersen:
Rebased again and added a workaround for forcing logout of additional browser tabs. Normally to implement this feature the oidc-client library has a userSignedOut Event which you can add a listener for (see: https://github.com/IdentityModel/oidc-client-js/wiki) but the event is only fired due to a call to your OIDC provider's check session iframe endpoint (which would also need to be included in the metadata section of the your oidc config) but that call fails because it doesn't have the right headers because OHIF has headers configured for cross origin isolation (see here for a discussion on how cross origin isolation can break oauth flows: https://blog.stackblitz.com/posts/cross-browser-with-coop-coep/). My workaround is for the manual signout tab to set a localstorage variable (signoutEvent) which is listened for in the other tabs. This variable also needs to be cleared from localstorage when the viewer first loads in case it remains from a previous logout. I've tested this solution and it seems to get the job done, I'm open to other solutions though.

platform/viewer/src/App.tsx
```diff
@@ -28,7 +28,7 @@ let commandsManager, extensionManager, servicesManager, hotkeysManager;
 
 function App({ config, defaultExtensions, defaultModes }) {
   const [init, setInit] = useState(null);
-
+  localStorage.removeItem("signoutEvent"); //Remove possible leftover from previous session
```
sedghi, line 31: This cleanup should happen at the component itself

platform/viewer/src/utils/OpenIdConnectRoutes.tsx
```diff
@@ -119,6 +128,15 @@ function OpenIdConnectRoutes({
...
 
+  //for multi-tab logout
+  window.addEventListener('storage', () => {
+    var signOutEvent = localStorage.getItem("signoutEvent");
+    if (signOutEvent) {
+      navigate(`/logout?redirect_uri=${encodeURIComponent(window.location.href)}`);
+      localStorage.removeItem("signoutEvent");
+    }
+  });
+
```
sedghi, line 139: This even binding should happen in a useEffect on mount and the clean up be also in the return statement of it

sedghi (changes requested):
See my comments about bindings and cleanup

Michael-Andersen:
@sedghi I have rebased and made the changes you requested. Can you take a look when you get a chance?

sedghi:
@Michael-Andersen I pushed a commit, can you check to see if it makes sense to you? https://github.com/OHIF/Viewers/pull/2840/commits/1bd2f180a13ddea45732867322d801badb3170a4#diff-ec71eefedc3ff13c4ee05a5a87e14faa7e1f64662a75a236c554796c2aff9beeR129-R145

Michael-Andersen:
@sedghi I pulled your change in and did some testing. Unfortunately the 'signoutEvent' is not removed from localstorage this way. If you login, logout, and then log back in, you can see the variable still exists in chrome devtools. If you keep that tab open and open the viewer in another tab it will immediately trigger the signout because the local storage variable is still there.

As for the event listener I find if you watch it in chrome devtools it will be destroyed with or without your addition.

So the way I had it on the previous commit works but another alternative is to just to put the removal of the left over localstorage variable at the beginning of the useeffect. Something like this:


```javascript
useEffect(() => {
    localStorage.removeItem('signoutEvent');
    const storageEventListener = event => {
      const signOutEvent = localStorage.getItem('signoutEvent');
      if (signOutEvent) {
        navigate(
          `/logout?redirect_uri=${encodeURIComponent(window.location.href)}`
        );
      }
    };

    window.addEventListener('storage', storageEventListener);
  }, []);
```

If you like this version better I can push it.

sedghi:
My only concern regarding your original commit was that your event listener for 'storage' does not get unsubscribed, so I guess you can mix my commit (which adds the unsubscribtion) and yours to achieve both goals?

Michael-Andersen:
Ok I've made my change and left in the storage event unsubscription from your commit.

sedghi (approved):

PR #2861, opened 2022-08-05
feat(syncGroups):Add registration for new synchronizers

wayfarer3130:
Adds a registration capability to add new synchronizers
Based off the hanging protocol PR, but is intended to merge after that one.
@sedghi and @swederik  - this is a modified design that allows creating new synchronizers.  There is also a bit more cleanup of types in the hanging protocol area to make that cleaner.  The only connection between hanging protocols and the syncrhonizers is the existing syncGroups - otherwise, that area is actually identical.  I do not yet have an example synchronizer fully worked out, but I didn't want to do that until you had seen it.

### DIFF
--- a/extensions/cornerstone/src/init.js
+++ b/extensions/cornerstone/src/init.js
@@ -259,6 +259,11 @@ export default async function init({
   //   }
   // };
 
+  const newStackCallback = evt => {
+    const { element } = evt.detail;
+    utilities.stackPrefetch.enable(element);
+  };
+
   function elementEnabledHandler(evt) {
     const { element } = evt.detail;
 
@@ -267,10 +272,10 @@ export default async function init({
       contextMenuHandleClick
     );
 
-    eventTarget.addEventListener(EVENTS.STACK_VIEWPORT_NEW_STACK, evt => {
-      const { element } = evt.detail;
-      utilities.stackPrefetch.enable(element);
-    });
+    eventTarget.addEventListener(
+      EVENTS.STACK_VIEWPORT_NEW_STACK,
+      newStackCallback
+    );
   }
 
   function elementDisabledHandler(evt) {
@@ -284,10 +289,11 @@ export default async function init({
       contextMenuHandleClick
     );
 
-    eventTarget.removeEventListener(EVENTS.STACK_VIEWPORT_NEW_STACK, evt => {
-      const { element } = evt.detail;
-      utilities.stackPrefetch.disable(element);
-    });
+    // TODO - consider removing the callback when all elements are gone
+    // eventTarget.removeEventListener(
+    //   EVENTS.STACK_VIEWPORT_NEW_STACK,
+    //   newStackCallback
+    // );
   }
 
   eventTarget.addEventListener(
--- a/extensions/cornerstone/src/services/SyncGroupService/SyncGroupService.ts
+++ b/extensions/cornerstone/src/services/SyncGroupService/SyncGroupService.ts
@@ -1,4 +1,8 @@
-import { synchronizers, SynchronizerManager } from '@cornerstonejs/tools';
+import {
+  synchronizers,
+  SynchronizerManager,
+  Synchronizer,
+} from '@cornerstonejs/tools';
 
 import { pubSubServiceInterface } from '@ohif/core';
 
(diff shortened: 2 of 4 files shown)


extensions/cornerstone/src/services/SyncGroupService/SyncGroupService.ts
```diff
@@ -6,20 +6,36 @@ const EVENTS = {
   TOOL_GROUP_CREATED: 'event::cornerstone::syncgroupservice:toolgroupcreated',
 };
 
+export type SyncCreator = (
```
wayfarer3130, line 9: This is the only file with relevant changes - the rest I was working on as I was figuring out how to do this.  Note that this does depend on a few changes in cornerstone 3d to allow synchronization between things other than source/target viewports, but that change is quite small, just in the event delivery mechanism.

extensions/cornerstone/src/init.js
```diff
@@ -267,10 +272,10 @@ export default async function init({
       contextMenuHandleClick
     );
 
-    eventTarget.addEventListener(EVENTS.STACK_VIEWPORT_NEW_STACK, evt => {
-      const { element } = evt.detail;
-      utilities.stackPrefetch.enable(element);
-    });
+    eventTarget.addEventListener(
```
wayfarer3130, line 275: This is a small performance enhancement that prevents continually adding new callbacks - there is exactly one, and it knows how to call the nested enable events.  It is only related to the rest of the changes in that it made debugging a LOT easier not to see multiple calls being made.

extensions/cornerstone/src/services/SyncGroupService/SyncGroupService.ts
wayfarer3130: Now the viewport grid also has a relevant change - but the main changes are in this file, the other one fixes a related bug in viewport grid.

extensions/cornerstone/src/services/SyncGroupService/SyncGroupService.ts
```diff
@@ -29,48 +45,71 @@ export default class SyncGroupService {
...
+      synchronizer = this._createSynchronizer(type, id, options);
+    }
+    return synchronizer;
+  }
+
   public addViewportToSyncGroup(
     viewportId: string,
     renderingEngineId: string,
-    syncGroups?: SyncGroup[]
+    syncGroups?: (SyncGroup | string)[]
```
wayfarer3130, line 82: This is just a convenience, as most of the time it is sufficient to just label the type.

extensions/cornerstone/src/services/SyncGroupService/SyncGroupService.ts
```diff
@@ -29,48 +45,71 @@ export default class SyncGroupService {
...
-      if (!synchronizer) {
-        synchronizer = this._createSynchronizer(type, id);
-      }
+      const synchronizer = this._getOrCreateSynchronizer(type, id, options);
 
       if (target && source) {
         synchronizer.add({
           viewportId,
           renderingEngineId,
+          options,
```
wayfarer3130, line 99: The options go in the source identifier object - that allows different viewports to apply different offset options.

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -47,12 +47,13 @@ export function ViewportGridProvider({ children, service }) {
         return { ...state, ...{ activeViewportIndex: action.payload } };
       }
       case 'SET_DISPLAYSET_FOR_VIEWPORT': {
-        const {
-          viewportIndex,
-          displaySetInstanceUIDs,
-          viewportOptions,
-          displaySetOptions,
-        } = action.payload;
+        const payload = action.payload;
```
wayfarer3130, line 50: The previous code was buggy in that calling setDisplaySetsForViewport would over-ride the viewport options and display set options EVEN if they were supposed to remain the same.  Just moved the default values here so that if values aren't provided, the old values or a default is used, and if the values are provided, then the new values replace the old ones.

wayfarer3130:
This change can be committed independently now from the hanging protocol change, but I would have to extract the HP stuff first, so that this one can go on top of v3-stable

extensions/cornerstone/src/init.js
```diff
@@ -284,10 +289,11 @@ export default async function init({
...
 
-    eventTarget.removeEventListener(EVENTS.STACK_VIEWPORT_NEW_STACK, evt => {
-      const { element } = evt.detail;
-      utilities.stackPrefetch.disable(element);
-    });
+    // TODO - consider removing the callback when all elements are gone
+    // eventTarget.removeEventListener(
+    //   EVENTS.STACK_VIEWPORT_NEW_STACK,
+    //   newStackCallback
+    // );
```
sedghi, line 296: I believe we need this back.

sedghi (commented):
Looks good to me, we can merge on this base branch and later do another review for `feat/cs3d-hp` into `v3-stable`

extensions/cornerstone/src/init.js
wayfarer3130: It was a no-op previously because the removeEventListener got called with a new object, so it was never in the list, and never would remove anything.  So, it isn't any worse than the previous behaviour, although I agree that eventually figuring out if no viewports are displayed to remove the listener is a good idea.  Or maybe not - just leaving it indefinitely and ensuring that it doesn't do anything when the viewport is another type also works.

sedghi (approved):
Looks great, awesome

PR #2866, opened 2022-08-09
feat: new Total Metabolic Tumor Volume tmtv

sedghi:
### PR Checklist


extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -195,9 +195,11 @@ const OHIFCornerstoneViewport = React.memo(props => {
             DisplaySetService
           );
 
+          const keepCamera = true;
```
wayfarer3130, line 198: This should be done/handled by one of the temporal synchronizations - ideally making the regular synchronizer also be able to apply the temporal version of the same synchronization - otherwise the keepCamera setting is wrong for many viewports and will distort the view.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -400,48 +400,55 @@ class CornerstoneViewportService implements IViewportService {
       volume.load();
     });
 
-    this.setVolumesForViewport(viewport, volumeInputArray);
+    return this.setVolumesForViewport(viewport, volumeInputArray);
```
wayfarer3130, line 403: Maybe a comment:
// This returns the async continuation only

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -400,48 +400,55 @@ class CornerstoneViewportService implements IViewportService {
...
+      csToolsUtils.jumpToSlice(viewport.element, {
+        imageIndex,
+      });
+    }
+
+    viewport.render();
   }
 
-  public updateViewport(viewportIndex, viewportData) {
+  public updateViewport(viewportIndex, viewportData, keepCamera = false) {
```
wayfarer3130, line 437: I really dislike introducing API like this here when the API really isn't general, but should be defined by the hanging protocol and/or external synchronizers - the problem is it adds it as API, which is then hard to change.
For now, could you comment:
@params keepCamera - Temporary API to preserve the camera data.  To be replaced by declared synchronizers/event listeners or other mechanisms.
so that people don't assume this will stay permanently.

extensions/cornerstone/src/services/ViewportService/Viewport.ts
```diff
@@ -19,7 +19,7 @@ export type ViewportOptions = {
   initialView?: string;
   syncGroups?: SyncGroup[];
   initialImageOptions?: InitialImageOptions;
-  customViewportOptions?: Record<string, unknown>;
+  customViewportProps?: Record<string, unknown>;
```
wayfarer3130, line 22: Is it a good idea to change the API surface here?  Probably this isn't used heavily enough yet to matter.

extensions/tmtv/src/Panels/PanelPetSUV.tsx
```diff
@@ -140,7 +164,7 @@ export default function PanelPetSUV({ servicesManager, commandsManager }) {
               }}
             />
             <Input
-              label={t('Patient Weight')}
+              label={t('Patient Weight (KG)')}
```
wayfarer3130, line 167: Technically kg is the correct unit

extensions/tmtv/src/Panels/PanelPetSUV.tsx
```diff
@@ -171,7 +195,7 @@ export default function PanelPetSUV({ servicesManager, commandsManager }) {
               }}
             />
             <Input
-              label={t('Half Life')}
+              label={t('Half Life (Seconds)')}
```
wayfarer3130, line 198: If you are using the SI units elsewhere, then (s) would be the right value here.

modes/tmtv/src/index.js
```diff
@@ -58,7 +58,7 @@ function modeFactory({ modeConfiguration }) {
       // Init Default and SR ToolGroups
       initToolGroups(toolNames, Enums, ToolGroupService, commandsManager);
 
-      const activateWindowLevel = () => {
+      const setWindowLevelActive = () => {
```
wayfarer3130, line 61: If you are going to make this change, you should also change the CLI templates to match.  However, I have a configurable mode that mostly abstracts this type of event into "this.setDefaultToolActive()", so if you are going to change it, can you hide the details of WHICH tool gets active.

platform/viewer/.webpack/writePluginImportsFile.js
```diff
@@ -36,56 +36,80 @@ function constructLines(input, categoryName) {
...
-      to: DIST_DIR,
-      toType: 'dir',
-    } : undefined;
-  }
-  ).filter(x => !!x);
-}
+const createCopyPluginPublicToDist = (SRC_DIR, DIST_DIR, plugins) => {
+  return plugins
+    .map((plugin) => {
+      const from = `${SRC_DIR}/../../../node_modules/${plugin.packageName}/public/`;
```
wayfarer3130, line 75: The SRC_DIR/../node_modules is used for link, while the SRC_DIR/../../../node_modules is used for package modules.  Ugly that it is different, but yarn is buggy.

platform/viewer/.webpack/writePluginImportsFile.js
```diff
@@ -36,56 +36,80 @@ function constructLines(input, categoryName) {
...
+          }
+        : undefined;
+    })
+    .filter((x) => !!x);
+};
+
+const createCopyPluginDistToDist = (SRC_DIR, DIST_DIR, plugins) => {
+  return plugins
+    .map((plugin) => {
+      const from = `${SRC_DIR}/../../../node_modules/${plugin.packageName}/dist/`;
```
wayfarer3130, line 91: This produces a nasty bug, where two extensions or modes exporting a build that both has sub-index files will over-write the dist directory with the same files.

wayfarer3130: The resolution of the bug is that the default cli mode creates output files in <pkg-name>/dist/umd/<pkg-name>/ so that deploying the dist directory just works.

platform/viewer/.webpack/writePluginImportsFile.js
```diff
@@ -58,34 +60,87 @@ function getFormattedWindowBlock(addToWindowLines) {
...
+            from,
+            to: DIST_DIR,
+            toType: 'dir',
+          }
+        : undefined;
+    })
+    .filter(x => !!x);
+};
+
+const createCopyPluginChunksToDistForBuild = (SRC_DIR, DIST_DIR, plugins) => {
```
wayfarer3130, line 115: This entire method and the calls to it are not required.  On windows, the way it ends up working causes the from file to not contain directory separators, which is mostly just a no-op, but occasionally it causes things to map to the wrong dist file, which means the dist value isn't available.  I suspect it works consistently in Linux.

PR #2867, opened 2022-08-10
Feat/v3 dynamic load

wayfarer3130:
Loads plugins dynamically from a directory using the name of the plugin rather than the module definition itself, so that configurations can list things like:
extensions: ["my-custom-extension"]
and then have it load from any of:
viewer package.json definition
viewer yarn link definition
umd file from a module, added to the distribution in the "umd/my-custom-extension/index.umd.js" file.  These can be auto-populated if the pluginConfig.js file references a directory containing the public/ distribution for the extension.

Example extensions that load include ecg-dicom and microscopy-dicom


platform/viewer/public/config/local_static.js
```diff
@@ -1,8 +1,14 @@
 window.config = {
   routerBasename: '/',
-  // whiteLabelling: {},
-  extensions: [],
-  modes: [],
+  // Don't list the extensions specifically - let the config mode define them
+  extensions: ['@ohif/extension-tmtv', '@radicalimaging/ecg-dicom', 'unknown'],
```
wayfarer3130, line 4: @swederik and @sedghi  - here is an example of loading the tmtv extension and the ecg dicom extensions.
The tmtv extension comes from the package.json linkage, while the ecg-dicom one comes from the Radical Imaging provided extension.  The viewer will load fine without either of them.

platform/viewer/src/umdLoader.js
```diff
@@ -0,0 +1,128 @@
+/**
```
wayfarer3130, line 1: This is a bunch of stuff to load umd modules - however, if no umd modules are declared, then the basics are essentially just import("@ohif/..."), almost the same as the v3-stable code, so this is almost unchanged to v3-stable.

platform/viewer/pluginConfig.js
```diff
@@ -0,0 +1,112 @@
+/**
```
wayfarer3130, line 1: Renamed to js to allow comments and easier syntax.
There are lots of changes in the file compared to the json file, but almost none of it is relevant - you can just return the json contents and it all works.  YOu can also just delete this file and use an existing json file and it also still all works.

platform/viewer/pluginConfig.js
```diff
@@ -0,0 +1,112 @@
...
+   * Packages needed to be known about, but not extensions/modes
+   */
+  packages: [
+    '@ohif/core',
+    '@ohif/ui',
+    '@ohif/extension-tmtv',
+    '@ohif/mode-tmtv',
+    'react',
+
+    // All the remaining packages are optional, included only if the
```
wayfarer3130, line 81: The definitions are all optional, but useful to provide as an example of how to include things.

PR #2868, opened 2022-08-12
fix(cli):Mode create wasn't quite working correctly.

wayfarer3130:
The cli wasn't creating modes or extensions that were quite buildable.


sedghi:
I don't think we need this any more

PR #2878, opened 2022-08-18
fix thumbnails too small in case of long list

salimkanoun:
Add a minimum height for thumbnail list, as the user can select an image in case of a long list of studies (adding a minimum height to interact)
 
discussed at #2798  (made a cleaner PR with the current v3 stable branch). (#2798 can be closed)

Probably also @sedghi for review


salimkanoun:
alerdy solved

Issue #2880, opened 2022-08-19
Unexpected token: punc ({) in BaseTool.js while building v3-stable


nigelheap:
### Describe the Bug

Build error in `@ohif/viewer` when running `yarn build` 

### What steps can we follow to reproduce the bug?

1. Clone the repo
2. `git checkout v3-stable` 
3. `yarn build`

npm: 8.15.0
node: v16.17.0
yarn: 1.22.18

The error seems to be coming from `@cornerstonejs/tools` package the issues that was recently updated with this merge 
https://github.com/OHIF/Viewers/commit/10b4ed5f8c9e884bef9aa8e456aef657725d0822

`node_modules/@cornerstonejs/tools/dist/esm/tools/base/BaseTool.js`
```js
19:   static { this.toolName = 'BaseTool'; }
```

```
@ohif/viewer: ERROR in app.bundle.8177b3b8d96a4391894e.js
@ohif/viewer: app.bundle.8177b3b8d96a4391894e.js from Terser plugin
@ohif/viewer: Unexpected token: punc ({) [webpack://../node_modules/@cornerstonejs/tools/dist/esm/tools/base/BaseTool.js:19,0][app.bundle.8177b3b8d96a4391894e.js:201443,11]
@ohif/viewer:     at js_error (Viewers/node_modules/terser/dist/bundle.min.js:546:11)
@ohif/viewer:     at croak (Viewers/node_modules/terser/dist/bundle.min.js:1270:9)
@ohif/viewer:     at token_error (Viewers/node_modules/terser/dist/bundle.min.js:1278:9)
@ohif/viewer:     at unexpected (Viewers/node_modules/terser/dist/bundle.min.js:1284:9)
@ohif/viewer:     at as_property_name (Viewers/node_modules/terser/dist/bundle.min.js:3074:20)
@ohif/viewer:     at concise_method_or_getset (Viewers/node_modules/terser/dist/bundle.min.js:2727:20)
@ohif/viewer:     at class_ (Viewers/node_modules/terser/dist/bundle.min.js:2683:22)
@ohif/viewer:     at statement (Viewers/node_modules/terser/dist/bundle.min.js:1458:24)
@ohif/viewer:     at _embed_tokens_wrapper (Viewers/node_modules/terser/dist/bundle.min.js:1335:26)
@ohif/viewer:     at block_ (Viewers/node_modules/terser/dist/bundle.min.js:2170:20)
[...]

PR #2881, opened 2022-08-19
Bump cornerstrone tools and core version to fix terser bug v3-stable

nigelheap:
Bump cornerstone tools and core version to fix terser build error
Related issue:  https://github.com/OHIF/Viewers/issues/2880

@sedghi @wayfarer3130


sedghi:
Thanks for your PR

PR #2886, opened 2022-08-22
Updates/rayhan

Rayhan-011:
### PR Checklist

### DIFF
--- a/extensions/cornerstone/src/ConnectedCornerstoneViewport.js
+++ b/extensions/cornerstone/src/ConnectedCornerstoneViewport.js
@@ -38,15 +38,15 @@ const mapStateToProps = (state, ownProps) => {
   // CINE
   let isPlaying =
     state.viewports.viewportSpecificData[0]['Modality'] === 'US' ? true : false;
-  let frameRate = 24;
+  let frameRate = 28;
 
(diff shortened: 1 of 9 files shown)


Rayhan-011:
1) The viewer needs the ability to load ultrasound cine and auto-play (without the user needing to press the play button). The framerate should be set automatically based on the stored frame time in the DICOM tag (0018,1063). See: https://github.com/OHIF/Viewers/blob/master/platform/ui/src/viewer/PlayClipButton.js

2) Also the animation should be smooth when frame times are short. In the existing code, the animation is not smooth when frame times are short (greater than 30fps). The code should instead set a reasonable framerate (like less than 60fps) and then when animation frames are needed, calculate the proper frame based on the actual elapsed time since the last frame. This would mean that the animation would play in real time even if the computer cannot process the frames fast enough. See: https://github.com/cornerstonejs/cornerstoneTools/blob/9ea015d9d8dbb2d1be43ca9c1f7d6bf61f153910/src/stackTools/playClip.js#L259

3) Reference lines should be displayed on the image. This feature is currently not available in the open source OHIF code but is available in Cornerstone. https://tools.cornerstonejs.org/examples/tools/reference-lines.html

sedghi:

PR #2894, opened 2022-08-25
feat: Allow configurable context menus

wayfarer3130:
This PR adds the ability to customize the measurements context menu in a mode, as well as moving the measurements change operations out of GUI action handlers and turned them into first class commands so that context menus can drive any commandManager available command.  This allows showing other types of context menus if desired, or entirely different behaviour based on either keyboard shortcuts or configured mouse clicks.


wayfarer3130:
@sedghi  - this is an example of using the toolbar service to add other types of "toolbars" - in this case, right click context menus.  We can use this as a straw man next week for our discussion.  I'm also going to try to get a second example of overlay configuration.

platform/ui/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -11,7 +12,7 @@ const ContextMenu = ({ items }) => {
       {items.map((item, index) => (
```
md-prog, line 12: @wayfarer3130 This line produces exception when `items` is undefined.
I would suggest to change `items.map....` to `(items || []).map...`

wayfarer3130: Added a specific check for undefined.

sedghi (commented):
Try to submit comments

extensions/cornerstone/src/initContextMenu.ts
```diff
@@ -0,0 +1,132 @@
...
+      'showContextMenu',
+      showContextMenuDefault
+    );
+
+  const showContextMenu = (contextMenuProps): void => {
+    customizationService.recordInteraction(
+      getShowContextMenu(),
+      contextMenuProps
+    );
+  };
```
sedghi, line 57: This requires explanation, I don't fully understand what you are doing here. So showContextMenu uses the customizationService to record interaction which uses the customization service to show context menu?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
+    if (options.nearbyToolData) {
+      defaultPointsPosition = commandsManager.runCommand(
+        'getToolDataActiveCanvasPoints',
+        { toolData: options.nearbyToolData }
+      );
+    }
+
+    contextMenuController.showContextMenu(
+      options,
+      viewerElement,
```
sedghi, line 110: what is viewerElement, is it enabledElement in cs3d language? there is no viewer element. You can also say viewportElement

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
+      defaultPointsPosition
+    );
+  }
+
   const actions = {
+    /** Show the specified context menu */
+    showViewerContextMenu: providedOptions => {
```
sedghi, line 117: also I prefer options, and later when you destruct `optionToUse` we use this naming pattern a lot through out the codebase both here, and in cs3d. vtk also uses it which is nice

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
+      const viewerElement = _getActiveEnabledElement();
+
+      const options = { ...providedOptions };
+      const { event: evt } = options;
+      const { useSelectedAnnotation, nearbyToolData, menuName } = options;
```
sedghi, line 122: merge

extensions/cornerstone/src/initContextMenu.ts
```diff
@@ -0,0 +1,132 @@
+import { eventTarget, Types, EVENTS } from '@cornerstonejs/core';
+import { Enums } from '@cornerstonejs/tools';
+import { setEnabledElement } from './state';
+
+const cs3DToolsEvents = Enums.Events;
+
+const showContextMenuDefault: Types.CommandCustomization = {
+  commands: [
+    {
+      commandName: 'showViewerContextMenu',
```
sedghi, line 10: There is `showContextMenu` and `showViewerContextMenu` what is the deal with these? Is there any non-viewer context menu?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
+
+      if (menuName) {
+        Object.assign(
+          options,
+          customizationService.getModeCustomization(
+            menuName,
+            defaultContextMenu
+          )
```
sedghi, line 130: I hate this pattern that you should pass the default inside. Compared to

```
customizationService.getModeCustomization(menuName) || defaultContextMenu
```

can we fix this?

extensions/cornerstone/src/commandsModule.ts
sedghi: what is `useSelectedAnnotation` I only see two of them in the whole repo.

extensions/cornerstone/src/initContextMenu.ts
```diff
@@ -0,0 +1,132 @@
...
+      contextMenuProps
+    );
+  };
+
+  const onRightClick = event => {
+    const nearbyToolData = findNearbyToolData(event);
+    showContextMenu({
+      event,
+      nearbyToolData,
+    });
```
sedghi, line 64: I honesetly think having this three functions refactored has added to complexity of reading it. It all can be under onRightClick

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
+        );
+      }
+
+      if (useSelectedAnnotation && !nearbyToolData) {
+        const firstAnnotationSelected = getFirstAnnotationSelected(
```
sedghi, line 135: can't find the import for getFirstAnnotationSelected

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
+          viewerElement
+        );
+        // filter by allowed selected tools from config property (if there is any)
+        if (
+          !options.allowedSelectedTools ||
```
sedghi, line 140: what is `allowedSelectedTools` ? similar to `useSelectedAnnotation` it is only used here. 
I guess these are happening since you are using them in your code base and just adding the bits in OHIF to make your code work.
If you think it is useful, then we should have a full documentation above showViewerContextMenu explainint all the options with their usecases

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
   }
 
+  function _showViewerContextMenu(viewerElement, options) {
+    let defaultPointsPosition = [];
+    if (options.nearbyToolData) {
+      defaultPointsPosition = commandsManager.runCommand(
+        'getToolDataActiveCanvasPoints',
+        { toolData: options.nearbyToolData }
+      );
+    }
```
sedghi, line 106: can't you just have all this inside the `showViewerContextMenu`, it is refactored, but again this refactoring doesn't do anything useful other than making the reader jump lines to find what the function is doing

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
+      );
+    },
+
+    // Measurement tool commands:
+    deleteMeasurement: ({ uid }) => {
+      if (uid) {
+        measurementServiceSource.remove(uid);
+      }
+    },
+    setLabel: ({ uid }) => {
```
sedghi, line 180: I know this was part of the right click and you put it here (thanks), can we name it `setMeasurementLabel` instead?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
+            updatedMeasurement.uid,
+            updatedMeasurement,
+            true
+          );
+        },
+        false
+      );
+    },
+
+    updateMeasurement: props => {
```
sedghi, line 205: This function needs to have a documentatoin to explain what is code and measurementKey

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,7 +96,129 @@ const commandsModule = ({ servicesManager }) => {
...
+      const { code, uid, measurementKey = 'finding' } = props;
+      const measurement = measurementService.getMeasurement(uid);
+      const updatedMeasurement = {
+        ...measurement,
+        [measurementKey]: code,
+        label: code.text,
```
sedghi, line 211: what happens if code is undefined? error at best case scenario, otherwise we loos the label

platform/core/src/extensions/ExtensionManager.ts
```diff
@@ -340,6 +340,7 @@ export default class ExtensionManager {
 
       return extensionModule;
     } catch (ex) {
+      console.log('Exception:', ex);
```
sedghi, line 343: Need more context, something along `Extension Manager::get:: ....` . So why we do need this since the error below has enogh information, maybe just add the exception there

platform/ui/src/types/index.ts
```diff
@@ -15,3 +14,9 @@ const StringNumber = PropTypes.oneOfType([PropTypes.string, PropTypes.number]);
 const StringArray = PropTypes.oneOfType([PropTypes.string, PropTypes.array]);
 
 export { StringNumber, StringArray, ThumbnailType };
+
+export type IProps = Record<string, unknown>;
+
+export type Predicate = (props: IProps) => boolean;
+
+export type Object = Record<string, unknown>;
```
sedghi, line 22: I don't think these add any useful type saftey other than just falsly giving the impression of being type safe. What is wrong with `Record<string, unknown>` ?

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
```diff
@@ -0,0 +1,235 @@
...
+      return;
+    }
+
+    const {
+      event,
+      subMenu,
+      menuId,
+      menus,
+      refs,
+      checkProps,
```
sedghi, line 49: checkProps is a weird name, what does it even mean/do?

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    return {
+      x: undefined,
+      y: undefined,
+    };
+  };
+
+  static _getCanvasPointsPosition = (points = [], viewerElementOfReference) => {
+    const viewerPos = ContextMenuController._getViewerElementDefaultPosition(
```
sedghi, line 176: there is no viewerElement

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
```diff
@@ -0,0 +1,235 @@
...
+      viewerElementOfReference
```
sedghi, line 177: viewerElementOfReference => element?

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
```diff
@@ -0,0 +1,235 @@
...
+  };
+
+  static _getEventDefaultPosition = eventDetail => ({
+    x: eventDetail && eventDetail.currentPoints.client[0],
+    y: eventDetail && eventDetail.currentPoints.client[1],
+  });
+
+  static _getViewerElementDefaultPosition = viewerElement => {
+    if (viewerElement) {
+      const boundingClientRect = viewerElement.getBoundingClientRect();
```
sedghi, line 162: so the element default position is its bottom right?

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
```diff
@@ -0,0 +1,235 @@
...
+      defaultPosition: ContextMenuController._getDefaultPosition(
+        defaultPointsPosition,
+        event?.detail,
+        activeViewerElement
+      ),
+      event,
+      content: ContextMenu,
+
+      onClickOutside: () =>
+        this.services.uiDialogService.dismiss({ id: 'context-menu' }),
```
sedghi, line 75: this.closeViewerContextMenu?

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
sedghi: all these require documentation, ideally above the function

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+      contentProps: {
+        items,
+        checkProps,
+        menus,
+        event,
+        subMenu,
+        eventData: event?.detail,
+        refs,
+        onRunCommands: item => {
```
sedghi, line 85: so `onRunCommands` runs each command? maybe we name it `run`?

sedghi: or `onRun`

platform/ui/src/components/ContextMenu/ContextMenuItemsBuilder.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  console.log('Menu chosen', menu?.id || 'NONE');
+
+  return menu;
+}
+
+/**
+ * Returns the menu from a list of possible menus, based on the actual state of component props and tool data nearby.
+ *
+ * @param checkProps
```
sedghi, line 88: checkprops need documentation

platform/ui/src/components/ContextMenu/ContextMenuItemsBuilder.ts
```diff
@@ -0,0 +1,195 @@
+import { Types } from '@ohif/ui';
```
sedghi, line 1: I don't understand a line of this file. Maybe since there is no example/doc on how menus can be created/added etc. 

what are `delegating, checkFunction, subMenu`?
what is `history` ?

Please add a proper documentation either as markdown or top of here to give exampel use cases. This is too much even for me

sedghi (changes requested):
Great stuff, This PR is so huge and yet so little in documentation. 
There is a great need for a separate markdown explaining different concepts such as `menus`, `checkProps` etc. etc. etc. etc.

Please consider adding such document.

extensions/cornerstone/src/initContextMenu.ts
wayfarer3130: Added explanation.

extensions/cornerstone/src/initContextMenu.ts
wayfarer3130: Cleaned that up - yes, there are differences between the menu commands.  This one is cornersone viewport specific, and there might be others which apply other viewport types or non-viewpor related context menus.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: No, the reason is that the default may need expansion inside to reference other registered GUI elements base on the uiType field.

wayfarer3130: But, for the cases that you CAN do that, we could change the pattern.  This case happens to be one of them.

extensions/cornerstone/src/initContextMenu.ts
wayfarer3130: It is more complex, but it allows defining and showing other types of context menus, and also allows other types of actions to be applied based on the incoming event.   For example, suppose I wanted right click to show the context menu, but control right click to bring up the labels dialog.  Binding that to a function like this allows defining a new command that can distinguish those two operations and run the right sub-commands to apply that.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: No, the point is that showViewerContextMenu sets up the conditions to show a generic context menu, while the showContextMenu actually displays a context menu.  As soon as there are other types of menus with different check conditions, they will need their own entry points.  Eventually the showContextMenu probably needs to move into @ohif/ui as it is not cornerstone specific.
showViewerContextMenu only works for CS3D viewports.

platform/core/src/extensions/ExtensionManager.ts
wayfarer3130: Actually removed context - this is a developer level thing just to ensure the stack trace is preserved when thigns go wrong, so the exception message alone is sufficient.

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
wayfarer3130: The actionType is RunCommands, so the name for the function has to be onRunCommands.
I can change to onRun and make the actionType 'Run' if you want?  I think RunCommands is more descriptive here, but up to you.

platform/ui/src/components/ContextMenu/ContextMenuItemsBuilder.ts
wayfarer3130: Please see the example file findingsContextMenu, as referenced in the documentation.

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
```diff
@@ -0,0 +1,213 @@
...
+ * to manage context menus based on the UI Customization Service.
+ * There are a few parts to this:
+ *    1. Basic controls to manage displaying and hiding context menus
+ *    2. Menu selection services, which use the UI customization service
+ *       to choose which menu to display
+ *    3. Menu item adapter services to convert menu items into displayable and actionable items.
+ *
+ * The format for a menu is defined in the exported type MenuItem
+ */
+export default class ContextMenuController {
```
sedghi, line 17: This file has a load of logic based on various `services`, I belive we should keep separating the ui from the business logic as much as we can. The `platform/ui` should only be react and react based ui components, the reset of logic should be injected (wrapped, etc.) somewhere else

platform/docs/docs/platform/services/ui/customization-service.md
```diff
@@ -275,6 +275,50 @@ const overlayItem: Types.UIOverlayItem = {
 
 This section can be used to specify various customization capabilities.
 
+## Context Menus
+The context menu customization is a relatively complex one, with two layers.
+The first layer is the choice of the command to run to display the context menu.
+For the case of the cornerstone viewer context menu, this is specified by a
```
sedghi, line 281: what is `the cornerstone viewer context menu`? maybe just viewer context menu?

platform/docs/docs/platform/services/ui/customization-service.md
```diff
@@ -275,6 +275,50 @@ const overlayItem: Types.UIOverlayItem = {
...
+customization called `showViewerContextMenu`.  Then, the actual menu is chosen
+by the menuName as specified in the showViewerContextMenu, OR by the actual
+items list in thew viewer context menu.
```
sedghi, line 284: This is confusing to me, what do we mean this is `specified` by a customization called `showViewerContextMenu`? I don't understand this part

platform/docs/docs/platform/services/ui/customization-service.md
```diff
@@ -275,6 +275,50 @@ const overlayItem: Types.UIOverlayItem = {
...
+shows the configuration for that:
+
+```js
+   customizationService.addCustomizations([{
+    id: 'showContextMenu',
+    commands: [
+      {
+        commandName: 'showCircularContextMenu',
+        commandOptions: {
+          menuName: 'cornerstoneContextMenu',
```
sedghi, line 300: I'm not sure If I like `cornerstoneContextMenu` or just `contextMenu`, including cornerstone in any name should be releated to rendering side of things (or tools), the context menu stuff is purely an app concern and having the library name in the names and ids probably confuses users

platform/docs/docs/platform/services/ui/customization-service.md
```diff
@@ -478,6 +522,21 @@ window.config = {
 
 <img src="../../../assets/img/customizable-overlay.png" />
 
+##  Customizable Context Menu
+
+The default cornerstone context menu can be customized by setting the
```
sedghi, line 527: again cornerstone context menu here

extensions/default/src/getCustomizationModule.tsx
```diff
@@ -44,69 +45,24 @@ export default function getCustomizationModule() {
...
-         *  },
-         *
-         * @see CustomizableViewportOverlay
-         */
         {
-          id: 'ohif.overlayItem',
-          uiType: 'uiType',
-          content: function (props) {
-            if (this.condition && !this.condition(props)) return null;
+          id: 'ohif.contextMenu',
```
sedghi, line 49: why are you removing the overlayItem customization? Michel recently added that

platform/core/src/extensions/ExtensionManager.ts
```diff
@@ -36,7 +36,9 @@ export interface ExtensionParams extends ExtensionConstructor {
  */
 export interface Extension {
   id: string;
-  preRegistration?: (p: ExtensionParams) => void;
+  preRegistration?: (p: ExtensionParams) => Promise<void> | void;
+  onModeExit?: () => void;
+  getHangingProtocolModule?: (p: ExtensionParams) => unknown;
```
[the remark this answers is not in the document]
wayfarer3130: It isn't being treated specially, I'm just starting to type all of them.  Yes, there are several getHangingProtocolModule's defined.
Add getCommandsModule as another example - this is so far just to type the parameters - I haven't gotten the response types defined yet.

extensions/default/src/getCustomizationModule.tsx
```diff
@@ -44,69 +45,24 @@ export default function getCustomizationModule() {
...
-            );
+            for (const menu of menus) {
+              const { items } = menu;
+              menu.items = [];
+              for (const item of items) {
+                menu.items.push(customizationService.applyType(item));
+              }
+            }
+            return ret;
           },
```
sedghi, line 66: I prefer this for readability

```js
applyType: function (customizationService: CustomizationService) {
  const clonedObject = { ...this };
  clonedObject.menus = this.menus.map(menu => ({ ...menu }));

  for (const menu of clonedObject.menus) {
    const originalItems = menu.items;
    menu.items = []; 

    for (const item of originalItems) {
      menu.items.push(customizationService.applyType(item));
    }
  }
  
  return clonedObject;
},
```

extensions/cornerstone/src/init.tsx
```diff
@@ -155,118 +151,12 @@ export default async function init({
   initWADOImageLoader(userAuthenticationService, appConfig);
 
   /* Measurement Service */
-  const measurementServiceSource = connectToolsToMeasurementService(
+  this.measurementServiceSource = connectToolsToMeasurementService(
```
sedghi, line 154: what is `this.`?

extensions/test-extension/src/custom-context-menu/codingValues.ts
```diff
@@ -0,0 +1,62 @@
+const codingValues = {
```
sedghi, line 1: some doc above this to explain what are these codes and where to read more?

extensions/test-extension/src/custom-context-menu/contextMenuCodeItem.ts
```diff
@@ -0,0 +1,27 @@
+import codingValues from './codingValues';
+
+const codeMenuItem = {
+  id: '@ohif/contextMenuAnnotationCode',
+  codingValues,
+
+  /** Applies the code value setup for this item */
+  applyType: function (customizationService) {
```
sedghi, line 8: can we come up with a better name for `applyType`? maybe along `transform` (since it is basically transforming)

extensions/test-extension/src/custom-context-menu/findingsContextMenu.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        // The example below shows how to include a delegating sub-menu,
+        // Only available on the @ohif/hp-extension.mn hanging protocol
+        // To demonstrate, select the 3x1 layout from the protocol menu
+        // and right click on a measurement.
+        {
+          label: 'IncludeSubMenu',
+          selector: ({ protocol }) => protocol?.id === '@ohif/hp-extension.mn',
+          delegating: true,
+          subMenu: 'orientationSelectionSubMenu',
```
sedghi, line 48: How is this `subMenu` when actionType is not `SubMenu`?

extensions/test-extension/src/getCustomizationModule.ts
```diff
@@ -0,0 +1,14 @@
...
+  codingValues,
+  contextMenuCodeItem,
+  findingsContextMenu,
+} from './custom-context-menu';
+
+export default function getCustomizationModule() {
+  return [
+    {
+      name: 'custom-context-menu',
+      value: [codingValues, contextMenuCodeItem, findingsContextMenu],
```
sedghi, line 11: how is `codingValues` passed here, it is just an object, the rest (contexxtMenuCodeItem and findingsContextMenu) on the other hand follow the interface for customization

platform/core/src/extensions/ExtensionManager.ts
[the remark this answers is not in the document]
wayfarer3130: It isn't special, I'm just starting to type things, in this case it is the parameters that I'm typing.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,9 +80,209 @@ const commandsModule = ({
...
+        const firstAnnotationSelected = getFirstAnnotationSelected(
+          viewerElement
+        );
+        // filter by allowed selected tools from config property (if there is any)
+        if (
+          !optionsToUse.allowedSelectedTools ||
+          optionsToUse.allowedSelectedTools.includes(
+            firstAnnotationSelected?.metadata?.toolName
+          )
+        ) {
```
sedghi, line 125: I prefer breaking this down

```js
  const isToolAllowed = !optionsToUse.allowedSelectedTools || optionsToUse.allowedSelectedTools.includes(firstSelectedAnnotation?.metadata?.toolName);

if (isToolAllowed){
  //
}
```

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -72,9 +80,209 @@ const commandsModule = ({
...
+        // TODO - remove this once measurement items customization is ready
+        const allCodes = [];
+        if (textLabel) allCodes.push(textLabel);
+        if (updatedMeasurement.finding) {
+          allCodes.push(updatedMeasurement.finding.CodeMeaning);
+        }
+        (updatedMeasurement.findingSites || []).forEach(it =>
+          allCodes.push(it.CodeMeaning)
+        );
+        updatedMeasurement.label = allCodes.join(', ');
```
sedghi, line 274: can you refactor this? since my guess is not everyone using `code`, so it is too much noise to read

platform/ui/src/components/ContextMenu/ContextMenuItemsBuilder.ts
```diff
@@ -0,0 +1,197 @@
...
+ */
+// menus category to be skipped when doing a depth search.
+const menuCategoryBlacklist = ['history'];
+
+/**
+ * Finds menu by menu id
+ *
+ * @returns Menu having the menuId
+ */
+export function findMenuByMenuId(menus, menuId): Menu {
```
sedghi, line 19: `findMenuById`

sedghi (commented):
Great revision, thanks a lot. See my new comments

extensions/test-extension/src/custom-context-menu/findingsContextMenu.ts
```diff
@@ -0,0 +1,100 @@
...
+          label: 'Delete Measurement',
+          commands: [
+            {
+              commandName: 'deleteMeasurement',
+            },
+          ],
+        },
+        {
+          label: 'Add Label',
+          commands: [
```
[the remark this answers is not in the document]
wayfarer3130: Yes, RunCommands is now the default, have removed it everywhere it was set specifically.

extensions/test-extension/src/custom-context-menu/findingsContextMenu.ts
wayfarer3130: Because subMenu is the name of another menu, in this case the delegating subMenu, and this item delegates to the sub-menu, it never gets run by itself.

extensions/test-extension/src/getCustomizationModule.ts
wayfarer3130: All values passed into the customization module are just objects.    This, and every other instance passed in gets used by someone using the customization service, but the only default/handling is that customizationType gets inheritted from, and transform gets called to update the object, but neither of those are used here.

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
wayfarer3130: Moved as requested.

extensions/cornerstone/src/init.tsx
wayfarer3130: It is the extension object - this is a really ugly case where a result of one call in the extension is re-used later on in another spot.  I think you set it up so the init was in the commands module previously, but that is just as ugly because it means you delay init till the commands module is setup.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: That is actually the primary update for code, and belongs in the commands module as that is the main behaviour of this call.  I think this belongs here - there will be a small update later to fix a few things once measurement customization works.

extensions/cornerstone/src/initContextMenu.ts
```diff
@@ -0,0 +1,129 @@
...
+        commandName: 'closeContextMenu',
+      },
+    ],
+  },
+  button3: {
+    commands: [
+      {
+        commandName: 'showViewerContextMenu',
+        commandOptions: {
+          menuName: 'cornerstoneContextMenu',
```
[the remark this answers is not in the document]
wayfarer3130: Agreed, updating.

extensions/default/src/CustomizeableContextMenu/types.ts
```diff
@@ -0,0 +1,123 @@
+import { Types } from '@ohif/core';
+
+/**
+ * SelectorProps are properties used to decide whether to select a manu or
```
jbocce, line 4: typo menu not manu

extensions/default/src/CustomizeableContextMenu/types.ts
```diff
@@ -0,0 +1,123 @@
...
+ */
+export interface SelectorProps {
+  // If the context menu is invoked in the context of a measurement, then it
+  // will contain the nearby tool data.
+  nearbyToolData?: Record<string, unknown>;
+
+  // The tool name for the nearby tool
+  toolName?: string;
+
+  // An annotation UID - this will be present if nearyToolData is present.
```
jbocce, line 20: type\o *nearbyToolData

platform/docs/docs/platform/services/ui/customization-service.md
```diff
@@ -478,6 +477,29 @@ window.config = {
...
+`cornerstoneContextMenu`.  For a full example, see `findingsContextMenu`.
+
+## Customizeable Cornerstone Viewport Click Behaviour
+
+The behaviour on clicking on the cornerstone viewport can be customized
+by setting the `cornerstoneViewportClickCommands`.  This is intended to
+support both the cornerstone 3D internal commands as well as things like
+context menus.  Currently it supports buttons 1-3, as well as modifier keys
+by associated a commands list with the button to click.  See `initContextMenu`
```
jbocce, line 499: *associating

platform/core/src/extensions/ExtensionManager.ts
```diff
@@ -331,7 +340,7 @@ export default class ExtensionManager {
     }
 
     try {
-      const extensionModule = getModuleFn({
+      const extensionModule = extension[getModuleFnName]({
```
jbocce, line 343: Why was this changed?

extensions/default/src/getCustomizationModule.tsx
```diff
@@ -109,6 +109,26 @@ export default function getCustomizationModule() {
...
         },
+
+        {
+          id: 'ohif.contextMenu',
+
+          /** Applies the customizationType to all the menu items */
+          transform: function (customizationService: CustomizationService) {
+            // Don't modify the children, as those are copied by reference
+            const clonedObject = { ...this };
+            clonedObject.menus = this.menus.map(it => ({ ...it }));
```
jbocce, line 120: Consider replacing `it` with `menu`.

extensions/default/src/getCustomizationModule.tsx
```diff
@@ -109,6 +109,26 @@ export default function getCustomizationModule() {
             );
           },
         },
+
+        {
+          id: 'ohif.contextMenu',
+
+          /** Applies the customizationType to all the menu items */
+          transform: function (customizationService: CustomizationService) {
+            // Don't modify the children, as those are copied by reference
```
jbocce, line 118: Does it makes sense to give an explicit reason as to why clones are being returned here? Pardon me if it should be obvious, but... 😊

extensions/default/src/commandsModule.ts
```diff
@@ -34,7 +42,60 @@ const commandsModule = ({ servicesManager, commandsManager }) => {
...
   const actions = {
+    /**
+     * Show the context menu.
+     * @param options.menuId defines the menu name to lookup, from customizationService
+     * @param options.defaultMenu contains the default menu set to use
+     * @param options.element is the element to show the menu within
+     * @param options.event is the event that caused the context menu
+     * @param options.selectorProps is the set of selection properties to use
+     */
+    showContextMenu: options => {
```
jbocce, line 60: Is it possible to `type` options here?

extensions/default/src/CustomizeableContextMenu/types.ts
```diff
@@ -0,0 +1,123 @@
+import { Types } from '@ohif/core';
+
+/**
+ * SelectorProps are properties used to decide whether to select a manu or
+ * menu item for display.
+ * An instance of SelectorProps is provided to the selector functions, which
+ * return true to include the item or false to exclude it.
+ * The point of this is to allow more specific conext menus which hide
+ * non-relevant menu options, optimizing the speed of selection of menus
+ * (See Bill Wallace's masters thesis for selection time versus complexity of user menus).
```
jbocce, line 10: That's awesome. However, as charming and intriguing as this is, consider excluding it OR if you have a link perhaps keep this and include the link.

extensions/default/src/CustomizeableContextMenu/ContextMenuController.tsx
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  static getDefaultPosition = (): Point => {
+    return {
+      x: 0,
+      y: 0,
+    };
+  };
+
+  static _getEventDefaultPosition = eventDetail => ({
```
jbocce, line 132: Shouldn't the various _get*Position methods have a return type of `Point` like getDefaultPosition does?

extensions/default/src/CustomizeableContextMenu/ContextMenuItemsBuilder.test.js
```diff
@@ -0,0 +1,29 @@
...
+    items: [],
+  },
+];
+
+const menuBuilder = new ContextMenuItemsBuilder();
+
+describe('ContextMenuItemsBuilder', () => {
+  test('findMenuDefault', () => {
+    expect(menuBuilder.findMenuDefault(menus, {})).toBe(menus[2]);
+    expect(menuBuilder.findMenuDefault(menus, { value: 'two' })).toBe(menus[1]);
```
jbocce, line 25: I must be missing something here, but shouldn't the second param to `findMenuDefault` have a `selectorProps` property? At least when I look at line 42 of ContextMenuItemsBuilder.ts.

extensions/default/src/CustomizeableContextMenu/ContextMenuItemsBuilder.ts
```diff
@@ -0,0 +1,193 @@
...
+ * See the `test-mode` examples for details.
+ *
+ * @param selectorProps
+ * @param {*} event event that originates the context menu
+ * @param {*} menus List of menus
+ * @param {*} menuIdFilter
+ * @returns
+ */
+export function getMenuItems(
+  selectorProps: Types.IProps,
```
jbocce, line 114: What is IProps?

platform/viewer/cypress/integration/measurement-tracking/OHIFStudyBrowser.spec.js
```diff
@@ -52,7 +52,7 @@ describe('OHIF Study Viewer Page', function() {
 
   it('performs double-click to load thumbnail in active viewport', () => {
     // Have to finish rendering the image before this works
-    cy.wait(250);
+    cy.wait(350);
```
jbocce, line 55: This test is not actually checking anything right?

platform/viewer/cypress/integration/measurement-tracking/OHIFContextMenuCustomization.spec.js
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  it('checks context menu customization', function () {
+    // Add length measurement
+    cy.addLengthMeasurement();
+    cy.get('[data-cy="prompt-begin-tracking-yes"]').click();
+    cy.get('[data-cy="measurement-item"]').click();
+
+    const [x1, y1] = [150, 100];
+    cy.get('@viewport')
```
jbocce, line 20: I suspect we are doing a 'down' and 'up' here to avoid  the race condition (of which I apologize) if a simple click were done instead? Is there not a way to wait for a condition to be true in Cypress?

jbocce (approved):

platform/core/src/extensions/ExtensionManager.ts
wayfarer3130: Because it doesn't associate the "this" property with the extension it is running, so if you need to store some variables
between init and other functions, then without this change, you can't do it.

jbocce: Cool. Thanks. Consider adding a comment maybe?

extensions/default/src/CustomizeableContextMenu/types.ts
wayfarer3130: I was wondering if anyone would comment - have removed it :-)

extensions/default/src/CustomizeableContextMenu/ContextMenuItemsBuilder.test.js
wayfarer3130: Yes, but selector props is really just a grab bag of things used for menu selection.

platform/viewer/cypress/integration/measurement-tracking/OHIFStudyBrowser.spec.js
wayfarer3130: It performs a double click - but the double click has to wait for the viewer to be ready to display something.

platform/viewer/cypress/integration/measurement-tracking/OHIFContextMenuCustomization.spec.js
wayfarer3130: That is right click, and I don't  know how to invoke click on a right click.

PR #2900, opened 2022-08-29
fix: cornerstone cache race condition

wayfarer3130:
In the cornerstone cache service, the viewport data changed was fired before the viewport data was actually fully changed.
Also, in panelPetSUV, the panel was being updated on display set changed, which was incorrect, because it actually depends on the viewport data changed event.  Updated it so that panel also uses the viewport data changed event.

### DIFF
--- a/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
+++ b/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
@@ -232,22 +232,22 @@ const OHIFCornerstoneViewport = React.memo(props => {
     }
 
     const loadViewportData = async () => {
-      const viewportData = await CornerstoneCacheService.getViewportData(
+      await CornerstoneCacheService.getViewportData(
         viewportIndex,
         displaySets,
         viewportOptions.viewportType,
         dataSource,
+        (viewportDataLoaded) => {
+          CornerstoneViewportService.setViewportDisplaySets(
+            viewportIndex,
+            viewportDataLoaded,
+            viewportOptions,
+            displaySetOptions
+          );
+          setViewportData(viewportDataLoaded);
+        },
         initialImageIndex
       );
-
-      CornerstoneViewportService.setViewportDisplaySets(
-        viewportIndex,
-        viewportData,
-        viewportOptions,
-        displaySetOptions
-      );
-
-      setViewportData(viewportData);
     };
 
     loadViewportData();
--- a/extensions/cornerstone/src/Viewport/Overlays/ViewportOrientationMarkers.tsx
+++ b/extensions/cornerstone/src/Viewport/Overlays/ViewportOrientationMarkers.tsx
@@ -137,7 +137,7 @@ function ViewportOrientationMarkers({
       }
 
       const imageIndex = imageSliceData.imageIndex;
-      const imageId = viewportData?.imageIds[imageIndex];
+      const imageId = viewportData.imageIds?.[imageIndex];
 
       // Workaround for below TODO stub
       if (!imageId) {
--- a/extensions/cornerstone/src/services/ViewportService/CornerstoneCacheService.ts
+++ b/extensions/cornerstone/src/services/ViewportService/CornerstoneCacheService.ts
@@ -32,6 +32,8 @@ const EVENTS = {
   VIEWPORT_DATA_CHANGED: 'event::cornerstone::viewportdatachanged',
 };
 
+export type IViewportData = StackData | VolumeData;
+
 class CornerstoneCacheService {
   stackImageIds: Map<string, string[]> = new Map();
   volumeImageIds: Map<string, string[]> = new Map();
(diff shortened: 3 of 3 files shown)


extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -228,12 +228,13 @@ const OHIFCornerstoneViewport = React.memo(props => {
...
         viewportIndex,
-        viewportData,
+        viewportDataLoaded,
         viewportOptions,
         displaySetOptions
       );
 
-      setViewportData(viewportData);
+      setViewportData(viewportDataLoaded);
+      CornerstoneCacheService.setViewportData(viewportDataLoaded, viewportIndex);
```
wayfarer3130, line 237: Explicitly call the cache service viewport data changed event, so that it happens after the actual change has completed.

extensions/cornerstone/src/Viewport/Overlays/ViewportOrientationMarkers.tsx
```diff
@@ -137,7 +137,7 @@ function ViewportOrientationMarkers({
       }
 
       const imageIndex = imageSliceData.imageIndex;
-      const imageId = viewportData?.imageIds[imageIndex];
+      const imageId = viewportData.imageIds?.[imageIndex];
```
wayfarer3130, line 140: The viewportData is already null checked, but the imageIds isn't, so move the check onto imageIds.

extensions/cornerstone/src/index.tsx
```diff
@@ -34,6 +35,13 @@ const OHIFCornerstoneViewport = props => {
   );
 };
 
+const registerCacheService = {
```
wayfarer3130, line 38: The cache is a global singleton, so register it this way.

extensions/tmtv/src/Panels/PanelPetSUV.tsx
```diff
@@ -89,8 +90,8 @@ export default function PanelPetSUV({ servicesManager, commandsManager }) {
 
   // get the patientMetadata from the StudyInstanceUIDs and update the state
   useEffect(() => {
-    const { unsubscribe } = DisplaySetService.subscribe(
-      DisplaySetService.EVENTS.DISPLAY_SETS_ADDED,
+    const { unsubscribe } = cornerstoneCacheService.subscribe(
```
wayfarer3130, line 93: And listen for the changes on the event that is actually fired when the values are changed, rather than an unrelated event that may or may not be fired at the right time.  
Note that this fixes a theoretical issue if the display sets being displayed got updated - something we will probably see in the future.

wayfarer3130:
@sedghi Could you take a look at this PR?  It is a straight fix for a race condition, no functional changes other than actually exposing the required event to listen for correctly (CornerstoneCacheService wasn't actually a service, and now it is).

extensions/cornerstone/src/index.tsx
sedghi: Can you move this to its own file similar to `ToolGroupService` and the rest?

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
sedghi: This doesn't make sense to me, the CacheService is returning the data itself, and we are setting it here again just for the purpose of event firing. I guess this is a fix that you have for the race condition you are seeing but will introduce more bugs. `setViewportData` is an async operation by react for state update. I guess the event was firing before this react state is set (I'm not even sure we need to keep track of the viewportData here at all, but have not had the chance to debug it)

sedghi (changes requested):
See my comments

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
wayfarer3130: Not everyone has access to "viewportData" in the react sense, but still needs to know when the viewport data has been loaded, so that they can rely on viewport specific things occurring.  The problem is, most users of the data use it by accessing the cornerstone3D viewport, which has multiple events which fire, so there isn't a good way of knowing which one to use.  This event now becomes a reliable "viewport is loaded" event.  I'm not totally happy with it yet, because it isn't cross viewport types - eg an ECG viewport should also be able to fire it, but currently can't.

extensions/cornerstone/src/index.tsx
wayfarer3130: The issue is that it is accessed both statically and as a real service.  How about I move the export object into the cache export as an additional export?

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -218,22 +218,22 @@ const OHIFCornerstoneViewport = React.memo(props => {
     }
 
     const loadViewportData = async () => {
-      const viewportData = await CornerstoneCacheService.getViewportData(
+      await CornerstoneCacheService.getViewportData(
         viewportIndex,
         displaySets,
         viewportOptions.viewportType,
         dataSource,
+        (viewportDataLoaded) => {
```
wayfarer3130, line 226: Pass in the necessary use of the viewport data to allow the event to fire after the data is stored/used

extensions/cornerstone/src/services/ViewportService/CornerstoneCacheService.ts
```diff
@@ -224,4 +228,12 @@ class CornerstoneCacheService {
 }
 
 const CacheService = new CornerstoneCacheService();
+
+export const registerCacheService = {
```
wayfarer3130, line 232: This addresses your comment about having the registration information in the Service file itself.

sedghi (approved):

PR #2912, opened 2022-09-05
fix: displaySetService no event not needed

sedghi:
### Fix
- We don't need to suppress events on displaySet service, we can just wait, since after the async metadata the part for instances add -> displaySet creation is syncronous.
- The PR broke the PanelSUV which is fixed now

### Add
- A new event on the HP (`PROTOCOL_CHANGED`), to let subscribers know that the protocol match for viewports is performed

### Refactor
- We had so many matchdetails in the HP service, I refactored all to be more clear and added many types for HP

### Note to discuss cc @swederik 

~~Todo: DisplaySetService makes the displayset when ALL instances are added
However, making a displayset only requires one instance and one instance only
you can look at the sopClassHandlers which you will see only the first instance
is used to make the displayset. This is a huge performance bottleneck
since we are waiting for all instances to be added before making the displayset
Making a displaySet can be broken down into 2 steps: 1) as soon as the first
instance is added, create the displayset and 2) when all instances are
added to the store, update the displayset with the rest of instances.
This improvement will make the first draw of the hung viewport way faster.~~

Update: since we are doing HP on DisplaySet the above comment is not valid anymore

### DIFF
--- a/platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
+++ b/platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
@@ -39,6 +39,12 @@ type DisplaySetMatchDetails = {
   sortingInfo: any;
 };
 
+type HangingProtocolMatchDetails = {
+  displaySetMatchDetails: Map<string, DisplaySetMatchDetails>;
+  viewportMatchDetails: ViewportMatchDetails[];
+  hpAlreadyApplied: boolean[];
+};
+
 class HangingProtocolService {
   studies: StudyMetadata[];
   protocols: Record<string, unknown>[];
@@ -116,11 +122,7 @@ class HangingProtocolService {
     // this.ProtocolEngine.reset()
   }
 
-  public getMatchDetails(): {
-    viewportMatchDetails: ViewportMatchDetails[];
-    displaySetMatchDetails: Map<string, DisplaySetMatchDetails>;
-    hpAlreadyApplied: boolean[];
-  } {
+  public getMatchDetails(): HangingProtocolMatchDetails {
     return {
       viewportMatchDetails: this.viewportMatchDetails,
       displaySetMatchDetails: this.displaySetMatchDetails,
@@ -330,6 +332,7 @@ class HangingProtocolService {
 
     this._broadcastChange(this.EVENTS.PROTOCOL_CHANGED, {
       viewportMatchDetails: this.viewportMatchDetails,
+      displaySetMatchDetails: this.displaySetMatchDetails,
       hpAlreadyApplied: this.hpAlreadyApplied,
     });
   }
@@ -665,6 +668,7 @@ class HangingProtocolService {
     this._broadcastChange(this.EVENTS.STAGE_CHANGE, {
       viewportMatchDetails: this.viewportMatchDetails,
       hpAlreadyApplied: this.hpAlreadyApplied,
+      displaySetMatchDetails: this.displaySetMatchDetails,
     });
     return true;
   }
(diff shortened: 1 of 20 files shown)


platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -96,16 +112,20 @@ class HangingProtocolService {
...
   }
 
-  public getDisplaySetsMatchDetails() {
-    return this.displaySetMatchDetails;
-  }
-
-  public getState() {
-    return [this.matchDetails, this.hpAlreadyApplied];
+  public getMatchDetails(): {
+    viewportMatchDetails: ViewportMatchDetails[];
```
wayfarer3130, line 120: Isn't this the same type one gets for the event, or else very similar?  If so, it should have it's own interface.

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -24,13 +24,20 @@ function defaultRouteInit({ servicesManager, studyInstanceUIDs, dataSource }) {
...
 
   const unsubscriptions = [];
-  // TODO: This should be baked into core, not manual?
-  // DisplaySetService would wire this up?
   const {
     unsubscribe: instanceAddedUnsubscribe,
   } = DicomMetadataStore.subscribe(
     DicomMetadataStore.EVENTS.INSTANCES_ADDED,
-    ({ StudyInstanceUID, SeriesInstanceUID, madeInClient = false }) => {
+    // Todo: DisplaySetService makes the displayset when ALL instances are added
```
wayfarer3130, line 31: I don't think this is the correct description - adding all instances at a series level isn't a huge performance bottleneck unless all instances are being loaded separately.  In fact, having all instances added makes the code quite a bit faster because it allows sharing initialization/setup, and avoids using the event loop to add instances, as would be required for adding a single item first and then adding additional items. 
There is also the vary serious issue that the wrong item can be shown - images can easily come back out of order and if the display ordering is such the user expects ordering from oldest to newest or vice-versa, and gets the opposite ordering initially, then they can end up looking at things while additional items are still loading and getting it wrong.
Finally, if the hanging protocol selects various possibilities of images, and first match gets shown initially, with later ones being still loading, then the wrong series can be shown entirely - that is even worse than wrong image, as it may not be obvious.  I think suggesting anything of the sort as to multiple phases is very premature.

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -42,45 +49,42 @@ function defaultRouteInit({ servicesManager, studyInstanceUIDs, dataSource }) {
...
+  // to display anything if a required match fails), so hold off the matches
   // here until the entire study is ready.
-  DisplaySetService.holdChangeEvents();
-  const allRetrieves = studyInstanceUIDs.map(StudyInstanceUID =>
-    dataSource.retrieve.series.metadata({ StudyInstanceUID })
-  );
+
   Promise.allSettled(allRetrieves).then(() => {
-    DisplaySetService.fireHoldChangeEvents();
+    const displaySets = DisplaySetService.getActiveDisplaySets();
```
wayfarer3130, line 62: You have re-introduced a bug here.  The retrieval of instances does NOT guarantee that display sets are generated yet, so you can randomly get matching or non-matching display sets.  For example, suppose we have
allRetrieves = [series1, series2]
Now, series2 has delivered it's display set in the getActiveDisplaySets(), but series1 has it's display set delivery still on the event queue.  So, displaySets = [series2 display set], which will, even on a 1x1 layout with no other constraints end up showing series 2 instead of series 1 on initial display.  I can show you more complex cases which completely fail to run, including your pt display set.  USUALLY it works on a fast retrieve, because the ordering of the event queue is such that it is usually done in time here, but randomly it will fail.

wayfarer3130 (approved):
Looks good to me now that we have discussed the timing of display set creation and added this note to the documentation.

Issue #2904, opened 2022-08-31
"Something went wrong" error after using scroll bar


JennyLeePMH:
### Describe the Bug
Using the scroll bar causes the "Something went wrong" error.  

### Steps to reproduce

1. Go to https://v3-demo.ohif.org/
2. Open any CT studies using Basic Viewer
3. Select any CT series
4. Use the bidirectional tool to draw anything. 
5. Click Yes to tracking measurements
6. Use the **scroll bar** to move to a different slice.   Don't use the mouse scroll wheel.
7. Repeat step 4-5 several times at different slice.
8. Go back to the study list
9. Select a different study using Basic Viewer
10. Get "Something went wrong"  error
11. See attached .mp4 for demonstration.

### Different steps to reproduce the same error

1. Go to https://v3-demo.ohif.org/
2. Open any CT studies with multiple series using Basic Viewer
3. Create a 4x4 viewports
4. Populate each viewports with series
5. Use the **scroll bar** on each viewport.  Don't use the mouse scroll wheel.
6. Switch back to 1x1 viewport
7. then, switch back to 4x4 viewports
8. Try to populate each viewports with series

https://user-images.githubusercontent.com/4551564/187762859-8446d507-86ef-479e-ab40-ebb7ee28a9f2.mp4

PR #2920, opened 2022-09-08
fix: imagScrollbar should not stop cine if not playing

sedghi:
https://github.com/OHIF/Viewers/issues/2904


sedghi:
https://github.com/OHIF/Viewers/issues/2904

PR #2925, opened 2022-09-13
feat: query string dynamic app config

Ouwen:
This PR allows for the app config to reference a json over network. This allows for
fast swapping of data source configuration, OIDC, etc, without recompiling the application.

If the `configUrl` query string is passed, the worklist and modes will load from the referenced json
rather than the default .env config

If there is no `configUrl` path provided, the default behaviour is used and there should not be any
deviation from current user experience.

Example usage:
`http://localhost:3000/?configUrl=http://localhost:3000/config/example.json`
The following will load the default.js settings, but with 6 web workers instead of 3 (check `window.config`) to verify.

### PR Checklist

### DIFF
--- a/extensions/default/src/ViewerLayout/index.tsx
+++ b/extensions/default/src/ViewerLayout/index.tsx
@@ -46,13 +46,21 @@ function ViewerLayout({
   const onClickReturnButton = () => {
     const { pathname } = location;
     const dataSourceIdx = pathname.indexOf('/', 1);
-    const search =
-      dataSourceIdx === -1
-        ? undefined
-        : `datasources=${pathname.substring(dataSourceIdx + 1)}`;
+    const query = new URLSearchParams(window.location.search);
+    const configUrl = query.get('configUrl');
+
+    const searchQuery = new URLSearchParams();
+    if (dataSourceIdx !== -1) {
+      searchQuery.append('datasources', pathname.substring(dataSourceIdx + 1));
+    }
+
+    if (configUrl) {
+      searchQuery.append('configUrl', configUrl);
+    }
+
     navigate({
       pathname: '/',
-      search,
+      search: decodeURIComponent(searchQuery.toString()),
     });
   };
 
--- a/platform/docs/docs/configuration/configurationFiles.md
+++ b/platform/docs/docs/configuration/configurationFiles.md
@@ -116,7 +116,26 @@ Here are a list of some options available:
 if auth headers are used, a preflight request is required.
 - `maxNumRequests`: The maximum number of requests to allow in parallel. It is an object with keys of `interaction`, `thumbnail`, and `prefetch`. You can specify a specific number for each type.
 - `showLoadingIndicator`: (default to true), if set to false, the loading indicator will not be shown when navigating between studies.
+- `dangerouslyUseDynamicConfig`: Dynamic config allows user to pass `configUrl` query string. This allows to load config without recompiling application. If the `configUrl` query string is passed, the worklist and modes will load from the referenced json rather than the default .env config. If there is no `configUrl` path provided, the default behaviour is used and there should not be any deviation from current user experience. This feature related to PR [OHIF#2925](https://github.com/OHIF/Viewers/pull/2925).<br/>
+Points to consider while using `dangerouslyUseDynamicConfig`:<br/>
+  - User have to enable this feature by setting `dangerouslyUseDynamicConfig.enabled:true`. By default it is `false`.
+  - Regex helps to avoid easy exploit. Dafault is `/.*/`. Setup your own regex to choose a specific source of configuration only.
+  - User must set `cross-origin:same-origin` to avoid potential harder exploit.
 
+  - Example config:
+    ```js
+    dangerouslyUseDynamicConfig: {
+      enabled: false,
+      regex: /.*/
+    }
+    ```
+
+  > Example 1, to allow numbers and letters in an absolute or sub-path only.<br/>
+`regex: /(0-9A-Za-z.]+)(\/[0-9A-Za-z.]+)*/`<br/>
+Example 2, to restricts to either hosptial.com or othersite.com.<br/>
+`regex: /(https:\/\/hospital.com(\/[0-9A-Za-z.]+)*)|(https:\/\/othersite.com(\/[0-9A-Za-z.]+)*)/` <br/>
+Example usage:<br/>
+`http://localhost:3000/?configUrl=http://localhost:3000/config/example.json`<br/>
 
 
 
--- a/platform/docs/docs/configuration/configurationFiles.md
+++ b/platform/docs/docs/configuration/configurationFiles.md
@@ -120,7 +120,7 @@ if auth headers are used, a preflight request is required.
 Points to consider while using `dangerouslyUseDynamicConfig`:<br/>
   - User have to enable this feature by setting `dangerouslyUseDynamicConfig.enabled:true`. By default it is `false`.
   - Regex helps to avoid easy exploit. Dafault is `/.*/`. Setup your own regex to choose a specific source of configuration only.
-  - User must set `cross-origin:same-origin` to avoid potential harder exploit.
+  - System administrators can return `cross-origin: same-origin` with OHIF files to disallow any loading from other origin. It will block read access to resources loaded from a different origin to avoid potential attack vector.
 
   - Example config:
     ```js
--- a/platform/viewer/public/assets/example.json
+++ b/platform/viewer/public/assets/example.json
@@ -1,0 +1,216 @@
+{
+    "routerBasename": "/",
+    "extensions": [],
+    "modes": [],
+    "showStudyList": true,
+    "maxNumberOfWebWorkers": 6,
+    "maxNumRequests": {
+      "interaction": 100,
+      "thumbnail": 75,
+      "prefetch": 10
+    },
+    "dataSources": [
+      {
+        "friendlyName": "dcmjs DICOMWeb Server",
+        "namespace": "@ohif/extension-default.dataSourcesModule.dicomweb",
+        "sourceName": "dicomweb",
+        "configuration": {
+          "name": "DCM4CHEE",
+          "wadoUriRoot": "https://domvja9iplmyu.cloudfront.net/dicomweb",
+          "qidoRoot": "https://domvja9iplmyu.cloudfront.net/dicomweb",
+          "wadoRoot": "https://domvja9iplmyu.cloudfront.net/dicomweb",
+          "qidoSupportsIncludeField": false,
+          "supportsReject": false,
+          "imageRendering": "wadors",
+          "thumbnailRendering": "wadors",
+          "enableStudyLazyLoad": true,
+          "supportsFuzzyMatching": false,
+          "supportsWildcard": true,
+          "staticWado": true,
+          "singlepart": "bulkdata,video,pdf"
+        }
+      },
+      {
+        "friendlyName": "dicom json",
+        "namespace": "@ohif/extension-default.dataSourcesModule.dicomjson",
+        "sourceName": "dicomjson",
+        "configuration": {
+          "name": "json"
+        }
+      },
+      {
+        "friendlyName": "dicom local",
+        "namespace": "@ohif/extension-default.dataSourcesModule.dicomlocal",
+        "sourceName": "dicomlocal",
+        "configuration": {}
+      }
+    ],
+    "defaultDataSourceName": "dicomweb",
+    "hotkeys": [
+      {
+        "commandName": "incrementActiveViewport",
+        "label": "Next Viewport",
+        "keys": [
+          "right"
+        ]
+      },
+      {
+        "commandName": "decrementActiveViewport",
+        "label": "Previous Viewport",
+        "keys": [
+          "left"
+        ]
+      },
+      {
+        "commandName": "rotateViewportCW",
+        "label": "Rotate Right",
+        "keys": [
+          "r"
+        ]
+      },
+      {
+        "commandName": "rotateViewportCCW",
+        "label": "Rotate Left",
+        "keys": [
+          "l"
+        ]
+      },
+      {
+        "commandName": "invertViewport",
+        "label": "Invert",
+        "keys": [
+          "i"
+        ]
+      },
+      {
+        "commandName": "flipViewportHorizontal",
+        "label": "Flip Horizontally",
+        "keys": [
+          "h"
+        ]
+      },
+      {
+        "commandName": "flipViewportVertical",
+        "label": "Flip Vertically",
+        "keys": [
+          "v"
+        ]
+      },
+      {
+        "commandName": "scaleUpViewport",
+        "label": "Zoom In",
+        "keys": [
+          "+"
+        ]
+      },
+      {
+        "commandName": "scaleDownViewport",
+        "label": "Zoom Out",
+        "keys": [
+          "-"
+        ]
+      },
+      {
+        "commandName": "fitViewportToWindow",
+        "label": "Zoom to Fit",
+        "keys": [
+          "="
+        ]
+      },
+      {
+        "commandName": "resetViewport",
+        "label": "Reset",
+        "keys": [
+          "space"
+        ]
+      },
+      {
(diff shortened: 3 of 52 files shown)


Ouwen:
@sedghi @swederik @wayfarer3130 
This might be related to #2867

swederik:
That is great and I'm sure @wayfarer3130 will love this. Thanks!

Ouwen:
Force push to rebase

wayfarer3130:
> @sedghi @swederik @wayfarer3130 This might be related to #2867

No, it isn't related to 2867 - that PR allows loading of UMD JavaScript modules for modes and extensions, whereas this one just allows hot-swapping the config file

Ouwen:
hi folks, just wanted to follow up on this PR!

extensions/default/src/ViewerLayout/index.tsx
```diff
@@ -37,7 +37,15 @@ function ViewerLayout({
   const navigate = useNavigate();
 
   const onClickReturnButton = () => {
-    navigate('/');
+    let query = new URLSearchParams(window.location.search)
```
sedghi, line 40: can you please use const if you are not assigning query? and others? I know it doesn't matter, but the linter would fix it when saved, so better to do it now

sedghi:
Can you rebase this and we can then merge?

extensions/default/src/ViewerLayout/index.tsx
Ouwen: sure

Ouwen:
@sedghi rebased and should be ready for merge

sedghi:
@Ouwen I tried `http://localhost:3000/?configUrl=http://localhost:3000/config/example.json` and it shows a black screen. I don't think it is expected right?

platform/viewer/src/index.js
```diff
@@ -17,11 +17,46 @@ import ReactDOM from 'react-dom';
...
+    }
+  }
+
+  if (configUrl) {
+    const response = await fetch(configUrl)
+    return response.json()
+  }
+
+  return null
+}
```
sedghi, line 41: Can you refactor this into its own file similar to `loadDynamicImports`?

platform/viewer/src/index.js
```diff
@@ -17,11 +17,46 @@ import ReactDOM from 'react-dom';
...
+    if (obj) {
+      const query = new URLSearchParams(obj.search)
+      configUrl = query.get('configUrl')
+    }
+  }
+
+  if (configUrl) {
+    const response = await fetch(configUrl)
+    return response.json()
+  }
```
sedghi, line 38: This can go to the `else`

sedghi (changes requested):
I tried the example you put in the description and didn't work as I expected, also had one criticial and one optional comment

platform/viewer/src/index.js
```diff
@@ -17,11 +17,46 @@ import ReactDOM from 'react-dom';
...
+      sessionStorage.getItem('ohif-redirect-to')
+    );
+    if (obj) {
+      const query = new URLSearchParams(obj.search)
+      configUrl = query.get('configUrl')
+    }
+  }
+
+  if (configUrl) {
+    const response = await fetch(configUrl)
```
wayfarer3130, line 36: This leaves some nasty cross-site scripting (CSS) issues.  I'd like to see that the configUrl requires matching a given URL path as defined in the original config file.  Something like a flag in the base configuration whose value is a regular expression defining the allowed paths.  Then, in an example configuration make it a hard URL based path.  You could even add an automatic prefix to it, and a default path, eg  prefixConfigUrl: "./dynamicConfig/" with a prefixCheck of something like /[a-zA-Z0-9\-]*\.json/

sedghi: @Ouwen Does the recent push handle Bill's comment above?

Ouwen: @sedghi, the recent push does not account for Bill's comment. There are some other issues with the Access-Control-Headers not being set properly when this dynamic config is used (separate issue).

Regarding cross-site scripting, my understanding is that it should be managed by cross-origin policy rather than forcing the prefix in javascript. This way, if someone is trying to send a POST or PUT request to a different domain via some type of injection it would be stopped by cross-origin.

So `cross-origin = 'same-origin'` is safest.

Happy to still implement the prefix check if needed. In the meantime I'm looking into the access control issue.

sedghi: Honestly we can't rely the user to setup CORS correctly. So I would prefer what Bill suggested above (to have some regular expression test on the config to throw error on non allowed addresses)

cupidchan:
We are trying to contribute to this PR by adding `prefixConfigUrl` which @wayfarer3130 brought up. However, after a second thought, we agree with @Ouwen that `cross-origin = 'same-origin'` is the proper approach to handle cross-site scripting concern because of the following reason:

1. If a hacker comes in, this new dynamic server feature will only allow him to change which server OHIF will connect to. Assuming, the server belongs to him, there is not much "security concern" since we are not leaking the data from the intended DICOM server.
2. If the hacker is really that knowledgeable to change the setting and steal the data from the other DICOM servers, he should also be able to fake the `prefixConfigUrl`
3. Such knowledgeable hacker should also be able to steal the data from the original DICOM server from OHIF, even without this PR. 

Therefore, I propose to let this PR in, with the documentation enhancement to explain how to set `cross-origin = 'same-origin'` properly.

wayfarer3130:
> We are trying to contribute to this PR by adding `prefixConfigUrl` which @wayfarer3130 brought up. However, after a second thought, we agree with @Ouwen that `cross-origin = 'same-origin'` is the proper approach to handle cross-site scripting concern because of the following reason:
> 
> 1. If a hacker comes in, this new dynamic server feature will only allow him to change which server OHIF will connect to. Assuming, the server belongs to him, there is not much "security concern" since we are not leaking the data from the intended DICOM server.
> 2. If the hacker is really that knowledgeable to change the setting and steal the data from the other DICOM servers, he should also be able to fake the `prefixConfigUrl`
> 3. Such knowledgeable hacker should also be able to steal the data from the original DICOM server from OHIF, even without this PR.
> 
> Therefore, I propose to let this PR in, with the documentation enhancement to explain how to set `cross-origin = 'same-origin'` properly.

No, that isn't the concern I have - the issue is that if the configuration specifies a valid/standard back end for the retrieve part, but specifies something like WADO URL retrieves for the retrieve part or perhaps for STOW, then one might end up retrieving valid information and storing to something not permitted.  I know you only parse hte JSON, but there are places where that JSON gets used in a way that might leak PHI.

cupidchan:
Thanks for your additional information @wayfarer3130! Let me try to understand and please correct me if I am wrong. 

Supposed there is really WADO URL/STOW being set using this PR @Ouwen developed,  I understand that there may be many places that JSON is used. However, how is this different than the configuration without this PR? I mean in the current set up, user can also configure a WADO and leaking PHI can still be the resulted just like people using this PR to set the server dynamically, right?

anarnoli:
@wayfarer3130 Are you suggesting to put a regex filter for `configUrl` so that it always return valid config json? Can you please help to understand more on it? @cupidchan has some valid point and your input awaiting.

Ouwen:
I think the worry is that someone would use:
`mydomain.com/?configUrl=https://mymaliciouslink.com/cookies_get.js`

A regex filter would prevent the above. However, it would not prevent the following:
`mydomain.com/?configUrl=https://mymaliciouslink.com/cookies_get_escape_fetch.js`

A malicious user would have to craft javascript which can escape the fetch call context (maybe with buffer overflow?). This would require proper `cross-origin` settings to prevent PII data from being set out.

The regex whitelist adds some protection, but ultimately the cross-origin should probably be same-site to be safest.

cupidchan:
Now, I see your point now @Ouwen! If I may rephrase, having the regex enhancement will NOT 100% prevent the hacker to inject malicious script. In order to prevent the true attack, we need to set the `same-site` appropriately, right? If that's the case, then I suggest to put that in the document to explain this to OHIF users, instead of adding a regex check to prevent the incident partially. Another way is to add a flag (default to false) to enable this feature, so that users must be well aware they turn on this feature.

anarnoli:
I do agree with @Ouwen thoughts and @cupidchan's suggestion for flag can be a good option. @wayfarer3130 do you agree with this logic?

Ouwen:
@anarnoli, let me know if the above suits your needs (required rebase)

platform/viewer/public/config/default.js
```diff
@@ -24,6 +24,11 @@ window.config = {
     prefetch: 25,
   },
   // filterQueryParam: false,
+  /* Dynamic config allows user to pass "configUrl" query string this allows to load config without recompiling application. The regex will ensure valid configuration source */
+  dangerouslyUseDynamicConfig: {
+    enabled: false,
+    regex: /.*/,
```
wayfarer3130, line 30: Please provide an example regex that queries locally only, eg:
```
regex: /(0-9A-Za-z.]+)(\/[0-9A-Za-z.]+)*/
That allows numbers and letters in an absolute or sub-path, but not any absolute, with a comment saying that this is a local URL only.  Might have an example showing other specific endpoints, eg:
regex: /(https:\/\/hospital.com(\/[0-9A-Za-z.]+)*)|(https:\/\/othersite.com(\/[0-9A-Za-z.]+)*)/
which restricts to either hosptial.com or othersite.com.  That is the entire point of the restriction that it is possible to choose a specific source of configuration only.

Comment (anarnoli) [reply platform/viewer/public/config/default.js:30]: @wayfarer3130 will below example works?

```
// regex will ensure valid configuration source and default is /.*/ which matches any character. To use this, setup your own regex to choose a specific source of configuration only.
// Example 1, to allow numbers and letters in an absolute or sub-path only.
// regex: /(0-9A-Za-z.]+)(\/[0-9A-Za-z.]+)*/
// Example 2, to restricts to either hosptial.com or othersite.com.
// regex: /(https:\/\/hospital.com(\/[0-9A-Za-z.]+)*)|(https:\/\/othersite.com(\/[0-9A-Za-z.]+)*)/
```

Comment (sedghi) [reply platform/viewer/public/config/default.js:30]: can we add documentation inside https://v3-docs.ohif.org/configuration/configurationFiles as well
Discussing the security concerns and also link to this PR?

Comment (anarnoli) [reply platform/viewer/public/config/default.js:30]: Yes @sedghi, I've raised PR to @Ouwen for "how to use dangerouslyUseDynamicConfig in configuration documentation".

File: platform/docs/docs/configuration/configurationFiles.md [hunk 6]
```diff
@@ -116,7 +116,26 @@ Here are a list of some options available:
...
 - `showLoadingIndicator`: (default to true), if set to false, the loading indicator will not be shown when navigating between studies.
-
+- `dangerouslyUseDynamicConfig`: Dynamic config allows user to pass `configUrl` query string. This allows to load config without recompiling application. If the `configUrl` query string is passed, the worklist and modes will load from the referenced json rather than the default .env config. If there is no `configUrl` path provided, the default behaviour is used and there should not be any deviation from current user experience. This feature related to PR [OHIF#2925](https://github.com/OHIF/Viewers/pull/2925).<br/>
+Points to consider while using `dangerouslyUseDynamicConfig`:<br/>
+  - User have to enable this feature by setting `dangerouslyUseDynamicConfig.enabled:true`. By default it is `false`.
+  - Regex helps to avoid easy exploit. Dafault is `/.*/`. Setup your own regex to choose a specific source of configuration only.
+  - User must set `cross-origin:same-origin` to avoid potential harder exploit.
```
wayfarer3130, line 123: This comment isn't clear - how about
- System adminstrators can return cross-origin: same-origin with OHIF files to disallow any loading from single origin sites.  This prevents ...
-

anarnoli: @wayfarer3130 Does below description makes sense to you?

- System administrators can return `cross-origin: same-origin` with OHIF files to disallow any loading from other origin. It will block read access to resources loaded from a different origin to avoid potential attack vector.

wayfarer3130: Sounds good to me.

platform/viewer/public/assets/example.json
wayfarer3130, line 1: This example isnt needed as it is only referenced in the documentation as a possibility to include.

platform/viewer/src/index.js
```diff
@@ -15,9 +15,10 @@ import ReactDOM from 'react-dom';
  * to the window for processing.
  */
 import loadDynamicImports, { loadRuntimeImports } from './pluginImports.js';
+import loadDynamicConfig from './loadDynamicConfig';
 
-loadDynamicImports().then(() => {
-  loadRuntimeImports(window.config).then(() => {
+Promise.all([loadDynamicImports(), loadDynamicConfig(window.config)]).then(
```
wayfarer3130, line 20: The dynamic config must get loaded first, because it might define dynamic imports to use.

platform/viewer/src/loadDynamicConfig.js
```diff
@@ -0,0 +1,31 @@
...
+    // If enabled then get configUrl param
+    let query = new URLSearchParams(window.location.search);
+    let configUrl = query.get('configUrl');
+
+    if (!configUrl) {
+      // Handle OIDC redirects
+      const obj = JSON.parse(sessionStorage.getItem('ohif-redirect-to'));
+      if (obj) {
+        const query = new URLSearchParams(obj.search);
+        configUrl = query.get('configUrl');
```
wayfarer3130, line 15: You never do anything with the configUrl beyond this point, so there is no point assigning anything here.

platform/viewer/src/routes/WorkList/WorkList.tsx
```diff
@@ -324,13 +324,13 @@ function WorkList({
...
             seriesInStudiesMap.has(studyInstanceUid)
               ? seriesInStudiesMap.get(studyInstanceUid).map(s => {
-                  return {
-                    description: s.description || '(empty)',
-                    seriesNumber: s.seriesNumber ?? '',
-                    modality: s.modality || '',
-                    instances: s.numSeriesInstances || '',
-                  };
-                })
+                return {
```
wayfarer3130, line 327: @sedghi - this type of formatting change is being introduced by the eslint stage.  Either we should disable it or we should enforce it.  I don't care either way, but just FYI

platform/viewer/public/assets/example.json
Ouwen: Hi @wayfarer3130, I think it may be good to include the `.json` example since the dynamic method cannot take `.js` files. Taking `js` files would require the use of javascript `eval` which may be dangerous.

platform/viewer/src/index.js
Ouwen: We can waterfall this promise in ~then~

platform/viewer/src/loadDynamicConfig.js
Ouwen: good catch

platform/viewer/src/routes/WorkList/WorkList.tsx
```diff
@@ -343,21 +343,26 @@ function WorkList({
...
+            if (filterValues.configUrl) {
+              query.append('configUrl', filterValues.configUrl);
+            }
             return (
               <Link
                 key={i}
                 to={`${dataPath ? '../../' : ''}${mode.routeName}${dataPath ||
-                  ''}?StudyInstanceUIDs=${studyInstanceUid}`}
-                // to={`${mode.routeName}/dicomweb?StudyInstanceUIDs=${studyInstanceUid}`}
+                  ''}?${decodeURIComponent(query.toString())}`}
```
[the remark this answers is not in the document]
Ouwen: Hmmm, the reason for this was just to have a easy way to manage the params without too much string manipulation. The main issue would be the replacement of special characters into code form like `%332%`. If the user needs to parse the URL, these code would be parsed into their utf-8 symbol by `new URLSearchParams`

sedghi:
@Ouwen @cupidchan The v3-stable has moved forward, you need to resolve the conflicts please

anarnoli:
Suggested changes by @wayfarer3130 covered in new PR https://github.com/OHIF/Viewers/pull/3391 to avoid conflict with `v3-stable` branch.

Issue #2911, opened 2022-09-05
Storybook in v3-stable branch does not work


eric-taix:
### Describe the Bug

You're trying to set our own styles to components and we'd like to see the result in the provided storybook. We are using the latest v3 branch (`v3-stable`) and even if the storybook server is started, the console displays lot of errors and no component is displayed.
### What steps can we follow to reproduce the bug?
We followed the file `platform/ui/README.md`:

1. `cd platform/ui` 
2. `yarn install`
3. `yarn storybook`

First steps work fine, but the third one generates the following error (snippet):
```
javascript modules 5.39 MiB
  modules by path ../../node_modules/ 5.39 MiB 919 modules
  modules by path ./ 662 bytes
    ./.storybook/manager.js + 2 modules 647 bytes [built] [code generated]
    ./util.inspect (ignored) 15 bytes [built] [code generated]
json modules 1.52 KiB
  ../../node_modules/character-entities-legacy/index.json 1.24 KiB [built] [code generated]
  ../../node_modules/character-reference-invalid/index.json 289 bytes [built] [code generated]
./static/favicon.png 42 bytes (javascript) 443 bytes (asset) [built] [code generated]
manager (webpack 5.72.1) compiled successfully in 24880 ms
99% done plugins webpack-hot-middlewarewebpack built preview 55a271a0b2ff6e7f3112 in 29325ms
WARN Force closed manager build
ModuleBuildError: Module build failed (from ../../node_modules/babel-loader/lib/index.js):
SyntaxError: C:\Users\eric.taix\development\intrasense\oa-ohif\platform\ui\.storybook\preview.js: Unexpected token, expected "," (21:21)

  19 |     inlineStories: true,
  20 |     container: ({ children, context }) => (
> 21 |       <DocsContainer context={context}>{children}</DocsContainer>
     |                      ^
  22 |     ),
  23 |     page: DocsPage,
  24 |     components: {
    at instantiate (C:\Users\eric.taix\development\intrasense\oa-ohif\node_modules\@babel\parser\lib\index.js:72:32)
```

I'd like to help to solve this issue, but I've no idea where to start. Any guidelines / links would be helpful.

PR #2926, opened 2022-09-13
fix: storybook with typescript

sedghi:
Storybook in `v3-stable` was broken as reported also by https://github.com/OHIF/Viewers/issues/2911

@eric-taix Can you see if this also fixes the issue you posted too?


eric-taix:
I gave it a try and yes, this PR fixes #2911 👍

PR #2931, opened 2022-09-14
fix(datasource): configurability of additional data source extensions

wayfarer3130:
This just fixes the configurability of additional data source extensions - without it, there isn't a way to define a second data source type without removing the default extension.

### DIFF
--- a/platform/core/src/types/Consumer.ts
+++ b/platform/core/src/types/Consumer.ts
@@ -1,0 +1,3 @@
+import Params from './ParamsType';
+type Consumer = (props: Params) => void;
+export default Consumer;
--- a/platform/core/src/types/Consumer.ts
+++ b/platform/core/src/types/Consumer.ts
@@ -1,3 +1,5 @@
-import Params from './ParamsType';
-type Consumer = (props: Params) => void;
+/**
+ * Just a function that consumes a single argument, with no return.
+ */
+type Consumer = (props: Record<string, unknown>) => void;
 export default Consumer;
--- a/platform/core/src/types/index.ts
+++ b/platform/core/src/types/index.ts
@@ -1,0 +1,9 @@
+import {
+  StudyMetadata,
+  SeriesMetadata,
+  InstanceMetadata,
+} from './StudyMetadata';
+
+import Consumer from './Consumer';
+
+export { StudyMetadata, SeriesMetadata, InstanceMetadata, Consumer };
--- a/platform/viewer/src/routes/DataSourceWrapper.tsx
+++ b/platform/viewer/src/routes/DataSourceWrapper.tsx
@@ -32,11 +32,16 @@ function DataSourceWrapper(props) {
     return acc.concat(mods);
   }, []);
 
-  // Grabbing first for now - should get active?
-  const name = webApiDataSources[0].name;
+  // Grabbing first defined for now - should get active
   // TODO: Why does this return an array?
-  const dataSource = extensionManager.getDataSources(name)[0];
-
+  const dataSource = webApiDataSources
+    .map(ds => extensionManager.getDataSources(ds.name)?.[0])
+    .find(it => it !== undefined);
+  if (!dataSource) {
+    throw new Error(
+      `No data source found for any of ${webApiDataSources.map(it => it.name)}`
+    );
+  }
   // Route props --> studies.mapParams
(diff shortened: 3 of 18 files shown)


platform/viewer/src/routes/DataSourceWrapper.tsx
```diff
@@ -32,11 +32,16 @@ function DataSourceWrapper(props) {
     return acc.concat(mods);
   }, []);
 
-  // Grabbing first for now - should get active?
-  const name = webApiDataSources[0].name;
+  // Grabbing first defined for now - should get active
   // TODO: Why does this return an array?
-  const dataSource = extensionManager.getDataSources(name)[0];
-
+  const dataSource = webApiDataSources
```
wayfarer3130, line 37: The logic is now to find the first defined data source - that isn't technically correct, as there should be a way to set it from configuration, but I was mostly just trying to get the base definitions working with an extension.  @sedghi  - if you want to use configuration, if you could suggest how to do that, I'd appreciate it.

sedghi:
Unit test is failing 

image

wayfarer3130:
@sedghi  - as requested, I moved the types into their own directory.  Also fixed the unit test.

platform/core/src/types/Consumer.ts
```diff
@@ -0,0 +1,3 @@
+import Params from './ParamsType';
+type Consumer = (props: Params) => void;
+export default Consumer;
```
sedghi, line 3: What is Consumer? I don't think we need this level of typings, but if you want, can we maybe call it like `Callback`?

sedghi: Also I could not comment on `StudyMetadata` but that file is empty and should be removed

platform/core/src/types/ParamsType.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * The type of generic parameter objects - this fixes issues with
+ * Object and object and any in many cases.
+ */
+type Params = Record<string, unknown>;
+export default Params;
```
sedghi, line 6: This is again another type that I don't think we need at all. `Record<string, unknown>` is well defined. What if there is like `Record<number, unknown>` or `Record<unkonwn, unknown`.

platform/core/src/types/index.ts
```diff
@@ -0,0 +1,16 @@
...
+} from './StudyMetadata';
+import Params from './ParamsType';
+import Consumer from './Consumer';
+
+// These are just Params right now, but eventually the idea is to
+// create a full interface object here - at that point they should
+// move to their own files.
+export type ICommandsManager = Params;
+
+export type IGuiCustomization = Params;
```
sedghi, line 14: `ICommandsManager = Params` again is so weird, lets just put it as `unkown` or `any` or at worst the `Record<string, unknown>` until we type it

sedghi: Also the IGuiCustomization stuff is not merged yet so let's remove it too

platform/core/src/types/Consumer.ts
wayfarer3130: That was a deleted file.

Add documentation to Consumer - that is a standard definition/actor type, and is useful to have a name for rather than declaring it every time it gets used - they are pretty common.

sedghi: Params are `Record<string, unknown` and `Consumer` is getting `Params` as its function argument. The reason I say we don't need this is that there can be multiple (if not infinite) combination of all of this 

Params that is not just one parameter, a `Record<string, unkonwn` and another optional parameter, then based on this you probably will say `myFunction(params: Params, anotherParam: AnotherType)`. We can't assume the parameters of a function are just in type of `Record<string, unkonwn` neither we can assume that a "consumer" is always taking one argument and is a void in return. What about another consumer which returns something? 

In summary, calling these Params and Consumer is a very general term which adds a lot to ambiguity, if they are event handler's parameters and consumer, then let's call them EventHandlerParam, and EventHandlerFunction or Consumer or whatever.

wayfarer3130:
I think I have all your issues resolved @sedghi, except that I left the Consumer as that is a well known name - ideally it would be a Consumer on a given type, but on a general class it works fairly well.

sedghi (approved):

PR #2933, opened 2022-09-15
feat: upgrade tailwind to version 3

sedghi:
Why? -> https://tailwindcss.com/blog/tailwindcss-v3

- Just in time engine
- More classes
- Arbitrary properties `m-[13px]`

### Considerations moving forward
Tailwind v3 classes does not overwrite the previous ones if they are coming after each other for instance `bg-black bg-primary-main` would not necessarily make the background with color of `primary-main`. It depends which CSS classes is created first. However, all some of the components in our UI library (e.g., Button.tsx) systems rely on such overwriting. There are two options moving forward for developers
- Using a third party library such as [tailwind merge](https://www.npmjs.com/package/tailwind-merge) to handle the merge 
- Introduce APIs via props (as done for Button currently) to overwrite defaults

Long term we should rework the library to enable the second option, until Tailwind introduce a `initial:` variant (if that ever happens)

### DIFF
--- a/extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/ActionButtons.tsx
+++ b/extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/ActionButtons.tsx
@@ -9,16 +9,22 @@ function ActionButtons({ onExportClick, onCreateReportClick }) {
 
   return (
     <React.Fragment>
-      <ButtonGroup color="black" size="inherit">
-        <Button className="text-base px-2 py-2" onClick={onExportClick}>
-          {t('Export')}
-        </Button>
-      </ButtonGroup>
       <Button
-        className="ml-2 text-base"
+        className="text-base px-2 py-2"
+        size="initial"
         variant="outlined"
-        size="small"
+        color="default"
+        border="secondary"
+        onClick={onExportClick}
+      >
+        {t('Export')}
+      </Button>
+      <Button
+        className="ml-2 px-2 text-base"
+        variant="outlined"
+        size="initial"
         color="black"
+        border="secondary"
         onClick={onCreateReportClick}
       >
         {t('Create Report')}
--- a/platform/ui/src/components/Button/Button.tsx
+++ b/platform/ui/src/components/Button/Button.tsx
@@ -10,6 +10,7 @@ const defaults = {
   disabled: false,
   fullWidth: false,
   rounded: 'medium',
+  border: 'default',
   size: 'medium',
   type: 'button',
   variant: 'contained',
@@ -23,7 +24,7 @@ const roundedClasses = {
   full: 'rounded-full',
 };
 
-const variantClasses = {
+const variants = {
   text: {
     default:
       'text-primary-light hover:bg-primary-light hover:text-white active:opacity-80 focus:bg-primary-light focus:text-white',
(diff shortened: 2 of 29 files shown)


platform/ui/src/components/Button/Button.tsx
```diff
@@ -176,6 +233,14 @@ Button.propTypes = {
     'inherit',
     'light',
   ]),
+  border: PropTypes.oneOf([
```
ladeirarodolfo, line 236: From current architecture stand point this design is very hard to maintain and scale.


I believe border props should have something like this
(from borderClasses)
none: '',
  light: 'border border-primary-light',
  primary: 'border border-primary-main',
  primaryActive: 'border border-primary-active',
  secondary: 'border border-secondary-light',
  white: 'border border-white',
  black: 'border border-black',
  
  
  also I have realized color is not consistent

ladeirarodolfo: we definetly one day should revisit how components has being implemented/design that decrease a lot the style customization

platform/ui/src/components/Button/Button.tsx
```diff
@@ -38,41 +39,92 @@ const variantClasses = {
...
 };
 
+const _getVariantClass = (variant, color, bgColor = null) => {
+  const defaultBackgroundColor = defaultVariantBackGroundColor[variant][color];
+
+  if (!bgColor) {
+    bgColor = defaultBackgroundColor;
+  }
+
+  return `${variants[variant][color]} ${bgColor}`;
```
ladeirarodolfo, line 115: what is the impact on component if bgColor is null?

platform/ui/src/components/Button/Button.tsx
```diff
@@ -38,41 +39,92 @@ const variantClasses = {
   },
   outlined: {
     default:
-      'border bg-trasparent border-primary-light text-primary-light hover:bg-primary-light hover:text-black focus:text-black focus:bg-primary-light active:opacity-80',
+      'text-primary-light hover:bg-primary-light hover:text-black focus:text-black focus:bg-primary-light active:opacity-80',
```
ladeirarodolfo, line 42: Not your fault but just sharing my frustration  when I see things like this on OHIF. Not DRY at all

platform/ui/src/components/StudyListPagination/StudyListPagination.tsx
```diff
@@ -33,7 +35,7 @@ const StudyListPagination = ({
         <div className="flex justify-between">
           <div className="flex items-center">
             <Select
-              id={"rows-per-page"}
+              id={'rows-per-page'}
```
ladeirarodolfo, line 38: maybe instead we should follow REact/JSX good practices and do not use curly braces for static content such strings
I.e
id="rows-per-page"

ladeirarodolfo (changes requested):
Hey Alli, thank for your PR. In general Pr looks good, there are only few items to cover (happy to discuss)

There is just one thing that I would like to bring up. There are some formatting changes that indicates linter/formatting is not covering all files. We should not see such changes on any PR at all. So, I have two considerations only:
- Ensure all files are captured (can be later)
- Ensure the actual formatting changes (your PR) is what the automatic would do (this might be true if config came later file changes)

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/ActionButtons.tsx
```diff
@@ -9,16 +9,22 @@ function ActionButtons({ onExportClick, onCreateReportClick }) {
...
   return (
     <React.Fragment>
-      <ButtonGroup color="black" size="inherit">
-        <Button className="text-base px-2 py-2" onClick={onExportClick}>
-          {t('Export')}
-        </Button>
-      </ButtonGroup>
       <Button
-        className="ml-2 text-base"
+        className="text-base px-2 py-2"
```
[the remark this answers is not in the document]
sedghi: Tailwind 3 has changed overriding classes so to make it look like previously I had to change the styles otherwise there would be UI regression

platform/ui/src/components/Button/Button.tsx
sedghi: line 111 it checks bgColor, if null then default color is used

PR #2936, opened 2022-09-16
feat: explicit layout apply via HangingProtocolService

sedghi:
### PR Checklist

### DIFF
--- a/extensions/cornerstone/src/commandsModule.ts
+++ b/extensions/cornerstone/src/commandsModule.ts
@@ -420,7 +420,7 @@ const commandsModule = ({ servicesManager }) => {
         (activeViewportIndex - 1 + viewports.length) % viewports.length;
       ViewportGridService.setActiveViewportIndex(nextViewportIndex);
     },
-    setHangingProtocol: (protocolId: string) => {
+    setHangingProtocol: ({ protocolId }) => {
       HangingProtocolService.setProtocol(protocolId);
     },
   };
--- a/extensions/cornerstone/src/commandsModule.ts
+++ b/extensions/cornerstone/src/commandsModule.ts
@@ -218,8 +218,6 @@ const commandsModule = ({ servicesManager }) => {
             CornerstoneViewportService,
           },
         });
-
-        return;
       }
     },
     rotateViewport: ({ rotation }) => {
@@ -422,121 +420,8 @@ const commandsModule = ({ servicesManager }) => {
         (activeViewportIndex - 1 + viewports.length) % viewports.length;
       ViewportGridService.setActiveViewportIndex(nextViewportIndex);
     },
-    activateMPRForActiveViewport: () => {
-      const { activeViewportIndex, viewports } = ViewportGridService.getState();
-      const viewportDisplaySetInstanceUIDs =
-        viewports[activeViewportIndex].displaySetInstanceUIDs;
-
-      const displaySetsToHang = viewportDisplaySetInstanceUIDs.map(
-        displaySetInstanceUID => {
-          const displaySet = DisplaySetService.getDisplaySetByUID(
-            displaySetInstanceUID
-          );
-
-          return displaySet;
-        }
-      );
-
-      if (displaySetsToHang.some(ds => !ds.isReconstructable)) {
-        UINotificationService.show({
-          title: 'Multiplanar reconstruction (MPR) ',
-          message:
-            'Cannot create MPR for this series since it is not reconstructable.',
-          type: 'warning',
-          displayTime: 3000,
-        });
-
-        return;
-      }
-
-      const matchingDisplaySets = {};
-
-      displaySetsToHang.forEach(displaySet => {
-        const {
-          displaySetInstanceUID,
-          SeriesInstanceUID,
-          StudyInstanceUID,
-        } = displaySet;
-
-        matchingDisplaySets[displaySetInstanceUID] = {
-          displaySetInstanceUID,
-          SeriesInstanceUID,
-          StudyInstanceUID,
-        } as Types.HangingProtocol.DisplaySetMatchDetails;
-      });
-
-      const hpViewports: Types.HangingProtocol.Viewport[] = [
-        'axial',
-        'sagittal',
-        'coronal',
-      ].map(viewportOrientation => {
-        return {
-          viewportOptions: {
-            toolGroupId: 'mpr',
-            viewportType: 'volume',
-            orientation: viewportOrientation,
-            initialImageOptions: {
-              preset: 'middle',
-            },
-            syncGroups: [
-              {
-                type: 'voi',
-                id: 'mpr',
-                source: true,
-                target: true,
-              },
-            ],
-          },
-          displaySets: viewportDisplaySetInstanceUIDs.map(
-            displaySetInstanceUID => {
-              return {
-                id: displaySetInstanceUID,
-              };
-            }
-          ),
-        };
-      });
-
-      const mprProtocol: Types.HangingProtocol.Protocol = {
-        id: 'mpr',
-        stages: [
-          {
-            id: 'mprStage',
-            name: 'mpr',
-            viewportStructure: {
-              layoutType: 'grid',
-              properties: {
-                rows: 1,
-                columns: 3,
-                layoutOptions: [
-                  {
-                    x: 0,
-                    y: 0,
-                    width: 1 / 3,
-                    height: 1,
-                  },
-                  {
-                    x: 1 / 3,
-                    y: 0,
-                    width: 1 / 3,
-                    height: 1,
-                  },
-                  {
-                    x: 2 / 3,
-                    y: 0,
-                    width: 1 / 3,
-                    height: 1,
-                  },
-                ],
-              },
-            },
-            displaySets: [],
-            viewports: hpViewports,
-          },
-        ],
-      };
-
-      HangingProtocolService.applyProtocol(mprProtocol, matchingDisplaySets);
+    setHangingProtocol: (protocolId: string) => {
+      HangingProtocolService.setProtocol(protocolId);
     },
   };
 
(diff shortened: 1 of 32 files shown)


extensions/cornerstone/src/commandsModule.ts
```diff
@@ -213,6 +218,8 @@ const commandsModule = ({ servicesManager }) => {
             CornerstoneViewportService,
           },
         });
+
+        return;
```
wayfarer3130, line 222: This seems an odd thing to have a return statement here - it returns at the end of the method regardless, and there isn't more code after that.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -415,6 +422,122 @@ const commandsModule = ({ servicesManager }) => {
...
+
+        matchingDisplaySets[displaySetInstanceUID] = {
+          displaySetInstanceUID,
+          SeriesInstanceUID,
+          StudyInstanceUID,
+        } as Types.HangingProtocol.DisplaySetMatchDetails;
+      });
+
+      const hpViewports: Types.HangingProtocol.Viewport[] = [
+        'axial',
```
wayfarer3130, line 469: Can you add acquisition here?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -415,6 +422,122 @@ const commandsModule = ({ servicesManager }) => {
...
+            displaySetInstanceUID => {
+              return {
+                id: displaySetInstanceUID,
+              };
+            }
+          ),
+        };
+      });
+
+      const mprProtocol: Types.HangingProtocol.Protocol = {
```
wayfarer3130, line 500: Please import this from it's own file

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: I see what you are doing, you are generating a hanging protocol to apply.  Is there any reason not to allow applying a generic hanging protocol, and just define it in the module in the normal way?  That way one could define buttons/commands to apply arbitrary hanging protocols, and you wouldn't have this large chunk of data generation in the middle of the commands module.

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -286,6 +286,23 @@ const toolbarButtons = [
...
+  //   id: 'MPR',
+  //   type: 'ohif.action',
+  //   props: {
+  //     icon: 'old-play',
+  //     label: 'MPR',
+  //     type: 'action',
+  //     commands: [
+  //       {
+  //         commandName: 'activateMPRForActiveViewport',
+  //         commandOptions: {},
```
wayfarer3130, line 300: I suggest defining a module exported hanging protocol name here that can be applied.  That way, all hanging protocols are defined in the same way, and a mode can changes the "MPR" layout to whatever they want just by switching the HP to apply.    I would also suggest specifying the default stage to start with, at least as a possible option.

platform/core/src/types/HangingProtocol.ts
```diff
@@ -0,0 +1,134 @@
+type DisplaySetInfo = {
```
wayfarer3130, line 1: Thank you for adding the typing for this - it really helps to get it typed.

platform/core/src/types/index.ts
```diff
@@ -0,0 +1,5 @@
+import type * as HangingProtocol from './HangingProtocol'
+
+export {
```
wayfarer3130, line 3: Why isn't this one export type { ...
It looks like it is otherwise the same as the import.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -415,6 +422,122 @@ const commandsModule = ({ servicesManager }) => {
         (activeViewportIndex - 1 + viewports.length) % viewports.length;
       ViewportGridService.setActiveViewportIndex(nextViewportIndex);
     },
+    activateMPRForActiveViewport: () => {
```
wayfarer3130, line 425: Can you call this applyHangingProtocol, taking the name of a HP extension export as the parameter and optionally a stage.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -392,7 +401,7 @@ class HangingProtocolService {
     }
 
     this.customImageLoadPerformed = false;
-    const { type: layoutType } = stageModel.viewportStructure;
+    const { layoutType } = stageModel.viewportStructure;
```
wayfarer3130, line 404: Is it necessary to rename the field name?  Probably it is, given the use of type as a keyword.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -313,7 +297,20 @@ class HangingProtocolService {
     return protocol;
   }
 
-  _setProtocol(protocol) {
+  public applyProtocol(
+    protocol: HangingProtocol.Protocol,
```
wayfarer3130, line 301: It should be possible to apply the protocol by name instead of handing it a full protocol in order to re-apply existing protocols, eg to go back to the previously applied protocol (un-apply the MPR mode).

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -313,7 +297,20 @@ class HangingProtocolService {
...
+    matchingDisplaySets: Record<string, HangingProtocol.DisplaySetMatchDetails>
+  ): void {
+    protocol = this._validateProtocol(protocol);
+    this.protocols.push(protocol);
```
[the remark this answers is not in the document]
wayfarer3130: It also changes the initial display conditions if you add it to the protocols list - what happens if you:
1. Display a study in the default protocol
2. Apply the MPR mode
3. Apply the "default" protocol again as a apply most applicable protocol - except this time it hits the MPR as the most applicable protocol.

I'd rather see the protocol name be the actual id of a protocol here to apply - the protocols can then be added in the default fashion, and to turn it off, just avoid setting any conditions on the protocol.   That also allows your MPR check to be generic, just call the study level protocol conditions.  Have the extension register a custom attribute that checks if the study contains a display set for which the response for the check is true.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -313,7 +297,20 @@ class HangingProtocolService {
     return protocol;
   }
 
-  _setProtocol(protocol) {
+  public applyProtocol(
+    protocol: HangingProtocol.Protocol,
+    matchingDisplaySets: Record<string, HangingProtocol.DisplaySetMatchDetails>
```
wayfarer3130, line 302: It should be allowed to supply the matchingDisplaySets optionally - it is entirely possible to just reapply a specific protocol without specifying the display sets

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
sedghi: Yes it was actually a bug, it is defined as layoutType in the protocols

sedghi:
@wayfarer3130 

1. Changed the HP Service to be able to register both Object and Protocol Generator (dynamic) protocols via `id` 
2. Changed the protocol registration to happen at the extension registration, previously it was done by the modes which was wrong, now a mode can define which protocol it should run with, if undefined we will run protocol Matching rules. 
3. Recent PR added inheritance to ViewportGrid which should not happen, as we cannot assume the next displaySets and previous displaySets on the same viewport share the same options.

sedghi:
I need to change docs as well, after we settle on the changes @sedghi

extensions/default/src/index.js
```diff
@@ -2,9 +2,9 @@ import getDataSourcesModule from './getDataSourcesModule.js';
 import getLayoutTemplateModule from './getLayoutTemplateModule.js';
 import getPanelModule from './getPanelModule';
 import getSopClassHandlerModule from './getSopClassHandlerModule.js';
-import getHangingProtocolModule from './getHangingProtocolModule.js';
```
wayfarer3130, line 5: Is there a good reason for this change here?  It seems an empty change...

extensions/tmtv/src/getHangingProtocolModule.js
```diff
@@ -15,6 +15,17 @@ const ptCT = {
   ],
   imageLoadStrategy: 'interleaveTopToBottom', // "default" , "interleaveTopToBottom",  "interleaveCenter"
   protocolMatchingRules: [
+    {
+      id: 'wauZK2QNEfDPwcAQo',
+      weight: 1,
+      attribute: 'ModalitiesInStudy',
+      constraint: {
+        contains: {
```
wayfarer3130, line 23: Use the contains: ['PT', 'CT'] syntax - it is a little easier to read.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -116,30 +94,91 @@ class HangingProtocolService {
...
+
+    return protocols;
   }
 
-  public addProtocols(protocols) {
-    protocols.forEach(protocol => {
-      if (this.protocols.indexOf(protocol) === -1) {
-        this.protocols.push(this._validateProtocol(protocol));
+  /**
+   * Returns the protocol with the given id
```
wayfarer3130, line 138: Please add documentation that it will call the generator function to create/update the protocol as necessary.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -524,6 +657,7 @@ class HangingProtocolService {
         const seriesMatchDetails = this.protocolEngine.findMatch(
           displaySet,
           seriesMatchingRules,
+          // Todo: why we have images here since the matching type does not have it
```
wayfarer3130, line 660: The reason to have images here is to allow getting the metadata to read attributes not included in the study by default - there are other possible patient/study level attributes.

wayfarer3130:
LGTM after the docs and couple of small changes are applied.

extensions/default/src/index.js
sedghi: That was a mistake

PR #2938, opened 2022-09-16
feat: Multiframe metadata provider

wayfarer3130:
The fix here is to read multiframe metadata in a way that is compatible with the volume viewing.


platform/core/src/classes/MetadataProvider.js
```diff
@@ -408,14 +415,18 @@ class MetadataProvider {
     //   SOPInstanceUID,
     // })
     // somewhere else
-    if (imageId.startsWith('wadors:')) {
+    if (
+      imageId.startsWith('wadors:') ||
+      imageId.startsWith('streaming-wadors:')
```
wayfarer3130, line 420: The base metadata was registered specifically, but the remaining ones were missing, so it was throwing an exception.  This allows parsing the streaming-wadors URL the same way the base URL is parsed.

platform/core/src/utils/combineFrameInstance.ts
```diff
@@ -0,0 +1,27 @@
+import Combine from './Combine';
+
+/**
+ * Combine the Per instance frame data, the shared frame data
```
wayfarer3130, line 4: This allows getting a proxy object that will just find the right instance data.

platform/core/src/utils/combineFrameInstance.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+const combineFrameInstance = (frame, instance) => {
+  const {
+    PerFrameFunctionalGroupsSequence,
+    SharedFunctionalGroupsSequence,
+  } = instance;
+  if (!PerFrameFunctionalGroupsSequence || frame === undefined) return instance;
+  const shared = Object.values(SharedFunctionalGroupsSequence[0])
+    .map(it => it[0])
+    .filter(it => it !== undefined && typeof it === 'object');
```
wayfarer3130, line 15: Technically, the non-object values should be merged too, but as the only thing tthere is private creators, and the private instance on that gets munged out of the return object, I don't think we really need them.

platform/core/src/utils/Combine.ts
```diff
@@ -0,0 +1,36 @@
...
+  get: (target, prop, receive) => {
+    for (const obj of target) {
+      if (prop in obj) {
+        return obj[prop];
+      }
+    }
+  },
+
+  set: (target, prop, value) => {
+    target[0][prop] = value;
```
md-prog, line 11: Due to the way that we set the first object when we use `set()` method, I think we need to add an explicit warning about the possible changes to the original objects used to create the proxy when use `set()` method on the proxy.

Or instead, we can have internal "empty" target object that sits at the index:0, removing the need of using empty object(`{}`) when we call `Combine()`

```
Combine({}, obj1, obj2, ...)
```

platform/core/src/utils/Combine.ts
```diff
@@ -0,0 +1,36 @@
...
+    return Object.keys(ret);
+  },
+};
+
+/**
+ * Combine multiple objects into a single response object, just by "merging" the whole kit an caboodle.
+ * @param args to proxy as though they were a hierarchy
+ * @returns
+ */
+export default function Combine(...args) {
```
sedghi, line 34: If this is merge, it should be called Merge. Also, how is is it different than loadash merge https://lodash.com/docs/4.17.15#merge which is already implemented and tested heavily

sedghi:
@md-prog @wayfarer3130 
Great stuff ! 💯 🥇 
See my comments. Question: does it also support stack of images? like multiframe ultrasound?

Build is failing see below 

image

platform/core/src/utils/Combine.ts
wayfarer3130: Added.

platform/core/src/utils/Combine.ts
wayfarer3130: It is neither merge nor assign - it is actually a proxy dynamic combine, which is kind of like Object.assign or _.assign, but has different performance characteristics, appropriate for the use it is getting.  Probably I should have used merge to avoid premature optimization.

wayfarer3130:
Fixed the PR requested changes - makes it mroe files changed, but a simpler change overall.
Added support for US multiframe as well (this was independent, but since it is MF, and easy to do, I included it)

sedghi (commented):
This has a bug in which the `combineFrameInstance` is getting called even if frame is undefined (which shouldn't)

wayfarer3130:
I'm not sure that was really a bug - the combine call just returns instance in that case, or returns the metadata for frame 1 if that is available.  However, I updated the PR with a check

sedghi:
Maybe not a bug for certainly readability 

```js
if (frameNumber){
    return combineFrameInstance(frameNumber, instance)
}

return instance
``` 

is my preference but it's fine I guess the way you wrote it too

PR #2939, opened 2022-09-17
Frametimes - Auto

Rayhan-011:
There are a few gaps in functionality that we need to add to the open source OHIF code:

1) The viewer needs the ability to load ultrasound cine and auto-play (without the user needing to press the play button). The framerate should be set automatically based on the stored frametime which is stored in the DICOM tag (0018,1063). See: https://github.com/OHIF/Viewers/blob/master/platform/ui/src/viewer/PlayClipButton.js

2) Also the animation should be smooth when frametimes are short. In the existing code, the animation is not smooth when frametimes are short (greater than 30fps). The code shoud instead set a resonable framerate (like less than 60fps) and then when animation frames are needed, calculate the proper frame based on actual elapsed time since the last frame. This would mean that the animation would play in real time even if the computer cannot process the frames fast enough. See: https://github.com/cornerstonejs/cornerstoneTools/blob/9ea015d9d8dbb2d1be43ca9c1f7d6bf61f153910/src/stackTools/playClip.js#L259

3) Reference lines should be displayed on the image. This is a feature that is currently not available in the open source OHIF code but is available in Cornerstone. https://tools.cornerstonejs.org/examples/tools/reference-lines.html


Rayhan-011:
@horner can you please check this pull request - Thanks

sedghi:
we should investigate this in v3

Rayhan-011:
> we should investigate this in v3

I committed my code with functionality addition that you can use in V3

sedghi:
Can you then maybe change the base to master?

jbocce:
This PR is being closed in favour of this new PR https://github.com/OHIF/Viewers/pull/3735 which targets the most current master branch. Thanks @Rayhan-011 for the inspiration and ideas!

PR #2946, opened 2022-09-20
V3 Use HP to prevent viewport trying to display RTPLAN, SR, REG, etc. 

Michael-Andersen:
When opening a study with series that don't render (such as RTPLAN, RTDOSE, RTSTRUCT, SR, REG) the non-rendering series may be selected for the main viewport so nothing will load there, forcing the use to double click a thumbnail to load the first series rather than doing so by default. 

I've edited the default hanging protocol to address this. There may be a more comprehensive way to do this, either by adding more non-rendering modalities or by using a list of the modalities that will render. I'm open to reformatting/expanding the fix I have started here.

### DIFF
--- a/extensions/default/src/getHangingProtocolModule.js
+++ b/extensions/default/src/getHangingProtocolModule.js
@@ -26,7 +26,33 @@ const defaultProtocol = {
           // Unused currently
           imageMatchingRules: [],
           // Matches displaysets, NOT series
-          seriesMatchingRules: [],
+          seriesMatchingRules: [
+            {
+              id: 'nRl6SgkPMx8HzLDVT',
+              weight: 1,
+              attribute: 'Modality',
+              constraint: {
+                doesNotContain: {
+                  value: 'RT',
+                },
+                doesNotEqual: {
+                  value: 'SR'
+                },
+              },
+              required: true,
+            },
+            {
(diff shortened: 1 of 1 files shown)


Michael-Andersen:
@sedghi I don't think I have the permissions to add you as a reviewer but during the last QIPCM-OHIF meeting you suggested I make a PR for this so you could take a look. My original logic was in ViewportGrid.ts checking available display sets for ImageSets but fixing this with the hanging protocol seemed like a better idea.

sedghi:
Hey @Michael-Andersen 
So is the reason that we don't have RT* and SEG and REG viewports yet?

Michael-Andersen:
@sedghi Yes basically. Are viewports for these types on the roadmap?

To be honest, today I am having trouble recreating the bug. I don't know if there was some change that made this bug disappear along the way or if today our PACS is sending the study in a way that the series to render is always the CT and not and RT*. This was just a minor user experience improvement anyway so feel free to close or leave this PR as you see fit. I will likely leave this logic in our deployment because I don't see the harm (until maybe a viewport for these modalities exists).

sedghi:
Yeah, the Hanging Protocol pieces have changed recently to work on displaySets rather than series, and that might be something that is fixing this

sedghi:
re open if happened again

PR #2948, opened 2022-09-21
fix: sop class handler isSingleImageModality always undefined

Ouwen:
There is a typo in instance.modality where Modality should be uppercase. Otherwise will always be undefined. This is why Xray with same series are still stacking rather than having separate display sets.


Ouwen:
@sedghi, many thanks for your help in reviewing all these PRs so quickly :)

PR #2952, opened 2022-09-23
feat(hangingProtocol):Compatibility with older hanging protocol definitions

wayfarer3130:
This adds backwards compatibility to the hanging protocol mode definitions by adding a "setActiveProtocol(s)", which allows defining the protocols currently active, independent of the full set.  This gets called on mode enter.

### DIFF
--- a/platform/core/src/extensions/ExtensionManager.js
+++ b/platform/core/src/extensions/ExtensionManager.js
@@ -286,8 +286,17 @@ export default class ExtensionManager {
 
   _initHangingProtocolsModule = (extensionModule, extensionId) => {
     const { HangingProtocolService } = this._servicesManager.services;
-    extensionModule.forEach(({ id, protocol }) => {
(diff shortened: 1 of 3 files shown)


sedghi (commented):
I see what you are trying to do here, however, having both `protocol` and `protocols` is not ideal. How about we identify it ourselves ? having `protocol` but the type is `string[] | string` and we identify and apply correctly

wayfarer3130:
> I see what you are trying to do here, however, having both `protocol` and `protocols` is not ideal. How about we identify it ourselves ? having `protocol` but the type is `string[] | string` and we identify and apply correctly

Unfortunately you may well want to have both of them - I can see a situation where a default protocol is always applied initially, but then a menu of available protocols is shown.  Also, we need to keep it backwards compatible as much as possible, so I don't see an alternative to having both.

wayfarer3130:
This isn't going to be merged as backwards compatibility was deemed a not required feature.

PR #2971, opened 2022-10-04
fix(hp):Make the updated hanging protocol compatible with modes

wayfarer3130:
There had been a change made to make the hanging protocols auto initialize, but that caused problems on mode init because the couldn't specify the hanging protocols any longer. THis change restores that, and also improves the API a little bit by allowing the display set selectors to be specified by id instead of positionally.

### DIFF
--- a/extensions/default/src/getHangingProtocolModule.js
+++ b/extensions/default/src/getHangingProtocolModule.js
@@ -9,6 +9,15 @@ const defaultProtocol = {
   editableBy: {},
   protocolMatchingRules: [],
   toolGroupIds: ['default'],
+  displaySetSelectors: {
+    defaultDisplaySetId: {
+      // Unused currently
+      imageMatchingRules: [],
+      // Matches displaysets, NOT series
+      seriesMatchingRules: [],
+      studyMatchingRules: [],
+    },
+  },
   stages: [
     {
       id: 'hYbmMy3b7pz7GLiaT',
@@ -20,16 +29,6 @@ const defaultProtocol = {
           columns: 1,
         },
       },
-      displaySets: [
-        {
-          id: 'displaySet',
-          // Unused currently
-          imageMatchingRules: [],
-          // Matches displaysets, NOT series
-          seriesMatchingRules: [],
-          studyMatchingRules: [],
-        },
-      ],
       viewports: [
         {
           viewportOptions: {
@@ -42,7 +41,7 @@ const defaultProtocol = {
           displaySets: [
             {
               options: [],
-              id: 'displaySet',
+              id: 'defaultDisplaySetId',
             },
           ],
         },
--- a/extensions/tmtv/src/getHangingProtocolModule.js
+++ b/extensions/tmtv/src/getHangingProtocolModule.js
@@ -1,10 +1,10 @@
 const ptCT = {
-  id: 'ptCT',
+  id: '@ohif/extension-tmtv.hangingProtocolModule.ptCT',
   locked: true,
   hasUpdatedPriorsInformation: false,
   name: 'Default',
   createdDate: '2021-02-23T19:22:08.894Z',
-  modifiedDate: '2021-02-23T19:22:08.894Z',
+  modifiedDate: '2022-10-04T19:22:08.894Z',
   availableTo: {},
   editableBy: {},
   toolGroupIds: [
@@ -16,37 +16,79 @@ const ptCT = {
   imageLoadStrategy: 'interleaveTopToBottom', // "default" , "interleaveTopToBottom",  "interleaveCenter"
   protocolMatchingRules: [
     {
-      id: 'wauZK2QNEfDPwcAQo',
-      weight: 1,
       attribute: 'ModalitiesInStudy',
       constraint: {
         contains: ['CT', 'PT'],
       },
-      required: false,
     },
     {
-      id: 'wauZK2QNEfDPwcAQo',
-      weight: 1,
       attribute: 'StudyDescription',
       constraint: {
-        contains: {
-          value: 'PETCT',
-        },
+        contains: 'PETCT',
       },
-      required: false,
     },
     {
-      id: 'wauZK2QNEfDPwcAQo',
-      weight: 1,
       attribute: 'StudyDescription',
       constraint: {
-        contains: {
-          value: 'PET/CT',
-        },
+        contains: 'PET/CT',
       },
-      required: false,
     },
   ],
+  displaySetSelectors: {
+    ctDisplaySet: {
+      seriesMatchingRules: [
+        {
+          weight: 1,
+          attribute: 'Modality',
+          constraint: {
+            equals: {
+              value: 'CT',
+            },
+          },
+          required: true,
+        },
+        {
+          attribute: 'SeriesDescription',
+          constraint: {
+            contains: 'CT',
+          },
+        },
+        {
+          attribute: 'SeriesDescription',
+          constraint: {
+            contains: 'CT WB',
+          },
+        },
+      ],
+    },
+    ptDisplaySet: {
+      seriesMatchingRules: [
+        {
+          attribute: 'Modality',
+          constraint: {
+            equals: 'PT',
+          },
+          required: true,
+        },
+        {
+          attribute: 'SeriesDescription',
+          constraint: {
+            contains: 'Corrected',
+          },
+        },
+        {
+          weight: 2,
+          attribute: 'SeriesDescription',
+          constraint: {
+            doesNotContain: {
+              value: 'Uncorrected',
+            },
+          },
+        },
+      ],
+    },
+  },
+
   stages: [
     {
       id: 'hYbmMy3b7pz7GLiaT',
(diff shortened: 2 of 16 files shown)


wayfarer3130:
@sedghi  - I added some documentation in the typescript files to help users understand how to create hanging protocols.  We might want to also update the general documentation pages on this to explain how to match etc, but the main change here is to allow setting the hanging protocol according to the discussion we had, from "hangingProtocol" as a string or list item.

extensions/default/src/getHangingProtocolModule.js
```diff
@@ -9,6 +9,15 @@ const defaultProtocol = {
   editableBy: {},
   protocolMatchingRules: [],
   toolGroupIds: ['default'],
+  displaySetSelectors: {
+    displaySet: {
```
sedghi, line 13: can we say `deafultDisplaySetId` here so that someone don't confuse it with actual `DisplaySet` here?

extensions/tmtv/src/getHangingProtocolModule.js
```diff
@@ -16,37 +16,87 @@ const ptCT = {
...
     {
-      id: 'wauZK2QNEfDPwcAQo',
-      weight: 1,
       attribute: 'StudyDescription',
       constraint: {
-        contains: {
-          value: 'PET/CT',
-        },
+        contains: 'PET/CT',
       },
```
sedghi, line 34: why we are removing the `id` and `weigth` and `required`? I assume weigth and required have defaults of `1` and `false`, but what about `id`?

extensions/tmtv/src/getHangingProtocolModule.js
```diff
@@ -16,37 +16,87 @@ const ptCT = {
...
+            },
+          },
+          required: true,
+        },
+        {
+          attribute: 'SeriesNumber',
+          constraint: {
+            equals: '4',
+          },
+        },
```
sedghi, line 55: I wanted to remove this, but can you on this PR?

platform/core/src/extensions/ExtensionManager.js
```diff
@@ -286,8 +286,15 @@ export default class ExtensionManager {
 
   _initHangingProtocolsModule = (extensionModule, extensionId) => {
     const { HangingProtocolService } = this._servicesManager.services;
-    extensionModule.forEach(({ id, protocol }) => {
-      HangingProtocolService.addProtocol(id, protocol);
+    extensionModule.forEach(({ id, name, protocol }) => {
```
sedghi, line 289: I'm confused, our protocols don't have `name`, where is this coming from?

platform/core/src/extensions/ExtensionManager.js
```diff
@@ -286,8 +286,15 @@ export default class ExtensionManager {
...
+      if (protocol) {
+        const extensionName = name ? `.${name}` : '';
+        // Only auto-register if protocol specified, otherwise let mode register
+        HangingProtocolService.addProtocol(
+          id || `${extensionId}${extensionName}`,
```
sedghi, line 294: is there any scenario that `id` is undefined? why are we having `||`. 
IMO `id` can represent one or an array of protocols .

I think the registration here is wrong. What is `extensionId` and `extensionName`. This is confusing

platform/docs/docs/platform/services/data/HangingProtocolService.md
```diff
@@ -65,7 +70,13 @@ There are two events that get publish in `HangingProtocolService`:
 
 Default initialization of the modes handles running the `HangingProtocolService`
 
+## Hanging Protocol Instance Definition
+A hanging protocol has an id and/or a name as provided by the extension, and
```
sedghi, line 74: This is confusing and unnecessary to me, why we need to have `name` as well as `id`?

platform/docs/versioned_docs/version-3.0/platform/modes/index.md
```diff
@@ -77,7 +77,7 @@ function modeFactory() {
       },
     ],
     extensions: extensionDependencies,
-    hangingProtocols: [],
+    hangingProtocol: [],
```
sedghi, line 80: this file is in versioned_docs, we don't need to modify it

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -123,8 +125,9 @@ class HangingProtocolService {
     // this.protocols is a map of protocols with the protocol id as the key
     // and the protocol or a function that returns a protocol as the value
     const protocols = [];
+    const keys = this.activeProtocolIds || this.protocols.keys();
```
sedghi, line 128: this will always return `this.activeProtocolIds`, did you want to check `.length`?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -184,6 +187,54 @@ class HangingProtocolService {
...
+  public setActiveProtocol(id: string, protocol?: Protocol): void {
+    if (!id) {
+      throw new Error(`Protocol ${protocol} added without an id`);
+    }
+    console.log('Active protocol set', id, protocol);
+    if (protocol && this.protocols.get(id) !== protocol) {
+      this.addProtocol(id, protocol);
+    }
+    if (!this.activeProtocolIds) {
+      this.activeProtocolIds = [];
```
sedghi, line 203: Do you want to initiate this in the constructor instead of here?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -184,6 +187,54 @@ class HangingProtocolService {
     this.protocols.set(protocolId, protocol);
   }
 
+  /** Sets the given protocol object as active.  Also sets the default
+   * for the given id to the specified protocol - allows updating which
+   * protocol is active for a given name.
+   */
+  public setActiveProtocol(id: string, protocol?: Protocol): void {
```
sedghi, line 194: It is weird that `setActiveProtocol` is pushing (instead of setting) into the active protocolIds list. I expected a set not append

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -184,6 +187,54 @@ class HangingProtocolService {
...
+  ): void {
+    if (typeof hangingProtocol === 'string') {
+      this.setActiveProtocols(extensionManager, [hangingProtocol]);
+      return;
+    }
+    this.activeProtocolIds = null;
+    if (!hangingProtocol) {
+      console.log('No active protocols, setting all to active');
+      return;
+    }
```
sedghi, line 225: shouldn't this be in the top?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -184,6 +187,54 @@ class HangingProtocolService {
...
+    this.activeProtocolIds = [];
```
sedghi, line 226: You are setting `activeProtocolIds` to null then to [] 3 lines later, this is werid

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -184,6 +187,54 @@ class HangingProtocolService {
...
+    hangingProtocol.forEach(extensionProtocol => {
+      const hangingProtocolModule = extensionManager.getModuleEntry(
+        extensionProtocol
+      );
+      this.setActiveProtocol(
+        hangingProtocolModule?.id || extensionProtocol,
+        hangingProtocolModule?.protocol
+      );
```
sedghi, line 234: It should bail out before setting active Protocol if HPModule not found with the string

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -184,6 +187,54 @@ class HangingProtocolService {
...
+    if (!hangingProtocol) {
+      console.log('No active protocols, setting all to active');
+      return;
+    }
+    this.activeProtocolIds = [];
+    hangingProtocol.forEach(extensionProtocol => {
+      const hangingProtocolModule = extensionManager.getModuleEntry(
+        extensionProtocol
+      );
+      this.setActiveProtocol(
```
sedghi, line 231: I think here we should also check if the protocolModule.protocol is actually one protocol or an array of protocols, since it can be based on what we discussed here

https://gist.github.com/sedghi/c3f277f743690a6cb1e717319e29b0b1

sedghi (changes requested):
Thanks for the PR. Almost there, see my comments please

extensions/tmtv/src/getHangingProtocolModule.js
wayfarer3130: Why do you need an id?  It isn't used anywhere.  Every item is referred to positionally.
weight defaults to 1 and required to false.

extensions/tmtv/src/getHangingProtocolModule.js
wayfarer3130: Removed.

platform/core/src/extensions/ExtensionManager.js
wayfarer3130: Removed.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: No, because null is used to indicate no specified hanging protocols, whereas this is used to specify only the specific set.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: Will rename it.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: No, then it wouldn't work - this specifically is done after the null assignment to reset to the default state, which is not to have a specified active list, but to make anything registered active.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: Removed the extension manager setting entirely from the HP Service, so it is just based on the id

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -184,6 +187,44 @@ class HangingProtocolService {
...
+  /** Add a given protocol object as active.  Can optionally specify which
+   * implementation of a protocol id is the active one.  This is typically
+   * used to allow different choices for the "default" or "mpr" protocols.
+   */
+  public addActiveProtocol(id: string, protocol?: Protocol): void {
+    if (!id) {
+      throw new Error(`Protocol ${protocol} added without an id`);
+    }
+    console.log('Active protocol set', id, protocol);
+    if (protocol && this.protocols.get(id) !== protocol) {
```
sedghi, line 199: Shouldn't this be retrieved from the `this.getProtocolById` so that the dynamic ones can also work?

platform/docs/docs/platform/services/data/HangingProtocolService.md
```diff
@@ -65,7 +70,13 @@ There are two events that get publish in `HangingProtocolService`:
...
+then a single protocol instance.  The default id for a protocol is the extension
+id, plus the name if provided.  Otherwise, a specific id can be specified.  The
+'default' name is used as the hanging protocol id when no other protocol applies.
```
sedghi, line 77: can you update this? (we don't need names anymore)

sedghi (commented):
Where is the extensionManager getModule stuff? that was needed for pooling together the protocols if it is something like 

```
hangingProtocol: [
      '@extension1.hangingprotocolModule.hp1-1',
      '@extension2.hangingprotocolModule.hp2-2',
    ],
```

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: No, the point is one can reconfigure a specific default here - maybe I don't really care about combining that functionality.  Will remove the protocol entirely.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
wayfarer3130: No, activeProtocolIds can be null, which means to use everything.

sedghi (approved):
Great work! Thanks

PR #2973, opened 2022-10-04
Multi-frame DICOM support

md-prog:
OHIFv3 lacks the support of annotations and measurements on multi-frame DICOM files particularly when we want to persist and load Measurement SRs on multi-frame DICOM files.

There are several issues being resolved in this PR:
  - Supporting of annotations and measurements on multi-frame files
  - Saving Measurement SRs on multi-frame DICOM files
  - Fixed loading of frames of multi-frame DICOM files : it was not loading the last frame ( loading Frame 1 twice, and not loading the last frame) , as you can guess, it was related to indexing of frames. Frame numbers are 1-based, not 0-based.
  - Fix jumping to the selected measurement/annotation on multi-frame files
  - Show correct instance id and frame number on the "Measurement Tracking" panel
  - Fixed a few "React" errors that was causing error logs in the browser console.

External dependencies:
     This depends on the changes of `dcmjs` library:  https://github.com/dcmjs-org/dcmjs/pull/317
     The changes has been merged and available on version 0.28.1, therefore updated the package.json for dcmjs library.

### DIFF
--- a/extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
+++ b/extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
@@ -221,11 +221,23 @@ function _checkIfCanAddMeasurementsToDisplaySet(
     const image = images[i];
     const { SOPInstanceUID } = image;
     if (SOPInstanceUIDs.includes(SOPInstanceUID)) {
-      const imageId = imageIdsForDisplaySet[i];
-
       for (let j = unloadedMeasurements.length - 1; j >= 0; j--) {
         const measurement = unloadedMeasurements[j];
         if (_measurementReferencesSOPInstanceUID(measurement, SOPInstanceUID)) {
+          let imageId;
+          if (newDisplaySet.isMultiFrame) {
+            const frameNumber =
+              (measurement.coords[0].ReferencedSOPSequence &&
+                measurement.coords[0].ReferencedSOPSequence
+                  .ReferencedFrameNumber) ||
+              1;
+            imageId =
+              imageIdsForDisplaySet[frameNumber - 1] ||
+              imageIdsForDisplaySet[i];
+          } else {
+            imageId = imageIdsForDisplaySet[i];
+          }
+
           addMeasurement(
             measurement,
             imageId,
--- a/extensions/cornerstone-dicom-sr/src/utils/addMeasurement.ts
+++ b/extensions/cornerstone-dicom-sr/src/utils/addMeasurement.ts
@@ -45,6 +45,12 @@ export default function addMeasurement(
 
   const annotationManager = annotation.state.getDefaultAnnotationManager();
 
+  // Create Cornerstone3D Annotation from measurement
+  const frameNumber =
+    (measurement.coords[0].ReferencedSOPSequence &&
+      measurement.coords[0].ReferencedSOPSequence.ReferencedFrameNumber) ||
+    1;
+
   const SRAnnotation: Types.Annotation = {
     annotationUID: measurement.TrackingUniqueIdentifier,
     metadata: {
@@ -61,6 +67,7 @@ export default function addMeasurement(
         TrackingUniqueIdentifier: measurementData.TrackingUniqueIdentifier,
         renderableData: measurementData.renderableData,
       },
+      frameNumber: frameNumber,
     },
   };
 
@@ -75,6 +82,7 @@ export default function addMeasurement(
   // It'd be super weird if it didn't anyway as a SCOORD.
   measurement.ReferencedSOPInstanceUID =
     measurement.coords[0].ReferencedSOPSequence.ReferencedSOPInstanceUID;
+  measurement.frameNumber = frameNumber;
   delete measurement.coords;
 }
 
--- a/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
+++ b/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
@@ -237,7 +237,7 @@ const OHIFCornerstoneViewport = React.memo(props => {
         displaySets,
         viewportOptions.viewportType,
         dataSource,
-        (viewportDataLoaded) => {
+        viewportDataLoaded => {
           CornerstoneViewportService.setViewportDisplaySets(
             viewportIndex,
             viewportDataLoaded,
@@ -394,7 +394,7 @@ function _jumpToMeasurement(
   viewportGridService
 ) {
   const targetElement = targetElementRef.current;
-  const { displaySetInstanceUID, SOPInstanceUID } = measurement;
+  const { displaySetInstanceUID, SOPInstanceUID, frameNumber } = measurement;
 
   if (!SOPInstanceUID) {
     console.warn('cannot jump in a non-acquisition plane measurements yet');
@@ -404,9 +404,14 @@ function _jumpToMeasurement(
     displaySetInstanceUID
   );
 
-  const imageIdIndex = referencedDisplaySet.images.findIndex(
+  // For multi-frame files, finding imageID simply by SOPInstanceUID won't work,
+  // as the "referencedDisplaySet.images" will only contain the Frame 1.
+  let imageIdIndex = referencedDisplaySet.images.findIndex(
     i => i.SOPInstanceUID === SOPInstanceUID
   );
+  if (referencedDisplaySet.isMultiFrame && frameNumber > 1) {
+    imageIdIndex = frameNumber - 1;
+  }
 
   // Todo: setCornerstoneMeasurementActive should be handled by the toolGroupManager
   //  to set it properly
--- a/extensions/cornerstone/src/initMeasurementService.js
+++ b/extensions/cornerstone/src/initMeasurementService.js
@@ -3,6 +3,7 @@ import { Enums, annotation } from '@cornerstonejs/tools';
 import { DicomMetadataStore } from '@ohif/core';
 
 import measurementServiceMappingsFactory from './utils/measurementServiceMappings/measurementServiceMappingsFactory';
+import getSOPInstanceAttributes from './utils/measurementServiceMappings/utils/getSOPInstanceAttributes';
 
 const { removeAnnotation } = annotation.state;
 
@@ -252,7 +253,23 @@ const connectMeasurementServiceToTools = (
         SOPInstanceUID
       );
 
-      const imageId = dataSource.getImageIdsForInstance({ instance });
+      let frameNumber = 1;
+      if (measurement.data && measurement.data.frameNumber) {
+        frameNumber = measurement.data.frameNumber;
+      } else if (
+        measurement.metadata &&
+        measurement.metadata.referencedImageId
+      ) {
+        frameNumber = getSOPInstanceAttributes(
+          measurement.metadata.referencedImageId
+        ).frameNumber;
+      }
+
+      let imageId = dataSource.getImageIdsForInstance({
+        instance,
+        frame: frameNumber,
+      });
+
       const annotationManager = annotation.state.getDefaultAnnotationManager();
       annotationManager.addAnnotation({
         annotationUID: measurement.uid,
@@ -269,6 +286,7 @@ const connectMeasurementServiceToTools = (
           handles: { ...data.annotation.data.handles },
           cachedStats: { ...data.annotation.data.cachedStats },
           label: data.annotation.data.label,
+          frameNumber: frameNumber,
         },
       });
     }
--- a/extensions/cornerstone/src/utils/measurementServiceMappings/ArrowAnnotate.js
+++ b/extensions/cornerstone/src/utils/measurementServiceMappings/ArrowAnnotate.js
@@ -69,6 +69,7 @@ const Length = {
       metadata,
       referenceSeriesUID: SeriesInstanceUID,
       referenceStudyUID: StudyInstanceUID,
+      frameNumber: mappedAnnotations[0].frameNumber || 1,
       toolName: metadata.toolName,
       displaySetInstanceUID: displaySet.displaySetInstanceUID,
       label: data.text,
@@ -90,9 +91,11 @@ function getMappedAnnotations(annotation, DisplaySetService) {
 
   const annotations = [];
 
-  const { SOPInstanceUID, SeriesInstanceUID } = getSOPInstanceAttributes(
-    referencedImageId
-  );
+  const {
+    SOPInstanceUID,
+    SeriesInstanceUID,
+    frameNumber,
+  } = getSOPInstanceAttributes(referencedImageId);
 
   const displaySet = DisplaySetService.getDisplaySetForSOPInstanceUID(
     SOPInstanceUID,
@@ -105,6 +108,7 @@ function getMappedAnnotations(annotation, DisplaySetService) {
     SeriesInstanceUID,
     SOPInstanceUID,
     SeriesNumber,
+    frameNumber,
     text,
   });
 
@@ -119,7 +123,7 @@ function getDisplayText(mappedAnnotations, displaySet) {
   const displayText = [];
 
   // Area is the same for all series
-  const { SeriesNumber, SOPInstanceUID } = mappedAnnotations[0];
+  const { SeriesNumber, SOPInstanceUID, frameNumber } = mappedAnnotations[0];
 
   const instance = displaySet.images.find(
     image => image.SOPInstanceUID === SOPInstanceUID
@@ -130,11 +134,10 @@ function getDisplayText(mappedAnnotations, displaySet) {
     InstanceNumber = instance.InstanceNumber;
   }
 
-  displayText.push(
-    InstanceNumber
-      ? `(S: ${SeriesNumber} I: ${InstanceNumber})`
-      : `(S: ${SeriesNumber})`
-  );
+  const instanceText = InstanceNumber ? ` I: ${InstanceNumber}` : '';
+  const frameText = displaySet.isMultiFrame ? ` F: ${frameNumber}` : '';
+
+  displayText.push(`(S: ${SeriesNumber}${instanceText}${frameText})`);
 
   return displayText;
 }
--- a/extensions/cornerstone/src/utils/measurementServiceMappings/Bidirectional.js
+++ b/extensions/cornerstone/src/utils/measurementServiceMappings/Bidirectional.js
@@ -67,6 +67,7 @@ const Bidirectional = {
       metadata,
       referenceSeriesUID: SeriesInstanceUID,
       referenceStudyUID: StudyInstanceUID,
+      frameNumber: mappedAnnotations[0].frameNumber || 1,
       toolName: metadata.toolName,
       displaySetInstanceUID: displaySet.displaySetInstanceUID,
       label: data.label,
@@ -92,30 +93,32 @@ function getMappedAnnotations(annotation, DisplaySetService) {
   Object.keys(cachedStats).forEach(targetId => {
     const targetStats = cachedStats[targetId];
 
-    let displaySet;
-
-    if (referencedImageId) {
-      const { SOPInstanceUID, SeriesInstanceUID } = getSOPInstanceAttributes(
-        referencedImageId
-      );
-
-      displaySet = DisplaySetService.getDisplaySetForSOPInstanceUID(
-        SOPInstanceUID,
-        SeriesInstanceUID
-      );
-    } else {
+    if (!referencedImageId) {
       throw new Error(
         'Non-acquisition plane measurement mapping not supported'
       );
     }
 
-    const { SeriesNumber, SeriesInstanceUID } = displaySet;
+    const {
+      SOPInstanceUID,
+      SeriesInstanceUID,
+      frameNumber,
+    } = getSOPInstanceAttributes(referencedImageId);
+
+    const displaySet = DisplaySetService.getDisplaySetForSOPInstanceUID(
+      SOPInstanceUID,
+      SeriesInstanceUID
+    );
+
+    const { SeriesNumber } = displaySet;
     const { length, width } = targetStats;
     const unit = 'mm';
 
     annotations.push({
       SeriesInstanceUID,
+      SOPInstanceUID,
       SeriesNumber,
+      frameNumber,
       unit,
       length,
       width,
@@ -171,7 +174,13 @@ function getDisplayText(mappedAnnotations, displaySet) {
   const displayText = [];
 
   // Area is the same for all series
-  const { length, width, SeriesNumber, SOPInstanceUID } = mappedAnnotations[0];
+  const {
+    length,
+    width,
+    SeriesNumber,
+    SOPInstanceUID,
+    frameNumber,
+  } = mappedAnnotations[0];
   const roundedLength = utils.roundNumber(length, 2);
   const roundedWidth = utils.roundNumber(width, 2);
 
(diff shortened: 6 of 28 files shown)


extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -221,11 +221,23 @@ function _checkIfCanAddMeasurementsToDisplaySet(
     const image = images[i];
     const { SOPInstanceUID } = image;
     if (SOPInstanceUIDs.includes(SOPInstanceUID)) {
-      const imageId = imageIdsForDisplaySet[i];
-
       for (let j = unloadedMeasurements.length - 1; j >= 0; j--) {
         const measurement = unloadedMeasurements[j];
         if (_measurementReferencesSOPInstanceUID(measurement, SOPInstanceUID)) {
```
wayfarer3130, line 226: I think I would include the frame number extracted from the imageId here, as _measurementReferencesSOPInstanceUID(measurement, SOPInstanceUID, frame#), as the code below ends up checking for some things that may not be true.  Specifically:
1. ReferencedFrameNumber may be an array of values
2. Frames do not need to be in order, nor all in the same display set.

The original logic, just say does a measurement reference this image id should work for all of them, as long as you extract both the sop instance uid and the frame#

wayfarer3130: This needs to be done by using the MetadataProvider to call the _getUIDsFromImageId (you will have to make it public), and then do the comparison on that returned data, as there is no guarantee that a single frame series doesn't have multiframe data in it, or that a multiframe series doesn't have two or more sop instances in it (they can get broken up for very large data sets).

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -404,9 +404,14 @@ function _jumpToMeasurement(
     displaySetInstanceUID
   );
 
-  const imageIdIndex = referencedDisplaySet.images.findIndex(
```
wayfarer3130, line 407: Revert to the const imageIdIndex, but iterate over the imageIds instead of the images, and use the MetadataProvider to get the sop uid and frame# for the given image id.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -221,11 +221,23 @@ function _checkIfCanAddMeasurementsToDisplaySet(
     const image = images[i];
     const { SOPInstanceUID } = image;
     if (SOPInstanceUIDs.includes(SOPInstanceUID)) {
-      const imageId = imageIdsForDisplaySet[i];
```
wayfarer3130, line 224: Discussed that we will iterate through the imageIds, and use the MetadataProvider.getUIDS... method to get the SOP + frame#

extensions/cornerstone/src/initMeasurementService.js
```diff
@@ -252,7 +253,23 @@ const connectMeasurementServiceToTools = (
         SOPInstanceUID
       );
 
-      const imageId = dataSource.getImageIdsForInstance({ instance });
```
wayfarer3130, line 255: revert the imageId and get the UIDs from it.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
md-prog: Fixed as discussed in https://github.com/OHIF/Viewers/pull/2973/commits/3ee38072d4309c1524a22a4da94bc41ee7bf4bd8 and https://github.com/OHIF/Viewers/pull/2973/commits/75b4cfcef27b972745ba2e3ea3f9546d5f8cf27d

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
md-prog: Fixed it in https://github.com/OHIF/Viewers/pull/2973/commits/75b4cfcef27b972745ba2e3ea3f9546d5f8cf27d

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
md-prog: Fixed in https://github.com/OHIF/Viewers/pull/2973/commits/75b4cfcef27b972745ba2e3ea3f9546d5f8cf27d

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -1,5 +1,6 @@
 import { SOPClassHandlerName, SOPClassHandlerId } from './id';
 import { utils, classes } from '@ohif/core';
+import metadataProvider from '@ohif/core/src/classes/MetadataProvider';
```
wayfarer3130, line 3: You should get the metadataProvider from either @ohif/core, in the line above, or from the init method in the extension.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -212,20 +213,27 @@ function _checkIfCanAddMeasurementsToDisplaySet(
     newDisplaySet
   );
 
-  for (let i = 0; i < images.length; i++) {
+  for (let i = 0; i < imageIdsForDisplaySet.length; i++) {
```
wayfarer3130, line 216: for(const imageId of imageIdsForDisplaySet) 
as that doesn't assume anything about the layout/design and just gets next until done.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -239,9 +247,23 @@ function _checkIfCanAddMeasurementsToDisplaySet(
...
+  SOPInstanceUID,
+  frameNumber
+) {
   const { coords } = measurement;
 
+  // NOTE: The ReferencedFrameNumber can be multiple values according to the DICOM
+  //  Standard. But for now, we will support only one ReferenceFrameNumber.
+  const ReferencedFrameNumber =
+    (measurement.coords[0].ReferencedSOPSequence &&
+      measurement.coords[0].ReferencedSOPSequence.ReferencedFrameNumber) ||
```
wayfarer3130, line 261: You should use ReferencedSOPSequence[0] - it is an array that happens to have attributes of child zero when of length 1,  but you shouldn't count on that.

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -404,17 +409,32 @@ function _jumpToMeasurement(
...
 
-  if (getEnabledElement(targetElement)) {
+  const enableElement = getEnabledElement(targetElement);
+  if (enableElement) {
+    // See how the jumpToSlice() of Cornerstone3D deals with imageIdx param.
+    const viewport = enableElement.viewport as IStackViewport | IVolumeViewport;
+    const { imageIds } = viewport;
+    let imageIdIndex;
+
+    for (let i = 0; i < imageIds.length; i++) {
```
wayfarer3130, line 425: You should use findIndex again, rather than implementing findIndex manually.

extensions/cornerstone/src/utils/measurementServiceMappings/Length.js
```diff
@@ -97,30 +98,32 @@ function getMappedAnnotations(annotation, DisplaySetService) {
...
     }
 
-    const { SeriesNumber, SeriesInstanceUID } = displaySet;
+    const {
+      SOPInstanceUID,
+      SeriesInstanceUID,
+      frameNumber,
+    } = getSOPInstanceAttributes(referencedImageId);
+
+    const displaySet = DisplaySetService.getDisplaySetForSOPInstanceUID(
```
wayfarer3130, line 113: This call had better take an optional frame number - it can be ignored for now, as we don't support multiframe splits, but there are some really important multiframe splits that I'm aware of.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -494,7 +494,7 @@ function createDicomWebApi(dicomWebConfig, UserAuthenticationService) {
         const NumberOfFrames = instance.NumberOfFrames;
 
         if (NumberOfFrames > 1) {
-          for (let i = 0; i < NumberOfFrames; i++) {
+          for (let i = 1; i <= NumberOfFrames; i++) {
```
wayfarer3130, line 497: If this is frame, call it frame in the for loop.

wayfarer3130:
@md-prog - a few small changes, then it looks good to me.
@sedghi - this fixes some important issues with multiframe and annotations.  It does not fix volume annotations yet - that is something we still need to discuss how to handle properly.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
sedghi: apparently your mention of the `@ohif/ core` has sent notification to all people in the core, this is funny :D

extensions/default/src/DicomWebDataSource/index.js
md-prog: resolved in commit ea05ac7db90672acc420f7c264bf04cd17af885a

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
md-prog: now importing metadataProvider from OHIF.classes.MetadataProvider.
Commit 1469ba76b4b2755f50f75b68048798183963a3d8

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
md-prog: Resolved in 9d380ba8e72953e0e76d3fa1212b015973093fa2

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
md-prog: Done in 52b4f8b6b14a00b6d85890dd3e7aea1580b1581c

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
md-prog: fixed in c48cbff4fd68e6848196fec07d0b49b7bffe0fd8

extensions/cornerstone/src/utils/measurementServiceMappings/Length.js
md-prog: Done in a979c66c6716974ebd4125e80ce331245039253a

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -404,17 +409,29 @@ function _jumpToMeasurement(
...
   // setCornerstoneMeasurementActive(measurement);
 
   viewportGridService.setActiveViewportIndex(viewportIndex);
 
-  if (getEnabledElement(targetElement)) {
+  const enableElement = getEnabledElement(targetElement);
+  if (enableElement) {
+    // See how the jumpToSlice() of Cornerstone3D deals with imageIdx param.
+    const viewport = enableElement.viewport as IStackViewport | IVolumeViewport;
+    const { imageIds } = viewport;
```
sedghi, line 422: imageIds is a private property, you need to access it with `get`

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -404,17 +409,29 @@ function _jumpToMeasurement(
...
+
+    const imageIdIndex = imageIds.findIndex(imageId => {
+      const {
+        SOPInstanceUID: SOPInstanceUID1,
+        frameNumber: frameNumber1,
+      } = getSOPInstanceAttributes(imageId);
+      return (
+        SOPInstanceUID1 === SOPInstanceUID &&
```
sedghi, line 430: I'm not a fan of appending 1 to the end of these variables, I prefer `aSOPInstanceUID` and `aFrameNumber`

Also use `===` instead of `==`

extensions/cornerstone/src/utils/measurementServiceMappings/ArrowAnnotate.js
```diff
@@ -69,6 +69,7 @@ const Length = {
       metadata,
       referenceSeriesUID: SeriesInstanceUID,
       referenceStudyUID: StudyInstanceUID,
+      frameNumber: mappedAnnotations[0].frameNumber || 1,
```
sedghi, line 72: can't `mappedAnnotations` be undefined or `[]` which makes `mappedAnnotations[0]` undefined? so how about `mappedAnnotaions?[0]?.frameNumber`?

extensions/cornerstone/src/utils/measurementServiceMappings/ArrowAnnotate.js
```diff
@@ -130,11 +135,10 @@ function getDisplayText(mappedAnnotations, displaySet) {
...
 
-  displayText.push(
-    InstanceNumber
-      ? `(S: ${SeriesNumber} I: ${InstanceNumber})`
-      : `(S: ${SeriesNumber})`
-  );
+  const instanceText = InstanceNumber ? ` I: ${InstanceNumber}` : '';
+  const frameText = displaySet.isMultiFrame ? ` F: ${frameNumber}` : '';
+
+  displayText.push(`(S: ${SeriesNumber}${instanceText}${frameText})`);
```
sedghi, line 141: This is for displayText so it will not get saved anywhere, but does the user care about the fact that it is multiframe (so is it necessary to have F:...)? I understand that InstanceNumber will be 1 all the time for multiframe.

cc @wayfarer3130

extensions/cornerstone/src/utils/measurementServiceMappings/Bidirectional.js
```diff
@@ -67,6 +67,7 @@ const Bidirectional = {
       metadata,
       referenceSeriesUID: SeriesInstanceUID,
       referenceStudyUID: StudyInstanceUID,
+      frameNumber: mappedAnnotations[0].frameNumber || 1,
```
sedghi, line 70: same

extensions/cornerstone/src/initMeasurementService.js
```diff
@@ -252,7 +253,24 @@ const connectMeasurementServiceToTools = (
...
+      let imageId;
+      let frameNumber = 1;
+
+      if (measurement.metadata && measurement.metadata.referencedImageId) {
```
sedghi, line 259: this can be simiplified to `measurement?.metadata?.referencedImageId`

extensions/cornerstone/src/initMeasurementService.js
```diff
@@ -252,7 +253,24 @@ const connectMeasurementServiceToTools = (
...
+        imageId = measurement.metadata.referencedImageId;
+        frameNumber = getSOPInstanceAttributes(
+          measurement.metadata.referencedImageId
+        ).frameNumber;
+      } else if (measurement.data && measurement.data.frameNumber) {
```
sedghi, line 264: `frameNumber` does not get stored in the `data` ever, why are you checking it here?

extensions/cornerstone-dicom-sr/src/utils/addMeasurement.ts
```diff
@@ -45,6 +45,13 @@ export default function addMeasurement(
 
   const annotationManager = annotation.state.getDefaultAnnotationManager();
 
+  // Create Cornerstone3D Annotation from measurement
+  const frameNumber =
+    (measurement.coords[0].ReferencedSOPSequence &&
+      measurement.coords[0].ReferencedSOPSequence[0] &&
+      measurement.coords[0].ReferencedSOPSequence[0].ReferencedFrameNumber) ||
```
sedghi, line 52: use optional chanining https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

extensions/cornerstone-dicom-sr/src/utils/addMeasurement.ts
```diff
@@ -61,6 +68,7 @@ export default function addMeasurement(
         TrackingUniqueIdentifier: measurementData.TrackingUniqueIdentifier,
         renderableData: measurementData.renderableData,
       },
+      frameNumber: frameNumber,
```
sedghi, line 71: we should not stick the `frameNumber` here, it is not a `data` per se. can't we just infer it from the referencedImageId? seeing if it has `frames/...` at the end?

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -212,20 +212,26 @@ function _checkIfCanAddMeasurementsToDisplaySet(
     newDisplaySet
   );
 
-  for (let i = 0; i < images.length; i++) {
+  for (const imageId of imageIdsForDisplaySet) {
     if (!unloadedMeasurements.length) {
       // All measurements loaded.
-      break;
+      return;
```
sedghi, line 218: can't we move this `if (!unloadedMeasurements.length) {` outside?

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
```diff
@@ -239,9 +245,24 @@ function _checkIfCanAddMeasurementsToDisplaySet(
...
+  frameNumber
+) {
   const { coords } = measurement;
 
+  // NOTE: The ReferencedFrameNumber can be multiple values according to the DICOM
+  //  Standard. But for now, we will support only one ReferenceFrameNumber.
+  const ReferencedFrameNumber =
+    (measurement.coords[0].ReferencedSOPSequence &&
+      measurement.coords[0].ReferencedSOPSequence[0] &&
+      measurement.coords[0].ReferencedSOPSequence[0].ReferencedFrameNumber) ||
```
sedghi, line 260: same

platform/core/src/services/DisplaySetService/DisplaySetService.js
```diff
@@ -75,7 +75,11 @@ export default class DisplaySetService {
     );
   };
 
-  getDisplaySetForSOPInstanceUID(SOPInstanceUID, SeriesInstanceUID) {
+  getDisplaySetForSOPInstanceUID(
+    SOPInstanceUID,
+    SeriesInstanceUID,
+    frameNumber
```
sedghi, line 81: frameNumber is not used

sedghi (changes requested):
Thanks @md-prog this is a great PR.

I put couple of comments above. There are couple of other issues 

1. When I try it locally it is not able to hydrate the SR for multiframe. It can draw and jump, but saving and hydration just puts everything ins the first slice (instance). 
To reproduce: open a multiframe study (I opened the brain MRI you gave me), draw length on two separate slices. Save report (to generate SR). 1) At this time the cornerstone viewport is unmounted and SR viewport mounts, which uses the SRDisplayTool (not hydrated yet), and as you see below it does not work 2) You can click on the -> to hydrate (or alternatively you can refresh the page and drag and drop SR and -> to hydrate), and that does not work either (don't have screenshot for it (but the same issue)

image

2. This might not be the issue in your PR but noticed the modalities are coming back as OT (Other) for some of the multiframes. Also the number of instances it says 1 (which is technically the case), but I guess we can show the number of frames here alternatively for the multiframe 

image

extensions/cornerstone/src/utils/measurementServiceMappings/ArrowAnnotate.js
md-prog: This could be the scenario:
When there are multiple annotations on different frames, and we display all those annotations on the right side Measurement Tracking panel. And when the user click on one of the annotations, it should jump the correct frame.
So displaying correct frame number would be great, I believe.

md-prog:
@sedghi, the issue No.1 mentioned in your last comment is related to dcmjs library's "lack of features" (if not defect).

https://github.com/dcmjs-org/dcmjs/issues/318

Basically, the lines of dcmjs library mentioned in above Github issue, are blocking us to save the annotations made on to the multi-frame file with the correct frame number information.

With this issue resolved, the issue should go. (During my development, I had to temporarily use the locally cloned dcmjs with the temporary fix to the issue.)

No.2 is not a new thing, I think.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
md-prog: No. it can't.
`unloadedMeasurements` is being modified within the loop. (Line 241)

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
md-prog: done in 471e9c6a294d746a6b81bc6c92e19233c677c6aa

extensions/cornerstone/src/utils/measurementServiceMappings/ArrowAnnotate.js
md-prog: done in fed74aae7ada884c3563b114ed8bc46938e325b3

extensions/cornerstone/src/utils/measurementServiceMappings/Bidirectional.js
md-prog: doen in fed74aae7ada884c3563b114ed8bc46938e325b3

extensions/cornerstone/src/initMeasurementService.js
md-prog: fixed in 9348a17652a77155bf7cd129fc17f2b4993dc785

extensions/cornerstone-dicom-sr/src/utils/addMeasurement.ts
md-prog: fixed in a9d22efe494178b6cbbdaaf971700f4b9e1bf517

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.js
md-prog: fixed in a9d22efe494178b6cbbdaaf971700f4b9e1bf517

extensions/cornerstone/src/initMeasurementService.js
md-prog: fixed in a9d22efe494178b6cbbdaaf971700f4b9e1bf517

extensions/cornerstone-dicom-sr/src/utils/addMeasurement.ts
md-prog: This is to conform with the changes we've made to dcmjs - cornerstone3D adapter: https://github.com/dcmjs-org/dcmjs/pull/317

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
md-prog: Fixed in 566ba8c72a8b6081d638b003bfd34383d6ef5f74

PR #2975, opened 2022-10-07
Bring v2 DicomTagBrowser to v3

Michael-Andersen:
Working with the v2 version of DicomTagBrowser, I made the necessary changes to make it work with v3. I've added it into the basic viewer mode under the "More Tools" dropdown. Currently I use the link-extension yarn cli command to link the folder but I guess we need to get this added into NPM so that it can added with yarn install? There's also an open issue with the styling to make it fit into v3 UI better.


Michael-Andersen:
@sedghi Tagging you for next steps with this. Feel free to send me style changes you'd like or any other changes you want.

sedghi:
Ok, I see that you are adding it as an extension. I think it is small enough to be just a command in the default extension commandsModule. So clicking on the button will invoke that command.

Also our design expert (Dan Rukas) has gave it a shot with the tag browser. You can see it below. It is basically using the components that we have already. However, I understand if you don't want to or don't have time to follow it exactly. My minimum requirement are 
1) DropDown and slider can be just in one row (no need to have two rows)
2) DropDown colors should match the design below (I think if you use the new dropdown it comes with it)

Optional:
- Good to have the table re designed but optional

Don't Do:
- Slider (I will take care of that)

image

Michael-Andersen:
@sedghi I've refactored to use the default extension. I've worked on the style as well. I'm using the standard OHIF dropdown and I've put the instance selector on the same line like you requested (I chose the header Series over Study, can change if you want). I left the slider alone. I have tried to make the table more like the StudyList. It didn't seem like there was any table component I could use wholesale so it was more a matter of comparing the styling and trying to duplicate. Let me know what you think. If there still some tweaks you'd like based on the new design let me know. I have also rebased this branch on top of v3-stable fyi.

Issue #2964, opened 2022-09-29
IDC: change message for segmentation when it fails to load because of orientation to


Punzo:
Change message for segmentation when it fails to load because of orientation tol (point user to settings in seg panel)

Punzo:
All relevant links on seg tol (logic, UI, etc.)

https://github.com/OHIF/Viewers/issues/2918

https://github.com/dcmjs-org/dcmjs/pull/304/files

generateToolState https://github.com/dcmjs-org/dcmjs/blob/master/src/adapters/Cornerstone/Segmentation_4X.js#L274

https://github.com/OHIF/Viewers/blob/e0fb439e988ad653f9af42b7424fc30bb2c0d279/extensions/dicom-segmentation/src/getOHIFDicomSegSopClassHandler.js#L160

https://github.com/OHIF/Viewers/blob/e0fb439e988ad653f9af42b7424fc30bb2c0d279/extensions/dicom-segmentation/src/getOHIFDicomSegSopClassHandler.js#L65

https://github.com/OHIF/Viewers/blob/e0fb439e988ad653f9af42b7424fc30bb2c0d279/extensions/dicom-segmentation/src/components/SegmentationSettings/SegmentationSettings.js#L139

Change message that you can see in video in https://github.com/OHIF/Viewers/pull/2924

PR #2981, opened 2022-10-14
Fix 2964

GitanjaliChhetri:
Fixes https://github.com/OHIF/Viewers/issues/2964

Fix summary:
Update the error message for segmentation load failure due to orientation tolerance value.
Direct user to change the tolerance value under Segmentation settings.


GitanjaliChhetri:
Discarding this in favor of https://github.com/OHIF/Viewers/pull/2982

Issue #2869, opened 2022-08-12
Segmentations converted from nii to dcm not rendered in OHIF Viewer


ccosmin97:
Hello,

I am having issues rendering dcm segmentations in OHIF Viewer(version number : 4.12.30).
The segmentations come from [ProstateX IDC collection](https://portal.imaging.datacommons.cancer.gov/explore/filters/?collection_id=prostatex), converted to nii using dcmqi, then converted back to dcm using dcmqi(v 1.2.5) again.

The dcm file originating from IDC buckets is rendered properly into OHIF, however the dcm file from dcmqi conversion from nii format is not rendered properly.

Here is the error reported by the OHIF Viewer :
![error_dcm_seg_ohif (1)](https://user-images.githubusercontent.com/72577931/184412717-a8b74022-3cd5-4ba8-bfec-5b2f88d6204e.jpg)

![ohif_console (1)](https://user-images.githubusercontent.com/72577931/184412763-39d38b8c-8fd3-42c4-9e32-70f205ec2af7.jpg)

To reproduce this issue, please see this [ipynb](https://github.com/ccosmin97/IDC-Prostate_segmentation/blob/main/prostateX_patID0004_dcmqi_test.ipynb).

Additional info :

- IDC bucket dcm
  -  [metadata](https://www.dropbox.com/s/6ewb178mjujzidz/1.2.276.0.7230010.3.1.4.1070885483.15960.1599120307.702_dcm_stats?dl=0) 
  - [dcm file](https://www.dropbox.com/s/2fxtsd2blwwcop2/1.2.276.0.7230010.3.1.4.1070885483.15960.1599120307.702.dcm?dl=0)
- DCMQI converted dcm(nii -> dcm)
  - [metadata](https://www.dropbox.com/s/l1rdu2315aks3xt/test_gt_0004_dcm_stats?dl=0)  
  - [dcm file](https://www.dropbox.com/s/xakvc4o4zg03ln2/test_gt_0004.dcm?dl=0)
- [Diff between two dcm metadata files](https://www.dropbox.com/s/9jszt3aa7ao5m95/diff-dcm-metadata?dl=0)

Thank you!

@fedorov

pieper:
Thanks for the report.  I think it's related this code: https://github.com/dcmjs-org/dcmjs/pull/220 but with your new data the error is larger than the (somewhat arbitrary) tolerance doesn't account for the differences due to numerical issues in your round trip.

[...]

PR #2982, opened 2022-10-14
fix: #2964 Update message for segmentation error loading due to orien…

GitanjaliChhetri:
Fixes https://github.com/OHIF/Viewers/issues/2964

Fix summary:
Update the error message for segmentation load failure due to orientation tolerance value.
Direct user to change the tolerance value under Segmentation settings.

segtol

### DIFF
--- a/extensions/dicom-segmentation/src/index.js
+++ b/extensions/dicom-segmentation/src/index.js
@@ -33,10 +33,17 @@ export default {
     const ExtendedSegmentationPanel = props => {
       const { activeContexts } = api.hooks.useAppContext();
       const onDisplaySetLoadFailureHandler = error => {
-        LoggerService.error({ error, message: error.message });
+        const message =
+          error.message.includes('orthogonal') ||
+          error.message.includes('oblique')
+            ? 'The segmentation has been detected as not planar,\
+    If you really think it is planar,\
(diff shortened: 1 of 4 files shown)


igoroctaviano (changes requested):
Hi @GitanjaliChhetri. Thanks for your contribution. 
The error message LGTM. Before moving forward I have some questions:

1. Do we want to display this information next to the thumbnail as an inconsistency warning as well? (when you hover the thumbnail warning icon)
2. I think we can add this error message only in the dicom-segmentation extension since this is a segmentation-specific error message, right? there's any other case why this would not be enough?

GitanjaliChhetri:
Thanks, @igoroctaviano !

1. We already display the inconsistency warning in the thumbnail. The warning message in the thumbnail is the original message from dcmjs. Please refer the picture.
image

2. The modified message aims at redirecting the user to segmentation panel settings in order to support viewing of segmentations that are detected as non-planar due to the 32 bit float rounding error (for example, for segmentations converted from nii  to dcm, original issue here https://github.com/OHIF/Viewers/issues/2869#issuecomment-1228331103) .
Since dcmjs is an external library, I think we are good keeping this modified message containing information about the UI of the upper level App within OHIF.

igoroctaviano (approved):
LGTM

PR #2984, opened 2022-10-17
feat(UICustomizationService):Add a customization service

wayfarer3130:
This is just the UI Customization Service, without any use of said service.   See the documentation for details, as well as related PR's for examples of how this is used.


platform/docs/docs/platform/managers/service.md
```diff
@@ -86,9 +87,16 @@ export default {
...
-> Note, the create method is critical for any custom service that you write and
+> Note: The create method is critical for any custom service that you write and
 > want to add to the list of services
 
+> Note: For typescript definitions, the service type should be exported
+> as part of the Types export on the module.  This is recommended going forward
+> and existing services will be migrated.  As well, the capitalization of the
+> name should be lower camel case, with the type being upper camel case.  In
+> the above example, the service instance should be `toolBarService` with the
+> class being `ToolBarService`.
```
sedghi, line 98: Are you proposing to change all the services from capitalized to lowercase? why?

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,196 @@
+---
+sidebar_position: 3
+sidebar_label: UI Modal Service
+---
+# UI Customization Service
+
+There are a lot of places where users may want to configure certain elements
```
sedghi, line 7: Great Introduction!

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,196 @@
...
+  },
+]);
+```
+
+Then, in the configuration file one might have a custom overlay definition:
+
+```js
+// in the APP_CONFIG file set the top right area to show the patient name
+// using PN: as a prefix when the study has a non-empty patient name.
+whiteLabeling: {
```
sedghi, line 62: Above, you explained by contextMenu as default registration, and here you are switching to the overlay top right. This will be better recieved if above also you show the default reigstration for the topRight.

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,196 @@
...
+  cornerstoneOverlayTopRight: {
+    id: 'cornerstoneOverlayTopRight',
+    items: [
+      {
+        id: 'PatientNameOverlay',
+        // Note the overlayItem as a parent type - this provides the
+        // rendering functionality to read the attribute and use the label.
+        customizationType: 'ohif.overlayItem',
```
sedghi, line 70: We have not talked about what is `customizationType` yet, and also `items` either. Maybe have a section above that each itesm will have its own properties (I think that is the case right?)

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,196 @@
...
+```js
+// in the APP_CONFIG file set the top right area to show the patient name
+// using PN: as a prefix when the study has a non-empty patient name.
+whiteLabeling: {
+  cornerstoneOverlayTopRight: {
+    id: 'cornerstoneOverlayTopRight',
+    items: [
+      {
+        id: 'PatientNameOverlay',
+        // Note the overlayItem as a parent type - this provides the
```
sedghi, line 68: Which parent Type?

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,196 @@
...
+    ],
+  }
+},
+```
+
+In the mode customization, the context menu is then customized:
+
+```js
+// First in ExtensionManager, the mode customizations are cleared by calling:
+UICustomizationService.onModeEnter();
```
sedghi, line 83: This is a redundant information IMO, user does not need to know that this is getting called in the extension manager. They only need to know that the following code will run on a mode enter

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,196 @@
...
+The return value then inherits the inheritted type value, so that the value can
+be typed and have default values and functionality provided.  The object
+can then be used in a way defined by the extension provided that customization
+point.
+
+```ts
+   cornerstoneOverlay = uiConfigurationService.getModeCustomization("cornerstoneOverlay", {customizationType: "ohif.cornerstoneOverlay", ...});
+   const { component: overlayComponent, props} = uiConfigurationService.getComponent(cornerstoneOverlay);
+   return (<defaultComponent {...props} overlay={cornerstoneOverlay}....></defaultComponent>);
+```
```
sedghi, line 118: If i'm not mistaken this is kind of behind the scene of UICustomizationService and how it works right? I don't think we need to explain that. Rather we need to show examples on how to configure stuff.

platform/core/src/extensions/MODULE_TYPES.js
```diff
@@ -1,5 +1,6 @@
 export default {
   COMMANDS: 'commandsModule',
+  CUSTOMIZATION: 'customizationModule',
```
sedghi, line 3: will we ever have other CUSTOMIZATIONS? if so this should be UI_CUSTOMIZATION

platform/core/src/extensions/ExtensionManager.js
```diff
@@ -40,13 +40,12 @@ export default class ExtensionManager {
...
 
-    const {
-      MeasurementService,
-      ViewportGridService,
-    } = _servicesManager.services;
+    const { MeasurementService } = _servicesManager.services;
 
+    for (const service of Object.values(_servicesManager.services)) {
+      service?.onModeEnter?.();
+    }
```
sedghi, line 47: To me a Service should not know anything about what is a `mode`. I understand it should clean up itself or init itself. But one can do that on a mode config `onModeEnter` and `onModeExit` like here  https://github.com/OHIF/Viewers/blob/v3-stable/modes/tmtv/src/index.js#L137-L150

platform/core/src/extensions/ExtensionManager.js
```diff
@@ -278,6 +278,7 @@ export default class ExtensionManager {
 
       return extensionModule;
     } catch (ex) {
+      console.log('Caught', ex);
```
sedghi, line 281: What is the purpose of this .log if the next line is thorwing error

platform/core/src/services/ToolBarService/ToolBarService.ts
```diff
@@ -52,19 +52,32 @@ export default class ToolBarService {
     this.buttons = {};
   }
 
+  onModeEnter() {
+    this.reset();
+  }
```
sedghi, line 57: The same as the other comment, this reset can be called on a modeEnter in the mode configuration factory. Having a `onModeEnter` on a Service is super weird to me

platform/core/src/services/ToolBarService/ToolBarService.ts
```diff
@@ -52,19 +52,32 @@ export default class ToolBarService {
...
+
   /**
    *
-   * @param {*} interaction
+   * @param {*} interaction - can be undefined to run nothing
+   * @param {*} extraOptions to include in the commands run
    */
-  recordInteraction(interaction) {
+  recordInteraction(interaction, extraOptions) {
```
sedghi, line 64: can we call this options? instead of extraoptions?

platform/core/src/services/_shared/pubSubServiceInterface.js
```diff
@@ -86,3 +87,21 @@ function _broadcastEvent(eventName, callbackProps) {
...
+    this._isValidEvent = _isValidEvent;
+    this.listeners = {};
+    this.unsubscriptions = [];
+  }
+
+  reset() {
+    this.unsubscriptions.forEach(unsub => unsub());
+    this.unsubscriptions = [];
+  }
+}
```
sedghi, line 107: Nice, this should have been a class from start

wayfarer3130:
@sedghi - for the first use of the service, how about we do a simple global customization - allow adding extra routes via customization, and then for the modes, show the overlay (which should be ready Monday).  

For the naming, I'm trying to avoid name collisions and conflicts with Type versus instance versus registration object.  According to what I understand of the code styling  TypeNames are upper camel case, instanceValues are lower camel case.  Constants are also upper camel case, but can't have the same name as the TypeName,  so that ends up with:
uiCustomizationService for an instanceValue
UICustomizationService for the class for the isntanceValue
UICustomizationRegistration for the registration object.

No, I'm not going to change everything to suite, nor am I going to make PubSubInterface used throughout, but where we have things that disagree with our stated naming conventions, I'd prefer to go back to the naming conventions.  That means I will add the single serve in the lower camel case naming, and leave the other services to be updated at some future point.  Does that mean I think the other ones SHOULD be updated - ABSOLUTELY they should get done, but when the time is appropriate.

platform/docs/docs/platform/managers/service.md
wayfarer3130: I'm eventually proposing that we do it using our specified naming convention, which is:
Type names are UpperCamelCase
instance names are lowerCamelCase
and one does not use the Type name as the instance name - they can be upper/lower camel case equivalent, but they don't overlap.  That prevents confusion when importing actual types.

The fact that we have lots of confusion over the naturalized DICOM data names is because the names should have gotten correctly capitalized.

My proposal is that as we add Type exports, we add a historical upper-case alias, and switch the canonical name to lowerCamelCase.

platform/docs/docs/platform/services/ui/ui-customization-service.md
wayfarer3130: It is part and parcel of how one defines customizations for use, so I think it is actually fairly important to define here.

platform/viewer/src/App.tsx
```diff
@@ -90,6 +91,9 @@ function App({ config, defaultExtensions, defaultModes }) {
 
   let authRoutes = null;
 
+  // Should there be a generic call to init on the extension manager?
+  uiCustomizationService.init(extensionManager);
```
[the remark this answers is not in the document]
wayfarer3130: The general contract is supposed to be lower camel case on instances of types.  We should switch to that as we get there.  I can do it differently if you really want to, but the naming gets long, and the ability to know what things are by name is extremely useful.  If you want, I can create an "alias" for all the services that allow fetching by either name, with the lower camel case being the preferred name, but that should be a separate PR.

platform/core/src/index.ts
```diff
@@ -28,6 +28,7 @@ import {
   pubSubServiceInterface,
   UserAuthenticationService,
   SegmentationService,
+  UICustomizationServiceRegistration,
```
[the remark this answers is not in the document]
wayfarer3130: The other services are WRONG - the values here are NOT SegmentationService etc, they are registration objects for them (and constants, hence upper camel case).  It is extremely ugly to have completely different types of things re-use the same name like that.

platform/core/src/extensions/ExtensionManager.js
wayfarer3130: That breaks the encapsulation between modes and services.  A mode on enter shouldn't know about any services which it doesn't specifically configure/manage.  That is, one should be able to add new services without touching every single mode if that service needs mode setup.  That should ALSO be true for AppInit, that is, there should be an 'onAppInit' call in the service that allows for service specific instantiation.

Putting service specific mode setup/cleanup into the Mode.tsx breaks the encapsulation between the Modes.tsx and the service.  Imagine if a service needed mode setup in the cornerstonejs extension.  Now you need to modify EVERY mode using the cornerstone extension.  That is what I mean by breaking encapsulation.  It would also require tweaking the Modes.tsx file to call the initialization service on setup, so you need an @ohif/ui change in order to call it non-generically from the Modes.tsx.

platform/core/src/services/ToolBarService/ToolBarService.ts
wayfarer3130: The problem is that it ties the Mode.tsx to the service definitions, so in order to add a new "reset" type service, you have to modify the Mode.tsx in core.  Suppose CornerstoneViewportService suddenly needs a reset - if it is in the service, then you just call it, and everything is good.  However, if it is in the Mode.tsx definition, then you need to update the @ohif/ui core code.  Again, that is what i mean by breaking encapsulation.

platform/core/src/services/ToolBarService/ToolBarService.ts
wayfarer3130: It is extra options because it is in addition to the default options specified by the commands.

platform/viewer/src/appInit.js
```diff
@@ -51,6 +52,7 @@ async function appInit(appConfigOrFunc, defaultExtensions, defaultModes) {
     UIViewportDialogService,
     MeasurementService,
     DisplaySetService,
+    [UICustomizationServiceRegistration, appConfig],
```
wayfarer3130, line 55: Make these go into appConfig.uiCustomizationService

platform/core/src/services/ToolBarService/ToolBarService.ts
wayfarer3130: Added the onModeEnter and onModeExit as requested.

extensions/default/src/getCustomizationModule.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+export default function getCustomizationModule() {
+  return [
+    {
+      name: 'customRoutes',
+      value: {
+        id: 'customRoutes',
```
sedghi, line 8: why `id` here is the same name as the `name` above. I don't like this repetition which is also confusing, can id be anything (I assume)? but for customizing the route the `name` should be `customRoutes`?

platform/core/src/extensions/MODULE_TYPES.js
```diff
@@ -1,5 +1,6 @@
 export default {
   COMMANDS: 'commandsModule',
+  UI_CUSTOMIZATION: 'customizationModule',
```
sedghi, line 3: This is confusing, why we are using the `customizationModule` but it is `UI_CUSTOMIZATION`. if you look at all other types here, they all match. Please stick to either `CustomizationModule` and `CustomizatoinService` or `UICustomizationModule` and `UICustomizationService`

platform/core/src/services/UICustomizationService/CustomizationType.ts
```diff
@@ -0,0 +1,4 @@
+export default interface CustomizationTypeInfo extends Record<string, unknown> {
```
sedghi, line 1: Again we should stick to either Customization and CustomizationType OR UICustomization and UICustomizationType. It should not be UICustomizationService + CustomizationType + CustomizationModule

platform/core/src/services/UICustomizationService/UICustomizationService.ts
```diff
@@ -0,0 +1,273 @@
...
+};
+
+type Obj = Record<string, unknown>;
+
+type ComponentReturn = {
+  component: ComponentType;
+  props?: Obj;
+};
+
+type NestedStrings = string[] | NestedStrings[];
```
sedghi, line 19: Can you move these to the CustomizationType file that you have

platform/core/src/services/UICustomizationService/UICustomizationService.ts
```diff
@@ -0,0 +1,273 @@
...
+      globalCustomizations: this.globalCustomizations,
+    });
+  }
+
+  findExtensionValue(value: string) {
+    const entry = this.extensionManager.getModuleEntry(value);
+    return entry;
+  }
+
+  initDefaults() {
```
sedghi, line 223: I see the order of methods in this class is not ideal. For instance, this function is called in the `init` but it is located almost at the bottom of the class. Function calls that are related should be almost close together with the caller being on top of callee.

platform/core/src/services/UICustomizationService/UICustomizationService.ts
```diff
@@ -0,0 +1,273 @@
...
+ * They default to the list ['*'] if not otherwise provided, which means to check
+ * every module for the given id and to load it/add it to the extensions.
+ */
+export default class UICustomizationService extends PubSubService {
+  commandsManager: Record<string, unknown>;
+  extensionManager: ExtensionManager;
+
+  modeCustomizations: Record<string, UICustomization> = {};
+  globalCustomizations: Record<string, UICustomization> = {};
+  configuration: UICustomizationConfiguration;
```
sedghi, line 69: I don't see this guy `UICustomizationConfiguration` being imported anywhere

platform/core/src/services/UICustomizationService/UICustomizationService.ts
```diff
@@ -0,0 +1,273 @@
...
+ *
+ * A customization value can be:
+ *   1. React function, taking (React, props) and returning a rendered component
+ *      For example, createLogoComponentFn renders a component logo for display
+ *   2. Custom UI component configuration, as defined by the component which uses it.
+ *      For example, context menus define a complex structure allowing site-determined
+ *      context menus to be set.
+ *   3. A string name, being the extension id for retrieving one of the above.
+ *
+ * The default values for the extension come from the app_config value 'whiteLabeling',
```
sedghi, line 55: Is this valid anymore? I thought you changed them so that it does not live in `whiteLabeling`

platform/core/src/services/UICustomizationService/types.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export type UICustomization =
+  | BaseUICustomization
+  | LabelUICustomization
+  | CommandUICustomization
+  | CodeUICustomization;
+
+export default UICustomization;
+
+export interface UIConfiguration {
```
sedghi, line 31: I think this is the one that need to be imported for the constructor?

platform/core/src/services/UICustomizationService/types.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface BaseUICustomization extends Record<string, unknown> {
+  id: string;
+  customizationType?: string;
+  description?: string;
+  label?: string;
+  commands?: Command[];
+}
+
+export interface LabelUICustomization extends BaseUICustomization {
```
sedghi, line 11: I see the BaseUICustomization has `label` why LabelUICustomization is adding it as well?

platform/core/src/extensions/ExtensionManager.js
sedghi: I still having hard time getting convinced this is actually a good idea, for the same reasons we talked e.g., in certain use cases like switching modes and keeping measurements measurementService should not reset. 

So then I don't see the benefit of splitting services to some that has onModeEnter and some that don't. Summoning @swederik

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,256 @@
...
+This service is a UI service in that part of the registration allows for registering
+UI components and types to deal with, but it does not directly provide an UI
+displayable elements unless customized to do so.
+
+## Registering Customizations
+There are several ways to register customizations.  The
+`APP_CONFIG.uiCustomizationService`
+field is used as a per-configuration entry.  This object can list single
+configurations by id, or it can list sets of customizations by referring to
+the `customizationModule` in an extension.
```
sedghi, line 25: can we have an example right below?

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,256 @@
...
+
+As well, extensions can register default customizations by providing a 'default'
+name key within the extension.  These are simply customizations loaded when
+the extension is loaded, and applied before the white labeling.
```
sedghi, line 29: No white labeling now, also maybe give the default customization example right after this paragraph

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,256 @@
...
+
+The type and parameters of a customization are defined by the user of the
+customization, based on the customization id.  For example, `cornerstoneOverlay`
+is a customization that is a React component, so it requires a react content,
+and optionally contentProps which are used to supply values to the content.
```
sedghi, line 34: again the example right after would benefit the reader

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,256 @@
...
+
+The extension can also supply a default parent instance to inherit values from.
+This allows the content or other parameters to be pre-filled, and only the
+required values changed.  The parent to use is specified by the `customizationType` field,
+and is simply the id of another customization object.
```
sedghi, line 39: example right after

platform/viewer/src/routes/index.tsx
```diff
@@ -47,7 +46,19 @@ const createRoutes = ({
...
+  const customRoutes = uiCustomizationService.getGlobalCustomization(
+    'customRoutes'
+  );
+  const customBakedInRoutes =
+    customRoutes?.customizeBakedInRoutes?.(bakedInRoutes) || bakedInRoutes;
+
+  const allRoutes = [
+    ...routes,
+    ...(customRoutes?.routes || []),
+    ...customBakedInRoutes,
```
sedghi, line 60: what is the difference between .routes and customeBakedInRoutes, please add a comment

sedghi (commented):
Great revision, I put some more comments, and I think we are there

extensions/default/src/getCustomizationModule.tsx
wayfarer3130: The name is part of the services definition name, so it will be combined with a full name to define a fully qualified name.  The id is the actual id of the customization to change.  This example happens to make them the same, but there may well be cases where one wants them to be different, for example, having a left and a right mammo hanging protocol set.  Both define a set of hanging protocols, so the id is identical between them, but the name might be leftMammo and rightMammo (this refers to not left/right images, but whether left is on left side or on right side).

platform/core/src/extensions/MODULE_TYPES.js
wayfarer3130: They don't all match already, but I agree, they probably should match in a well defined fashion.

extensions/default/src/getCustomizationModule.tsx
wayfarer3130: Changed the example to show the usage of both items - name is now helloPage, and customRoutes is the id.

platform/core/src/services/UICustomizationService/CustomizationType.ts
wayfarer3130: Agreed, this is ugly.  I've deleted customizationType, as that one wasn't being used.  Will fix the others.

platform/core/src/services/UICustomizationService/UICustomizationService.ts
wayfarer3130: Done

platform/core/src/services/UICustomizationService/UICustomizationService.ts
wayfarer3130: Updated, but see if you think more changes are necessary.

platform/core/src/services/UICustomizationService/types.ts
wayfarer3130: I removed this one - it was old/not well setup.  Probably it should eventually get added to the constructor as a properly typed item.

platform/docs/docs/platform/services/ui/ui-customization-service.md
```diff
@@ -0,0 +1,256 @@
...
+    {
+      path: '/myroute',
+      children: MyRouteReactFunction,
+    }
+  ],
+  customizeBakedInRoutes: (bakedInRoutes) => {
+    const ret = [...bakedInRoutes];
+    ret[0] = { path: '/', children: MyCustomSearchPage };
+    return ret;
+  }
```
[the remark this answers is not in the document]
wayfarer3130: Removed the custom routes - solved a better/simpler way.

platform/docs/docs/platform/services/ui/ui-customization-service.md
wayfarer3130: Added one.

platform/docs/docs/platform/services/ui/ui-customization-service.md
wayfarer3130: Added.

sedghi (approved):
Great service to customize stuff. Thank you

Georift:
Hey there @sedghi @wayfarer3130, I'm just updating the #3009 PR to fix merge issues with this, I don't think much of this relates and I'll probably just fix the merge conflicts and call it a day.

However while digging into the documentation this change I noticed my local environment doesn't seem to have the same types as the examples:
```typescript
import { Types } from '@ohif/ui';

// UIContextMenu is not found
const customContextMenu: Types.UIContextMenu
```

I also checked the deployed package `@ohif/ui`, but as there is no version 3 released I don't imagine this applys?

Are these types coming soon in another branch? Or maybe an example for docs?

I'd be very interested in hearing what the OHIF way is with types, I've been dreaming of building some types/intelisense for the config object.

GlebVST:
Hey guys, I noticed a few types are not resolving in the `CustomizationService`:

```
configuration: UICustomizationConfiguration;
```

and also:

```
protected setConfigGlobalCustomization(
    configuration: AppConfigCustomization
  ):

```

I can't find the `UICustomizationConfiguration` and `AppConfigCustomization` declarations anywhere - are they lost during merge or what?

PR #2997, opened 2022-10-31
Remove unused code from DICOM SR parsing

GitanjaliChhetri:
Within TID 1500, sub-template TID 1600 (Image Library) is not required while parsing SR for image references for annotations and planar measurements. The same is obtained from sub template TID 1501>TID 300>TID 320.
_getReferencedImagesList_ is not referenced anywhere else in the code and should be removed. 
	
Tested that annotations from DICOM SRs still work.


igoroctaviano (approved):
LGTM 🥇

PR #3000, opened 2022-11-01
Display series specified by URL query parameters in viewer

TFRadicalImaging:
This change allow the filtering passed by the query parameter being made only over the OHIF client without need any change on the static-dicomweb to make it works. The user can pass through the URL the following query parameter: seriesInstaceUID or SeriesInstanceUID, retrieve the study information, and based on the series requested by the URL, it will only do the request regarding the series applied. This change was requests by @Zaid-Safadi and it's working as expected. If you take a look on the code, the only issue was on the filter option, what was using the wrong value to make it works.

### DIFF
--- a/extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.js
+++ b/extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.js
@@ -41,8 +41,17 @@ export default class StaticWadoClient extends api.DICOMwebClient {
     const { queryParams } = options;
     if (!queryParams) return searchResult;
     const filtered = searchResult.filter(study => {
-      for (const key of Object.keys(StaticWadoClient.studyFilterKeys)) {
-        if (!this.filterItem(key, queryParams, study)) return false;
+      for (const key of Object.keys(StaticWadoClient.seriesFilterKeys)) {
+        if (
+          !this.filterItem(
+            key,
+            queryParams,
+            study,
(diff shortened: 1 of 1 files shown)


extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.js
```diff
@@ -56,8 +65,17 @@ export default class StaticWadoClient extends api.DICOMwebClient {
     const { queryParams } = options;
     if (!queryParams) return searchResult;
     const filtered = searchResult.filter(study => {
-      for (const key of Object.keys(StaticWadoClient.studyFilterKeys)) {
-        if (!this.filterItem(key, queryParams, study)) return false;
+      for (const key of Object.keys(StaticWadoClient.seriesFilterKeys)) {
```
md-prog, line 68: For Series, shouldn't we send the combination of the `studyFilterKeys` and `seriesFilterKeys` ?

md-prog (commented):
Otherwise looks good.

md-prog (approved):

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.js
md-prog: if `searchForSeries()` is called after `searchForStudies()` , then it would be fine.

PR #3019, opened 2022-11-14
Fix #3020 - Elliptical Annotations not being added/tracked when created outside the actual image area

### LINKED ISSUE #3020: Elliptical Annotations not being added/tracked when created outside the actual image area
- When creating an elliptical annotations outside the actual image area, it won't either start the measurement tracking, or display the added annotation to the tracking panel
- This was due to the "area" calculated for the annotation would be "undefined" and that was causing exception
- We have `preventHandleOutsideImage` flag for each of the annotations, but this flag seems to be a more buggy.
   If we perform some of the monkey-test with this flag on, some of annotations that shouldn't be added are getting added to the tracking panel and some annotations that are supposed to be added are not being displayed properly.

md-prog:
The easiest fix for the issue would be handling of the cases where "area" becomes "undefined"


sedghi (approved):
Looks great, thanks

PR #3025, opened 2022-11-18
fix: Correctly ignore directories in CopyWebpackPlugin in webpack config

HarmvZ:
Previously, the `config/` and `html-templates/` directories were copied to the `dist/` folder when building because of an incorrect reference in the webpack config.


sedghi (approved):
Looks great thanks

PR #3026, opened 2022-11-19
Fix/v33 remaining bugs

sedghi:
- warn user for overlapping segments which is not currently supported
- more general MPR which can be applied on any porotocol; not only on default HP like before


extensions/cornerstone/src/commandsModule.ts
```diff
@@ -160,48 +146,51 @@ const commandsModule = ({ servicesManager }) => {
...
 
-        // Set the new toolName to be active
-        toolGroup.setToolActive(toolName, {
-          bindings: [{ mouseButton: Enums.MouseBindings.Primary }],
-        });
+      const activeToolName = toolGroup.getActivePrimaryMouseButtonTool();
 
-        return;
+      if (activeToolName) {
+        // Todo: this is a hack to prevent the crosshairs to stick around
```
wayfarer3130, line 178: Discuss in slack

extensions/cornerstone/src/utils/mpr/toggleMPRHangingProtocol.ts
```diff
@@ -0,0 +1,285 @@
+import { Enums } from '@cornerstonejs/tools';
```
wayfarer3130, line 1: Thank you for splitting this out - it makes it easier to understand to have things in smaller pieces.

wayfarer3130 (changes requested):
Lets look at the set active/inactive and see if that code can be moved into the Tools instances themselves, rather than having the if condition on the two types.  Otherwise the changes look good.

sedghi:
@wayfarer3130 Sure, as discussed privately, we will do it after the release

wayfarer3130 (approved):
Changes look good now - just make sure we have an issue to update the handling of tools with respect to whether they are active etc.

PR #3029, opened 2022-11-21
feat(datasource):Support querying different data sources

wayfarer3130:
This change adds the ability to show the worklist against different data sources by using a path /:datasourcename.  It also fixes the various sub-paths so that they return to the correct parent data source name.  For example, in the multiple.js configuration, there are data sources:  default, aws and e2e.  Then, one can query three different back ends with:
http://localhost:3000/search/default
http://localhost:3000/search/aws
http://localhost:3000/search/e2e
You can pull/try those three examples (assuming you have a dicomweb instance on port 5000 running).

### DIFF
--- a/extensions/default/src/ViewerLayout/index.tsx
+++ b/extensions/default/src/ViewerLayout/index.tsx
@@ -2,6 +2,8 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useNavigate } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
+import { useLocation } from 'react-router';
+
 import {
   SidePanel,
   ErrorBoundary,
@@ -11,7 +13,6 @@ import {
   useModal,
   LoadingIndicatorProgress,
 } from '@ohif/ui';
-
 import i18n from '@ohif/i18n';
 import { hotkeys } from '@ohif/core';
 import { useAppConfig } from '@state';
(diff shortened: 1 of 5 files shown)


wayfarer3130:
@Zaid-Safadi - could you review this PR please?

platform/viewer/src/routes/index.tsx
```diff
@@ -20,6 +20,12 @@ const bakedInRoutes = [
     private: true,
     props: { children: WorkList },
   },
+  {
```
sedghi, line 23: A mode can be configured in the URL to use a custom data source, why you are adding another route for data source name? 

https://github.com/OHIF/Viewers/blob/v3-stable/platform/viewer/src/routes/buildModeRoutes.tsx#L53

sedghi: you can use `/tmtv/dicomjson` and `/tmtv/customDataSource` as a result of that

wayfarer3130: Because there isn't a data source name route that allows showing the query page on a given data source.  Suppose I have two data sources, "default" and "aws" (as in the multiple.js configuration)  The first data source found would previously have defined the default worklist datasource, and there wouldn't be a way of querying to anything else.
Now, you can do: http://localhost:3000/aws to bring up a query page against the second data source.

sedghi: So I don't like breaking out convention that the first argument after `/` is the modeId. I believe having `/aws` and `/tmtv` is confusing. Can you put the data source name for the worklist as a param? something like `myviewer.com/search?datasource=aws` or similar?

wayfarer3130: Changed so that the convention is follow - that is,
/search/:dataSourceName
for anything except the default data source name, which is still on /

Zaid-Safadi (approved):
LGTM

sedghi:
@wayfarer3130 Can we please stick to the same convention of how we do filtering on mrn, etc. 

`http://localhost:3000/?description=sdf&patientName=name`

so 

`http://localhost:3000/?datasourcename=aws`

sedghi:
Actually I do believe this is just another query like others that we have here

https://github.com/OHIF/Viewers/blob/v3-stable/platform/viewer/src/routes/WorkList/WorkList.tsx#L513-L530

Also look how Ouwen is doing his configURL PR here #2925  which is very similar

PR #3036, opened 2022-11-23
fix(misc): Update the data source name, provide  a /datasources example customization for debug

wayfarer3130:
Change the datasource name to be datasources, add a page for viewing the available queries and fix some imports to avoid multiples.


extensions/default/src/Panels/DataSourceSelector.tsx
```diff
@@ -0,0 +1,40 @@
...
+              {ds.sourceName}
+            </Button>
+            <br />
+          </>
+        ))}
+    </div>
+  );
+}
+
+export default DataSourceSelector;
```
sedghi, line 40: Since this is becoming part of UI, we need to get a design for it. Can this PR not include this?

wayfarer3130: This GUI on the level of the debug page, for use for testing purposes.  Final/real GUI should be done direct in the search page.  It isn't even included by default, you  have to enable it as a customization, although the multiple.js config file enables it.  I'd prefer to get something in sooner so we don't keep on changing names of parameters, so if you really want, I will delete all GUI elements and just incorporate it into an extension.  A real GUI requires significant discussion about multi versus single select, how to display the data source, how to handle query against one back end and retrieve on another, fallback for error conditions, ordering of results etc.

sedghi: Yeah, certainly there are a lot of variants (multi select etc.), if it is temporary and will move to the worklist, then can we just not add it in this PR? I'm in discussion with Dan to design that

extensions/default/src/Panels/DataSourceSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+import classnames from 'classnames';
+import { useNavigate } from 'react-router-dom';
+
+import { Button } from '@ohif/ui';
+
+function DataSourceSelector() {
+  console.log('Rendering multi-data sources initial page');
+  const navigate = useNavigate();
+
+  const dsConfigs = window.config.dataSources;
```
sedghi, line 11: We shouldn't encourage people doing this. You can grab it from the extension manager

gholamrezaeesaeed:
سلام خوبی آقا علیرضا
من یه سئوال در مورد خطای CORS دارم
ممنون میشم اگه لطف کنی راهنمایی کنی خیلی روش کار کردم اما درست نشد

PR #3041, opened 2022-11-25
[Draft] feat: switch to oidc-client-ts since our current oidc-client is deprecated

sedghi:
switch to oidc-client-ts since the oidc-client is deprecated. It significantly reduces the bundle size as well. However, it seems like it only supports `response_type` of `code` and for our current `id_token token` we need to add `client_secret` as well

https://github.com/authts/oidc-client-ts/blob/main/docs/migration.md
https://github.com/perarnborg/vuex-oidc/pull/187
https://github.com/authts/oidc-client-ts/issues/152
https://github.com/authts/oidc-client-ts/issues/571

Old client in build analyzer
image

### DIFF
--- a/.webpack/webpack.base.js
+++ b/.webpack/webpack.base.js
@@ -146,7 +146,7 @@ module.exports = (env, argv, { SRC_DIR, DIST_DIR }) => {
         ),
       }),
       // Uncomment to generate bundle analyzer
-      new BundleAnalyzerPlugin(),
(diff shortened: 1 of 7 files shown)


sedghi:
Closing as I implemented it here https://github.com/OHIF/Viewers/pull/4234

sedghi:
Fixed here https://github.com/OHIF/Viewers/pull/4234

PR #3044, opened 2022-11-25
feat(imageLoader):Add an nth image loader strategy (#10)

wayfarer3130:
Add an nth image loader, making the appearance of volumes much nicer when they are large by creating some detail everywhere immediately.


extensions/cornerstone/src/utils/getNthFrames.js
```diff
@@ -0,0 +1,29 @@
+export default function getNthFrames(imageIds) {
```
sedghi, line 1: can you add some docs to this function?

sedghi (commented):
Some docs request, otherwise looks great

wayfarer3130:
Added docs as requested.

Issue #3050, opened 2022-11-29
Measurements not working correctly in TMTV mode


BrandonDatUHN:
Repository URL

https://github.com/OHIF/Viewers/tree/v3-stable

Version number

3.3.0

Build number

0

Browser

Chrome 107.0.0

OS

Windows 10

testing on https://v3-demo.ohif.org/tmtv?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2403.459769504433903221904322299373

Loading the PET-CT from OHIF demo Site in TMTV then clicking the ellipse tool will not draw an ellipse.  A point will be added but you can't delete it.  If you add more they will not be shown.  If you leave and go back to the study list then return all ROIs will now display BUT you can't delete or move them.

Video of the behaviour below.

https://watch.screencastify.com/v/jRb7Q8QadK2UvazhoN92

BrandonDatUHN:
Working with fix.

Issue #3049, opened 2022-11-29
MPR Not working with V3-stable


mokbelt:
MPR is not working on v3-stable after building.
It working normally with V2.

mokbelt:
When I click on MPR, I get this error, and nothing opens on the UI (howhewver working normally on V2).
I am using Chrome.

createFloat32SharedArray.js:3 Uncaught (in promise) Error: Your page is NOT cross-origin isolated, see https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated
    at cm (createFloat32SharedArray.js:3:15)
    at sw (cornerstoneStreamingImageVolumeLoader.js:50:26)
    at Zt (volumeLoader.js:42:30)
    at Module.Kt (volumeLoader.js:76:24)
    at e.<anonymous> (CornerstoneCacheService.ts:181:37)
    at c (CornerstoneCacheService.ts:2:1)
    at Generator._invoke (CornerstoneCacheService.ts:2:1)
    at Generator.next (CornerstoneCacheService.ts:2:1)
    at aN (CornerstoneCacheService.ts:2:1)
    at o (CornerstoneCacheService.ts:2:1)

![image](https://user-images.githubusercontent.com/56265911/204652431-fa09fb40-bb60-4774-8e33-faf614f46829.png)

ranasrule:
> When I click on MPR, I get this error, and nothing opens on the UI (howhewver working normally on V2). I am using Chrome.
> 
> createFloat32SharedArray.js:3 Uncaught (in promise) Error: Your page is NOT cross-origin isolated, see https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated at cm (createFloat32SharedArray.js:3:15) at sw (cornerstoneStreamingImageVolumeLoader.js:50:26) at Zt (volumeLoader.js:42:30) at Module.Kt (volumeLoader.js:76:24) at e. (CornerstoneCacheService.ts:181:37) at c (CornerstoneCacheService.ts:2:1) at Generator._invoke (CornerstoneCacheService.ts:2:1) at Generator.next (CornerstoneCacheService.ts:2:1) at aN (CornerstoneCacheService.ts:2:1) at o (CornerstoneCacheService.ts:2:1)
> 
> ![image](https://user-images.githubusercontent.com/56265911/204652431-fa09fb40-bb60-4774-8e33-faf614f46829.png)

Already discussed  and solved at the below link

https://github.com/OHIF/Viewers/issues/3033

[...]

Issue #3043, opened 2022-11-25
unable to see text because of its colour


JennyLeePMH:
> **Before Creating an issue**
>
> - Are you running the latest version?
> - Are you reporting to the correct repository?
> - Did you search existing issues?

### When saving an SR, unable to see text on button because of its colour

See attached.

```js
Please use code blocks to show formatted errors or code snippets
```
![odd_colour_measurement](https://user-images.githubusercontent.com/4551564/204035248-fb274cd3-0f38-4ae0-a80d-6d1e3d8faf07.PNG)

> :warning: Reports we cannot reproduce are at risk of being marked stale and
> closed. The more information you can provide, the more likely we are to look
> into and address your issue.

JennyLeePMH:
Forgot to mention, this is from OHIF's v3-stable branch.

Georift:
I can see this on the current [demo url](https://v3-demo.ohif.org/). Likely related to this change: #3031?

JennyLeePMH:
thanks.

Issue #3039, opened 2022-11-24
Cine Mode will play but not stop


BrandonDatUHN:
This bug is specific to the latest release Repository URL

https://github.com/OHIF/Viewers/tree/v3-stable

Version number

3.3.0

Build number

0

Browser

Chrome 107.0.0

OS

Windows 10

If you open an image and click cine mode it will play but the stop button doesn't work.  What's worse is that now all images are stuck in cine mode.  Even going back to the study list and opening the next study it will keep scrolling.  This also shows an SBOX_MEMORY error at the end (we've removed that on our deployment by setting the workers to 1)

https://watch.screencastify.com/v/JCl6afOUlXTjS91zNVAB

BrandonDatUHN:
Fixed now.

PR #3052, opened 2022-12-01
fix: various bugs with v33 re cine and colors

sedghi:
Bugs fixed
- https://github.com/OHIF/Viewers/issues/3050
- https://github.com/OHIF/Viewers/issues/3039
- https://github.com/OHIF/Viewers/issues/3043
- https://github.com/OHIF/Viewers/issues/3024

Add /debug to add additional info for debugging (currently limited info)
- https://github.com/OHIF/Viewers/issues/3049

image

### DIFF
--- a/platform/ui/src/components/Button/Button.tsx
+++ b/platform/ui/src/components/Button/Button.tsx
@@ -44,7 +44,8 @@ const variants = {
       'text-primary-main hover:opacity-80 active:opacity-100 focus:opacity-80',
     secondary:
       'text-secondary-light hover:opacity-80 active:opacity-100 focus:opacity-80',
-    white: 'text-white hover:opacity-80 active:opacity-100 focus:opacity-80',
+    white:
+      'text-black hover:bg-primary-main focus:bg-primary-main hover:border-black focus:border-black',
     black:
       'text-white hover:bg-primary-main focus:bg-primary-main hover:border-black focus:border-black',
     primaryActive:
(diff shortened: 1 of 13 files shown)


swederik (approved):

platform/ui/src/components/Button/Button.tsx
```diff
@@ -46,7 +46,7 @@ const variants = {
       'text-secondary-light hover:opacity-80 active:opacity-100 focus:opacity-80',
     white: 'text-white hover:opacity-80 active:opacity-100 focus:opacity-80',
     black:
-      'text-black hover:bg-primary-main focus:bg-primary-main hover:border-black focus:border-black',
+      'text-white hover:bg-primary-main focus:bg-primary-main hover:border-black focus:border-black',
```
wayfarer3130, line 49: This should be text-black - the specific behaviour is that this is specified by the color attribute of the button, which specifies the text color for use - specifically look at line 47, which has text-white.  If it is desire to use white, then specify white as the annotation.

wayfarer3130 (changes requested):
Need to discuss the white/black text issue.

platform/ui/src/components/Button/Button.tsx
sedghi: This is actually a problem with the PR we recently changed, there are various places in UI that are broken as a result, my recommendation is to revert that change (which was an oversight from me since I didn't check all the components), and a new PR that probably add a new variant for the app you are working on. How about that?

wayfarer3130 (approved):
Looks good now.

PR #3053, opened 2022-12-01
Add CT volume in PET ROI Thresholding

rodrigobasilio2022:
### PR Checklist
- Adds the CT volume data in ROI Thresholding
- Adds a new line in ROI Threshold panel to configure CT ranges

- This PR enables the user to create REOI by filtering both CT and PET data, facilitating the creation of roi in anatomical body parts with specific CT intensity ranges


rodrigobasilio2022:
> There are obvious UI bugs here
> 
> image
> 
> * CT and PT font size or weight is different
> * There is some extra padding on their left
> * The second row should not repeat lower and upper

Done

PR #3054, opened 2022-12-01
Fix Dicom Tag Browser slider doesn't update when series changed

Michael-Andersen:
In the Dicom Tag Browser if you move the instance slider and then change series, the slider won't update to the first position even though the new series displayed will start with the first instance. 

I've fixed this by giving the slider the displaySetInstanceUID as a key to force it to re-render when the series changes. I'm open to other solutions to the bug.

@sedghi Tagging you for review


sedghi (approved):
Thanks

PR #3058, opened 2022-12-04
Hp image matching

Zaid-Safadi:
### PR Checklist

- [X] Brief description of changes
The main change is to enable the hanging protocol to consider the "imageMatchingRules" by filtering out images from the displayset that don't match the rules.
In addition, the following changes has been made:
1. support autoplay in video viewport
2. fix loading resources with http/https url

### DIFF
--- a/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
+++ b/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
@@ -105,6 +105,7 @@ const OHIFCornerstoneViewport = React.memo(props => {
     servicesManager,
     onElementEnabled,
     onElementDisabled,
+    needsRerendering,
     // Note: you SHOULD NOT use the initialImageIdOrIndex for manipulation
     // of the imageData in the OHIFCornerstoneViewport. This prop is used
     // to set the initial state of the viewport's first image to render
@@ -278,6 +279,25 @@ const OHIFCornerstoneViewport = React.memo(props => {
     // handle the default viewportType to be stack
     if (!viewportOptions.viewportType) {
       viewportOptions.viewportType = STACK;
+    }
+
+    if (needsRerendering) {
+      displaySets.forEach(displaySet => {
+        const viewportInfo = CornerstoneViewportService.getViewportInfoByIndex(
+          viewportIndex
+        );
+
+        if (viewportInfo.hasDisplaySet(displaySet.displaySetInstanceUID)) {
+          const viewportData = viewportInfo.getViewportData();
+          const currentViewportData = (viewportData) ? viewportData : { viewportType: STACK };
+          CornerstoneCacheService.invalidateViewportData(currentViewportData,
+            displaySet.displaySetInstanceUID,
+            dataSource,
+            DisplaySetService);
+
+          displaySet.needsRerendering = false;
+        }
+      });
     }
 
     const loadViewportData = async () => {
--- a/extensions/cornerstone/src/initWADOImageLoader.js
+++ b/extensions/cornerstone/src/initWADOImageLoader.js
@@ -54,7 +54,7 @@ export default function initWADOImageLoader(
       // we should set this flag to false.
       convertFloatPixelDataToInt: false,
     },
-    beforeSend: function(xhr) {
+    beforeSend: function (xhr, imagId) {
       const headers = UserAuthenticationService.getAuthorizationHeader();
 
       // Request:
@@ -62,12 +62,12 @@ export default function initWADOImageLoader(
       // whatever transfer-syntax the origin server provides.
       // For now we use image/jls and image/x-jls because some servers still use the old type
       // http://dicom.nema.org/medical/dicom/current/output/html/part18.html
-      const xhrRequestHeaders = {
+      const xhrRequestHeaders = isWadoRs(imagId) ? {
         Accept: appConfig.omitQuotationForMultipartRequest
           ? 'multipart/related; type=application/octet-stream'
           : 'multipart/related; type="application/octet-stream"',
         // 'multipart/related; type="image/x-jls", multipart/related; type="image/jls"; transfer-syntax="1.2.840.10008.1.2.4.80", multipart/related; type="image/x-jls", multipart/related; type="application/octet-stream"; transfer-syntax=*',
-      };
+      } : {};
 
       if (headers && headers.Authorization) {
         xhrRequestHeaders.Authorization = headers.Authorization;
@@ -81,6 +81,10 @@ export default function initWADOImageLoader(
   });
 
   initWebWorkers(appConfig);
+}
+
+function isWadoRs(imageId) {
+  return imageId.indexOf('wadors:') == 0;
 }
 
 export function destroy() {
--- a/extensions/cornerstone/src/initWADOImageLoader.js
+++ b/extensions/cornerstone/src/initWADOImageLoader.js
@@ -62,6 +62,9 @@ export default function initWADOImageLoader(
       // whatever transfer-syntax the origin server provides.
       // For now we use image/jls and image/x-jls because some servers still use the old type
       // http://dicom.nema.org/medical/dicom/current/output/html/part18.html
+      // The Accept header should only be set to a multipart/related type in
+      // wado-rs request. Other requests (e.g. wado-uri) that don't support
+      // this value in the accept header may cause the server to reject the request as unacceptable.
       const xhrRequestHeaders = isWadoRs(imagId) ? {
         Accept: appConfig.omitQuotationForMultipartRequest
           ? 'multipart/related; type=application/octet-stream'
@@ -84,7 +87,7 @@ export default function initWADOImageLoader(
 }
 
 function isWadoRs(imageId) {
-  return imageId.indexOf('wadors:') == 0;
+  return imageId.indexOf('wadors:') === 0;
 }
 
 export function destroy() {
--- a/extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
+++ b/extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
@@ -3,7 +3,6 @@ import { utils } from '@ohif/core';
 
 import getCornerstoneViewportType from '../../utils/getCornerstoneViewportType';
 import {
-  StackData,
   StackViewportData,
   VolumeViewportData,
 } from '../../types/CornerstoneCacheService';
@@ -12,8 +11,8 @@ const VOLUME_IMAGE_LOADER_SCHEME = 'streaming-wadors';
 const VOLUME_LOADER_SCHEME = 'cornerstoneStreamingImageVolume';
 
 class CornerstoneCacheService {
-  stackImageIds: Map<string, string[]> = new Map();
-  volumeImageIds: Map<string, string[]> = new Map();
+  stackImageIdsMap: Map<string, string[]> = new Map();
+  volumeImageIdsMap: Map<string, string[]> = new Map();
 
   constructor(servicesManager) {
     this.servicesManager = servicesManager;
@@ -68,15 +67,15 @@ class CornerstoneCacheService {
   }
 
   public async invalidateViewportData(
-    viewportData: StackData | VolumeViewportData,
+    viewportData: StackViewportData | VolumeViewportData,
     invalidatedDisplaySetInstanceUID: string,
     dataSource,
     DisplaySetService
   ) {
     if (viewportData.viewportType === Enums.ViewportType.STACK) {
       const displaySet = DisplaySetService.getDisplaySetByUID(invalidatedDisplaySetInstanceUID)
-      if (this.stackImageIds.has(invalidatedDisplaySetInstanceUID)) {
-        this.stackImageIds.delete(invalidatedDisplaySetInstanceUID);
+      if (this.stackImageIdsMap.has(invalidatedDisplaySetInstanceUID)) {
+        this.stackImageIdsMap.delete(invalidatedDisplaySetInstanceUID);
         return await this._getStackViewportData(
           dataSource,
           [displaySet],
@@ -114,13 +113,13 @@ class CornerstoneCacheService {
     // For Stack Viewport we don't have fusion currently
     const displaySet = displaySets[0];
 
-    let stackImageIds = this.stackImageIds.get(
+    let stackImageIds = this.stackImageIdsMap.get(
       displaySet.displaySetInstanceUID
     );
 
     if (!stackImageIds) {
       stackImageIds = this._getCornerstoneStackImageIds(displaySet, dataSource);
-      this.stackImageIds.set(displaySet.displaySetInstanceUID, stackImageIds);
+      this.stackImageIdsMap.set(displaySet.displaySetInstanceUID, stackImageIds);
     }
 
     const { displaySetInstanceUID, StudyInstanceUID } = displaySet;
@@ -175,7 +174,7 @@ class CornerstoneCacheService {
 
       const volumeId = `${volumeLoaderSchema}:${displaySet.displaySetInstanceUID}`;
 
-      let volumeImageIds = this.volumeImageIds.get(
+      let volumeImageIds = this.volumeImageIdsMap.get(
         displaySet.displaySetInstanceUID
       );
 
@@ -191,7 +190,7 @@ class CornerstoneCacheService {
           imageIds: volumeImageIds,
         });
 
-        this.volumeImageIds.set(
+        this.volumeImageIdsMap.set(
           displaySet.displaySetInstanceUID,
           volumeImageIds
         );
--- a/extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
+++ b/extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
@@ -3,6 +3,7 @@ import { utils } from '@ohif/core';
 
 import getCornerstoneViewportType from '../../utils/getCornerstoneViewportType';
 import {
+  StackData,
   StackViewportData,
   VolumeViewportData,
 } from '../../types/CornerstoneCacheService';
@@ -67,34 +68,42 @@ class CornerstoneCacheService {
   }
 
   public async invalidateViewportData(
-    viewportData: VolumeViewportData,
+    viewportData: StackData | VolumeViewportData,
     invalidatedDisplaySetInstanceUID: string,
     dataSource,
     DisplaySetService
   ) {
     if (viewportData.viewportType === Enums.ViewportType.STACK) {
-      throw new Error('Invalidation of StackViewport is not supported yet');
+      const displaySet = DisplaySetService.getDisplaySetByUID(invalidatedDisplaySetInstanceUID)
+      if (this.stackImageIds.has(invalidatedDisplaySetInstanceUID)) {
+        this.stackImageIds.delete(invalidatedDisplaySetInstanceUID);
+        return await this._getStackViewportData(
+          dataSource,
+          [displaySet],
+          null
+        );
+      }
+    } else {
+      // Todo: grab the volume and get the id from the viewport itself
+      const volumeId = `${VOLUME_LOADER_SCHEME}:${invalidatedDisplaySetInstanceUID}`;
+
+      const volume = cs3DCache.getVolume(volumeId);
+
+      if (volume) {
+        cs3DCache.removeVolumeLoadObject(volumeId);
+      }
+
+      const displaySets = viewportData.data.map(({ displaySetInstanceUID }) =>
+        DisplaySetService.getDisplaySetByUID(displaySetInstanceUID)
+      );
+
+      const newViewportData = await this._getVolumeViewportData(
+        dataSource,
+        displaySets
+      );
+
+      return newViewportData;
     }
-
-    // Todo: grab the volume and get the id from the viewport itself
-    const volumeId = `${VOLUME_LOADER_SCHEME}:${invalidatedDisplaySetInstanceUID}`;
-
-    const volume = cs3DCache.getVolume(volumeId);
-
-    if (volume) {
-      cs3DCache.removeVolumeLoadObject(volumeId);
-    }
-
-    const displaySets = viewportData.data.map(({ displaySetInstanceUID }) =>
-      DisplaySetService.getDisplaySetByUID(displaySetInstanceUID)
-    );
-
-    const newViewportData = await this._getVolumeViewportData(
-      dataSource,
-      displaySets
-    );
-
-    return newViewportData;
   }
 
   private _getStackViewportData(
--- a/platform/core/src/classes/MetadataProvider.js
+++ b/platform/core/src/classes/MetadataProvider.js
@@ -443,17 +443,7 @@ class MetadataProvider {
       };
     }
 
-    // Maybe its a non-standard imageId
-    // check if the imageId starts with http:// or https:// using regex
-    // Todo: handle non http imageIds
-    let imageURI;
-    const urlRegex = /^(http|https):\/\//;
-    if (urlRegex.test(imageId)) {
-      imageURI = imageId;
-    } else {
-      imageURI = imageIdToURI(imageId);
-    }
-
+    const imageURI = imageIdToURI(imageId);
     const uids = this.imageURIToUIDs.get(imageURI);
     const frameNumber = imageId.split(/\/frames\//)[1];
 
--- a/platform/core/src/services/DisplaySetService/DisplaySetService.js
+++ b/platform/core/src/services/DisplaySetService/DisplaySetService.js
@@ -30,6 +30,7 @@ const findInstance = (instance, displaySets) => {
 
 export default class DisplaySetService {
   constructor() {
+    this._unfilteredDisplaySets = new Map();
     this.activeDisplaySets = [];
     this.listeners = {};
     this.EVENTS = EVENTS;
@@ -46,6 +47,10 @@ export default class DisplaySetService {
   _addDisplaySetsToCache(displaySets) {
     displaySets.forEach(displaySet => {
       displaySetCache.push(displaySet);
+
+      const instances = displaySet.images || displaySet.others || [];
+      this._unfilteredDisplaySets.set(displaySet.displaySetInstanceUID,
+        [...instances]);
     });
   }
 
@@ -260,5 +265,46 @@ export default class DisplaySetService {
       }
     }
     return allDisplaySets;
+  }
+
+
+  filterDisplaySetInstances(displaySetInstanceUID, filteredInstances) {
+    const displaySet = this.getDisplaySetByUID(displaySetInstanceUID);
+    let changed = false;
+
+    if (displaySet) {
+      const instances = displaySet.images || displaySet.others || [];
+
+      if (instances.length === filteredInstances.length) return; //shortcut
+      for (let index = instances.length - 1; index >= 0; index--) {
+        if (!filteredInstances.includes(instances[index])) {
+          instances.splice(index, 1);
+          changed = true;
+        }
+      }
+    }
+
+    if (changed) {
+      this._broadcastEvent(EVENTS.DISPLAY_SETS_FILTERED, displaySet);
+    }
+  }
+
+  clearDisplaySetFilters(displaySetInstanceUID) {
+    const displaySet = this.getDisplaySetByUID(displaySetInstanceUID);
+    const allInstances = this._unfilteredDisplaySets.get(displaySetInstanceUID);
+    const displaySetInstances = displaySet.images || displaySet.others;
+
+    if (displaySetInstances.length !== allInstances.length) {
+      displaySetInstances.length = 0;
+      displaySetInstances.push(...allInstances);
+      this._broadcastEvent(EVENTS.DISPLAY_SETS_FILTERED, displaySet);
+    }
+  }
+
+  clearAllDisplaySetsFilters() {
+    displaySetCache.forEach(displaySet => {
+      this.clearDisplaySetFilters(displaySet.displaySetInstanceUID)
+    }
+    )
   }
 }
--- a/platform/core/src/services/DisplaySetService/DisplaySetService.js
+++ b/platform/core/src/services/DisplaySetService/DisplaySetService.js
@@ -30,10 +30,10 @@ const findInstance = (instance, displaySets) => {
 
 export default class DisplaySetService {
   constructor() {
-    this._unfilteredDisplaySets = new Map();
     this.activeDisplaySets = [];
     this.listeners = {};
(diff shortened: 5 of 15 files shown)


jbocce:
Just a comment for future commits... several of the commit comments for this PR do not follow the guidelines set out in https://docs.google.com/presentation/d/1WQeKPfVqXpnkn09eJus9_xVX3VNshzsGPOBvE-XVzp8/edit#slide=id.g14247aab95c_0_46 slide 34. In particular, the doc changes commit comment should be prefixed with "doc:" and the very first commit comment should be prefixed with either "feat:" or "fix:".

extensions/dicom-video/src/viewports/OHIFCornerstoneVideoViewport.tsx
```diff
@@ -30,6 +32,9 @@ function OHIFCornerstoneVideoViewport({ displaySets }) {
         controlsList="nodownload"
         preload="auto"
         className="w-full h-full"
+        crossOrigin={(appConfig.videoViewport && appConfig.videoViewport.crossOrigin) ? appConfig.videoViewport.crossOrigin : undefined}
```
jbocce, line 35: Are these new settings in the config? If so, should we add them in the default config just to "show" that they are available? Also should we document these in the extension's README instead? Apologies if I am totally off base with this. :)

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -280,6 +281,25 @@ const OHIFCornerstoneViewport = React.memo(props => {
...
 
+    if (needsRerendering) {
+      displaySets.forEach(displaySet => {
+        const viewportInfo = CornerstoneViewportService.getViewportInfoByIndex(
+          viewportIndex
+        );
+
+        if (viewportInfo.hasDisplaySet(displaySet.displaySetInstanceUID)) {
+          const viewportData = viewportInfo.getViewportData();
+          const currentViewportData = (viewportData) ? viewportData : { viewportType: STACK };
```
jbocce, line 292: Consider replacing the two lines with a single line. Not a big deal, just something to consider. :)

```const currentViewportData = viewportInfo.getViewportData() ?? { viewportType: STACK };```

File: platform/core/src/services/DisplaySetService/DisplaySetService.js [hunk 3]
```diff
@@ -245,4 +250,45 @@ export default class DisplaySetService {
...
+
+  filterDisplaySetInstances(displaySetInstanceUID, filteredInstances) {
+    const displaySet = this.getDisplaySetByUID(displaySetInstanceUID);
+    let changed = false;
+
+    if (displaySet) {
+      const instances = displaySet.images || displaySet.others || [];
+
+      if (instances.length === filteredInstances.length) return; //shortcut
+      for (let index = instances.length - 1; index >= 0; index--) {
```
jbocce, line 263: Is it not possible to use instances.filter here? Perhaps I am missing something?

platform/core/src/services/HangingProtocolService/HangingProtocolService.test.js
```diff
@@ -14,6 +16,7 @@ const testProtocol = {
   ],
   displaySetSelectors: {
     displaySetSelector: {
+      imageMatchingRules: [],
```
jbocce, line 19: Should we not add a test for this new functionality?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -175,7 +175,21 @@ class HangingProtocolService {
...
-    return protocol;
+    if (protocol instanceof Function) {
+      try {
+        const { protocol: generatedProtocol } = this._getProtocolFromGenerator(
+          protocol
+        );
+
+        return generatedProtocol;
+      } catch (error) {
+        console.warn(
```
jbocce, line 186: Is ```console.error better here?

jbocce: Is ```console.error``` better here?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -175,7 +175,21 @@ class HangingProtocolService {
   public getProtocolById(id: string): HangingProtocol.Protocol {
     const protocol = this.protocols.get(id);
 
-    return protocol;
+    if (protocol instanceof Function) {
```
jbocce, line 178: Does it make sense to instead check for ```protocol instanceof HangingProtocol.ProtocolGenerator```?

jbocce (commented):
See my comments that are part of this review.

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -280,6 +281,25 @@ const OHIFCornerstoneViewport = React.memo(props => {
       viewportOptions.viewportType = STACK;
     }
 
+    if (needsRerendering) {
+      displaySets.forEach(displaySet => {
+        const viewportInfo = CornerstoneViewportService.getViewportInfoByIndex(
+          viewportIndex
+        );
+
+        if (viewportInfo.hasDisplaySet(displaySet.displaySetInstanceUID)) {
```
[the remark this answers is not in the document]
Zaid-Safadi: no need to check. the check already exists in line 284:
`    if (needsRerendering) {`

The needsRerendering flag is in fact `displaySet.needsRerendering` and is passed by the parent control
what we are really doing here is telling the cache to invalidate this displaySet data (imageIds) so next time the viewport requests the displaySet, it gets an updated version.

platform/core/src/services/DisplaySetService/DisplaySetService.js
Zaid-Safadi: instances.filter may work too if it is supported on both "images" and "others".  I think they are a simple arrays so it should but doing a loop will work just fine too.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
Zaid-Safadi: I think the reasoning behind using a warn rather than an error, is because we are handling this and not totally failing the execution of the app. @sedghi might have an opinion on this

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
Zaid-Safadi: I think the "instanceof" will fail. ProtocolGenerator is a type not a class where we create an instance from. In the "getHangingProtocolModule", the protocol generator is defined as a standard function but conforms to this type. 
Not sure if there is a better way to check the method signature is valid before execution

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -1197,6 +1245,29 @@ class HangingProtocolService {
           highestSeriesMatchingScore
         );
 
+        if (imageMatchingRules && imageMatchingRules.length) {
+          const instances = displaySet.images || displaySet.others || [];
+          const filteredInstances = [];
+          for (let instanceIndex = instances.length - 1; instanceIndex >= 0; instanceIndex--) {
```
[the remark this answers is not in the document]
Zaid-Safadi: different people have different coding styles, unless there is a true advantage of using forEach over an for loop, there is nothing wrong with either way

extensions/dicom-video/src/viewports/OHIFCornerstoneVideoViewport.tsx
Zaid-Safadi: I've added a config for autoPlay but will leave the cross-origin out as per documentation, adding this tag will change the behavior regardless which option is picked

platform/core/src/services/HangingProtocolService/HangingProtocolService.test.js
Zaid-Safadi: I agree and I think we should. I've added a TODO for now in the code to do this. I'll try to get to it soon

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -577,7 +591,22 @@ class HangingProtocolService {
       return;
     }
 
-    const protocol = this._validateProtocol(foundProtocol);
+    let protocol;
+    if (foundProtocol instanceof Function) {
+      try {
```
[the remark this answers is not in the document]
Zaid-Safadi: good idea, we already have getProtocolById which I've used here.

Zaid-Safadi:
great comments @jbocce, thanks for the review. I believe I've answered all your comments, let me know if you still have any questions.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
jbocce: I think you are correct. There is not a better way.

Zaid-Safadi:
> Just a comment for future commits... several of the commit comments for this PR do not follow the guidelines. In particular, the doc changes commit comment should be prefixed with "doc:" and the very first commit comment should be prefixed with either "feat:" or "fix:".

rebased and updated commit messages

extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
```diff
@@ -67,34 +68,42 @@ class CornerstoneCacheService {
   }
 
   public async invalidateViewportData(
-    viewportData: VolumeViewportData,
+    viewportData: StackData | VolumeViewportData,
```
sedghi, line 71: why you didn't use the StackViewportData for this? I hate the inconsistensies in the type names here

extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
```diff
@@ -67,34 +68,42 @@ class CornerstoneCacheService {
...
-    }
-
-    // Todo: grab the volume and get the id from the viewport itself
-    const volumeId = `${VOLUME_LOADER_SCHEME}:${invalidatedDisplaySetInstanceUID}`;
+      const displaySet = DisplaySetService.getDisplaySetByUID(invalidatedDisplaySetInstanceUID)
+      if (this.stackImageIds.has(invalidatedDisplaySetInstanceUID)) {
+        this.stackImageIds.delete(invalidatedDisplaySetInstanceUID);
+        return await this._getStackViewportData(
+          dataSource,
+          [displaySet],
```
sedghi, line 82: I was confused to read this at first, then I saw our stackImageIds is a map. Can we rename it to `stackImageIdsMap`?

sedghi: (not your fault)

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -83,6 +83,10 @@ export default function initWADOImageLoader(
   initWebWorkers(appConfig);
 }
 
+function isWadoRs(imageId) {
+  return imageId.indexOf('wadors:') == 0;
```
sedghi, line 87: ===

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -54,20 +54,20 @@ export default function initWADOImageLoader(
...
       // For now we use image/jls and image/x-jls because some servers still use the old type
       // http://dicom.nema.org/medical/dicom/current/output/html/part18.html
-      const xhrRequestHeaders = {
+      const xhrRequestHeaders = isWadoRs(imagId) ? {
         Accept: appConfig.omitQuotationForMultipartRequest
           ? 'multipart/related; type=application/octet-stream'
           : 'multipart/related; type="application/octet-stream"',
         // 'multipart/related; type="image/x-jls", multipart/related; type="image/jls"; transfer-syntax="1.2.840.10008.1.2.4.80", multipart/related; type="image/x-jls", multipart/related; type="application/octet-stream"; transfer-syntax=*',
-      };
+      } : {};
```
sedghi, line 70: please add a comment of why we are checking if wadors

extensions/dicom-video/src/viewports/OHIFCornerstoneVideoViewport.tsx
```diff
@@ -30,6 +32,9 @@ function OHIFCornerstoneVideoViewport({ displaySets }) {
...
+        autoPlay={(appConfig.videoViewport && appConfig.videoViewport.autoPlay) ? appConfig.videoViewport.autoPlay : false}
+        muted={(appConfig.videoViewport && appConfig.videoViewport.autoPlay) ? appConfig.videoViewport.autoPlay : undefined}
```
sedghi, line 37: I don't like how we are adding viewport-specific config now to the config in the config. 
I'm pretty sure the new customizationService by @wayfarer3130 can handle these sort of stuff. No?

platform/core/src/classes/MetadataProvider.js
```diff
@@ -443,17 +443,7 @@ class MetadataProvider {
...
-    // Todo: handle non http imageIds
-    let imageURI;
-    const urlRegex = /^(http|https):\/\//;
-    if (urlRegex.test(imageId)) {
-      imageURI = imageId;
-    } else {
-      imageURI = imageIdToURI(imageId);
-    }
-
+    const imageURI = imageIdToURI(imageId);
```
sedghi, line 446: I'm pretty sure there was a reson we had the regex check, why we don't need it anymore?

platform/core/src/services/DisplaySetService/DisplaySetService.js
```diff
@@ -261,4 +266,45 @@ export default class DisplaySetService {
...
     return allDisplaySets;
   }
+
+
+  filterDisplaySetInstances(displaySetInstanceUID, filteredInstances) {
+    const displaySet = this.getDisplaySetByUID(displaySetInstanceUID);
+    let changed = false;
+
+    if (displaySet) {
+      const instances = displaySet.images || displaySet.others || [];
```
sedghi, line 276: you are donig this || || || a lot, can we make that a private method?

platform/core/src/services/DisplaySetService/DisplaySetService.js
```diff
@@ -261,4 +266,45 @@ export default class DisplaySetService {
     }
     return allDisplaySets;
   }
+
+
+  filterDisplaySetInstances(displaySetInstanceUID, filteredInstances) {
```
sedghi, line 271: These are all public APIs, they should have documentation, AND also please add doc to the DisplaySetService on what is this filtering and how people can use it

platform/core/src/services/DisplaySetService/DisplaySetService.js
```diff
@@ -30,6 +30,7 @@ const findInstance = (instance, displaySets) => {
 
 export default class DisplaySetService {
   constructor() {
+    this._unfilteredDisplaySets = new Map();
```
sedghi, line 33: I would put this at the end of constructor personally (instead of first thing), but not important

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -1202,6 +1233,29 @@ class HangingProtocolService {
...
+              && imagesMatchDetails.details.failed.length === 0) {
+              filteredInstances.push(instance);
+            }
+          }
+
+          DisplaySetService.filterDisplaySetInstances(displaySet.displaySetInstanceUID, filteredInstances);
+          if (instances.length === 0) {
+            return;
+          }
+        }
```
sedghi, line 1257: Can you also please modify the HangingProtocolService doc with an example on what is `imageMatchingRules` and what can be done with it?

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -280,6 +281,25 @@ const OHIFCornerstoneViewport = React.memo(props => {
       viewportOptions.viewportType = STACK;
     }
 
+    if (needsRerendering) {
```
sedghi, line 284: So this block says, if it is `needsRerendering` it means it need to invalidate its stack? I'm confused since that was not the intention for having `needsRerendering` I beleive. The intension originally was when displaySetService metadata is changed via DisplaySetService.EVENTS.DISPLAY_SET_SERIES_METADATA_INVALIDATED, (if metadata is wrong, and user corrects it in the app at runtime) the data should gets invalidated. Can't you just put this logic in the same `useEffect` above?

sedghi (changes requested):
Getting there, please see my comments

sedghi:
Also was thinking `imageMatchingRules` might not be the best name, since it can be apply to non-images like RT, SEG, SR. Right? How about `instanceMatchingRules` ? 

CC @wayfarer3130

platform/viewer/public/config/default.js
```diff
@@ -17,6 +17,9 @@ window.config = {
     thumbnail: 75,
     prefetch: 10,
   },
+  videoViewport: {
+    autoPlay: true
+  },
```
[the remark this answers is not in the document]
Zaid-Safadi: maybe but not sure how that works, first look seems like an overkill tbh but maybe I am missing how it works as I can't see a clear example for reading config values.

platform/core/src/classes/MetadataProvider.js
Zaid-Safadi: because this check introduced a bug by not passing the imageId to the `imageIdToURI` function in the http/https case. the function removes the protocol from the imageId, which is usually wadors:, wadourl:..., and maintains a valid URI, so while the concept of having the regex to verify if it is a standard URL then it doesn't need to "convert" makes sense, that check should be in that method `imageIdToURI` not here. 

For the code here, the imageURI is being used as a lookup in a map for the UIDs, which is maintained in this MetadataProvider class and uses the same approach for adding UIDs (calling `imageIdToURI`) directly without the regex check.

so the options we have:
1. Remove the check here, it should have 0 drawbacks and least effort
2. Do the regex test everywhere needed in the MetadataProvider - not good option at all as you are getting 0 value from that vs. option 1 and more code + effort
3. Update the method `imageIdToURI` to test for the regex. Utlimately, this should be our NorthStar. However, this is a shared method and used several other areas in the code and they all need to be tested if we change the behavior. Something that I don't intend doing in this PR unless there is a reason we need to.

extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
Zaid-Safadi: must have picked the wrong type by mistake

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
Zaid-Safadi: Yes, this block means to invalidate the stack if "needsRendering" is true. 

I am also confused because unless my search function is not working, I am not seeing `needsRendering` used anywhere in the code (although I believe I've seen it a while back when I implemented this).

Either way, I don't see why this property would be only tied to the "metadata" when it changes. In my case, I set this to `true` in the app (custom mode) after a process that involves setting hp, filtering,..., then I set this to true to let the viewport knows it should re-render. Maybe it also can be set to true when the metadata changes then you can maybe get rid off the event subscription but not sure.

For "DISPLAY_SET_SERIES_METADATA_INVALIDATED" it is now fired by the PanelPetSUV in the tmtv extention. 
Note that the viewer didn't use to "invalidate" a stack before anyway as the CacheService didn't support it until this PR.

Right now, it seems to me the two logics while kind of similar, they need to co-exist

Zaid-Safadi:
> Also was thinking `imageMatchingRules` might not be the best name, since it can be apply to non-images like RT, SEG, SR. Right? How about `instanceMatchingRules` ?
> 
> CC @wayfarer3130

I agree. I followed the existing naming but I like `instanceMatchingRules` better

extensions/dicom-video/src/viewports/OHIFCornerstoneVideoViewport.tsx
Zaid-Safadi: oK, I looked at the docs and the service and the intent of the service might be solving for this. However, I am not clear how to configure the config values in the service still. I see there is supposed to be support for it but I don't see the service used anywhere. 
I am thinking I can create a "getCustomizationModule" in the dicom-video extention and set these as default values there so others can extend them. if that's the case it makes sense to me. 

can someone show me what the format of the value looks like to get this setup?

platform/core/src/services/DisplaySetService/DisplaySetService.js
Zaid-Safadi: weird, me too!

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
Zaid-Safadi: sure, I think it is important. added a section to the hpModule docs as well as the DisplayService for the instance level matching and filtering capability.

Zaid-Safadi:
> Getting there, please see my comments

great comments @sedghi !!! I think we are close. I believe the only outstanding item is the customizationService. If I can get some clarification on it hopefully that would be it

platform/viewer/public/config/default.js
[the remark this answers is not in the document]
sedghi: maybe you can take the viewport changes out of this PR then? and we discuss it in another PR? since the rest looks fine to go

extensions/dicom-video/src/viewports/OHIFCornerstoneVideoViewport.tsx
sedghi: @md-prog can we get some help/insight from you here

platform/core/src/classes/MetadataProvider.js
sedghi: I don't understand, hmm. Isn't it the fact that if imageId is `http/https...` then it is already imageURI and no need to be converted?

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
sedghi: Yeah that was necessary a while back and unnecessary now. Ok, how about we change the event to `DISPLAYSET_INVALIDATED` ?

sedghi:
@Zaid-Safadi I answered the unresolved issues. I would suggest to take out the viewport parts and let's scope this PR to only HP

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
wayfarer3130: None of this will be necessary if a NEW display set is created in the filter event  below.

platform/core/src/services/DisplaySetService/DisplaySetService.js
```diff
@@ -261,4 +268,71 @@ export default class DisplaySetService {
...
+  /**
+   * Updates the displaySet to include only the provided
+   * 'filteredInstances' for rendering. This filtering should happen in a
+   * different component (e.g HangingProtocolService), which then calls this
+   * method to update the displaySet and notifies other components
+   * @param {string} displaySetInstanceUID the displaySet UID
+   * @param {any[]} filteredInstances - the filtered array of instances to
+   * be rendered
+  */
+  filterDisplaySetInstances(displaySetInstanceUID, filteredInstances) {
```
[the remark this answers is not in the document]
sedghi: Hmm I guess Bill has a great point here. Nice catch

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -54,20 +54,23 @@ export default function initWADOImageLoader(
...
       // Request:
       // JPEG-LS Lossless (1.2.840.10008.1.2.4.80) if available, otherwise accept
       // whatever transfer-syntax the origin server provides.
       // For now we use image/jls and image/x-jls because some servers still use the old type
       // http://dicom.nema.org/medical/dicom/current/output/html/part18.html
-      const xhrRequestHeaders = {
+      // The Accept header should only be set to a multipart/related type in
+      // wado-rs request. Other requests (e.g. wado-uri) that don't support
+      // this value in the accept header may cause the server to reject the request as unacceptable.
+      const xhrRequestHeaders = isWadoRs(imagId) ? {
```
wayfarer3130, line 68: This is fairly ugly to have the isWadoRs know about the specific type/call.  Also, the configuration should be data source related so that one can configure it as part of hte actual configuration item, rather than globally.  Can you move this into it's own PR so that we can talk about the issue related to omitQuotationForMultipartRequest separately?

extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
```diff
@@ -11,8 +11,8 @@ const VOLUME_IMAGE_LOADER_SCHEME = 'streaming-wadors';
 const VOLUME_LOADER_SCHEME = 'cornerstoneStreamingImageVolume';
 
 class CornerstoneCacheService {
-  stackImageIds: Map<string, string[]> = new Map();
-  volumeImageIds: Map<string, string[]> = new Map();
+  stackImageIdsMap: Map<string, string[]> = new Map();
```
wayfarer3130, line 14: Is it necessary to rename these?

sedghi: it is not a list of imageIds as `stackImageIds` say, it is a map between displaySetInstanceUIDs and their imageIds.

extensions/dicom-video/src/viewports/OHIFCornerstoneVideoViewport.tsx
wayfarer3130: const videoViewportConfig = customizationService.getModeCustomization("videoViewportConfig", {
  autoPlay: false, muted: false })

then just use videoViewportConfig.  You can define a custom version of it in the APP_CONFIG file.

platform/core/src/types/HangingProtocol.ts
```diff
@@ -75,9 +75,9 @@ type ViewportStructure = {
  * it won't be selected.
  */
 type DisplaySetSelector = {
-  // The image matching rule (not currently implemented) selects which image to
-  // display initially, only for stack views.
-  imageMatchingRules?: MatchingRule[];
+  // The instance matching rules selects which instance to
```
wayfarer3130, line 78: The description is correct, but the implementation is NOT agreeing with the description.  The implementation changes the internal representation of a display set, effectively removing elements from the display set.  The selectors are supposed to change what is displayed, not the underlying values, and that is what is causing all the problems with actually changing display sets.   Also, there are other types of bugs introduced by this description because it isn't at all clear that one cannot make selections from multiple series, for example:
Series Matching Rules   has Modality=SR as the match
Instance Matching Rules matches basic DICOM SR
expectation is to match ALL series containing DICOM SR instances.
There are also rules such as match one display set for each DICOM SR without a history object mapping it, which require custom code to implement.

Zaid-Safadi:
> @Zaid-Safadi I answered the unresolved issues. I would suggest to take out the viewport parts and let's scope this PR to only HP

removed the video viewport changes

platform/core/src/types/HangingProtocol.ts
Zaid-Safadi: > The description is correct, but the implementation is NOT agreeing with the description. The implementation changes the internal representation of a display set, effectively removing elements from the display set. The selectors are supposed to change what is displayed, not the underlying values, and that is what is causing all the problems with actually changing display sets. 

This is incorrect. The implementation agrees with the description and the general logic of the hanging protocol matching rules.
You need to look at the `HangingProtocolService. _matchImages function to see the logic, but in summary,  if a "required" matching rule fails on a study, then that study is completely ignored/removed from the display, same thing for series, if a "required" matchingRule fails, the series is removed from the display, for this implementation to support "instanceMatchingRules" the implementation will always assume the rules "Required" and do the same on the instance level and REMOVE instances that fails the "InstanceMatchingRules" from the display

Otherwise, how would you describe the "InstanceMatchingRules" logic to work? 

> Also, there are other types of bugs introduced by this description because it isn't at all clear that one cannot make selections from multiple series, for example:
> Series Matching Rules has Modality=SR as the match
> Instance Matching Rules matches basic DICOM SR
> expectation is to match ALL series containing DICOM SR instances.
> There are also rules such as match one display set for each DICOM SR without a history object mapping it, which require custom code to implement.

I am not sure I completely follow your examples above, @wayfarer3130. How is the current implementation causes a bug in this case?
The SeriesMatchingRule will select the DisplaySets with the Modality=SR, what value is in the "InstanceMatchingRule" that will cause a filtering issue? 
I am not even sure why there would be a use-case where "InstanceMatchingRules"  are applicable with a SeriesMatchingRules of Modality `SR` but even if there is, then this is the user choice for defining these rules.

platform/core/src/services/DisplaySetService/DisplaySetService.js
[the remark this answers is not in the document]
Zaid-Safadi: Create a new displaySet is not filtering an existing displaySet. This function filters an existing displaySet. There is not need for a new displaySet when the user wants to filter an existing one. 

I am sure there are cases when the user may want to have multiple displaySets each displaying some part of a Series, but that is completely different than what is being achieved by this function. 

The DisplaySet is the "View" that will be displayed on the screen, it is not the "Modal", the actual DICOM Metadata that we receive is the Modal that these DisplaySets are created from. We create displaySets to represent what we want to show on the screen the "VIEW" If the user wants to filter the DisplaySet that they want to show on the screen, then they are free to do that. In your solution you assume there is always an "unfiltered" DisplaySet which is not the case here, and it is not the case for HangingProtocol that this is used by.

Why does the user has to create a DisplaySet that contains everything, then another "DisplaySet" that has the actual "View" when they only need to show the filtered one?

extensions/cornerstone/src/initWADOImageLoader.js
Zaid-Safadi: This is fixing a bug that is caused by the code assuming the WADO Image Loader is always requesting a wado-uri. I am not sure why someone would assume this but this check ensure we only add this when wado-rs is being requested. 
Personally, I don't like this piece of code entirely. 
This code should be moved inside the loader where it belongs, having the loader need to be configured like this in every application doesn't make sense.
Any future PRs should focus on refactoring this code.

sedghi:
@matthiasg I guess you can go ahead with the PR. Message me on Slack if you need assitance

matthiasg:
Hi @sedghi did you mean to reference someone else ?

sedghi:
@matthiasg sorry to bother you, yes

PR #3068, opened 2022-12-07
fix(grid):Grid service wasn't being reset

wayfarer3130:
There were a number of interacting issues, which were causing the old display set to be redisplayed.  The easiest way to demonstrate this was to display a video initially and then redisplay it, causing the display sets to get duplicated, but another way was to just show successive studies going back/forth to the studies page and seeing the flash of the old study.

### DIFF
--- a/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
+++ b/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
@@ -136,7 +136,7 @@ class CornerstoneViewportService implements IViewportService {
   public destroy() {
     this._removeResizeObserver();
     this.viewportGridResizeObserver = null;
-    this.renderingEngine.destroy();
+    this.renderingEngine?.destroy?.();
     this.viewportsDisplaySets.clear();
     this.renderingEngine = null;
     cache.purgeCache();
--- a/platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
+++ b/platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
@@ -49,6 +49,9 @@ function createStudyMetadata(StudyInstanceUID) {
       );
 
       if (existingSeries) {
+        // Only add instances not already present, so generate a map
+        // of existing instances and filter the to add by things
+        // already present.
         const sopMap = {};
         existingSeries.instances.forEach(
           it => (sopMap[it.SOPInstanceUID] = it)
--- a/platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
+++ b/platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
@@ -49,7 +49,12 @@ function createStudyMetadata(StudyInstanceUID) {
       );
 
       if (existingSeries) {
-        existingSeries.instances.push(...instances);
+        const sopMap = {};
+        existingSeries.instances.forEach(
+          it => (sopMap[it.SOPInstanceUID] = it)
+        );
+        const newInstances = instances.filter(it => !sopMap[it.SOPInstanceUID]);
+        existingSeries.instances.push(...newInstances);
       } else {
         const series = createSeriesMetadata(instances);
         this.series.push(series);
--- a/platform/core/src/services/DisplaySetService/DisplaySetService.js
+++ b/platform/core/src/services/DisplaySetService/DisplaySetService.js
@@ -194,6 +194,11 @@ export default class DisplaySetService {
     }
   };
 
+  onModeExit() {
+    this.getDisplaySetCache().length = 0;
+    this.activeDisplaySets.length = 0;
+  }
+
   makeDisplaySetForInstances(instancesSrc, settings) {
     let instances = instancesSrc;
     const instance = instances[0];
--- a/platform/core/src/services/DisplaySetService/DisplaySetService.js
+++ b/platform/core/src/services/DisplaySetService/DisplaySetService.js
@@ -53,7 +53,12 @@ export default class DisplaySetService {
     const activeDisplaySets = this.activeDisplaySets;
 
     displaySets.forEach(displaySet => {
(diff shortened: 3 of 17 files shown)


extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -136,7 +136,7 @@ class CornerstoneViewportService implements IViewportService {
   public destroy() {
     this._removeResizeObserver();
     this.viewportGridResizeObserver = null;
-    this.renderingEngine.destroy();
+    this.renderingEngine?.destroy?.();
```
wayfarer3130, line 139: The renderingEngine was null, so the destroy dereference threw an exception.  This caused failures to destroy the rest of the data.

platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
```diff
@@ -49,7 +49,12 @@ function createStudyMetadata(StudyInstanceUID) {
       );
 
       if (existingSeries) {
-        existingSeries.instances.push(...instances);
+        const sopMap = {};
```
wayfarer3130, line 52: The DicomMetadataStore was growing on each subsequent view of the same study, because the instances were being added regardless.

platform/core/src/services/DisplaySetService/DisplaySetService.js
```diff
@@ -194,6 +194,11 @@ export default class DisplaySetService {
     }
   };
 
+  onModeExit() {
```
wayfarer3130, line 197: The display set cache was growing without bound because it wasn't getting cleared.

platform/core/src/services/ViewportGridService/ViewportGridService.ts
```diff
@@ -96,6 +96,10 @@ class ViewportGridService {
     this.serviceImplementation._reset();
   }
 
+  public onModeExit(): void {
```
wayfarer3130, line 99: The reset wasn't being called because it wasn't defined as a function in the API, causing the flash of the old instance (this is the ultimate cause of the overall issue, everything else is related but additional).

platform/ui/src/components/Button/Button.tsx
```diff
@@ -235,6 +235,7 @@ Button.propTypes = {
     'black',
     'inherit',
     'light',
+    'translucent',
```
wayfarer3130, line 238: Just a warning message.

platform/viewer/src/components/ViewportGrid.tsx
```diff
@@ -452,6 +452,7 @@ function _getViewportComponent(
     }
   }
 
+  console.log("Can't show displaySet", SOPClassHandlerId, displaySets[0]);
```
wayfarer3130, line 455: This is useful for debugging - we shouldn't hit this unless things have gone wrong.

md-prog (commented):
@wayfarer3130  I was about to report the same issue:

And from what I found, it was adding the same displaySet again and again to the activeDisplaySets, as it is not checking for any duplicacy.

On my local, I was able to fix it by adding

```
!activeDisplaySets.includes(displaySet) &&
        activeDisplaySets.push(displaySet);
```
to line 56 of `DisplaySetService.js`

platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
sedghi: can you add this comment in the code please? we shouldn't use github for commenting/explaining code if it is required explanation. Something like

`Filter ....`

platform/core/src/services/DisplaySetService/DisplaySetService.js
sedghi: what do you mean? so it was re-creating the same displaySets? we should not create instead. This enforcing of the DS clearance on mode is not a universal thing that every one wants for performance reasons.

sedghi:
> @wayfarer3130 I was about to report the same issue:
> 
> And from what I found, it was adding the same displaySet again and again to the activeDisplaySets, as it is not checking for any duplicacy.
> 
> On my local, I was able to fix it by adding
> 
> ```
> !activeDisplaySets.includes(displaySet) &&
>         activeDisplaySets.push(displaySet);
> ```
> 
> to line 56 of `DisplaySetService.js`

@wayfarer3130 Let's do this as @md-prog has suggested. Otherwise see my comments

platform/core/src/services/DicomMetadataStore/createStudyMetadata.js
wayfarer3130: Done

wayfarer3130:
> > @wayfarer3130 I was about to report the same issue:
> > And from what I found, it was adding the same displaySet again and again to the activeDisplaySets, as it is not checking for any duplicacy.
> > On my local, I was able to fix it by adding
> > ```
> > !activeDisplaySets.includes(displaySet) &&
> >         activeDisplaySets.push(displaySet);
> > ```
> > 
> > to line 56 of `DisplaySetService.js`
> 
> @wayfarer3130 Let's do this as @md-prog has suggested. Otherwise see my comments

@sedghi - I did what was suggested, but it is insufficient, because there are conflicts between calling the notification service and delivering data etc which cause problems.

platform/core/src/services/DisplaySetService/DisplaySetService.js
wayfarer3130: You have to clear the DS values on the end of the mode because there is no guarantee that the same DSes would be created, and there isnt' any good way to check it nor to ensure that the right callbacks are made as the DSes get created.  Creating the DS is relatively cheap once one has the DICOM metadata, even for large numbers of frames/instances, so I don't think that is an issue.

sedghi (approved):
Looks great thanks

Issue #3067, opened 2022-12-06
Eye icon for segment is not shown when segment name is long


fedorov:
See segment 4 in the screenshot below

![image](https://user-images.githubusercontent.com/313942/205953171-deea6897-0d79-4738-958d-c211a281a9ab.png)

https://viewer.imaging.datacommons.cancer.gov/viewer/1.3.6.1.4.1.14519.5.2.1.7311.5101.170561193612723093192571245493?seriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.7311.5101.206828891270520544417996275680,1.2.276.0.7230010.3.1.3.1070885483.15960.1599120307.701

PR #3073, opened 2022-12-12
Re IDC #3067: Show eye-icon when segment label name is long

GitanjaliChhetri:
Fixes https://github.com/OHIF/Viewers/issues/3067

Updated css property to wrap the segment label name.
image


GitanjaliChhetri:
Updated python version to 3.8 in runtime.txt for Netlify rules.

PR #3080, opened 2022-12-15
Fix/sr hydration non tracking

wayfarer3130:
Adds a save and restore capability to the non-tracking mode in OHIF, similar to that in OHIF v2.


extensions/cornerstone-dicom-sr/src/commandsModule.js
```diff
@@ -37,6 +37,12 @@ const _generateReport = (
   // Add in top level series options
   Object.assign(dataset, options);
 
+  // Set the default character set as UTF-8
+  // https://dicom.innolitics.com/ciods/nm-image/sop-common/00080005
+  if (typeof dataset.SpecificCharacterSet === 'undefined') {
+    dataset.SpecificCharacterSet = 'ISO_IR 192';
```
ladeirarodolfo, line 43: any chance for this dataset.SpecificCharacterSet to be null?

wayfarer3130: It will always be null because it isn't one of the fields copied from anywhere else.  Not saying it can't be provided in the future, but right now it will get into the if.

ladeirarodolfo (approved):
LGTM

extensions/default/src/Panels/PanelMeasurementTable.tsx
```diff
@@ -56,6 +68,64 @@ export default function PanelMeasurementTable({
...
 
+  async function clearMeasurements() {
+    MeasurementService.clearMeasurements();
+  }
+
+  async function createReport() {
+    // filter measurements that are added to the active study
+    const activeViewport = viewports[activeViewportIndex];
+    const measurements = MeasurementService.getMeasurements();
+    const displaySet = DisplaySetService.getDisplaySetByUID(
```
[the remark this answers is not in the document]
wayfarer3130: What isn't undefined?  The display set service should be defined, as should measurements.  Did you mean activeViewport?

extensions/cornerstone-dicom-sr/src/viewports/OHIFCornerstoneSRViewport.tsx
```diff
@@ -69,18 +71,25 @@ function OHIFCornerstoneSRViewport(props) {
...
+    // if no panels from measurement-tracking extension is used, this code will trun
+    trackedMeasurements = null;
+    sendTrackedMeasurementsEvent = (eventName, { displaySetInstanceUID }) => {
+      MeasurementService.clearMeasurements();
+      hydrateStructuredReport(
+        { servicesManager, extensionManager },
+        displaySetInstanceUID
+      );
+    };
   }
```
sedghi, line 93: Can you double check if you really need this

sedghi (commented):
Please look into the comment for sr context, otherwise looks great

extensions/cornerstone-dicom-sr/src/viewports/OHIFCornerstoneSRViewport.tsx
wayfarer3130: Yes, that code in the old version sent a tracked measurement event, which internally called the rehydrate event AFTER showing a dialog asking if it should continue tracking.  In the new code, if it is NOT tracked measurements - that is, only if it is actually running without the tracked measurement panel, it directly calls the rehydrate, which is the code that has to be called to actually load the new measurements into the measurement service.

PR #3086, opened 2022-12-21
fix: rework inconsistencies in split button implementation

jbocce:
The primary button will always be the layout grid selector and the MPR toggle is in the drop down menu. This is the first step to later adding other (e.g. 3D) layout HPs to the menu.

Summary of changes/fixes:
- SplitButton now allows for its primary button to be any button component specified by a UI type.
- SplitButton passes a className to its primary button component to dictate how it should be rendered based on its active/toggle status, hovering, menu visibility etc.
- The default menu item renderer for the SplitButton, renders acitve toggle items correctly.
- Items/buttons in the toolbar are now all the correct and same size.
- Added ToolbarService.getButtonComponentForUIType to fetch a button component for a given UI type.
- Spacing between toolbar items is now done at the Toolbar level to avoid double spacing and to ensure uniform spacing.
- The drop down menu closes when clicking the primary button.
- Show a mouse down click effect when clicking the primary button.

Items to investigate still:
The following are all slightly contrary to the design specs of the UI:
- Is the click effect on the primary button satisfactory because with the hover border effect, the button appears slightly smaller during the mouse down/up operation? (MEDIUM)
- On hover over the drop down button, the right-hand-side corners of the primary button appear rounded and NOT square. Also the hover over the drop down button border effects make the primary button appear smaller. (MEDIUM)
- Should there be hover effects when hovering over either tooltip of the SplitButton? (LOW)

### PR Checklist

### DIFF
--- a/extensions/default/src/Toolbar/Toolbar.tsx
+++ b/extensions/default/src/Toolbar/Toolbar.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react';
+import classnames from 'classnames';
 
 export default function Toolbar({ servicesManager }) {
   const { ToolBarService } = servicesManager.services;
@@ -46,15 +47,18 @@ export default function Toolbar({ servicesManager }) {
         // These can... Trigger toolbar events based on updates?
         // Then sync using useEffect, or simply modify the state here?
         return (
-          <Component
-            key={id}
-            id={id}
-            {...componentProps}
-            bState={buttonState}
-            isActive={isActive}
-            onInteraction={args => ToolBarService.recordInteraction(args)}
-            servicesManager={servicesManager}
-          />
+          // The margin for separating the tools on the toolbar should go here and NOT in each individual component (button) item.
+          // This allows for the individual items to be included in other UI components where perhaps alternative margins are desired.
+          <div key={id} className={classnames('mr-1')}>
+            <Component
+              id={id}
+              {...componentProps}
+              bState={buttonState}
+              isActive={isActive}
+              onInteraction={args => ToolBarService.recordInteraction(args)}
+              servicesManager={servicesManager}
+            />
+          </div>
         );
       })}
     </>
--- a/extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
+++ b/extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
@@ -55,6 +55,24 @@ function LayoutSelector({
   const onInteractionHandler = () => setIsOpen(!isOpen);
   const DropdownContent = isOpen ? OHIFLayoutSelector : null;
 
+  const onSelectionHandler = ({ numRows, numCols }) => {
+    if (HangingProtocolService.getActiveProtocol().protocol.id === 'mpr') {
+      ToolBarService.recordInteraction({
+        groupId: 'MPR',
+        itemId: 'MPR',
+        interactionType: 'toggle',
+        commands: [
+          {
+            commandName: 'toggleMPR',
+            commandOptions: {},
+            context: 'CORNERSTONE',
+          },
+        ],
+      });
+    }
+    viewportGridService.setLayout({ numRows, numCols });
+  };
+
   return (
     <ToolbarButton
       id="Layout"
@@ -68,26 +86,7 @@ function LayoutSelector({
           <DropdownContent
             rows={rows}
             columns={columns}
-            onSelection={({ numRows, numCols }) => {
-              if (
-                HangingProtocolService.getActiveProtocol().protocol.id === 'mpr'
-              ) {
-                ToolBarService.recordInteraction({
-                  groupId: 'MPR',
-                  itemId: 'MPR',
-                  interactionType: 'toggle',
-                  commands: [
-                    {
-                      commandName: 'toggleMPR',
-                      commandOptions: {},
-                      context: 'CORNERSTONE',
-                    },
-                  ],
-                });
-              }
-
-              viewportGridService.setLayout({ numRows, numCols });
-            }}
+            onSelection={onSelectionHandler}
           />
         )
       }
--- a/extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
+++ b/extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
@@ -56,6 +56,10 @@ function LayoutSelector({
   const DropdownContent = isOpen ? OHIFLayoutSelector : null;
 
   const onSelectionHandler = ({ numRows, numCols }) => {
+    // TODO Introduce a service to persist the state of the current hanging protocol/app.
+
+    // TODO Here the layout change will amount to a change of hanging protocol as specified by the extension for this layout selector tool
+    // followed by the change of the grid itself.
     if (HangingProtocolService.getActiveProtocol().protocol.id === 'mpr') {
       ToolBarService.recordInteraction({
         groupId: 'MPR',
--- a/extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
+++ b/extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
@@ -6,7 +6,13 @@ import {
   useViewportGrid,
 } from '@ohif/ui';
 
-function LayoutSelector({ rows, columns, className, servicesManager }) {
+function LayoutSelector({
+  rows,
+  columns,
+  className,
+  servicesManager,
+  ...rest
+}) {
   const [isOpen, setIsOpen] = useState(false);
   const [disableSelector, setDisableSelector] = useState(false);
   const [viewportGridState, viewportGridService] = useViewportGrid();
@@ -56,6 +62,7 @@ function LayoutSelector({ rows, columns, className, servicesManager }) {
       icon="tool-layout"
       onInteraction={onInteractionHandler}
       className={className}
+      rounded={rest.rounded}
       dropdownContent={
         DropdownContent !== null && (
           <DropdownContent
--- a/extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
+++ b/extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
@@ -6,12 +6,12 @@ import {
   useViewportGrid,
 } from '@ohif/ui';
 
-function LayoutSelector({ rows, columns, servicesManager }) {
+function LayoutSelector({ rows, columns, className, servicesManager }) {
   const [isOpen, setIsOpen] = useState(false);
   const [disableSelector, setDisableSelector] = useState(false);
   const [viewportGridState, viewportGridService] = useViewportGrid();
 
-  const { HangingProtocolService } = servicesManager.services;
+  const { HangingProtocolService, ToolBarService } = servicesManager.services;
 
   const closeOnOutsideClick = () => {
     if (isOpen) {
@@ -24,12 +24,6 @@ function LayoutSelector({ rows, columns, servicesManager }) {
       HangingProtocolService.EVENTS.PROTOCOL_CHANGED,
       evt => {
         const { protocol } = evt;
-
-        if (protocol.id === 'mpr') {
-          setDisableSelector(true);
-        } else {
-          setDisableSelector(false);
-        }
       }
     );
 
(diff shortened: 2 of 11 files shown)


extensions/default/src/Toolbar/Toolbar.tsx
```diff
@@ -46,15 +47,18 @@ export default function Toolbar({ servicesManager }) {
...
-            id={id}
-            {...componentProps}
-            bState={buttonState}
-            isActive={isActive}
-            onInteraction={args => ToolBarService.recordInteraction(args)}
-            servicesManager={servicesManager}
-          />
+          // The margin for separating the tools on the toolbar should go here and NOT in each individual component (button) item.
+          // This allows for the individual items to be included in other UI components where perhaps alternative margins are desired.
+          <div key={id} className={classnames('mr-1')}>
```
jbocce, line 52: This fixes the inconsistent spacing that was occurring is the previous rendition of the toolbar. Note that the spec says that there should be 4px between toolbar items.

extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
```diff
@@ -61,12 +55,30 @@ function LayoutSelector({ rows, columns, servicesManager }) {
...
       icon="tool-layout"
       onInteraction={onInteractionHandler}
+      className={className}
       dropdownContent={
         DropdownContent !== null && (
           <DropdownContent
             rows={rows}
             columns={columns}
             onSelection={({ numRows, numCols }) => {
+              if (
```
jbocce, line 65: I think my prettier formatter plugin put the if condition on a new line. Thoughts?

extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
```diff
@@ -61,12 +55,30 @@ function LayoutSelector({ rows, columns, servicesManager }) {
...
+                HangingProtocolService.getActiveProtocol().protocol.id === 'mpr'
+              ) {
+                ToolBarService.recordInteraction({
```
jbocce, line 68: Instead of disabling the layout selector during MPR, toggle MPR off before switching to the new layout. Eventually this is where we can apply the previous state with the new layout.

platform/ui/src/components/SplitButton/SplitButton.tsx
```diff
@@ -31,11 +29,17 @@ const classes = {
   Primary: ({ primary, isExpanded }) =>
     classNames(
       baseClasses.Primary,
-      primary.isActive && !isExpanded
-        ? 'bg-primary-light rounded-tr-md rounded-br-md active'
+      primary.isActive
```
jbocce, line 32: The className for the primary button is calculated here and passed down to the button as a prop.

platform/ui/src/components/SplitButton/SplitButton.tsx
```diff
@@ -107,7 +105,7 @@ const SplitButton = ({
 
         setState(state => ({
           ...state,
-          primary: !isAction ? { ...item, index } : state.primary,
+          primary: !isAction && isRadio ? { ...item, index } : state.primary,
```
jbocce, line 108: We can now say isRadio=false for the split button and it will work. Yay! I guess this was another fix that I forgot to put in my git commit message.

jbocce:
@sedghi perhaps we can discuss the "Items still to investigate" listed at the top. Thanks so much.

extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
wayfarer3130: I would suggest creating the callback function up above, and then just assigning it here.  That way you create it in a context that has more space and shouldn't break the spacing here.  IMHO, prettier produces extremely ugly code at times.

extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
wayfarer3130: It isn't just MPR that we want to be able to toggle, but there are other hanging protocols which have the same sort of behaviour such as multi-study comparison.  Is there some way we can use the hanging protocol service to record any "toggled on" hanging protocols, and their previous state?  That way, we can toggle any hanging protocol on if desired, and switch between them.  For example a 3d rendering of a volume, a 1x1 stack mode, an mpr mode, a compare to previous series with the same description (in PR right now) etc all want the "toggle" capability.

platform/ui/src/components/SplitButton/SplitButton.tsx
wayfarer3130: This is getting quite ugly - this needs some way to document the behaviour and why the various behaviour is as it is so that in the future people can figure out how to fix bugs in it.

platform/ui/src/components/SplitButton/SplitButton.tsx
wayfarer3130: Thats useful to have.

extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
```diff
@@ -61,12 +61,31 @@ function LayoutSelector({ rows, columns, servicesManager }) {
...
+                    {
+                      commandName: 'toggleMPR',
+                      commandOptions: {},
+                      context: 'CORNERSTONE',
+                    },
+                  ],
+                });
+              }
+
               viewportGridService.setLayout({ numRows, numCols });
```
wayfarer3130, line 89: I think the right way to do this is to call recordInteraction on the grid layout button, and that should consist of two items, first a toggleMPR item, and second setLayout command.  That way, modes can configure additional custom behaviour if they want, and it avoids needing to know about any special behaviour here.  See more comments below.

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -295,26 +296,41 @@ const toolbarButtons = [
...
-        {
-          commandName: 'toggleMPR',
-          commandOptions: {},
-          context: 'CORNERSTONE',
+      isRadio: false,
+      groupId: 'Layouts',
+      primary: {
+        id: 'Layout',
+        type: 'action',
+        uiType: 'ohif.layoutSelector',
```
wayfarer3130, line 307: Here you can add a command list of things to execute when the layout is toggled.  The first item should be one to
turn off any toggled layouts (eg toggleMPR for now), while the second should be a new command that simply calls setLayout(rows,columns) after getting rows/columns from the arguments.
That allows in the future to do things like have custom layouts configured in the hanging protocol and switching to those - for example, a 1x1 layout might just "toggle" the viewport to a single volume view 1x1 rather than  switching away from MPR mode.

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -295,26 +296,41 @@ const toolbarButtons = [
...
+          columns: 3,
         },
+      },
+      secondary: {
+        icon: 'chevron-down',
+        label: '',
+        isActive: true,
+        tooltip: 'More Layouts',
+      },
+      items: [
```
wayfarer3130, line 319: Eventually we will need a computed child list rather than a static one - something that can list various stages as well as other hanging protocols.  Should think about how to handle that, but probably it doesn't belong here right now.

platform/ui/src/components/SplitButton/SplitButton.tsx
jbocce: I agree it is quite ugly, but I would say it was not much prettier before. The intended behaviour is described by the OHIF specs laid out in Zeplin and this code fulfills that behaviour. Is that document enough to guide bug fixing?

Unfortunately the complexity is a result of how intertwined the look-and-feel behaviour of the two buttons have been defined in the OHIF specs. To be honest here, my past dealings with split buttons is that in essence the two buttons are separate buttons when it comes to look and feel. A quick look around at bootstrap and React MUI (https://mui.com/material-ui/react-button-group/#split-button) further confirms this. For example, I would say in general it is very rare for a split button that hovering over one of the two buttons affects the look of the other. However, the OHIF spec gives the primary button rounded corners on the left and square corners on the right when hovering over the secondary, but the primary has rounded corners when active. Now that my change allows for a separate button to be the primary it makes the logic here difficult, but doable.

If the two buttons were a bit more independent look-and-feel wise, this ugliness would go away. Thoughts?

platform/ui/src/components/SplitButton/SplitButton.tsx
jbocce: Thanks.

extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
jbocce: Yes agreed. This set of commits (and PR) was just "step one" in a series of changes that will eventually be adding other layouts to the drop down menu. So this logic here is by no means permanent. 

This set of commits was to transition the split button to be more flexible (previously it was not easy to add a custom button as the primary) and do a proof of concept whereby the MPR button was put into the drop down menu. I did not explain this in the PR since I figured it would be @sedghi doing the review and he understood this.

The actual solution (i.e. next step) discussed with @sedghi is to introduce a StateSyncService that will preserve various states of the app (e.g. HP, the viewports and their contents, window level, etc.) when an HP is applied. When a layout change is requested, instead of doing a recordInteraction to toggle whatever layout existed prior, the previous state is  restored and the new layout applied to that old state.

I hope this addresses this and the concerns/comments that follow. Does it?

modes/longitudinal/src/toolbarButtons.js
jbocce: Yes, perhaps later.

extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
jbocce: See previous comment [regarding StateSyncService.](https://github.com/OHIF/Viewers/pull/3086#discussion_r1062566151)

modes/longitudinal/src/toolbarButtons.js
jbocce: See previous comment [regarding StateSyncService.](https://github.com/OHIF/Viewers/pull/3086#discussion_r1062566151)

extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
jbocce: Great suggestion. Thanks. I have just pushed in a change for this.

modes/longitudinal/src/toolbarButtons.js
wayfarer3130: OK, it can go in a separate PR - I agree, the general change layout solution needs to be thought out more carefully.

platform/ui/src/components/SplitButton/SplitButton.tsx
wayfarer3130: Can you create an issue to separate out the buttons?  I think that would definitely make it cleaner overall - it was ugly before and is just as ugly now.

extensions/default/src/Toolbar/ToolbarLayoutSelector.tsx
```diff
@@ -55,20 +55,38 @@ function LayoutSelector({ rows, columns, servicesManager }) {
   const onInteractionHandler = () => setIsOpen(!isOpen);
   const DropdownContent = isOpen ? OHIFLayoutSelector : null;
 
+  const onSelectionHandler = ({ numRows, numCols }) => {
+    if (HangingProtocolService.getActiveProtocol().protocol.id === 'mpr') {
```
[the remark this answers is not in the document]
jbocce: Sure thing.

platform/ui/src/components/SplitButton/SplitButton.tsx
jbocce: I am not sure that a simple issue would do here. Separating the buttons amounts to a change in design/spec. I am not sure we want to change the spec any time soon. Thoughts @sedghi ?

modes/longitudinal/src/toolbarButtons.js
jbocce: Ok. I added a bunch of TODOs. In essence, I agree with you. Changing grid layout amounts to: saving the current app/HP state; switching to some hanging protocol (likely default) and then changing to the grid selected.  Currently the toggleMPR command does the state saving and restoring, but it shouldn't.

jbocce: We can discuss this further, but I think maybe in the next PR the HangingProtocolService should (perhaps via another service) save the current app/hp state when switching to another hp/stage. Furthermore when switching back to a previously used hp/stage, that previous state should be restored.

sedghi (changes requested):
This is a great PR fixing a lot of inconsistencies with our toolbar button.

Having just one layout right now (MPR), let's move that back outside, until we have more "common" layouts, so that people don't get confused. 

We should have a talk with Dan for the design below, but this PR solves couple of bugs as you saw for the drop down.  

image

jbocce:
> This is a great PR fixing a lot of inconsistencies with our toolbar button.
> 
> Having just one layout right now (MPR), let's move that back outside, until we have more "common" layouts, so that people don't get confused.
> 
> We should have a talk with Dan for the design below, but this PR solves couple of bugs as you saw for the drop down.
> 
> image

Thanks and I will update.

sedghi (approved):

PR #3090, opened 2022-12-30
feat:Preserve viewport contents/location on layout change

wayfarer3130:
Fix the viewport content/location values.


wayfarer3130:
@sedghi  - I don't have this one working fully yet - mostly it works, except that on returning to the original layout I start seeing exceptions.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -96,7 +96,7 @@ class CornerstoneViewportService implements IViewportService {
   }
 
   public getViewportId(viewportIndex: number): string {
-    return `viewport-${viewportIndex}`;
+    return this.viewportsInfo[viewportIndex]?.getViewporId?.();
```
sedghi, line 99: This has a typo

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -199,7 +199,10 @@ class CornerstoneViewportService implements IViewportService {
       newViewportId = publicViewportOptions.viewportId;
       viewportInfo.setViewportId(newViewportId);
 
-      renderingEngine.disableElement(viewportId);
+      if (viewportId) {
+        console.log('Would disable element', viewportId);
+        // renderingEngine.disableElement(viewportId);
```
sedghi, line 204: why this is commented out? if the viewportId changes, we would want to disable the previous one. Example: when we switch hanging protocol e.g., MPR

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -89,35 +83,44 @@ export function ViewportGridProvider({ children, service }) {
...
-            yPos = row * h;
+        const hasOptions = layoutOptions?.length;
+        const viewports = [];
+        let activeViewportIndex;
+
+        for (let col = 0; col < numCols; col++) {
+          for (let row = 0; row < numRows; row++) {
+            const pos = col + row * numCols;
+            const layoutOption = layoutOptions[pos];
+            const viewportId = layoutOption?.viewportId || `${col},${row}`;
```
sedghi, line 94: I think `viewport-1-2` is more readable than `1,2`

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -238,14 +241,15 @@ export function ViewportGridProvider({ children, service }) {
...
-    ({ layoutType, numRows, numCols, layoutOptions = [] }) =>
+    ({ layoutType, numRows, numCols, layoutOptions = [], newLayout = {} }) =>
       dispatch({
         type: 'SET_LAYOUT',
         payload: {
           layoutType,
           numRows,
           numCols,
           layoutOptions,
+          newLayout,
```
sedghi, line 252: newLayout is not used any where in the code, do you need it?

sedghi (changes requested):
see my comments

sedghi:
Also tried it and it has a bug for going back and forth to mpr. I think it is related to you not disabling the viewport when the id changes

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
wayfarer3130: Where?  The ?. is intentional to avoid calling an undefined value.

platform/ui/src/contextProviders/ViewportGridProvider.tsx
wayfarer3130: Removed

platform/ui/src/contextProviders/ViewportGridProvider.tsx
wayfarer3130: Changed.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
wayfarer3130: Changed

wayfarer3130:
Need to rethink this a while, see if a solution can be found.  Will re-open if appropriate.

sedghi:
@wayfarer3130 Could you please summarize the issued you found for this so far?

PR #3097, opened 2023-01-05
fix: multiframe image prefetch logic

slak44:
When the StudyPrefetcher is enabled, it prefetches the images of the current stack (and a prop is passed to react-cornerstone-viewport to stop it from prefetching).

Unfortunately, the code in StudyPrefetcher that computed which imageIds should be fetched was broken for multiframe images. As a result, nothing except the current image got loaded (triggering a load on every stack scroll, which is not great).

The first issue was that StudyPrefetcher used a non existent property `numFrames`, which made it think all images have a single frame (`undefined > 1` is false).
The second issue was that the getImageId standalone function delegates to InstanceMetadata#getImageId, but without passing along its arguments (the `frame` argument, in particular).
The third issue was that InstanceMetadata#getImageId cached its return value, but that's not ok to do when the frame argument can vary.

### DIFF
--- a/platform/core/src/classes/StudyPrefetcher.js
+++ b/platform/core/src/classes/StudyPrefetcher.js
@@ -494,7 +494,7 @@ export class StudyPrefetcher {
 
     // TODO: This duplicates work done by the stack manager
     displaySet.images.forEach(image => {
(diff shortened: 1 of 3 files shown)


sedghi:

PR #3102, opened 2023-01-09
fix: OHIF-386 Disable Export and Create Report buttons if there are no tracked measurements

jbocce:
Use the disabled Button component variant when no tracked measurements are present. Made the buttons look as per the spec.

### PR Checklist


jbocce:
> This has a change to the button UI (export color)
> 
> before
> 
> image
> 
> now image
> 
> I guess you were following the design in Zeplin
> 
> image
> 
> But there is this one too
> 
> image
> 
> So I guess you need to ask Dan which one is it

Yes, I spoke with Dan on Friday and furthermore you, Dan, James and I met today and confirmed the changes. In particular:
- the two buttons should have the same look (prior to this PR they did not)
- the drop down is likely from an older version of the spec where more than one export type was possible (now there is only one)
- using 50% opacity and 'not allowed' cursor for disabled buttons is acceptable

sedghi (approved):

PR #3116, opened 2023-01-13
fix: Use @cornerstonejs/adapters for rehydrating structure reports, and sort the saved SR objects to after image display sets in reverse order

wayfarer3130:
This change uses the @cornerstonejs/adapters for loading and saving of DICOM SR.  
Also, fixes the import issue preventing restore of DICOM SR in tracking mode, and fixes the sorting issue where the DICOM series were NOT being sorted as expected, but were just being displayed in incoming order.  The sort order expected is:
High priority series (those with images and for immediate display)
   * Sort by increasing series number, then Series Date/Time (increasing)
Low priority series (SEG, SR, KO, GSPS etc)
   * Sort by decreasing series number, then decreasing Series Data/Time

This sort order puts the rehydrated DICOM SR AFTER the images, but with the most recent one first on the list.


sedghi (approved):
Looks good. Please update the PR description and explain the sortStudy changes

PR #3136, opened 2023-01-27
feat(Calibration Tool): add a calibration tool that can manually recalibrate an image and measurements on it

md-prog:
### Context

Manual Calibration Tool implementation.
A DICOM image usually have calibration information (or pixel spacing information) that represents the mapping of real-world size and the pixels in the image. Sometimes this information is not available and even when it is available, we may want to manually re-calibrate the image.

#### UX

- Similar to the line(length) tool
- but once the line is drawn, a popup shows up to collect input about the physical distance represented by the line.
- Click OK, then the viewport and all the measurements will be updated

#### Notes

- The re-calibration performed by this tool would be temporary and won't be persisted/uploaded back to the source image yet.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 10
- [x] Node version: 16.15.0
- [x] Browser: Chrome 109.0.5414.121 (Official Build) (64-bit)

### DIFF
--- a/extensions/cornerstone/src/initMeasurementService.js
+++ b/extensions/cornerstone/src/initMeasurementService.js
@@ -144,7 +144,10 @@ const connectToolsToMeasurementService = servicesManager => {
       } = annotationAddedEventDetail;
       const { toolName } = metadata;
 
-      if (toolName === toolNames.CalibrationLine) {
+      if (
+        csToolsEvent.type === completedEvt &&
+        toolName === toolNames.CalibrationLine
+      ) {
         // show modal to input the measurement (mm)
         onCompletedCalibrationLine(servicesManager, csToolsEvent)
           .then(
--- a/extensions/cornerstone/src/initMeasurementService.js
+++ b/extensions/cornerstone/src/initMeasurementService.js
@@ -1,6 +1,8 @@
 import { eventTarget } from '@cornerstonejs/core';
 import { Enums, annotation } from '@cornerstonejs/tools';
-import { DicomMetadataStore, MeasurementService } from '@ohif/core';
+import { DicomMetadataStore } from '@ohif/core';
+import { toolNames } from './initCornerstoneTools';
+import { onCompletedCalibrationLine } from './tools/CalibrationLineTool';
 
 import measurementServiceMappingsFactory from './utils/measurementServiceMappings/measurementServiceMappingsFactory';
 import getSOPInstanceAttributes from './utils/measurementServiceMappings/utils/getSOPInstanceAttributes';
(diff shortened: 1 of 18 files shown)


extensions/cornerstone/src/initMeasurementService.js
```diff
@@ -66,14 +68,25 @@ const initMeasurementService = (
...
 
+  // On the UI side, the Calibration Line tool will work almost the same as the
+  // Length tool
+  MeasurementService.addMapping(
+    csTools3DVer1MeasurementSource,
+    'CalibrationLine',
+    Length.matchingCriteria,
+    Length.toAnnotation,
+    Length.toMeasurement
+  );
```
sedghi, line 79: I don't understand why you are adding it to the measurementService? The tools added here store state in the measurementService, and need mappings. But to me, the calibration tool is a tool that runs a call back on finish. Do we really need mapping here?

extensions/cornerstone/src/initMeasurementService.js
```diff
@@ -95,15 +108,35 @@ const connectToolsToMeasurementService = (
...
+            },
+            () => true
+          )
+          .finally(() => {
+            // we don't need the calibration line lingering around, remove the
+            // annotation from the display
+            removeAnnotation(annotationUID);
+            removeMeasurement(csToolsEvent);
+            // this will ensure redrawing of annotations
+            CornerstoneViewportService.resize();
```
sedghi, line 126: Yeah I thought you would do this, so we don't need the measurement service mapping.

extensions/cornerstone/src/tools/CalibrationLineTool.ts
```diff
@@ -0,0 +1,127 @@
...
+import getActiveViewportEnabledElement from '../utils/getActiveViewportEnabledElement';
+
+/**
+ * Calibration Line tool works almost the same as the
+ */
+class CalibrationLineTool extends LengthTool {
+  static toolName = 'CalibrationLine';
+
+  _renderingViewport: any;
+  _lengthToolRenderAnnotation = this.renderAnnotation;
```
sedghi, line 14: interesting

sedghi:
See my comment. I don't think you need to add it to measurement service since it is not getting displayed in the measurement panel, nore getting exported right?

extensions/cornerstone/src/initMeasurementService.js
md-prog: You're right, the code is not necessary

arturojain:
What OHIF version does this PR support?

sedghi (approved):
This looks good to me, please rebase and we can merge

md-prog:
@sedghi I am sorry for the delay. Rebased, fixed (some code changes required after rebase), and tested.

PR #3138, opened 2023-02-02
fix(Segmentation): bump packages to fix segmentation rendering

sedghi:
It bump packages to fix seg rendering

+ it adds a test mode which later will get used in cypress to test hanging protocols + other features inside a mode
+ Changing the numberOfWorkers from 4 to 3 since some windows systems have issues with more than 3 web workers
+ Add ReferenceLines to basic mode

### DIFF
--- a/modes/longitudinal/src/initToolGroups.js
+++ b/modes/longitudinal/src/initToolGroups.js
@@ -175,7 +175,10 @@ function initMPRToolGroup(extensionManager, ToolGroupService, commandsManager) {
       { toolName: toolNames.Angle },
       { toolName: toolNames.SegmentationDisplay },
     ],
-    disabled: [{ toolName: toolNames.Crosshairs }],
+    disabled: [
+      { toolName: toolNames.Crosshairs },
+      { toolName: toolNames.ReferenceLines },
+    ],
 
     // enabled
     // disabled
--- a/extensions/cornerstone-dicom-sr/package.json
+++ b/extensions/cornerstone-dicom-sr/package.json
@@ -45,7 +45,7 @@
   "dependencies": {
     "@babel/runtime": "7.16.3",
     "classnames": "^2.2.6",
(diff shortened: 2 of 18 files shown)


jbocce:
I checked out the deploy preview and got into a situation where MPR was no longer working. I had the following exception in the console. However, I have not been able to reproduce this behaviour since. I am including it here as an FYI, but I will try again in a bit.

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getImageLoadRequests')
    at nthLoader.ts:78:27
    at Array.map (<anonymous>)
    at dw (nthLoader.ts:78:6)
    at e.value (HangingProtocolService.ts:390:12)
    at e.<anonymous> (CornerstoneViewportService.ts:456:37)
    at c (CornerstoneViewportService.ts:2:1)
    at Generator._invoke (CornerstoneViewportService.ts:2:1)
    at Generator.next (CornerstoneViewportService.ts:2:1)
    at BL (CornerstoneViewportService.ts:2:1)
    at i (CornerstoneViewportService.ts:2:1)
```
image

modes/longitudinal/src/initToolGroups.js
```diff
@@ -175,7 +175,10 @@ function initMPRToolGroup(extensionManager, ToolGroupService, commandsManager) {
       { toolName: toolNames.Angle },
       { toolName: toolNames.SegmentationDisplay },
     ],
-    disabled: [{ toolName: toolNames.Crosshairs }],
+    disabled: [
+      { toolName: toolNames.Crosshairs },
+      { toolName: toolNames.ReferenceLines },
```
jbocce, line 180: Please add this information in the summary/description of the PR.

jbocce (approved):
Other than a few suggestions, I think it looks good.

jbocce (approved):
Changes look good.

PR #3140, opened 2023-02-03
feat(SR): Measurement adapter mappings and save to same series

wayfarer3130:
Add mappings for RectangleROI and Angle to the tools
Save the DICOM SR to the same series as the first found SR series with the same description to allow updating an SR
Add partial mappings for Cobb Angle, Planar Freehand, both of which have some issues but at least allow showing up in the measurements table even if they fail to restore correctly.


extensions/cornerstone/src/utils/measurementServiceMappings/utils/selection.ts
```diff
@@ -0,0 +1,44 @@
+import { annotation as cs3dToolAnnotationUtils } from '@cornerstonejs/tools';
```
wayfarer3130, line 1: The code in the 3 utils classes is used by the freehand mapping to centralize how the mapping is defined/handled.  Probably we will use these utilities further in the future.

extensions/cornerstone/src/index.tsx
```diff
@@ -150,4 +161,19 @@ const cornerstoneExtension: Types.Extensions.Extension = {
...
+  getEnabledElement,
+  annotation: {
+    setAnnotationColor,
+    setAnnotationLineDash,
+    getAnnotationColor,
+    setAnnotationSelected,
+    setAnnotationVisibility,
+    isAnnotationSelected,
+  },
+};
```
sedghi, line 176: why we are exporting these like this instead of `getUtilityModule`? What is the difference

extensions/cornerstone/src/registerCustomCornerstoneTools.ts
```diff
@@ -0,0 +1,132 @@
...
+const { BaseTool } = csTools;
+
+/**
+ * Object to store registered Custom Tools (classes)
+ */
+const customTools = {};
+
+/**
+ * Auxiliary class to store original customTools/toolNames to be restore on a fallback approach.
+ */
```
sedghi, line 14: Please add a lot more explanation here. I don't understand what is the reason for this Class? please explain one or two usecases

extensions/cornerstone/src/registerCustomCornerstoneTools.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.toolNames = {};
+  }
+
+  setTo(toolName, from, to) {
+    if (from[toolName]) {
+      to[toolName] = from[toolName];
+    }
+  }
+
+  addToCache(CustomTool) {
```
sedghi, line 27: this can be just `add` since the className is `customToolCache` so it will be `customToolCache.add` instead of `customToolCache.addToCache` same for revert

extensions/cornerstone/src/registerCustomCornerstoneTools.ts
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Auxiliary class to store original customTools/toolNames to be restore on a fallback approach.
+ */
+class CustomToolCache {
+  constructor() {
+    this.customTools = {};
+    this.toolNames = {};
+  }
+
+  setTo(toolName, from, to) {
```
sedghi, line 21: make this private

extensions/cornerstone-dicom-sr/src/commandsModule.js
```diff
@@ -29,15 +29,13 @@ const _generateReport = (
 
   const report = MeasurementReport.generateReport(
     filteredToolState,
-    metaData,
-    utilities.worldToImageCoords
+    OHIF.classes.MetadataProvider,
```
sedghi, line 32: metaData was correct since it uses .get and it will act based on priority. Why you are forcing it to use the OHIF one? it should be based on priorities like elseWhere

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -105,6 +105,7 @@ function _getDisplaySetsFromSeries(
     StudyInstanceUID,
     SOPClassHandlerId,
     SOPClassUID,
+    others: instances,
```
sedghi, line 108: is `others` a well-known dicom thing? it is too general. `instances` instead?

extensions/default/src/Panels/PanelMeasurementTable.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
+import { ServicesManager } from '@ohif/core';
```
sedghi, line 3: you had the `@ohif/core` couple of lines below, import from that

extensions/default/src/utils/getSameSeriesOptions.ts
```diff
@@ -0,0 +1,38 @@
+import { DisplaySetService } from '@ohif/core';
+
+import getNextSRSeriesNumber from './getNextSRSeriesNumber';
+
+export default function getSameSeriesOptions(
+  SeriesDescription: string,
+  displaySetService: DisplaySetService
+): Record<string, unknown> {
+  const activeDisplaySets = displaySetService.getActiveDisplaySets();
+  const srDisplaySets = activeDisplaySets.filter(ds => ds.Modality === 'SR');
```
sedghi, line 10: When I read the function name i didn't get the idea it is SR only, please rename this function

extensions/default/src/utils/getSameSeriesOptions.ts
```diff
@@ -0,0 +1,38 @@
...
+  const sameSeries = srDisplaySets.find(
+    ds => ds.SeriesDescription === SeriesDescription
+  );
+  if (sameSeries) {
```
sedghi, line 14: I prefer less indentation, you can say

```
if (!sameSeries){
  return
}

// rest of logic
```

sedghi: oh you can't do it here, nevermind

extensions/default/src/utils/getSameSeriesOptions.ts
```diff
@@ -0,0 +1,38 @@
+import { DisplaySetService } from '@ohif/core';
+
+import getNextSRSeriesNumber from './getNextSRSeriesNumber';
+
+export default function getSameSeriesOptions(
```
sedghi, line 5: please add documentation above

platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -529,6 +526,7 @@ class MeasurementService extends PubSubService {
       measurement = toMeasurementSchema(sourceAnnotationDetail);
       measurement.source = source;
     } catch (error) {
+      console.log('Failed to map', error);
```
sedghi, line 529: give more information 
`MeasurementService::annotation ........`

extensions/default/src/utils/getSameSeriesOptions.ts
sedghi: options doesn't make sense for the output

sedghi (changes requested):
See my comments

# Seems like the SRDisplayTool is broken with the new additions:

When exporting the angle, the SR viewports shows weird (I drew angle and exported this is what was shown after)

image

The cobb angle export also has some weird intermediate display

image

The freehand shows 
image


# It is pretty wierd having multiple freehands
since we don't have stats, i don't know what is the best way

image

# Freehand doesn't hydrate gives error

Export Freehand SR, refresh the page, the import again => error

image

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
wayfarer3130: others is what is used consistently in the getSopClassHandler modules for things other than images.  I agree that instances would be nicer, but would need to rename a bunch of them.

platform/core/src/services/MeasurementService/MeasurementService.ts
wayfarer3130: The stack trace has all that information, this is just to be able to debug.

extensions/cornerstone/src/index.tsx
wayfarer3130: If they are exported like that, then they are accessible directly with correct typing.  The getUtilityModule export works, but doesn't take advantage of direct linking to things.  Some items should be exported direct, and others in getUtilityModule.

wayfarer3130:
Still can't get the CobbAngle to render correctly on initial display, or the Freeform to load after saving (from the SR panel to the measurements), but the rest of them seem pretty good now.
To run, it DOES require an updated DICOM SR module, which is a dependency of this, so I'm not removing the WIP until that is gone.

extensions/cornerstone-dicom-sr/src/tools/toolNames.ts
```diff
@@ -1,15 +1,15 @@
...
-  SREllipticalROI: SREllipticalROI.toolName,
-  SRArrowAnnotate: SRArrowAnnotate.toolName,
+  SRLength: 'SRLength',
+  SRBidirectional: 'SRBidirectional',
+  SREllipticalROI: 'SREllipticalROI',
+  SRArrowAnnotate: 'SRArrowAnnotate',
+  SRAngle: 'SRAngle',
+  SRCobbAngle: 'SRCobbAngle',
+  SRRectangleROI: 'SRRectangleROI',
+  SRPlanarFreehandROI: 'SRPlanarFreehandROI',
```
sedghi, line 12: Let's remove the names if you don't have them in tools/*

sedghi (approved):

PR #3141, opened 2023-02-03
fix: Create typed services and allow lower case access to agree with naming conventions

wayfarer3130:
Move to typescript typed services
Use the correct naming convention on registration of services, but leave the old names for a while
Show a few examples of use, including all examples of CineService => cineService and ToolBarService => toolbarService to show the correct naming convention.

### Future plan:
To rename all class instance to lower case.


wayfarer3130:
@sedghi - this is just switching to typescript to make it more consistent.  No functional changes.

sedghi (approved):
Let's wait until CI passes. Great initiative

PR #3143, opened 2023-02-05
fix(SR): fix Length and Bidirectional tools

lscoder:
### Context

Annotations are not rendered properly after loading a SR and enabling measurements tracking (OHIF-341).

### Changes & Results

- Changed how Length and Bidirectional annotations are rendered to make it similar to the original annotation added by the user before saving the SR.
- SRs does not support labels that were moved to different location
- No changes pushed to ellipse and arrow annotations

### Testing

- Add some Length and/or Bidirectional annotations
- Create a SR
- Load the SR created
- Once SR is loaded enable measurements tracking
- Compare how annotations are rendered before saving the SR, after loading the SR and after enabling measurements tracking.

### DIFF
--- a/.docker/Nginx-Orthanc/docker-compose.yml
+++ b/.docker/Nginx-Orthanc/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3.5'
 
 services:
   orthanc:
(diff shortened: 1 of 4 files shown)


sedghi (approved):
Great work

PR #3155, opened 2023-02-08
feat(findingSite): Allow customizing the measurement panel and show site/finding

wayfarer3130:
This change fixes saving and restoring of the site/finding information associated with a measurement, and IF the measurement has a site/finding, then adds it to the measurement display.

Must be merged after the context menu changes as it depends on that.

### DIFF
--- a/extensions/cornerstone-dicom-sr/src/utils/hydrateStructuredReport.js
+++ b/extensions/cornerstone-dicom-sr/src/utils/hydrateStructuredReport.js
@@ -11,6 +11,25 @@ const CORNERSTONE_3D_TOOLS_SOURCE_VERSION = '0.1';
 
 const supportedLegacyCornerstoneTags = ['cornerstoneTools@^4.0.0'];
 
+const convertCode = (codingValues, code) => {
+  if (!code || code.CodingSchemeDesignator === 'CORNERSTONEJS') return;
+  const ref = `${code.CodingSchemeDesignator}:${code.CodeValue}`;
+  const ret = { ...codingValues[ref], ref, ...code, text: code.CodeMeaning };
+  return ret;
+};
+
+const convertSites = (codingValues, sites) => {
+  if (!sites || !sites.length) return;
+  const ret = [];
+  // Do as a loop to convert away from Proxy instances
+  for (let i = 0; i < sites.length; i++) {
+    // Deal with irregular conversion from dcmjs
+    const site = convertCode(codingValues, sites[i][0] || sites[i]);
+    if (site) ret.push(site);
+  }
+  return (ret.length && ret) || undefined;
+};
+
 /**
  * Hydrates a structured report, for default viewports.
  *
@@ -20,8 +39,16 @@ export default function hydrateStructuredReport(
   displaySetInstanceUID
 ) {
   const dataSource = extensionManager.getActiveDataSource()[0];
-  const { measurementService, displaySetService } = servicesManager.services;
+  const {
+    measurementService,
+    displaySetService,
+    customizationService,
+  } = servicesManager.services;
 
+  const codingValues = customizationService.getCustomization(
+    'codingValues',
+    {}
+  );
   const displaySet = displaySetService.getDisplaySetByUID(
     displaySetInstanceUID
   );
@@ -171,6 +198,15 @@ export default function hydrateStructuredReport(
         CORNERSTONE_3D_TOOLS_SOURCE_VERSION
       );
       annotation.data.label = getLabelFromDCMJSImportedToolData(toolData);
+      annotation.data.finding = convertCode(
+        codingValues,
+        toolData.finding?.[0]
+      );
+      annotation.data.findingSites = convertSites(
+        codingValues,
+        toolData.findingSites
+      );
+      annotation.data.site = annotation.data.findingSites?.[0];
 
       const matchingMapping = mappings.find(
         m => m.annotationType === annotationType
--- a/extensions/cornerstone/src/commandsModule.ts
+++ b/extensions/cornerstone/src/commandsModule.ts
@@ -222,16 +222,6 @@ function commandsModule({ servicesManager, commandsManager }) {
             updatedMeasurement.findingSites = [code];
           }
         }
-        // TODO - remove this once measurement items customization is ready
-        const allCodes = [];
-        if (textLabel) allCodes.push(textLabel);
-        if (updatedMeasurement.finding) {
-          allCodes.push(updatedMeasurement.finding.CodeMeaning);
-        }
-        (updatedMeasurement.findingSites || []).forEach(it =>
-          allCodes.push(it.CodeMeaning)
-        );
-        updatedMeasurement.label = allCodes.join(', ');
       }
       measurementService.update(
         updatedMeasurement.uid,
--- a/platform/core/src/services/MeasurementService/MeasurementService.ts
+++ b/platform/core/src/services/MeasurementService/MeasurementService.ts
@@ -459,7 +459,12 @@ class MeasurementService extends PubSubService {
       log.warn(`Measurement ID not found. Generating UID: ${internalUID}`);
     }
 
+    const annotationData = data.annotation.data;
+
     const newMeasurement = {
+      finding: annotationData.finding,
+      findingSites: annotationData.findingSites,
+      site: annotationData.findingSites?.[0],
       ...measurement,
       modifiedTimestamp: Math.floor(Date.now() / 1000),
       uid: internalUID,
@@ -472,7 +477,7 @@ class MeasurementService extends PubSubService {
         measurement: newMeasurement,
       });
     } else {
-      log.info(`Measurement added.`, newMeasurement);
+      log.info('Measurement added', newMeasurement);
       this.measurements[internalUID] = newMeasurement;
       this._broadcastEvent(this.EVENTS.RAW_MEASUREMENT_ADDED, {
         source,
@@ -553,26 +558,27 @@ class MeasurementService extends PubSubService {
       );
     }
 
+    const oldMeasurement = this.measurements[internalUID];
+
     const newMeasurement = {
+      ...oldMeasurement,
       ...measurement,
       modifiedTimestamp: Math.floor(Date.now() / 1000),
       uid: internalUID,
     };
 
-    if (this.measurements[internalUID]) {
+    if (oldMeasurement) {
       // TODO: Ultimately, each annotation should have a selected flag right from the soure.
       // For now, it is just added in OHIF here and in setMeasurementSelected.
-      newMeasurement.selected = this.measurements[internalUID].selected;
       this.measurements[internalUID] = newMeasurement;
       if (isUpdate) {
-        this._broadcastEvent(this.EVENTS.MEASUREMENT_UPDATED, {
+       this._broadcastEvent(this.EVENTS.MEASUREMENT_UPDATED, {
           source,
           measurement: newMeasurement,
           notYetUpdatedAtSource: false,
         });
       } else {
         log.info('Measurement added.', newMeasurement);
-        this.measurements[internalUID] = newMeasurement;
         this._broadcastEvent(this.EVENTS.MEASUREMENT_ADDED, {
           source,
           measurement: newMeasurement,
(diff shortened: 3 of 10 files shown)


platform/core/src/extensions/ExtensionManager.js
```diff
@@ -281,6 +281,7 @@ export default class ExtensionManager {
 
       return extensionModule;
     } catch (ex) {
+      console.log('Exception:', ex);
```
sedghi, line 284: what is the need for log, when there is Error below?

platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -549,26 +554,27 @@ class MeasurementService extends PubSubService {
...
       uid: internalUID,
     };
 
-    if (this.measurements[internalUID]) {
+    if (oldMeasurement) {
       // TODO: Ultimately, each annotation should have a selected flag right from the soure.
       // For now, it is just added in OHIF here and in setMeasurementSelected.
-      newMeasurement.selected = this.measurements[internalUID].selected;
       this.measurements[internalUID] = newMeasurement;
       if (isUpdate) {
```
sedghi, line 570: Why this is removed? I'm pretty sure @joe added that recently for selection

platform/ui/src/types/index.ts
```diff
@@ -15,3 +14,9 @@ const StringNumber = PropTypes.oneOfType([PropTypes.string, PropTypes.number]);
 const StringArray = PropTypes.oneOfType([PropTypes.string, PropTypes.array]);
 
 export { StringNumber, StringArray, ThumbnailType };
+
+export type IProps = Record<string, unknown>;
+
+export type Predicate = (props: IProps) => boolean;
+
+export type Object = Record<string, unknown>;
```
sedghi, line 22: I don't think any of these are needed. They give feeling of being typesafe, though providing no value at all regarding type

platform/ui/src/components/MeasurementTable/MeasurementItem.tsx
```diff
@@ -24,6 +25,16 @@ const MeasurementItem = ({
   const onMouseEnter = () => setIsHovering(true);
   const onMouseLeave = () => setIsHovering(false);
 
+  const { site, finding } = item;
+  const useLabel = label || site?.text || finding?.text || '(empty)';
```
sedghi, line 29: labelToUse instead of useLabel?

platform/ui/src/components/MeasurementTable/MeasurementItem.tsx
```diff
@@ -24,6 +25,16 @@ const MeasurementItem = ({
...
+  let useDisplayText = displayText;
```
sedghi, line 30: same for displayText

platform/ui/src/components/MeasurementTable/MeasurementItem.tsx
```diff
@@ -24,6 +25,16 @@ const MeasurementItem = ({
...
+  if (finding && finding?.text !== useLabel) {
+    useDisplayText = [finding.text, ...useDisplayText];
+  }
+  if (site && site?.text !== useLabel) {
+    useDisplayText = [site.text, ...useDisplayText];
+  }
```
sedghi, line 36: I don't think these belong here, these logics should be a level up, and providing displayText and label as before, so that MeasurementItem becomes untouched

platform/ui/src/components/MeasurementTable/MeasurementTable.tsx
```diff
@@ -1,13 +1,31 @@
...
-const MeasurementTable = ({ data, title, onClick, onEdit }) => {
+const MeasurementTable = ({
+  data,
+  title,
+  onClick,
+  onEdit,
+  servicesManager,
+}) => {
+  servicesManager = servicesManager as ServicesManager;
+  const { customizationService } = servicesManager.services;
```
sedghi, line 16: So the idea was UI not get touched with the ohif core (services, managers, etc.), can we achieve the same thing by doing what you are doing with customizationService at the panel level?

platform/ui/src/components/ContextMenu/ContextMenuController.tsx
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * The context menu controller is a helper class that knows how
+ * to manage context menus based on the UI Customization Service.
+ * There are a few parts to this:
+ *    1. Basic controls to manage displaying and hiding context menus
+ *    2. Menu selection services, which use the UI customization service
+ *       to choose which menu to display
+ *    3. Menu item adapter services to convert menu items into displayable and actionable items.
+ */
+export default class ContextMenuController {
```
sedghi, line 90: ideally this should be the first thing in this file, and all the private `_methods` be below for readability. One should not need to scroll to the middle of the file to find the docstring

extensions/cornerstone-dicom-sr/src/utils/hydrateStructuredReport.js
```diff
@@ -166,17 +193,26 @@ export default function hydrateStructuredReport(
...
+        toolData.findingSites
+      );
+      annotation.data.site = annotation.data.findingSites?.[0];
 
       const matchingMapping = mappings.find(
         m => m.annotationType === annotationType
       );
 
-      MeasurementService.addRawMeasurement(
+      measurementService.addRawMeasurement(
```
sedghi, line 215: I don't like how each PR will have one or two change of letter casing. Those PRs should be separate and a complete rename of one service like what we did for Cine and Toolbar. Please take these out

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -11,29 +11,53 @@ import {
...
-    CornerstoneViewportService,
-    HangingProtocolService,
-    UINotificationService,
+    uiDialogService,
+    cornerstoneViewportService,
+    hangingProtocolService,
+    uiNotificationService,
+    customizationService,
+    displaySetService,
+    measurementService,
```
sedghi, line 37: same for these, let's have separate PRs for group of services, the approach to touch anything we see first makes the scope of this PR much larger for review (either for reviewer or for the reader later), and also doesn't achieve the end goal of changing the service name fully

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -68,16 +92,151 @@ const commandsModule = ({ servicesManager }) => {
...
+    contextMenuController.showContextMenu(
+      options,
+      viewerElement,
+      defaultPointsPosition
+    );
+  }
+
   const actions = {
+    /** Show the specified context menu */
+    showViewerContextMenu: providedOptions => {
```
sedghi, line 117: isn't this name redundant? do we have a non-viewer context menu?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -68,16 +92,151 @@ const commandsModule = ({ servicesManager }) => {
...
+          options,
+          customizationService.getModeCustomization(
+            menuName,
+            defaultContextMenu
+          )
+        );
+      }
+
+      if (useSelectedAnnotation && !nearbyToolData) {
+        const firstAnnotationSelected = getFirstAnnotationSelected(
```
sedghi, line 135: where is this getFirstAnnotationSelected imported?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -68,16 +92,151 @@ const commandsModule = ({ servicesManager }) => {
...
+      return (
+        nearbyToolData ??
+        cstUtils.getAnnotationNearPoint(element, canvasCoordinates)
+      );
+    },
+
+    // Measurement tool commands:
+    deleteMeasurement: ({ uid }) => {
+      if (uid) {
+        measurementServiceSource.remove(uid);
```
sedghi, line 177: I know you copy pasted from before, but this should be from measurementService itself. Then you can put the `measurementServiceSource` out of this file like before

extensions/cornerstone/src/commandsModule.ts
sedghi: In addition to above, it makes testing a lot harder! if we follow what you are doing right now, then we should test rigorously all components for each PR (segmentation, measurements, hangign protocol, etc. etc.). Please don't do this, and let's have separate PRs for each service

platform/core/src/extensions/ExtensionManager.js
wayfarer3130: The error below swallows this stack trace, so there is nothing obvious any longer.  It makes possible debugging extension load issues - and those shouldn't occur in production, so I don't see the problem with the exception.

platform/core/src/services/MeasurementService/MeasurementService.ts
wayfarer3130: It is implicit in the way that the new measurement is created, so was redundant.  It wasn't the only thing that was getting updated, which is why I changed how the update happens.

platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -549,26 +554,27 @@ class MeasurementService extends PubSubService {
       );
     }
 
+    const oldMeasurement = this.measurements[internalUID];
+
     const newMeasurement = {
+      ...oldMeasurement,
```
wayfarer3130, line 560: This is where the isSelected gets set - along with finding, findingSites, site, ...

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: The service isn't changing, it is just using typed access to the service.  When touching a file, I think we should use the new naming/typing because it makes it much easier to get the design right.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: Not yet, but I would guess that we will have them - for example, a context menu on the measurements in the measurements table.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: Added the import

extensions/cornerstone/src/initContextMenu.ts
```diff
@@ -0,0 +1,132 @@
...
+      'showContextMenu',
+      showContextMenuDefault
+    );
+
+  const showContextMenu = (contextMenuProps): void => {
+    customizationService.recordInteraction(
+      getShowContextMenu(),
+      contextMenuProps
+    );
+  };
```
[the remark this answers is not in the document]
wayfarer3130: Created a run option in the CommandsManager that is typed/managed - all I wanted was to have a way to run a set of commands like the toolbar does.

platform/ui/src/components/MeasurementTable/MeasurementTable.tsx
wayfarer3130: That is what I'm mostly trying to achieve, but the basic data needs to be available to the panel in a format it can detect, so that does require some changes to the surrounding code.  I can pull out all the changes related to actually displaying the data if it exists, although that only gets invoked if you are already using data that has the extra information.  The parsers have parsed it all along, just that it has been ignored or thrown out.

platform/ui/src/types/index.ts
wayfarer3130: Left predicate, but removed the others.  Predicate is a standard definition in many programming languages, so is useful to have.

platform/ui/src/components/MeasurementTable/MeasurementItem.tsx
wayfarer3130: Fixed

platform/ui/src/components/MeasurementTable/MeasurementTable.tsx
wayfarer3130: The UI should be unchanged unless you specifically add customizations - that is, it looks identical until you start adding customizations.

platform/ui/src/components/MeasurementTable/MeasurementItem.tsx
```diff
@@ -11,6 +11,7 @@ const MeasurementItem = ({
   isActive,
   onClick,
   onEdit,
+  item,
```
sedghi, line 14: I don't think this item is getting used, right?

sedghi (approved):
Looks good thanks

PR #3163, opened 2023-02-10
feat(ImageOverlayViewerTool) - add ImageOverlayViewer tool that can render image overlay (pixel overlay) of the DICOM images

md-prog:
### Context

DICOM images can have Overlay Plane and this overlay can be one of two types: "G" for graphics, and "R" for "ROI".
We want to be able to render "G" type overlays: image overlays or pixel overlays, over the DICOM image.

### Changes & Results

- Add new tool ImageOverlayViewerTool to `@ohif/extension-cornerstone` extension
- Add commands to "toggle" the image overlays

image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 10
- [x] Node version: 16.15.0
- [x] Browser: Version 109.0.5414.121 (Official Build) (64-bit)


extensions/cornerstone/src/commandsModule.ts
```diff
@@ -445,6 +446,27 @@ const commandsModule = ({ servicesManager }) => {
...
+      const viewportInfo = CornerstoneViewportService.getViewportInfoByIndex(
+        activeViewportIndex
+      );
+
+      const viewportId = viewportInfo.getViewportId();
+      const toolGroup = ToolGroupService.getToolGroupForViewport(viewportId);
+
+      toolGroup.setToolConfiguration(ImageOverlayViewerTool.toolName, {
+        showOverlays: toggledState,
+      });
```
sedghi, line 468: I don't think there is a need for `showOverlays` as option, since the disabled tool will do the job for you. So instead here you can just disabled it, and if toggle on enable it

sedghi: Doing it through disable/enable also makes the overlay off/on for all viewports in the toolgroup which aligns how other tools work

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * Image Overlay Viewer tool is not a traditional tool that requires user interactin.
+ * But it is used to display Pixel Overlays. And it will provide toggling capability.
+ *
+ * The documentation for Overlay Plane Module of DICOM can be found in [C.9.2 of
+ * Part-3 of DICOM standard](https://dicom.nema.org/medical/dicom/2018b/output/chtml/part03/sect_C.9.2.html)
+ */
+class ImageOverlayViewerTool extends BaseTool {
+  static toolName = 'ImageOverlayViewer';
```
sedghi, line 23: Follow how we are naming tools in cs3d, this causes some weird error in build

`  static toolName;` 

`SegmentationDisplayTool.toolName = 'SegmentationDisplay';
export default SegmentationDisplayTool;
`

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+  private _getCachedStat(targetId: string, createIfMissing = true): CachedStat {
+    if (!this._cachedStats[targetId] && createIfMissing)
+      this._cachedStats[targetId] = {
+        color: [127, 127, 127, 255], // color (r,g,b,a), default color: gray
+        overlays: [],
+      };
+    return this._cachedStats[targetId];
+  }
+
+  renderAnnotation = (enabledElement, svgDrawingHelper) => {
```
sedghi, line 50: hmmm, i know why you need this here, but I guses we need to fix it in cs3d later. Add a todo for this please

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+import { guid } from '@ohif/core/src/utils';
+
+interface CachedStat {
+  color: number[];
+  overlays: (unknown & {
+    _id: string;
+    pixelDataRaw: ArrayBuffer;
+    type: 'G' | 'R';
+  })[];
+}
```
sedghi, line 13: This is too confusing, please refactor into two or more types

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+    this._renderingViewport = viewport;
+
+    const imageId = this._getReferencedImageId(viewport);
+    if (!imageId) return;
+
+    const targetId = this.getTargetId(viewport);
+    const cachedStat = this._getCachedStat(targetId);
+
+    let overlays;
+    ({ overlays } = metaData.get('overlayPlaneModule', imageId));
```
sedghi, line 64: why not `let {overlays}` ?

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
sedghi: Also please add comment on what is G or R

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+    overlays = overlays.filter(overlay => overlay.type === 'G');
```
sedghi, line 65: so we don't render R (whatever it is)? please add commment why

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+    // no overlays
+    if (!overlays || overlays.length <= 0) return;
+
+    overlays.forEach(async (overlay, idx) => {
```
sedghi, line 70: Can you do please map and promiseAll instead of forEach?

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+      try {
+        if (!cachedStat.overlays[idx]) {
+          // first time load? let's load data
+          if (overlay.pixelData) {
```
sedghi, line 74: what happens if there is no pixelData? this doesn't cover it

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
md-prog: I thought of that, but in all other tools the renderAnnotation() was not a async function, so didn't use the `Promise.all`

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+    ) {
+      // pixelData of overlayPlane module is an array of bits corresponding
+      // to each of the underlying pixels of the image.
+      // let's create pixel data from bit array of overlay data
+      const { pixelDataRaw, rows: height, columns: width } = overlayData;
+      const pixelDataView = new DataView(pixelDataRaw);
+      const totalBits = width * height;
+      const color = this.configuration.fillColor;
+
+      const canvas = document.createElement('canvas');
```
sedghi, line 120: Can we remove this canvas at the end if we don't need it?

sedghi (changes requested):
Great work. 
See my minor comments. The major comment is to clean up all the cachedStats, canvases, svg nodes etc. that you create on destroy. Maybe this is something we need to add at cs3d level that when a toolGroup is destroyed each tool should run their own destroy as we see here it is storing a lot of state that will later cause memory leak IMO

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
sedghi: discard my comment, since I saw you are rendering svg

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+    const targetId = this.getTargetId(viewport);
+    const cachedStat = this._getCachedStat(targetId);
+
+    let overlays;
+    ({ overlays } = metaData.get('overlayPlaneModule', imageId));
+    overlays = overlays.filter(overlay => overlay.type === 'G');
+
+    // no overlays
+    if (!overlays || overlays.length <= 0) return;
```
wayfarer3130, line 68: This line needs to be before line 65

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+    const { viewport } = enabledElement;
+    this._renderingViewport = viewport;
+
+    const imageId = this._getReferencedImageId(viewport);
+    if (!imageId) return;
+
+    const targetId = this.getTargetId(viewport);
+    const cachedStat = this._getCachedStat(targetId);
+
+    let overlays;
```
wayfarer3130, line 63: Prefer to use const and assign a new variable rather than filtering the original array.

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+    });
+
+    return true;
+  };
+
+  private _renderOverlay(enabledElement, svgDrawingHelper, overlayData) {
+    if (
+      !overlayData.color ||
+      this.configuration.fillColor != overlayData.color
+    ) {
```
wayfarer3130, line 111: Please add the actual generation of the dataUrl into a separate function and store the dataUrl as the primary output of the overall asynchronous function which loads the overlay data.  That way, you cache just the dataUrl, and turning it on/off is just whether you include the SVG element or not.  Also, that generation should probably occur as part of general image loading, which separates out the timing from the creation of the tool object.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: @md-prog - can you apply this change to do it through the enable/disable toolgroup manager?

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
+import { metaData, StackViewport, Types } from '@cornerstonejs/core';
+import { utilities } from '@cornerstonejs/core';
+import { BaseTool } from '@cornerstonejs/tools';
+import { guid } from '@ohif/core/src/utils';
+
+interface CachedStat {
+  color: number[];
+  overlays: (unknown & {
+    _id: string;
+    pixelDataRaw: ArrayBuffer;
```
wayfarer3130, line 10: Consider making this just the raw overlay data, plus the color, eg:
color: Types.Color;
overlayData: URL;
size: vec2;
position: vec2;

That way the tool is added to the image area once it is actually loaded, and is for a specific image - so in fact you could consider adding it to the measurements panel if desired.

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,258 @@
...
+      configuration: {
+        showOverlays: true,
+        fillColor: [255, 127, 127, 255],
+      },
+    }
+  ) {
+    super(toolProps, defaultToolProps);
+  }
+
+  private _getCachedStat(targetId: string, createIfMissing = true): CachedStat {
```
wayfarer3130, line 41: Extract the code out into static methods which can be run against the metadata object, and result in tool data to add to the data structure.  Then overlays can be added externally, which does require some tweaks, but you won't be leaking memory quite the same way, nor creating them in an asynchronous fashion that doesn't quite work.

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -423,6 +423,14 @@ const toolbarButtons = [
             },
           ]
         ),
+        // TODO: we need a new icon for image overlay toggler button
+        _createToggleButton('ImageOverlay', 'link', 'Image Overlay', [
+          {
+            commandName: 'toggleImageOverlay',
```
wayfarer3130, line 429: Consider a generic toggle operation which can take a tool group to toggle between active and disabled here.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: Also, make it a generic enable/disable a named tool group - which means it is generally applicable for other tool types if desired.

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
wayfarer3130: Actually that is exactly how we are naming things - it is an equivalent line to the one you gave but is typescripty rather than javascripty.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -612,6 +613,27 @@ function commandsModule({ servicesManager, commandsManager }) {
...
       toolGroup.setToolEnabled(ReferenceLinesTool.toolName);
     },
+
+    /**
+     * Toggle Image Overlays
+     *
+     * @param param0
+     */
+    toggleImageOverlay: ({ toggledState }) => {
+      toggledState = !toggledState; // revert the state because unlike other toggle tools, this starts as enabled
```
wayfarer3130, line 623: You can just add a comment to line 633 and reverse the two conditions to set enabled/disabled.  Then, add the comment right over that to indicate that the "on" toggled stage means pixel overlay hidden.

sedghi:
@md-prog Can you please resolve the conflict for this PR, and also apply remaining comments?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -646,6 +647,26 @@ function commandsModule({
       }
       stateSyncService.store(storeState);
     },
+
+    /**
+     * Toggle Image Overlays
+     *
+     * @param param0
+     */
+    toggleImageOverlay: ({ toggledState }) => {
```
sedghi, line 656: I'm surprised we have a separate method for this here. It is a toggle similar to Crosshairs and every other toggle we have. We should treat it the same

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+    );
+
+    cachedStats[imageId] = {
+      color: color,
+      overlays: overlays.filter(overlay => overlay),
+    };
+  }
+
+  return cachedStats[imageId];
+};
```
sedghi, line 132: I prefer seeing the main piece of code when I open the file instead of seeing the helpers. The helpers should move to the bottom of the source code

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+  renderAnnotation = (enabledElement, svgDrawingHelper) => {
+    // overlays are toggled off by configuration
+    if (this._mode !== 'Enabled') return false;
+
+    const { viewport } = enabledElement;
+    this._renderingViewport = viewport;
+
+    const imageId = this._getReferencedImageId(viewport);
+    if (!imageId) return;
```
sedghi, line 178: please follow the new eslint rule -> no same line if statement

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+    const { overlays } = metaData.get('overlayPlaneModule', imageId);
```
sedghi, line 180: we should cache this, as we hit this A LOT! so `this.overlayCache.get(imageId)` if not found grab from metadata and add it to the cache

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+ *
+ * @param imageId
+ * @param overlayMetadata
+ * @param color {number[]} - color (r,g,b,a), default color: gray
+ * @returns
+ */
+const getCachedStat = async (
+  imageId: string,
+  overlayMetadata: any[],
+  color: number[] = [127, 127, 127, 255]
```
sedghi, line 91: there is no need for default color here, as the configuration has a default

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+): Promise<CachedStat> => {
+  if (
+    !cachedStats[imageId] ||
+    !isSameColor(cachedStats[imageId].color, color)
+  ) {
```
sedghi, line 96: I prefer less indentation. So can we early return on the opposite condition and make indentation one less?

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+        .filter(overlay => overlay.pixelData)
+        .map(async (overlay, idx) => {
+          let pixelData = null;
+          if (overlay.pixelData.Value) {
+            pixelData = overlay.pixelData.Value;
+          } else if (overlay.pixelData.retrieveBulkData) {
+            pixelData = await overlay.pixelData.retrieveBulkData();
+          }
+
+          if (!pixelData) return;
```
sedghi, line 108: ditto

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+    if (!imageId) return;
+
+    const { overlays } = metaData.get('overlayPlaneModule', imageId);
+    // no overlays
+    if (!overlays || overlays.length <= 0) return;
+
+    getCachedStat(imageId, overlays, this.configuration.fillColor).then(
+      cachedStat => {
+        cachedStat.overlays.forEach(overlay => {
+          this._renderOverlay(enabledElement, svgDrawingHelper, overlay);
```
sedghi, line 187: it is odd that getCachedStat is not a class method but renderOverlay is

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+    let referencedImageId;
+
+    if (viewport instanceof StackViewport) {
+      referencedImageId = targetId.split('imageId:')[1];
+    }
+
+    return referencedImageId;
+  }
+}
```
sedghi, line 284: I think this is already implemented in either BaseAnnotationTool or AnnotationTool

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+function _setNewAttributesIfValid(attributes, svgNode: SVGElement) {
+  Object.keys(attributes).forEach(key => {
+    const newValue = attributes[key];
+    if (newValue !== undefined && newValue !== '') {
+      svgNode.setAttribute(key, newValue);
+    }
+  });
+}
```
sedghi, line 305: These are already implemented in the cs3d drawingSVG, can we import it from there?

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,307 @@
...
+   *
+   * @param enabledElement
+   * @param svgDrawingHelper
+   * @param overlayData
+   * @returns
+   */
+  private _renderOverlay(enabledElement, svgDrawingHelper, overlayData) {
+    const { viewport } = enabledElement;
+    const imageId = this._getReferencedImageId(viewport);
+    if (!imageId) return;
```
sedghi, line 206: ditto

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -692,14 +696,19 @@ function commandsModule({
...
+      commandFn: actions.setSourceViewportForReferenceLinesTool,
     },
     storePresentation: {
       commandFn: actions.storePresentation,
       storeContexts: [],
       options: {},
     },
+    toggleImageOverlay: {
+      commandFn: actions.toggleImageOverlay,
+      storeContexts: [],
```
sedghi, line 709: storeContexts

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -692,14 +696,19 @@ function commandsModule({
...
-      commandFn: actions.toggleReferenceLines,
+    setSourceViewportForReferenceLinesTool: {
+      commandFn: actions.setSourceViewportForReferenceLinesTool,
     },
     storePresentation: {
       commandFn: actions.storePresentation,
       storeContexts: [],
       options: {},
     },
+    toggleImageOverlay: {
```
sedghi, line 707: remove I guess

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -15,7 +17,10 @@ const { windowLevelPresets } = defaults;
  * @param {*} icon
  * @param {*} label
  */
-function _createButton(type, id, icon, label, commands, tooltip, uiType) {
+function _createButton(type, id, icon, label, commands, tooltip, uiType, isActive) {
+  if (type === 'toggle' && isActive != null) {
```
sedghi, line 21: can we please have this as part of the return and ToolbarService handles it?

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,262 @@
...
+    if (!imageId) {
+      return;
+    }
+
+    const overlays =
+      this._cachedOverlayMetadata.get(imageId) ??
+      metaData.get('overlayPlaneModule', imageId)?.overlays;
+
+    // no overlays
+    if (!overlays || overlays.length <= 0) {
```
sedghi, line 51: most likely it should be ?.length !== 0

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -0,0 +1,262 @@
...
+  color: number[]; // [r, g, b, a]
+  overlays: {
+    // ...overlayPlaneModule
+    _id: string;
+    type: 'G' | 'R'; // G for Graphics, R for ROI
+    color?: number[]; // Rendered color [r, g, b, a]
+    dataUrl?: string; // Rendered image in Data URL expression
+  }[];
+}
+
```
sedghi, line 261: can we put this on top?

sedghi (commented):
one last iteration and we are done

sedghi (commented):
Thanks for taking care of this long awaiting PR Joe

sedghi (approved):

PR #3166, opened 2023-02-10
rename other services

sedghi:


platform/core/src/classes/HotkeysManager.ts
```diff
@@ -76,8 +76,10 @@ export class HotkeysManager {
       }
       definitions.forEach(definition => this.registerHotkeys(definition));
     } catch (error) {
-      const { UINotificationService } = this._servicesManager.services;
-      UINotificationService.show({
+      const {     uiNotificationService,
+ } = this._servicesManager.services;
```
wayfarer3130, line 80: Can you fix the formatting here?

wayfarer3130 (approved):
Will fix the one comment in the new branch.

PR #3173, opened 2023-02-14
Fix/Suggestion(MeasurementTracking):prevent empty sr description when export

dxlin:
### Context
Suggestion to prevent empty description for DICOM SR Report export. Prevent and alert when users attempts to save with empty description.

Questions:
1. Alert currently using browser, would bringing UINotificationService be preferred?
2. Alternatively could use states to disable the save button; however, ran into issues with react hooks when implementation attempted to turn actions passed to Dialog component to state -> is there any examples of a different approach to this?

### Changes & Results
-use _handleFormSubmit when enter is pressed (so same code is ran)
-add empty description check before allow processing within _handleFormSubmit
-if empty, stop event resolution and send alert

### Testing
1. Load Image in Basic Viewer
3. Select length tool
4. Create length
5. Start tracking
6. Change to Measurements panel (on right panel)
7. Click "Create Report" button
8. Press enter with empty description
-> Should ignore and prevent save
9. Click save button with empty description
-> Should ignore and prevent save
10. Type description, then delete (so description is blank again)
11. Press enter with empty description
-> Should ignore and prevent save
12. Click save button with empty description
-> Should ignore and prevent save
13. Type description, click save/press enter
-> Should allow save routine to proceed

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 10
- [x] "Node version: 16.15.1 
- [x] "Browser: Chrome 110.0.5481.96
  


sedghi:
Thanks
we shouldn't alert, it should just disable the save button

JennyLeePMH:
@sedghi , please take a look at Daniel's latest commit.

sedghi:
The button is not disabled

dxlin:
The upload function is disabled when description is empty as noted in the tests; the button itself is not in the disabled state.

As noted in point 2 of my description, I ran into issues with hooks when attempted to implement via state and would appreciate guidance/examples if this is the preferred method and has been implemented elsewhere.

sedghi:
The mouse hover should show a disable icon and also button should be disabled too look at the measurement panel when there is no measurements 

image

dxlin:
The measurement button isn't based off the dialog component and states can be passed to it when changed. The dialog however sets the state of the buttons at time of creation via the 'actions' array, which I am struggling to modify on the fly.

My attempts revolved around modifying this 'actions' array to pass a disabled state for save button if description is empty; however, that's when I ran into hook issues.

JennyLeePMH:
@sedghi , 
@dxlin had trouble disabling the save button on the dialog component.  The measurement panel is not a dialog component, so he could not re-use the example.  Specifically, on the dialog component, he ran into "issues with react hooks when implementation attempted to turn actions passed to Dialog component to state".  

Currently, when the report name is empty and user clicks "save", nothing is saved.  Let me know if this is reasonable for you.  It's ok for  us.

PR #3180, opened 2023-02-15
feat (cypress): add more e2e tests for MPR and add test mode and extension

sedghi:
- Add test mode and test extension for isolation of the tests
- Add hanging protocol selection through URL (useful for testing) `http://localhost:3000/viewer?StudyInstanceUIDs=1.3.12.2.1107.5.2.32.35162.30000015050317233592200000046&hangingprotocolId=mpr` 
- Add MPR e2e tests 
- Add measurement panel relabling test
- Add measurement panel jump to measurement test
- fix unit tests
- It makes the query params in the URL case insensitive 
- Upgrade cypress and babel


extensions/cornerstone/src/init.tsx
```diff
@@ -77,18 +77,18 @@ export default async function init({
...
-      message:
-        'Cross Origin Isolation is not enabled, volume rendering will not work (e.g., MPR)',
-      type: 'warning',
-    });
-  }
-
-  if (cornerstone.getShouldUseCPURendering()) {
-    _showCPURenderingModal(uiModalService, hangingProtocolService);
-  }
+  // if (!window.crossOriginIsolated) {
```
wayfarer3130, line 80: This seems like something which could be turned off in the config for the cornerstone extension, as part of the overall configuration file.  Then, the testing can turn it off, but otherwise we can leave it on because it causes issues.

extensions/cornerstone/src/utils/mpr/toggleMPRHangingProtocol.ts
```diff
@@ -58,26 +58,19 @@ export default function toggleMPRHangingProtocol({
...
-    // Unable to create MPR, so be sure to return to the cached/original protocol.
-    hangingProtocolService.setProtocol(
-      cachedState.protocol.id,
-      viewportMatchDetails,
-      restoreErrorCallback
-    );
+  // What is the current active protocol and stage number to restore later
+  const { protocol, stage } = hangingProtocolService.getActiveProtocol();
 
+  const restoreErrorCallback = error => {
```
wayfarer3130, line 64: Please dump the error message to console.log, as otherwise it can be hard to figure out why something went wrong, and it is too late to grab the error message.

extensions/test-extension/README.md
```diff
@@ -0,0 +1 @@
+# TMTV Extension
```
wayfarer3130, line 1: Could you drop a 1 paragraph desription here please?

extensions/test-extension/src/index.tsx
```diff
@@ -0,0 +1,22 @@
+import { id } from './id';
+import init from './init';
+
+/**
+ *
+ */
+const testExtension = {
```
wayfarer3130, line 7: Please type this, now that we have Types.Extensions.Extension as a valid type.

extensions/test-extension/src/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+   * Only required property. Should be a unique value across all extensions.
+   */
+  id,
+  preRegistration({
+    servicesManager,
+    commandsManager,
+    extensionManager,
+    configuration = {},
+  }) {
+    init({ servicesManager, commandsManager, extensionManager, configuration });
```
wayfarer3130, line 18: Please import as "preRegistration" and include the function directly, rather than adding an extra hop here.  That avoids the extra communication about parameters between index.tsx and init

extensions/test-extension/src/init.js
```diff
@@ -0,0 +1,15 @@
+/**
+ *
+ * @param {Object} servicesManager
+ * @param {Object} configuration
+ * @param {Object|Array} configuration.csToolsConfig
+ */
+export default function init({ servicesManager, extensionManager }) {
```
wayfarer3130, line 7: Call the file init.ts  and then type the parameters/return.

modes/basic-test-mode/package.json
```diff
@@ -38,7 +38,8 @@
     "@ohif/extension-cornerstone-dicom-sr": "^3.0.0",
     "@ohif/extension-dicom-pdf": "^3.0.1",
     "@ohif/extension-dicom-video": "^3.0.1",
-    "@ohif/extension-measurement-tracking": "^3.0.0"
+    "@ohif/extension-measurement-tracking": "^3.0.0",
+    "@ohif/extension-test": "0.0.1"
```
wayfarer3130, line 42: ^3.0.1

platform/viewer/.webpack/writePluginImportsFile.js
```diff
@@ -66,7 +67,23 @@ function getRuntimeLoadModesExtensions() {
...
     '  const newModes = await modeFactory(modes,extensions);\n' +
     '  newModes.forEach(newMode => modes.push(newMode));\n' +
-    '}\n}\n'
+    '}\n}\n\n\n' +
+    'async function loadRuntimeImports(config) {\n' +
+    '  if (config && config.modes && config.modes.length) {\n' +
+    '    for (const modeName of config.modes) {\n' +
+    '      const existingMode = modes.find(mode => mode.id === modeName);\n' +
+    '      if (!existingMode) {\n' +
+    '        if (modeName === `@ohif/mode-basic-test-mode`) {\n' +
```
wayfarer3130, line 76: This should be generic code generated from the plugingConfig.json file dynamic entries list

platform/viewer/cypress/integration/volume/MPR.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    cy.get('.cornerstone-canvas').should('have.length', 1);
+
+    // should not have any div under it
+    cy.get('[data-cy="thumbnail-viewport-labels"]')
+      .eq(2)
+      .find('div')
+      .should('have.length', 0);
+  });
+
+  it('should correctly render Crosshairs for MPR', () => {
```
wayfarer3130, line 95: Generally great set of tests - what about a test to ensure the crosshairs are turned after MPR is exited?

wayfarer3130 (commented):
A few changes as commented, the only big one is to make the pluginConfig.json more dynamic so that any module, both modes and extensions can be elected to be lazy load.

platform/viewer/.webpack/writePluginImportsFile.js
sedghi: Will address in a separate PR as discussed

platform/viewer/cypress/integration/volume/MPR.spec.js
sedghi: I will add test with Percy

wayfarer3130 (approved):
Great set of additional tests - thank you Alireza

PR #3188, opened 2023-02-22
fix: state setter for download aspect toggle

slak44:
### Context

Clicking the aspect ratio button in the ViewportDownloadForm fails due to an error. When shadowing a class member name in js, you can't use the shadowed name to initialize the shadowing variable.

image

### Changes & Results

I fixed the syntax to avoid the error, and used the setter with callback, which should always be used when the new state depends on the current state in react.

### Testing

Click the aspect ratio button, and it will now work.

#### PR

#### Code

  etc.)

  additions or removals.


sedghi:

PR #3191, opened 2023-02-22
Fix servicesManager use before initialization

Pierre-Baptiste:
### Context

in AppInit.js, when `appConfigOrFunc` is defined as a function, it is being called with `servicesManager`. 
`servicesManager` is only initialized after, which results in the app crashing.

Screenshot 2023-02-22 at 18 47 55

 

### Changes & Results

Move the definition of `appConfig` down, after the initialization of `servicesManager`.

### Testing

you could reproduce this by using a function as configs.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: macOS 13.1"
- [x] "Node version: 16.14.2"
- [x] "Chrome 110.0.5481.100, Firefox 110.0"

### DIFF
--- a/platform/viewer/src/appInit.js
+++ b/platform/viewer/src/appInit.js
@@ -24,12 +24,6 @@ import {
  * @param {object[]} defaultExtensions - array of extension objects
  */
 async function appInit(appConfigOrFunc, defaultExtensions, defaultModes) {
-  const appConfig = {
-    ...(typeof appConfigOrFunc === 'function'
-      ? appConfigOrFunc({ servicesManager })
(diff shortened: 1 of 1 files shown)


sedghi (approved):
This looks good. Thanks!
Just curious what use case you have that you need a function for app config? (if you can share)

Pierre-Baptiste:
Thank you for merging ! 🙏🏻 

(I still struggle with a rendering problem, I submitted an issue #3196)

We use a function for app config so we can use a dynamic url as root. We get that url from a query string we provide when opening the app.

PR #3198, opened 2023-02-24
feat(CinePlayer and ViewportActionBar)

jbocce:
### Context

See OHIF issue #3123 for complete details.

### Changes & Results

new look cine control implemented
new custom blue color in tailwind config for various hover backgrounds in the cine control
new icons added for cine

Tooltip component now can be placed top (center) on hover
Tooltip component border colour now consistent with specs

fixed NPE in ViewportActionBar

upgraded tailwind to 3.2.7 in platform/ui
fixed issues in various button components brought by tailwind 3.2.7 where classes now need important flag

fixed issue with InputRange component so that the tracked value property can change externally
InputRange component can now optionally show its label

added new measurement tracking state service to hydrate SR without prompting
segmentation can also now be hydrated without prompting

### Testing

Visit this [url](https://delicate-monstera-5e0eac.netlify.app/).
The turn on cine. The new cine control should be centered at the bottom of the viewport.
Try to play/pause cine.
The x button exits cine completely.
The playback frame rate can be increased via the > button and decreases with the < button.
Hovering over the FPS, popups up a slider/input range control to more finely/quickly change the FPS.

When hydrating SR or SEG using the LOAD button in the ViewportActionBar, the hydration should happen immediately with no prompt.

For segmentation, use the various input range components in the side bar for Outline, Fill and Opacity.

The buttons in the top toolbar should look-and-feel as in the UI spec.

#### PR

- [x] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [x] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [x] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: 18.10.0"
- [x] "Browser: Chrome  110.0.5481.177"
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #3186, opened 2023-02-17
feat(ViewportActionBar): OHIF issue #3123

jbocce:
### Context

See issue https://github.com/OHIF/Viewers/issues/3123 for complete details.

### Changes & Results
Coding changes summary:

- Renamed previous viewport action bar to be LegacyViewportActionBar
- Components LegacyViewportActionBar depends on also renamed: LegacyCinePlayer and LegacyPatientInfo
- New Viewport coded to specs in issue
- added React hook useResizeObserver
- added some tailwind classes
- clicks and hovers for the viewport action bar of a non-active viewport now function WITHOUT having to activate the viewport first

With a shorter viewport action bar there is more vertical space for the viewport and its images/content.

A new React hook: useResizeObserver will come in handy when components resize and some action needs to be taken.

The useResizeObserver was used to remove and add various components of the viewport action bar depending on the viewport's width: 
- the measurement tracking navigation arrows are only present for viewports at least 300 pixels wide
- the series description is removed when its area to show the actual description is zero width

Icons were changed to better suit the shorter viewport.

### Testing

Change to various grid layouts and browser window sizes. The viewport action bar should resize smoothly with no wrapping. Series description and study date should truncate with ellipses when not enough space is present for them.
The patient info icon should be clickable to display the patient into in a tooltip type bubble.
Ensure measurement tracking navigation with the arrows works properly. 
The measurement tracking navigation arrows are only present for viewports at least 300 pixels wide.
Ensure that the various status icons and actions for SEG and SR measurement tracking work as expected.
Ensure that clicks and hovers for the viewport action bar of a non-active viewport should function WITHOUT having to activate the viewport first.

#### PR

- [x] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [x] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [x] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: 18.10.0"
- [x] "Browser: Chrome 110.0.5481.100
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #3204, opened 2023-02-28
feat(ViewportActionBar and CinePlayer): Add new design for action bar and cine player

jbocce:
This pull request is for merging the feature branch feat/viewport-actionbar-enhancements to v3-stable.

Detailed PRs have already been performed on the feature branch. In particular...

- https://github.com/OHIF/Viewers/pull/3198
- https://github.com/OHIF/Viewers/pull/3186

This pull request is simply the merging of the changes in those PRs to v3-stable


sedghi:
great work indeed. 

Could you please look into these

1. The arrows don't go to hide when minimize happens sometimes when there is no focus?

[1c74d097-2258-4ed4-a3f4-81289cade88a.webm](https://user-images.githubusercontent.com/7490180/221885387-fe86ae2d-7caf-4fad-8e0d-bf22c7a7d23e.webm)

2. The tooltip for CINE is too low, almost touching the text
image

3. The tool tip hover don't stay when moving to the extremes 
[ae1d193c-cb36-46a2-a3ab-1b50b456b9a8.webm](https://user-images.githubusercontent.com/7490180/221886199-9aac736c-34d3-4333-8b5c-274066e9c155.webm)

4. I asked Erik and he also thinks < > should navigate in the measurements of the current viewport being selected. It is an easy fix since you just need to filter the measurement [here](https://github.com/OHIF/Viewers/blob/v3-stable/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.tsx#L278-L329), can we have that for this PR as well?

jbocce:
> great work indeed.
> 
> Could you please look into these
> 
> 1. The arrows don't go to hide when minimize happens sometimes when there is no focus?
> 
> 2. The tooltip for CINE is too low, almost touching the text
>    image
> 3. The tool tip hover don't stay when moving to the extremes
>    [ae1d193c-cb36-46a2-a3ab-1b50b456b9a8.webm](https://user-images.githubusercontent.com/7490180/221886199-9aac736c-34d3-4333-8b5c-274066e9c155.webm)
> 4. I asked Erik and he also thinks < > should navigate in the measurements of the current viewport being selected. It is an easy fix since you just need to filter the measurement [here](https://github.com/OHIF/Viewers/blob/v3-stable/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.tsx#L278-L329), can we have that for this PR as well?

Thanks for the feedback Alireza. I will address each of these in the next commit. In summary...

1. I was not correctly using the DOM ref in the React component - I incorrectly was using the ref as if setting the ref would cause a render which I fully new better, but stuff happens. I will store the DOM element in state as the ref changes and all will be good.
2. I have moved the tooltip up. However there is only so high it can go because if it goes too high there will be a gap between the hover area and the the tooltip and the tooltip will disappear as the mouse travels from the hover area to the tooltip.
3. From the video you posted it was evident that the mouse was travelling out the side in to the < > buttons' area which was NOT a hover area. Now the hover area includes the FPS text plus the < > buttons.
4. I think I have added the correct logic.

sedghi:
Found more 

image

is not the previous color

- Toolbar color is different 

https://user-images.githubusercontent.com/7490180/221958649-26cf7fb7-943d-4cb1-8a9b-8fb5ba666f07.mp4

jbocce:
> Found more
> 
> image
> 
> is not the previous color
> 
> * Toolbar color is different

You have a good eye @sedghi! These regressions were caused by an update to tailwind 3.2.7. There is no actual need for the update and since there were so many UI regressions because of the update, I think it is best to rollback to 3.2.4 - especially considering there are likely more regressions that are not easy to spot at this point in time.

sedghi (approved):
Looks great! Thanks

Issue #3135, opened 2023-01-27
Right side panel to auto start as closed but auto open when needed


sedghi:
### Request

The side panel on the right is automatically opened. But we have an option to pass the `rightPanelDefaultClosed` [here](https://github.com/OHIF/Viewers/blob/v3-stable/modes/longitudinal/src/index.js#L165).

![image](https://user-images.githubusercontent.com/7490180/215119319-4512c47d-5698-487b-8750-b480ac359ff8.png)

This feature request is to start with the closed state, and the Panels themselves hook into the desired events to open themselves when needed. 

- Measurement panel should listen to measurement modified/added and open itself
- Segmentation panel should listen to segmentation added event 
- My custom AI panel can listen to backend event ...

Probably we should pass in openCallback from the ViewerLayout component that update [the state](https://github.com/OHIF/Viewers/blob/v3-stable/extensions/default/src/ViewerLayout/index.tsx#L228) when the event happens.

**What feature or change would you like to see made?**

I would like to see an option added to the OHIF Viewers that allows users to start with all side panels in a closed state, and have the panels open themselves when an event is triggered, such as a measurement modified/added or segmentation added event. Additionally, custom AI panels could listen to backend events.

**Why should we prioritize this feature?**

This feature would help to keep the OHIF Viewers interface clutter-free, and give users more control over their user experience. By having side panels automatically open when particular events are triggered, users would not need to manually open them to access additional tools or perform certain tasks.

PR #3212, opened 2023-03-03
feat(SidePanel): Left or right side panel to auto start as closed but auto open when needed

jbocce:
### Context

OHIF issue #3135

### Changes & Results

- Added a method to set a callback for a Panel to invoke when it is ready to be shown (e.g. its data is loaded).
- Implemented such methods for both the segmentation and measurement panels.
- The SidePanel component now adds a callback to Panel components so that it will automatically open a Panel that was initially closed and yet to be opened.
- Updated the OHIF documentation accordingly.

#### Test A
1. Launch OHIF - the right side panel should be closed/collapsed.
2. Add a measurement.
3. The right side panel should open.

#### Test B
1. Launch OHIF with a study with an SR series that tracked measurement - the right side panel should be closed/collapsed initially.
2. Drag and drop or double click the SR series.
3. Load the measurements.
4. The right side panel should open.

#### Test C
1. Launch OHIF with a study with an SR series that tracked measurement - the right side panel should be closed/collapsed initially.
2. Open and close the right side panel.
3. Drag and drop or double click the SR series.
4. Load the measurements.
5. The right side panel should remain closed.

#### Test D
1. Launch OHIF with a study with a segmentation series - the right side panel should be closed/collapsed initially.
3. Drag and drop or double click the segmentation series.
4. The right side panel should open.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11" 
- [x] "Node version: 18.10.0 
- [x] "Browser: Chrome 110.0.5481.178"
  


sedghi (commented):
I see what you are doing in general. I think it is fine, but we need a new `PanelService` that can better manage all this logic + in future other use cases such as `PanelService.setPanelActive(panelId)` or `PanelService.movePanel(panelId, 'bottom')`.

I think `default` extension can be responsible for such service creation and registration (maybe it is a core service? maybe). Then your `contentReadyCallback` can be **one** of many ways a panel can set itself to active then other people can just add other callbacks that they want a panel to get activated (using event subs) without touching the source code for the panel such as `PanelService.addPanelActivationCallback(panelId, callback)` and `PanelService` will **add** it to the array of activations for each of the panels.

jbocce:
> I see what you are doing in general. I think it is fine, but we need a new `PanelService` that can better manage all this logic + in future other use cases such as `PanelService.setPanelActive(panelId)` or `PanelService.movePanel(panelId, 'bottom')` (no need to design future use cases, but I see the need for `PanelService`)
> 
> I think `default` extension can be responsible for such service creation and registration (maybe it is a core service? maybe). Then your `contentReadyCallback` (don't like the name) can be **one** of many ways a panel can set itself to active then other people can just add other callbacks that they want a panel to get activated (using event subs) without touching the source code for the panel. Example: `PanelService.addPanelActivationCallback(panelId, callback)` and `PanelService` will **add** it to the array of activations for each of the panels.

Thanks @sedghi. That all makes sense to me. Some comments/replies...
- the `PanelService` feels like a core service to me because `PanelModule`s are core
- similar to you not liking the name `contentReadyCallback` I do not like the name `panelActivationCallback`. Why? Because IMHO 'activation' refers to something being activated or being currently active. A panel that is hidden and not part of the UI can be neither of those. How about  `readyToShowCallback ` instead?

sedghi:
@jbocce activation is tied to show right now (if you click when closed it becomes activate + shown), but also if you have it open and you click on another tab to make it activated it switches to that. I don't see any scenario where a panel is activated but not shown. Do you?

platform/core/src/services/PanelService/PanelService.ts
```diff
@@ -0,0 +1,96 @@
...
+export const EVENTS = {
+  ACTIVATE_PANEL: 'event::panelService:activatePanel',
+};
+
+export type ActivatePanelEvent = {
+  panelId: string;
+  forceActive: boolean;
+};
+
+export default class PanelService extends PubSubService {
```
jbocce, line 13: Once we are ok with this service, its usage and its API, I will add the appropriate documentation for it.

extensions/cornerstone-dicom-seg/src/index.tsx
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 
 import getSopClassHandlerModule from './getSopClassHandlerModule';
 import PanelSegmentation from './panels/PanelSegmentation';
+import { Types } from '@ohif/core';
```
sedghi, line 6: above relative imports please

platform/docs/docs/platform/services/data/PanelService.md
```diff
@@ -0,0 +1,42 @@
...
+| ACTIVATE__PANEL       | Fires a `ActivatePanelEvent` when a particular panel should be activated.     |
+
+
+## API
+
+### Panel Activation
+
+- `activatePanel`: Fires the `ACTIVATE_PANEL` event for a particular panel (id).
+An optional `forceActive` flag can be passed that when `true` forces the
+panel to activate. Ultimately, it is up to the panel's container whether it
```
sedghi, line 29: The `forceActive` can benefit from more explanation. It is not clear what `forces to activate` mean

platform/docs/docs/platform/services/data/PanelService.md
```diff
@@ -0,0 +1,42 @@
...
+is appropriate to activate the panel. For instance, if the user opened and then
+closed the panel, the container may decide the user knows best and will not
+open the panel.
```
sedghi, line 32: Same for this section, it is not clear when reading what does `For instance, if the user opened and then closed the panel...`, what is container? what is decide? how it is decided? through the `forceActive`?

platform/docs/docs/platform/services/data/PanelService.md
```diff
@@ -0,0 +1,42 @@
...
+
+- `addActivatePanelTriggers`: Creates and returns event subscriptions that when
+fired will activate the specified panel with an optional `forceActive` flag
+(see `activatePanel`). This allows for panel activation to be directly triggered
+by some other event(s). When the triggers are no longer needed, simply
+unsubscribe to the returned subscriptions.
```
sedghi, line 38: it benefits from measurement_added example

platform/docs/docs/platform/services/data/index.md
```diff
@@ -20,6 +20,7 @@ We maintain the following non-ui Services:
 - [Toolbar Service](../data/ToolBarService.md)
 - [Measurement Service](../data/MeasurementService.md)
 - [Customization Service](customization-service.md)
+- [Panel Service](../data/PanelService.md)
```
sedghi, line 23: we need to add it to `sidebars.js` as well

platform/ui/src/components/SidePanel/SidePanel.tsx
```diff
@@ -67,15 +68,22 @@ const position = {
...
   className,
   activeTabIndex: activeTabIndexProp,
   tabs,
+  openWhenPanelActivated,
 }) => {
+  const panelService: PanelService = servicesManager.services.panelService;
+
   const { t } = useTranslation('SidePanel');
 
+  const [hasBeenOpened, setHasBeenOpened] = useState(
```
sedghi, line 82: please add a comment above that this tracks it has been opened from the begining of the app

jbocce: actually it does not track from the beginning of the app, only for the life of the SidePanel. So if one returns to the Search page, this will get reset.

platform/docs/docs/platform/extensions/modules/panel.md
```diff
@@ -64,7 +68,17 @@ providing the component with its.
 
 New: You can easily add multiple panels to the left/right side of the viewer
 using the mode configuration. As seen below, the `leftPanels` and `rightPanels`
-accept an `Array` of the `IDs`.
+accept an `Array` of the `IDs`. The default state of either/both the left and
+right side panels may also be specified. The default state includes whether the side
+panel should be closed initially and if closed initially, whether the side panel
+should open as soon as one of its child panels requests to activate via a
```
sedghi, line 74: child panels => child panels tabs? or just tabs?

extensions/measurement-tracking/src/getPanelModule.tsx
```diff
@@ -38,4 +45,21 @@ function getPanelModule({
...
+  // Do not force activation so as to respect the state the user may have left the UI in.
+  return panelService.addActivatePanelTriggers(
+    trackedMeasurementsPanelId,
+    measurementService,
+    [
+      measurementService.EVENTS.MEASUREMENT_ADDED,
+      measurementService.EVENTS.RAW_MEASUREMENT_ADDED,
+    ]
+  );
+}
```
sedghi, line 63: same for this to go inside modeEnter of mode

modes/longitudinal/src/index.js
```diff
@@ -161,8 +161,15 @@ function modeFactory() {
...
               leftPanels: [tracked.thumbnailList],
+              // leftPanelDefaultState: {
+              //   closed: true,
+              //   openWhenPanelActivated: true,
+              // },
               rightPanels: [dicomSeg.panel, tracked.measurements],
-              // rightPanelDefaultClosed: true, // optional prop to start with collapse panels
+              rightPanelDefaultState: {
+                closed: true,
+                openWhenPanelActivated: true,
```
sedghi, line 171: I guess we can just use the old one which was `rightPanelDefaultClosed` and we don't need the `openWhenPanelActivated` anymore, in the SidePanel you can just check if that panel has some registered triggers, if so it means `openWhenPanelActivated`, right?

platform/core/src/services/PanelService/PanelService.ts
```diff
@@ -0,0 +1,85 @@
...
+    eventName: string,
+    panelId: string,
+    callback: (event: PanelEvent) => void
+  ): Subscription {
+    if (!this.panelPubSub.has(panelId)) {
+      this.panelPubSub.set(panelId, new PubSubService(EVENTS));
+    }
+
+    return this.panelPubSub.get(panelId).subscribe(eventName, callback);
+  }
```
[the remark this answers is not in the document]
sedghi: nermind i'm wrong

platform/core/src/services/PanelService/PanelService.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Adds a mapping of events (sourceTriggerEvents) broadcasted by sourcePubService that
+   * when fired/broadcasted must in turn activate the panel with the given id.
+   * <p>
```
sedghi, line 60: what does <p> do here?

sedghi (commented):
Great work, getting there, see my comments please

platform/ui/src/components/SidePanel/SidePanel.tsx
sedghi: if so please add that comment there

modes/longitudinal/src/index.js
jbocce: There is no way to check if a panel has triggers registered because the PanelService does NOT store the triggers. So I think we should leave it like this.

platform/core/src/index.test.js
```diff
@@ -40,6 +40,7 @@ describe('Top level exports', () => {
       'DicomMetadataStore',
       'pubSubServiceInterface',
       'PubSubService',
+      'PanelService',
```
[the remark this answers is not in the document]
jbocce: Um... I really don't know.

platform/docs/docs/platform/extensions/modules/panel.md
jbocce: Well this whole md file does NOT mention tab(s) a single time, so I am not going to introduce that now. I think child panels is just fine.

platform/docs/docs/platform/services/data/PanelService.md
jbocce: Ok I will update the text. If it is still unclear perhaps you can help me come up with some wording that we can use.

platform/docs/docs/platform/services/data/PanelService.md
jbocce: Ok I will update the text. If it is still unclear perhaps you can help me come up with some wording that we can use.

platform/docs/docs/platform/services/data/index.md
jbocce: I am not sure what you mean. The deployed docs does have it in the sidebar already from what I can tell...

image

modes/longitudinal/src/index.js
jbocce: Furthermore, because the `PanelService` has a `activatePanel` method anyone can call that to activate the panel WITHOUT a trigger. So I am definitely going to leave it as is.

extensions/measurement-tracking/src/getPanelModule.tsx
```diff
@@ -11,9 +15,10 @@ function getPanelModule({
   commandsManager,
   extensionManager,
   servicesManager,
-}) {
+}): Types.Panel[] {
   return [
     {
+      id: '@ohif/extension-measurement-tracking.panelModule.seriesList',
```
sedghi, line 21: I'm not sure why we are adding id here. Extension manager dynaicly will create this id from the `name` here https://github.com/OHIF/Viewers/blob/v3-stable/platform/core/src/extensions/ExtensionManager.ts#L374

platform/core/src/types/PanelModule.ts
```diff
@@ -0,0 +1,30 @@
...
+  id: string;
+  name: string;
+  iconName: string;
+  iconLabel: string;
+  label: string;
+  component: React.FC;
+};
+
+type ActivatePanelTriggers = {
+  sourcePubService: PubSubService;
```
sedghi, line 13: shouldn't this be `sourcePubSubService`?

sedghi: also having `PanelService.addActivatePanelTriggers` it already has the trigger in the name, maybe we rename the second key to `sourceEvents`?

sedghi (commented):
Two tiny comments and then we are good to go

platform/core/src/types/PanelModule.ts
jbocce: This all makes sense.

extensions/measurement-tracking/src/getPanelModule.tsx
jbocce: Yes this bothered me a bit too. Well the `id` needs to be added to each individual panel module (i.e. each tab) so that `SidePanel` can register activation listeners for each of its tabs. I am about to push code where the `ExtensionManager` now sets the id for the various extension modules including the panel modules. I am not 100% certain this is the best way and if not I am sure you will be able to help me find something better. 😊

sedghi (approved):

sedghi:
Great work! thank you 🎉

Issue #3229, opened 2023-03-09
2D MPR crashes on some CTs with RangeError: Array buffer allocation failed.


muddasar-de:
We have the CTs with the following details, 
Dimensions: 512 x 512 
Number of slices: 2500 to 3500
OS: Windows 
Browsers: Google Chrome and Firefox 
System RAM: 16GB 

This problem is for both Local and Cloud environment(AWS) as well.

pieper:
Please provide details about the CT data that fails, like the slice dimensions, number of slices, and what clinical procedure this comes from.  Also describe your environment, like what OS, browser, how much memory your machine has, etc.

muddasar-de:
> Please provide details about the CT data that fails, like the slice dimensions, number of slices, and what clinical procedure this comes from.  Also describe your environment, like what OS, browser, how much memory your machine has, etc.

We have the CTs with the following details, 
Dimensions: 512 x 512 
Number of slices: 2500 to 3500
OS: Windows 
Browsers: Google Chrome and Firefox 
System RAM: 16GB 

This problem is for both Local and Cloud environment(AWS) as well.

pieper:
Thanks for the extra details.  I have also seen array buffer allocation fail for a scan with 2000 slices.  I'm not sure there's a planned work around.  @sedghi has may be able to comment.

muddasar-de:
> Thanks for the extra details.  I have also seen array buffer allocation fail for a scan with 2000 slices.  I'm not sure there's a planned work around.  @sedghi has may be able to comment.

Appreciate for the response and help.
Looking for the @sedghi response.

sedghi:
There are some plans for this. @Ouwen has written a POC for how to split the slices into different sections and let the user see the first sectoin, until it scrolls to the bottom and it automatically triggers going to next section, maybe he can talk about that. In addition to these kinds of tricks, we are working on the INT_16 texture size which reduces the memory in half, but seems like it has some bugs in Chromium and Safari, more on that here

- https://bugs.chromium.org/p/chromium/issues/detail?id=1408247

Ouwen:
Hi @muddasar-de for desktop use you can increase the memory limit from 1GB to 2GB and it should work with this PR into cornerstone3D-beta:

https://github.com/cornerstonejs/cornerstone3D-beta/pull/420

There is also a partial load PR I am working on which is based on invalidating the displaySetUID

Ouwen:
[...]

PR #3235, opened 2023-03-13
Handle large volumes

Ouwen:
Fixes https://github.com/OHIF/Viewers/pull/3235

### Context

Some volumes can be 2500-3500 large and it is challenging to load these in memory. This PR helps by limiting the volume size and adds a callback function on the ImageSet class which can be used to filter out or change the metadata of an ImageSet / displaySet

### Changes & Results
Large volumes are evenly sampled when they go above the maxFramesInVolume config. Changes have been made to the ImageSet class and config has been added to the CornerstoneCacheService.

### Testing
1. Open the viewer to http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095722.1

2. Open MPR mode to go into the volume mode (stack is unaffected by these changes)

3. Run the following on the console/window
`services.CornerstoneCacheService.setMaxFramesInVolume(20)`

4. You should see the volume update due to the DisplaySet being invalidated. You can also use the DisplaySetService to set the ImageMapper and invalidate. This functionality is left to be placed in extensions as use cases can vary.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [x] "Browser: Chrome Version 110.0.5481.177 (Official Build) (arm64)
  

### DIFF
--- a/extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
+++ b/extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
@@ -22,9 +22,24 @@ class CornerstoneCacheService {
 
   stackImageIds: Map<string, string[]> = new Map();
   volumeImageIds: Map<string, string[]> = new Map();
+  maxFramesInVolume = Infinity;
+  servicesManager: ServicesManager;
 
-  constructor(servicesManager) {
+  constructor(servicesManager: ServicesManager) {
     this.servicesManager = servicesManager;
+  }
+
+  public setMaxFramesInVolume(numFrames: number): void {
+    this.maxFramesInVolume = numFrames || Infinity;
+    const { displaySetService } = this.servicesManager.services;
+    const displaySets = displaySetService.getActiveDisplaySets();
+    displaySets.forEach(displaySet =>
+      displaySetService.setDisplaySetMetadataInvalidated(displaySet.getUID())
+    );
+  }
+
+  public getMaxFramesInVolume(): number {
+    return this.maxFramesInVolume;
   }
 
   public getCacheSize() {
@@ -186,9 +201,22 @@ class CornerstoneCacheService {
           displaySet,
           dataSource
         );
+        const distributedCopy = (items, n) => {
+          const elements = [items[0]];
+          const totalItems = items.length - 2;
+          const interval = totalItems / (n - 2);
+          for (let i = 1; i < n - 1; i++) {
+            elements.push(items[Math.floor(i * interval)]);
+          }
+          elements.push(items[items.length - 1]);
+          return elements;
+        };
 
         volume = await volumeLoader.createAndCacheVolume(volumeId, {
-          imageIds: volumeImageIds,
+          imageIds:
+            volumeImageIds.length > this.maxFramesInVolume
+              ? distributedCopy(volumeImageIds, this.maxFramesInVolume)
+              : volumeImageIds,
         });
 
         this.volumeImageIds.set(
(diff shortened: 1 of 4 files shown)


extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
```diff
@@ -186,9 +201,22 @@ class CornerstoneCacheService {
           displaySet,
           dataSource
         );
+        const distributedCopy = (items, n) => {
```
wayfarer3130, line 204: This is going to cause rendering issues when items.length ~= n because it is just going to throw out a certain set of instances, which causes the interval between images to be off by a factor of 2.  When n << items.length, it works ok,
because the interval becomes nearly consistent again, or if n = items.length / integer value, then one can sub-select evenly.  It is possible to write a inconsistent sample viewer, but currently we assume in the configuration that the sample volume is consistent, and don't account for this type of differences.

wayfarer3130: You could write a decimator instead - that takes out every other image, or as a general implementation, leaves 1 image, takes k images out, then repeats.  Decimators leave consistent spacing.

platform/core/src/classes/ImageSet.ts
```diff
@@ -10,13 +10,15 @@ const OBJECT = 'object';
...
+  private rawImages;
+
   constructor(images) {
     if (Array.isArray(images) !== true) {
       throw new Error('ImageSet expects an array of images');
     }
 
-    // @property "images"
-    Object.defineProperty(this, 'images', {
+    // @property "rawImages"
```
wayfarer3130, line 20: Just set instances to the complete list - instances is used elsewhere to mean all the instance values, whereas images is used for the images for display.

platform/core/src/classes/ImageSet.ts
```diff
@@ -32,6 +34,24 @@ class ImageSet {
     });
   }
 
+  get images(): any {
```
wayfarer3130, line 37: The general design of the display sets is that they are fixed representations of specific groups of DICOM objects - this mapping allows redefining that, and can cause problems when rendering different image sets.  I'd rather have a mapping function produce a new ImageSet entirely rather than having the images be variable.

wayfarer3130: It isn't used anywhere here, and I don't see any testing behaviour for it - I think this is a bad idea without careful consideration of what it means for other classes which expect to the representation of display sets to be consistent.

JeffersonCurveBeam:
Wondering if there was any plan to merge this PR soon - we are facing an issue addressed by this PR https://github.com/OHIF/Viewers/issues/3229#issuecomment-1535847521

In essence using the 2D MPR feature on our image will consume the entire machine's RAM and crash the browser temporarily on a 3-month old machine.

sedghi:
this is useful we should add it

extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
wayfarer3130: I added a decimator to CS3D in the HTJ2K progressive branch if you want one.

Ouwen:
Hey @sedghi, sorry for the radio silence from my side. I've been under pressure on finishing my dissertation ha

sedghi:
@Ouwen Oh wow, Good luck, nothing is expected, just was trying to include you in the loop

james-hanks:
@sedghi comment and close

sedghi:
So, after more thinking about this feature, I believe it will not adequately address the problem of large volumes as we desire. However, changing to a 16-bit texture might be the solution for most use cases, and in the future, we can consider the chunk and overlap solution.

https://docs.ohif.org/faq#how-do-i-handle-large-volumes-for-mpr-and-volume-rendering

Anyway, thanks @Ouwen for this

sedghi:
I will reopen for visiblity

PR #3237, opened 2023-03-13
fix(crosshairs): suppressed error for crosshair

Ouwen:
Suppress console error in the case the annotation is a Crosshair

### DIFF
--- a/extensions/cornerstone/src/initMeasurementService.js
+++ b/extensions/cornerstone/src/initMeasurementService.js
@@ -50,6 +50,20 @@ const initMeasurementService = (
 
   measurementService.addMapping(
     csTools3DVer1MeasurementSource,
+    'Crosshairs',
+    Length.matchingCriteria,
+    () => {
+      console.warn('Crosshairs mapping not implemented.');
+      return {};
+    },
+    () => {
+      console.warn('Crosshairs mapping not implemented.');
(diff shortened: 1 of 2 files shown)


platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -517,6 +517,7 @@ class MeasurementService extends PubSubService {
     }
 
     let measurement = {};
+    if (annotationType === 'Crosshairs') return;
```
wayfarer3130, line 520: Could you just add a mapping for crosshairs, even if it doesn't do anything?  There are other tools like probe which could also get empty mappings to suppress errors.

Ouwen: That makes sense, I'll have Crosshairs pretty much copy the mapping for Probe tool

sedghi (approved):

platform/core/src/services/MeasurementService/MeasurementService.ts
sedghi: This has caused an unexpected behaviour in OHIF recently when you open MPR and you open Crosshairs since it has mapping it prompts for tracking. I will remove this in a PR 

image

PR #3265, opened 2023-03-20
feat(URL): add param for initial series and sop uids to display

wayfarer3130:
### Context

In OHIF v2, it was possible to navigate to a specific series or SOP UID by providing URL parameters.  This is misisng in
v3 till now - this change adds it.

### Changes & Results
A URL like: 
http://localhost:3000/basic-test?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125113417.1&displaySet.SeriesInstanceUID=1.3.6.1.4.1.25403.345050719074.3824.20170125113545.4&displaySet.SOPInstanceUID=1.3.6.1.4.1.25403.345050719074.3824.20170125113558.2

can be used to navigate to the 101's instance in the third series.

### Testing

Use the above URL to navigate to the given position.  Try without the SOP instance or without the series to navigate to specified locations. (identical without series, image 1 without SOP)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/extensions/default/src/getHangingProtocolModule.js
+++ b/extensions/default/src/getHangingProtocolModule.js
@@ -45,6 +45,13 @@ const defaultProtocol = {
             greaterThan: { value: 0 },
           },
         },
+        {
+          attribute: 'isDisplaySetFromUrl',
+          weight: 10,
+          constraint: {
+            equals: true,
+          },
+        },
       ],
       // Can be used to select matching studies
       // studyMatchingRules: [],
@@ -65,7 +72,11 @@ const defaultProtocol = {
           viewportOptions: {
             viewportType: 'stack',
             toolGroupId: 'default',
-            // initialImageOptions: {
+            initialImageOptions: {
+              type: 'custom',
+              attribute: 'sopInstanceLocation',
+            },
+            // Other options for initialImageOptions:
             //   index: 180,
             //   preset: 'middle', // 'first', 'last', 'middle'
             // },
--- a/extensions/default/src/getHangingProtocolModule.js
+++ b/extensions/default/src/getHangingProtocolModule.js
@@ -7,7 +7,7 @@ const defaultProtocol = {
   hasUpdatedPriorsInformation: false,
   name: 'Default',
   createdDate: '2021-02-23T19:22:08.894Z',
-  modifiedDate: '2021-02-23T19:22:08.894Z',
+  modifiedDate: '2023-04-01',
   availableTo: {},
   editableBy: {},
   protocolMatchingRules: [],
@@ -33,8 +33,6 @@ const defaultProtocol = {
   },
   displaySetSelectors: {
     defaultDisplaySetId: {
-      // Unused currently
-      imageMatchingRules: [],
       // Matches displaysets, NOT series
       seriesMatchingRules: [
         // Try to match series with images by default, to prevent weird display
@@ -45,6 +43,8 @@ const defaultProtocol = {
             greaterThan: { value: 0 },
           },
         },
+        // This display set will select the specified items by preference
+        // It has no affect if nothing is specified in the URL.
         {
           attribute: 'isDisplaySetFromUrl',
           weight: 10,
@@ -72,11 +72,14 @@ const defaultProtocol = {
           viewportOptions: {
             viewportType: 'stack',
             toolGroupId: 'default',
+            // This will specify the initial image options index if it matches in the URL
+            // and will otherwise not specify anything.
             initialImageOptions: {
               type: 'custom',
               attribute: 'sopInstanceLocation',
             },
-            // Other options for initialImageOptions:
+            // Other options for initialImageOptions, which can be included in the default
+            // custom attribute, or can be provided directly.
             //   index: 180,
             //   preset: 'middle', // 'first', 'last', 'middle'
             // },
--- a/extensions/default/src/getHangingProtocolModule.js
+++ b/extensions/default/src/getHangingProtocolModule.js
@@ -75,8 +75,7 @@ const defaultProtocol = {
             // This will specify the initial image options index if it matches in the URL
             // and will otherwise not specify anything.
             initialImageOptions: {
-              type: 'custom',
-              attribute: 'sopInstanceLocation',
+              custom: 'sopInstanceLocation',
             },
             // Other options for initialImageOptions, which can be included in the default
             // custom attribute, or can be provided directly.
--- a/platform/core/src/services/HangingProtocolService/isDisplaySetFromUrl.ts
+++ b/platform/core/src/services/HangingProtocolService/isDisplaySetFromUrl.ts
@@ -1,0 +1,53 @@
+import { utils } from '@ohif/core';
+
+/** Indicates if the given display set is the one specified in the
+ * displaySet parameter in the URL
+ * The parameters are:
+ *    displaySet.SeriesInstanceUID
+ *    displaySet.SOPInstanceUID
+ */
+const isDisplaySetFromUrl = (displaySet): boolean => {
+  const params = new URLSearchParams(window.location.search);
+  const displaySetSeriesInstanceUID = utils.splitComma(
+    params.getAll('displaySet.SeriesInstanceUID')
+  );
+  let ret;
+  if (displaySetSeriesInstanceUID.length) {
+    const foundSeries = displaySetSeriesInstanceUID.indexOf(
+      displaySet.instance?.SeriesInstanceUID
+    );
+    if (foundSeries === -1) return false;
+    ret = true;
+  }
+  const displaySetSOPInstanceUID = utils.splitComma(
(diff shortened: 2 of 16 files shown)


platform/docs/docs/configuration/url.md
```diff
@@ -110,17 +110,31 @@ You can open more than one study in the Viewer by adding the `StudyInstanceUIDs`
 :::tip
 
 You can ues this feature to open a current and prior study in the Viewer.
-Read more in the [Hanging Protocol Module](../platform/extensions/modules/hpModule.md#matching-on-prior-study-with-uid) section
+Read more in the [Hanging Protocol Module](../platform/extensions/modules/hpModule.md#matching-on-prior-study-with-uid) section.  You can also use commas to separate
+values.
```
sedghi, line 114: can you provide couple of examples pleaes? 
- if you want two Studies ...
- if you want to do one study and two series

Do we have filtering for SOP too?
If so, that as well

platform/docs/docs/configuration/url.md
```diff
@@ -110,17 +110,31 @@ You can open more than one study in the Viewer by adding the `StudyInstanceUIDs`
...
 
 ```js
 /viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095722.1&SeriesInstanceUID=1.3.6.1.4.1.25403.345050719074.3824.20170125095748.1
 ```
 
 This will only open the viewer with one series (one displaySet).
+
+Alternatively, you can open the study on a specified series and/or sop instance
+using:
+`displaySet.SeriesInstanceUID` and/or `displaySet.SOPInstanceUID`
```
sedghi, line 131: as discussed on the call, `displaySet.` might be a little bit weird (need dicom prior knowledge kind of), how about `initialSeriesInstanceUID` and `initialSOPInstanceUID`

platform/core/src/utils/index.test.js
```diff
@@ -13,6 +13,7 @@ describe('Top level exports', () => {
       'sortStudyInstances',
       'sortStudySeries',
       'sortingCriteria',
+      'splitComma',
```
sedghi, line 16: is there any point in this file? do you agree we should get rid of this?

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -273,8 +273,10 @@ export default function ModeRoute({
       const filters =
         Array.from(query.keys()).reduce(
           (acc: Record<string, string>, val: string) => {
-            if (val !== 'StudyInstanceUIDs') {
-              if (['seriesInstanceUID', 'SeriesInstanceUID'].includes(val)) {
+            const lowerVal = val.toLowerCase();
+            if (lowerVal !== 'studyInstanceuids') {
+              // Note the query to filter the series UIDs is filterSeriesInstanceUID
+              if (lowerVal === 'filterseriesinstanceuid') {
```
sedghi, line 279: I thought you didn't change the `SeriesInstnaceUID` filter key, what is `filterseriesinstanceuid`

extensions/default/src/getHangingProtocolModule.js
```diff
@@ -65,7 +72,11 @@ const defaultProtocol = {
           viewportOptions: {
             viewportType: 'stack',
             toolGroupId: 'default',
-            // initialImageOptions: {
+            initialImageOptions: {
+              type: 'custom',
+              attribute: 'sopInstanceLocation',
+            },
```
sedghi, line 78: first of all we shouldn't change this since it is the default HP, I think just the documentation is enough. Both in PR and the MD docs

sedghi: can we mix this together? so we have 

``` js
initialImageOptions = {
  index: 180
}
```

or 


```js
initialImageOptions = {
  preset: 'middle'
}
```

or

```js
initialImageOptions = {
  custom: 'sopInstanceLocation'
}
```

platform/core/src/services/HangingProtocolService/isDisplaySetFromUrl.ts
```diff
@@ -0,0 +1,53 @@
...
+ * The parameters are:
+ *    displaySet.SeriesInstanceUID
+ *    displaySet.SOPInstanceUID
+ */
+const isDisplaySetFromUrl = (displaySet): boolean => {
+  const params = new URLSearchParams(window.location.search);
+  const displaySetSeriesInstanceUID = splitComma(
+    params.getAll('displaySet.SeriesInstanceUID')
+  );
+  let ret;
```
sedghi, line 14: please carry over the discussions from the previous PRs. Probably this is an old code though, `ret` is not clear

extensions/default/src/getHangingProtocolModule.js
```diff
@@ -45,6 +45,13 @@ const defaultProtocol = {
             greaterThan: { value: 0 },
           },
         },
+        {
+          attribute: 'isDisplaySetFromUrl',
+          weight: 10,
+          constraint: {
+            equals: true,
+          },
+        },
```
sedghi, line 54: I'm not sure why we are weighting things here, we should just always check for all `isDisplaySetFromURL`, and just use that for hanging. We shouldn't add this here. What is the purpose? There is only always one Series and one instance that can get hung first, so we can do it always

platform/core/src/services/HangingProtocolService/isDisplaySetFromUrl.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+  const displaySetSOPInstanceUID = splitComma(
+    params.getAll('displaySet.SOPInstanceUID')
+  );
+  if (displaySetSOPInstanceUID.length) {
+    if (!displaySet.images) return false;
+    const instance = displaySet.images.find(
+      instance =>
+        displaySetSOPInstanceUID.indexOf(instance.SOPInstanceUID) !== -1
+    );
```
sedghi, line 30: how about 

```js
const hasMatchingInstance = displaySet.images.some(instance => displaySetSOPInstanceUIDs.includes(instance.SOPInstanceUID));

return hasMatchingInstance
```

platform/core/src/services/HangingProtocolService/isDisplaySetFromUrl.ts
```diff
@@ -0,0 +1,53 @@
...
+    return instance !== undefined;
+  }
+  return ret;
+};
+
+/** Returns the index location of the requested image, or the defaultValue in this */
+function sopInstanceLocation(displaySets) {
+  const displaySet = displaySets[0];
+  if (!displaySet) return this.defaultValue;
```
sedghi, line 39: can't we just return null and the caller checks and use default itself? or just pass it as second argument

platform/core/src/services/HangingProtocolService/isDisplaySetFromUrl.ts
```diff
@@ -0,0 +1,53 @@
...
+  const displaySetSOPInstanceUID = splitComma(
+    params.getAll('displaySet.SOPInstanceUID')
+  );
+  if (displaySetSOPInstanceUID.length === 0 || !displaySet.images) {
+    return this.defaultValue;
+  }
+  const instanceIndex = displaySet.images.findIndex(
+    instance => displaySetSOPInstanceUID.indexOf(instance.SOPInstanceUID) !== -1
+  );
+  return instanceIndex === -1 ? this.defaultValue : { index: instanceIndex };
```
sedghi, line 50: how we can make sure the `defaultValue` follows the contract for the options in the viewport? it is odd than if found it is returning it as `{index: instanceIndex}`

sedghi (commented):
see my comments, great stuff

platform/core/src/utils/splitComma.ts
```diff
@@ -0,0 +1,13 @@
+/** Splits a list of stirngs by commas within the strings */
+const splitComma = (strings: string[]): string[] => {
```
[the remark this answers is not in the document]
wayfarer3130: No, that logic splits things up differently and doesn't work - probably there is a flatMap solution, but I don't know what it is.

extensions/default/src/getHangingProtocolModule.js
wayfarer3130: The initialSeries/SOP parameters are used by the default protocol to choose the initial display set to hang and this is how that gets applied.

wayfarer3130: It has no affect if the user does not supply the parameters.

wayfarer3130: The point of this change is that one can now add custom attributes to do this sort of thing, so it is no longer baked into the core framework, but can be done in various ways as required.  So, the main mode/hanging protocol code no longer needs to know about hanging protocol details around how instances are matched.

PR #3273, opened 2023-03-22
[WIP] feat: Save DICOM SR to the same series

wayfarer3130:
### Context

It becomes very confusing to have a lot of series all with the same series description, but differing date/times and content. 
This PR changes the save so items get saved to the same series.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


pieper:
@fedorov from a quick look I think @wayfarer3130 is not overwriting the SR instance, but adding the new instance to the same series as existing SRs.  I don't know if this is allowed in the standard (not sure why it wouldn't be) but if there's a chain of edits of the "same" SR then to me it sounds good to put them in the same series.

dclunie:
"adding the new instance to the same series as existing SRs. I don't know if this is allowed in the standard (not sure why it wouldn't be)" - criteria for same Series are defined in the standard [here](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/chapter_A.html#sect_A.1.2.3) - same equipment may be a barrier here, since arguably two deployed instances of the same viewer used on different computers at different times are not really "the same equipment".

fedorov:
@wayfarer3130 would you be interested to get together to discuss this feature together with the members of the Imaging Data Commons team (David Clunie, Steve Pieper and myself)? We would be interested in such a discussion.

sedghi:
@wayfarer3130 Could you please revisit this after the new replies ? Thanks

PR #3280, opened 2023-03-24
fix: Store hotkeys to specified name

wayfarer3130:
### Context

Hotkeys are stored incorrectly on a lot of people's machines.  This PR adds the ability
to use named hotkeys, as specified by hotkeyName in the mode object.
That also allows different modes to use different hotkey sets.

Finally, to allow updating of hotkeys without needing to manually clear them, the
default name has been changed to add a version number to it.

### Testing

Using the basic test mode, set a custom hotkey
Test your change
View the default longitudinal mode, test that the hotkey change is NOT effective
Go back into basic test mode, check that the hotkey is effective.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/platform/docs/docs/platform/modes/index.md
+++ b/platform/docs/docs/platform/modes/index.md
@@ -346,11 +346,16 @@ function modeFactory() {
     id: '',
     id: '',
     displayName: '',
-    hotkeyName: 'myHotkeyLocalStorageName',
     /*
     ...
     */
-    hotkeys: [..hotkeys.defaults.hotkeyBindings, ...myHotkeys],
+    hotkeys: {
+      // The name in preferences to use for this set of hotkeys
+      // Allows defining different sets for different modes
+      name: 'custom-hotkey-name',
+      // And the actual custom values here.
+      hotkeys:[..hotkeys.defaults.hotkeyBindings, ...myHotkeys]
+    },
   }
 }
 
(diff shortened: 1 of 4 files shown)


platform/docs/docs/platform/modes/index.md
```diff
@@ -344,6 +346,7 @@ function modeFactory() {
     id: '',
     id: '',
     displayName: '',
+    hotkeyName: 'myHotkeyLocalStorageName',
```
sedghi, line 349: It is weird that we have `hotkeyName` and then below we have `hotkeys` hmmmmmm

Does it make sense to have the `hotkeys` array an array of objects 

```js
hotkeys: [{name: myHotKeyLocal, hotkeys:[{....}]}]
```

so that we use the first name be default, BUT if the mode wants it can run with the other one? and later we can have the UI to switch between them?

sedghi (commented):
see my comment please

platform/docs/docs/platform/modes/index.md
wayfarer3130: That would be incompatible with what exists now, and we are starting to have modes.  Mostly, all that needs to be specified by the mode is the hotkeyName, and then it can get the default values.

wayfarer3130: I think it actually might make sense at some point to define the set of hotkeys elsewhere entirely, and just be able to read them by hotkey name.

sedghi: Yeah I think we need hotkeyModule, but at least bring the hotkeyName close to the definition of hotkeys

PR #3286, opened 2023-03-28
fix(viewports): The display of linked viewports during drag and drop has a race

wayfarer3130:
### Context

If multiple react state events occur near each other, react can decide to combine this, or can serialize them, or any combination thereof.  As well, any number of renderings or none can occur in between these.  This combination can lead to viewports
being in the midst of being created or destroyed when being rendered, or a partial new rendering occur, causing exceptions.

### Changes & Results

There are two parts to the fix.  First, not reusing viewport Id's at all means that create/destroy events on single viewport id's don't occur.  Second, making the drag and drop and segmentation updates atomic (single operation done at once) means it occurs consistently and in the same order and doesn't invoke multiple renders.  This also, incidentally, makes it faster.

### Testing

To test, the easiest way is to open up the test:e2e:serve environment, and view the SEG containing series in basic test mode
Try it with various layouts and protocol stages with:
Seg in one viewport, primary images in one or two other viewports
Ensure that you can load the seg, and that after you can:
1. Add the SEG to a blank viewport
2. Navigate each image series separately
3. Change layouts still
4. Note that during changing layouts there are some issues with remembering the positioning of multiple viewports containing the same series.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/extensions/cornerstone-dicom-seg/src/utils/_hydrateSEG.ts
+++ b/extensions/cornerstone-dicom-seg/src/utils/_hydrateSEG.ts
@@ -31,8 +31,6 @@ async function _hydrateSEGDisplaySet({
     displaySetInstanceUID
   );
 
-  viewportGridService.setDisplaySetsForViewports(updatedViewports);
-
   // Todo: fix this after we have a better way for stack viewport segmentations
 
   // check every viewport in the viewports to see if the displaySetInstanceUID
@@ -50,7 +48,7 @@ async function _hydrateSEGDisplaySet({
     );
 
     if (shouldDisplaySeg) {
-      viewportGridService.setDisplaySetsForViewport({
+      updatedViewports.push({
         viewportIndex: index,
         displaySetInstanceUIDs: viewport.displaySetInstanceUIDs,
         viewportOptions: {
@@ -61,6 +59,9 @@ async function _hydrateSEGDisplaySet({
       });
     }
   });
+
+  // Do the entire update at once
+  viewportGridService.setDisplaySetsForViewports(updatedViewports);
 
   return true;
 }
--- a/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
+++ b/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
@@ -156,7 +156,7 @@ class CornerstoneViewportService extends PubSubService
    * Disables the viewport inside the renderingEngine, if no viewport is left
    * it destroys the renderingEngine.
    *
-   * This is called when the element goes away entire - with new viewportId's
+   * This is called when the element goes away entirely - with new viewportId's
    * created for every new viewport, this will be called whenever the set of
    * viewports is changed, but NOT when the viewport position changes only.
    *
@@ -283,6 +283,9 @@ class CornerstoneViewportService extends PubSubService
     const viewport = renderingEngine.getViewport(viewportId);
     this._setDisplaySets(viewport, viewportData, viewportInfo, presentations);
 
+    // The broadcast event here ensures that listeners have a valid, up to date
+    // viewport to access.  Doing it too early can result in exceptions or
+    // invalid data.
     this._broadcastEvent(this.EVENTS.VIEWPORT_DATA_CHANGED, {
       viewportData,
       viewportIndex,
@@ -391,8 +394,6 @@ class CornerstoneViewportService extends PubSubService
       }
     }
 
-    // There is a bug in CS3D that the setStack does not
-    // navigate to the desired image.
     viewport.setStack(imageIds, initialImageIndexToUse).then(() => {
       viewport.setProperties(properties);
       const camera = presentations.positionPresentation?.camera;
--- a/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
+++ b/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
@@ -155,9 +155,14 @@ class CornerstoneViewportService extends PubSubService
   /**
    * Disables the viewport inside the renderingEngine, if no viewport is left
    * it destroys the renderingEngine.
+   *
+   * This is called when the element goes away entire - with new viewportId's
+   * created for every new viewport, this will be called whenever the set of
+   * viewports is changed, but NOT when the viewport position changes only.
+   *
    * @param viewportIndex
    */
-  public disableElement(viewportIndex: number) {
+  public disableElement(viewportIndex: number): void {
     const viewportInfo = this.viewportsInfo.get(viewportIndex);
     if (!viewportInfo) {
       return;
@@ -620,6 +625,10 @@ class CornerstoneViewportService extends PubSubService
     }
 
     const viewportInfo = this.getViewportInfo(viewport.id);
+
+    if (!viewportInfo) {
+      console.warn('Viewport info not defined for', viewport.id);
+    }
 
     const toolGroup = toolGroupService.getToolGroupForViewport(viewport.id);
     csToolsUtils.segmentation.triggerSegmentationRender(toolGroup.id);
--- a/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
+++ b/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
@@ -259,12 +259,6 @@ class CornerstoneViewportService extends PubSubService
     viewportInfo.setDisplaySetOptions(displaySetOptions);
     viewportInfo.setViewportData(viewportData);
 
-    this._broadcastEvent(this.EVENTS.VIEWPORT_DATA_CHANGED, {
-      viewportData,
-      viewportIndex,
-      viewportId,
-    });
-
     const element = viewportInfo.getElement();
     const type = viewportInfo.getViewportType();
     const background = viewportInfo.getBackground();
@@ -288,6 +282,12 @@ class CornerstoneViewportService extends PubSubService
 
     const viewport = renderingEngine.getViewport(viewportId);
     this._setDisplaySets(viewport, viewportData, viewportInfo, presentations);
+
+    this._broadcastEvent(this.EVENTS.VIEWPORT_DATA_CHANGED, {
+      viewportData,
+      viewportIndex,
+      viewportId,
+    });
   }
 
   public getCornerstoneViewport(
@@ -393,9 +393,7 @@ class CornerstoneViewportService extends PubSubService
 
     // There is a bug in CS3D that the setStack does not
     // navigate to the desired image.
-    viewport.setStack(imageIds, 0).then(() => {
-      // The scroll, however, works fine in CS3D
-      viewport.scroll(initialImageIndexToUse);
+    viewport.setStack(imageIds, initialImageIndexToUse).then(() => {
       viewport.setProperties(properties);
       const camera = presentations.positionPresentation?.camera;
       if (camera) viewport.setCamera(camera);
--- a/platform/core/src/extensions/ExtensionManager.ts
+++ b/platform/core/src/extensions/ExtensionManager.ts
@@ -283,6 +283,11 @@ export default class ExtensionManager {
             // Default for most extension points,
             // Just adds each entry ready for consumption by mode.
             extensionModule.forEach(element => {
+              if (!element.name) {
+                throw new Error(
+                  `Extension ID ${extensionId} module ${moduleType} element has no name`
+                );
+              }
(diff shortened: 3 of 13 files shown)


extensions/cornerstone-dicom-seg/src/utils/_hydrateSEG.ts
```diff
@@ -62,6 +60,9 @@ async function _hydrateSEGDisplaySet({
     }
   });
 
+  // Do the entire update at once
```
wayfarer3130, line 63: As it says, this change just does the entire update in a single operation, preventing inconsistencies and odd rendering effects.

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -34,6 +34,16 @@ const DEFAULT_STATE = {
 
 export const ViewportGridContext = createContext(DEFAULT_STATE);
 
+const isReuseableViewport = (oldViewport, newViewport) => {
```
wayfarer3130, line 37: This code isn't strictly necessary, but it does reuse viewports when they are the same size/display sets, and are immediately redisplayed, and otherwise gets a new viewport id every time.

platform/core/src/services/ViewportGridService/ViewportGridService.ts
```diff
@@ -75,18 +71,8 @@ class ViewportGridService extends PubSubService {
...
-    displaySetOptions,
-  }) {
-    this.serviceImplementation._setDisplaySetsForViewport({
-      viewportIndex,
-      displaySetInstanceUIDs,
-      viewportOptions,
-      displaySetOptions,
-    });
+  public setDisplaySetsForViewport(viewport) {
+    this.serviceImplementation._setDisplaySetsForViewports([viewport]);
```
sedghi, line 75: this doesn't makes sense, the function is called `setDisplaySetsForViewport ` which should (as it sounds like), set some displaySet props and viewport props and set it, right now it is setting just viewport?

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -90,53 +101,61 @@ export function ViewportGridProvider({ children, service }) {
...
+          // we can't assume the new displaySet can inherit the previous
+          // displaySet's or viewportOptions at all. For instance, dragging
+          // and dropping a SEG/RT displaySet without any viewportOptions
+          // or displaySetOptions should not inherit the previous displaySet's
+          // which might have been a PDF Viewport. The viewport itself
+          // will deal with inheritance if required. Here is just a simple
+          // provider.
+          const { viewportIndex, displaySetInstanceUIDs } = updatedViewport;
+          const previousViewport = viewports[viewportIndex] || {};
+          const viewportOptions = { ...updatedViewport.viewportOptions };
```
sedghi, line 132: we say in the comment there should not be any inheritance, yet this is inheritance, either the comment is wrong or this line

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -90,53 +101,61 @@ export function ViewportGridProvider({ children, service }) {
...
+
+          const displaySetOptions = updatedViewport.displaySetOptions || [];
+          if (displaySetOptions.length === 0) {
+            // Only copy index 0, as that is all that is currently supported by this
+            // method call.
+            displaySetOptions.push({
+              ...previousViewport.displaySetOptions?.[0],
+            });
```
sedghi, line 140: what about our updated `displaySetOptions`, why here we are using an old one, always 0 which doesn't makes sense either

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -90,53 +101,61 @@ export function ViewportGridProvider({ children, service }) {
...
+          }
+
+          let newView = {
```
sedghi, line 143: is this new viewport? then please newViewport instead of newView

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -46,6 +56,7 @@ const reuseViewport = (idSet, viewport, stateViewports) => {
   const oldIds = {};
   for (const oldViewport of stateViewports) {
     const { viewportId: oldId } = oldViewport;
+    if (!oldId) continue;
```
sedghi, line 59: is there a scenario where there is no viewportId? I don't think so, when?

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -90,53 +101,61 @@ export function ViewportGridProvider({ children, service }) {
...
+            viewportOptions,
+            displaySetOptions,
+            viewportLabel: viewportLabels[viewportIndex],
+          };
+          viewportOptions.presentationIds = getPresentationIds(
+            newView,
+            viewports
+          );
+
+          newView = reuseViewport(idSet, newView, state.viewports);
```
sedghi, line 155: I'm still not clear on what `reuseViewport` do, I guess it finds a reusable viewport right? so first of all `getReusableViewport`? since `reuseViewport` is action
Otherwise, the `reuseViewport` function can benefit a lot from more function documentation + in-line doc

platform/core/src/services/ViewportGridService/ViewportGridService.ts
wayfarer3130: What do you want me to call viewport when I set viewport?  I don't really care what it is called, but everywhere else we call it viewport, so I'd like to have a consistent new name for it if we are going to rename it.

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -56,32 +67,32 @@ const reuseViewport = (idSet, viewport, stateViewports) => {
...
+          ...oldViewport.viewportOptions,
+
+          viewportId: oldViewport.viewportId,
+        },
+      };
+    }
   }
-  throw new Error('No ID found');
+  // Find a viewport instance number different from earlier viewports
+  const viewportId = 'viewport-' + viewportCounter;
```
sedghi, line 86: does it makes sense instead of letting viewportId just goes to infinit (100000), we create it based on displaySetInstanceUIDs?

sedghi (commented):
please see my comments

platform/ui/src/contextProviders/ViewportGridProvider.tsx
wayfarer3130: No, it is updatedViewport.viewportOptions, NOT previousViewport.viewportOptions, so it is simply updating the viewport.

platform/ui/src/contextProviders/ViewportGridProvider.tsx
wayfarer3130: That is fairly long standing code - it is moved here, but is consistent with previous behaviour.  I think the issue is that if the provided update doesn't include display set options, then we need to provide something here.

platform/core/src/services/ViewportGridService/ViewportGridService.ts
sedghi: can't you just use the props? like the object that we already had?

platform/ui/src/contextProviders/ViewportGridProvider.tsx
wayfarer3130: Yes, when a viewport has no matching contents it doesn't necessarily get a viewportId

platform/ui/src/contextProviders/ViewportGridProvider.tsx
wayfarer3130: That causes all sorts of issue - you then need to figure out a unique addendum to the viewport to make it unique again because you can have two viewports shown at once.  Also, during race conditions where there are lots of updates to viewports you get some ugly errors.

platform/core/src/services/ViewportGridService/ViewportGridService.ts
```diff
@@ -38,9 +37,6 @@ class ViewportGridService extends PubSubService {
     if (setActiveViewportIndexImplementation) {
       this.serviceImplementation._setActiveViewportIndex = setActiveViewportIndexImplementation;
     }
-    if (setDisplaySetsForViewportImplementation) {
-      this.serviceImplementation._setDisplaySetsForViewport = setDisplaySetsForViewportImplementation;
```
ladeirarodolfo, line 42: For sake of backward compatibility I would say to not remove this. Instead we just factory to the new reducer you have created.

Also, if we decide to really change this we need to ensure the usage of it is also updated (cornerstone-dicom-seg, cornerstone-dicom-sr)

ladeirarodolfo (changes requested):

platform/ui/src/contextProviders/ViewportGridProvider.tsx
jbocce: So if it is possible that a viewport gets no id, then isn't the fact that we are using the `viewportId` as a key for a `ViewportPane` in `ViewportGrid.tsx` problematic?

jbocce (commented):
Looks good. Just some comments to address.

platform/core/src/services/ViewportGridService/ViewportGridService.ts
wayfarer3130: It is backwards compatible - I added a function that supports that directly, but I'm really against having two functions to call in the back end as it makes it harder to figure out what is happening.  We are still pre-release on v3-stable so don't have to absolutely worry about backwards compatibility.

platform/ui/src/contextProviders/ViewportGridProvider.tsx
wayfarer3130: Yes, it probably is an issue - maybe this doesn't happen any longer, but I'd rather leave the check in as it will create a viewportId for any missing viewport Ids.

jbocce: Sorry I am a bit confused. Is it still possible for a Viewport NOT to have a viewportId?

wayfarer3130: AFAIK it cannot happen, but I prefer to have a check there just in case.

jbocce: Just to be clear. I agree with leaving the check. I was more concerned with the use of `viewportId` as a key for a component.

platform/core/src/extensions/ExtensionManager.ts
```diff
@@ -283,6 +283,11 @@ export default class ExtensionManager {
             // Default for most extension points,
             // Just adds each entry ready for consumption by mode.
             extensionModule.forEach(element => {
+              if (!element.name) {
```
wayfarer3130, line 286: Not defining the name causes loading issues in various places, so it is worth having a check here on load.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -388,9 +393,7 @@ class CornerstoneViewportService extends PubSubService
 
     // There is a bug in CS3D that the setStack does not
     // navigate to the desired image.
-    viewport.setStack(imageIds, 0).then(() => {
-      // The scroll, however, works fine in CS3D
-      viewport.scroll(initialImageIndexToUse);
+    viewport.setStack(imageIds, initialImageIndexToUse).then(() => {
```
jbocce, line 396: With this change I think we have to remove the comment above.

jbocce (commented):
The commits since my last review look good.

I did find one issue that works on demo but not on this PRs deploy preview. Here are the steps...

1. Launch a series into MPR.
2. Double click one of the viewports to one up.
3. Swap in a different reconstructable series.
4. The viewer goes black with the following exception in the dev tools console...

```
react_devtools_backend.js:2655 TypeError: Cannot read properties of undefined (reading 'viewportIndex')
    at ViewportGridProvider.tsx:195:36
    at ViewportGridProvider.tsx:190:19
    at Object.co [as useReducer] (react-dom.production.min.js:159:447)
    at t.useReducer (react.production.min.js:23:173)
    at yn (ViewportGridProvider.tsx:325:41)
    at io (react-dom.production.min.js:157:137)
    at jo (react-dom.production.min.js:180:154)
    at Ys (react-dom.production.min.js:269:343)
    at Ol (react-dom.production.min.js:250:347)
```

PR #3287, opened 2023-03-28
fix(volumeLoad): should not have missing slices when loading

sedghi:
### Context
Loading volume viewports sometimes resulted in missing slices 

## Loading
The problem in cornerstone3D is solved first by updating texture always, and the problem here was that we shouldn't run imageLoadStrategy if defined AND custom .load, it caused problem. 

## increase prefetch queue 
Server can handle more, increased from 10 to 25

## mprAnd3DVolumeViewport
Added modality match so that the demo can be run on more studies that don't mount the CT first

### Testing

Before
-https://v3-demo.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095722.1&hangingProtocolId=mprAnd3DVolumeViewport

image

Now
- https://deploy-preview-3287--ohif-platform-viewer.netlify.app/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2403.871108593056125491804754960339&hangingProtocolId=mprAnd3DVolumeViewport

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 
- [x] Node version: 18.10.0 
- [x] Browser: Chrome 111.0.5563.111
  


extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -506,19 +506,16 @@ class CornerstoneViewportService extends PubSubService
...
-      if (
-        hangingProtocolService.runImageLoadStrategy({
-          viewportId: viewport.id,
-          volumeInputArray,
-        })
-      ) {
-        // Fallback to the default strategy if the custom one fails
-        return;
-      }
+      return hangingProtocolService.runImageLoadStrategy({
```
sedghi, line 509: we should not fallback to default, the imageLoadStrategy should handle when it is ready to continue, not here, if customImageLoad is defined we should not .load anymore

extensions/cornerstone/src/getHangingProtocolModule.ts
```diff
@@ -184,6 +184,15 @@ const mprAnd3DVolumeViewport = {
           },
           required: true,
         },
+        {
+          attribute: 'Modality',
+          constraint: {
+            equals: {
```
wayfarer3130, line 190: consider contains or containsI: ['ct', 'mr']

extensions/cornerstone/src/init.tsx
```diff
@@ -167,9 +167,26 @@ export default async function init({
           viewportId
         );
 
+        const ohifViewport = cornerstoneViewportService.getViewportInfo(
```
wayfarer3130, line 170: Thank you for getting this added - I wonder if the code should migrate somewhere it can be called generically?  Maybe just a helper method taking the viewport, stateSyncService and cornerstoneViewportService?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -506,19 +506,16 @@ class CornerstoneViewportService extends PubSubService
...
+        viewportId: viewport.id,
+        volumeInputArray,
+      });
     }
 
     volumeToLoad.forEach(volume => {
-      volume.load();
+      if (!volume.loadStatus.loaded && !volume.loadStatus.loading) {
```
wayfarer3130, line 516: What about cancelled?

platform/viewer/public/config/default.js
```diff
@@ -18,7 +18,7 @@ window.config = {
   maxNumRequests: {
     interaction: 100,
     thumbnail: 75,
-    prefetch: 10,
+    prefetch: 25,
```
wayfarer3130, line 21: The prefetch number is heavily dependent on protocol - running http/2 or /3 allows for high prefetch numbers, but a prefetch of 25 is too high for /1.1 or /1.0.  The issue is that it starts blocking the fetch queue.  I think I would add a comment here explaining a bit of that - I would also try as high as 100 and see if it makes a difference.  If you consider a latency of 200 ms, for a fetch size of 200k, then the transfer time probably on the 20 ms time frame, and you want to completely consume all the latency time with transfer time, so that means in a perfect overlap situation, you need 20/200 = 10 prefetch, which in a practical world with random transfer times, you want 2-3 times that amount, so 25 is a good practical amount.

extensions/cornerstone-dicom-sr/package.json
```diff
@@ -46,7 +46,7 @@
     "@babel/runtime": "^7.20.13",
     "classnames": "^2.3.2",
     "@cornerstonejs/adapters": "^0.4.1",
-    "@cornerstonejs/core": "^0.36.5",
-    "@cornerstonejs/tools": "^0.55.1"
+    "@cornerstonejs/core": "^0.37.0",
+    "@cornerstonejs/tools": "^0.57.1"
```
wayfarer3130, line 50: Why not latest - I think that is 0.38.0, 0.58.0 and adapters 0.6.0
Also, please ensure you update every package.json correctly - getting mismatches is ugly for including the libraries.

extensions/cornerstone/src/init.tsx
sedghi: yeah it should eventually get moved

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
sedghi: we should be able to resume the cancel, so I don't think so

extensions/cornerstone/src/getHangingProtocolModule.ts
sedghi: the preset we use it not for MR, it is CT bone, so until we add ui for changing that I guess it is fine

jbocce (approved):

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
wayfarer3130: If so, you need to remove the cancelled flag in the load then - that can wait for now, but you might want to test it.

wayfarer3130 (approved):
LGTM

PR #3290, opened 2023-03-30
feat(tmtv): add more stages to pt/ct

sedghi:
### PR Checklist

### DIFF
--- a/extensions/tmtv/src/getHangingProtocolModule.js
+++ b/extensions/tmtv/src/getHangingProtocolModule.js
@@ -1,3 +1,16 @@
+import {
+  ctAXIAL,
+  ctCORONAL,
+  ctSAGITTAL,
+  fusionAXIAL,
+  fusionCORONAL,
+  fusionSAGITTAL,
+  mipSAGITTAL,
+  ptAXIAL,
+  ptCORONAL,
+  ptSAGITTAL,
+} from './utils/hpViewports';
+
 const ptCT = {
   id: '@ohif/extension-tmtv.hangingProtocolModule.ptCT',
   locked: true,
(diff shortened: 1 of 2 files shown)


extensions/tmtv/src/utils/hpViewports.ts
```diff
@@ -0,0 +1,438 @@
+const ctAXIAL = {
```
wayfarer3130, line 1: This is a nice way of defining things so that the hanging protocol doesn't get quite so big.

extensions/tmtv/src/getHangingProtocolModule.js
```diff
@@ -173,425 +186,110 @@ const ptCT = {
...
+        ptCORONAL,
+        fusionAXIAL,
+        fusionSAGITTAL,
+        fusionCORONAL,
+        mipSAGITTAL,
+      ],
+      createdDate: '2021-02-23T18:32:42.850Z',
+    },
+    {
+      name: 'stage-2',
```
wayfarer3130, line 203: Could we give it a more descriptive name?  The name is intended to be displayable directly to users, so maybe 'Fusion 2x2' would work?

extensions/tmtv/src/getHangingProtocolModule.js
```diff
@@ -173,425 +186,110 @@ const ptCT = {
...
+        ptSAGITTAL,
+        ptCORONAL,
+        fusionAXIAL,
+        fusionSAGITTAL,
+        fusionCORONAL,
+        mipSAGITTAL,
+      ],
+      createdDate: '2021-02-23T18:32:42.850Z',
+    },
+    {
```
wayfarer3130, line 202: Can we add the convention that stages get comments on them, or maybe even that they get their own const and just get included?  I find it hard to notice when going between stages, so I'd prefer to have a bit more separation.  Not something I've done before, but something I think we should do.

wayfarer3130 (approved):
Just some minor rearranging and one renaming issue.

PR #3293, opened 2023-03-30
feat(4d): 1x3 PET layout with some panels on the left

lscoder:
Initial commit that loads PET/CT 4D study (same one used on PET/CT 4D example) in the "Data Preparation" stage (1x3 PET - no colormap applied so far) and adds some empty panels on the left.

There is a method that depends on cornerstone but I updated it to run locally without having to update the library.

### DIFF
--- a/extensions/default/src/getSopClassHandlerModule.js
+++ b/extensions/default/src/getSopClassHandlerModule.js
@@ -4,7 +4,39 @@ import ImageSet from '@ohif/core/src/classes/ImageSet';
 import isDisplaySetReconstructable from '@ohif/core/src/utils/isDisplaySetReconstructable';
 import { id } from './id';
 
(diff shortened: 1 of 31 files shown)


extensions/default/src/getSopClassHandlerModule.js
```diff
@@ -199,7 +238,9 @@ const sopClassUids = [
   sopClassDictionary.EnhancedUSVolumeStorage,
 ];
 
-function getSopClassHandlerModule() {
+function getSopClassHandlerModule(appContextParam) {
+  appContext = appContextParam;
```
sedghi, line 242: I prefer prop drilling in this case: passing the extension manager below into the getDisplaySetsFromSeries, and then into makeDisplaySets and then the function you like

PR #3300, opened 2023-04-02
WIP feat: expose hanging protocol stage to default

Ouwen:
### Context
This PR is meant to expose the stage index as an initial setting from the mode. The final implementation needs some discussion before it gets merged.
 - HangingProtocolService options are a bit inconsistent, rather than one options `{}` with named key inputs it is a mix of Object and positional args
 - Should stages have a matching engine similar to HP?
 - Modes currently force matching engine to run for a protocol even if one is explicitly provided.

### Changes & Results
This PR tries to expose stage selection or stage matching to the user.

@sedghi

### DIFF
--- a/platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
+++ b/platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
@@ -334,7 +334,11 @@ export default class HangingProtocolService extends PubSubService {
    *        the studies to display in viewports.
    * @param protocol is a specific protocol to apply.
    */
-  public run({ studies, displaySets, activeStudy }, protocolId) {
+  public run(
+    { studies, displaySets, activeStudy },
+    protocolId,
+    { stageIndex }
+  ) {
     this.studies = [...(studies || this.studies)];
     this.displaySets = displaySets;
     this.setActiveStudyUID((activeStudy || studies[0])?.StudyInstanceUID);
@@ -346,7 +350,7 @@ export default class HangingProtocolService extends PubSubService {
 
     if (protocolId && typeof protocolId === 'string') {
       const protocol = this.getProtocolById(protocolId);
-      this._setProtocol(protocol);
+      this._setProtocol(protocol, { stageIndex });
       return;
     }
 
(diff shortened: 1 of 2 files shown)


platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -334,7 +334,11 @@ export default class HangingProtocolService extends PubSubService {
    *        the studies to display in viewports.
    * @param protocol is a specific protocol to apply.
    */
-  public run({ studies, displaySets, activeStudy }, protocolId) {
+  public run(
+    { studies, displaySets, activeStudy },
+    protocolId,
+    { stageIndex }
```
wayfarer3130, line 340: Can you pass all the options here rather than extracting stageIndex?  There are reasons it might be desirable to pass initial display conditions to the run method.

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -75,7 +79,8 @@ function defaultRouteInit(
     // hanging protocol in the mode configuration
     hangingProtocolService.run(
       { studies, activeStudy, displaySets },
-      hangingProtocolId
+      hangingProtocolId,
+      { stageIndex: hangingProtocolStageIndex || 0 }
```
wayfarer3130, line 83: This should not pass in 0 if it isn't specified in the URL.  There is matching rules to decide the initial protocol stage, and specifically passing it in will disable some of those.

platform/viewer/src/routes/Mode/Mode.tsx
```diff
@@ -294,28 +305,26 @@ export default function ModeRoute({
...
+          hangingProtocolStageIndex: hangingProtocolStageIndexToUse,
+        });
+      }
+
+      return defaultRouteInit({
+        servicesManager,
+        studyInstanceUIDs,
+        dataSource,
+        filters,
+        hangingProtocol: hangingProtocolIdToUse,
```
wayfarer3130, line 325: This needs to start getting passed in as a set of options to the route init, as there may well be others that we want here.

sedghi:
@Ouwen Also for this one, any one free you have to work on this?

PR #3305, opened 2023-04-03
feat: Add a new hanging protocol @ohif/mn

wayfarer3130:
### Context

Added some hanging protocols which are available for use in the URL, but are not otherwise available by default.
`@ohif/mn`
and
`@ohif/seg`
to load in mn mode and segmentation default

### Changes & Results

Added the two new hanging protocols, only activated when specified

### Testing

Using default.js configuration, test with the two URLS:
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095722.1&hangingProtocolId=@ohif/mnGrid

and

http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.32722.99.99.270210390546167243599316128636641521105&hangingProtocolId=@ohif/seg

Also, this now works with the initialSeriesInstanceUID changes:
http://localhost:3000/viewer/ohif?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.4792.2001.105216574054253895819671475627&hangingProtocolId=@ohif/mnGrid&initialSeriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.4792.2001.299324598221034771323901632218,&initialSeriesInstanceuid=1.3.6.1.4.1.14519.5.2.1.4792.2001.210542687020950294776803479438,1.3.6.1.4.1.14519.5.2.1.4792.2001.246998110832292351156382625158
compare that with/without the initialseriesinstanceuid setting

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


extensions/default/src/hpMN.ts
```diff
@@ -0,0 +1,255 @@
+import { Types } from '@ohif/core';
+
+/**
+ * This hanging protocol can be activated on the primary mode by directly
+ * referencing it in a URL or by directly including it within a mode.
+ * It is not included in the viewer mode by default.
+ */
+const hpMN: Types.HangingProtocol.Protocol = {
```
jbocce, line 8: So I am guessing that the name MN refers to various M x N layouts - is that correct? Perhaps let's think of a more meaningful name like `hpVariousGridLayouts`? That said, if I am missing something and MN has meaning then by all means leave as is.

jbocce (approved):
Looks great. Thanks for the testing links. Approved.

extensions/default/src/hpMN.ts
```diff
@@ -0,0 +1,255 @@
...
+        {
+          viewportOptions: {
+            toolGroupId: 'default',
+            allowUnmatchedView: true,
+          },
+          displaySets: [
+            {
+              matchedDisplaySetsIndex: 1,
+              id: 'defaultDisplaySetId',
+              reuseId: '1-0',
```
sedghi, line 156: reuseId is from before

sedghi (changes requested):
see my comment

extensions/default/src/hpMN.ts
wayfarer3130: How about hpMxNGrid?

wayfarer3130: And @ohif/MNGrid - probably that means hpMNGrid would work

extensions/default/src/getHangingProtocolModule.js
```diff
@@ -92,211 +94,6 @@ const defaultProtocol = {
       ],
       createdDate: '2021-02-23T18:32:42.850Z',
     },
-
```
wayfarer3130, line 95: This really doesn't belong in the default HP, and the @ohif/mnGrid hp is now available on the URL, so can be directly applied.

platform/core/src/services/HangingProtocolService/custom-attribute/isDisplaySetFromUrl.ts
```diff
@@ -1,4 +1,4 @@
-import { getSplitParam } from '../../utils';
+import { getSplitParam } from '../../../utils';
```
jbocce, line 1: For whatever reason I cannot make comments on any other line other than this first line.

So on line 34, consider using `const displaySet = displaySets ?? displaySets[0]`;

jbocce (commented):
One minor comment to address.

jbocce:
The URL to test initialSeriesInstanceUID did not work for me.

platform/core/src/services/HangingProtocolService/custom-attribute/isDisplaySetFromUrl.ts
wayfarer3130: The syntax is
const displaySet = displaySets?.[0]
for a condition dereference - the ?? would mean take the display sets list if it is non null.
Applied that change.

PR #3306, opened 2023-04-03
fix(DicomTagBrowser): Fix parsing problems that was limiting nested tags rendering

igoroctaviano:
The recursive call was getting the wrong value during parsing. 
This fix gets the metadata once and uses the recursive call correctly to avoid parsing errors of nested tags.


sedghi:
I tested this and worked fine, however too slow for the SEG, see here https://deploy-preview-3306--ohif-platform-viewer.netlify.app/viewer?StudyInstanceUIDs=1.3.12.2.1107.5.2.32.35162.30000015050317233592200000046
Can we at least make it non-blocking with some text saying calculating or whatever?

sedghi:
@jbocce  can you please take care of this

jbocce:
> When I open SEG it shows the loading indicator, then I go to CT and shows CT and when I come back to SEG it shows the loading indicator again. I did not expect the second loading indicator. Why?

The loading indicator is shown because of how long it takes to render 1000+ rows regardless of if the same list of rows is displayed multiple times.

igoroctaviano:
> > When I open SEG it shows the loading indicator, then I go to CT and shows CT and when I come back to SEG it shows the loading indicator again. I did not expect the second loading indicator. Why?
> 
> The loading indicator is shown because of how long it takes to render 1000+ rows regardless of if the same list of rows is displayed multiple times.

@jbocce this approach worked for you? I tried here but couldn't see the loading indicator + the UI is frozen.

jbocce:
> > > When I open SEG it shows the loading indicator, then I go to CT and shows CT and when I come back to SEG it shows the loading indicator again. I did not expect the second loading indicator. Why?
> > 
> > 
> > The loading indicator is shown because of how long it takes to render 1000+ rows regardless of if the same list of rows is displayed multiple times.
> 
> @jbocce this approach worked for you? I tried here but couldn't see the loading indicator + the UI is frozen.

@igoroctaviano yes it did work for me, but it was an (obviously feeble) attempt to at least show that the UI is frozen via a progress indicator. To be honest, I was trying the minimal effort for this to add it to the v3 release. Regardless, even if the indicator was shown/seen, the user still could not do things like close the dialogue because the UI is still blocked, so it is a no-go. Apologies, it was just an attempt, knowing the actual solution would need more effort. Hopefully I have explained myself properly. 😊

FYI @sedghi .

platform/ui/src/components/InputRange/InputRange.tsx
```diff
@@ -72,14 +86,16 @@ const InputRange: React.FC<{
           inputClassName ? inputClassName : ''
         }`}
         style={{
-          background: `linear-gradient(to right, #5acce6 0%, #5acce6 ${rangeValuePercentage -
-            10}%, #3a3f99 ${rangeValuePercentage + 10}%, #3a3f99 100%)`,
+          background: solidTrackBackground
+            ? '#3a3f99'
```
sedghi, line 90: let's make this a configurable option in the props, same for the circle background, it can have defaults but should be able to get configured too

extensions/default/src/DicomTagBrowser/DicomTagBrowser.tsx
```diff
@@ -1,27 +1,35 @@
...
 import DicomTagTable from './DicomTagTable';
 import './DicomTagBrowser.css';
-import { InputRange, Select, Typography } from '@ohif/ui';
 
 const { ImageSet } = classes;
 const { DicomMetaDictionary } = dcmjs.data;
 const { nameMap } = DicomMetaDictionary;
 
 const DicomTagBrowser = ({ displaySets, displaySetInstanceUID }) => {
+  const excludedColumnIndicesForFilter: Set<number> = new Set([1]);
```
sedghi, line 17: can benefit from a comment above of what is it

extensions/default/src/DicomTagBrowser/DicomTagBrowser.tsx
```diff
@@ -54,64 +62,125 @@ const DicomTagBrowser = ({ displaySets, displaySetInstanceUID }) => {
...
+                solidTrackBackground={true}
+              />
+            </div>
+          )}
+        </div>
       </div>
-      <DicomTagTable
-        rows={getFormattedRowsFromTags(activeDisplaySet, instanceNumber)}
-      ></DicomTagTable>
+      <div className="w-full h-1 bg-black"></div>
```
sedghi, line 149: please add a Todo that this (search) should get refactored in the future

extensions/default/src/DicomTagBrowser/DicomTagBrowser.tsx
```diff
@@ -54,64 +62,125 @@ const DicomTagBrowser = ({ displaySets, displaySetInstanceUID }) => {
...
+
+  const debouncedSetFilterValue = useMemo(() => {
+    return debounce(setFilterValue, 200);
+  }, []);
+
+  useEffect(() => {
+    return () => {
+      debouncedSetFilterValue?.cancel();
+    };
+  }, []);
```
sedghi, line 102: I prefer the [] useEffects at the top of the list of useEffects, not sure about you

sedghi: nevermind there is no other

extensions/default/src/DicomTagBrowser/DicomTagBrowser.tsx
```diff
@@ -54,64 +62,125 @@ const DicomTagBrowser = ({ displaySets, displaySetInstanceUID }) => {
...
+      metadata = activeDisplaySet.images[instanceNumber - 1];
+    } else {
+      metadata = activeDisplaySet.instance || activeDisplaySet;
+    }
+    const tags = getSortedTags(metadata);
+    return getFormattedRowsFromTags(tags, metadata);
+  }, [instanceNumber, selectedDisplaySetInstanceUID]);
+
+  const filteredRows = useMemo(() => {
+    const filterValueLowerCase = filterValue.toLowerCase();
```
sedghi, line 77: if (filterValue === "") return early?

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
```diff
@@ -1,97 +1,203 @@
...
+    }
+
+    listRef.current.scrollTo(0);
+    listRef.current.resetAfterIndex(0);
+  }, [rows]);
+
+  /**
+   * When the browser window resizes, update the row virtualization (i.e. row heights)
+   */
+  React.useEffect(() => {
```
sedghi, line 86: remove React.

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
```diff
@@ -1,97 +1,203 @@
...
+    function handleResize() {
```
sedghi, line 87: handleResize could (maybe should because of useEffect dependency thingy in react) be outside

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
```diff
@@ -1,97 +1,203 @@
...
+      listRef.current.resetAfterIndex(0);
+    }
+
+    window.addEventListener('resize', handleResize);
+
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
+  });
```
sedghi, line 96: should it be on the []? this will get run everytime

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
```diff
@@ -1,97 +1,203 @@
...
+
+  const Row = ({ index, style }) => {
```
sedghi, line 98: put outside? (it will regenerate this function on each render)

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
```diff
@@ -1,97 +1,203 @@
...
+      >
+        <div className="px-3 w-4/24">{row[0]}</div>
+        <div className="px-3 w-2/24">{row[1]}</div>
+        <div className="px-3 w-6/24">{row[2]}</div>
+        <div className="px-3 w-5/24 grow">{row[3]}</div>
+      </div>
+    );
+  };
+
+  const [tagHeaderElem, setTagHeaderElem] = useState(null);
```
sedghi, line 118: I did not expect useStates this down, maybe move them to top?

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
```diff
@@ -1,97 +1,203 @@
...
+  const [vrHeaderElem, setVrHeaderElem] = useState(null);
+  const [keywordHeaderElem, setKeywordHeaderElem] = useState(null);
+  const [valueHeaderElem, setValueHeaderElem] = useState(null);
+
+  const tagRef = elem => {
+    if (elem) {
+      setTagHeaderElem(elem);
+    }
+  };
```
sedghi, line 127: shouldn't these be actual refs (useRef instead of states)?

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
```diff
@@ -1,97 +1,203 @@
...
+      setKeywordHeaderElem(elem);
+    }
+  };
+  const valueRef = elem => {
+    if (elem) {
+      setValueHeaderElem(elem);
+    }
+  };
+
+  const isHeaderRendered = () => tagHeaderElem != null;
```
sedghi, line 144: no need to recreate this on every render, just use tagHeaderElemen !== null in the return

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
```diff
@@ -1,97 +1,203 @@
...
+  /**
+   * Get the item/row size. We use the header column widths to calculate the various row heights.
+   * @param index the row index
+   * @returns the row height
+   */
+  const getItemSize = index => {
```
sedghi, line 150: inside a useCallback?

sedghi (commented):
great work indeed, see my comments please

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
jbocce: Yes, there should be a [] here. Good catch.

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
jbocce: I think once I add the `[]` dependency, this effect should be fine.

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
jbocce: I think I will use a `useCallback` here.

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
jbocce: The virtualized table does NOT render until the header is rendered because the header column widths are used to determine the row heights in the table. Therefore whenever the refs change (in particular the first time the refs are set), we want to trigger a render of the table. I do not think simple refs trigger a re-render, so I opted for putting them in states. Is there a better/alternative way?

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
jbocce: I will change to a `useCallback`

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
jbocce: yes

extensions/default/src/DicomTagBrowser/DicomTagTable.tsx
jbocce: And come to think of it, I should debounce the resize event too.

platform/ui/src/components/InputRange/InputRange.tsx
jbocce: Will do for the track, but the circle/thumb is handled in the css file and I am not sure it can be done inline.

jbocce:
> great work indeed, see my comments please

Thank you and thanks for the feedback.

sedghi (approved):
Looks awesome and smooth

PR #3307, opened 2023-04-03
fix(SRTool): Ellipse Display for DICOMSR Tool

IbrahimCSAE:
### Context

Fixes ellipse not rendering correctly when viewport is rotated. This was fixed in the native cornerstone3D tool and it's adapters, but also needed to be applied to the DICOMSRDisplayTool. 

See comments on the pull request

[Pull#479](https://github.com/cornerstonejs/cornerstone3D-beta/pull/479)

### Changes & Results

Added condition to check for the rotation and select the top/bottom/left/right canvas points based on it

Now Ellipse renders correctly in all scenarios, when drawn using the native tool, or when stored and restored, It will render correctly before/after re-hydration.

GIF 6

_______________________________________________________________________________________________________________________

### DIFF
--- a/extensions/cornerstone-dicom-sr/src/tools/DICOMSRDisplayTool.ts
+++ b/extensions/cornerstone-dicom-sr/src/tools/DICOMSRDisplayTool.ts
@@ -307,13 +307,26 @@ export default class DICOMSRDisplayTool extends AnnotationTool {
 
       const ellipsePointsWorld = data;
 
+      const rotation = viewport.getRotation();
+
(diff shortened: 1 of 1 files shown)


sedghi:
didn't we address this in Cornerstone3D? why do we need edit here as well?

IbrahimCSAE:
> didn't we address this in Cornerstone3D? why do we need edit here as well?

Because the DICOMSRTool handles the rendering instead of cornerstone3D tool if the load button has not been clicked yet in the OHIF viewer

sedghi (approved):

PR #3308, opened 2023-04-04
Refactored validators to allow for array input.

Sofien-Sellami:
### Context
Add an Array input for constraint value validation 

### Changes & Results
the inputs that are accepted by constraint type are: Array, string or {value :} 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: 
- [x] "Node version: 
- [x] "Browser:
  


salimkanoun:
maybe better if equals and doesNotEqual  only accept array of length 1 to not overlap with contains and doesNotContain 
same greaterThan make sens to only one value (integer or array of 1 integer)

wayfarer3130:
> maybe better if equals and doesNotEqual only accept array of length 1 to not overlap with contains and doesNotContain. Also, greaterThan make sense to only one value (integer or array of 1 integer)
> 
> probably needs @wayfarer3130 feedback

No, contains and doesNotContain are different operations from equals and doesNotEqual - one of them is a string containment, while the other is a equal or not equal.  Specifically:

equals  (array, value)
is true if any member of array is === to any member of value

notEqual (array, value)
is true if no member of array is equal to any member of value

salimkanoun:
Oh yes ok I see.
Great, we will wait for your review of this PR

platform/core/src/services/HangingProtocolService/lib/validator.js
```diff
@@ -1,69 +1,80 @@
 import validate from 'validate.js';
 
 validate.validators.equals = function(value, options, key, attributes) {
```
wayfarer3130, line 3: I think this needs some documentation here to specify exactly what it does - and also to think about the use cases a little bit.  There are two use cases I can think about:
1. That a vector is equal to a specified vector - for example  [1,0,0] === [1,0,0]
2. That a single value is equal to any one of a specified set, eg Modality === CT or MR
3. That any one of a set is equal to any one of another set, eg  ModalitiesInStudy being [CT,SR,SEG] has any value equal to either SR or SEG

I've usually used contains for case 2,3,  so I wonder if we should specify that equals implements #1 here?
IS there a case where we want exact equality of any of a set of values to any of a set of instances?

I know I had you change this so it supported #2, but now that I see it I wonder if we actually need to support #1 properly?  Could I ask you to update it so that it is full array equals?  That is, so that we can test [1,0,0]===[1,0,0] by position?

salimkanoun: The issue I see is that depending on the tag the value is either a single value or an array of value (as well as the constraint value)

So maybe the implementation could be :
- Check that value and contraint are comparable, if value is a single value, the constraint has to be a single value or an array of length 1.
- if the value is a single value, the single value constraint or position 0 of the array constraint should be strict equality (===)
- if the value is an array, constraint have to be an array of the same length and all member should match strict equality (===)
- Maybe for array to array comparison we shouldn't care about item order as we might have no control about the value order (modalities in study ['PT', 'CT'] or ['CT', 'PT']) so we will check that array has full size intersection no matter of item position.

Does it fit for you ? Is there another case we are missing.

And for doesNotEquals we will follow the same strict logic (array intersection of 0 if both array)

salimkanoun: Dear @wayfarer3130 , we thought on it there are some additional issues. If Equals is strict equality (unless order), and contains a different operation (string containement), then we will miss the option to make "on of the string" ("equal on of"/ array.includes).
So we might need some other terms ? Equals / includes / contains  ? 

Also the validate js is in public archive from end 2021 with no additional commit from 2y+.
The maintained validation library in js seems to be JOI https://github.com/hapijs/joi

Will you keep validate.js in long term ?

PR #3309, opened 2023-04-05
fix: Service consistency typing

wayfarer3130:
### Context

The general design of the services in OHIF is as a class instance, where the class has public methods which are callable,
and the class has a static registration object that is used to construct the instance.  

### Changes & Results

No functional changes, but in the cornerstone module switched to static REGISTRATION objects rather than dynamic ones.
Typed UIModalService and UINotificationService

### Testing

Load a DICOM SR, check the notification service shows the banner
Save a DICOM SR, check that the save notification appears.
Run the integration tests.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


extensions/cornerstone/src/types/CornerstoneServices.ts
```diff
@@ -0,0 +1,16 @@
+import { Types } from '@ohif/core';
+import ToolGroupService from './services/ToolGroupService';
+import SyncGroupService from './services/SyncGroupService';
+import SegmentationService from './services/SegmentationService';
+import CornerstoneCacheService from './services/CornerstoneCacheService';
+import CornerstoneViewportService from './services/ViewportService/CornerstoneViewportService';
+
+interface CornerstoneServices extends Types.Services {
```
sedghi, line 8: I don't think this is a good idea, we should be able to just extend the Types.Services and still be Types.Services everywhere, I don't like how each extension can just extend and claim it with their own name. Probably you can just do that inside the `registerServices` right?

sedghi (commented):
see my only comment

extensions/cornerstone/src/types/CornerstoneServices.ts
wayfarer3130: The problem is that there isn't any way to inject the type from the cornerstone extension into the @ohif/core component, so I don't really see how one could include the actual type there.  If you find a way, please let me know.

PR #3312, opened 2023-04-05
feat(App): Loads mode extensions on demand

wayfarer3130:
### Context

The declaration for the plugin imports file written by writePluginImports didn't allow for the config file to list the extensions/modes by name rather than by hard coded value, causing the initial bundle size to include all plugins. 

### Changes & Results

Split the generated code into bundles by extensions/modes so that they are loaded on demand by string name rather than being incorporated at build time.

### Testing

Load the APP_CONFIG=config/multiple.js to show loading of dev test mode and other extensions.  System should still work as before.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


jbocce (commented):
I ran the `multiple.js` configuration locally and was able to launch the various modes. I assume the non-default extensions loaded fine, but is there a way for me to verify this?

modes/basic-dev-mode/src/index.js
```diff
@@ -183,7 +186,7 @@ function modeFactory({ modeConfiguration }) {
       },
     ],
     extensions: extensionDependencies,
-    hangingProtocol: [ohif.hangingProtocol],
+    hangingProtocol: 'default',
```
jbocce, line 189: Perhaps remove line 13 then?

platform/viewer/src/appInit.js
```diff
@@ -87,25 +93,40 @@ async function appInit(appConfigOrFunc, defaultExtensions, defaultModes) {
...
+    ...(appConfig.modes || []),
+    ...defaultModes,
+  ]);
 
-    const mode = modeFactory(modeConfig);
+  // This is a terrible setup here - todo, change this to use an alternate
+  // name so as to not confuse the issue
+  appConfig.modes = [];
+  // This is a preferred name
+  appConfig.loadedModes = [];
```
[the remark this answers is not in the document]
wayfarer3130: Removed the update to the modes - now it just applies to loadedModes

jbocce (approved):

platform/viewer/pluginConfig.json
```diff
@@ -1,48 +1,55 @@
 {
   "extensions": [
+    {
+      "packageName": "@ohif/extension-default"
+    },
     {
       "packageName": "@ohif/extension-cornerstone",
-      "version": "3.0.0"
+      "default": false
```
sedghi, line 8: why we are getting rid of the version here?

wayfarer3130: It is meaningless - only external packages are meaningful to have the version number.

platform/public/demo.html
```diff
@@ -0,0 +1,6 @@
+<html>
```
jbocce, line 1: What is this all about?

wayfarer3130: It is just a demonstration that one can server up additional files coming from a specified directory.  That allows us to split off static files if we want, and have the directory get included automatically.

platform/viewer/pluginConfig.json
sedghi: maybe right now is meaningless but in future when we package properly and version them it is needed. So I don't think we remove and add back again, just let it be there

platform/public/demo.html
wayfarer3130: A place to put demo loads/alternate configs etc.

PR #3318, opened 2023-04-10
fix: Jump to measurement wasn't jumping when presentation already existed

wayfarer3130:
### Context

Changed the jump to measurement.  It now only jumps a single active viewport, all controlled by calls into commandsManager.
This has the stage setup to allow for jumping to 3d measurements as well.

### Changes & Results

Added a clearPresentation which is intended to clear the presentation information to allow immediate jump to a specified
image location without needing any back and forth/set timeouts to handling jumping.

Updated first/last image to have a new function jumpToImage instead of the old implementation

Updated ViewportGrid to call either clearPresentation/set display sets for viewports as a pair when no viewport currently holds the given view, OR to call jumpToImage to jump to a specified image.  This separates the functionality for viewport grid, leaving it in viewport grid services, and the functionality for actually jumping to a slice, which is clearly cornerstone functionality (for cornerstone viewports - other viewports may need other behaviour).

### Testing

Create a number of measurements on different images within a couple of series.
Navigate to OTHER images in the series, then drag a different series onto the viewport.
Navigate to the measurement by clicking on it.  The specified measurement should appear.

Note that the basic navigation still works with acquisition plane measurements, but this change has NOT added the ability to jump to non-acquisition plane viewports.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


extensions/cornerstone/src/commandsModule.ts
```diff
@@ -10,7 +10,7 @@ import {
   utilities as cstUtils,
   ReferenceLinesTool,
 } from '@cornerstonejs/tools';
-import { ServicesManager } from '@ohif/core';
+import { Types as OhifTypes } from '@ohif/core';
```
sedghi, line 13: OHIFTypes instea of `OhifTypes`

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -632,148 +667,110 @@ function commandsModule({ servicesManager, commandsManager }) {
...
-      options: {},
     },
     toggleReferenceLines: {
       commandFn: actions.toggleReferenceLines,
-      storeContexts: [],
-      options: {},
+    },
+    getToolGroup: {
+      commandFn: actions.getToolGroup,
     },
```
sedghi, line 774: we have a `toolGroupService`, any logic you need from here better be inserted in that service

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -491,33 +493,55 @@ function commandsModule({ servicesManager, commandsManager }) {
...
+        console.log('*** No position info found', viewport.viewportOptions);
       }
-
-      // Set slice to first slice
-      const options = { imageIndex: 0 };
-      cstUtils.jumpToSlice(viewport.element, options);
+      return {
+        ...viewport.viewportOptions,
+        initialImageOptions: { index: imageIndex },
+      };
```
sedghi, line 527: why `clearPresentation` is returning something? then I guess the name `clearPresentaiton` is not descriptive of its logic

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -491,33 +493,55 @@ function commandsModule({ servicesManager, commandsManager }) {
...
+        viewport.viewportOptions?.presentationIds?.positionPresentationId;
+      if (positionId && positionPresentationStore[positionId]) {
+        const updatePosition = {
+          ...positionPresentationStore[positionId],
+          initialImageIndex: imageIndex,
+        };
+        positionPresentationStore = {
+          ...positionPresentationStore,
+          [positionId]: updatePosition,
+        };
```
sedghi, line 519: I guess this can be merged 

```js
positionPresentationStore[positionId] = {
  ...positionPresentationStore[positionId],
  initialImageIndex: imageIndex,
};
```

platform/viewer/src/components/ViewportGrid.tsx
```diff
@@ -190,46 +216,55 @@ function ViewerViewportGrid(props) {
...
+        if (matchingViewports.length) {
+          const viewport = findMeasurementViewport(
+            matchingViewports,
+            viewportIndex,
+            measurement
+          );
+          // TODO - check the orientation of the viewport, and
+          // update it to match the provided image, changing the
+          // camera and type as required.
+          commandsManager.runCommand('jumpToImage', { viewport, imageIndex });
```
sedghi, line 235: having `platform/viewer` logic calling a command that will run inside extensions is very werid, and not ideal at all. We should avoid these calls, as you can see there is not other instance of `runCommand` inside platform/viewer folder

sedghi (commented):
see my comments

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: This is specific to the interaction between cornerstone and toolGroupService, so no, you can't put it into the toolGroupService.  It is just exposing a method that sometimes is needed elsewhere.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: Convention is that acronyms longer than 3 characters are treated as words with an upper/lower letter to start, and the remainder lower case.  So, OhifTypes is correct.

wayfarer3130:
@jbocce - as requested, updated this with a merge pull.

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -515,9 +526,26 @@ function _subscribeToJumpToMeasurementEvents(
...
+    props => {
+      cacheJumpToMeasurement = props;
+      const {
+        viewportIndex: jumpIndex,
+        measurement,
+        isConsumed,
+        priority,
+      } = props;
       if (!measurement) return;
+      if (isConsumed) {
```
sedghi, line 539: use 

```js
if (!measurement || isConsumed) return
```

platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -83,6 +83,20 @@ const VALUE_TYPES = {
   ROI_THRESHOLD_MANUAL: 'value_type::roiThresholdManual',
 };
 
+// Contains priority levels for the jump to image.  This allows firing the
+// jump to image several times with increasing priority to apply the change
+// to different sets of viewports until some viewport handles it.
+const JUMP_TO_MEASUREMENT = {
+  // The grid layout should be updated to include this
+  GRID_PRIORITY: 2,
```
sedghi, line 91: I would recommend this

```js
const JUMP_PRIORITY = {
  ALL_VIEWPROTS: 'allviewports', 
  ANY_VIEWPORT: 'anyviewport',
  SELECTED_VIEWPORT: 'selectedpriority'
}

Comment (wayfarer3130) [reply platform/core/src/services/MeasurementService/MeasurementService.ts:91]: The issue is that it needs to be numeric because it needs to increase, so one can't just name them - they are inherently ordered.

File: extensions/cornerstone/src/commandsModule.ts (hunk 2 above)
Comment (wayfarer3130) [reply extensions/cornerstone/src/commandsModule.ts:738]: Will separate this out into it's own PR.

File: platform/core/src/services/_shared/pubSubServiceInterface.js [hunk 8]
```diff
@@ -103,4 +103,20 @@ export class PubSubService {
...
     this.unsubscriptions = [];
   }
+
+  /**
+   * Creates an event that records whether or not someone
+   * has consumed it.  Call eventData.consume() to consume the event.
+   * Check eventData.isConsumed to see if it is consumed or not.
+   * @param props - to include in the event
+   */
+  protected createConsumableEvent(props) {
```
sedghi, line 113: protected is only in TS files, I'm seeing an error here

sedghi (approved):

PR #3338, opened 2023-04-25
[Bugfix] Calibration tool applied to annotations multiple times, altering their positions and measurements

md-prog:
### Context

 Calibration tool (implemented in #3136) was applied to annotations multiple times, altering their positions and measurements, when we change viewing images
 

### How to reproduce the bug

  1. Open a study with multiple images
  2. Select one image, draw an annotation.
  3. Calibrate the image using the Calibration Tool
  4. Notice the measurement changes on calibration on the measurement panel
  5. Select another image, (change the image rendered on the same viewport), and then go back to the original image
  6. Notice the measurement changes even though we did nothing to change it.

### Analysis

   The calibration tool performs its function by overriding `calibratedPixelSpacing` metadata. On each changes, the viewport re-renders. And in the StackViewport implementation, it will compare the `calibratedPixelSpacing` metadata with `imagePlaneModule` (`rowPixelSpacing` and `columnPixelSpacing`) metadata - original spacing. If these two are different, it will shoot out an "SPACING_CALIBRATED" event, which will then update the annotations and their measurements.

   Therefore if we update only the `calibratedPixelSpacing` metadata, and change the images, the "CALIBRATED" event will be triggered multiple times, which changes measurements multiple times. 
  The way that it updates the annotations and measurements also contributes to this issue, as they use the relative scale value to re-calculate, rather than the spacing values.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/extensions/cornerstone/src/tools/CalibrationLineTool.ts
+++ b/extensions/cornerstone/src/tools/CalibrationLineTool.ts
@@ -1,12 +1,16 @@
 import { metaData } from '@cornerstonejs/core';
 import { LengthTool, utilities } from '@cornerstonejs/tools';
+import { classes } from '@ohif/core';
+
 import callInputDialog from '../utils/callInputDialog';
 import getActiveViewportEnabledElement from '../utils/getActiveViewportEnabledElement';
 
 const { calibrateImageSpacing } = utilities;
+const metadataProvider = classes.MetadataProvider;
 
 /**
- * Calibration Line tool works almost the same as the
+ * Calibration Line tool works almost the same as the Line tool, but used for
+ * image length calibration.
  */
 class CalibrationLineTool extends LengthTool {
   static toolName = 'CalibrationLine';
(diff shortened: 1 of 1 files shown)


extensions/cornerstone/src/tools/CalibrationLineTool.ts
```diff
@@ -81,20 +85,48 @@ export function onCompletedCalibrationLine(servicesManager, csToolsEvent) {
...
+    calibrateImageSpacing(imageId, renderingEngine, rowSpacing, colSpacing);
+
+    // NOTE: in the StackViewport, if the row and colum pixel spacing from
+    // 'calibratedPixelSpacing' and the 'imagePlaneModule' metadata are different,
+    // it triggers the "IMAGE_SPACING_CALIBRATED" event, which applies calibration
+    // multiple times to the annotations.
+    // So, let's override imagePlaneModule.
+    // This override should happen in the next event cycle, as the rendering is
+    // going on after above function call.
+    setTimeout(() => {
```
wayfarer3130, line 109: @md-prog  - please update this change to be the one from FUBerlin - they are contributing that code back to the OSS, but we need to use the final version that is compatible with the released CS3D

md-prog:
A more safe fix has been applied to the `@cornerstonejs/core` package : https://github.com/cornerstonejs/cornerstone3D-beta/commit/06419303b5bf8901645f4c74bc25cb8eabf279c8

PR #3364, opened 2023-05-04
fix(cli): better support for external modules and toolGroups in templates

sedghi:
### Context

Previously the ohf-cli were not modifying the webpack and as a side effect external install dependencies in extensions and modes were not getting recognized. Also our templates did not have any toolbar and toolGorups

### Changes & Results

- Modify the webpack for the linked modes/extensions
- Add default toolGroups and toolbar to mode templates
- Fix a thumbnail style bug
- default data source name in configs was way down 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


modes/basic-dev-mode/src/index.js
```diff
@@ -135,8 +135,11 @@ function modeFactory({ modeConfiguration }) {
       ]);
     },
     onModeExit: ({ servicesManager }) => {
```
wayfarer3130, line 137: I'd suggest just making the toolGroupService be auto-destroy on mode exit - that prevents needing to know about it in the onModeExit, and the entire method can then be dropped.

platform/cli/templates/mode/src/initToolGroups.js
```diff
@@ -0,0 +1,260 @@
...
+  commandsManager,
+  toolGroupId
+) {
+  const utilityModule = extensionManager.getModuleEntry(
+    '@ohif/extension-cornerstone.utilityModule.tools'
+  );
+
+  const { toolNames, Enums } = utilityModule.exports;
+
+  const tools = {
```
wayfarer3130, line 13: Consider exporting some of these sets of values from the appropriate mode, and inheritting them, and then just re-use them.  The toolNames doesn't need to come from the utility module, so it can just be directly included, then this can all be statically defined, and the new mode creator can choose to use the defaults or modify it as they please.

wayfarer3130 (changes requested):
Given that you are changing this, I'd really like to see the default functions/setup inheritted from either basic-dev-mode or longitudinal.

platform/cli/templates/mode/src/toolbarButtons.js
```diff
@@ -0,0 +1,608 @@
+// TODO: torn, can either bake this here; or have to create a whole new button type
+// Only ways that you can pass in a custom React component for render :l
+import {
```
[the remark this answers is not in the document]
sedghi: I'm not sure I follow what you say here, are you saying in the cli tool I copy the file from the longitudinal mode? since modes depending on other modes are just not good

wayfarer3130 (approved):

PR #3370, opened 2023-05-05
fix: Toggle different study

wayfarer3130:
### Context

Drag and drop of a different study to the MPR mode, and then turning off the MPR mode should still work and preserve the old hanging protocol correctly even if the active study has changed.

### Changes & Results

It is probably difficult to see this change in the current version of OHIF because once a study is active in terms of the hanging protocol, there isn't an option to change which study is active.  

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


extensions/default/src/commandsModule.ts
```diff
@@ -279,38 +280,50 @@ const commandsModule = ({
...
         activeStudy,
       } = hangingProtocolService.getActiveProtocol();
       const { toggleHangingProtocol } = stateSyncService.getState();
-      const storedHanging = `${
-        activeStudy.StudyInstanceUID
-      }:${protocolId}:${stageIndex | 0}`;
+      // Just use the protocol id as the toggle id for toggling
+      // as the stage shouldn't matter, and one wants to restore even across
+      // active studies.
+      const toggleID = protocolId;
```
wayfarer3130, line 295: It is now possible to toggle on a hanging protcool with multiple stages, and then it won't toggle off until a different hanging protocol is applied or the toggle is directly called.  That allows for toggling on a mode such as MPR, and having commands which either apply a given stage or toggle the stage on if it is off, and those same commands will toggle the HP off.

sedghi:
@wayfarer3130 my PR which reduces the complexity of viewportGrid should be reviewed/merged first before this since that is the way to go forward

sedghi:
I can't reproduce what you mentioned in the Context in the PR description, it shows the first DS as you wanted

https://github.com/OHIF/Viewers/assets/7490180/d866480b-91a5-4100-8427-a74ee7bc1a03

sedghi:
I gave it another shot, but couldn't replicate the problem. So, I'll hold off on merging this for now until we can reproduce it step by step.

PR #3372, opened 2023-05-08
feat(custom displayOptions): add custom options to every level of hp

sedghi:
## Enhanced Custom attributes
Previously you could set your custom options using `{custom: 'mycustomattribute'}` inside the viewportOptions which was very handy, but had some limitations. First: you could not set it if you had array of objects or objects of objects etc. (nested) and you could not have it for displaySet options. This PR addresses those limitatoins. 

## Cleaner API for colormaps preset etc.
This PR also refactors various codes regarding colormap and preset and move them down inside Cornerstone3D 

## Move thumbnail rendering to use GPU
This PR switch the thumbnail rendering to GPU which was previously CPU

## Enhanced colormap setting in HP
Now you can set the opacityMapping for range to opacity function for a colormap in the HP, look the tmtv 

## Fix PT NAC window level issue 
Previously we were hard coding the PT VOI to be 5, and 2.5 but this PR makes it computed based on metadata via custom `getPTVOIRange`

### Testing

Everything should work like before in tmtv regarding the initial colormap and later colormap changes

PT NAC should have correct VOI on the thumbnail and also tmtv (if someone drag and drops it, which is turned off in the demo)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


wayfarer3130:
Thanks for including this - was just about to start it now, so just reviewing makes it easy.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -556,9 +555,13 @@ function commandsModule({
         return actorEntry.uid.includes(displaySetInstanceUID);
       });
 
-      const { actor: volumeActor } = actorEntry;
+      const { actor: volumeActor, uid: volumeId } = actorEntry;
 
-      setColormap(volumeActor, colormap);
+      // viewport.setProperties({ colormap }, volumeId);
+      const mapper = volumeActor.getMapper();
+      mapper.setSampleDistance(1.0);
```
wayfarer3130, line 562: Can you add a comment about what this is doing?  It feels really opaque to me why you added this code.

extensions/tmtv/src/utils/hpViewports.ts
```diff
@@ -312,10 +311,12 @@ const fusionSAGITTAL = {
     },
     {
       options: {
-        colormap: 'hsv',
+        colormap: {
+          name: 'hsv',
+          opacityMapping: [[0.1, 0.9]],
```
wayfarer3130, line 316: Is there an example of a non-double array instance?  Just seems odd to have a double array here.

extensions/cornerstone/src/commandsModule.ts
sedghi: nothing removed

extensions/cornerstone/src/services/ViewportService/Viewport.ts
```diff
@@ -63,7 +63,7 @@ export type PublicDisplaySetOptions = {
   blendMode?: string;
   slabThickness?: number;
   colormap?: string;
-  presetName?: string;
+  displayPreset?: string;
```
jbocce, line 66: Does this need to be changed and/or added in any of the (*.md) docs?

jbocce:
Please fix the cypress tests and preview deployment.

modes/tmtv/src/index.js
```diff
@@ -131,6 +133,43 @@ function modeFactory({ modeConfiguration }) {
         'RectangleROIStartEndThreshold',
         'fusionPTColormap',
       ]);
+
+      // For the hanging protocol we need to decided on the window level
```
jbocce, line 137: sp *decide

extensions/tmtv/src/utils/hpViewports.ts
sedghi: I'm not sure what is the best way to do this other than array of arrays, the idea is min value is mapped to 0 and max value is mapped to 1, and there is a linear interpolation for the rest. But if you provide this opacityMapping in the HP it can override that interpolation and add a point there which in this case it will add mapping of 0.1 to 0.9 opacity. Theoretically people can add as many points as they want. Other options is to do something like

```js
opacityMapping: [
 { value: 0.1, opacity: 0.9}
]
```

extensions/cornerstone/src/services/ViewportService/Viewport.ts
sedghi: I'm pretty sure I updated it everywhere

jbocce: Does it perhaps make sense to update the list of display set options in this section of the docs? https://v3-docs.ohif.org/platform/extensions/modules/hpModule#viewports

image

extensions/tmtv/src/utils/hpViewports.ts
wayfarer3130: I'd prefer an array of objects, as I find that easier to understand.

jbocce: I agree with Bill.

ladeirarodolfo:
minor contribution. 
your Pr should ensure version for this current other PR
https://github.com/cornerstonejs/cornerstone3D-beta/pull/602/files

jbocce:
> minor contribution. your Pr should ensure version for this current other PR https://github.com/cornerstonejs/cornerstone3D-beta/pull/602/files
> 
> Also the thumb, generated by gpu process, does not produce same result as cpu. See
> 
> Screen Shot 2023-05-09 at 11 09 45 AM

I was not able to reproduce the offset camera issue with the thumbnails that @ladeirarodolfo saw.

jbocce:
Looks like there is a `yarn.lock` conflict.

jbocce (commented):
Looks good. Just some minor documentation suggestions/changes.

jbocce (approved):
Approved

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -619,46 +635,68 @@ export default class HangingProtocolService extends PubSubService {
...
+        // Try to recompute the viewport options based on the current
+        // viewportIndex that needs update but from its old/original un-computed
+        // viewport & displaySet options
+        if (originalProtocolStage) {
+          const originalViewport = originalProtocolStage.viewports[index];
+          const originalViewportOptions = originalViewport.viewportOptions;
+          const originalDisplaySetOptions = originalViewport.displaySets;
+
+          const recomputedViewportOptions = this.getComputedOptions(
+            JSON.parse(JSON.stringify(originalViewportOptions)),
```
wayfarer3130, line 656: Can you do a deep copy here rather than parse(stringify

modes/tmtv/src/index.js
```diff
@@ -131,6 +133,43 @@ function modeFactory({ modeConfiguration }) {
         'RectangleROIStartEndThreshold',
         'fusionPTColormap',
       ]);
+
+      // For the hanging protocol we need to decide on the window level
+      // based on whether the SUV is corrected or not, hence we can't hard
+      // code the window level in the hanging protocol but we add a custom
+      // attribute to the hanging protocol that will be used to get the
+      // window level based on the metadata
+      hangingProtocolService.addCustomAttribute(
```
wayfarer3130, line 142: It feels like this should come as part of an extension rather than being in the mode directly.  I'm ok leaving it here for now, but eventually this sort of logic should get incorporated into an extension object.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -49,6 +51,9 @@ export default class HangingProtocolService extends PubSubService {
   studies: StudyMetadata[];
   // stores all the protocols (object or function that returns an object) in a map
   protocols: Map<string, Protocol>;
+  // protocols copy for the case where protocols has computed values which override
+  // the original values
+  protocolsCopy: Map<string, Protocol>;
```
wayfarer3130, line 56: Could we call one of them protocolsCustomized and the other just protocols?  protocolsCopy is really unclear to me as to what it means.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -135,6 +140,7 @@ export default class HangingProtocolService extends PubSubService {
     this._commandsManager = commandsManager;
     this._servicesManager = servicesManager;
     this.protocols = new Map();
+    this.protocolsCopy = new Map();
```
wayfarer3130, line 143: Please initialize in the declaration object - makes it immediately obvious what it is.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -294,6 +300,10 @@ export default class HangingProtocolService extends PubSubService {
     }
 
     this.protocols.set(protocolId, protocol);
+
+    // also insert it in the copy of the protocols map object, but make a deep
+    // copy of the protocol object first
+    this.protocolsCopy.set(protocolId, this._copyProtocol(protocol));
```
wayfarer3130, line 306: This should be done on use/starting to update the given protocol, so that a copy is made as required, rather than copying everything.  Since the protocols can be reasonably large, it reduces startup cost by not copying until needed - and indeed, not until usable.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -611,6 +621,12 @@ export default class HangingProtocolService extends PubSubService {
       return defaultReturn;
     }
 
+    const originalProtocol = this.protocolsCopy.get(this.protocol.id);
+    let originalProtocolStage;
+    if (!(originalProtocol instanceof Function)) {
```
wayfarer3130, line 626: If the copy is done at the time the protocol is assigned, with the original/unaltered version, then it helps startup time as well as avoiding any of these instanceof functions - because the function will have already been called correctly.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -619,46 +635,68 @@ export default class HangingProtocolService extends PubSubService {
...
+
+          const recomputedViewportOptions = this.getComputedOptions(
+            JSON.parse(JSON.stringify(originalViewportOptions)),
+            [newDisplaySetInstanceUID]
+          );
 
-            let {
-              displaySetInstanceUID: displaySetInstanceUIDToUse,
-            } = this.displaySetMatchDetails.get(id);
+          if (recomputedViewportOptions) {
```
wayfarer3130, line 660: Just assign viewport.viewportOptions = recomputedViewportOptions - that way in the future the getComputed call can compute the entire viewportOptions if desired, or just individual elements of it.  It also puts all the logic into the getComputed call, and you always start with the original instance of it, and produce a new instance (which may be identical to the old one)

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -681,27 +719,46 @@ export default class HangingProtocolService extends PubSubService {
...
+      );
+    }
+
     const computed = { ...options };
-    let displaySets;
-    for (const key in computed) {
-      const value = computed[key];
-      if (!value) continue;
+
+    const processValue = (value: any, parent: any, key: string) => {
```
wayfarer3130, line 733: You should check here for whether options has a custom attribute right here - and then call getComputedOptions recursively rather than processValue.  That allows you to replace the entire options, eg:
viewportOptions: { custom: 'myViewportOptions' }

wayfarer3130 (changes requested):
There are a few things that would be nice to change in the hanging protocol computed options area.  I think they are worthwhile because it is so hard to minimize startup costs that anything which can easily be done in that area is a good idea.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
sedghi: As talked we can do this later

modes/tmtv/src/index.js
sedghi: As talked we will do this later

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -619,46 +633,61 @@ export default class HangingProtocolService extends PubSubService {
...
+        // Try to recompute the viewport options based on the current
+        // viewportIndex that needs update but from its old/original un-computed
+        // viewport & displaySet options
+        if (originalProtocolStage) {
+          const originalViewport = originalProtocolStage.viewports[index];
+          const originalViewportOptions = originalViewport.viewportOptions;
+          const originalDisplaySetOptions = originalViewport.displaySets;
+
+          viewport.viewportOptions = this.getComputedOptions(
+            JSON.parse(JSON.stringify(originalViewportOptions)),
```
wayfarer3130, line 654: At least put the copy into getComputedOptions, as it is part of that operation

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -619,46 +633,61 @@ export default class HangingProtocolService extends PubSubService {
...
-            acc.displaySetInstanceUIDs.push(displaySetInstanceUIDToUse);
-            acc.displaySetOptions.push(displaySet);
+        const displaySetInstanceUIDs = [];
+        const displaySetOptions = [];
 
-            return acc;
-          },
-          { displaySetInstanceUIDs: [], displaySetOptions: [] }
-        );
+        for (const prevDisplaySet of viewport.displaySets) {
```
wayfarer3130, line 678: Can you drop this into a function - just that the code is already pretty long, and making it into a single line makes the whole function easier to understand.

wayfarer3130 (changes requested):
Just a couple of small final cleanup changes

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
sedghi: I put a copy at the end of the getComputed, but we need this copy here, it is the copy of the original, if we put it inside the computed it is just deep copying recursively which might not be necessary

jbocce (approved):

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -611,6 +619,38 @@ export default class HangingProtocolService extends PubSubService {
       return defaultReturn;
     }
 
+    const originalProtocol = this._originalProtocol;
+    let originalProtocolStage;
+    if (!(originalProtocol instanceof Function)) {
+      originalProtocolStage = originalProtocol.stages[this.stageIndex];
+    }
+
+    function updateDisplaySetInstanceUIDs(
```
wayfarer3130, line 628: Please define this outside the function - you aren't taking accessing any local variables AFAIK, so you can just call it as either:
HangingProtocolService.updateDisplaySetInstanceUIDs
or this.update....

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -681,28 +720,57 @@ export default class HangingProtocolService extends PubSubService {
...
-              displaySetUIDs.indexOf(displaySet.displaySetInstanceUID) !== -1
+  ): any {
+    // Base case: if options is an array, map over the array and recursively call getComputedOptions
+    if (Array.isArray(options)) {
+      return options.map(option =>
+        this.getComputedOptions(option, displaySetUIDs)
+      );
+    }
+
+    if (typeof options === 'object' && options !== null) {
```
wayfarer3130, line 733: You might return null if options is null, and then assume options is an array, so that you nest indentation 1 less, which makes it a bit easier to understand, eg:
if (Array...) ...

if( options===null ) return null;
// Assume options is object here - it has to be given the possibilities

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -681,28 +720,57 @@ export default class HangingProtocolService extends PubSubService {
...
+      // If options is an object with a custom attribute, compute a new options object
+      if ('custom' in options) {
```
wayfarer3130, line 735: Prefer options.custom because what you actually care about is whether it is set to a value, and 0 doesn't count.  The in test can be useful at times, but isn't the right choice here.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -681,28 +720,57 @@ export default class HangingProtocolService extends PubSubService {
...
-    let displaySets;
-    for (const key in computed) {
-      const value = computed[key];
-      if (!value) continue;
-      if (value.custom) {
-        if (!displaySets) {
-          displaySets = this.displaySets.filter(
-            displaySet =>
-              displaySetUIDs.indexOf(displaySet.displaySetInstanceUID) !== -1
+  ): any {
```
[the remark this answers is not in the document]
sedghi: no it will not be as it is recursive

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -681,28 +717,54 @@ export default class HangingProtocolService extends PubSubService {
...
+    const newOptions = {} as Record<string, unknown>;
+    for (const key in options) {
+      // if not undefined
+      if (options[key]) {
+        newOptions[key] = this.getComputedOptions(options[key], displaySetUIDs);
+      }
+    }
+
+    // give the copy
+    return JSON.parse(JSON.stringify(newOptions));
```
wayfarer3130, line 767: You have just implicitly copied options, so you don't need to do it again.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -681,28 +717,54 @@ export default class HangingProtocolService extends PubSubService {
...
+          `Custom key "${customKey}" not found in customAttributeRetrievalCallbacks.`
+        );
+      }
+
+      const callback = this.customAttributeRetrievalCallbacks[customKey]
+        .callback;
+      const newOptions = callback.call(options, displaySets);
+
+      if (newOptions === undefined) {
+        return options.defaultValue;
```
wayfarer3130, line 751: How about:
newOptions = options.defaultValue
so that you have the option of having nested computed values in the response.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -619,45 +633,44 @@ export default class HangingProtocolService extends PubSubService {
...
+        // Try to recompute the viewport options based on the current
+        // viewportIndex that needs update but from its old/original un-computed
+        // viewport & displaySet options
+        if (originalProtocolStage) {
+          const originalViewport = originalProtocolStage.viewports[index];
+          const originalViewportOptions = originalViewport.viewportOptions;
+          const originalDisplaySetOptions = originalViewport.displaySets;
+
+          viewport.viewportOptions = this.getComputedOptions(
+            JSON.parse(JSON.stringify(originalViewportOptions)),
```
wayfarer3130, line 654: Really, this is unneeded - the getComputed options does not modify the provided value.

wayfarer3130 (approved):

PR #3376, opened 2023-05-09
fix: Load issues with changes to lazy mode/extension loading

wayfarer3130:
### Context

There are three changes being made in this PR to resolve the issues with the lazy loading:

* Fix the Mode.tsx loading timeline to take into account the end of loading the modules in more places
* Fix the loading of modules so that the service itself can handle the module init in order to load default values after service init
* Add documentation and configuration for a custom PLUGIN_CONFIG file to allow updating the plugin config (this was supposed to have been added to the original change but got missed - it allows custom versions of the plugins without necessary changing any code in OHIF)

### Changes & Results

Before this PR, the render methods would start too early on the Mode init, causing renders to fail.  As well, the service init on the customization service was handling the load of default modules, which was too early to complete the load, so this caused Mode loaded modules to not get initialized.

### Testing

Test loading the segmentation, exiting the mode and coming back in and loading the SEG/RT again.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


sedghi:
I will close this PR in favor of the other PR that works for now until you get back to this and fix the rest of the issues https://github.com/OHIF/Viewers/pull/3377/files

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -1524,4 +1525,14 @@ export default class HangingProtocolService extends PubSubService {
   _copyProtocol(protocol: Protocol) {
     return JSON.parse(JSON.stringify(protocol));
   }
+
+  /** Register the hanging protocols from the extension module */
+  public initModule(extensionModule, extensionId: string): void {
```
wayfarer3130, line 1530: extensonModule => hangingProtocolModule
remove extensionId - it is still there, but unused

platform/core/src/services/CustomizationService/CustomizationService.ts
```diff
@@ -87,6 +88,16 @@ export default class CustomizationService extends PubSubService {
     });
   }
 
+  /** Register the customizations from the extensions */
+  public initModule(extensionModule, extensionId: string): void {
```
wayfarer3130, line 92: Remove extensionId

platform/core/src/extensions/ExtensionManager.ts
```diff
@@ -312,6 +318,20 @@ export default class ExtensionManager {
...
+      if (!element.name) {
+        throw new Error(
+          `Extension ID ${extensionId} module ${moduleType} element has no name`
+        );
+      }
+      const id = `${extensionId}.${moduleType}.${element.name}`;
+      element.id = id;
+      this.modulesMap[id] = element;
+    });
+    service?.initModule?.(extensionModule, extensionId);
```
wayfarer3130, line 332: service?.onServiceModuleRegistration?.(....

sedghi:
we should continue on this later

wayfarer3130:
Will reconsider this in the future with a more extensive change

PR #3385, opened 2023-05-11
fix(Measurements): in MeasurementService keep a separate set of unmapped measurements

jbocce:
### Context

Measurements cannot be deleted in the non-acquisition views of MPR or TMTV.

### Changes & Results

When a measurement is created there is a lookup of various metadata based on the image id. The image id does not exist for the non-acquisition views/series. An exception is thrown as  a result. In `MeasurementService` we now catch this exception and add the measurement to a set of "unmapped" measurement.

### Testing

Attempt to delete an measurement that was added in OHIF. Note that measurements added in non-acquisition views will NOT be tracked.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 
- [x] Node version: 16.14.0
- [x] Browser: Chrome 113.0.5672.93
  

### DIFF
--- a/extensions/tmtv/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
+++ b/extensions/tmtv/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
@@ -4,6 +4,8 @@ export default function getSOPInstanceAttributes(imageId) {
   if (imageId) {
     return _getUIDFromImageID(imageId);
   }
+
+  return {};
 }
 
 function _getUIDFromImageID(imageId) {
--- a/extensions/tmtv/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
+++ b/extensions/tmtv/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
@@ -4,8 +4,6 @@ export default function getSOPInstanceAttributes(imageId) {
   if (imageId) {
     return _getUIDFromImageID(imageId);
   }
-
-  return {};
 }
 
 function _getUIDFromImageID(imageId) {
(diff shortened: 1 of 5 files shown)


extensions/tmtv/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
```diff
@@ -4,6 +4,8 @@ export default function getSOPInstanceAttributes(imageId) {
   if (imageId) {
     return _getUIDFromImageID(imageId);
   }
+
+  return {};
```
sedghi, line 8: I used `measurementService.getMeasurements` and it gave all the measurements (including those that are empty and with no SOP). We should keep them separate as they are non-exportable and we don't have support for SR for SCOORD3D

jbocce: Yes good catch, thank you so much. I will make updates.

platform/core/src/services/MeasurementService/MeasurementService.test.js
```diff
@@ -429,5 +451,75 @@ describe('MeasurementService.js', () => {
 
       expect(updateCallbackWasCalled).toBe(false);
     });
+
+    it('subscribers do NOT receive add unmapped measurements event', () => {
```
sedghi, line 455: really like it

platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -112,6 +112,7 @@ class MeasurementService extends PubSubService {
   public readonly VALUE_TYPES = VALUE_TYPES;
 
   private measurements = new Map();
+  private unmappedMeasurements = new Set();
```
sedghi, line 115: could take benefit from some comments above it about its purpose

platform/core/src/services/MeasurementService/MeasurementService.test.js
jbocce: thanks

sedghi (approved):

PR #3391, opened 2023-05-14
feat: query string dynamic app config

anarnoli:
### Context
This PR allows for the app config to reference a json over network. This allows for
fast swapping of data source configuration, OIDC, etc, without recompiling the application.

If the `configUrl` query string is passed, the worklist and modes will load from the referenced json
rather than the default .env config

If there is no `configUrl` path provided, the default behaviour is used and there should not be any
deviation from current user experience.

**Points to consider**

- User have to enable this feature by setting `dangerouslyUseDynamicConfig.enabled:true`. By default it is `false`.
- Regex helps to avoid easy exploit
- User must set `cross-origin = same-origin` to avoid potential harder exploit

**Sample local config**
```
/* Dynamic config allows user to pass "configUrl" query string this allows to load config without recompiling application. The regex will ensure valid configuration source */
  dangerouslyUseDynamicConfig: {
    enabled: false,
    regex: /.*/,
  }
```

**Example usage**

`http://localhost:3000/?configUrl=http://localhost:3000/config/example.json`

This PR is related to another PR https://github.com/OHIF/Viewers/pull/2925

#### PR

#### Code

  etc.)

  additions or removals.

### DIFF
--- a/extensions/default/src/ViewerLayout/index.tsx
+++ b/extensions/default/src/ViewerLayout/index.tsx
@@ -50,6 +50,8 @@ function ViewerLayout({
     //   dataSourceIdx === -1
     //     ? undefined
     //     : `datasources=${pathname.substring(dataSourceIdx + 1)}`;
+
+    // Todo: Handle parameters in a better way.
     const query = new URLSearchParams(window.location.search);
     const configUrl = query.get('configUrl');
 
--- a/extensions/default/src/ViewerLayout/index.tsx
+++ b/extensions/default/src/ViewerLayout/index.tsx
@@ -46,13 +46,25 @@ function ViewerLayout({
   const onClickReturnButton = () => {
     const { pathname } = location;
     const dataSourceIdx = pathname.indexOf('/', 1);
-    const search =
-      dataSourceIdx === -1
-        ? undefined
-        : `datasources=${pathname.substring(dataSourceIdx + 1)}`;
+    // const search =
+    //   dataSourceIdx === -1
+    //     ? undefined
+    //     : `datasources=${pathname.substring(dataSourceIdx + 1)}`;
+    const query = new URLSearchParams(window.location.search);
+    const configUrl = query.get('configUrl');
+
+    const searchQuery = new URLSearchParams();
+    if (dataSourceIdx !== -1) {
+      searchQuery.append('datasources', pathname.substring(dataSourceIdx + 1));
+    }
+
+    if (configUrl) {
+      searchQuery.append('configUrl', configUrl);
+    }
+
     navigate({
       pathname: '/',
-      search,
+      search: decodeURIComponent(searchQuery.toString()),
     });
   };
 
--- a/platform/viewer/public/config/default.js
+++ b/platform/viewer/public/config/default.js
@@ -26,15 +26,15 @@ window.config = {
   // filterQueryParam: false,
   defaultDataSourceName: 'dicomweb',
   /* Dynamic config allows user to pass "configUrl" query string this allows to load config without recompiling application. The regex will ensure valid configuration source */
-  dangerouslyUseDynamicConfig: {
-    enabled: false,
-    // regex will ensure valid configuration source and default is /.*/ which matches any character. To use this, setup your own regex to choose a specific source of configuration only.
-    // Example 1, to allow numbers and letters in an absolute or sub-path only.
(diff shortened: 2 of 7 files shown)


anarnoli:
@sedghi @wayfarer3130 - request to review this PR.

@cupidchan FYI.

extensions/default/src/ViewerLayout/index.tsx
```diff
@@ -46,13 +46,25 @@ function ViewerLayout({
...
-        ? undefined
-        : `datasources=${pathname.substring(dataSourceIdx + 1)}`;
+    // const search =
+    //   dataSourceIdx === -1
+    //     ? undefined
+    //     : `datasources=${pathname.substring(dataSourceIdx + 1)}`;
+    const query = new URLSearchParams(window.location.search);
+    const configUrl = query.get('configUrl');
+
+    const searchQuery = new URLSearchParams();
```
wayfarer3130, line 56: Can you just add a TODO here to handle the remembering of the parameters in a better way?  Really we should add the relevant parameters to the other queries, and then restore the full query string on going back, but that is NOT your problem here, it is just that I want a reminder that having the viewer page know about hte query page isn't a good idea.

platform/viewer/public/config/default.js
```diff
@@ -25,6 +25,16 @@ window.config = {
   },
   // filterQueryParam: false,
   defaultDataSourceName: 'dicomweb',
+  /* Dynamic config allows user to pass "configUrl" query string this allows to load config without recompiling application. The regex will ensure valid configuration source */
+  dangerouslyUseDynamicConfig: {
```
wayfarer3130, line 29: You can leave the configuration commented out with the enabled: true set to show how to enable it.  That way the entire block just works if someone uncomments it.

platform/viewer/src/loadDynamicConfig.js
```diff
@@ -0,0 +1,23 @@
+export default async config => {
+  const useDynamicConfig = config.dangerouslyUseDynamicConfig;
+
+  // Check if dangerouslyUseDynamicConfig enabled
+  if (useDynamicConfig.enabled) {
```
wayfarer3130, line 5: This has to be useDynamicConfig?.enabled so that if the value isn't configured, it won't throw.

platform/viewer/src/loadDynamicConfig.js
```diff
@@ -0,0 +1,23 @@
...
+    // If enabled then get configUrl query-string
+    let query = new URLSearchParams(window.location.search);
+    let configUrl = query.get('configUrl');
+
+    if (configUrl) {
+      // validate regex
+      const regex = useDynamicConfig.regex;
+
+      if (regex.test(configUrl)) {
```
wayfarer3130, line 14: Use:
configUrl.match(regex) so that if a string is provided int he configuration, it doesn't fail to work.

platform/viewer/src/routes/WorkList/WorkList.tsx
```diff
@@ -345,12 +345,17 @@ function WorkList({
...
+            if (filterValues.configUrl) {
+              query.append('configUrl', filterValues.configUrl);
+            }
+            query.append('StudyInstanceUIDs', studyInstanceUid);
             return (
               <Link
                 key={i}
                 to={`${dataPath ? '../../' : ''}${mode.routeName}${dataPath ||
-                  ''}?StudyInstanceUIDs=${studyInstanceUid}`}
+                  ''}?${decodeURIComponent(query.toString())}`}
```
wayfarer3130, line 358: This fails for the situation where the configUrl includes parameters, such as:
configUrl=https://myhost/config?a=b&c=d
because the & is encoded/decoded.  The decodeURIComponent, as the name says, needs to apply to individual components.

wayfarer3130 (changes requested):
Just on bug to fix and a couple of comments/very small improvements.

platform/viewer/public/config/default.js
anarnoli: I agree @wayfarer3130 .

platform/viewer/src/routes/WorkList/WorkList.tsx
anarnoli: @wayfarer3130  Do you mean decodeURIComponent should be removed and original queryString should be passed?

wayfarer3130 (approved):

PR #3406, opened 2023-05-18
feat(Wado) : Customizable TS

salimkanoun:
### Changes & Results

This PR customize the dicom web Accept request headers to specifiy a dicom transfer syntax (to be specified in ohif config), to optimize bandwith usage

### Testing

To be tested in a Orthanc / DCM4CHE backend, payload answer shall be smaller using a compressed TS

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Linux Ubuntu
- [x] "Node version: 18
- [x] "Browser: Chome


platform/viewer/public/config/default.js
```diff
@@ -16,6 +16,7 @@ window.config = {
   showCPUFallbackMessage: true,
   showLoadingIndicator: true,
   strictZSpacingForVolumeViewport: true,
+  requestTransferSyntaxUID: '1.2.840.10008.1.2.4.80',
```
sedghi, line 19: most likely this is put here by mistake

salimkanoun: it was to show the example, i remove it

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -54,20 +54,39 @@ export default function initWADOImageLoader(
...
       // Request:
       // JPEG-LS Lossless (1.2.840.10008.1.2.4.80) if available, otherwise accept
       // whatever transfer-syntax the origin server provides.
       // For now we use image/jls and image/x-jls because some servers still use the old type
       // http://dicom.nema.org/medical/dicom/current/output/html/part18.html
+
+      //Initialize the accept header content
+      const acceptHeader = ['multipart/related']
+      //Storing the requesting transfer syntax specified in OHIF config
+      const requestTransferSyntaxUID = appConfig.requestTransferSyntaxUID
```
wayfarer3130, line 69: This parameter has to come from the data source configuration and NOT the appConfig generally, because different back ends can handle things differently.

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -54,20 +54,39 @@ export default function initWADOImageLoader(
...
+
+      if (requestTransferSyntaxUID) {
+        //Check TS is valid
+        if (!Object.keys(typeForTS).includes(requestTransferSyntaxUID)) {
+          console.warn(requestTransferSyntaxUID + 'is unexpected')
+        } else {
+          //Fetch type header according to requesting TS
+          let type = typeForTS[requestTransferSyntaxUID]
+          if (type === 'application/octet-stream' && !appConfig.omitQuotationForMultipartRequest) {
```
wayfarer3130, line 78: This isn't just for application/octet stream, but is for all types.
Also, the configuration has to come from the data source and not the appConfig generally.

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -54,20 +54,39 @@ export default function initWADOImageLoader(
...
       // http://dicom.nema.org/medical/dicom/current/output/html/part18.html
+
+      //Initialize the accept header content
+      const acceptHeader = ['multipart/related']
+      //Storing the requesting transfer syntax specified in OHIF config
+      const requestTransferSyntaxUID = appConfig.requestTransferSyntaxUID
+
+      if (requestTransferSyntaxUID) {
+        //Check TS is valid
+        if (!Object.keys(typeForTS).includes(requestTransferSyntaxUID)) {
```
wayfarer3130, line 73: This is a slow way of doing things - just use if( !typeForTS[requestTransferSyntaxUID] ) is sufficient.

extensions/cornerstone/src/initWADOImageLoader.js
sedghi: makes sense

salimkanoun: Agreed but how do you know in this function which is the dicom archive parameter being used ?

wayfarer3130, line 69: This should be the default TSUID, but the actual request should be able to over-ride the requested TSUID.  Why not just allow configuring the default accept header - INCLUDING making it empty?  If it were set as an array of strings, then an empty array wouldn't set it, and would allow the default response to be sent.  Also, it is permissable to request single part types directly, and it would be nice to allow that to be configured using the same mechanism - for example:
dataSource....defaultImageAccept: []   would not pass any accept header (this is required to work on all DICOMweb servers)
dataSource...defaultImageAccept: ['multipart/related; type="image/jphc"', 'multipart/related; type="image/jpeg"] would allow for either HTJ2K or JPEG responses - something that can be very important when fetching existing lossy encoded images.
Then,
dataSource...defaultImageAccept: ['image/jpeg', 'image/jp2'] would return JPEG or JPEG2000 responses.

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -83,6 +102,28 @@ export default function initWADOImageLoader(
   initWebWorkers(appConfig);
 }
 
+/*
+taken from
+https://hg.orthanc-server.com/orthanc-dicomweb/file/tip/Plugin/WadoRsRetrieveFrames.cpp
+*/
+const typeForTS = {
```
wayfarer3130, line 109: Hard coding this prevents adding custom types here such as the single part, or adding priority.  Suggest dropping this table.

wayfarer3130 (changes requested):
I'd really like to see this switch to a specific list of accept headers as there are cases that having a hard coded value will not allow for, such as lossy encoded JPEG for some images, and JPEG-LS for others, or for single bit images having a different compression scheme.

extensions/cornerstone/src/initWADOImageLoader.js
salimkanoun: moving this comment in dicomwebdatasource index.js (WIP)

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -59,23 +59,104 @@ function createDicomWebApi(dicomWebConfig, userAuthenticationService) {
     supportsReject,
     staticWado,
     singlepart,
+    requestTransferSyntaxUID,
+    omitQuotationForMultipartRequest
   } = dicomWebConfig;
 
   const dicomWebConfigCopy = JSON.parse(JSON.stringify(dicomWebConfig));
 
+  const acceptHeader = ['multipart/related']
```
sedghi, line 68: What I thought Bill's comment was (@wayfarer3130 ) was that the configuration should come from the data source not the logic 

like this 

```
{
      friendlyName: 'dcmjs DICOMWeb Server',
      sourceName: 'dicomweb',
      configuration: {
        name: 'aws',
     	acceptHeader: 'Blahblahblah',
        wadoUriRoot: 'https://d33do7qe4w26qo.cloudfront.net/dicomweb',
        qidoRoot: 'https://d33do7qe4w26qo.cloudfront.net/dicomweb',
        wadoRoot: 'https://d33do7qe4w26qo.cloudfront.net/dicomweb',

        supportsReject: false,
        imageRendering: 'wadors',
        thumbnailRendering: 'wadors',
        enableStudyLazyLoad: true,
        supportsFuzzyMatching: false,
        supportsWildcard: true,
        staticWado: true,
        singlepart: 'bulkdata,video',
        bulkDataURI: {
          enabled: true,
          relativeResolution: 'studies',
        },
      },
```

note the      	acceptHeader: 'Blahblahblah',

salimkanoun: Yes I was thinking on it. But why limiting it to the accept header ? We could also let the user define the whole header including authentification.
Maybe we need both way
- one for regular users that will have options such as existing now and in which we added the requested TS
- a header property for advanced user that would want to build their herder from scratch (will allow to set a custom authorization for each dicom web server as for now we have 1 authorization service that apply for all)

What do you think ?

sedghi: for this PR I beleive the accept header is good enough, and we can do the header property in the next PR

salimkanoun: OK let's try this way

salimkanoun:
hi gues, I don't get it, i don't see any effect of the header option after my changes, 
Are we sure instances retrieve request are made by DICOMwebClient  ?

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -60,23 +60,47 @@ function createDicomWebApi(dicomWebConfig, userAuthenticationService) {
...
     staticWado,
     singlepart,
+    omitQuotationForMultipartRequest,
+    acceptHeader
   } = dicomWebConfig;
 
   const dicomWebConfigCopy = JSON.parse(JSON.stringify(dicomWebConfig));
 
+  //If accept header is not specified, set as multipart
+  let formattedAcceptHeader = acceptHeader ?? []
```
wayfarer3130, line 70: You are dealing with slightly different lists - the formatted accept header is a single line accept header that has parts separated by semi-colon, while the acceptHeader is a list of header values that are complete - for example:
acceptHeader: [ 'multipart/related; type=image/jpls; q=1', 'multipart/related; type=application/octet-stream; q=0.1' ]
so that one can prioritize the headers.

wayfarer3130 (changes requested):
Need to fix the accept headers setting

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -60,23 +60,47 @@ function createDicomWebApi(dicomWebConfig, userAuthenticationService) {
     supportsReject,
     staticWado,
     singlepart,
+    omitQuotationForMultipartRequest,
```
wayfarer3130, line 63: Making this parameter data source specific is definitely useful, thank you.

extensions/default/src/DicomWebDataSource/index.js
salimkanoun: it will mean I have to split each header by the ; separator to add the quotation and reserialize it ? 
At it going to provided by the user as an array, the user can already decide priority when building the array that we are going to use "as it"? can you detail the usecase for quotation ? 

In my testing my header made in this file doesn't change anything in dicom web requests, do you have an idea why my changes are not propagated in the requests ?

salimkanoun:
Hi guys, here a new iteration,

Some remaining questions,

- I created a generateAcceptHeader function to build the accept header using config but for now this fuction is duplicated in initWadoImageLoader and index of DicomWebDataSource. I wanted to make a common function that is simply imported, but the question is where to put this function ? We are in both extension, i guess it would be wired to import a file from another extension as i guess you want them code-independant. (or maybe as beforeSend should diseapear in favor of DicomWebDataSource we let it duplicated for now ?)

- Orthanc does not seems to take account of accept priority using q=, i wonder if dcm4che implement it properly?

- What will be the strategy getAuthorizationHeader that is not scoped by source ? I think we can pass source name in argument so the implementation of this function can return a specific authorization header for each source. But i see this method being called in CornestoneCacheService and dicomLoaderService (without understing what they stand for)

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -83,6 +79,34 @@ export default function initWADOImageLoader(
...
 
+const generateAcceptHeader = (configAcceptHeader = [], omitQuotationForMultipartRequest = false) => {
+  let acceptHeader = []
+  if (configAcceptHeader.length === 0) {
+    acceptHeader.push('multipart/related; type=application/octet-stream')
+  } else {
+    acceptHeader = configAcceptHeader
+  }
+
+  if (!omitQuotationForMultipartRequest) {
```
wayfarer3130, line 90: The point of the configuration accept header is to provide a header as is, without needing any tweaks.  The point of providing a transfer syntax UID is that you can easily configure a single TSUID in the right type, and not need to worry about it - also, so that the TSUID can come from the query response, eg in an AvailableTransferSyntaxUID method.
So, if the config accept header  is provided, use it - either as a comma separated string, or as an array (I had forgotten you could do either), and if it is provided as a transfer syntax UID, then use the omit quotation for multipart request, and provide a single header response in that TSUID.

extensions/cornerstone/src/initWADOImageLoader.js
```diff
@@ -83,6 +79,34 @@ export default function initWADOImageLoader(
   initWebWorkers(appConfig);
 }
 
+const generateAcceptHeader = (configAcceptHeader = [], omitQuotationForMultipartRequest = false) => {
```
wayfarer3130, line 82: If you extract this function into a separate file, then it becomes easy to duplicate in two places - there isn't really a good solution for sharing it otherwise, but having it as a file makes it fairly easy.

extensions/cornerstone/src/initWADOImageLoader.js
salimkanoun: Ok thanks, here is a new version

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -60,23 +60,44 @@ function createDicomWebApi(dicomWebConfig, userAuthenticationService) {
...
   const dicomWebConfigCopy = JSON.parse(JSON.stringify(dicomWebConfig));
 
+  //Generate accept header depending on config params
+  let formattedAcceptHeader = utils.generateAcceptHeader(
+    acceptHeader,
+    requestTransferSyntaxUID,
+    omitQuotationForMultipartRequest
+  );
+
+  const authHeaders = userAuthenticationService.getAuthorizationHeader();
```
wayfarer3130, line 77: You can't add the authorization header and then forget about it - the problem is that it gets updated/changed as it gets renewed, so you need to fetch it at the time of use.  I would suggest creating a method that generates the full set of headers/updates them as required.

platform/core/src/utils/generateAcceptHeader.ts
```diff
@@ -0,0 +1,58 @@
+const generateAcceptHeader = (
+  configAcceptHeader = [],
+  requestTransferSyntaxUID = null,
+  omitQuotationForMultipartRequest = false
+): string[] => {
+  //if acceptedHeader is passed by config use it as it.
+  if (configAcceptHeader.length > 0) {
```
wayfarer3130, line 7: This looks good - thank you for updating that.

extensions/default/src/DicomWebDataSource/index.js
salimkanoun: OK I got it that's why the get authorization header was duplicated, will fix it

salimkanoun:
Now authorization header being updated at each request (uses same logic that initial code)

wayfarer3130 (approved):
Thank you for all the work on this - it looks good to me.

sedghi (changes requested):
we should move the omitQuoatation in all configs too

image

salimkanoun:
Here it is

sedghi (approved):

PR #3407, opened 2023-05-18
fix(Measurement): fixed line dashing of measurements in various scenarios

jbocce:
### Context

For a stack viewport, when swapping in a new series either by double click or drag and drop, if the swapped in series had tracked measurements, they were still being rendered with a dashed line.

For a volume viewport (like MPR), the acquisition plane viewport with tracked measurements would render the measurements with dashed lines on first entering MPR; double clicking to/from one up in MPR and swapping in a different series into the MPR view.

### Changes & Results

For a StackViewport simply rely on the SeriesInstanceUID being part of the tracked measurements or not.

For a VolumeViewport first check the image id of the current image in the viewport and then check the SeriesInstanceUID (like for a StackViewport) if an id is returned. To reliably get the id of the current image, listen for the VOLUME_VIEWPORT_NEW_VOLUME event.

### Testing

Track measurements and swap in a different series and return to the tracked series ensuring measurements are always rendered with the correct dashing.

In MPR, ensure that the plane of acquisition view displays solid line measurements when it is being tracked.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 
- [x] Node version: 16.14.0
- [x] Browser: 113.0.5672.127
  

### DIFF
--- a/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.tsx
+++ b/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import OHIF, { utils } from '@ohif/core';
 
@@ -8,6 +8,7 @@ import { useTranslation } from 'react-i18next';
 
 import { annotation } from '@cornerstonejs/tools';
 import { useTrackedMeasurements } from './../getContextModule';
+import { BaseVolumeViewport, Enums, VolumeViewport } from '@cornerstonejs/core';
 
 const { formatDate } = utils;
 
@@ -34,6 +35,7 @@ function TrackedCornerstoneViewport(props) {
   const [trackedMeasurements] = useTrackedMeasurements();
   const [isTracked, setIsTracked] = useState(false);
   const [trackedMeasurementUID, setTrackedMeasurementUID] = useState(null);
+  const [viewportElem, setViewportElem] = useState(null);
 
   const { trackedSeries } = trackedMeasurements.context;
   const viewportId = viewportOptions.viewportId;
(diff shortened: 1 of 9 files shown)


extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.tsx
```diff
@@ -55,6 +57,62 @@ function TrackedCornerstoneViewport(props) {
...
+  }, []);
+
+  useEffect(updateIsTracked, [updateIsTracked]);
+
+  /**
+   * This useEffect is needed for volume viewports so that an current image id
+   * can be reliably fetched once the volume is actually loaded.
+   */
+  useEffect(() => {
+    viewportElem?.addEventListener(
```
sedghi, line 103: can you put this addEventListener in the onElementEnabled and the clean up in the onElementDisable so that we get rid of this effect?

jbocce: Yes that would be a very good thing to do. However I tried to do but noticed a few things...

- in `OHIFCornerstoneViewport`, `onElementDisabled` is at present passing a null argument (for `viewportInfo`) because once `cornerstoneViewportService.disableElement` is called, `viewportInfo` is null (see https://github.com/OHIF/Viewers/blob/f2004aee9eae1680c0f2e4c46f102015d197c747/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx#L340)
- I tried to instead get the `viewportInfo` prior to calling `disableElement`, but then `viewportInfo.element` was null in the `onElementDisabled` callback

I checked other places that are using `onElementEnabled` and `onElementDisabled` and each of the other places uses them as they are here (see https://github.com/OHIF/Viewers/blob/f2004aee9eae1680c0f2e4c46f102015d197c747/extensions/cornerstone-dicom-rt/src/viewports/OHIFCornerstoneRTViewport.tsx#L86).

I think it might be best to leave as is for now. Thoughts?

jbocce: You are correct. I can leave the state and make this work. Thanks.

sedghi (approved):
I tested it, worked great, thanks

Issue #3112, opened 2023-01-11
Previous Next toolbar buttons


olivert1969:
What feature or change would you like to see made?
meteor version had Previous and Next toolbar buttons - very efficient for echocardiography workflow
these buttons are missing from v3
meteor used moveDisplaySets - also missing from OHIF v3

any suggestions?

sedghi:
Can you describe what those buttons should do?

olivert1969:
move displayset to previous or next DICOM instance in the Ultrasound series
was fully functional in Meteor version

sedghi:
Can you create a video on the meteor version to show how it looks? I joined the project after the Meteor version and I'm curious to know how it worked before

olivert1969:
[OHIF Viewer - Google Chrome - 12 January 2023 - Watch Video](https://www.loom.com/share/ecfd60bd9f464c7f9a2f4046556b915e)

sedghi:
That's what I expected. Adding buttons back to the toolbar is not in the roadmap, but fixing the keyboard shortcut to go to next/previous displayset is something we need to fix (CC @jbocce )

xuliangming:
请教影像克隆下来的界面跟官网的不一样 克隆下来的工具栏没有窗宽窗位的功能

sedghi:
We had this [here](https://github.com/OHIF/Viewers/blob/368b8308a4fa620123677d4e23b78bf83eba251a/platform/viewer/src/appExtensions/GenericViewerCommands/commandsModule.js#L8), we will add it to commands and you can add the toolbar icon to execute that command to not have regression with V2

olivert1969:
thanks for this - greatly appreciated - will avoid LOTS of additional clicks by ultrasound users







---- On Wed, 17 May 2023 11:54:26 -0400 Alireza ***@***.***> wrote ---





We had this https://github.com/OHIF/Viewers/blob/368b8308a4fa620123677d4e23b78bf83eba251a/platform/viewer/src/appExtensions/GenericViewerCommands/commandsModule.js#L8, we will add it to commands and you can add the toolbar icon to execute that command to not have regression with V2

PR #3409, opened 2023-05-19
fix(command): Added command and hotkey binding for navigate to next and previous display set.

jbocce:
### Context

In prior versions of OHIF, it was possible to associate a button with a command to navigate the series in the active viewport forward or backward.

See https://github.com/OHIF/Viewers/issues/3112

### Changes & Results

The `updateViewportDisplaySet` command was added to match the equivalent command in v2. `PgUp` and `PgDn` hotkey bindings were added to navigate to the previous and next display sets respectively. An option was included to exclude non-image modalities for the navigation.

### Testing

View a multi-series study in OHIF and use the `PgUp` and `PgDn` keys to navigate the series in the active viewport.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 
- [x] Node version: 16.14.0
- [x] Browser: Chrome 113.0.5672.127
  

### DIFF
--- a/extensions/default/src/commandsModule.ts
+++ b/extensions/default/src/commandsModule.ts
@@ -560,7 +560,8 @@ const commandsModule = ({
         'RTDOSE',
       ];
 
-      // Sort the display sets as per the hanging protocol service which is also consistent with the thumbnail list.
+      // Sort the display sets as per the hanging protocol service viewport/display set scoring system.
+      // The thumbnail list uses the same sorting.
       const dsSortFn = hangingProtocolService.getDisplaySetSortFunction();
       const currentDisplaySets = [...displaySetService.activeDisplaySets];
 
--- a/extensions/default/src/commandsModule.ts
+++ b/extensions/default/src/commandsModule.ts
@@ -560,10 +560,11 @@ const commandsModule = ({
         'RTDOSE',
       ];
 
-      const currentDisplaySets = displaySetService.activeDisplaySets;
+      // Sort the display sets as per the hanging protocol service which is also consistent with the thumbnail list.
+      const dsSortFn = hangingProtocolService.getDisplaySetSortFunction();
+      const currentDisplaySets = [...displaySetService.activeDisplaySets];
 
-      // Sort the display sets as per the thumbnail/study browser list
-      utils.sortBySeriesDate(currentDisplaySets);
+      currentDisplaySets.sort(dsSortFn);
 
       const { activeViewportIndex, viewports } = viewportGridService.getState();
 
--- a/extensions/default/src/commandsModule.ts
+++ b/extensions/default/src/commandsModule.ts
@@ -23,6 +23,11 @@ export type HangingProtocolParams = {
   stageId?: string;
 };
 
+export type UpdateViewportDisplaySetParams = {
+  direction: number;
+  excludeNonImageModalities?: boolean;
+};
+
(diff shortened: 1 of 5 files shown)


extensions/default/src/commandsModule.ts
```diff
@@ -541,6 +546,82 @@ const commandsModule = ({
...
+        'SM',
+        'RTSTRUCT',
+        'RTPLAN',
+        'RTDOSE',
+      ];
+
+      const currentDisplaySets = displaySetService.activeDisplaySets;
+
+      // Sort the display sets as per the thumbnail/study browser list
+      utils.sortBySeriesDate(currentDisplaySets);
```
sedghi, line 566: so isn't the currentDisplaySets sorted by default? I'm confused why we need to sort again.

jbocce: I found that is was NOT sorted. Without this line, the navigation would jump to display sets in an order different than what appears in the thumbnail list.

platform/core/src/defaults/hotkeyBindings.js
```diff
@@ -76,18 +76,24 @@ const bindings = [
...
-  //   keys: ['pageup'],
-  //   isEditable: true,
-  // },
-  // {
-  //   commandName: 'previousViewportDisplaySet',
-  //   label: 'Previous Series',
-  //   keys: ['pagedown'],
-  //   isEditable: true,
-  // },
+  {
```
sedghi, line 79: can you please add it to all configs? thanks (I guess some have it already)

sedghi: ignore this

extensions/default/src/commandsModule.ts
```diff
@@ -541,6 +546,83 @@ const commandsModule = ({
...
+      const nonImageModalities = [
+        'SR',
+        'SEG',
+        'SM',
+        'RTSTRUCT',
+        'RTPLAN',
+        'RTDOSE',
+      ];
+
+      // Sort the display sets as per the hanging protocol service which is also consistent with the thumbnail list.
```
sedghi, line 563: can we add more comment about this is the same scoring system in HP

sedghi (commented):
Looks good, see my comments please

sedghi:
Having a long thumbnail and using page down and up should ideally scroll the thumbnail list too. Would it be possible for you to add the scroll to view for the thumbnail that is getting selected? it would be something like this

```
const scrollbar = document.querySelector('.ohif-scrollbar');

// element that is active = find by class active
scrollbar.scrollTo({
// that element
      behavior: 'smooth',
  });
```

jbocce:
> Having a long thumbnail and using page down and up should ideally scroll the thumbnail list too. Would it be possible for you to add the scroll to view for the thumbnail that is getting selected? it would be something like this
> 
> ```
> const scrollbar = document.querySelector('.ohif-scrollbar');
> 
> // element that is active = find by class active
> scrollbar.scrollTo({
> // that element
>       behavior: 'smooth',
>   });
> ```

Yes this makes sense. Note that we currently do not have a class on the active thumbnail and searching for 'ohif-scrollbar' does not cut it. Hopefully what I did is satisfactory. Let me know.

sedghi (approved):
Looks great thanks!

PR #3417, opened 2023-05-23
MPR: add viewport id to identify them in order to customize for Hairs tool

md-prog:
### Context

CrossHairs tool uses viewport id to set the reference colors for the control lines. (see `getReferenceLineColor()` definition of CrossHairs tool)
Currently viewport ids are automatically generated (usually having an increasing sequence number).

In this PR, it tries to set the IDs that will be persisted into the viewports, so that we can identify the viewport from the Hairs tool when customizing reference colors. (note this will not be the `viewport.viewportId` but `viewport.viewportOptions.id`)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/extensions/cornerstone/src/getHangingProtocolModule.ts
+++ b/extensions/cornerstone/src/getHangingProtocolModule.ts
@@ -89,6 +89,7 @@ const mpr: Types.HangingProtocol.Protocol = {
       viewports: [
         {
           viewportOptions: {
+            id: 'viewport-0', //!Important: This ID is used by Hairs tool to identify the viewport
             toolGroupId: 'mpr',
             viewportType: 'volume',
             orientation: 'axial',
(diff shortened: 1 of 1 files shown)


extensions/cornerstone/src/getHangingProtocolModule.ts
```diff
@@ -211,6 +214,7 @@ const mprAnd3DVolumeViewport = {
       viewports: [
         {
           viewportOptions: {
+            id: 'viewport-0', //!Important: This ID is used by Hairs tool to identify the viewport
```
sedghi, line 217: where is this id getting used? i'm pretty sure we don't have that logic in OHIF

PR #3422, opened 2023-05-25
fix/new demo data server

sedghi:
### Context & Changes

With the new demo server (static-wado) and addition of more modalities, more bugs appeared in how OHIF handle static wado. 

RT:
- better error handling 
- make the default server use bulkDataURI to fetch

SM:
- Metadata handled for consistensie in the imageType
- relative bulkDataURI is converted to the full

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


extensions/cornerstone-dicom-rt/src/loadRTStruct.js
```diff
@@ -55,42 +56,54 @@ async function checkAndLoadContourData(instance, datasource) {
...
-
-            ContourData.push(value);
-            startIndex = endIndex + 1;
-            endIndex = dataUint8Array.indexOf('\\', startIndex);
-            endIndex === -1 ? (endIndex = dataUint8Array.length) : endIndex;
-            numbersParsed++;
+  instance.ROIContourSequence.forEach(ROIContour => {
+    try {
+      ROIContour.ContourSequence.forEach((Contour, contourIndex) => {
+        const promise = resolvedPromises[index++];
```
wayfarer3130, line 62: Could this be changed to a map by contourIndex?  I'm just concerned that we might get out of sync on this and end up with the wrong contour somewhere.  Mapping the promises definitely rather than by position just feels safer.

extensions/cornerstone-dicom-rt/src/loadRTStruct.js
```diff
@@ -55,42 +56,54 @@ async function checkAndLoadContourData(instance, datasource) {
...
+          } else {
+            // Otherwise, proceed with existing logic
+            const uint8Array = new Uint8Array(promise.value);
+            const textDecoder = new TextDecoder();
+            const dataUint8Array = textDecoder.decode(uint8Array);
+            if (
+              typeof dataUint8Array === 'string' &&
+              dataUint8Array.includes('\\')
+            ) {
+              const numSlashes = (dataUint8Array.match(/\\/g) || []).length;
```
wayfarer3130, line 79: Could you write a custom function that counts them - doing this as a regular expression works, but the efficiency isn't very good, and some of these arrays can be huge.

wayfarer3130: Thinking about it, just try dataUint8Array.split(/\\/).map(parseFloat)  That uses a regular expression once, but then you get the entire array of strings as a result - and if it needs to be more efficient in the future, it is easy to implement an efficient version of it.

modes/tmtv/src/index.js
```diff
@@ -189,16 +189,24 @@ function modeFactory({ modeConfiguration }) {
...
-      return (
+      const isValid =
         modalities_list.includes('CT') &&
         modalities_list.includes('PT') &&
-        !invalidModalities.some(modality => modalities_list.includes(modality))
-      );
+        !invalidModalities.some(modality =>
+          modalities_list.includes(modality)
+        ) &&
+        // This is study is a 4D study with PT and CT and not a 3D study for the tmtv
```
wayfarer3130, line 202: Prefix with TODO: Find a better way to identify 4D studies

platform/viewer/src/routes/WorkList/WorkList.tsx
```diff
@@ -343,6 +343,7 @@ function WorkList({
 
             const isValidMode = mode.isValidMode({
               modalities: modalitiesToCheck,
+              StudyInstanceUID: studyInstanceUid,
```
wayfarer3130, line 346: Can you just pass the ...study here?

wayfarer3130 (changes requested):
Hopefully not to hard to update

extensions/cornerstone-dicom-rt/src/loadRTStruct.js
```diff
@@ -7,12 +7,13 @@ async function checkAndLoadContourData(instance, datasource) {
     return Promise.reject('Invalid instance object or ROIContourSequence');
   }
 
-  const promises = [];
-  let counter = 0;
+  const promisesMap = new Map();
```
wayfarer3130, line 10: Thank you.

extensions/default/src/utils/getDirectURL.js
```diff
@@ -57,31 +51,19 @@ const getDirectURL = (config, params) => {
...
-    return `${wadoRoot}/studies/${StudyInstanceUID}/${acceptUri}`;
+
+  if (tag === 'PixelData' || tag === 'EncapsulatedDocument') {
+    return `${wadoRoot}/studies/${StudyInstanceUID}/series/${SeriesInstanceUID}/instances/${SOPInstanceUID}/rendered`;
   }
-  throw new Error('BulkDataURI in unknown format:' + BulkDataURI);
+
+  // since we fixed the bulkDataURI to be absolute (in case of the relative),
+  // we can just return it here
+  return BulkDataURI;
```
wayfarer3130, line 66: This should be the accept URI, because the DICOMweb standard states that the default is multipart related, and then separately states that the accept parameter is the URL parameter equivalent of the accept header.

extensions/dicom-microscopy/src/utils/cleanDenaturalizedDataset.ts
```diff
@@ -38,7 +44,15 @@ export default function cleanDenaturalizedDataset(obj: any): any {
         delete obj[key].Value;
       } else if (Array.isArray(obj[key].Value) && obj[key].vr) {
         if (obj[key].Value.length === 1 && obj[key].Value[0].BulkDataURI) {
-          obj[key].BulkDataURI = obj[key].Value[0].BulkDataURI;
+          // If the BulkDataURI is not a study level URI, then we need to convert the
+          // relative URI to an absolute URI (see open PR https://github.com/dcmjs-org/dicomweb-client/pull/54)
+          if (!obj[key].Value[0].BulkDataURI.includes('studies/')) {
```
wayfarer3130, line 49: You need to call fixBulkDataURI here - the test here as you have it isn't safe, because a perfectly value bulkdata URI is:
https://server/bulkdata/hashcode.dat
There is NO requirement at all that the bulkdata URI contain the studies path.  IF it is a path relative URI, then it will be relative to the metadata call that was used to retrieve it, but nothing says it has to be in the dicomweb path even.

wayfarer3130 (approved):
Couple of small changes - marking approved as I may not be able to easily get back to it today, and there are really just the two issues.

wayfarer3130 (approved):

Issue #3361, opened 2023-05-03
v3: Add basic GCP url support 


igoroctaviano:
### Request
- Add support to retrieve study/series from GCP using dicomweb:
http://localhost:3000/projects/your-project/locations/us/datasets/your-dataset/dicomStores/test-store/study/123

sanjay3290:
Hey Guys, any update or workaround for this?

igoroctaviano:
> Hey Guys, any update or workaround for this?

@sanjay3290 With this PR you can accomplish this: https://github.com/OHIF/Viewers/pull/3543

igoroctaviano:
Now with https://github.com/OHIF/Viewers/pull/3543 merged the first part of this issue is done. We can support gcp URLs (study level) using this config:
```
{
    friendlyName: 'GCP DICOMWeb Server',
    namespace: '@ohif/extension-default.dataSourcesModule.dicomweb',
    sourceName: 'gcpdicomweb',
    configuration: {
      name: 'gcpdicomweb',
      qidoSupportsIncludeField: false,
      imageRendering: 'wadors',
      thumbnailRendering: 'wadors',
      enableStudyLazyLoad: true,
      supportsFuzzyMatching: false,
      supportsWildcard: false,
      singlepart: 'bulkdata,video,pdf',
      useBulkDataURI: false,
      onConfiguration: (dicomWebConfig, options) => {
        const { params } = options;
        const { project, location, dataset, dicomStore } = params;
        const pathUrl = `https://healthcare.googleapis.com/v1/projects/${project}/locations/${location}/datasets/${dataset}/dicomStores/${dicomStore}/dicomWeb`;
        return {
          ...dicomWebConfig,
          wadoRoot: pathUrl,
          qidoRoot: pathUrl,
          wadoUri: pathUrl,
          wadoUriRoot: pathUrl,
        };
      },
    },
  },
```

sanjay3290:
how can i send dynamic params for the below method during runtime.? or it can be set only during the build time?

onConfiguration: (dicomWebConfig, options) => {
          const { params } = options;
          const { project, location, dataset, dicomStore } = params;
          const pathUrl = `https://healthcare.googleapis.com/v1/projects/${project}/locations/${location}/datasets/${dataset}/dicomStores/${dicomStore}/dicomWeb`;
          return {
            ...dicomWebConfig,
            wadoRoot: pathUrl,
            qidoRoot: pathUrl,
            wadoUri: pathUrl,
[...]

PR #3427, opened 2023-05-30
DicomWebDataSource: Add support for dynamic config values via config hook

igoroctaviano:
- #3361 
- Expose `showStudyList` as a layout option: Disable rendering of the return button per layout config (valid for data sources not supporting the worklist)
- `noWorklistButton` mode option: Disable rendering of the mode button in the expanded worklist item
- New DicomWebDataSource`onConfiguration` option: Currently only available for DicomWebDataSource. This option allows interception of the data source configuration for dynamic values or configuration options dependent on URL params or query params e.g.:
```
  {
    friendlyName: 'GCP DICOMWeb Server',
    sourceName: 'gcpdicomweb',
    configuration: {
      name: 'gcpdicomweb',
      imageRendering: 'wadors',
      thumbnailRendering: 'wadors',
      enableStudyLazyLoad: true,
      supportsFuzzyMatching: false,
      supportsWildcard: false,
      singlepart: 'bulkdata,video,pdf',
      useBulkDataURI: false,
      onConfiguration: (dicomWebConfig, options) => {
        const { params } = options;
        const { project, location, dataset, dicomStore } = params;
        const pathUrl = `https://healthcare.googleapis.com/v1/projects/${project}/locations/${location}/datasets/${dataset}/dicomStores/${dicomStore}/dicomWeb`;
        return {
          ...dicomWebConfig,
          wadoRoot: pathUrl,
          qidoRoot: pathUrl,
          wadoUri: pathUrl,
          wadoUriRoot: pathUrl,
        };
      },
    },
  },
```


platform/app/src/routes/WorkList/WorkList.tsx
```diff
@@ -197,6 +200,7 @@ function WorkList({
   useEffect(() => {
     const fetchSeries = async studyInstanceUid => {
       try {
+        await dataSource.initialize({ params, query });
```
sedghi, line 203: This doesn't resonate with me, we are initializing the data source 3 times after this PR (used to be one time in the mode.tsx). 

Should we call it something else?
Should we skip it if already initialized?

sedghi (commented):
please see my comments

platform/app/src/routes/WorkList/WorkList.tsx
```diff
@@ -341,6 +345,10 @@ function WorkList({
 
             const modalitiesToCheck = modalities.replaceAll('/', '\\');
 
+            if (mode.noWorklistButton === true) {
```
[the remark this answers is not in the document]
igoroctaviano: - noWorklistButton is to disable rendering a mode button here if I want to have a hidden mode or something like that:

Screenshot 2023-06-13 at 13 49 03
- isReturnEnabled was already there I just exposed it so people can disable the return to worklist button per mode instead of a global config option (current implementation)

platform/app/src/routes/WorkList/WorkList.tsx
igoroctaviano: Good call. I added a logic to skip if its already initialized. I added this code here so we want initialization before loading things.

platform/app/src/routes/WorkList/WorkList.tsx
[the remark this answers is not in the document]
jbocce: Please add `noWorklistButon` to the table of mode properties found in this [section](https://docs.ohif.org/platform/modes/#anatomy) of our docs. However, I am not sure if this is any different from simply not including the mode in the config file from the get-go. What is the use case for this? Would it be to include the mode (for URL viewer launch), but exclude it in the worklist?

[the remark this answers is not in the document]
jbocce: Also consider where to document `isReturnEnabled`. Maybe in the docs for [Layout template](https://docs.ohif.org/platform/extensions/modules/layout-template#overview)? We do mention the default extension ViewerLayout there, but I am not sure that is the best place. If we are adding it then we should have people know about it. Just a suggestion. :)

platform/docs/docs/configuration/configurationFiles.md
```diff
@@ -134,8 +134,37 @@ Example 2, to restricts to either hosptial.com or othersite.com.<br/>
 `regex: /(https:\/\/hospital.com(\/[0-9A-Za-z.]+)*)|(https:\/\/othersite.com(\/[0-9A-Za-z.]+)*)/` <br/>
 Example usage:<br/>
 `http://localhost:3000/?configUrl=http://localhost:3000/config/example.json`<br/>
-
-
+- `onConfiguration`: Currently only available for DicomWebDataSource. This option allows interception of the data source configuration for dynamic values or configuration options that are dependent of url params or query params. Example:
```
jbocce, line 137: Consider fleshing out this description to include more detail about what the particular example below is doing.

platform/app/src/routes/WorkList/WorkList.tsx
jbocce: I too am concerned with having multiple initialize calls. One thing that concerns be the most is that prior to this PR, one could use some of the methods without calling `initialize` first. In particular, `getConfig` is one such method and I regret to let you know that now upload is broken with this PR. So in the future I can see more "places" where initialize has to be called. Is there a more central/early location to call initialize? I feel bad suggesting there is a problem without an actual suggestion of how to fix it, but I am open to discussing it. Thoughts?

image.png

platform/app/src/routes/WorkList/WorkList.tsx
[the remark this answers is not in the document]
igoroctaviano: @jbocce Yes, the idea is to have a mode that does not render a button from a worklist e.g. a URL only mode. This way I don't have to build an URL for gcp from a worklist.

jbocce (commented):
A few comments to address. Thank you so much.

platform/app/src/routes/WorkList/WorkList.tsx
igoroctaviano: > I too am concerned with having multiple initialize calls. One thing that concerns be the most is that prior to this PR, one could use some of the methods without calling `initialize` first. In particular, `getConfig` is one such method and I regret to let you know that now upload is broken with this PR. So in the future I can see more "places" where initialize has to be called. Is there a more central/early location to call initialize? I feel bad suggesting there is a problem without an actual suggestion of how to fix it, but I am open to discussing it. Thoughts?

Maybe we should place the initialize somewhere else. Currently it doest work like I would expect, if I go to the worklist, another code from the dicomweb datasource executes before initialize is called. I'll revisit the PR and see where I place it so we don't have to call it twice.

jbocce: image

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -51,47 +51,58 @@ const metadataProvider = classes.MetadataProvider;
...
 
-  const wadoConfig = {
-    url: wadoRoot,
-    staticWado,
-    singlepart,
-    headers: userAuthenticationService.getAuthorizationHeader(),
-    errorInterceptor: errorHandler.getHTTPErrorHandler(),
-  };
+  const implementation = {
+    initialize: ({ params, query }) => {
```
sedghi, line 62: This method should be able to grab a new config too to override the old one

igoroctaviano: Updated the code so the initialize is called once as it should be.

platform/docs/docs/configuration/configurationFiles.md
igoroctaviano: Updated! Thanks.

platform/app/src/routes/WorkList/WorkList.tsx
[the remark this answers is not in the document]
igoroctaviano: Updated, thanks!

platform/app/src/routes/buildModeRoutes.tsx
```diff
@@ -45,21 +46,23 @@ export default function buildModeRoutes({
       const path = `/${mode.routeName}/${dataSourceName}`;
 
       // TODO move up.
```
jbocce, line 48: To avoid duplicate code on line 75, please consider performing this TODO. I respect it if you were to decline, but I figure I would mention it.

platform/app/src/routes/buildModeRoutes.tsx
```diff
@@ -45,21 +46,23 @@ export default function buildModeRoutes({
...
-      const children = () => (
+      const children = (...props) => (
```
jbocce, line 49: Correct me if I am wrong, but the `ModeRoute` at present really does not care about `props`, so consider not including them here. Or am I missing something?

extensions/default/src/ViewerLayout/index.tsx
```diff
@@ -285,6 +286,7 @@ ViewerLayout.propTypes = {
   commandsManager: PropTypes.instanceOf(CommandsManager),
   servicesManager: PropTypes.instanceOf(ServicesManager),
   // From modes
+  isReturnEnabled: PropTypes.bool,
```
jbocce, line 289: Um, shouldn't this be `showStudyList` instead?

platform/core/src/DataSources/IWebApiDataSource.js
```diff
@@ -73,6 +74,7 @@ function create({
     getImageIdsForDisplaySet,
     getImageIdsForInstance,
     getConfig: getConfig || defaultGetConfig,
+    getStudyInstanceUIDs,
```
jbocce, line 77: Consider adding a default implementation here.

platform/docs/docs/platform/modes/index.md
```diff
@@ -112,6 +113,12 @@ export default mode;
       </td>
       <td align="left">actual name of the mode being displayed for each study in the study summary panel</td>
     </tr>
+    <tr>
+      <td align="left">
+          noWorklistButton
+      </td>
+      <td align="left">wether or not to display the mode button in the expanded worklist item</td>
```
jbocce, line 120: typo: wh*ether

platform/app/src/routes/WorkList/WorkList.tsx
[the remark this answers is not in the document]
jbocce: Thanks for updating the docs. However, I noticed that upload is still broken. But the good news is that I think it can be simply fixed by adding a null check on line 453 as such...

```
dicomUploadComponent && dataSource.getConfig()?.dicomUploadEnabled
```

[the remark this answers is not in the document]
igoroctaviano: Sorry, what do you mean by broken? i couldn't see anything in the logs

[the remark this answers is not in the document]
jbocce: As I briefly mentioned in my first review [here](https://github.com/OHIF/Viewers/pull/3427#discussion_r1231449990), if you try to run the `config/local_orthanc.js` or `config/local_dcm4chee.js` you will notice that OHIF does not even run at all.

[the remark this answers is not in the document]
jbocce: Both of those data source configurations have dicom upload enabled.

extensions/default/src/DicomWebDataSource/index.js
jbocce: I think @sedghi's suggestion above still needs to be addressed, although I am not sure if this PR is necessarily the best place to address it. The suggestion stems from a community request for being able to dynamically change/update the data source configuration via a `postMessage` from a parent/ancestor when OHIF is embedded in an iframe. I think it is going to be tricky because the `postMessage` could arrive at any arbitrary time.

jbocce (commented):
A few more comments to address. Thank you so much.

platform/docs/docs/platform/modes/index.md
igoroctaviano: Good catch, updated, thanks!

platform/core/src/DataSources/IWebApiDataSource.js
igoroctaviano: Added, thanks!

extensions/default/src/ViewerLayout/index.tsx
igoroctaviano: Yes, good catch, thank you.

platform/app/src/routes/buildModeRoutes.tsx
igoroctaviano: Updated! thanks :)

platform/app/src/routes/buildModeRoutes.tsx
igoroctaviano: Updated, thanks!

extensions/default/src/DicomWebDataSource/index.js
igoroctaviano: Added the option to override the config 👍 One could call initialize again using postMessage.

jbocce: Thanks for this. I suppose if ever `initialize` is called with a `config`, it would be up to the caller to do "something" so that the new config is applied "everywhere" that the former had already been applied to... perhaps even a page/router refresh... but you don't have to worry about that... I just got lost in thought... lol. Thanks again.

jbocce (changes requested):
Thanks for all the updates. I have done a bit more testing and noticed a couple of problems...

- the end-to-end tests are failing; please have a look
- the `/local` and `/localBasic` routes no longer appear to work; I suspect that `DicomLocalDataSource` might need some tweaks; I suspect some of the other data sources might too, in particular `DicomJSONDataSource`; please have a look

extensions/default/src/DicomWebDataSource/index.js
anarnoli: Actually I was struggling with the `postMessage` implementation for override the config with new one. Can you please suggest step to override config using this PR?

jbocce: @anarnoli , I know we have spoken outside of this PR, but I just wanted to record what we spoke about here for completeness...

- In the `preRegistration` method of your extension you should be able to grab an instance of `extensionManager `.
- `extensionManager` has a `getDataSources` method.
- Loop through the data sources returned, looking for the data source you want to update the configuration for.
- Once you have the data source to (re)configure, call the new `initialize` method (of this PR) on it with the new configuration.
- Then the question I have is that the above process can happen at ANY point in time. For example, a study might already have been loaded and then the `postMessage` arrives. We need to add some thought about how to apply the new configuration. One way might be to use the React router to simply refresh the page. Not sure that will work, but worth a shot.

extensions/default/src/DicomJSONDataSource/index.js
```diff
@@ -41,10 +42,10 @@ const findStudies = (key, value) => {
 };
 
 function createDicomJSONApi(dicomJsonConfig) {
-  const { name, wadoRoot } = dicomJsonConfig;
+  const { wadoRoot } = dicomJsonConfig;
 
   const implementation = {
-    initialize: async ({ params, query, url }) => {
+    initialize: async ({ query, url }) => {
```
jbocce, line 48: Do we need/want to handle the `config` option here like we do in `DicomWebDataSource`?

jbocce (changes requested):
Thanks for the updates.

Thank you for fixing `/local`, I tested it and it looks good.

The `end-to-end` tests are still failing, please have a look.

I also tried your branch with the local DICOM json demo (https://docs.ohif.org/configuration/datasources/dicom-json/#local-demo) and it had the following error and it did not display the study. I tried master and it works fine.

image

platform/app/src/routes/index.tsx
```diff
@@ -15,15 +15,24 @@ import PrivateRoute from './PrivateRoute';
 const bakedInRoutes = [
   {
     path: '/debug',
-    children: Debug,
+    children: DataSourceWrapper,
+    props: {
```
sedghi, line 19: can we have some doc about who is child of whom since there are two children props here, thanks

igoroctaviano: What do you mean by two children? I'm just configuring the wrapper to be the parent of any route since any route could retrieve or access data from a data source, that's all

igoroctaviano:
> Thanks for the updates.
> 
> Thank you for fixing `/local`, I tested it and it looks good.
> 
> The `end-to-end` tests are still failing, please have a look.
> 
> I also tried your branch with the local DICOM json demo (https://docs.ohif.org/configuration/datasources/dicom-json/#local-demo) and it had the following error and it did not display the study. I tried master and it works fine.
> 
> image

Thanks, I'll do that

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -522,10 +522,26 @@ function createDicomWebApi(dicomWebConfig, userAuthenticationService) {
     getConfig() {
       return dicomWebConfigCopy;
     },
+    getStudyInstanceUIDs({ params, query }) {
```
sedghi, line 525: we should check this PR if having multiple studyUIDs separated with comma work, this one 

https://docs.ohif.org/configuration/url#studyinstanceuids

platform/app/src/routes/index.tsx
sedghi: I know what you are doing, but for someone who reads the code it might not be evident. So some docs would be good since there are two children props

sedghi:
Apparently now that we have wrapped the mode with datasource there are multiple unnecessary xhr for the study list 

image

platform/app/src/routes/index.tsx
igoroctaviano: I'm still not sure what you mean by two children, I can only see one for the route and another for the children. I'm also not sure yet why we should explain such technical detail about having a wrapper for the routes. If the data source wrapper is already documented I think that's all people need to know? Again, I'm open to suggestions on where I could add this information

igoroctaviano:
> @igoroctaviano Apparently now that we have wrapped the mode with datasource there are multiple unnecessary xhr for the study list
> 
> https://deploy-preview-3427--ohif-dev.netlify.app/viewer?StudyInstanceUIDs=2.16.840.1.114362.1.11972228.22789312658.616067305.306.2
> 
> image

Thanks for cathing this. I'll step back with this implementation and think about doing something else. I think the wrapper should be a root wrapper (every route should automatically be a child of it). Looks like there's too much stuff going on under the hood, maybe the wrapper was not meant to be why I'm using it for.

jbocce:
> Thanks for cathing this. I'll step back with this implementation and think about doing something else. I think the wrapper should be a root wrapper (every route should automatically be a child of it). Looks like there's too much stuff going on under the hood, maybe the wrapper was not meant to be used like I did.

@igoroctaviano 

I did not think of mentioning this earlier, but I have some code that is destined to be used for the Google Cloud API data source whereby a datasource can be added dynamically (i.e. A data source's configuration is created on the fly and then it is added as a new/separate data source). This is similar to what this PR is trying to do except that instead of altering an existing data source, a new one is created. One could think of editing an existing as simply creating a new one. Furthermore, the code I have allows for the active data source to be set/changed on the fly.

Thoughts? Would that satisfy what this PR is trying to do in essence?

igoroctaviano:
> > Thanks for cathing this. I'll step back with this implementation and think about doing something else. I think the wrapper should be a root wrapper (every route should automatically be a child of it). Looks like there's too much stuff going on under the hood, maybe the wrapper was not meant to be used like I did.
> 
> @igoroctaviano
> 
> I did not think of mentioning this earlier, but I have some code that is destined to be used for the Google Cloud API data source whereby a datasource can be added dynamically (i.e. A data source's configuration is created on the fly and then it is added as a new/separate data source). This is similar to what this PR is trying to do except that instead of altering an existing data source, a new one is created. One could think of editing an existing as simply creating a new one. Furthermore, the code I have allows for the active data source to be set/changed on the fly.
> 
> Thoughts? Would that satisfy what this PR is trying to do in essence?

I do have a new data source specifically for GCP in another branch and it works fine. I thought of having this solution here in the open source because I thought that maybe other people might leverage this dynamic config for dicomweb. If you guys think its better to just have people creating their own custom data source I'm cool closing this PR 👍

sedghi:
Do we need to close this after the recent merge by @jbocce

PR #3473, opened 2023-06-13
fix(Button): redesign of existing Button component

jbocce:
### Context

New Button component designs thanks to @dan-rukas (see the screen shots below). The redesign improves various items:
1. A consistent look-and-feel with the rest of OHIF with better button styles, sizes and colours

2. A reduction in complexity for developers using the `Button`. The previous `Button` component allowed for many (too many) variant, colour and border permutations, making for a confusing developer experience. Furthermore, with so many permutations, maintaining a consistent button look-and-feel proved to be very difficult. Now there are four basic permutations:
```
<Button type={ButtonEnums.type.primary} size={ButtonEnums.size.medium}/>
<Button type={ButtonEnums.type.secondary} size={ButtonEnums.size.medium}/>
<Button type={ButtonEnums.type.primary} size={ButtonEnums.size.small}/>
<Button type={ButtonEnums.type.secondary} size={ButtonEnums.size.small}/>
```

3. A disabled button is no longer a variant or type, but instead just a `boolean` attribute of the `Button` like so...
```
<Button disabled={true}}/>
```

image

image

image

image

### Changes & Results

- Copied ui/Button to ui/LegacyButton and replaced Button with LegacyButton in ButtonGroup components.
- Altered the `Button` component to match the new styles.
- Updated various uses of the `Button` component.
- Uses of  `Button` component that did NOT match the new design were changed to `LegacyButton`.
- `LegacyButton` uses will be changed/addressed in future redesigns.
- Updated the `Button` Storybook.

### Testing

All the various uses of `Button` should reflect the new design. Those that do NOT include...
- `ButtonGroup` buttons
image
- various Worklist buttons like those for launching modes and clearing filters

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 
- [x] Node version:  16.14.0 
- [x] Browser: Chrome  114.0.5735.110
  


extensions/default/src/Panels/ActionButtons.tsx
```diff
@@ -2,20 +2,23 @@ import React from 'react';
...
+import { LegacyButton, ButtonGroup } from '@ohif/ui';
 
 function ActionButtons({ onExportClick, onCreateReportClick }) {
   const { t } = useTranslation('MeasurementTable');
 
   return (
     <React.Fragment>
       <ButtonGroup color="black" size="inherit">
-        <Button className="px-2 py-2 text-base" onClick={onExportClick}>
+        <LegacyButton className="px-2 py-2 text-base" onClick={onExportClick}>
```
sedghi, line 13: maybe add a todo that we will revisit the button group later and that is why we are using the legacyButton here

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/ActionButtons.tsx
```diff
@@ -2,31 +2,26 @@ import React from 'react';
...
       <Button
-        className="text-base px-2 py-2"
-        size="initial"
-        variant={disabled ? 'disabled' : 'outlined'}
-        color="black"
-        border="primaryActive"
         onClick={onExportClick}
         disabled={disabled}
+        type={ButtonType.secondary}
+        size={ButtonSize.small}
```
sedghi, line 16: Hmmm, I was fine with the ButtonType to be exported, until here that I see ButtonSize is also another export. Could we group them all under `ButtonStyle` and then we have `ButtonStyle.type.secondary`, and `ButtonStyle.size.small`. Or anyother name that make it a group. I'm saying this since we will probably have this pattern of a UI component with its style Enums. Maybe even ButtonEnums

sedghi (commented):
Looks great, see my comments and maybe update the PR before the first image to explain that this is done to reduce the complexity of variants and colors and etc etc, and maybe just mention the 4 different syntaxes they can use now (instead of who knows how many before)

jbocce:
> Looks great, see my comments and maybe update the PR before the first image to explain that this is done to reduce the complexity of variants and colors and etc etc, and maybe just mention the 4 different syntaxes they can use now (instead of who knows how many before)

Thanks. I updated the PR overview.

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking/ActionButtons.tsx
jbocce: Makes sense. I made changes... hopefully they are what you meant.

sedghi (approved):
Looks great, thanks for taking care of this

Issue #3360, opened 2023-05-03
[Feature Request] Mechanism to communicate to the users inconsistencies within a single series


igoroctaviano:
### Request
Alert icon (lower right corner under the thumbnail in v2) indicating series that have geometry issues, and explanation of what those are, is missing; e.g., see series 2 for this study

### Example series that motivate some discussion what should be done:
*  https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc_tcia_dev/dicomStores/idc_tcia/study/1.3.6.1.4.1.14519.5.2.1.8162.4009.498881901170332065029863238265 - first instance is a localizer
* https://idc-sandbox-000.firebaseapp.com/projects/idc-dev-etl/locations/us-central1/datasets/idc_tcia_dev/dicomStores/idc_tcia/study/1.3.6.1.4.1.14519.5.2.1.8162.4009.498881901170332065029863238265?seriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.8162.4009.329297441360267384643622758743-  one series has 2 subvolumes with different slice and pixel spacing

At the moment, OHIF Viewer does not have any indication about those inconsistencies, and presents the slices as if they belong to a single continuous volume.

![image](https://user-images.githubusercontent.com/313942/106286951-39970a80-6214-11eb-893a-d83ccc7272d5.png)

### Example in v2
![image](https://user-images.githubusercontent.com/313942/113185008-e6e1cb00-9223-11eb-9c4f-89cd360af456.png)
![image](https://user-images.githubusercontent.com/313942/113185067-fc56f500-9223-11eb-8c3d-c028b3557612.png)

fedorov:
Per discussion today, @jlulloaa will investigate what, if any, are the geometry checks implemented in v3, where in the code they are located, how can they be extended, etc ...

Related to this, here is an interesting edge case from IDC where frames within the same series do not have consistent origin: https://viewer.imaging.datacommons.cancer.gov/viewer/1.3.6.1.4.1.14519.5.2.1.7009.9004.229337801792248347333415175865?seriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.7009.9004.172787163177205374234578117444

From the IDC perspective, it would also be highly desirable to be able to see image series that have geometry issues in 2D. If this is not possible, we will be in a situation where (potentially) many images from IDC that were viewable in v2 will no longer be viewable in v3!

jlulloaa:
[...]

PR #3499, opened 2023-06-27
feat(Thumbnail): Display set messages support & displaying of series inconsistencies in the thumbnail

rodrigobasilio2022:
When the user has studies with series supposed to be reconstructable but they arent, at first its hard to identify them and know why a particular series cant be viewed in MPR mode. This PR proposes a warning button in thumbnail list that highlights problems in series and help users know geometry issues in the series. It is related to issue IDC #3360 

Below I added some studies from the IDC database that has series with different problems

- Study with a series that change position during scan
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.9004.229337801792248347333415175865

- Study with a series with irregular pixel spacing
http://localhost:3000/viewer?StudyInstanceUIDs=1.2.840.113654.2.55.69925400202012637024302712643359784845

- Studies has at least a series that can be viewed in MPR mode, despite the problems
http://localhost:3000/viewer?StudyInstanceUIDs=1.2.840.113654.2.55.268978779666003033721491501200337886584 (S. 3)
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.3098.5025.250976478345829511981090188621 (S. 5)

- Study has a serie with different orientations that wrongly can be reconstructed in current OHIF version. My fixes prevent that.
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.55696427943949105811475330286060 (S. 601)

- Studies with series with MIP images and which the first image appears to be an localizer for the series in another orientation
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2406.131628405171680429117301556825
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.16051849653578423672035750287473

- Study with series composed by two volumes of data 
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.16051849653578423672035750287473 (S. 2)


extensions/default/src/validateInstances.js
```diff
@@ -0,0 +1,228 @@
...
+    checkSingleFrame(sortedInstances, warnings);
+  }
+
+  if (!isReconstructable) {
+    warnings.push('Series is not a reconstructable 3D volume.');
+  }
+  return warnings;
+}
+
+function checkMultiFrame(multiFrameInstance, warnings) {
```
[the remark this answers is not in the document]
rodrigobasilio2022: All the functions called in checkMultiframe are already in the core. I am just using them. The other functions a broke in smaller parts to isolate the check and possible be used in core

extensions/default/src/validateInstances.js
```diff
@@ -0,0 +1,228 @@
...
+  const isMultiframe = firstInstance.NumberOfFrames > 1;
+  const Modality = firstInstance.Modality;
+  if (!constructableModalities.includes(Modality)) {
+    return warnings;
+  }
+
+  // Can't reconstruct if all instances don't have the ImagePositionPatient.
+  if (
+    !isMultiframe &&
+    !instances.every(instance => instance.ImagePositionPatient)
```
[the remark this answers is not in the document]
rodrigobasilio2022: I started the process

extensions/default/src/validateInstances.js
```diff
@@ -0,0 +1,228 @@
...
+  hasOrientation,
+  hasPosition,
+  _getPerpendicularDistance,
+  _getSpacingIssue,
+  reconstructionIssues,
+  _isSameOrientation,
+  constructableModalities,
+} from '@ohif/core/src/utils/isDisplaySetReconstructable';
+
+function addWarning(warnings, warning) {
```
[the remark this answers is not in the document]
rodrigobasilio2022: I agree that this function shouldnt be here, but where is the ebst place to put it? @ohif/core/src/utils? Right now warnings is a list of strings. It could be renamed to issues, that could be warnings or errors, and it could be a list of objects

jbocce (commented):
See my suggestions and comments.

extensions/default/src/validateInstances.js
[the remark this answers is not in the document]
jbocce: So I do not know what the requirements are. But if only one extension (like you currently have) is creating the warnings and adding them to each display set, then the process you have defined for adding warnings is fine.

Adding a code for each warning so that it can be translated is a good idea too. The warning/message code is not only useful for translation, but also for maybe in the future for a button/tool to decide if it should be disabled or not based on the code. In my opinion, instead of the message deciding what it applies to, the UI/area decides based on the code how it interprets the message code.

If indeed it is a requirement for any extension/anywhere to add a warning then right now, the only way I can think of allowing extensions (and other places) to add warnings is maybe via the `DisplaySetService` (e.g. maybe `DisplaySetService.addDisplaySetMessage`). The reasons I say `DisplaySetService` are:
- a message/warning is tied to a display set
- if anywhere in the code (e.g. an extension) can add a warning then a notification (e.g. via `DisplaySetService._broadcastEvent(EVENTS.DISPLAY_SET_MESSAGE_ADDED)`) should be sent to the listeners (e.g. `PanelStudyBrowser`) that might care about those messages and react accordingly

Again I am not the expert and I might be totally wrong. These are my humble suggestions. 😊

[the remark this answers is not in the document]
rodrigobasilio2022: I created two objects in the ohif core: displaySetMessage object that has three attributes: id (number, represents which message), displayLocation (number, where to put the message), hangingProtocolId (string, name of the protocol id to be used in the future). I also created displaySetMessageList, that handles a list of displaySetMessages. 

I think these two objects are sufficient for now. I prefer not to commit to creating a method in displaySetService for now until we better understand what we want to do.

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.tsx
```diff
@@ -477,6 +477,7 @@ function _mapDisplaySets(
         seriesDate: formatDate(ds.SeriesDate),
         numInstances: ds.numImageFrames,
         countIcon: ds.countIcon,
+        warnings: ds.warnings,
```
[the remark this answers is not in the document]
rodrigobasilio2022: Changed the name to messages. Each message object now has an attribute, displayLocation, that indicates where to put it (THUMBNAIL, CONSOLE, SNACKBAR_NOTIFICATION). To display in thumbnails, it filters the message list to display only thumbnail messages.

platform/app/public/config/default.js
```diff
@@ -40,33 +40,19 @@ window.config = {
...
+        name: 'IDC',
 
         // new server
-        wadoUriRoot: 'https://d33do7qe4w26qo.cloudfront.net/dicomweb',
-        qidoRoot: 'https://d33do7qe4w26qo.cloudfront.net/dicomweb',
-        wadoRoot: 'https://d33do7qe4w26qo.cloudfront.net/dicomweb',
-        qidoSupportsIncludeField: false,
-        supportsReject: false,
+        wadoUriRoot:
+          'https://testing-proxy.canceridc.dev/current/viewer-only-no-downloads-see-tinyurl-dot-com-slash-3j3d9jyp/dicomWeb',
```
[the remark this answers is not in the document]
rodrigobasilio2022: Yes. It is just for test purposes. But before merge I will restore this file

platform/core/src/utils/formatPN.js
```diff
@@ -6,7 +6,13 @@ export default function formatPN(name) {
     return;
   }
 
-  const nameToUse = name.Alphabetic ?? name;
+  let nameToUse;
+  const _nameToUse = name.Alphabetic ?? name;
+  if (_nameToUse instanceof String) {
```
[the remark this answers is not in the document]
rodrigobasilio2022: Ok. I just changed that because for some reason there is an entry in the IDC dicomweb server that returns an object ( {} ) in the name and prevents OHIF to display the study list

platform/ui/src/components/ThumbnailNoImage/ThumbnailNoImage.tsx
```diff
@@ -56,6 +58,23 @@ const ThumbnailNoImage = ({
               </div>
             </Tooltip>
             <span className="ml-4 text-base text-blue-300">{seriesDate}</span>
+            {warnings?.length && (
```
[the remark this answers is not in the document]
rodrigobasilio2022: Now warnings are converted to messages objects that have a method called messages.thumbnailContents() that returns the same contents that getWarningsContent  was returning.

platform/ui/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -78,6 +81,23 @@ const Thumbnail = ({
             <Icon name={countIcon || 'group-layers'} className="w-3 mr-2" />
             {` ${numInstances}`}
           </div>
+          {warnings?.length ? (
```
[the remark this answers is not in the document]
rodrigobasilio2022: Already commented in there

extensions/default/src/validateInstances.js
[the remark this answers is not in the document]
jbocce: Thanks @rodrigobasilio2022.

I will not have time to further review the code until Tuesday.

However, I still believe that a `Message` object should NOT have a `displayLocation` and should NOT have a `hangingProtocolId`. Extensions, modes and UIs should decide themselves which messages are applicable and where to show the messages. Take the following use case...

Use Case A
A hospital adds a their own OHIF mode and wants to display a certain message (id) in a side panel. In the proposed design, would they have to alter the existing/default extension where the message is created and the display location is set to change the location? That feels awkward to me. Sites using OHIF would typically not want to edit the core library code or default extensions to make their own customizations.

Likewise goes for the case where a hospital adds its own extension and hanging protocol. Would they have to edit the message in the default extension or OHIF core to apply or disallow the hanging protocol?

I might be wrong here, but we should think about this.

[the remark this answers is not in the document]
rodrigobasilio2022: But where to display and which hangingProtocolId to be used, if any,  is a decision of the extension. Right now the extension can only create messages for thumbnails, with no hangingprotocolIds but that is a first version because right now we don't have any demands for location and hanging protocol just for messages in thumbnails. I can remove the fields too

extensions/default/src/validateInstances.ts
```diff
@@ -0,0 +1,319 @@
...
+/**
+ * Checks if a series is reconstructable to a 3D volume.
+ *
+ * @param {Object[]} instances An array of `OHIFInstanceMetadata` objects.
+ */
+export default function validateInstances(
+  instances: Array<any>,
+  isReconstructable: boolean
+): displayServiceMessageList {
+  const messages = new displayServiceMessageList();
```
igoroctaviano, line 44: I don't feel like we should instantiate a new service class from a validation method. If you decided to create a new service to handle messages it should have an instance e.g. ```platform/core/src/services/UINotificationService/index.ts``` and be instantiated and shared along with other OHIF services across the codebase.

Also, the service names should be Uppercased and have simple naming .e.g MessageService. That's why I thought about creating a new manager just for validation in my own sketch (ValidationManager) but it could also be MessageManager or something alike.

extensions/default/src/validateInstances.ts
```diff
@@ -0,0 +1,319 @@
...
+ * @returns
+ */
+function checkSeriesDimensions(instances) {
+  const firstImage = instances[0];
+  const firstImageRows = toNumber(firstImage.Rows);
+  const firstImageColumns = toNumber(firstImage.Columns);
+
+  for (let i = 1; i < instances.length; i++) {
+    const instance = instances[i];
+    const { Rows, Columns } = instance;
```
igoroctaviano, line 148: I think all these checks should be implemented as separate files and unit tested.

igoroctaviano (changes requested):
Reviewed. Left some comments.
👍

extensions/default/src/validateInstances.js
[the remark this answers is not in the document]
rodrigobasilio2022: Removed both attirbutes.

extensions/default/src/validateInstances.ts
rodrigobasilio2022: Itsa not a service. Changed the name to DisplaySetMessageList

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
```diff
@@ -0,0 +1,87 @@
...
+  MULTIFRAME_NO_POSITION_INFORMATION: 6,
+  MISSING_FRAMES: 7,
+  IRREGULAR_SPACING: 8,
+  INCONSISTENT_DIMENSIONS: 9,
+  INCONSISTENT_COMPONENTS: 10,
+  INCONSISTENT_ORIENTATIONS: 11,
+  INCONSISTENT_POSITION_INFORMATION: 12,
+};
+
+const STRINGS = {
```
jbocce, line 16: I am not sure defining the strings here is best. We should run this by others, but I think we should just define the codes here and in a new locale file (e.g. `platform\i18n\src\locales\en-US\Messages.json`) add the messages where the key is the code. That way the messages can be properly translated. `Messsages.json` might look something like the following...

```
  "1": "No valid instances found in series.",
  "2": "Series has missing position information.",
  ... etc
```

Then every individual UI that wants to display a message can use their regular pattern to translate the message.

```
  const { t } = useTranslation('Messages');
  t(displaySet.messsages[0].code);
```

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
```diff
@@ -0,0 +1,87 @@
+const CODES = {
```
jbocce, line 1: We should check with @sedghi if this is a good location for this file.

platform/ui/src/components/DisplaySetMessageListTooltip/DisplaySetMessageListTooltip.tsx
```diff
@@ -0,0 +1,48 @@
...
+                    {index + 1}. {message.toString()}
+                  </li>
+                ))}
+              </ol>
+            </div>
+          }
+        >
+          <Icon
+            name={
+              messages.includes([DisplaySetMessage.CODES.NOT_RECONSTRUCTABLE])
```
jbocce, line 31: In my opinion, I find it a bit odd that the thumbnail decides whether a message is an error or warning. Perhaps we should consider if the message itself has a field that dictates this and can be set when the message is created.

extensions/default/src/getSopClassHandlerModule.js
```diff
@@ -92,7 +96,7 @@ function getSopClassUids(instances) {
  * @param {SeriesMetadata} series The series metadata object from which the display sets will be created
  * @returns {Array} The list of display sets created for the given series object
  */
-function getDisplaySetsFromSeries(instances) {
+function _getDisplaySetsFromSeries(instances, servicesManager) {
```
jbocce, line 99: I don't think `servicesManager` is used in this function. Please remove it.

extensions/default/src/getSopClassHandlerModule.js
```diff
@@ -203,12 +207,14 @@ const sopClassUids = [
   sopClassDictionary.EnhancedUSVolumeStorage,
 ];
 
-function getSopClassHandlerModule() {
+function getSopClassHandlerModule({ servicesManager }) {
```
jbocce, line 210: You will likely need to remove `servicesManager` from here too.

extensions/default/src/validateInstances.ts
```diff
@@ -0,0 +1,51 @@
...
+import { constructableModalities } from '@ohif/core/src/utils/isDisplaySetReconstructable';
+import { DisplaySetMessage, DisplaySetMessageList } from '@ohif/core';
+import checkMultiFrame from './utils/validations/checkMultiframe';
+import checkSingleFrames from './utils/validations/checkSingleFrames';
+/**
+ * Checks if a series is reconstructable to a 3D volume.
+ *
+ * @param {Object[]} instances An array of `OHIFInstanceMetadata` objects.
+ */
+export default function validateInstances(
```
jbocce, line 11: It is great that you refactored the various validation methods and put them into the `validations` directory. However, some validation methods add messages to the messages list while others return `boolean`. Maybe we should name them accordingly so they are easy to identify which are which? This is just a suggestion. Not a big deal.

jbocce: Also, consider making the valdiation methods more typecript compliant.

platform/ui/src/components/DisplaySetMessageListTooltip/DisplaySetMessageListTooltip.tsx
rodrigobasilio2022: This was not required so I removed the if so it will always be warnings

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
rodrigobasilio2022: Done

extensions/default/src/Panels/PanelStudyBrowser.tsx
```diff
@@ -317,6 +317,7 @@ function _mapDisplaySets(displaySets, thumbnailImageSrcMap) {
         numInstances: ds.numImageFrames,
         countIcon: ds.countIcon,
         StudyInstanceUID: ds.StudyInstanceUID,
+        messages: ds.messages,
```
sedghi, line 320: we need the same change for panelStudyBrowserTracking too

platform/ui/src/components/DisplaySetMessageListTooltip/DisplaySetMessageListTooltip.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Tooltip from '../Tooltip';
+import Icon from '../Icon';
+import { useTranslation } from 'react-i18next';
+
```
sedghi, line 6: require documentation

platform/ui/src/components/DisplaySetMessageListTooltip/DisplaySetMessageListTooltip.tsx
```diff
@@ -0,0 +1,42 @@
...
+const DisplaySetMessageListTooltip = ({ messages }): React.ReactNode => {
+  const { t } = useTranslation('Messages');
+  if (messages.size()) {
+    return (
+      <div>
+        <Tooltip
+          position="left"
+          tight={true}
+          content={
+            <div className="text-left max-w-40">
```
sedghi, line 16: is it series or displaySet messages

platform/i18n/src/locales/en-US/Messages.json
```diff
@@ -0,0 +1,14 @@
...
+  "3": "Series is not a reconstructable 3D volume.",
+  "4": "Multi frame series don't have pixel measurement information.",
+  "5": "Multi frame series don't have orientation information.",
+  "6": "Multi frame series don't have position information.",
+  "7": "Series has missing frames.",
+  "8": "Series has irregular spacing.",
+  "9": "Series has inconsistent dimensions between frames.",
+  "10": "Series has frames with inconsistent number of components.",
+  "11": "Series has frames with inconsistent orientations.",
+  "12": "Series has inconsistent position information."
```
sedghi, line 13: These messages are appearing on a thumbnail which are for displaySets not series, right?

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
```diff
@@ -0,0 +1,59 @@
...
+  MULTIFRAME_NO_POSITION_INFORMATION: 6,
+  MISSING_FRAMES: 7,
+  IRREGULAR_SPACING: 8,
+  INCONSISTENT_DIMENSIONS: 9,
+  INCONSISTENT_COMPONENTS: 10,
+  INCONSISTENT_ORIENTATIONS: 11,
+  INCONSISTENT_POSITION_INFORMATION: 12,
+};
+
+class DisplaySetMessage {
```
sedghi, line 16: both classes require documentation

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  public include(messageId: number): boolean {
+    for (let i = 0; i < this.messages.length; i++) {
+      if (this.messages[i].id === messageId) {
+        return true;
+      }
+    }
+    return false;
+  }
```
sedghi, line 47: return this.messages.some(message => message.id === messageId);

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  public includes(messageIdList: number[]): boolean {
+    for (let i = 0; i < messageIdList.length; i++) {
+      if (!this.include(messageIdList[i])) {
+        return false;
+      }
+    }
+    return true;
+  }
```
sedghi, line 56: same

extensions/default/src/getSopClassHandlerModule.js
```diff
@@ -208,7 +211,9 @@ function getSopClassHandlerModule() {
     {
       name: sopClassHandlerName,
       sopClassUids,
-      getDisplaySetsFromSeries,
+      getDisplaySetsFromSeries: instances => {
+        return _getDisplaySetsFromSeries(instances);
+      },
```
sedghi, line 216: why? you don't need this

extensions/default/src/getSopClassHandlerModule.js
```diff
@@ -18,8 +19,9 @@ const makeDisplaySet = instances => {
     value: isReconstructable,
     averageSpacingBetweenFrames,
   } = isDisplaySetReconstructable(instances);
-
   // set appropriate attributes to image set...
+  const messages = validateInstances(instances, isReconstructable);
```
sedghi, line 23: it is not validating (no error is thrown as validation implies) it is more `getDisplaySetMessages` or similar

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
sedghi: CODES can be static property of the DisplaySetMessageList

sedghi: like we have for events

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
```diff
@@ -0,0 +1,59 @@
...
+  id: number;
+
+  constructor(id: number) {
+    this.id = id;
+  }
+
+  public static get CODES() {
+    return CODES;
+  }
+}
```
sedghi, line 26: I don't understand the purpose for this class, the list is enough

extensions/default/src/validateInstances.ts
```diff
@@ -0,0 +1,51 @@
...
+): DisplaySetMessageList {
+  const messages = new DisplaySetMessageList();
+  if (!instances.length) {
+    messages.addMessage(DisplaySetMessage.CODES.NO_VALID_INSTANCES);
+  }
+
+  const firstInstance = instances[0];
+  if (firstInstance.ImageType.includes('LOCALIZER')) {
+    return messages;
+  }
```
sedghi, line 23: need a comment why the logic is like this

extensions/default/src/validateInstances.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  const isMultiframe = firstInstance.NumberOfFrames > 1;
```
sedghi, line 25: this line can be moved below for readability

extensions/default/src/validateInstances.ts
```diff
@@ -0,0 +1,51 @@
...
+    messages.addMessage(DisplaySetMessage.CODES.NO_POSITION_INFORMATION);
+  }
+
+  const sortedInstances = sortInstancesByPosition(instances);
+
+  if (isMultiframe) {
+    checkMultiFrame(sortedInstances[0], messages);
+  } else {
+    checkSingleFrames(sortedInstances, messages);
+  }
```
sedghi, line 45: isMultiframe ? checkMultiFrame(sortedInstances[0], messages) : checkSingleFrames(sortedInstances, messages);

extensions/default/src/utils/validations/checkMultiframe.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (!hasOrientation(multiFrameInstance)) {
+    messages.addMessage(DisplaySetMessage.CODES.MULTIFRAME_NO_ORIENTATION);
+  }
+
+  if (!hasPosition(multiFrameInstance)) {
+    messages.addMessage(
+      DisplaySetMessage.CODES.MULTIFRAME_NO_POSITION_INFORMATION
+    );
+  }
+}
```
sedghi, line 32: checkMultiframe should not deal with message additions, as it is the case for other checks

extensions/default/src/utils/validations/checkSeriesComponents.ts
```diff
@@ -0,0 +1,21 @@
+import toNumber from '@ohif/core/src/utils/toNumber';
+
+/**
+ * Check if the series has frames with different samplesPerPixel
+ * @param {*} instances
+ * @returns
+ */
+export default function checkSeriesComponents(instances) {
```
sedghi, line 8: rename function, doesn't convey whta the doc says

extensions/default/src/utils/validations/checkSeriesDimensions.ts
```diff
@@ -0,0 +1,22 @@
...
+  for (let i = 1; i < instances.length; i++) {
+    const instance = instances[i];
+    const { Rows, Columns } = instance;
+
+    if (Rows !== firstImageRows || Columns !== firstImageColumns) {
+      return false;
+    }
+  }
+  return true;
+}
```
sedghi, line 22: These functions are not checks, they return booleans. Throughout the repos we use is/areSometing, I recommend `areAllImageDimensionsEqual`

sedghi (commented):
great work, see my comments

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
rodrigobasilio2022: Done

platform/ui/src/components/DisplaySetMessageListTooltip/DisplaySetMessageListTooltip.tsx
rodrigobasilio2022: Done

platform/ui/src/components/DisplaySetMessageListTooltip/DisplaySetMessageListTooltip.tsx
rodrigobasilio2022: I dont know if the OHIF users, not programmers, knows about what is displaySet, but I can change if you want

platform/i18n/src/locales/en-US/Messages.json
rodrigobasilio2022: Yes

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
rodrigobasilio2022: Done

extensions/default/src/getSopClassHandlerModule.js
rodrigobasilio2022: Undone

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
rodrigobasilio2022: The idea is in the future put more attributes in this message class

extensions/default/src/utils/validations/checkMultiframe.ts
rodrigobasilio2022: The idea is to separate the messages for single frame and multiframe. What do you suggest?

platform/ui/src/components/DisplaySetMessageListTooltip/DisplaySetMessageListTooltip.tsx
sedghi: DisplaySet is a DICOM construct, see what is done in OHIF v2

extensions/default/src/utils/validations/checkMultiframe.ts
sedghi: Not important

sedghi:
[This](https://deploy-preview-3499--ohif-dev.netlify.app/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.30380506825315291544089774688247) is not good  

image

Let me talk to Dan to design this quickly

fedorov:
Would it make sense to use the one from v2, which is nicer, and is already familiar to the users?

image

sedghi:
probably we will use the warning icon that we have in other parts of v3 for consistency

sedghi:
We should change the messages from saying "series has" to "displaySet has"

rodrigobasilio2022:
Any other comment?

sedghi:
So this is the design

image

It should open on the right and it has a new icon (thicker) 
https://www.dropbox.com/s/j5a8vw52uh0c4r5/status-alert-warning.svg?dl=0

rodrigobasilio2022:
I do not know how to open to the right because as the tooltip is inside the thumbnail list it will be hidden

rodrigobasilio2022:
Made all changes except for the right position

sedghi:
@igoroctaviano Can you please fix the css for the tooltip to appear on the right side thanks

igoroctaviano:
> @igoroctaviano Can you please fix the css for the tooltip to appear on the right side thanks

Done, FYI I tried to use overflow-visible (tailwind) but it didn't work but inline styles did. It might be some conflicting tailwind style that needs further investigation.

sedghi:
thanks Igor! 
@dan-rukas I think the width is too small right?

image

igoroctaviano:
> > > @igoroctaviano Can you please fix the css for the tooltip to appear on the right side thanks
> > 
> > 
> > Done, FYI I tried to use overflow-visible (tailwind) but it didn't work but inline styles did. It might be some conflicting tailwind style that needs further investigation.
> 
> FYI @igoroctaviano, unfortunately setting the overflow visible for the thumbnail list causes it to no longer scroll! For example, this study https://deploy-preview-3499--ohif-dev.netlify.app/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.3671.4754.298665348758363466150039312520 has 7 series.
> 
> Before... image
> 
> After... image
> 
> Maybe try just `overflow-x: visible`?

It's a good catch and that's why it's hard to make the tooltip float outside of its relative parent when inside a scrollable div. It's a common UI conundrum. @sedghi I think this is not worth the trouble considering that we might need to refactor the left panel.

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
sedghi: I think the use of strings is fine and is in harmony with the rest of the i18n that we have

sedghi (approved):

sedghi:
@jbocce @igoroctaviano can you approve this if suited?

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
```diff
@@ -0,0 +1,49 @@
...
+  public addMessage(messageId: number): void {
+    const message = new DisplaySetMessage(messageId);
+    this.messages.push(message);
+  }
+
+  public size(): number {
+    return this.messages.length;
+  }
+
+  public include(messageId: number): boolean {
```
jbocce, line 40: Consider renaming this `includesMessage`.

platform/core/src/services/DisplaySetService/DisplaySetMessage.ts
```diff
@@ -0,0 +1,49 @@
...
+    return this.messages.some(message => message.id === messageId);
+  }
+
+  public includes(messageIdList: number[]): boolean {
```
jbocce, line 44: Consider renaming this `includesAllMessages` or maybe `includesEveryMessage`

extensions/default/src/utils/validations/areAllImageComponentsEqual.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Check if all voxels in series images has same number of components (samplesPerPixel)
+ * @param {*} instances
+ * @returns
+ */
+export default function areAllImageComponentsEqual(
+  instances: Array<any>
+): boolean {
+  const firstImage = instances[0];
```
jbocce, line 11: For this and all of the other validation functions, consider adding a safety check of something like 

```
if (!instances?.length) { 
  return false;
}
```

jbocce (commented):
See my comments.

rodrigobasilio2022:
> @sedghi @rodrigobasilio2022 It looks good to me except for one hiccup I found during my tests:
> 
> http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2406.131628405171680429117301556825
> 
> If you try to see the last series (thumbnail) tooltip, it's cut in half, and if you try to scroll down to see you cant: Screen Recording 2023-08-09 at 09 15 06 (1)

It seems we are not be able to resolve this unless we can show the tooltip in the right as designed, when you can enlarge the width of the tooltip box

jbocce (approved):
Looks good. Thanks so much.

igoroctaviano:
> I feel like not merging now honestly. First it was ok let's merge it although it should have appeared on the right side, now is it has weird behaviour when it is the last item. We don't know when the left panel is reworked maybe in 6 months, and I don't want to ship buggy feature.

@sedghi I'll try to find a fix for this specific bug myself later today 👍 @rodrigobasilio2022 feel free to do the same.

igoroctaviano:
@sedghi I pushed a new tooltip that uses React portals to avoid these overflow issues. Can you please take a look?

platform/ui/src/components/Tooltip/PortalTooltip.tsx
```diff
@@ -0,0 +1,102 @@
...
+import PropTypes from 'prop-types';
+
+import Card from './PortalTooltipCard';
+
+const portalNodes = {};
+
+/**
+ * A portal based tooltip.
+ */
+export default class ToolTip extends React.Component {
```
jbocce, line 12: I sort of expected this to be named `PortalTooltip` to match the name of the file. I think I prefer it being named `PortalTooltip` but maybe I am being picky. Also note that the other/existing tool tip component is named `Tooltip` (i.e. small t for tip). We should be consistent one way or the other.

fedorov:
Why introducing this completely foreign to most users concept "Display set"? That thumbnail corresponds to a DICOM series, why not call it "DICOM series"?

platform/ui/src/components/Tooltip/PortalTooltip.tsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+
+import Card from './PortalTooltipCard';
+
+const portalNodes = {};
+
+/**
+ * A portal based tooltip.
```
jbocce, line 10: React portals sound cool - until now I did not know about them. Thanks for thinking of this. However I am bit surprised that we need to use the older style class type components. I did a quick search of the code base and noticed that this would be the first/only component that uses the old style - but I might be wrong. Are we actually forced to use this? I did see this https://react.dev/reference/react-dom/createPortal#rendering-to-a-different-part-of-the-dom, does it help so that we can use a function component instead?

igoroctaviano: > React portals sound cool - until now I did not know about them. Thanks for thinking of this. However I am bit surprised that we need to use the older style class type components. I did a quick search of the code base and noticed that this would be the first/only component that uses the old style - but I might be wrong. Are we actually forced to use this? I did see this https://react.dev/reference/react-dom/createPortal#rendering-to-a-different-part-of-the-dom, does it help so that we can use a function component instead?

I repurposed this component. It's not exposed and not worth moving the trouble migrating it to function component. I added it just to workaround this specific issue, we should refactor the left panel so we can go back and use the original tooltip instead. Also, this is my last effort to address this specific issue without touching the left panel.

igoroctaviano:
> Why introducing this completely foreign to most users concept "Display set"? That thumbnail corresponds to a DICOM series, why not call it "DICOM series"?

The concept of a display set has persisted since the first version of OHIF, this tooltip is tied to the display set messages which is part of the display set. I think this was done like that for consistency and also for extensibility once that messages could extend beyond just series.

sedghi (changes requested):
@igoroctaviano Can you please add comment on top of each PortalToolTip and PortalTooltipCard on where the code is being copied over? thanks

igoroctaviano:
> @igoroctaviano Can you please add comment on top of each PortalToolTip and PortalTooltipCard on where the code is being copied over? thanks

Done, thanks.

sedghi (approved):
Looks great thanks for this. 
Also thanks @igoroctaviano for adding the portal

PR #3505, opened 2023-06-29
fix(measurements): Update the calibration tool to match changes in CS3D

wayfarer3130:
### Context

The calibration tool in CS3D changed the input behaviour a bit, and also fixed some formatting/warning issues on labels, so this is a corresponding fix in OHIF to agree with those changes.

### Changes & Results

Use the new scaling parameter in the calibrate image from CS3D
Use the units and formatting from CS3D

### Testing

Check that the calibrate tool still works.  It should not resize windows any longer.
Save a study with calibration on - it should save the correct positions of measurements now, but will NOT restore the calibration line on a fresh load.
Check that px units and "mm USER" shows up for user calibrated images.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


wayfarer3130:
This is the equivalent set of change to fix the display of the user calibraiton label and px units on the OHIF side, using the updated calibration changes.

platform/core/src/classes/MetadataProvider.ts
```diff
@@ -506,11 +477,54 @@ const metadataProvider = new MetadataProvider();
 
 export default metadataProvider;
 
+const WADO_IMAGE_LOADER = {
```
sedghi, line 480: what is the benefit of moving this down here?

wayfarer3130: Primarily that the function size doesn't keep on growing - there are a hundred plus modules in DICOM, and having them all returned in a single function is very ugly.  This breaks it up so that they are separate functions rather than keeping it all in a single gigantic function.
Also allows adding more modules after the fact, based on having the instance, rather than needing to register them with the CS3D metadata provider which doesn't assume anything about having instances.

sedghi (approved):

Issue #3357, opened 2023-05-03
[Feature Request]: v3: Viewer "read only" mode


igoroctaviano:
### Features
- Disable editing SEGs / RTs, etc
- Disable tools (tool by tool basis) e.g. never enable brush (could be done by mode config)
- Is it possible to disable all editing features at once (both editing of the images, but also editing of things like segment labels)? turn off hydration
- Render the measurements in the right panel but keep the read-only viewport (don't hydrate): When showing the intermediary viewport, don't popup and show the measurements in the right panel (panels are hydration dependent so maybe this is going to be a big change)

fedorov:
@rodrigobasilio2022 specifically:
* it should be possible (probably via deploy configuration?) to disable the ability of the user to change the labels of the segments in the screenshot below:
![image](https://github.com/OHIF/Viewers/assets/313942/ef0c857d-5743-4ad5-b000-047205b876b7)
* it should be possible to disable renaming of the segment group:
![image](https://github.com/OHIF/Viewers/assets/313942/d809fba2-c51d-422f-804a-753a515ccd72)
* it should be possible to configure disable of the measurement tools:
![image](https://github.com/OHIF/Viewers/assets/313942/04891e69-fc5f-46b3-8741-7669134a2d60)

Please let me know if you have any questions!

igoroctaviano:
@rodrigobasilio2022 just for reference, here are some notes based on the last meeting about this feature request:

- A new mode should be created without the editing tools
- The new mode should disable measurement hydration (which disables the editing of measurements)
- The measurement panel (the one that does not track measurements) could be modified so it would consume the unmapped measurements (non-hydrated measurements) so the user could see the measurement list without hydrating the measurements (this would require a new integration between the measurement service and the normal measurement panel)

james-hanks:
Use case: UI component as workflow solution https://www.loom.com/share/5606990bfe5c408094dfd2bf6a78abb8?t=124

rodrigobasilio2022:
And already merged.

igoroctaviano:
Done! #3586

vishnusureshperumbavoor:
How to make a viewport readonly in v3

rodrigobasilio2022:
Add the variable `disableEditing` in the default.js or the configuration file you are using

vishnusureshperumbavoor:
> Add the variable `disableEditing` in the default.js or the configuration file you are using

disableEditing: true,

like this??

PR #3506, opened 2023-06-29
Add edit prevention in OHIF

rodrigobasilio2022:
This PR is related to issue IDC #3357 , that want to prevent editing in OHIF by disabling renaming of Segmentation and Segments and the use of measurements tools. It adds a new attribute in the configuration called disableEditing to control the feature.

### DIFF
--- a/extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
+++ b/extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
@@ -2,20 +2,16 @@ import React, { useEffect, useState, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { SegmentationGroupTable } from '@ohif/ui';
 import callInputDialog from './callInputDialog';
-
 import { useTranslation } from 'react-i18next';
 
 export default function PanelSegmentation({
   servicesManager,
   commandsManager,
   extensionManager,
+  appConfig,
 }) {
-  const {
-    segmentationService,
-    uiDialogService,
-    uiNotificationService,
-  } = servicesManager.services;
-  const disableEditing = extensionManager._appConfig?.disableEditing;
+  const { segmentationService, uiDialogService } = servicesManager.services;
+  const disableEditing = appConfig?.disableEditing;
 
   const { t } = useTranslation('PanelSegmentation');
   const [selectedSegmentationId, setSelectedSegmentationId] = useState(null);
@@ -73,13 +69,6 @@ export default function PanelSegmentation({
     };
   }, []);
 
-  const disableEditingNotification = () => {
-    uiNotificationService.show({
-      title: 'Segmentation Edit',
-      message: 'Segmentation editing is disabled in this viewer.',
-      type: 'error',
-    });
-  };
   const onSegmentationClick = (segmentationId: string) => {
     segmentationService.setActiveSegmentationForToolGroup(segmentationId);
   };
@@ -225,13 +214,10 @@ export default function PanelSegmentation({
           activeSegmentationId={selectedSegmentationId || ''}
           onSegmentationClick={onSegmentationClick}
           onSegmentationDelete={onSegmentationDelete}
-          onSegmentationEdit={
-            disableEditing ? disableEditingNotification : onSegmentationEdit
-          }
+          onSegmentationEdit={onSegmentationEdit}
           onSegmentClick={onSegmentClick}
-          onSegmentEdit={
-            disableEditing ? disableEditingNotification : onSegmentEdit
-          }
+          onSegmentEdit={onSegmentEdit}
+          disableEditing={disableEditing}
           onSegmentColorClick={onSegmentColorClick}
           onSegmentDelete={onSegmentDelete}
           onToggleSegmentVisibility={onToggleSegmentVisibility}
@@ -297,6 +283,8 @@ PanelSegmentation.propTypes = {
   commandsManager: PropTypes.shape({
     runCommand: PropTypes.func.isRequired,
   }),
+  appConfig: PropTypes.object.isRequired,
+  extensionManager: PropTypes.object,
   servicesManager: PropTypes.shape({
     services: PropTypes.shape({
       segmentationService: PropTypes.shape({
--- a/extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
+++ b/extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
@@ -8,8 +8,14 @@ import { useTranslation } from 'react-i18next';
 export default function PanelSegmentation({
   servicesManager,
   commandsManager,
+  extensionManager,
 }) {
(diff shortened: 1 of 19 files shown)


platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -329,6 +329,12 @@ export default function ModeRoute({
         commandsManager,
       });
 
+      const disableEditing = extensionManager._appConfig?.disableEditing;
+      if (disableEditing) {
+        const { toolbarService } = servicesManager.services;
```
wayfarer3130, line 334: This absolutely doesn't belong here - really, the right way to do this is to create a new mode that simply doesn't have the editing tools.  That would obviate all the rest of this too, and would become obvious as to how to handle it.

rodrigobasilio2022: I don't want to put it here either. The requirement is to disable the editing tools based on config. Where is the best place to do that, considering the user just want to change the config, not create a new mode...

rodrigobasilio2022: I put it here because it's the best place I know that works for every mode created.

igoroctaviano: @rodrigobasilio2022 I think for this issue requirement: ```Disable tools (tool by tool basis) e.g. never enable brush``` @wayfarer3130 is right, we can disable tools (tool by tool basis) by just not including them in a new mode.

igoroctaviano (changes requested):
Reviewed. Left some comments.
👍

extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
```diff
@@ -200,9 +225,13 @@ export default function PanelSegmentation({
           activeSegmentationId={selectedSegmentationId || ''}
           onSegmentationClick={onSegmentationClick}
           onSegmentationDelete={onSegmentationDelete}
-          onSegmentationEdit={onSegmentationEdit}
+          onSegmentationEdit={
+            disableEditing ? disableEditingNotification : onSegmentationEdit
```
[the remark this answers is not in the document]
rodrigobasilio2022: Please update the issue page with the requirements.

platform/app/src/routes/Mode/Mode.tsx
rodrigobasilio2022: Please update the requirements in the issue page

extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
```diff
@@ -67,6 +73,13 @@ export default function PanelSegmentation({
     };
   }, []);
 
+  const disableEditingNotification = () => {
+    uiNotificationService.show({
+      title: 'Segmentation Edit',
+      message: 'Segmentation editing is disabled in this viewer.',
```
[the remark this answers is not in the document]
rodrigobasilio2022: Ok

platform/app/src/routes/Mode/Mode.tsx
sedghi: We should sync to discuss the requirements for IDC mode, since there are certain behaviours and tools that IDC don't want or wants and is not enabled/disabled in OHIF right now. But the correct way moving forward is IDC having its own mode (or even set of modes)

rodrigobasilio2022: I removed the modifications in this file until further discussion

extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
```diff
@@ -8,8 +8,14 @@ import { useTranslation } from 'react-i18next';
...
   commandsManager,
+  extensionManager,
 }) {
-  const { segmentationService, uiDialogService } = servicesManager.services;
+  const {
+    segmentationService,
+    uiDialogService,
+    uiNotificationService,
+  } = servicesManager.services;
+  const disableEditing = extensionManager._appConfig?.disableEditing;
```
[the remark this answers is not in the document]
rodrigobasilio2022: I refactored tha

extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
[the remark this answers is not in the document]
rodrigobasilio2022: Changed

extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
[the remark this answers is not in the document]
rodrigobasilio2022: As RT uses Segmentations, its hard to do that now. I think we need further discussion about this topic

wayfarer3130:
My suggestion on how to configure this is to add a new extension with a commands module that over-rides the default commands modules from SEG, SR, RT etc, and then I can provide an updated command handler that takes an ordered list of the commands modules names.  Let me give a concrete example for disabling the dialog popup asking about loading:
1. Move the dialog asking if the data should be loaded into a command module as a single command, say in extension-rt
2. Create a new copy of the command module in disable-editing-extension
3. In the disable editing commands module, have the show dialog just return immediately
4. Configure the 'disable-editing-extension' to load, and configure the command module in it to run before cornerstone and the rt commands module (this is done in the mode)

If you do it this way, depending on the mode you can have completely different behaviour in the load module setup.  Indeed, that could be the spot that you add things to the right handle panel instead.

sedghi:
I generally like what Bill said above in his 4 steps. Just wondering how we can do 

```Configure the 'disable-editing-extension' to load, and configure the command module in it to run before cornerstone and the rt commands module (this is done in the mode)
```

How to make sure they are running before cornerstone? I don't remember we have such way other than context which we kind of ditched

wayfarer3130:
> How to make sure they are running before cornerstone? I don't remember we have such way other than context which we kind of ditched

Ibrahim just merged a PR to FlexView that includes the ability to specify the ordering of the lookup in the commands manager. That change is specifically tagged to get merged back to OHIF, and is a 1 file change, so should be fairly easy to include in this PR in order to get it in earlier.  @rodrigobasilio2022 - if you want to get that from Ibrahim that would be great.

rodrigobasilio2022:
Ok. I think I can do that, but I was worried about the disable hydration code, because if it is not correct there is no point in reading the variable, wherever it is stored. So can you can verify if the code is correct? All hydrations can be auto loaded (RT, SR, SEG). Also, I agree the best way is to using modes to disabel or not specific tools. If we gona use modes, maybe its best to pass a configuration to the extensions inside a mode. Is there a way to do that in OHIF? Like adding a config alonside the extensions declaration in a mode?

image

igoroctaviano:
Assigned to @rodrigobasilio2022.
@rodrigobasilio2022 please add progress updates here.

rodrigobasilio2022:
No need anymore

Issue #3482, opened 2023-06-21
[Bug] The "Track measurement" dialog does not appear when the left studies panel is folded.


Yonatanmh:
### Describe the Bug

If the left studies panel in the viewer is closed before creating any annotations, the "tracked measurements for this series" dialog does not appear at all, leaving the measurements untracked. This could lead to confusion for users as they may need to realize they need to open the panel before making the first annotations in order to mark the series as a tracked series.

### Steps to Reproduce

1. Open the viewer.
2. Fold the left studies panel.
3. Create an annotation (for example, a length tool annotation).

### The current behavior

The "Track measurement" dialog does not appear when the left studies panel is folded.

### The expected behavior

The "Track measurement" dialog should appear even if the left studies panel is folded.

### OS

macOS 13.4

### Node version

19.7.0

### Browser

Chrome 114.0.5735.133 (Official Build) (arm64)

Yonatanmh:
Based on my understanding, when the panel is folded, the `PanelStudyBrowserTracking` component is not rendered, and as a result, the _useEffect_ hook, which triggers the `‘TRACK_SERIES’` event that in turn displays the requested dialog, does not get executed at all. 
One potential solution could be to move the contents of this _useEffect_ to the parent `SidePanel` component. The `SidePanel` component appears to continue to be rendered even when the studies panel is folded, so this might be a more robust place to handle this functionality.

PR #3530, opened 2023-07-12
fix(Measurement tracking): Side panel study browser can now be closed and measurements are still tracked

jbocce:
### Context

Fixes issue https://github.com/OHIF/Viewers/issues/3482

### Changes & Results

Moved the effect for firing the measurement tracking state machine events from PanelStudyBrowserTracking to TrackedCornerstoneViewport. Since at least one TrackedCornerstoneViewport is always on screen when markups are added, measurements can be tracked even if the side study panel is closed or not even included in the app.

Ensured that the state machine events are only fired once per measurement addition.

### Testing

Follow the steps in issue #3482 and ensure the desired outcome is achieved.

In general, measurements should be able to be tracked regardless of the visibility and presence of the side panel study browser.

Also ensure that only one measurement tracking state machine event is fired per measurement added. This can be done by putting a break point [here](https://github.com/jbocce/Viewers/blob/65f27c7c357475faacadf4ee8dfe0c49117ccc07/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.tsx#L176) and ensure it is hit only once per measurement added.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 
- [x] Node version: 16.14.0
- [x] Browser: Chrome 114.0.5735.199
  

### DIFF
--- a/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.tsx
+++ b/extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.tsx
@@ -154,6 +154,13 @@ function TrackedCornerstoneViewport(props) {
     };
   }, [isTracked]);
 
+  /**
+   * The effect for listening to measurement service measurement added events
+   * and in turn firing an event to update the measurement tracking state machine.
+   * The TrackedCornerstoneViewport is the best place for this because when
+   * a measurement is added, at least one TrackedCornerstoneViewport will be in
+   * the DOM and thus can react to the events fired.
+   */
   useEffect(() => {
     const added = measurementService.EVENTS.MEASUREMENT_ADDED;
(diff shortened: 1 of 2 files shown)


extensions/measurement-tracking/src/viewports/TrackedCornerstoneViewport.tsx
```diff
@@ -149,6 +154,48 @@ function TrackedCornerstoneViewport(props) {
     };
   }, [isTracked]);
 
+  useEffect(() => {
```
sedghi, line 157: This would appreaciate a note/comment about why this belongs here

sedghi (commented):
Looks good, thanks

sedghi (approved):

PR #3534, opened 2023-07-12
Fix the wrong aspect ratio load

wayfarer3130:
### Context

The resize listener in OHIFCornerstoneViewport wasn't firing the resize most of the time on DICOM SR load because the wrong parameter was being applied.
Also added fixes to OHIFCornerstoneViewport to provide ready events for image rendered so that the tests can proceed properly.  That allows testing the actual expansion/collapse of the viewport rendering properly because it is possible to know when it has been updated.
Finally, a tweak to the scrollbar size to reflect the number of images rather than being a constant size.

### Testing

This test only fails if one has added automatically opening of the measurement panel.  I'm actually not sure how to reproduce this inside vanilla OHIF.

Find a study that has a non square image (SIIM Sally dataset with XC works) in fact any dataset with an SR saved is sufficient
Save a DICOM SR to the study
Double click the DICOM SR to load into 1x1 mode
Click the load button, watching the position of measurements.  Often, previously, they wouldn't move, AND the aspect ratio of non-square images would change as actually viewed.  This is not 100%, but was 20-50%

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


sedghi:
I remember this weird effect when I used throttle for resize, that is why I made it debounce, though throttle seem much more smooth, but it has these werid flickerings

[OHIF-Viewer (1).webm](https://github.com/OHIF/Viewers/assets/7490180/f0242315-837b-4b76-ac8a-b9b6c8a315ad)

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -328,6 +350,11 @@ const OHIFCornerstoneViewport = React.memo(props => {
       Enums.Events.ELEMENT_ENABLED,
       elementEnabledHandler
     );
+    const elementToListen = elementRef.current;
```
sedghi, line 353: if this is for tests, I much rather have it inside cypress, since we have cornerstone on the window, you can grab it get the viewport, get the element, subscribe to it and just run a set interval until your closure state becomes true for the `dataRendered`

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -486,13 +518,11 @@ const OHIFCornerstoneViewport = React.memo(props => {
...
-      <div className="viewport-wrapper">
+      <div className="viewport-wrapper" data-rendered={dataRendered}>
         <ReactResizeDetector
-          handleWidth
-          handleHeight
-          skipOnMount={true} // Todo: make these configurable
-          refreshMode={'debounce'}
-          refreshRate={200} // transition amount in side panel
+          skipOnMount={true}
+          refreshMode="throttle"
```
sedghi, line 524: see my other message that has the video displaying the flickering effect

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
wayfarer3130: Right now the primary use is for tests, but there are numerous other applications for which we need this sort of state:
* Clearing views when switching to a different image/view that is taking a while (to prevent hte user having the wrong image)

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
wayfarer3130: I see the issue, debounce is definitely wrong in the current setup, but maybe there is another parameter I can use.  debounce on trailing edge would work

PR #3540, opened 2023-07-17
test: Integration test wait until rendered/queried instead of timed

wayfarer3130:
### Context

Added support to directly use the rendered/queried results.  

### Changes & Results

Use the cornerstone renderedState to wait for dicom image load instead of timing it.
Added a querying parameter to the worklist to record whether a query is in progress or not.  Exposed as a data-querying element (this is already re-rendering, so no significant change on render timing).

### Testing

Run the integration tests - they should pass (once new CS3D is incorporated)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/platform/app/cypress/support/commands.js
+++ b/platform/app/cypress/support/commands.js
@@ -46,7 +46,7 @@ Cypress.Commands.add('openStudy', PatientName => {
   cy.openStudyList();
   cy.get('#filter-patientNameOrId').type(PatientName);
   // cy.get('@getStudies').then(() => {
-  cy.wait(1000);
+  cy.waitQueryList();
 
   cy.get('[data-cy="study-list-results"]', { timeout: 5000 })
     .contains(PatientName)
@@ -65,7 +65,6 @@ Cypress.Commands.add(
       ) {
         cy.openStudyInViewer(StudyInstanceUID, otherParams);
         cy.waitDicomImage();
-        cy.wait(2000);
       }
     });
   }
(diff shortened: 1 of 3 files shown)


platform/app/cypress/support/commands.js
```diff
@@ -198,54 +200,24 @@ Cypress.Commands.add('expectMinimumThumbnails', (seriesToWait = 1) => {
...
-              $cornerstone.Enums.Events.ELEMENT_ENABLED,
-              onEnabled
+Cypress.Commands.add('waitDicomImage', () => {
+  cy.window()
+    .its('cornerstone')
+    .should($cornerstone => {
+      const enabled = $cornerstone.getEnabledElements();
+      if (enabled?.length) {
+        enabled.forEach((item, i) => {
+          if (item.viewport.viewportStatus !== 'rendered') {
```
sedghi, line 210: we should grab 'rendered' from Enums

wayfarer3130: App doesn't include @cornerstonejs/core as that is part of the extension.  I'd rather not have the dependency at that level, so that kind of leaves directly encoding the value.  If you prefer me to add the dependency then I can, but I really feel all cornerstone dependencies should be in extensions.

sedghi: you are already using it in line 207, $cornerstone

PR #3549, opened 2023-07-19
fix isReuseableViewport method

m00n620:
### Context

`isReuseableViewport` method is checking 4 conditions(position, displaySets, its width and height). As a result, viewport id is being updated when we change the grid layout or drag new series into the existing viewport.

I think we don't need to check displaySets and viewport size(width, height) in this method.

### Changes & Results

Removed 3 conditions, and it's working well with ReferenceLinesTool.


sedghi:
I don't think you need this anymore and the issue is resolved, right? I'm closing this

PR #3558, opened 2023-07-21
fix(rerendering): reduce the re-rendering of the viewport grid

sedghi:
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
+++ b/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
@@ -10,13 +10,7 @@ import {
   utilities as csUtils,
 } from '@cornerstonejs/core';
 import { MeasurementService } from '@ohif/core';
-import {
-  CinePlayer,
-  useCine,
-  useViewportGrid,
-  Notification,
-  useViewportDialog,
-} from '@ohif/ui';
+import { Notification, useViewportDialog } from '@ohif/ui';
 import {
   IStackViewport,
   IVolumeViewport,
@@ -28,6 +22,7 @@ import './OHIFCornerstoneViewport.css';
 import CornerstoneOverlays from './Overlays/CornerstoneOverlays';
 import getSOPInstanceAttributes from '../utils/measurementServiceMappings/utils/getSOPInstanceAttributes';
 import CornerstoneServices from '../types/CornerstoneServices';
+import CinePlayer from '../components/CinePlayer';
 
 const STACK = 'stack';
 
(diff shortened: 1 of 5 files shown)


extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -505,25 +432,11 @@ const OHIFCornerstoneViewport = React.memo(props => {
...
-            }
-            onFrameRateChange={frameRate =>
-              cineService.setCine({
-                id: activeViewportIndex,
-                frameRate,
-              })
-            }
-          />
-        )}
+        <CinePlayer
```
wayfarer3130, line 435: I really like this setup with just instantiating the component rather than passing all sorts of extra services in.

platform/ui/src/components/SplitButton/SplitButton.tsx
```diff
@@ -183,77 +183,83 @@ const SplitButton = ({
 
   const listItemRenderer = renderer || DefaultListItemRenderer;
 
-  return (
-    <OutsideClickHandler onOutsideClick={outsideClickHandler}>
```
wayfarer3130, line 187: If you take a look at the https://www.npmjs.com/package/react-outside-click-handler documentation, it mentions that there is a disabled parameter which is specifically designed to enable/disable this, without mounting/unmounting the entire tree.  The cost of setting disabled is also much lower than that of adding/removing the tree, and that makes the GUI more responsive.  Could you use disabled instead?

extensions/cornerstone/src/components/CinePlayer/CinePlayer.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { useEffect } from 'react';
+import { CinePlayer, useCine, useViewportGrid } from '@ohif/ui';
+import { Enums, eventTarget } from '@cornerstonejs/core';
+
+function WrappedCinePlayer({
+  enabledVPElement,
+  viewportIndex,
+  servicesManager,
+}) {
+  const { toolbarService } = servicesManager.services;
```
wayfarer3130, line 10: Could you consider getting the customization service here and if the customization service defines a cine player, then returning that CINE player instead of this one?  That way the user can replace the CINE player entirely should the mode decide they want to do that.  I've seen several implementations of different CINE players, so it is clearly something that people want to customize in the mode.

wayfarer3130 (changes requested):
Would like to see the disabled use on the tree rather than the remove/add of the entire outside click handler.

wayfarer3130 (approved):

PR #3573, opened 2023-08-01
feat(MetadataLoading) : enable metadata loading from images

gpasquie:
### Context

The goal of this PR is to allow the MetadataProvider to read the metadata of dicom images directly without making an extra WADO-RS Metadata request to the PACS.
Internaly it uses dataSetCacheManager of dicomImageLoader to access the DICOM dataset. Then dcmjs parse the dataset and add the metadata to the MetadataStore.
This PR allow the DicomJSON source to provide only the image UID + instance number without the need of putting all the image metadata into the final .json.

You can enable or disable this behavior with the variable automaticallyLoadDicomMetadata in the config file.

### Testing

This can be tested using a regular JSON DataSource where only SeriesInstanceUID, StudyInstanceUID, SOPClassUID and InstanceNumber must be present.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS:  Windows 10
- [] Node version: 
- [x] Browser: Chrome 115.0.5790.110, Firefox 115.0.3

### DIFF
--- a/extensions/default/src/DicomJSONDataSource/index.js
+++ b/extensions/default/src/DicomJSONDataSource/index.js
@@ -206,6 +206,9 @@ function createDicomJSONApi(dicomJsonConfig) {
                 imageId: instance.url,
                 ...series,
                 ...study,
+                // We need to force this flag to false in case the client wants to load the metadata from the image
+                // directly with automaticallyLoadDicomMetadata
+                imageMetadataLoaded: false
               };
               delete obj.instances;
               delete obj.series;
--- a/platform/core/src/classes/MetadataProvider.js
+++ b/platform/core/src/classes/MetadataProvider.js
@@ -7,6 +7,10 @@ import fetchPaletteColorLookupTableData from '../utils/metadataProvider/fetchPal
 import toNumber from '../utils/toNumber';
 import combineFrameInstance from '../utils/combineFrameInstance';
 
+import dicomImageLoader from '@cornerstonejs/dicom-image-loader';
+import dcmjs from 'dcmjs';
+
+
 class MetadataProvider {
   constructor() {
     // Define the main "metadataLookup" private property as an immutable property.
@@ -32,6 +36,22 @@ class MetadataProvider {
       writable: false,
       value: new Map(),
     });
+
+    Object.defineProperty(this, 'automaticallyLoadDicomMetadata', {
+      configurable: false,
+      enumerable: false,
+      writable: true,
+      value: false,
+    });
+  }
+
+  setLoadDicomMetadata(option) {
+    if (option === undefined)
+      return
+
+    console.log("mdr" + option)
+
+    this.automaticallyLoadDicomMetadata = option
   }
 
   addImageIdToUIDs(imageId, uids) {
@@ -65,11 +85,12 @@ class MetadataProvider {
       frameNumber,
     } = uids;
 
-    const instance = DicomMetadataStore.getInstance(
+    let instance = DicomMetadataStore.getInstance(
       StudyInstanceUID,
       SeriesInstanceUID,
       SOPInstanceUID
     );
+
     return (
       (frameNumber && combineFrameInstance(frameNumber, instance)) || instance
     );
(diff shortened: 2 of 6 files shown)


sedghi:
so this is only for the dicomJSON right? since for the dicomweb there is no extra call to the PACS and the metadata provider already have all the necessary information from the `instance` call

gpasquie:
Maybe we missed something in the recent changes, but in the previous versions, OHIF was doing a WADO-RS Metadata request before retrieving the instances, so we thought this could be useful as well for the dicomweb mode, as long as you know the studies, series and instances UIDs.
But in our case, yes, we are mostly interested in reducing the metadata in the JSON datasource. Actually, we are trying to open OHIF from a DICOM Key Object Selection (KOS), which usually does not contain all the images metadata (rows, column, ...).

sedghi:
I see. 
We indeed fire a metadata for wadors since that is needed to layout the viewer before the instance call for the dicomweb.
I will review your PR, the immediate thing that comes to my attention is the added dependency of cornerstone and dcmjs into the core

```
import dicomImageLoader from '@cornerstonejs/dicom-image-loader';
import dcmjs from 'dcmjs';
```

which ideally should not happen. But I will give my in-depth comments today/tomorrow

Thanks!

platform/core/src/services/DicomMetadataStore/DicomMetadataStore.ts
```diff
@@ -95,6 +95,29 @@ function _getInstanceByImageId(imageId) {
   }
 }
 
+function _updateInstance(StudyInstanceUID, SeriesInstanceUID, SOPInstanceUID, metadata) {
```
sedghi, line 98: I prefer `_updateMetadataForInstance` for consistency with next method

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -92,8 +93,16 @@ export default class HangingProtocolService extends PubSubService {
...
-      callback: displaySet => displaySet.isReconstructable ?? false,
+      // We need to check again if display set is reconstructable in case we've
+      // got some new metadata in the instances since the last time we've check
+      callback: displaySet => {
+        const {
+          value: isReconstructable,
+          missingFrames: missingFrames,
+        } = isDisplaySetReconstructable(displaySet.instances);
+
+        return isReconstructable && missingFrames == 0
```
sedghi, line 104: I think an idea would be to fire and event in the dicomMetadataStore when an instance is updated and maybe recalculate the displaySet.isReconstructable in an event handler? 

This will most likely break the dicomweb or if not break would slow it down

extensions/default/src/DicomJSONDataSource/index.js
```diff
@@ -206,6 +206,9 @@ function createDicomJSONApi(dicomJsonConfig) {
                 imageId: instance.url,
                 ...series,
                 ...study,
+                // We need to force this flag to false in case the client wants to load the metadata from the image
+                // directly with automaticallyLoadDicomMetadata
+                imageMetadataLoaded: false
```
sedghi, line 211: how about only make it false if only series/studies/sop UID are provided. For other use cases where we have the full metadata in the JSON this doesn't makes sense to be false and later in the provider it checks and reads again etc.

platform/core/src/classes/MetadataProvider.js
```diff
@@ -7,6 +7,10 @@ import fetchPaletteColorLookupTableData from '../utils/metadataProvider/fetchPal
 import toNumber from '../utils/toNumber';
 import combineFrameInstance from '../utils/combineFrameInstance';
 
+import dicomImageLoader from '@cornerstonejs/dicom-image-loader';
+import dcmjs from 'dcmjs';
```
sedghi, line 11: I thought these are new dependencies but seems like we already have them although I prefer if we can find a way to not call dicomImageLoader in the provider here, any thoughts?

platform/core/src/classes/MetadataProvider.js
```diff
@@ -65,11 +85,12 @@ class MetadataProvider {
       frameNumber,
     } = uids;
 
-    const instance = DicomMetadataStore.getInstance(
+    let instance = DicomMetadataStore.getInstance(
```
sedghi, line 88: const was fine I guess

platform/core/src/classes/MetadataProvider.js
```diff
@@ -126,6 +147,52 @@ class MetadataProvider {
   _getCornerstoneDICOMImageLoaderTag(wadoImageLoaderTag, instance) {
     let metadata;
 
+    // NOTE : we only check if the flag is set and set to false to prevent other
+    // non compatible data source to work even if automaticallyLoadDicomMetadata
+    // is set.
+    // To make your data source compatible with this you'll have to explicitly
+    // set the instance.imageMetadataLoaded to false
+    if (this.automaticallyLoadDicomMetadata && (instance.imageMetadataLoaded === false)) {
```
sedghi, line 155: as I said above, I prefer not to call the dicomImageLoader here, but even with calling it here, I don't think we need a new flag. We can derive the flag status from whether it has all necessary metadata (we should create a schema for that somewhere to validate againts, that also helps people to know what metadata is necessary) and if not we automatically switch to the dicomloader+dcmjs how about that?

sedghi: @wayfarer3130 any thought on where to call the dicomImageLoader + dcmjs other than here? (ideally they should not be here)

sedghi (commented):
see my comments. Thanks again!

platform/core/src/classes/MetadataProvider.js
wayfarer3130: What I would suggest is having a lazy load function to call that if it is defined in the metadata, it can be called to replace the metadata with a full load version of it - which then removes the function from the metadata.  That way, different providers can provide different versions of the function, some of which might do the dcmjs thing, and others might load JSON metadata, while others might just read bulkdata information which is missing.

extensions/default/src/DicomJSONDataSource/index.js
wayfarer3130: If you follow the suggestions further in this, then you can figure out when to load appropriately just by which loader it is calling, so there wouldn't be any setting here at all.

platform/core/src/classes/MetadataProvider.js
sedghi: I like this

sedghi:
can you give this a look please? @moalsad

moalsad:
> can you give this a look please? @moalsad

I agree that dicomImageLoader + dcmjs should be placed somewhere else more specific to the (JSON) data source.

gpasquie:
Replaced by https://github.com/OHIF/Viewers/pull/3935

PR #3574, opened 2023-08-02
feat(4D-ROI): ROI Segmentation panel (4D)

lscoder:
- Updated `ToolGroupServices` to make it possible to add tool instances (eg: circular brush/eraser tools are instances of BrushTool)
- Updated `calculateSUVPeak` to support 4D datasets since volumes can have multiple arrays of `scalarData`
- Added ROI Segmentation panel to 4D mode (started with a copy of that same panel from TMTV)
- Updated ROI Segmentation panel layout and created a `BrushConfiguration` component to make it possible to change all brush-related tools settings
- Added all segmentation tools (circular brush, circular eraser, sphere brush, sphere eraser, threshold brush, rectangle ROI threshold, rectangle scissor, circle scissor, sphere scissor and paint fill)
- Updated the toolbar to keep all segmentation related tools in a single group
- Updated `thresholdSegmentationByRectangleROITool` and `calculatedSUVPeak` from TMTV to also support `Rectangle ROI Threshold` (only `Rectangle ROI Threshold (start/end)` was supported before)
- Created `ToolService` to be able to trigger and listen to events (activated tools)
- Fixed 4D mode HP (PT viewports were using a wrong `toolGroup`) + a few other changes to make it similar to HP from TMTV
- Fixed some issues in `ViewportWindowLevel` related to segmentations


extensions/cornerstone-dynamic-volume/src/getHangingProtocolModule.ts
```diff
@@ -300,29 +298,10 @@ const defaultProtocol = {
...
-        preset: 'middle', // 'first', 'last', 'middle'
-      },
-    },
-    displaySets: [
-      {
-        id: 'defaultDisplaySetId',
-        matchedDisplaySetsIndex: -1,
-      },
-    ],
-  },
```
lscoder, line 325: Just trying to make it more similar to TMTV

lscoder, line 325: Just trying to make it more similar to TMTV

modes/preclinical-4d/src/index.tsx
```diff
@@ -101,39 +110,39 @@ function modeFactory({ modeConfiguration }) {
...
+            protocolId: preclinical4d.hangingProtocol,
             stageId: 'dataPreparation',
           },
         },
         {
           id: 'registration',
           name: 'Registration',
           hangingProtocol: {
-            protocolId: 'default4D',
+            protocolId: preclinical4d.hangingProtocol,
```
sedghi, line 121: I thought in each step we are applying a new protocol no?

extensions/cornerstone-dynamic-volume/src/getHangingProtocolModule.ts
```diff
@@ -280,7 +277,8 @@ function getFusionViewports() {
 }
 
 const defaultProtocol = {
-  id: 'default4D',
+  id:
```
sedghi, line 280: i don't think you should make the full pathname here, if you say 'default' it automatically should create the  `@ohif/extension-cornerstone-dynamic-volume.hangingProtocolModule.default` if i'm not mistaken

extensions/cornerstone/src/services/ToolGroupService/ToolGroupService.ts
```diff
@@ -316,16 +288,17 @@ export default class ToolGroupService {
...
-    toolNames.forEach(toolName => {
-      // Initialize the toolConfig if no configuration is provided
-      const toolConfig = configs[toolName] ?? {};
-
-      // if (volumeUID) {
-      //   toolConfig.volumeUID = volumeUID;
-      // }
-
-      toolGroup.addTool(toolName, { ...toolConfig });
+    ['active', 'passive', 'enabled', 'disabled'].forEach(toolStateName => {
```
sedghi, line 291: can we use enums for these?

extensions/cornerstone/src/init.tsx
```diff
@@ -76,18 +76,13 @@ export default async function init({
...
     uiNotificationService,
-    cineService,
     cornerstoneViewportService,
     hangingProtocolService,
     toolGroupService,
     viewportGridService,
     stateSyncService,
   } = servicesManager.services as CornerstoneServices;
 
-  window.services = servicesManager.services;
```
sedghi, line 87: please don't remove them they are here for debugging the demos

extensions/cornerstone/src/services/ToolService/ToolService.ts
```diff
@@ -0,0 +1,152 @@
...
+  };
+
+  public getActivePrimaryMouseButtonTool(toolGroupId?: string): string {
+    const { toolGroupService } = this.servicesManager
+      .services as CornerstoneServices;
+
+    const toolGroup = toolGroupService.getToolGroup(toolGroupId);
+
+    return toolGroup?.getActivePrimaryMouseButtonTool();
+  }
```
sedghi, line 70: this is something that I think toolGroupService should handle not toolService

extensions/cornerstone/src/services/ToolService/ToolService.ts
```diff
@@ -0,0 +1,152 @@
...
+import { Enums as cstEnums } from '@cornerstonejs/tools';
+import { CornerstoneServices } from '../../types';
+
+const EVENTS = {
+  TOOL_ACTIVATED: 'event::cornerstone::toolservice:toolActivated',
+  PRIMARY_TOOL_ACTIVATED:
+    'event::cornerstone::toolservice:primaryToolActivated',
+};
+
+class ToolService extends PubSubService {
```
sedghi, line 12: I don't understand why we need another service. It appears that you are only subscribing to an event and just publishing again. You can do that in the init.ts for the cornerstone extension

sedghi: if you really need a service I recommend you using `toolGroupService` for this

sedghi (commented):
thanks, see my comments

extensions/cornerstone-dynamic-volume/src/getHangingProtocolModule.ts
lscoder: I changed the id when I was comparing it with TMTV HP and I saw that one using `@ohif/extension-tmtv.hangingProtocolModule.ptCT` as id

modes/preclinical-4d/src/index.tsx
sedghi: hmmm this is confusing then and requires the a preexisting knowledge of the hanging protcol and that it should have multiple stages defined. Explicit code is much better in this situation so at least put a stage index or something as the argument

lscoder: I'm not sure if I'm following because all workflow stages are using the same `protocolId` but each one of them has a different `stageId`.

extensions/cornerstone/src/services/ToolGroupService/ToolGroupService.ts
lscoder: Updated to use the Enum from `cornerstoneTools`

extensions/cornerstone/src/init.tsx
lscoder: I always search for `window.`, `debug` and `console.` when I run `git diff` before pushing and I deleted those lines by accident

extensions/cornerstone/src/services/ToolService/ToolService.ts
lscoder: Updated

extensions/cornerstone/src/services/ToolService/ToolService.ts
lscoder: Moved to `toolGroupService`

modes/preclinical-4d/src/index.tsx
sedghi: Nevermind I understand now. Also when did we choose to add `workflow` to the mode root key? I thought we are using it in the onModeEnter and register the steps via the workflowService no?

Also seems like we have stages in the HP and stages in workflow which is a lot confusing we should think about how to rename the workflow stage to something else. I think we should opt in for `WorkflowStep` and `WorkflowStepService`

extensions/cornerstone-dynamic-volume/src/getHangingProtocolModule.ts
sedghi: does it work if you say `default` here? if so please do so

modes/preclinical-4d/src/index.tsx
lscoder: Renamed "stages" to "steps" everywhere

extensions/cornerstone-dynamic-volume/src/getHangingProtocolModule.ts
lscoder: It works if we set it to default because it would use the default HP from default mode but I've reverted my change renaming the HP back to `default4D`

lscoder:
- Renamed workflow "stages" to "steps" everywhere
- ROI Segmentation is available only when user gets to "ROI Quantification" workflow step
- Fixed issues when updating left/right panels content based on workflow settings or moving across different tabs because components were losing their state (broke WorkflowSteps dropdown, WindowLevels)
- Fixed "Rectangle ROI Threshold (start/end)" which was not working on fusion viewports
- Removed "Threshold" setting from circular and sphere brush tools (only brush size is available)
- "Rectangle ROI Threshold" now have settings similar to "Rectangle ROI Threshold (start/end)" but without start/end buttons

sedghi (approved):

PR #3580, opened 2023-08-07
feature(Mode): Ability to overwrite mode configuration and small improvements

igoroctaviano:
### Context

Allowing the override of mode's configuration through viewer config and small improvements.

### Changes & Results

- [Feature] Ability to override module configuration
- [Improvement] Hardening of patientName util
- [Improvement] Hardening default layout header's return click handler (validating dataSource before appending query params)
- [Improvement] Only render display name button in the worklist if defined in the mode configuration
- [Improvement] Pass appConfig to onModeExit and onModeEnter hooks

What are the effects of this change?
- Before vs After
- Screenshots / GIFs / Videos
-->

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/extensions/default/src/ViewerLayout/index.tsx
+++ b/extensions/default/src/ViewerLayout/index.tsx
@@ -50,13 +50,16 @@ function ViewerLayout({
     //   dataSourceIdx === -1
     //     ? undefined
     //     : `datasources=${pathname.substring(dataSourceIdx + 1)}`;
-
+    
     // Todo: Handle parameters in a better way.
     const query = new URLSearchParams(window.location.search);
     const configUrl = query.get('configUrl');
 
+    const dataSourceName = pathname.substring(dataSourceIdx + 1);
+    const existingDataSource = extensionManager.getDataSources(dataSourceName);
+
     const searchQuery = new URLSearchParams();
-    if (dataSourceIdx !== -1) {
+    if (dataSourceIdx !== -1 && existingDataSource) {
       searchQuery.append('datasources', pathname.substring(dataSourceIdx + 1));
     }
 
@@ -66,7 +69,7 @@ function ViewerLayout({
 
     navigate({
       pathname: '/',
-      search: decodeURIComponent(searchQuery.toString()),
+      search: decodeURIComponent(searchQuery.toString())
     });
   };
 
--- a/platform/docs/docs/configuration/configurationFiles.md
+++ b/platform/docs/docs/configuration/configurationFiles.md
@@ -123,7 +123,7 @@ if auth headers are used, a preflight request is required.
   - Example config:
   ```js
     modesConfiguration: {
-      ['@ohif/mode-longitudinal']: {
+      '@ohif/mode-longitudinal': {
         displayName: 'Custom Name',
         routeName: 'customRouteName',
           routes: [
--- a/platform/docs/docs/configuration/configurationFiles.md
+++ b/platform/docs/docs/configuration/configurationFiles.md
@@ -124,7 +124,30 @@ if auth headers are used, a preflight request is required.
   ```js
     modesConfiguration: {
       ['@ohif/mode-longitudinal']: {
-        displayName: 'New Name'
+        displayName: 'Custom Name',
+        routeName: 'customRouteName',
+          routes: [
+            {
+              path: 'customPath',
+              layoutTemplate: () => {
+                /** Custom Layout */
+                return {
+                  id: ohif.layout,
+                  props: {
+                    leftPanels: [tracked.thumbnailList],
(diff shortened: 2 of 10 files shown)


platform/docs/docs/configuration/configurationFiles.md
```diff
@@ -119,6 +119,15 @@ Here are a list of some options available:
 - `omitQuotationForMultipartRequest`: Some servers (e.g., .NET) require the `multipart/related` request to be sent without quotation marks. Defaults to `false`. If your server doesn't require this, then setting this flag to `true` might improve performance (by removing the need for preflight requests). Also note that
 if auth headers are used, a preflight request is required.
 - `maxNumRequests`: The maximum number of requests to allow in parallel. It is an object with keys of `interaction`, `thumbnail`, and `prefetch`. You can specify a specific number for each type.
+- `modesConfiguration`: Allows overriding modes configuration.
+  - Example config:
+  ```js
+    modesConfiguration: {
+      ['@ohif/mode-longitudinal']: {
```
sedghi, line 126: why it is an array here?

sedghi (commented):
see my comments

platform/docs/docs/configuration/configurationFiles.md
igoroctaviano: The array is just to define a property with special characters like the mode id (just a map).
I'll add more examples 👍

extensions/default/src/ViewerLayout/index.tsx
```diff
@@ -50,13 +50,16 @@ function ViewerLayout({
...
     //     ? undefined
     //     : `datasources=${pathname.substring(dataSourceIdx + 1)}`;
-
+    
     // Todo: Handle parameters in a better way.
     const query = new URLSearchParams(window.location.search);
     const configUrl = query.get('configUrl');
 
+    const dataSourceName = pathname.substring(dataSourceIdx + 1);
+    const existingDataSource = extensionManager.getDataSources(dataSourceName);
```
jbocce, line 59: Are we anticipating a data source to be in the URL that is NOT an actual, configured data source? If so, under what circumstances would this occur? I am a bit confused as to why we need this check.

igoroctaviano: I might have a mode that does not use `/${mode.routeName}/${dataSourceName}`. This means that my mode can rely on the defaultDataSource from the viewer config to define its dataSource. This way I don't need to add the datasources query param and change the defaultDataSource of the worklist coming from a random mode URL e.g. /myModeURL/otherParamWhichIsNotNecessarilyADataSourceName (no dataSource being used)

platform/docs/docs/configuration/configurationFiles.md
sedghi: I don't understand why you can't say [{ '@ohif/mode-longitudinal': {...}, { '@ohif/mode-segmentation':{...}]

and you need to say ['@ohif/mode-longitudinal']

igoroctaviano: > I don't understand why you can't say [{ '@ohif/mode-longitudinal': {...}, { '@ohif/mode-segmentation':{...}]
> 
> and you need to say ['@ohif/mode-longitudinal']

I can use array but I think its a lot easier to use a map instead? so later I can just const config = modesConfiguration[modeId] instead of iterating the array looking for the config.

jbocce: Furthermore, I am trying to test this PR and when I go to save the configuration js file, the `['@ohif/mode-longitudinal']:` automatically changes to `'@ohif/mode-longitudinal':`.

igoroctaviano: It's not required to use the brackets, this is just an explicit way of doing it. I'll update the docs.

extensions/default/src/ViewerLayout/index.tsx
jbocce: Ok cool. Thanks.

platform/docs/docs/configuration/configurationFiles.md
jbocce: Is it worth mentioning in the docs that although the `modeConfiguration` is passed to the mode factory it is up to the particular mode itself if and how it applies the configuration? That is, for example, `basic-dev-mode` and `basic-test-mode` do not currently use this `modeConfiguration`. Just a thought.

igoroctaviano: Agree, updated.

jbocce (approved):
Looks good. Approved.

Issue #3356, opened 2023-05-03
[Feature Request] Auto load derived display sets when opening series


igoroctaviano:
### Request

Derived display sets are automatically loaded when the series is open in the viewport instead of double-clicking in the derived display set thumbnail in the left panel and accepting the hydration through a dialog

### Context

In OHIF v2 when clicking on a series with derived display sets, all derived data loads automatically.
This is useful for read-only viewers that do not require the hydration workflow in v3, and just want all data related to a given loaded series to be displayed instantly.

### Possible solutions

- Add a new hanging protocol option to automatically load derived display sets for series being loaded in the viewport which would disable all hydration dialogs 
- Add a new option to the viewer configuration

james-hanks:
#3506

fedorov:
For the sake of background, here is an example in v2 that demonstrates how v2 handles a series that has 2 segmentation series: https://viewer.imaging.datacommons.cancer.gov/viewer/1.3.6.1.4.1.14519.5.2.1.7311.5101.170561193612723093192571245493. We are basically trying to get the functionality that proved to work very well in v2 for IDC and not only for IDC.

james-hanks:
@rodrigobasilio2022 could you let us know if this is fixed on IDC's side and OHIF has anything to look into?

rodrigobasilio2022:
This functionality is under review and works in IDC side, as a complement of modes

igoroctaviano:
@wayfarer3130 @sedghi do you guys think it would be useful to have a new property added to the hanging protocol to automatically load the latest derived display set for the active series? If so, I'll keep this ticket open and do that for the community. This is how v2 used to work, you didn't have to click on the SR and open the SR viewport. For now, we added this functionality only to the mode.

wayfarer3130:
I would suggest not loading the latest of any derived instances, but matching a display set instance, and taking the first item of the display set instance that matches.  That way a hanging protocol can modify how the matching rules work for what to load.  I'd also suggest doing that for segmentation data, but you might want to allow more than 1 matching rule for segmentation to succeed

igoroctaviano:
@wayfarer3130 @sedghi I'll close this issue for IDC purposes but I created another one for the better long term solution proposed by @wayfarer3130: https://github.com/OHIF/Viewers/issues/3879

PR #3585, opened 2023-08-09
Exposes hydration functions from seg and rt extensions

rodrigobasilio2022:
This PR is related to the issue #3356, and exposes hydration functions to be used in an IDC mode that auto hydrates  derived series (SEG, RTSTRUCT, SR) when user opens the main series.


sedghi:
This does not belong here. This should be the mode in the IDC mode/extension repo (not fork, just mode/extension repo that you should have similar to what radicalImaging have [here](https://github.com/RadicalImaging/OHIFExtensionsAndModes)

Only pieces of it that relates to the core should be put as the PR

rodrigobasilio2022:
Understand that. Do you think the file modes/autoload/src/loadDerivedDisplaySets.ts could be leaved in main repo to allow auto hydration or no?

sedghi (approved):

PR #3586, opened 2023-08-09
feat(configuration): read only configuration

rodrigobasilio2022:
This PR is related to issue IDC #3357 , that want to prevent editing in OHIF. It creates a new variable that prevents the edition in Segmentation table (SEG and RTSTRUCT) and open measurements with locking enabled on creation


sedghi:
You should take out the autoLoad mode from this PR and only keep the pieces that is required to be changed in the core.

rodrigobasilio2022:
Already done that

extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/TrackedMeasurementsContext.tsx
```diff
@@ -26,7 +26,7 @@ const SR_SOPCLASSHANDLERID =
  * @param {*} param0
  */
 function TrackedMeasurementsContextProvider(
-  { servicesManager, commandsManager, extensionManager }, // Bound by consumer
+  { servicesManager, commandsManager, extensionManager, appConfig }, // Bound by consumer
```
sedghi, line 29: You don't need to pass the appConfig as props in most of the changed files. Take a look at how we use `useAppConfig` hook

rodrigobasilio2022: Changed as requested

platform/app/public/config/default.js
```diff
@@ -1,4 +1,5 @@
 window.config = {
+  disableEditing: true,
```
sedghi, line 2: can you remove this please

sedghi: we should add it to the documentation in the configuration though

sedghi (changes requested):
see my comments

platform/app/public/config/default.js
rodrigobasilio2022: Ok

platform/app/public/config/default.js
```diff
@@ -1,5 +1,7 @@
 window.config = {
-  disableEditing: true,
+  // This variable disables editing in OHIF, hiding edit buttons in segmentation
+  // panel and locking already stored measurements.
+  //disableEditing: true,
```
sedghi, line 4: I rather have this inside documentation and not here at all

sedghi (approved):
I think this is good to go. Please add your comments @igoroctaviano @wayfarer3130

igoroctaviano (approved):
LGTM

PR #3560, opened 2023-07-26
fix(state synchronization): SEG and RT viewports should keep their before hydration voi 

sedghi:
fixes #3410 

### Context

As described in the issue above, when having SEG or RT viewport, and the user clicks on the load/hydrate, the old
viewport VOI LUT is not preserved and is removed. The reason for this is that this action is an "update" 
action, and we only storePresentation state on "set" actions (during disable element). 

### Changes & Results

- refactored the storePresentation into a command so that we can call it from both set and update actions.
- In addition the getPresentationIds should not increment the index at the end of the id if the action is an update (
  if in the same viewportIndex and displaySetInstanceUid we are requesting a new presentation id)

### Testing

- https://deploy-preview-3560--ohif-dev.netlify.app/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.1706.8374.643249677828306008300337414785
- 1x2 
- d&d SEG on the right
- Invert the left viewport 
- Load the SEG (hydrate)
- see that the left viewport is inverted (was not before the fix)

#### PR

- [x] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [x] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [x] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #3591, opened 2023-08-09
feat(grid): remove viewportIndex and only rely on viewportId

sedghi:
## Context
During the development of OHIF, the reliance on viewport index has proven to be problematic. This issue has seen in several ways:

- Inconsistent Viewport Indexing with Layout Changes: A change from a 2x2 to a 3x3 grid alters the viewport index, causing unnecessary re-rendering, even though the display sets remain the same. This has resulted in a lack of optimization in grid rendering during stage and layout changes.

- Viewport Movement Problems: for moving a viewport (similar to moving a person's window in Zoom, which we have in plan to support in future), the change in viewport index forces React to re-render the whole grid (as right now we have the key as viewportIndex). While the element itself hasn't changed, the viewport index change causes a problematic full re-render.

## Changes Made
- Removed Viewport Index: All occurrences of the viewport index have been replaced with viewportId. This prevents the above issues related to the inconsistency of the viewport index.
- Changed Viewport Grid State: The viewport grid state has been modified from an array to a map, using viewport IDs as keys. This provides a more stable mapping of viewports.

## Migration Guide
You should not rely on viewportIndex and activeViewportIndex anymore. 

Instead try 

```
const { activeViewportId, viewports } = viewportGridService.getState() OR useViewportGrid()
```

`Viewports` also has been changed from an array to a Map 

```
viewports: new Map(
    Object.entries({
      default: {
        viewportId: 'default',
        displaySetInstanceUIDs: [],
        viewportOptions: {
          viewportId: 'default',
        },
        displaySetSelectors: [],
        displaySetOptions: [{}],
        x: 0, // left
        y: 0, // top
        width: 100,
        height: 100,
        viewportLabel: null,
      },
    })
  ),

```

So you need to `viewports.get(viewportId` instead of `viewports[viewportIndex]`

## Testing
The implemented changes have been tested, and the following features are working as expected:

- Rendering a default viewport
- Rendering a TMTV
- Rendering segmentation and RT struct
- Enabling double-click on a thumbnail
- Changing layout, going MPR, adding measurements
- Toggle One Up works
- Viewport Labels (ABCD) works
- Presentation States is preserved
- Change from 2x2 to 3x3 works

Cheers


extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -75,45 +72,17 @@ class CornerstoneViewportService extends PubSubService
...
-      viewportOptions.viewportId =
-        this.getViewportId(viewportIndex) || `viewport-${viewportIndex}`;
-    }
-    const { viewportId } = viewportOptions;
-    const viewportInfo = new ViewportInfo(viewportIndex, viewportId);
-    if (!viewportInfo.viewportId) {
-      throw new Error('Should have viewport ID afterwards');
-    }
-
+  public enableViewport(viewportId: string, elementRef: HTMLDivElement): void {
```
wayfarer3130, line 78: Why doesn't this have the viewport options any longer - are those always already provided?

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
```diff
@@ -37,32 +37,33 @@ export default function toggleStackImageSync({
...
-        displaySetInstanceUID
-      );
-
-      if (displaySet && displaySet.isReconstructable) {
-        return true;
-      }
+  const viewportsArray = Array.from(viewports.values())
+    .filter(
+      viewport =>
+        viewport.displaySetInstanceUIDs &&
```
wayfarer3130, line 46: viewport.displaySetInstanceUIDs?.length is shorter/easier to read.

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
```diff
@@ -37,32 +37,33 @@ export default function toggleStackImageSync({
...
+
+      for (const displaySetInstanceUID of displaySetInstanceUIDs) {
+        const displaySet = displaySetService.getDisplaySetByUID(
+          displaySetInstanceUID
+        );
 
-      return false;
-    }
-  });
+        if (displaySet && displaySet.isReconstructable) {
```
wayfarer3130, line 58: displaySet?.isReconstructable

extensions/default/src/commandsModule.ts
```diff
@@ -432,14 +432,14 @@ const commandsModule = ({
         // This findOrCreateViewport returns either one of the updatedViewports
         // returned from the HP service OR if there is not one from the HP service then
         // simply returns what was in the previous state.
-        const findOrCreateViewport = (viewportIndex: number) => {
+        const findOrCreateViewport = (pos: number) => {
```
wayfarer3130, line 435: Could we call is position - pos is a little too short to be clearly understood.

sedghi:
@jbocce can you try again? I can't reproduce it

sedghi:
I don't understand haha, can you try in cognito or something else?

https://github.com/OHIF/Viewers/assets/7490180/3aef504b-2359-47b4-814e-2c5a5df8170d

jbocce:
> I don't understand haha, can you try in cognito or something else?
> 
I just tried incognito, Firefox and Edge. ALL show the same problem.

sedghi:
Ok i will look

extensions/dicom-microscopy/src/getCommandsModule.ts
```diff
@@ -124,15 +124,15 @@ export default function getCommandsModule({
     },
 
     incrementActiveViewport: () => {
-      const { activeViewportIndex, viewports } = viewportGridService.getState();
-      const nextViewportIndex = (activeViewportIndex + 1) % viewports.length;
-      viewportGridService.setActiveViewportIndex(nextViewportIndex);
+      const { activeViewportId, viewports } = viewportGridService.getState();
+      const nextViewportIndex = (activeViewportId + 1) % viewports.size;
```
wayfarer3130, line 128: This seems wrong - you should need to find the index and then add to it.  Suggest implementing this once, with a delta value and then set the delta appropriately.  Alternatively, add a viewportGridService.getActiveIndex() function that does this, and perhaps the setActiveIndex,  just as functions that apply correctly.

extensions/dicom-microscopy/src/index.tsx
```diff
@@ -63,16 +63,16 @@ export default {
...
       return (
         <MicroscopyViewport
           servicesManager={servicesManager}
           extensionManager={extensionManager}
           commandsManager={commandsManager}
-          activeViewportIndex={activeViewportIndex}
-          setViewportActive={(viewportIndex: number) => {
-            viewportGridService.setActiveViewportIndex(viewportIndex);
+          activeViewportId={activeViewportId}
+          setViewportActive={(viewportId: number) => {
```
wayfarer3130, line 74: This had better be a string here as the id is a string.

extensions/dicom-microscopy/src/getCommandsModule.ts
```diff
@@ -124,15 +124,15 @@ export default function getCommandsModule({
     },
 
     incrementActiveViewport: () => {
```
wayfarer3130, line 126: Please remove this function from microscopy - it belongs in the default one as a generic function that sets the next/previous viewport active.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -231,40 +193,28 @@ class CornerstoneViewportService extends PubSubService
   /**
    * Uses the renderingEngine to enable the element for the given viewport index
```
jbocce, line 194: Please reword this comment.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -231,40 +193,28 @@ class CornerstoneViewportService extends PubSubService
...
     const renderingEngine = this.getRenderingEngine();
-    const viewportId =
-      publicViewportOptions.viewportId || this.getViewportId(viewportIndex);
-    if (!viewportId) {
-      throw new Error('Must define viewportId externally');
-    }
 
-    const viewportInfo = this.viewportsById.get(viewportId);
+    const viewportInfo = this.viewportsById.get(
+      publicViewportOptions.viewportId
```
jbocce, line 211: I expected that `viewportId` would be used here. Why `publicViewportOptions.viewportId` instead?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -907,25 +811,25 @@ class CornerstoneViewportService extends PubSubService
...
     if (viewportInfo?.contains(displaySetInstanceUID, referencedImageId)) {
-      return activeViewportIndex;
+      return activeViewportId;
     }
 
     return (
       [...this.viewportsById.values()].find(viewportInfo =>
         viewportInfo.contains(displaySetInstanceUID, referencedImageId)
-      )?.viewportIndex ?? -1
+      )?.viewportId ?? -1
```
jbocce, line 832: The return type of this function is a string yet -1 is returned when no viewport is found. Maybe use `null` or `undefined` instead?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -412,10 +412,31 @@ export default class HangingProtocolService extends PubSubService {
     );
   }
 
+  /**
```
jbocce, line 415: Please consider updating the HP service docs pages. https://docs.ohif.org/platform/services/data/HangingProtocolService

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -555,16 +576,18 @@ export default class HangingProtocolService extends PubSubService {
...
   }
 
-  getViewportsRequireUpdate(viewportIndex, displaySetInstanceUID) {
+  getViewportsRequireUpdate(viewportId, displaySetInstanceUID) {
     const newDisplaySetInstanceUID = displaySetInstanceUID;
     const protocol = this.protocol;
     const protocolStage = protocol.stages[this.stageIndex];
     const protocolViewports = protocolStage.viewports;
-    const protocolViewport = protocolViewports[viewportIndex];
+    const protocolViewport = protocolViewports.find(
```
jbocce, line 584: As a sanity check, consider checking if `protocolViewport` is `undefined` before proceeding.

platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -649,18 +649,15 @@ class MeasurementService extends PubSubService {
    * merely causes it to fire the event with the isConsumed set to true.
    */
 
-  public jumpToMeasurement(
-    viewportIndex: number,
-    measurementUID: string
-  ): void {
+  public jumpToMeasurement(viewportId: string, measurementUID: string): void {
```
jbocce, line 652: Please update the OHIF docs accordingly.

platform/core/src/services/UIViewportDialogService/index.js
```diff
@@ -22,12 +22,12 @@ const serviceImplementation = {
 };
 
 /**
- * Show a new UI viewport dialog on the specified viewportIndex;
+ * Show a new UI viewport dialog on the specified viewportId;
```
jbocce, line 25: Update any docs accordingly.

extensions/default/src/commandsModule.ts
```diff
@@ -410,8 +410,8 @@ const commandsModule = ({
         }
         // There is a state to toggle back to. The viewport that was
         // originally toggled to one up was the former active viewport.
-        const viewportIndexToUpdate =
-          toggleOneUpViewportGridStore.activeViewportIndex;
+        const viewportIdsToUpdate =
```
jbocce, line 413: Why plural? I think it should be `viewportIdToUpdate` no?

extensions/default/src/commandsModule.ts
```diff
@@ -432,14 +432,14 @@ const commandsModule = ({
         // This findOrCreateViewport returns either one of the updatedViewports
         // returned from the HP service OR if there is not one from the HP service then
         // simply returns what was in the previous state.
-        const findOrCreateViewport = (viewportIndex: number) => {
+        const findOrCreateViewport = (pos: number) => {
           const viewport = updatedViewports.find(
-            viewport => viewport.viewportIndex === viewportIndex
+            viewport => viewport.pos === pos
```
jbocce, line 437: This is a bit weird because from what I can tell, a viewport is not storing its `pos` anywhere... correct? So I think this check no longer makes sense.  Furthermore, I think this check is from a time when somebody could toggle to one up, swap in a different series and then return to the previous layout and the swapped in series would occupy the viewport that was toggled to/from one up. From what I can tell that no longer works, so if we don't care about that feature then I think even the call to `getViewportsRequireUpdate` above is no longer necessary.

extensions/default/src/findViewportsByPosition.ts
```diff
@@ -17,7 +17,7 @@ import { StateSyncService, Types } from '@ohif/core';
 export const findOrCreateViewport = (
   hangingProtocolService,
   viewportsByPosition,
-  viewportIndex: number,
+  viewportId: string,
```
jbocce, line 20: Looking at ViewportGridProvider, I believe this argument is `pos` (i.e. position) and not `viewportId`, no?

extensions/default/src/findViewportsByPosition.ts
```diff
@@ -74,24 +74,19 @@ const findViewportsByPosition = (
...
   const initialInDisplay = [];
 
-  for (const viewport of viewports) {
+  viewports.forEach(viewport => {
     if (viewport.positionId) {
       const storedViewport = {
         ...viewport,
         viewportOptions: { ...viewport.viewportOptions },
       };
       viewportsByPosition[viewport.positionId] = storedViewport;
```
jbocce, line 83: I am not sure why we are no longer doing the deleting. Is there a reason?

extensions/dicom-microscopy/src/services/MicroscopyService.ts
```diff
@@ -293,7 +293,7 @@ export default class MicroscopyService extends PubSubService {
    * active viewport.
    *
    * @param {Object} viewer Third-party viewer API's object to be managed
-   * @param {Number} viewportIndex The index of the viewport to load the viewer
+   * @param {Number} viewportId The viewport Id where the viewer will be loaded
```
jbocce, line 296: String not Number

extensions/cornerstone-dicom-seg/src/utils/_hydrateSEG.ts
```diff
@@ -38,7 +38,7 @@ async function _hydrateSEGDisplaySet({
   // (if already is not using it) since Cornerstone3D currently only supports
   // volume viewport for segmentation
   viewports.forEach((viewport, index) => {
-    if (index === viewportIndex) {
+    if (index === viewportId) {
```
jbocce, line 41: This does not seem right, perhaps `viewport.viewportId === viewportId` instead?

extensions/cornerstone-dicom-seg/src/utils/_hydrateSEG.ts
```diff
@@ -49,7 +49,7 @@ async function _hydrateSEGDisplaySet({
 
     if (shouldDisplaySeg) {
       updatedViewports.push({
-        viewportIndex: index,
+        viewportId: index,
```
jbocce, line 52: Perhaps `viewportId: viewport.viewportId` instead???

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -322,15 +323,11 @@ function commandsModule({
         return;
       }
 
-      const filteredViewports = viewports.filter(viewport => {
-        if (!viewport.viewportOptions) {
-          return false;
+      const filteredViewports = Array.from(viewports.values()).filter(
+        viewport => {
+          return toolGroupViewportIds.includes(viewport.viewportId);
```
jbocce, line 328: Why do we no longer check `if(!viewport.viewportOptions)`?

extensions/cornerstone/src/init.tsx
```diff
@@ -283,9 +283,8 @@ export default async function init({
   );
 
   viewportGridService.subscribe(
-    viewportGridService.EVENTS.ACTIVE_VIEWPORT_INDEX_CHANGED,
-    ({ viewportIndex, viewportId }) => {
-      viewportId = viewportId || `viewport-${viewportIndex}`;
+    viewportGridService.EVENTS.ACTIVE_VIEWPORT_ID_CHANGED,
```
jbocce, line 286: Update the ViewportGridService docs please.

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.tsx
```diff
@@ -419,20 +419,19 @@ function _mapDisplaySets(
...
-              ) {
-                acc.push(viewportData.viewportLabel);
-              }
-              return acc;
-            }, []);
+      const viewportIdentificator = [];
+
+      if (numPanes !== 1) {
+        viewports.forEach(viewportData => {
+          if (
```
jbocce, line 426: Why are we no longer excluding viewports that are "beyond" the number of panes?

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -6,120 +6,125 @@ import React, {
...
-  for (const oldViewport of stateViewports) {
-    const { viewportId: oldId } = oldViewport;
-    if (!oldId) {
-      // This occurs on startup, so skip re-using it
-      continue;
-    }
-    if (viewportIdSet.has(oldId)) {
-      // oldId is already used - we can't reuse it
-      continue;
+const getActiveViewportId = (state: DefaultState, newViewports: Map) => {
```
jbocce, line 70: This kinda confuses me with the `state.activeViewportId`. Please add JS doc comments for it and consider renaming the function to something like `determineActiveViewportId` (I don't have a good name, but the current name confuses me a bit).

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -213,32 +201,39 @@ export function ViewportGridProvider({ children, service }) {
...
+
             viewport.positionId = positionId;
+
+            // If the viewport doesn't have a viewportId, we create one
+            if (!viewport.viewportOptions?.viewportId) {
+              const randomUID = utils.uuidv4().substring(0, 8);
+              viewport.viewportOptions.viewportId = `viewport-${randomUID}`;
+            }
+
+            if (!viewport.viewportId) {
```
jbocce, line 228: Does it make sense to simply not do this check (i.e. remove the if) and just always set `viewport.viewportId` to safeguard against (potentially) having a different viewportId in viewportOptions?

jbocce (changes requested):
Thanks so much for all this work. Please see my comments and use cases the might be issues.

Use cases (copied from above)

Launch a study say [deploy-preview-3591--ohif-dev.netlify.app/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5](https://deploy-preview-3591--ohif-dev.netlify.app/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5)
Change the window level of the series displayed (or scroll the stack)
Swap in a different series to replace the one that was changed (either via double click or drag and drop)
Swap the series that was originally changed back in. Notice that the changes are gone. In [viewer-dev.ohif.org](https://viewer-dev.ohif.org/) the changes are preserved.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
sedghi: we were only using viewportOptions to grab viewportId, so just using viewportId now

extensions/dicom-microscopy/src/getCommandsModule.ts
sedghi: I really don't want to give people the ability to rely on index again, the purpose of this PR is to really remove viewportIndex once for all

extensions/default/src/commandsModule.ts
sedghi: I don't think right now we have the feature you are saying

> somebody could toggle to one up, swap in a different series and then return to the previous layout and the swapped in series would occupy the viewport that was toggled to/from one up

Can you try it in the https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2403.334240657131972136850343327463

extensions/default/src/findViewportsByPosition.ts
sedghi: that is not what we use for `findViewportsByPositoin` it is getting bound as layoutFindOrCreate

extensions/default/src/findViewportsByPosition.ts
sedghi: it is causing problems for the new "no viewportIndex world", maybe it was needed when we were messed up in the "viewport index world"

extensions/cornerstone/src/commandsModule.ts
sedghi: since previously viewportId was in the options and now it is not

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.tsx
sedghi: Not sure what was that logic

platform/ui/src/contextProviders/ViewportGridProvider.tsx
sedghi: sure

sedghi:
@wayfarer3130 @jbocce I fixed the reported bugs and a handful of more bugs and tests, can you look again?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -907,25 +827,25 @@ class CornerstoneViewportService extends PubSubService
...
    */
-  public getViewportIndexToJump(
-    activeViewportIndex: number,
+  public getViewportIdToJump(
+    activeViewportId: string,
     displaySetInstanceUID: string,
     cameraProps: unknown
-  ): number {
-    const viewportInfo = this.viewportsInfo.get(activeViewportIndex);
+  ): string | null {
```
jbocce, line 838: Do we really need the | null?

extensions/default/src/commandsModule.ts
jbocce: Well the feature used to be there, but this PR was not the one that removed it.

That aside, this check for `viewport.pos === pos` does not really make sense because from what I can see, we are NEVER saving/storing this position per viewport and furthermore `HangingProtocolService.getViewportsRequireUpdate` does NOT return it (i.e. `pos`) as one of the fields so I am confused as to why we even need to check for `updatedViewports`. It is likely all we have to do in `findOrCreateViewport` is to return the previous state.

extensions/default/src/findViewportsByPosition.ts
jbocce: I am fairly certain that ViewportGridProvided does call this `findOrCreateViewport` function, but notice that the `viewportId` argument is NOT actually used, but if it were it would instead be `position`. @wayfarer3130 could you please confirm this?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -770,8 +753,8 @@ function commandsModule({
     toggleReferenceLines: {
       commandFn: actions.toggleReferenceLines,
     },
-    storePresentation: {
-      commandFn: actions.storePresentation,
+    resetCameraLayout: {
```
jbocce, line 756: Hmm... `storePresentation` has been moved/removed, yet I still see commands being run for it. I believe those were added for the PR #3560. I believe that functionality is broken now.

jbocce:
A couple more things I found...

1. I noted this in the cornerstone commandsModule file, but I confirmed that the bug fixed by PR #3560 is broken now.
2. For PET SUV, when I change the patient weight and click `Reload Data`, I get the following exception in the console...
```
Uncaught (in promise) ReferenceError: viewportIndex is not defined
    at Object.callback (OHIFCornerstoneViewport.tsx:270:11)
    at pubSubServiceInterface.js:85:16
    at W.c [as _broadcastEvent] (pubSubServiceInterface.js:84:31)
    at W.setDisplaySetMetadataInvalidated (DisplaySetService.ts:159:10)
    at onClick (PanelPetSUV.tsx:142:23)
    at onClick (Button.tsx:93:7)
    at Object.je (react-dom.production.min.js:52:317)
    at Qe (react-dom.production.min.js:52:471)
    at react-dom.production.min.js:53:35
```

jbocce (changes requested):
It is coming along... see my comments.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
sedghi: it can return null, so yes

extensions/cornerstone/src/commandsModule.ts
sedghi: yeah I didn't meant to commit it

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
jbocce: But null is assignable to string, no?

sedghi:
@jbocce Thanks for thorough review, applied/fixed based on your comment

extensions/default/src/commandsModule.ts
```diff
@@ -432,14 +432,16 @@ const commandsModule = ({
         // This findOrCreateViewport returns either one of the updatedViewports
         // returned from the HP service OR if there is not one from the HP service then
         // simply returns what was in the previous state.
-        const findOrCreateViewport = (viewportIndex: number) => {
+        const findOrCreateViewport = (position: number) => {
           const viewport = updatedViewports.find(
-            viewport => viewport.viewportIndex === viewportIndex
+            viewport => viewport.pos === position
```
jbocce, line 437: I am a bit concerned that when double clicking to one-up while in MPR and then swapping in a new (reconstructable) series, the user is left stuck in one-up until exiting MPR (i.e. they cannot double click again to return to the 3x1 MPR view). I understand that this PR was NOT responsible for that, but the changes here brought that to light.

I am still concerned that there is no `viewport.pos` defined for this code to ever work.

jbocce:
Great work fixing the issues I found! Thanks so much.

I think the following comments should be discussed and/or addressed...
- https://github.com/OHIF/Viewers/pull/3591#discussion_r1303091533
- https://github.com/OHIF/Viewers/pull/3591#discussion_r1304321413
- https://github.com/OHIF/Viewers/pull/3591#discussion_r1304284150

sedghi:
@jbocce applied all three comments

sedghi:
@jbocce I think I got it

extensions/default/src/commandsModule.ts
```diff
@@ -432,14 +428,50 @@ const commandsModule = ({
...
+          return updatedViewportsViaHP.length > 1 &&
+            updatedViewportsViaHP[position]
+            ? {
+                viewportOptions,
+                displaySetOptions,
+                ...updatedViewportsViaHP[position],
+              }
+            : newViewports[position];
+
+          // const viewport = updatedViewports.find(
```
jbocce, line 463: Please consider removing this commented out code.

extensions/default/src/commandsModule.ts
```diff
@@ -432,14 +428,50 @@ const commandsModule = ({
...
+          //   viewport => viewport.pos === position
+          // );
+
+          // return viewport
+          //   ? { viewportOptions, displaySetOptions, ...viewport }
+          //   : Array.from(toggleOneUpViewportGridStore.viewports.values())[
+          //       position
+          //     ];
+          return Array.from(toggleOneUpViewportGridStore.viewports.values())[
```
jbocce, line 472: I believe this is an unreachable return statement.

platform/core/src/utils/subscribeToNextViewportGridChange.ts
```diff
@@ -24,10 +24,6 @@ function subscribeToNextViewportGridChange(
         viewportGridService.EVENTS.LAYOUT_CHANGED,
         callback
       ),
-      viewportGridService.subscribe(
```
jbocce, line 27: I am a little bit concerned with removing this. At present it seems ok, but I suspect at some point there may be a scenario where the grid state changes and the one-up state sync becomes stale. Furthermore, perhaps this utility function should take a list of events instead of hardcoding them. Thoughts?

platform/core/src/utils/subscribeToNextViewportGridChange.ts
```diff
@@ -24,10 +24,6 @@ function subscribeToNextViewportGridChange(
...
-        viewportGridService.EVENTS.GRID_STATE_CHANGED,
```
jbocce, line 28: At the very least the doc comment for this utility function needs to be updated to exclude the `GRID_STATE_CHANGED` event.

jbocce:
> @jbocce I think I got it working for the double click too

Yes. Thank you for that. See my comments.

platform/core/src/utils/subscribeToNextViewportGridChange.ts
sedghi: i would worry about it at some point where for some scenarios (that we don't know now) would break.

jbocce (approved):
Approved.

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -329,15 +323,25 @@ const OHIFCornerstoneViewport = React.memo(props => {
...
       let measurement;
-      if (cacheJumpToMeasurementEvent?.viewportIndex === viewportIndex) {
+      if (cacheJumpToMeasurementEvent?.viewportId === viewportId) {
         measurement = cacheJumpToMeasurementEvent.measurement;
         // Delete the position presentation so that viewport navigates direct
         presentations.positionPresentation = null;
         cacheJumpToMeasurementEvent = null;
       }
 
+      // Note: This is a hack to get the grid to re-render the OHIFCornerstoneViewport component
```
wayfarer3130, line 333: I think this needs to get documented somewhere that other people can see how it gets used.  I agree, it is a hack, but sometimes a necessary hack.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -701,10 +670,11 @@ function commandsModule({
       options: { rotation: -90 },
     },
     incrementActiveViewport: {
-      commandFn: actions.incrementActiveViewport,
+      commandFn: actions.changeActiveViewport,
```
wayfarer3130, line 673: This is definitely cleaner

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
```diff
@@ -37,32 +37,29 @@ export default function toggleStackImageSync({
...
+
+      for (const displaySetInstanceUID of displaySetInstanceUIDs) {
+        const displaySet = displaySetService.getDisplaySetByUID(
+          displaySetInstanceUID
+        );
 
-      return false;
-    }
-  });
+        if (displaySet?.isReconstructable) {
```
wayfarer3130, line 54: return !!displaySet?.isReconstructable is a one line return for this.

platform/app/cypress/integration/customization/OHIFDoubleClick.spec.js
```diff
@@ -16,12 +16,18 @@ describe('OHIF Double Click', () => {
       .should('be.eq', numExpectedViewports);
 
     for (let i = 0; i < numExpectedViewports; i += 1) {
+      cy.wait(2000);
```
wayfarer3130, line 19: If you merge the CINE fix in CS3D, then this wait gets removed and you just do a waitDicomImage() above line 18.
I'd really like to see it do waitDicomImage instead of cy.wait as it is much more reliable, and is fast on fast systems, and slow on slower connections.

platform/core/src/services/HangingProtocolService/HPMatcher.js
```diff
@@ -74,14 +74,6 @@ const match = (
       errorMessages = ['Something went wrong during validation.', e];
     }
 
-    console.log(
```
wayfarer3130, line 77: Could you move this into log4js or the like instead of removing - comment it out for now, and mark it as a change for log4js rather than removing entirely becuase it is really hard to fix hanging protocols without logs indicating why things didn't work or matched unexpectedly.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -1471,15 +1510,15 @@ export default class HangingProtocolService extends PubSubService {
...
-            // Todo: why we have images here since the matching type does not have it
-            {
-              studies: this.studies,
-              instance: displaySet.images?.[0],
-              displaySetMatchDetails: this.displaySetMatchDetails,
-              displaySets: studyDisplaySets,
-            }
+          displaySet,
+          seriesMatchingRules,
+          // Todo: why we have images here since the matching type does not have it
```
wayfarer3130, line 1515: I don't think we need the actual instance any longer.

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -130,65 +139,49 @@ export function ViewportGridProvider({ children, service }) {
...
           // Use the newly provide viewportOptions and display set options
           // when provided, and otherwise fall back to the previous ones.
           // That allows for easy updates of just the display set.
-          const { viewportIndex, displaySetInstanceUIDs } = updatedViewport;
-          const previousViewport = viewports[viewportIndex] || {};
-          const viewportOptions = {
-            ...(updatedViewport.viewportOptions ||
-              previousViewport.viewportOptions),
-          };
+          const viewportOptions = merge(
```
wayfarer3130, line 152: Does this need a real merge?  It feels like it might cause mixtures that aren't intended, where one version has a value for key A, and the next one has a value for keyB, but doesn't specifically null out keyA, then keyA still applies.
Not saying it doesn't, just asking.

wayfarer3130 (approved):
There are some smallish changes, and a couple of questions, but I think this is important enough that we should proceed on it.
If you decide to include any of the fixes, that is great, but otherwise I wouldn't worry about them - you decide which comments are still worth applying.

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -63,6 +63,10 @@ function areEqual(prevProps, nextProps) {
     return false;
   }
 
+  if (nextProps.viewportOptions.needsRerendering) {
```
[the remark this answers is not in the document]
sedghi: later in the code

platform/ui/src/contextProviders/ViewportGridProvider.tsx
sedghi: I think it is fine

Issue #3541, opened 2023-07-17
[Feature Request] Fallback page for study/series/instance not found and invalid url


igoroctaviano:
### What feature or change would you like to see made?

Display a splash screen with information about the study, series, or instance not found depending on the query params used and which was not found.

Cases:
- Study not found
- Series not found (just show a notification tooltip?)
- Instance not found (just show a notification tooltip?)
- Page not found 

### With loading spinner enabled: displays 404 in the console + infinite loading 
<img width="1529" alt="Screenshot 2023-07-17 at 13 50 07" src="https://github.com/OHIF/Viewers/assets/13886933/498949b7-b2a5-45a7-8148-2421de8182f3">

### Loading spinner disabled: displays 'no mrn found' + blank screen
<img width="1534" alt="Screenshot 2023-07-17 at 13 48 41" src="https://github.com/OHIF/Viewers/assets/13886933/264cc8f1-e3bf-4f0c-b7de-d570cbb0fff6">

### Why should we prioritize this feature?

It's not clear what happened trying to load using an invalid URL.

rodrigobasilio2022:
PR  merged. Will be integrated to IDC in the next update.

PR #3592, opened 2023-08-10
feat: Create fallback for Server, Study and Series

rodrigobasilio2022:
This PR is related to issue #3541 and creates a fallback for unreachable servers and problematic study. For server it will display a black page saying server unreachable

image

And for study it will display

image

It also checks in case of using a SereisInstanceUID filter in the url if the filter was used. If not it indicates it with a snackbar notification

image

To test the server fallback, one needs to change in default.js file the qidoRoot to an unreachable server
To test the study fallback one can use the following localhost link:

Wrong: 
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.2564676639130103327764017034747167424581

Correct one:
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.256467663913010332776401703474716742458

To test the series fallback one should use the following link:
Wrong one : 
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.256467663913010332776401703474716742458&SeriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.404451122123901597115412596819231980351

Correct one: 
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.256467663913010332776401703474716742458&SeriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.40445112212390159711541259681923198035


sedghi:
please provide testing steps. You should not remove the template of the PR which includes context, changes, testing and other stuff

rodrigobasilio2022:
Add test steps

extensions/default/src/Panels/PanelStudyBrowser.tsx
```diff
@@ -69,6 +69,11 @@ function PanelStudyBrowser({
         studyInstanceUid: StudyInstanceUID,
       });
 
+      if (!qidoForStudyUID.length) {
+        window.open('/notfoundstudy', '_self');
```
sedghi, line 73: you should use react-router for this not window

sedghi: also you should do some checking like `!qidoForStudyUID?.length`

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -43,6 +45,46 @@ function defaultRouteInit(
         SeriesInstanceUID
       );
 
+      /**
+       * This function is used to check if the filter is used. Its intend is to
+       * warn the user in case of link with a SeriesInstanceUID was called
+       * @param instances
+       * @returns
+       */
+      function isSeriesFilterUsed(instances) {
```
sedghi, line 54: we should refactor all of this in utils or something

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -43,6 +45,46 @@ function defaultRouteInit(
...
+              return true;
+            } else {
+              return false;
+            }
+          } else {
+            return false; // no images so the filter didn't work
+          }
+        } else {
+          return true; // no filter so its ok
+        }
```
sedghi, line 69: this can written better

```
if (!series){
  return true
}

if (!instances.length){
  return false
}

return seriesInstanceUIDs.includes(instances[0].SeriesInstanceUID);
```

sedghi (commented):
see my comments

platform/app/src/routes/index.tsx
```diff
@@ -10,9 +10,56 @@ import Debug from './Debug';
 import NotFound from './NotFound';
 import buildModeRoutes from './buildModeRoutes';
 import PrivateRoute from './PrivateRoute';
+import PropTypes from 'prop-types';
+import { Link } from 'react-router-dom';
+
+const NotFoundServer = ({
+  message = 'Sorry, image server is unreachable.',
```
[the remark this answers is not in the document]
jbocce: This one is a bit tricky for me. How about...

`"Unable to query for studies at this time."`

Do we want them to "check" something like their data source configuration or perhaps network connection to the data source?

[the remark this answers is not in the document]
rodrigobasilio2022: I think we need to inform that

rodrigobasilio2022:
Done all changes requested

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -43,6 +46,23 @@ function defaultRouteInit(
         SeriesInstanceUID
       );
 
+      // checks if the series filter was used, if it exists
+      const seriesInstanceUIDs = filters?.seriesInstanceUID;
+      if (
+        seriesInstanceUIDs &&
+        seriesInstanceUIDs.length &&
```
sedghi, line 53: seriesInstanceUIDs?.length

platform/app/src/utils/isSeriesFilterUsed.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * This function is used to check if the filter is used. Its intend is to
+ * warn the user in case of link with a SeriesInstanceUID was called
+ * @param instances
+ * @returns
+ */
+export default function isSeriesFilterUsed(instances, filters) {
+  const seriesInstanceUIDs = filters?.seriesInstanceUID;
+  if (seriesInstanceUIDs) {
```
jbocce, line 9: I am a bit confused by the logic of this function. If I am not mistaken, this check on line 9 and subsequent return on line 10 means that when there ARE seriesInstanceUIDs line 15 will NOT get executed. Furthermore when seriesInstanceUIDs is null or undefined then line 15 will cause an error. Furthermore, the JS Doc comment for this function is confusing too.

Lastly I tried this test URL which I thought might show the notification warning/error but does NOT!

https://deploy-preview-3592--ohif-dev.netlify.app/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.256467663913010332776401703474716742458&SeriesInstanceUID=blah.blah

rodrigobasilio2022: You are conpletely right. When I refactor the function, I forgot the !

platform/app/src/utils/isSeriesFilterUsed.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * This function is used to check if the filter is used. Its intend is to
+ * warn the user in case of link with a SeriesInstanceUID was called
+ * @param instances
+ * @returns
+ */
+export default function isSeriesFilterUsed(instances, filters) {
```
jbocce, line 7: I am still confused as to what this function is intended to do based on its name and JS docs. By the now (corrected) logic, it seems to check if the series for the given `instances` is to be "kept" (return true) or filtered out (return false). If so then consider naming this function `isSeriesNotFiltered` or maybe use `isSeriesFiltered` and reverse when true and false are returned.

Furthermore, should this URL show a warning? 
http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095258.1&seriesinstanceuid=non.existent,1.3.6.1.4.1.25403.345050719074.3824.20170125095258.2
I had expected it to. But maybe I do not undertand the use case.

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -43,6 +46,22 @@ function defaultRouteInit(
...
+      if (
+        seriesInstanceUIDs?.length &&
+        !isSeriesFilterUsed(seriesMetadata.instances, filters) &&
+        !issuedWarningSeries.includes(seriesInstanceUIDs[0])
+      ) {
+        // stores the series instance filter so it shows only once the warning
+        issuedWarningSeries.push(seriesInstanceUIDs[0]);
+        uiNotificationService.show({
+          title: 'Series filter',
+          message: `Series ${seriesInstanceUIDs[0]} could not be found in the current study`,
```
jbocce, line 60: Considering this message will only display when the list specified in `seriesinstanceuid` has NO corresponding series in the study, then shouldn't we output the entire filter list (and not just the first entry) and say something instead like the following?

"Each of the series: ${seriesInstanceUIDs}  are not part of the current study. The entire study is being displayed"

Not sure of the wording here, but maybe somebody else can help with that too.

rodrigobasilio2022: I think its fine. I added 'in filter'

jbocce (approved):

sedghi (approved):

igoroctaviano (approved):
LGTM

@rodrigobasilio2022 please update your PR titles to the semantic version style (look out for other PRs title for reference)

Issue #3421, opened 2023-05-25
[Feature Request] v3: impossible to know which of the segmentation series is being displayed


fedorov:
### What feature or change would you like to see made?

Indicate to the user which segmentation series is being displayed

### Why should we prioritize this feature?

There may be more than one segmentation series in the study. In v2, it was possible to access individual segmentations in the UI, but in v3 I do not see any UI elements to indicate which series is being shown. For a demonstration and to see the differences between v2 and v3, please see #3420 that has links and screenshots.

03shek:
I think it's because after loading segmentations with "YES", the SEG names aren't auto-populating. The user must manually rename them again.
![image](https://github.com/OHIF/Viewers/assets/115667958/6558b78b-1248-4d30-bf34-78fca40991a3)

jbocce:
@fedorov, would you consider #3426 covering this request?

fedorov:
I have hard times understanding what exactly is requested in that issue.

03shek:
@fedorov If there's a name for a SEG or RTSTRUCT file on the left, when that file is loaded, that name should be present in the collapsible SEG list above its ROIs

fedorov:
Thank you for the clarification! I am not sure what "name" means, but showing SeriesDescription for the loaded SEG/RTSTRUCT would do it for me.

jbocce:
Duplicate of #3426.

jbocce:
Hi @fedorov, here is a [PR](https://github.com/OHIF/Viewers/pull/3517) that was just merged, could you please have a look to determine if it addresses this feature request? Thanks.

fedorov:
Can you please confirm if it is live in https://viewer.ohif.org/?

fedorov:
I'd say the current implementation is not sufficient. Neither SeriesDescription, nor SeriesNumber are required to have unique values. Would it be possible to highlight the corresponding series in the list on the left?

fedorov:
Also, what is the semantics of that name shown on the right? If it is supposed to match SeriesDescription, why is it editable? Are you planning to overwrite SeriesDescription at the source if the user renames it? This would not be a good idea.

03shek:
[...]

PR #3594, opened 2023-08-10
feat(thumbnail highlight): Thumbnails of hydrated series are now highlighted

jbocce:
### Context

Thumbnails representing hydratable series (i.e. SR, SEG, RTSTRUCT, etc.) are now highlighted to indicate when the item is hydrated in OHIF. 
Addresses OHIF issue #3421.
Also, thumbnails with no image that are 'active' (i.e. the viewport is selected in the viewer) are now also highlighted appropriately with a border just like the image thumbnails.
The highlighting is per as described [here](https://github.com/OHIF/Viewers/issues/3421#issuecomment-1632988610)

### Changes & Results

SEG and RTSTRUCT display sets now have their `isHydrated` flag set to true when they are hydrated. Hydrated SR display sets for tracked measurements already had this flag set appropriately.

The `ThumbnailNoImage` modality label colours and the "list-bullets" icon color when the thumbnail is hydrated are now set based on whether the associated display set is hydrated or not. Also a change was made in `ThumbnailNoImage` so that it indicates an active series (i.e. when visible and highlighted in an OHIF viewport) with a border just like the `Thumbnail` (with an image).

### Testing

A few test cases to consider...

Test A
1. View a study with tracked measurements.
2. Double click or drag and drop the SR into a viewport.
3. Opt NOT to track the measurements and simply view the SR in the viewport.
4. The thumbnail for the SR should have a border indicating that it is the active viewport. Activating a different viewport should remove the border around the thumbnail.
image

5. Similar behaviour should occur when loading a  study with SEG or RTSTRUCT and opting NOT to hydrate the segmentation.

Test B
1. View a study with tracked measurements.
2. Double click or drag and drop the SR into a viewport.
3. Opt to track the measurements.
4. The thumbnail for the hydrated SR should look similar to below...
image

5. Similar behaviour should occur when loading a  study with SEG or RTSTRUCT after opting to hydrate the segmentation.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 
- [x] Node version: 16.14.0 
- [x] Browser: Chrome 115.0.5790.171
  


sedghi (approved):

sedghi (changes requested):
So something came to my mind to try again and seems to be a bug. it is where a segment is hydrated then removed, it still shows it as hydrated

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -1132,6 +1134,18 @@ class SegmentationService extends PubSubService {
...
+  private _setDisplaySetIsHydrated(
+    displaySetUID: string,
+    isHydrated: boolean
+  ): void {
+    const {
+      DisplaySetService: displaySetService,
+    } = this.servicesManager.services;
+    const displaySet = displaySetService.getDisplaySetByUID(displaySetUID);
+    displaySet.isHydrated = isHydrated;
+    displaySetService.setDisplaySetMetadataInvalidated(displaySetUID);
```
jbocce, line 1146: Not sure if this is the best way, but at the very least it will start a discussion.

platform/ui/src/components/ThumbnailList/ThumbnailList.tsx
```diff
@@ -36,6 +36,7 @@ const ThumbnailList = ({
           imageSrc,
           messages,
           imageAltText,
+          isHydrated,
```
sedghi, line 39: it is kind of weird that we have `isHydrated` even for regular imagees (e.g., CT, PT, etc.), can we rename it to something else? I don't know what but basically it is `is this display set hydrated given that it is derived (rt, seg, sr)`

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
sedghi: hmmm, I see why you need this, but at the same time we have another listener for this at the OHIFCornerstoneViewport which invalidates the data. Maybe we need another event to decouple this?

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.tsx
```diff
@@ -248,9 +248,29 @@ function PanelStudyBrowserTracking({
       }
     );
 
+    const SubscriptionDisplaySetMetaDataInvalidated = displaySetService.subscribe(
+      displaySetService.EVENTS.DISPLAY_SET_SERIES_METADATA_INVALIDATED,
```
sedghi, line 252: we should add the same logic to the PanelStudyBrowswer too

sedghi (commented):
see my comments

platform/ui/src/components/ThumbnailList/ThumbnailList.tsx
jbocce: How about calling it `isADerivedAndHydratedDisplaySet`?

sedghi: How about `isHydratedForDerivedDisplaySet`

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
jbocce: How about a new `DisplaySetService` event called `DISPLAY_SET_UPDATED`?

sedghi: I'm not sure really, it is kind of the same event we are invalidating a metadata, but at the same time we don't want to hit that other listener in the OHIFCornerstoneViewport, aaaaah
I hate having `DISPLAY_SET_UPDATED` and `DISPLAY_SET_SERIES_METADATA_INVALIDATED`
Let's think till we meet

sedghi (approved):
Looks great.  thanks a lot!

igoroctaviano:
@fedorov with this change, do you think we could close https://github.com/OHIF/Viewers/issues/3421?

PR #3611, opened 2023-08-21
fix(memory leak): array buffer was sticking around in volume viewports

sedghi:
### Context

purge cache was not freeing up memory since the deactivate texture calls had an exception which caused the scalardata to hang around. 

Also previously we were terminating the web workers when going back to the worklist since that prob helped with the memory leak but has performance issues for invoking them again, that step is now removed, so you should expect perf gains going back and forth between studies and worklist

CC: @chafey (re web worker)

### Changes & Results

re-order the destroy calls

### Testing

after
CleanShot 2023-08-21 at 18 41 38

Before 
CleanShot 2023-08-21 at 18 44 00
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


wayfarer3130 (approved):

extensions/cornerstone-dicom-sr/package.json
```diff
@@ -44,9 +44,9 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.20.13",
-    "@cornerstonejs/adapters": "^1.10.3",
-    "@cornerstonejs/core": "^1.10.3",
-    "@cornerstonejs/tools": "^1.10.3",
+    "@cornerstonejs/adapters": "^1.11.1",
```
wayfarer3130, line 47: Could we get the prefetch fixes in as well?  That issue is also related to how the memory gets handled, and I'd like to see the two of them combined.

Issue #3480, opened 2023-06-20
Improve parsing/loading “time to first image”


igoroctaviano:
### Request

Investigate and implement possible improvements to the performance when loading the sample study below.

### Context

The time to first image (the time that it takes to render the first image in the viewport after a full reload) was higher in the new v3 version compared to OHIF v2 (legacy).

V2: https://testing-viewer.canceridc.dev/viewer/1.3.12.2.1107.5.1.4.60175.30000008042114404745300000007 - 13 sec

V3: https://ohif-v3.web.app/viewer?StudyInstanceUIDs=1.3.12.2.1107.5.1.4.60175.30000008042114404745300000016 - 35 sec

fedorov:
@igoroctaviano the issue discussed in the document where you transferred this from is not about IDC - it is in general an observation that for that study it takes much longer for the first image to show up in the viewer than in v2. There are no segmentations in that study.

igoroctaviano:
@fedorov there's also the issue with the totalsegmentator dataset. Do we need an independent issue for that?

fedorov:
Yes, we need a separate issue comparing the performance of loading large SEG between v2 and v3.

rodrigobasilio2022:
PR opened

igoroctaviano:
@rodrigobasilio2022 can you please follow up with @sedghi to get the PR merged?

sedghi:
also this one is finished?

igoroctaviano:
@sedghi yes, it's finished.

sedghi:
We just release the OHIF 3.8, you can find more details here https://ohif.org/release-notes/3p8/
If you still encounter this issue in 3.8, please re-open this.

PR #3612, opened 2023-08-22
feat(hp): enable OHIF to run with partial metadata for large studies at the cost of less effective hanging protocol

rodrigobasilio2022:
### Context
This PR is related to the issue #3480 , that complains about the excessive time spend to display the first image of studies with a very large number of series. Compared to OHIF v2, it takes 3x the time to display the first image.

This error is related to the to when the the hanging protocol is fired. The hanging protocol waits until all series metadata are loaded before open a series in the study

### Changes & Results

This PR makes two changes in the OHIF code. First, it defines a new config variable, minDisplaySetsToRunHP, that defines the mininum number of received displaySets to run the hanging protocol. The second change is related to the series thumbnails retrieval, changing the await function calls for the .then resolve calls, to not block the main context every time a new displayset is added. Should be noted that this setting could affect the correct hanging protocol application and should be used only when necessary. 
Another change is related to how ohif retrieves the series metadata: this PR add extra dicom tag request to include Series Number in the series metadata to correctly order the series before starting request. The ordered list of series is important because when ohif detects it is retrieving a large study (more than 100 series), it will retrieve first a small number of series, 10, to start the hanging protocol application
Also, we add appConfig in onModeEnter to set minDisplaySetsToRunHP in modes

### Testing

To test it, I configured the default.js pointing to IDC database (its important to set supportsWildcard: false with this database), set minDisplaySetsToRunHP to 10 and log the times with these links:

Study 1:
(v2) https://testing-viewer.canceridc.dev/viewer/1.3.12.2.1107.5.1.4.60175.30000008042114404745300000007
(v3) http://localhost:3000/viewer?StudyInstanceUIDs=1.3.12.2.1107.5.1.4.60175.30000008042114404745300000007

Study 2:
(v2) https://testing-viewer.canceridc.dev/viewer/1.3.12.2.1107.5.1.4.60175.30000008042114404745300000016
(v3) http://localhost:3000/viewer?StudyInstanceUIDs=1.3.12.2.1107.5.1.4.60175.30000008042114404745300000016 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 10
- [x] Node version: 16.14.0
- [x] Browser: Chrome 116.0.5845.97, Firefox 116.0.3


extensions/default/src/Panels/PanelStudyBrowser.tsx
```diff
@@ -123,24 +123,28 @@ function PanelStudyBrowser({
...
-      const newImageSrcEntry = {};
       const displaySet = displaySetService.getDisplaySetByUID(
         dSet.displaySetInstanceUID
       );
       const imageIds = dataSource.getImageIdsForDisplaySet(displaySet);
       const imageId = imageIds[Math.floor(imageIds.length / 2)];
 
       // TODO: Is it okay that imageIds are not returned here for SR displaySets?
-      if (imageId && !displaySet?.unsupported) {
+      if (
```
wayfarer3130, line 133: A better choice is to configure OHIF to use the server side rendered thumbnails.  Those only need series instance UID, so it should be possible to just render those as simple requests.

platform/docs/docs/configuration/configurationFiles.md
```diff
@@ -111,13 +111,15 @@ window.config = ({ servicesManager } = {}) => {
 ## Configuration Options
 
 Here are a list of some options available:
+- `minDisplaySetsToRunHP`:  If configured, sets the minimum display sets needed to run the hanging protocol and start displaying images. Used when OHIF needs to handle studies with several series and it is required that the first image should be loaded faster. Only changed if really necessary.
```
sedghi, line 114: what is `  panel and locking already stored measurements.`? 

Please also explain that HP migh break if you chose to run without the full metadata

extensions/default/src/DicomWebDataSource/wado/retrieveMetadataLoader.js
```diff
@@ -12,20 +12,27 @@ export default class RetrieveMetadataLoader {
...
    */
   constructor(
     client,
     studyInstanceUID,
     filters = {},
-    sortCriteria,
-    sortFunction
+    sortCriteria = undefined,
+    sortFunction = undefined,
+    thresholdLargeStudy = 100,
```
sedghi, line 26: I recommend more meaningul names 

such as `largeStudySeriesCountThreshold` and `initialSeriesFetchSize`

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -25,14 +25,45 @@ const { getSplitParam } = utils;
...
+
+  /**
+   * Function to apply the hanging protocol when the minimum number of display sets were
+   * received or all display sets retrieval were completed
+   * @returns
+   */
+  function applyHangingProtocol() {
+    // check if a hanging protocol is already applied. If so abort execution
+    if (hangingProtocolService.getActiveProtocol().protocol) {
+      return;
```
sedghi, line 46: this doesn't make sure it is already applied. it means that it only exists

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -63,6 +94,18 @@ function defaultRouteInit(
...
       }
       displaySetService.makeDisplaySets(seriesMetadata.instances, madeInClient);
+
+      // if minimum number of display sets to run a hanging protocol is defined
+      // check if it was reached to apply a hanging protocol
+      if (minDisplaySetsToRunHP) {
+        const displaySets = displaySetService.getActiveDisplaySets();
+        if (
+          displaySets?.length > minDisplaySetsToRunHP &&
+          !hangingProtocolService.getActiveProtocol().protocol
```
sedghi, line 104: same here

sedghi (changes requested):
see my comments

platform/app/src/routes/Mode/Mode.tsx
rodrigobasilio2022: See the comments above.

platform/app/src/routes/Mode/Mode.tsx
rodrigobasilio2022: The protocol attribute inside the hangingProtocolService object is assigned in three ocasions: in the creation of the object, in reset method, which is called whenever OHIF enters a new mode, and in the setProtocol method, which has a try/catch block that returns to the previous state in case of a problem. So if the protocol attribute is assigned it means someone called the setProtocol function. So if it is assigned and not applied it is an internal state error, that could happen without this code, right?

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -25,14 +25,45 @@ const { getSplitParam } = utils;
...
+  const minDisplaySetsToRunHP = appConfig?.minDisplaySetsToRunHP;
+
+  /**
+   * Function to apply the hanging protocol when the minimum number of display sets were
+   * received or all display sets retrieval were completed
+   * @returns
+   */
+  function applyHangingProtocol() {
+    // check if a hanging protocol is already applied. If so abort execution
+    if (hangingProtocolService.getActiveProtocol().protocol) {
```
sedghi, line 45: I still feel like this is not the correct check

rodrigobasilio2022: @wayfarer3130 Can you point the best suited test?

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -67,11 +67,29 @@ export default class StaticWadoClient extends api.DICOMwebClient {
...
+    // and Series Number in the series metadata returned to better sort series
+    // in preLoad function
+    const commaSeparatedFields = [
+      '00080021',
+      '00080031',
+      '0008103E',
+      '00200011',
+    ].join(',');
+
+    const queryParamsMinimun = {
```
wayfarer3130, line 80: queryParamsMinimum

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -67,11 +67,29 @@ export default class StaticWadoClient extends api.DICOMwebClient {
   }
 
   async searchForSeries(options) {
+    // asking to include Series Date, Series Time, Series Description
+    // and Series Number in the series metadata returned to better sort series
+    // in preLoad function
+    const commaSeparatedFields = [
```
wayfarer3130, line 73: Please call this includeField.  Also, define it statically up above.

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -67,11 +67,29 @@ export default class StaticWadoClient extends api.DICOMwebClient {
...
+      '00080031',
+      '0008103E',
+      '00200011',
+    ].join(',');
+
+    const queryParamsMinimun = {
+      includefield: commaSeparatedFields,
+    };
+    const updatedOptions = {
+      ...options,
```
wayfarer3130, line 84: What happens if options already has a query params?  You need to join this. 
Also, the include field should only be applied at the required level.
Note that static wado client is used for more back ends than just static wado ones.  It also sometimes supports real queries, so I wouldn't just delete the remaining queries.

extensions/default/src/DicomWebDataSource/wado/retrieveMetadataLoader.js
```diff
@@ -12,20 +12,27 @@ export default class RetrieveMetadataLoader {
...
    */
   constructor(
     client,
     studyInstanceUID,
     filters = {},
-    sortCriteria,
-    sortFunction
+    sortCriteria = undefined,
+    sortFunction = undefined,
+    largeStudySeriesCountThreshold = 100,
```
wayfarer3130, line 26: This is probably a bit high - I would guess around 50 is where it starts hitting issues.  Also, the value should come from the datasource being queried, and not be provided globally - dififerent data sources might behave sufficiently different that it doesn't make sense to apply the same behaviour to them.

extensions/default/src/DicomWebDataSource/wado/retrieveMetadataLoaderAsync.js
```diff
@@ -74,7 +74,7 @@ export default class RetrieveMetadataLoaderAsync extends RetrieveMetadataLoader
     return sortStudySeries(
       naturalized,
       sortCriteria ||
-        sortingCriteria.seriesSortCriteria.seriesInfoSortingCriteria,
+      sortingCriteria.seriesSortCriteria.seriesInfoSortingCriteria,
```
wayfarer3130, line 77: The sorting criteria should be globally configured, although a better choice might be to get it from the customization framework so that it can be set in the mode itself.

platform/app/src/routes/Mode/Mode.tsx
wayfarer3130: This is all in one file so you can use a local variable to know whether or not you have applied a protocol in this mode.

extensions/default/src/DicomWebDataSource/wado/retrieveMetadataLoaderAsync.js
```diff
@@ -92,6 +92,21 @@ export default class RetrieveMetadataLoaderAsync extends RetrieveMetadataLoader
...
+      while (seriesAsyncLoader.hasNext()) {
+        const promise = seriesAsyncLoader.next();
+        promises.push(promise);
+        firstGroup.push(promise);
+        if (firstGroup.length >= this.initialSeriesFetchSize) {
+          break;
+        }
+      }
+      // wait for all promises in the first group to be settled
+      await Promise.allSettled(firstGroup);
```
wayfarer3130, line 107: This doesn't quite feel like it should be here - this method is supposed to return immediately, after having initiated the various requests, just providing the set of promises.  What you want is to be able to group the requests so that there are no more than N of them concurrently, but let them proceed as fast as possible, just allowing that many at once.
Then, once the first set of instances is settled, you can return, but you might well get many more than just than just N of them.  Probably return 2 promises, one of which is the initial group list - which you always return regardless of size, and the second of which is the all settled on everything promise.  Code is a bit of a pain for this as you really do care that it is the first N of them, and not just a random set.

wayfarer3130: Thinking about how to implement that, what you want is a set of promises which don't initiate the actual retrieve, but have an initiate retrieve call as part of the promise.  The retrieve initiate call does:
1. Setup the http retrieve
2. When the http retrieve completes, find the next promise in order to retrieve, and call that initiate method
3. Complete the promise
Then, the retrieve metadata loader async knows nothing about retrieve a first data set, it simply always knows to retrieve the metadata in teh sorted order, working in groups of N retrieves at a time, where N is defined by the data source.

platform/app/src/routes/Mode/Mode.tsx
wayfarer3130: You don't want a check here at all - what you want is to just apply the hanging protocol when there first N items are applied, and the remaining items will simply continue retrieving until done.

extensions/default/src/DicomWebDataSource/wado/retrieveMetadataLoaderAsync.js
wayfarer3130: Thought about this more last night, what you want to do is ALWAYS use an allSettled - but the choice of which promises are in the allSettled simply depends on the configured number of items to include that are required to run the hanging protocol.  Then, when the allSettled completes, you apply the hanging protocol.
The code described in the async loader ensures that the instances are loaded in order.  It is also possible to add another initiate if specific instances are required in other orders.

wayfarer3130 (changes requested):
The logic about what is required to initiate the hanging protocol needs to be in the Mode definition, while the logic about the preload in order needs to be in the async loader, as described in the comments.

platform/app/src/routes/Mode/Mode.tsx
rodrigobasilio2022: A created a variable inside the function that is set to true when the hanging protocol is applied

rodrigobasilio2022: The allSettle that calls the applyHangingProtocol is in study level. I dont have access to the series promises. And even if I had it, I wouldnt use because hangingprotocol sees displaySets, not seriesMetadata

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -82,24 +125,7 @@ function defaultRouteInit(
   // until we run the hanging protocol matching service.
 
   Promise.allSettled(allRetrieves).then(() => {
```
wayfarer3130, line 127: Just change this to select the promises to wait for before applying the hanging protocol - I know this is the series retrieves, not the display sets, but I think it is reasonable to apply here.

platform/app/src/routes/Mode/Mode.tsx
wayfarer3130: You do have access to the series promises - the allSettled below that I just commented on is where you have access to the series level promises.  You run the HP when enough series are loaded, not when the given display sets are actually available.

extensions/default/src/DicomWebDataSource/wado/retrieveMetadataLoaderAsync.js
rodrigobasilio2022: I feel like creating this will make a lot of internal changes that I dont know this PR should cover, because all the code expects the promises array has one promise per serie

platform/app/src/routes/Mode/Mode.tsx
rodrigobasilio2022: Need to understand better that... I am not sure about that

extensions/default/src/Panels/PanelStudyBrowser.tsx
rodrigobasilio2022: I doens appear to be supported abroad. I tried /studies/{study}/series/{series}/thumbnail  with Orthanc and it didnt work.

platform/app/src/routes/Mode/Mode.tsx
rodrigobasilio2022: Changed as discussed in the meeting

rodrigobasilio2022:
Updated the code and fix microscopy opening

extensions/default/src/getHangingProtocolModule.js
```diff
@@ -14,6 +14,7 @@ const defaultProtocol = {
   editableBy: {},
   protocolMatchingRules: [],
   toolGroupIds: ['default'],
+  minSeriesLoadedToRunHP: 10,
```
wayfarer3130, line 17: This shouldn't be a simple value, but should be something like:
runHPCondition: {
  minSeriesLoaded: 10,
}
then, that condition can have other child elements added in the future to define more conditions.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -432,13 +442,21 @@ function createDicomWebApi(dicomWebConfig, userAuthenticationService) {
...
+        }
+        return promise.then(instances => {
           storeInstances(instances);
-        })
-      );
-      await Promise.all(seriesDeliveredPromises);
-      setSuccessFlag();
+        });
+      });
+      if (returnPromises) {
```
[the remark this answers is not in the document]
igoroctaviano: @rodrigobasilio2022 comment for you

igoroctaviano:
> The logic about what is required to initiate the hanging protocol needs to be in the Mode definition, while the logic about the preload in order needs to be in the async loader, as described in the comments.

@rodrigobasilio2022 this comment is still valid?

rodrigobasilio2022:
> > The logic about what is required to initiate the hanging protocol needs to be in the Mode definition, while the logic about the preload in order needs to be in the async loader, as described in the comments.
> 
> @rodrigobasilio2022 this comment is still valid?

No

extensions/default/src/DicomWebDataSource/index.js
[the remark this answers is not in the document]
rodrigobasilio2022: I don't think this is ugly because in mode.tsx this function is not called directly: there is another function that checks in the server configuration if lazy study loading is ok. If it's not, it will retrieve data synchronously, not considering this function. So either way I will need to check the returning value of the async function to see if the series promises are defined to use them. The way implemented in this PR changed the code the minimum, to try to avoid inserting some regression bugs.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -90,6 +90,9 @@ function createDicomWebApi(dicomWebConfig, userAuthenticationService) {
         singlepart: dicomWebConfig.singlepart,
         headers: userAuthenticationService.getAuthorizationHeader(),
         errorInterceptor: errorHandler.getHTTPErrorHandler(),
+        largeStudySeriesCountThreshold:
```
wayfarer3130, line 93: You aren't actually using the large study series count threshold anywhere, and that value doesn't belong in the data source, as it should simply fetch the initialSeriesFetchSize instances regardless of the study size.

rodrigobasilio2022: Removed both variables

extensions/default/src/DicomWebDataSource/index.js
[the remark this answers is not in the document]
wayfarer3130: @sedghi - I really dislike the returning of different types/paths here because it means you need to know how the function is behaving fundamentally to know whether you can wait on various elements.  What I was suggestion was to always return the same type of value, but to have attributes of the returned promise available to get sub-parts of the promise.    That is, make series.metadata return a Promise<Map<string,Promise<SeriesMetadata>>> 
Then, only for the Mode version of it, if the resolution of the promise is a Map, then it contains the additional promises, which are then sub-selected to wait for the required set - all the time, based just on the hanging protocols defined/being applied, and if it isn't a Map, then don't wait for anything inside the nested promise.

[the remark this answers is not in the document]
rodrigobasilio2022: I agree this is an important point and we should wait for @sedghi comments

sedghi (commented):
I will wait for the multi retrieved to be merged and this PR updated before review, since there are too many overlapping zones

igoroctaviano:
@rodrigobasilio2022 can you please finish the comments/follow up with @sedghi  so we can merge this PR asap? I think you doing it its going to be a lot faster. Thanks in advance!

igoroctaviano:
> I will wait for the multi retrieved to be merged and this PR updated before review, since there are too many overlapping zones

@rodrigobasilio2022 can you please address the conflicts? we are probably going to merge this one first considering that the multi server implementation is going to change.

PR #3631, opened 2023-09-04
WIP fix(SR): Fix/sr SCOORD3D not loading

rodrigobasilio2022:
### Context

This PR is related to IDC issue #1215 and wants to enable OHIF to support SCOORD3D annotations and SR polyline annotations  in IDC that didnt follow the OHIF / Cornerstone standards. It introduces the idea to customize the dicomSr extension by checking if 'dicomSrExtensionCustomizations' id is defined in a modeCustomization. If yes it checks if a convertAnnotation function is defined and execute it if it exists. This function will be responsible for converting the non conformant SRs to OHIF/Cornerstone compatible ones by rewriting the SR metadata. An example of such function is defined in here. https://github.com/ImagingDataCommons/ohif-idc-mode/pull/3

### Changes & Results

This PR adds checking inside dicomsr extension to use, if defined a function called converAnnotations that should perform the proper and necessary converstions to support non conformant SRs, in this case SCOORD3D point annotations and polyline annotations. 

The changes only affect modes that register the dicomSRExtensionCustomizations in customizationServices.

### Testing

To test it, one should create a new mode or modify the longitudinal mode following this PR https://github.com/ImagingDataCommons/ohif-idc-mode/pull/3.

And need to setup a local dicomweb server and add the studies in the following dropbox link:
https://www.dropbox.com/s/98rylgt25b2sm9r/planar_annotations.zip?dl=0

And open the respective studies. Once opened, just click in the SR series, that originally has only two annotations, but one will see that ten new arrow annotations were added in the measurements tabel, were one can jump to all of them.

image

image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 10
- [x] Node version: 16.14.0
- [x] Browser: Chrome 116.0.5845.97, Firefox 116.0.3


extensions/cornerstone/src/initMeasurementService.js
```diff
@@ -375,7 +375,12 @@ const connectMeasurementServiceToTools = (
           label: data.annotation.data.label,
           frameNumber: frameNumber,
         },
-      });
+      };
+      if (data.annotation.data?.polyline) {
```
igoroctaviano, line 379: Can we make this more generic so it works for all graphic types?

extensions/cornerstone-dicom-sr/src/utils/getClosestInstanceInfo.ts
```diff
@@ -0,0 +1,100 @@
...
+  // todo: this does not assume orientation yet, but that can be added later
+  const displaySetInstanceUID = displaySet.displaySetInstanceUID;
+  return displaySet.instances.reduce((closestInstanceInfos, instance) => {
+    const distance = planeDistance(targetPoint, instance);
+
+    // the threshold is half of the slicethickness or 5 mm
+    const threshold = 0.1; //(instance?.SliceThickness || 5) / 2;
+
+    if (distance < threshold) {
+      const closestInstanceInfo = {
```
igoroctaviano, line 74: Can you please rename this file? I'm not sure what get closest instance info means here.

extensions/cornerstone-dicom-sr/src/utils/convertAnnotations.ts
```diff
@@ -0,0 +1,237 @@
...
+
+        const mappedTrackingIdentifier = `${cornerstoneTag}:${toolName}`;
+        TrackingIdentifierGroup.TextValue = mappedTrackingIdentifier;
+      }
+    }
+  }
+  return { newlyGeneratedMeasurementGroups, newlyFoundReferencedImages };
+}
+
+export function convertAnnotations(dataset, displaySetService) {
```
igoroctaviano, line 206: Same here, not sure what convertAnnotation means here.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -155,9 +155,15 @@ function _load(displaySet, servicesManager, extensionManager) {
   const dataSources = extensionManager.getDataSources();
   const dataSource = dataSources[0];
 
+  const referencedImages = convertAnnotations(displaySet.instance, displaySetService);
```
igoroctaviano, line 158: Maybe this should be called getReferencedImages? Alos, why you need this property?

igoroctaviano (changes requested):
Code Review boostrap

extensions/cornerstone/src/initMeasurementService.js
rodrigobasilio2022: I changed to put all attributes in data.annotation.data. Is that correct @sedghi?

extensions/cornerstone-dicom-sr/src/utils/convertAnnotations.ts
rodrigobasilio2022: This function will be removed from here and put in the IDC mode

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
rodrigobasilio2022: The functino returns all the referencedImages array to not call the extension function _getReferencedImagesList that will break with IDC metadata

fedorov:
I would like to have a test instance of this PR with OHIF that can be pointed to a Google DICOM store, before this is merged.

rodrigobasilio2022:
In my tests I am doing exactly that.

fedorov:
>  In my tests I am doing exactly that.

Very good. In this case I hope it should be easy to create a setup where I would be able to do it as well.

rodrigobasilio2022:
I will talk to @igoroctaviano so I can create something like the https://ohif-v3.web.app/

extensions/cornerstone/src/initMeasurementService.js
rodrigobasilio2022: I make this import because here an important attribute, polyline, isn´t imported. That could be the case for another annotations.

igoroctaviano:
@sedghi can't we have something like https://github.com/OHIF/Viewers/blob/v2-legacy/extensions/cornerstone/src/tools/DICOMSRDisplayTool.js for v3/cornerstone3d? This way we have a generic cs tool that renders any kind of scoord3d graphic type.

sedghi:
@igoroctaviano Don't we have it already in OHIF in measurement tracking?

igoroctaviano:
> @igoroctaviano Don't we have it already in OHIF in measurement tracking?

@rodrigobasilio2022 please check if we could use this tool then.

rodrigobasilio2022:
It already has. The problem is that it doesnt enable measure jump and, of course, it didn't open the any of these SR examples

igoroctaviano:
> It already has. The problem is that it doesnt enable measure jump and, of course, it didn't open the any of these SR examples

Can we make it work then? what we can do to fix the jump and the support?

rodrigobasilio2022:
> The reason I would like to do this is because I want to confirm it works not only for the test samples, but also for the similar annotations we create. In the past, with v2, the implementation was not working the way we expected. I don't think there is any urgency to merge this PR, it is more important to confirm it works as expected first.

https://d3w0jbfdn2j3pj.cloudfront.net/

rodrigobasilio2022:
No this is not ready for review, as we need to add the functionality described in the issue, maybe creating a new MeasurementTable

sedghi:
You mean on the SR view, you want measurement panel to show annotations? but right now we only show for hydrated?

fedorov:
@rodrigobasilio2022 before the code review, I would like to have access to a test instance to confirm it works as intended.

igoroctaviano:
@rodrigobasilio2022 whats the status of this PR based on our last iteration? Have you made the changes? I think you doing it its going to be a lot faster than handing it off. Let me know what you think.

PR #3632, opened 2023-09-06
feat/segmentation mode

sedghi:
### Context

Adds new segmentation mode with brushes, add new segmentation and export capabilities

~~Requires this PR to be merged first https://github.com/cornerstonejs/cornerstone3D/pull/771~~

Fixes various issues such as https://github.com/OHIF/Viewers/issues/3660 

#### Add new segmentation on viewports and edit via Segmentation tools and export

https://github.com/OHIF/Viewers/assets/7490180/f644582c-6ba0-470f-bc30-75ab298b0e1d

####  toolGroupService.createToolGroupAndAddTools

this function has changed to make it simpler, now the tool hosts the configuration too

Before

```js
const tools = {
  passive: [
    {
      toolName: toolNames.ArrowAnnotate,
    },
  ],
  // the rest of tools
}

const configuration = {
  // the rest of configuration
  [toolNames.ArrowAnnotate]: {
    getTextCallback: (callback, eventDetails) => {
      commandsManager.runCommand("arrowTextCallback", {
        callback,
        eventDetails,
      })
    },

    changeTextCallback: (data, eventDetails, callback) =>
      commandsManager.runCommand("arrowTextCallback", {
        callback,
        data,
        eventDetails,
      }),
  },
}

toolGroupService.createToolGroupAndAddTools("toolGroupId", tools, configuration)

```

After 

```js
const tools = {
  // the rest
  passive: [
    {
      toolName: toolNames.ArrowAnnotate,
      configuration: {
        getTextCallback: (callback, eventDetails) => {
          commandsManager.runCommand("arrowTextCallback", {
            callback,
            eventDetails,
          })
        },

        changeTextCallback: (data, eventDetails, callback) =>
          commandsManager.runCommand("arrowTextCallback", {
            callback,
            data,
            eventDetails,
          }),
      },
    },
  ],
  // the rest of tools
}

toolGroupService.createToolGroupAndAddTools('toolGroupId', tools);

```

### Testing

- Segment on https://deploy-preview-3632--ohif-dev.netlify.app/segmentation?StudyInstanceUIDs=2.16.840.1.114362.1.11972228.22789312658.616067305.306.2
- For export (create report) you should run it with your local server (e.g., orthanc), however, you should be able to Download the segmentation 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


SachidanandAlle:
Thank for getting this out as a basic feature in ohif viewer 👍

jbocce:
I have not started looking at the code changes yet, but I have started to briefly test the new segmentation mode.

It is awesome to see that the segmentations added in non-mpr, display in mpr. Furthermore, the segmentation tools appear to work very well in mpr as well! Good stuff!

I have found the following issues thus far:

1. There is a noticeable delay when clicking Add Segmentation. About 1-2 seconds (it does vary - and appears longer when running locally against orthanc). Is this expected? If feels a bit long to me.
2. If one selects any brush before creating a segmentation and then clicks and drags on an image, nothing occurs in the UI to suggest a segmentation must be created first. Indeed a warning message is output in the console, but I suspect most users will not have the console open. Perhaps the brush tools can be disabled until a segmentation is created? Note also that the following exception is also thrown in the console...
   ```
   BrushTool.ts:112 Uncaught Error: Not implemented yet
    at Rr.preMouseDownCallback (BrushTool.ts:112:13)
    at HTMLDivElement.fg (mouseDown.ts:60:38)
    at se (triggerEvent.ts:37:13)
    at mouseDownListener.ts:214:23
    at mt (mouseDownListener.ts:406:3)
    at HTMLDocument.dt (mouseDownListener.ts:245:7)
   ```
3. Creating a subsequent segmentation (i.e. via the Add New Segmentation menu item) does not appear to work.

4. Deleting the first (and only) segmentation throws an exception in the console and the segmentation remains in the side panel. However the segmentation (annotations) on the images are removed.
   ```
   Uncaught TypeError: Cannot set properties of undefined (setting 'isHydrated')
    at ht._setDisplaySetIsHydrated (SegmentationService.ts:1148:16)
    at ht.remove (SegmentationService.ts:1313:10)
    at onSegmentationDelete (PanelSegmentation.tsx:65:25)
    at onClick (SegmentationDropDownRow.tsx:67:17)
    at onClick (Dropdown.tsx:53:11)
    at Object.je (react-dom.production.min.js:52:317)
    at Qe (react-dom.production.min.js:52:471)
    at react-dom.production.min.js:53:35
    at Pr (react-dom.production.min.js:100:68)
    at Or (react-dom.production.min.js:101:380)
   ```	
5. There is some weird rounding that occurs when changing the outline and fill opacity for the segmentation.

ScreenHunter Sep  08 09 23

6. When using the sphere brush tool immediately after using the sphere ROI threshold tool, the annotation added by the brush tool does not "act" like a sphere - that is to say navigating forward and backward to neighbouring slices do NOT show the segmented (sphere) annotation as expected. Perhaps the brush tool shows that sphere is selected, but it is really circle that is selected?

https://github.com/OHIF/Viewers/assets/109477394/214b3aa1-b1e9-4f89-bfb8-c54accc50afd

7. I have occassionally (but rarely) seen tooltips for the various segmentation tool (icons). The tooltips should be shown on hover consistently in my opinion to give a better idea of what each tool is/does.

8. When adjusting the radius of any of the circle or sphere tools using the slider, fractions of a mm are displayed. However, I am not able to type in a fraction. Is this expected?

9. I noticed varying delays when first dragging the sphere brush tool. For example, doing a quick initial drag from one point to another draws the initial sphere, but then there is a delay to draw the remainder of the "path" to the second point.
Furthermore, the are occassions where some of the path between the two points is "skipped" (i.e. not filled in).

https://github.com/OHIF/Viewers/assets/109477394/aed014e5-1240-4933-98a6-5573b11d0fda

10. Download segmentation throws an exception and nothing is downloaded.
   ```
   Uncaught TypeError: U is not a function
    at downloadSegmentation (commandsModule.ts:181:7)
    at jt.runCommand (CommandsManager.ts:162:14)
    at onSegmentationDownload (PanelSegmentation.tsx:186:21)
    at onClick (SegmentationDropDownRow.tsx:81:23)
    at onClick (Dropdown.tsx:53:11)
    at Object.je (react-dom.production.min.js:52:317)
    at Qe (react-dom.production.min.js:52:471)
    at react-dom.production.min.js:53:35
    at Pr (react-dom.production.min.js:100:68)
    at Or (react-dom.production.min.js:101:380)
   ```	
11. The lock icon for a segment appears strangely.
ScreenHunter Sep  08 09 52

12. It is not obvious what clicking the lock icon is supposed to do.

13. Delete all segments (but leave the segmentation) and notice that the tools are still available and will draw annotations on the images. I would expect the tools to not even be selectable.

14. Should the measurement tools be available in this new mode? If they are not, then shouldn't we remove them from the top toolbar?

15. When switching back to a tool in the top toolbar, both the last active segmentation tool and the recently selected tool in the top tool bar show as active/selected.

two tools show as active


16. Add a sphere segment in non-mpr. Change the opacity to something more opaque. Enter and then exit mpr, the opacity now appears to be non-existent (i.e. transparent). Furthermore, adjusting the opacity for the segmentation has NO effect on the segment at all.

17. I was able to Create a Report (i.e. export the segmentation), however when attempting to hydrate the segmentation I see a gray box overlaying the images.


https://github.com/OHIF/Viewers/assets/109477394/ef3591f7-9aeb-4e71-a29d-805aed695941


18. It appears that the segmentation tools can only be applied to the viewport that was active when the segmentation was created. Attempting to apply the tools to another viewport throws the following exception.
   ```
   C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\tools\segmentation\BrushTool.ts:112 Uncaught Error: Not implemented yet
    at BrushTool.preMouseDownCallback (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\tools\segmentation\BrushTool.ts:112:1)
    at HTMLDivElement.mouseDown (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\eventDispatchers\mouseEventHandlers\mouseDown.ts:60:1)
    at triggerEvent (C:\dev\Viewers\node_modules\@cornerstonejs\core\src\utilities\triggerEvent.ts:37:1)
    at _doMouseDown (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\eventListeners\mouse\mouseDownListener.ts:214:1)
    at _doStateMouseDownAndUp (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\eventListeners\mouse\mouseDownListener.ts:406:1)
    at HTMLDocument._onMouseDrag (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\eventListeners\mouse\mouseDownListener.ts:245:1)
   ```	

https://github.com/OHIF/Viewers/assets/109477394/5d1abe4e-8f23-4bf5-adfd-2c8605b50c1b

19. In some cases, creating a segmentation, changes the window level of the active viewport.

https://deploy-preview-3632--ohif-dev.netlify.app/segmentation?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.3671.4754.298665348758363466150039312520


https://github.com/OHIF/Viewers/assets/109477394/01f35cd7-a477-400b-8fcb-607b438db7c0

sedghi:
Thanks for the thorough review 

Item 1 -> The initial delay is kind of inevitable right now since a volume need to be created from the stack viewport (similar to mpr)
Item 9 -> I don't see such delay or mis-behavior, can you try again?
Item 12 -> what do you propose? 
Item 15 -> I can't reproduce it anymore, maybe I fixed it in items 1-14 
Item 17 -> Can't reproduce it anymore
Item 19 -> it doesn't, it just moved the slice to the middle, i think i fixed that too

extensions/cornerstone/src/Viewport/Overlays/CustomizableViewportOverlay.tsx
```diff
@@ -254,7 +254,7 @@ function CustomizableViewportOverlay({
         instanceNumber,
       };
 
-      if (item.customizationType === 'ohif.overlayItem.windowLevel') {
+      if (item.customizationType === 'alireza') {
```
jbocce, line 257: Revert this?

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
```diff
@@ -46,6 +47,15 @@ class ImageOverlayViewerTool extends AnnotationDisplayTool {
     this._cachedOverlayMetadata = new Map();
   };
 
+  protected getReferencedImageId(viewport: IStackViewport | IVolumeViewport): string {
```
jbocce, line 50: Why can't we use the one from AnnotationDisplayTool?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -589,7 +585,20 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
       // otherwise, check if the hydrated segmentations are in the same FOR
       // as the primary displaySet, if so add the representation (since it was not there)
       const { id: segDisplaySetInstanceUID, type } = segmentation;
-      const segFrameOfReferenceUID = this._getFrameOfReferenceUID(segDisplaySetInstanceUID);
+      let segFrameOfReferenceUID = this._getFrameOfReferenceUID(segDisplaySetInstanceUID);
+
+      if (!segFrameOfReferenceUID) {
+        // if the segmentation does not have a FOR, we might check the
```
jbocce, line 591: It is confusing that the word segmentation appears twice in this sentence. Should this say instead, "if the segmentation display set does not..."?

jbocce: Also using FOR is confusing, I would stick with 'frame of reference uid' in the comment.

platform/ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -28,22 +60,31 @@ const Dropdown = ({ id, children, showDropdownIcon, list, titleClassName }) => {
...
+              ))}
+            </div>
+          )}
+          {title.length <= maxCharactersPerLine && (
+            <Typography className={itemsClassName}>{title}</Typography>
+          )}
+        </div>
       </div>
     );
   }, []);
```
jbocce, line 86: I think the `useCallback` dependency should included (at least) `maxCharactersPerLine`, but I suppose if it does not change then it might not matter.

platform/ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -118,9 +161,10 @@ const Dropdown = ({ id, children, showDropdownIcon, list, titleClassName }) => {
     </div>
   );
 };
-
 Dropdown.defaultProps = {
   showDropdownIcon: true,
+  maxTextLength: 10,
```
jbocce, line 166: Should this be `maxCharactersPerLine` instead?

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -281,15 +292,19 @@ class SegmentationService extends PubSubService {
...
-    segmentationId: string,
-    segmentIndex: number,
-    isLocked: boolean
-  ): void {
+  public setSegmentLocked(segmentationId: string, segmentIndex: number, isLocked: boolean): void {
     const suppressEvents = false;
     this._setSegmentLocked(segmentationId, segmentIndex, isLocked, suppressEvents);
   }
 
+  public toggleSegmentLocked(segmentationId: string, segmentIndex: number): void {
```
jbocce, line 300: Add to docs?

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
```diff
@@ -2,7 +2,7 @@ import calculateViewportRegistrations from './calculateViewportRegistrations';
 
 // [ {
 //   synchronizerId: string,
-//   viewports: [ { viewportId: string, renderingEngineId: string, index: number } , ...]
+//   viewports: [ { viewportId: number, renderingEngineId: string, index: number } , ...]
```
jbocce, line 5: I realize this is in a comment, but I think `viewportId` is a string.

extensions/cornerstone-dicom-seg/src/panels/Toolbox/ThresholdSettingsPreset.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const handleRangeChange = newRange => {
+    const selectedOption = options.find(o => o.value === selectedPreset);
+
+    if (newRange[0] === selectedOption.range[0] && newRange[1] === selectedOption.range[1]) {
+      return;
+    }
+
+    onRangeChange(newRange);
+
+    const updatedOptions = options.map(o => {
```
jbocce, line 45: Hmmm... I believe it would be safest to do this as part of the update function that can be passed to `setOptions`?

```
setOptions((options) => {
  return options.map(...)
});
```

extensions/cornerstone-dicom-seg/src/panels/Toolbox/ThresholdSettingsPreset.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const selectedPresetRange = options.find(ds => ds.value === selectedPreset).range;
+
+  return (
+    <div>
+      <div className="bg-secondary-light h-[1px]"></div>
+      <div className="mt-1 text-[13px] text-white">Threshold</div>
+      <div className="mt-1 w-1/2">
+        <Select
+          isClearable={false}
+          onChange={handlePresetChange}
```
jbocce, line 67: Where is `handlePresetChange`?

extensions/cornerstone-dicom-seg/src/panels/Toolbox/ThresholdSettingsPreset.tsx
```diff
@@ -0,0 +1,88 @@
...
+    range: [-770, 1730] as [number, number],
+  },
+  {
+    value: 'Brain',
+    label: 'Brain',
+    range: [0, 80] as [number, number],
+  },
+];
+
+function ThresholdSettings({ onRangeChange }) {
```
jbocce, line 32: Is this referenced somewhere?

extensions/cornerstone-dicom-seg/src/panels/SegmentationToolbox.tsx
```diff
@@ -0,0 +1,372 @@
+import React, { useCallback, useEffect, useState, useReducer } from 'react';
+import { AdvancedToolbox, InputDoubleRange, useViewportGrid } from '@ohif/ui';
+import { Types } from '@ohif/extension-cornerstone';
+import { utilities } from '@cornerstonejs/tools';
+
+const { segmentation: segmentationUtils } = utilities;
+
+const ACTIONS = {
+  SET_BRUSH_SIZE: 'SET_BRUSH_SIZE',
```
jbocce, line 9: Is this needed anymore? I don't think it is being used.

platform/ui/src/components/AdvancedToolbox/AdvancedToolbox.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { useState, useEffect } from 'react';
+import classnames from 'classnames';
+import { PanelSection, Icon, Tooltip } from '../../components';
+import ToolSettings from './ToolSettings';
+
+const AdvancedToolbox = ({ title, items }) => {
+  const [isActive, setIsActive] = useState(null);
```
jbocce, line 7: `isActive` sounds like a boolean to me, yet I believe this state is for the name of the active item. So consider using `[activeItemName, setActiveItemName]` instead.

platform/ui/src/components/AdvancedToolbox/ToolSettings.tsx
```diff
@@ -0,0 +1,78 @@
...
+      });
+    });
+
+    return buttons;
+  };
+
+  return (
+    <div className="space-y-2 py-2 text-white">
+      {options?.map(option => {
+        if (option.type === 'range') {
```
jbocce, line 26: Consider making `range`, `radio` and `custom` enums or constants maybe?

extensions/cornerstone-dicom-seg/src/panels/SegmentationToolbox.tsx
```diff
@@ -0,0 +1,372 @@
...
+
+  return (
+    <AdvancedToolbox
+      title="Segmentation Tools"
+      items={[
+        {
+          name: 'Brush',
+          icon: 'icon-tool-brush',
+          disabled: !toolsEnabled,
+          active: state.activeTool === 'CircularBrush' || state.activeTool === 'SphereBrush',
```
jbocce, line 215: Consider introducing constants for the following (note that this list is not exhaustive, so please check for more constants to add):

-'CircularBrush'
-'SphereBrush'
-'Brush'
-'Circle'
-'ThresholdCircularBrush'
- etc...

modes/segmentation/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
```
jbocce, line 1: Do we need this file considering it is not consistent with that at the root?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -1381,7 +1391,7 @@ export default class HangingProtocolService extends PubSubService {
     const matchingScores = [];
     let highestSeriesMatchingScore = 0;
 
-    console.log('ProtocolEngine::matchImages', studyMatchingRules, seriesMatchingRules);
+    // console.log('ProtocolEngine::matchImages', studyMatchingRules, seriesMatchingRules);
```
jbocce, line 1394: Do we want to uncomment this?

platform/ui/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -1,150 +1,63 @@
...
+        );
       })}
     </div>
   );
 };
 
 ButtonGroup.propTypes = {
-  children: PropTypes.node.isRequired,
+  buttons: PropTypes.arrayOf(PropTypes.object).isRequired,
+  orientation: PropTypes.oneOf(['horizontal', 'vertical']),
```
jbocce, line 55: To avoid using strings, consider declaring this as you did below`PropTypes.oneOf(Object.values(ButtonEnums.orientation))`.

platform/ui/src/utils/getMaxDigits.ts
```diff
@@ -0,0 +1,7 @@
+const getMaxDigits = (maxValue: number, step: number) => {
```
jbocce, line 1: Please add a comment for this function.

extensions/cornerstone-dicom-seg/src/commandsModule.ts
```diff
@@ -0,0 +1,378 @@
...
+
+      const segmentation = segmentationService.getSegmentation(segmentationId);
+
+      if (!segmentation) {
+        throw new Error('No segmentation found');
+      }
+
+      const { label } = segmentation;
+      const SeriesDescription =
+        promptResult.value !== undefined && promptResult.value !== ''
```
jbocce, line 323: How about just `promptResult.value` ?

platform/ui/src/components/InputDoubleRange/InputDoubleRange.css
```diff
@@ -0,0 +1,5 @@
+.input-range-thumb-design {
+  border: 2px solid #090c29;
+  border-radius: 50%;
+  background: #5acce6;
```
jbocce, line 4: Can we use `@apply bg-primary-light` here instead?

platform/ui/src/components/InputDoubleRange/InputDoubleRange.css
```diff
@@ -0,0 +1,5 @@
+.input-range-thumb-design {
+  border: 2px solid #090c29;
```
jbocce, line 2: `@apply border-[2px] border-solid border-primary-dark`?

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
```diff
@@ -0,0 +1,219 @@
...
+      <span className={classNames(labelClassName ?? 'text-white')}>
+        {val}
+        {unit}
+      </span>
+    );
+  };
+
+  useEffect(() => {
+    return () => {
+      window.removeEventListener('mousemove', handleMouseMove);
```
jbocce, line 94: I might be wrong but I think we might have a slight memory leak here - ableit unlikely to occur. During a drag of either thumb, `InputDoubleRange` is rerendered constantly and as such a "new" `handleGlobalMouseUp` and `handleMouseMove` functions are created for each render. When these `removeEventListener` functions are called, the referenced `handleGlobalMouseUp` and `handleMouseMove` are NOT the ones that were added in `handleMouseDown`. However this would only be a problem if the component were unmounted during a drag. So maybe we don't need to worry about it too much, but I think we can solve the problem by simply using `useCallback` for each listner function.

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const rangeValuePercentageStart = ((rangeValue[0] - minValue) / (maxValue - minValue)) * 100;
+  const rangeValuePercentageEnd = ((rangeValue[1] - minValue) / (maxValue - minValue)) * 100;
+
+  return (
+    <div className={`flex items-center space-x-2${containerClassName ?? ''}`}>
+      {showLabel && LabelOrEditableNumber(rangeValue[0], 0)}
+      <div
+        className="relative flex h-10 w-full items-center"
+        onMouseDown={handleMouseDown}
+        onMouseMove={handleMouseMove}
```
jbocce, line 178: I don't think we need to add the `onMouseMove` and `onMouseUp` here since they are being added on the window.

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
```diff
@@ -0,0 +1,219 @@
...
+      if (selectedThumb === 1 && clampedValue <= rangeValue[0]) {
+        return;
+      }
+
+      // Update the correct values in the rangeValue array
+      const updatedRangeValue = [...rangeValue];
+      updatedRangeValue[selectedThumb] = clampedValue;
+      setRangeValue(updatedRangeValue);
+
+      // No need to call onChange here, since we're updating local state+
```
jbocce, line 155: I am confused by this comment...if there is no need to call onChange, then why are we doing it on the next line?

platform/ui/src/components/InputRange/InputRange.css
```diff
@@ -4,23 +4,32 @@ input[type='range'] {
 
 input[type='range']::-webkit-slider-thumb {
   -webkit-appearance: none;
-  border: none;
+  border: 2px solid #090c29;
```
jbocce, line 7: Can we use @apply in here?

platform/ui/src/components/InputRange/InputRange.tsx
```diff
@@ -38,72 +42,76 @@ const InputRange: React.FC<{
...
+          value={rangeValue}
+          className={`h-[3px] appearance-none rounded-md ${inputClassName ?? ''}`}
+          style={{
+            background: `linear-gradient(to right, #5acce6 0%, #5acce6 ${rangeValuePercentage}%, #3a3f99 ${rangeValuePercentage}%, #3a3f99 100%)`,
+          }}
+          onChange={handleChange}
+          id="myRange"
+          step={step}
+        />
+        <div style={{ width: labelWidth }}>
```
jbocce, line 111: If `showLabel` is `false`, the `div` will be empty so do we still want to have this empty `div` occupy "width" space?

platform/ui/src/components/PanelSection/PanelSection.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react';
+import { Icon } from '@ohif/ui';
+import PropTypes from 'prop-types';
+
+const PanelSection = ({ title, children, actionIcons = [] }) => {
```
jbocce, line 5: Add `children` and `actionIcons` to propTypes below.

platform/ui/src/components/PanelSection/PanelSection.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [isChildrenVisible, setChildrenVisible] = useState(true);
```
jbocce, line 6: `areChildrenVisible` reads better in English, but it is not a big deal.

platform/ui/src/components/SegmentationGroupTable/__stories__/segmentationGroupTable.stories.mdx
```diff
@@ -55,7 +54,7 @@ SegmentationGroupTable is a component that renders the SegmentationGroupTables.
           segments: [
             null,
             {
-              label: 'Segment 1',
+              label: 'Segment 1 Segment 1Segment 1Segment 1Segment',
```
jbocce, line 57: If we are trying to show how a long label is handled, then I would suggest changing this to something like 'This is a very long segment label'.

platform/ui/src/components/SegmentationGroupTable/SegmentationGroupTable.tsx
```diff
@@ -31,71 +41,123 @@ const SegmentationGroupTable = ({
...
-                key={id}
-                label={label}
+    <div className="flex min-h-0 flex-col bg-black text-[13px] font-[300]">
+      <PanelSection
+        title="Segmentation"
+        actionIcons={
+          activeSegmentation && [
+            {
+              name: 'settings-bars',
+              onClick: () => setIsConfigOpen(!isConfigOpen),
```
jbocce, line 81: I think it is safest to use `setIsConfigOpen(isOpen => !isOpen)`

platform/ui/src/components/SegmentationGroupTable/SegmentationGroupTable.tsx
```diff
@@ -31,71 +41,123 @@ const SegmentationGroupTable = ({
...
+      activeSegmentationIdToSet = segmentations[0].id;
+    }
+
+    // If there is no segmentation, set the active segmentation to null
+    if (segmentations?.length === 0) {
+      activeSegmentationIdToSet = null;
+    }
+
+    setActiveSegmentationId(activeSegmentationIdToSet);
+  }, [segmentations, activeSegmentationId]);
```
jbocce, line 67: I don't think `activeSegmentationId` should be a dependency for this `useEffect`.

extensions/cornerstone/src/tools/ImageOverlayViewerTool.tsx
sedghi: Since that one requires worldPos and other coordinates, which we don't have here

platform/ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -5,16 +5,48 @@ import classnames from 'classnames';
...
+    }
+    return acc;
+  }, 0);
+
+  maxCharactersPerLine = maxCharactersPerLine ?? longestTitle;
+
   const DropdownItem = useCallback(({ id, title, icon, onClick }) => {
+    // Split the title into lines of length maxCharactersPerLine
+    const lines = [];
+    for (let i = 0; i < title.length; i += maxCharactersPerLine) {
```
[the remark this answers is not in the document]
sedghi: I mean what choice do we have, it is either a very wide dropdown as a result of only one of the items being longs or break like this. In practice I have not seen this happening

extensions/cornerstone-dicom-seg/src/panels/Toolbox/ThresholdSettingsPreset.tsx
sedghi: I forgot to remove this file, it is outdated and we don't need it

modes/segmentation/src/initToolGroups.ts
```diff
@@ -0,0 +1,206 @@
...
+        bindings: [{ mouseButton: Enums.MouseBindings.Auxiliary }],
+      },
+      {
+        toolName: toolNames.Zoom,
+        bindings: [{ mouseButton: Enums.MouseBindings.Secondary }],
+      },
+      { toolName: toolNames.StackScrollMouseWheel, bindings: [] },
+    ],
+    passive: [
+      { toolName: toolNames.CircleScissors },
```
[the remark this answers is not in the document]
sedghi: One is for deafult and one is for MPR

extensions/cornerstone-dicom-seg/src/commandsModule.ts
sedghi: but we don't want ' '

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
sedghi: Thanks for noticing this, but as you said let's worry when we have a bug for this

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
sedghi: No they should be there, they decide which thumb is dragged and move it

modes/segmentation/src/initToolGroups.ts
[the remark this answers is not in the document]
jbocce: I understand, but I just thought maybe we could maybe introduce a constant and use the constant twice instead of the whole chunk of code. No big deal either way.

extensions/cornerstone-dicom-seg/src/commandsModule.ts
jbocce: The empty string ('') is falsy, so I think we are ok, no? https://developer.mozilla.org/en-US/docs/Glossary/Falsy

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
jbocce: Ok. I find it odd that we need both these AND the listeners on the window, but perhaps I am missing something. In fact, I think maybe the mouse move listener on the window has a stale/null selected thumb and might be doing nothing. Maybe only the mouse up listener on the window is useful. Anyway please have a another look.

platform/ui/src/components/SegmentationGroupTable/SegmentationGroupTable.tsx
```diff
@@ -1,28 +1,38 @@
...
+import NoSegmentationRow from './NoSegmentationRow';
+import AddSegmentRow from './AddSegmentRow';
+import SegmentationGroupSegment from './SegmentationGroupSegment';
 
 const SegmentationGroupTable = ({
   segmentations,
+  // segmentation initial config
+  segmentationConfig,
+  // UI show/hide
+  disableEditing,
```
jbocce, line 15: Please ensure the propTypes below are up-to-date.

platform/ui/src/components/SegmentationGroupTable/SegmentationGroupSegment.tsx
```diff
@@ -21,200 +21,159 @@ const SegmentItem = ({
...
-        onMouseLeave={() => setIsSegmentIndexHovering(false)}
+        className={classnames('bg-primary-dark group/number grid w-[32px] place-items-center', {
+          'bg-primary-light border-primary-light rounded-l-[4px] border text-black': isActive,
+          'border-primary-dark border': !isActive,
+        })}
+        onMouseEnter={() => setIsNumberBoxHovering(true)}
+        onMouseLeave={() => setIsNumberBoxHovering(false)}
       >
-        {isSegmentIndexHovering && showSegmentDelete ? (
+        {isNumberBoxHovering && showDelete ? (
```
jbocce, line 49: Not a big deal, but I believe instead of using `onMouseEnter` and `onMouseLeave` to keep track of the hover we could also add a `group` class to the parent `div` and then have class names `hidden group-hover:block` on the `Icon`.

platform/ui/src/components/SegmentationGroupTable/SegmentationGroupSegment.tsx
```diff
@@ -21,200 +21,159 @@ const SegmentItem = ({
...
-            <div className="px-1">
+          <div className="flex items-center py-1 hover:cursor-pointer">{label}</div>
+        </div>
+        <div
+          className={classnames(
+            'absolute right-0 top-0 flex flex-row-reverse rounded-lg pr-[8px] pt-[3px]',
+            {}
+          )}
+        >
+          {!isVisible && !isRowHovering && (
```
jbocce, line 97: It is a bit confusing that each icon is rendered here when NOT hovering and then rendered on line 131 when hovering. It would be great if we only had to "look" in one place for each icon. But if it is too much trouble, then that is ok.

platform/ui/src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
jbocce: Well because it is an edge case we should leave it, but I suspect there would be something we could do with css https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_text/Wrapping_breaking_text

platform/ui/src/components/SegmentationGroupTable/SegmentationDropDownRow.tsx
```diff
@@ -0,0 +1,138 @@
+import React from 'react';
+import { Select, Icon, Dropdown } from '../../components';
+import PropTypes from 'prop-types';
+
+function SegmentationDropDownRow({
```
jbocce, line 5: Update propTypes below.

jbocce (changes requested):
It is looking good. Lots of 'neat' GUI controls.

Have a look at my comments. I will now do some more testing.

jbocce:
I believe the former issues are handled for the most part. I think for item 17, perhaps the reconstruction for some series/studies is just slow with Orthanc and the gray box appears for a "while". Here are some more things I found.

a. After hydrating a saved/reported segmentation, the thumbnails in the side study browser go blank.
b. The stack sync tool appears to be broken. For example, when navigating one of two axial series on screen and the tool is activated only one scrolls. Actually, it appears that the even though I click the tool to activate it, it appears not selected/active.
c. After deleting the last segmentation, the brush tools are still active and using them throws exceptions in the console.
d. After deleting all segments, the last tool is still active and it can be used.
e. I noticed that setting the fill opacity to 0 makes the outline disappear (i.e. I think it goes transparent). Is this expected? I would have expected the outline to remain but with no fill.
f. So I still see the issue where the window level appears to change when adding a segmentation (i.e. item 19 from before) and after having a good look at the images and the various stack scroll bars/sliders) the middle slice is NOT being moved to from what I can tell. If anything I would say the window level of the middle image is being applied yet the current/last image remains on screen.

https://github.com/OHIF/Viewers/assets/109477394/ac247769-0b8b-405a-89d8-84f8041dd56c

platform/ui/src/components/SegmentationGroupTable/SegmentationGroupSegment.tsx
sedghi: seems like there is another bug

CleanShot 2023-09-14 at 13 10 34

jbocce:
The stack sync tool appears to work now. However, once a segmentation is added/drawn, the scroll syncing stops. Is this expected because now we have a volume viewport?

jbocce:
I still see the problem where after deleting the last segmentation, the brush tools are still active and using them throws exceptions in the console.

jbocce:
I slide the fill opacity all the way to the left and the label says 4%. Is this expected?

image

jbocce:
I am still seeing some weirdness with the threshold tool slider. Consider the following steps...

1. Drag the far left thumb of the slider up and over to the left. Drag the mouse over the word Mode in the panel above.
2. Notice it is selecting/highlighting text during the drag this is not desirable.
3. Release the drag when over the word Mode. Notice that text is still highlighted.
4. Try to drag the left thumb again. Notice the weird dragging of text that is happening.
5. Release the mouse button and continue moving the mouse and notice that we are "stuck" in drag mode.

https://github.com/OHIF/Viewers/assets/109477394/0b116551-5251-4c11-a156-cf3b207f6154

jbocce:
After saving a seg report, if the newly saved series/report is double clicked. The following exception is thrown in the console...

Uncaught TypeError: segDisplaySet.instance is not iterable (cannot read property Symbol(Symbol.iterator))
    at _getReferencedDisplaySetMetadata (C:\dev\Viewers\extensions\cornerstone-dicom-seg\src\viewports\OHIFCornerstoneSEGViewport.tsx:379:1)
    at OHIFCornerstoneSEGViewport (C:\dev\Viewers\extensions\cornerstone-dicom-seg\src\viewports\OHIFCornerstoneSEGViewport.tsx:68:1)
    at renderWithHooks (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:14985:1)
    at updateFunctionComponent (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:17356:1)
    at mountLazyComponent (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:17677:1)
    at beginWork (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:19055:1)
    at HTMLUnknownElement.callCallback (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:3994:1)
    at invokeGuardedCallback (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:4056:1)
    at beginWork$1 (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:23964:1)
    
 

https://github.com/OHIF/Viewers/assets/109477394/059ce55b-5300-4f57-99f9-0e55ec6cf12e

jbocce:
Consider removing the rectangle tool for segmentation mode.

jbocce:
Deleting a hyrdrated seg should remove the corresponding highlight in the study list/thumbnail.

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
```diff
@@ -0,0 +1,225 @@
...
+  const rangeValuePercentageStart = ((rangeValue[0] - minValue) / (maxValue - minValue)) * 100;
+  const rangeValuePercentageEnd = ((rangeValue[1] - minValue) / (maxValue - minValue)) * 100;
+
+  return (
+    <div className={`flex items-center space-x-2${containerClassName ?? ''}`}>
+      {showLabel && LabelOrEditableNumber(rangeValue[0], 0)}
+      <div
+        className="relative flex h-10 w-full items-center"
+        onMouseDown={handleMouseDown}
+        // onMouseMove={handleMouseMove}
```
jbocce, line 183: Remove if not needed.

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
```diff
@@ -0,0 +1,225 @@
...
+    } else if (distanceToEndThumb < 10) {
+      selectedThumbRef.current = 1;
+    }
+
+    window.addEventListener('mousemove', handleMouseMove);
+    window.addEventListener('mouseup', handleGlobalMouseUp);
+  };
+
+  const handleMouseMove = e => {
+    console.debug('🚀 ~ selectedThumb:', selectedThumbRef.current);
```
jbocce, line 134: I don't think we want this during a drag.

jbocce (changes requested):
Just a few issues to address.

sedghi:
> The stack sync tool appears to work now. However, once a segmentation is added/drawn, the scroll syncing stops. Is this expected because now we have a volume viewport?

Yes it is because it is converted to volume viewport, ideally we should have viewportSync instead of `stack` viewport sync which handles both stack and volume viewport

> I still see the problem where after deleting the last segmentation, the brush tools are still active and using them throws exceptions in the console.

Probably not built correctly?

https://github.com/OHIF/Viewers/assets/7490180/47bfe02d-743d-433e-b3d1-2844a5c1ec17

> I slide the fill opacity all the way to the left and the label says 4%. Is this expected?

We don't have proper 0 right now as it removes the outline as we discussed, so I just put it as min 4 for now

>

jbocce:
> I still see the problem where after deleting the last segmentation, the brush tools are still active and using them throws exceptions in the console.

This looks fixed now - thanks. However, the window level tool appears to be active, yet there is no indication that it is active. (i.e. the tool  button does not indicate that it is selected).

jbocce:
It appears that adding segments to a different oriented series is NOT support or at least throws an exceptions. Is there a better way to handle this? Perhaps it is something we address later, but I figured it would be worth mentioning here at least in case there was something that we could easily do (perhaps disable the tools or something).

```
Uncaught Error: Only source data the same dimensions/size/orientation as the segmentation currently supported.
    at o.Ka (fillSphere.ts:117:11)
    at o.applyActiveStrategy (BaseTool.ts:85:39)
    at Rr._endCallback (BrushTool.ts:383:10)
    at se (triggerEvent.ts:37:13)
    at HTMLDocument.ct (mouseDownListener.ts:327:5)
```
https://github.com/OHIF/Viewers/assets/109477394/0fa50225-d96a-410c-b66e-c20ce6481ebd

jbocce:
Looks like there is a failing cypress test. Please have a look.

jbocce:
I noticed that there is an exception thrown when saving measurements that does not appear to exist in master. However, there appears to be an exception thrown when hydrating measurements in master. I will log a bug for that.

jbocce:
Hmm. Not sure why I didn't see this earlier, but when deleting a segmentation I am getting the following exception 

Perhaps it is only seen when displaying multiple series/orientations???

```
Uncaught TypeError: viewport.removeVolumeActors is not a function
    at removeLabelmapFromElement (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\tools\displayTools\Labelmap\removeLabelmapFromElement.ts:22:1)
    at _removeLabelmapFromToolGroupViewports (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\tools\displayTools\Labelmap\labelmapDisplay.ts:485:1)
    at Object.removeSegmentationRepresentation (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\tools\displayTools\Labelmap\labelmapDisplay.ts:116:1)
    at _removeSegmentation (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\stateManagement\segmentation\removeSegmentationsFromToolGroup.ts:74:1)
    at C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\stateManagement\segmentation\removeSegmentationsFromToolGroup.ts:57:1
    at Module.removeSegmentationsFromToolGroup (C:\dev\Viewers\node_modules\@cornerstonejs\tools\src\stateManagement\segmentation\removeSegmentationsFromToolGroup.ts:56:1)
    at C:\dev\Viewers\extensions\cornerstone\src\services\SegmentationService\SegmentationService.ts:1993:1
    at SegmentationService._removeSegmentationFromCornerstone (C:\dev\Viewers\extensions\cornerstone\src\services\SegmentationService\SegmentationService.ts:1981:1)
```


https://github.com/OHIF/Viewers/assets/109477394/705d24e9-f9f4-4152-a322-dfb53acd216c

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
```diff
@@ -0,0 +1,223 @@
...
+  const rangeValuePercentageStart = ((rangeValue[0] - minValue) / (maxValue - minValue)) * 100;
+  const rangeValuePercentageEnd = ((rangeValue[1] - minValue) / (maxValue - minValue)) * 100;
+
+  return (
+    <div className={`flex select-none items-center space-x-2 ${containerClassName ?? ''}`}>
+      {showLabel && LabelOrEditableNumber(rangeValue[0], 0)}
+      <div
+        className="relative flex h-10 w-full items-center"
+        onMouseDown={handleMouseDown}
+        onMouseUp={handleMouseUp}
```
jbocce, line 182: It is great to see that we don't need the onMouseMove here any more. Do we still need the onMouseUp?

platform/core/src/services/ViewportGridService/ViewportGridService.ts
```diff
@@ -74,6 +74,11 @@ class ViewportGridService extends PubSubService {
     return this.serviceImplementation._getState();
   }
 
+  public getActiveViewportId() {
```
jbocce, line 77: Please add this to the docs if you haven't already done so. Thanks.

extensions/cornerstone-dicom-seg/src/panels/SegmentationToolbox.tsx
```diff
@@ -0,0 +1,404 @@
...
+              value: state.activeTool,
+              values: [
+                { value: TOOL_TYPES.THRESHOLD_CIRCULAR_BRUSH, label: 'Circle' },
+                { value: TOOL_TYPES.THRESHOLD_SPHERE_BRUSH, label: 'Sphere' },
+              ],
+              onChange: value => setToolActive(value),
+            },
+            {
+              type: 'custom',
+              id: 'threshold-range',
```
jbocce, line 358: Consider using `'segmentation-threshold-range'` to guarantee better uniqueness. I have seen browsers do crazy things when ids are no unique.

jbocce (changes requested):
Just a few issues to address.

jbocce:
The following exception is thrown when attempting to relaunch Segmentation mode after adding a segmentation in the last session.

```
ToolGroupService.ts:166 Uncaught (in promise) Error: ToolGroup default already exists
    at ot.createToolGroup (ToolGroupService.ts:166:13)
    at ot.createToolGroupAndAddTools (ToolGroupService.ts:188:28)
    at initToolGroups.ts:53:20
    at g (initToolGroups.ts:78:3)
    at Object.onModeEnter (index.tsx:58:7)
    at Mode.tsx:334:13
    at Mode.tsx:394:5
    at Ul (react-dom.production.min.js:262:359)
    at t.unstable_runWithPriority (scheduler.production.min.js:18:343)
    at qa (react-dom.production.min.js:122:325)
```


https://github.com/OHIF/Viewers/assets/109477394/f316e048-02fa-481d-bec5-4e62c5aa6c60

platform/ui/src/components/InputDoubleRange/InputDoubleRange.tsx
```diff
@@ -0,0 +1,222 @@
...
+  };
+
+  useEffect(() => {
+    return () => {
+      window.removeEventListener('mousemove', handleMouseMove);
+      window.removeEventListener('mouseup', handleGlobalMouseUp);
+    };
+  }, []);
+
+  const handleMouseUp = () => {
```
jbocce, line 102: This I think is no longer referenced so it can be removed.

platform/docs/docs/platform/services/ui/viewport-grid-service.md
```diff
@@ -32,6 +32,7 @@ is expected to support, [check out it's interface in `@ohif/core`][interface]
 | `reset()`                                                             | Resets the default states                           |
 | `getNumViewportPanes()`                                               | Gets the number of visible viewport panes           |
 | `getLayoutOptionsFromState(gridState)`                                | Utility method that produces a `ViewportLayoutOptions` based on the passed in state|
+| `getActiveViewportId()`                                | Returns the viewportId of the active viewport in the grid|
```
jbocce, line 35: Add a space between viewport and id.

jbocce (changes requested):
Just a couple more things.

jbocce:
Choosing a tool/item from either split button menu (i.e. measurement or more tools) no longer shows the tool as selected. Also the mpr button does not render as selected when in mpr mode.

https://github.com/OHIF/Viewers/assets/109477394/9e36aa15-4073-4df0-9cde-8bb49f2f82db

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -94,8 +94,9 @@ export default class ToolbarService extends PubSubService {
             commandsManager.runCommand(commandName, commandOptions, context);
           });
 
-          // only set the primary tool if no error was thrown
-          this.state.primaryToolId = itemId;
+          // only set the primary tool if no error was thrown and if the itemId is not
+          // undefined set the first tool as the primary tool
```
jbocce, line 98: It is not clear from the comment why we are setting the primaryToolId the way we are. Why might `itemId` be undefined? Also the comment suggests that the first tool is used if `itemId` is not undefined.

sedghi: for backward compatibility i put it like this, we don't really need itemId

extensions/default/src/Toolbar/ToolbarButtonWithServices.tsx
```diff
@@ -0,0 +1,75 @@
+import { ToolbarButton } from '@ohif/ui';
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+
+function ToolbarSplitButtonWithServices({
```
jbocce, line 5: Shouldn't this be `ToolbarButtonWithServices`?

jbocce:
I am still seeing some weirdness with the toolbar buttons...

When stack image sync is active, I sometimes see this... 
image
and sometimes this...
image

extensions/default/src/Toolbar/ToolbarSplitButtonWithServices.tsx
```diff
@@ -0,0 +1,192 @@
...
+    toggles: {},
+    groups: {},
+  });
+
+  const [state, setState] = useState({
+    primary,
+    items: getSplitButtonItems(items).filter(item =>
+      isRadio && !isAction ? item.id !== primary.id : true
+    ),
+    isHovering: false,
```
jbocce, line 57: Are `isHovering` and `isExpanded` actually used somewhere?

extensions/default/src/Toolbar/ToolbarSplitButtonWithServices.tsx
```diff
@@ -0,0 +1,192 @@
...
+        />
+      )}
+    />
+  );
+}
+
+ToolbarSplitButtonWithServices.propTypes = {
+  isRadio: PropTypes.bool,
+  isAction: PropTypes.bool,
+  bState: PropTypes.shape({
```
jbocce, line 152: Please remove and/or update this list such that it is accurate.

jbocce (approved):
Some minor comments, but otherwise approved.

Thanks for the tremendous effort on this.

PR #3635, opened 2023-09-06
fix(hotkeys): preserve hotkeys if changed, and reduce re-rendering

### LINKED ISSUE #3598: [Bug] User preference hotkeys not reloaded from local storage
### Describe the Bug

You can change hotkeys and save them, and they persist for the duration of the tab. However, on refresh (and likely on going from one Mode to another) the hotkeys get reset to default.

I think the problem is [here](https://github.com/OHIF/Viewers/blob/daf734e09e843f5474301124aab0429289436878/platform/app/src/routes/Mode/Mode.tsx#L173).

The change was good, insofar as if you add a hotkey object to a Mode, it will use the name member to set and retrieve the hotkeys. However, the default behaviour is to _write_ the hotkeys to 'hotkey-definitions', but read them from 'hotkey-definitions-v2'.

Fixed locally by changing above line to the expected 'hotkey-definitions'. 

@wayfarer3130 made the change, and may be able to add more context.

### Steps to Reproduce

1. Go to a [viewer](https://viewer.ohif.org/).
2. Open any BasicViewer.
3. Change a hotkey in Preferences.
4. Save.
5. Reload the page.
6. Note the hotkeys are back to defaults.

### The current behavior

No hotkeys are restored from LS.

### The expected behavior

User preference hotkeys are restored from LS.

### OS

Windows 10

### Node version

16.15.0-slim

### Browser

Chrome 115.0.5790.110

sedghi:
### Context

fixes #3598 

Hotkeys were not getting restored after refresh. 

There were too many re-renderings happening because of modals and useTranslation

### Changes & Results

Fixes the re-renderings as shown below by refactoring and not changing language if the same

Current behaviour 

https://github.com/OHIF/Viewers/assets/7490180/864562db-997d-4178-9a08-dd657a976edd

After this PR
- No re-rendering -> No `search for studies` 

### Testing

- Changed hotkeys in the user preferences should be preserved after refresh
- the app should feel faster after save preferences (since there is no re-renderings)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/platform/app/src/routes/Mode/Mode.tsx
+++ b/platform/app/src/routes/Mode/Mode.tsx
@@ -164,7 +164,7 @@ export default function ModeRoute({
 
   // Preserve the old array interface for hotkeys
   const hotkeys = Array.isArray(hotkeyObj) ? hotkeyObj : hotkeyObj?.hotkeys;
-  const hotkeyName = hotkeyObj?.name || 'hotkey-definitions-v2';
+  const hotkeyName = hotkeyObj?.name || 'hotkey-definitions';
 
   // An undefined dataSourceName implies that the active data source that is already set in the ExtensionManager should be used.
   if (dataSourceName !== undefined) {
--- a/extensions/default/src/ViewerLayout/ViewerHeader.tsx
+++ b/extensions/default/src/ViewerLayout/ViewerHeader.tsx
@@ -1,0 +1,118 @@
+import React from 'react';
+import { useNavigate } from 'react-router-dom';
+import { useTranslation } from 'react-i18next';
+import { useLocation } from 'react-router';
+
+import { ErrorBoundary, UserPreferences, AboutModal, Header, useModal } from '@ohif/ui';
+import i18n from '@ohif/i18n';
+import { hotkeys } from '@ohif/core';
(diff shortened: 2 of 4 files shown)


platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -164,7 +164,7 @@ export default function ModeRoute({
 
   // Preserve the old array interface for hotkeys
   const hotkeys = Array.isArray(hotkeyObj) ? hotkeyObj : hotkeyObj?.hotkeys;
-  const hotkeyName = hotkeyObj?.name || 'hotkey-definitions-v2';
+  const hotkeyName = hotkeyObj?.name || 'hotkey-definitions';
```
jbocce, line 167: Does this affect anyone migrating from v2?

sedghi: I don't think that was supposed to be v2, prob just a typo

jbocce (approved):

jbocce:
Looks great. Great catch on the rerendering!

PR #3645, opened 2023-09-11
Add codespell config + github action, run and fix a good number of typos

yarikoptic:
- I also added a fix to the filename (two instances) : continuous-integration.md 
- I was not sure how to fix wording in following snippet which did not sound proper English construct to me (but I am not "native" ;))

```shell
❯ git grep -A3 :StudyBrowser -- platform/docs/docs/platform/services/ui/customization-service.md
platform/docs/docs/platform/services/ui/customization-service.md:* Name: 'class:StudyBrowser'
platform/docs/docs/platform/services/ui/customization-service.md-* Attributes:
platform/docs/docs/platform/services/ui/customization-service.md-** `true` for the is active true text color
platform/docs/docs/platform/services/ui/customization-service.md-** `false` for the is active false text color.
``` 

With this action/config you should spot now whenever someone is about to introduce a typo.  Workflow would fail but would not automagically fix anything -- could be done locally by running [codespell](https://github.com/codespell-project/codespell) directly or just manually.


sedghi (approved):
Thanks this looks great!

.codespellrc
```diff
@@ -0,0 +1,6 @@
+[codespell]
+skip = .git,*.pdf,*.svg,yarn.lock,*.min.js,locales
```
sedghi, line 2: oh does this mean it will run on js, ts, jsx, and tsx files too? How powerful is it to not mess around with real js code and only fix typos in the function documents

yarikoptic: well, you can see how well it did so far ;)

indeed JS and alike are typically trickier but not entirely hopeless -- worse comes to worse you can disable for those if starts bothering.  codespell doesn't do any programming language specific parsing -- so can't tell code from documentation. Just extracts words and finds typos in the words.

sedghi: Great explanation thanks again!

sedghi (approved):

PR #3646, opened 2023-09-12
Fix(health imaing): studies not loading from healthimaging if imagepositionpatient is missing

IbrahimCSAE:
### Context

The bug was introduced in the new feature "DisplaySet Messages." It didn't have proper error handling, causing promises not to settle, leading to infinite loading in the viewer.

I added proper error handling so this issue doesn't occur again and errors are correctly logged in the console.


sedghi (approved):

PR #3650, opened 2023-09-13
docs(links): Adding references to two used markdown links

yarikoptic:
I saw

image

There could be more.  I would expect there might be an option in the documentation system to error out if such links are not provided.  Or may be there is a way to provide some central location for such links to avoid duplicating them in every file?
Also I would like to refer you to a nice tool  https://linkcheck.github.io/linkchecker/  which could be used on CI to once in a while check if references used in the documentation are still live.
Here is an example of use of it in BIDS specification project -- does "wonders" ;) https://github.com/bids-standard/bids-specification/blob/HEAD/.circleci/config.yml#L52


sedghi (approved):

PR #3663, opened 2023-09-19
fix(StackSync): Miscellaneous fixes for stack image sync

wayfarer3130:
### Context

The stack image sync has two major issues:
  1. The stack image sync stays active in the toolbar when navigating display sets, but is inactive in the new viewport
  2. The stack image sync was setting the navigation position for FOR linked series, which prevented FOR navigation 

### Changes & Results

Changed the toolbar buttons to contain the listeners for re-activation in various situations
Applied the re-activation to:
  * Stack Prefetch
  * Stack Image Sync
  * Reference Lines

Removed the calls to the navigation position so that the FOR linking works again

Added an integration test for preserving the stack syncing on display set navigation

### Testing
Display 2x2 layout

Activate stack sync
Navigate display sets to linked studies
Display a new display set
Ensure that the stack sync is still active in the toolbar and in the image
Note that stack sync does NOT apply until an image is navigated.  This is pre-existing and has not been fixed.

Activate reference lines
Change the layout
Click on different display sets
Reference lines should be active on the selected viewport

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/extensions/cornerstone/src/commandsModule.ts
+++ b/extensions/cornerstone/src/commandsModule.ts
@@ -267,6 +267,18 @@ function commandsModule({
       toolbarService.recordInteraction(props);
     },
 
+    // Enable or disable a toggleable command, without calling the activation
+    // Used to setup already active tools from hanging protocols
+    toolbarSetActive: props => {
+      toolbarService.setActive(props.toolId, props.isActive ?? true);
+    },
+
+    // Enable stack prefetch on the given element
+    stackPrefetch: props => {
+      const { element } = props;
+      cstUtils.stackPrefetch.enable(element);
+    },
+
     setToolActive: ({ toolName, toolGroupId = null, toggledState }) => {
       if (toolName === 'Crosshairs') {
         const activeViewportToolGroup = toolGroupService.getToolGroup(null);
@@ -344,6 +356,28 @@ function commandsModule({
           },
         ],
       });
+    },
+    toggleReferenceLines: ({ toggledState }) => {
+      try {
+        const { viewportId } = _getActiveViewportEnabledElement();
+        const toolGroup = toolGroupService.getToolGroupForViewport(viewportId);
+
+        if (!toggledState) {
+          toolGroup.setToolDisabled(ReferenceLinesTool.toolName);
+          return;
+        }
+
+        toolGroup.setToolConfiguration(
+          ReferenceLinesTool.toolName,
+          {
+            sourceViewportId: viewportId,
+          },
+          true // overwrite
+        );
+        toolGroup.setToolEnabled(ReferenceLinesTool.toolName);
+      } catch (e) {
+        console.log("Couldn't activate reference lines");
+      }
     },
     showDownloadViewportModal: () => {
       const { activeViewportId } = viewportGridService.getState();
@@ -551,7 +585,6 @@ function commandsModule({
 
     toggleStackImageSync: ({ toggledState }) => {
       toggleStackImageSync({
-        getEnabledElement,
         servicesManager,
         toggledState,
       });
@@ -562,6 +595,11 @@ function commandsModule({
 
       const viewportId = viewportInfo.getViewportId();
       const toolGroup = toolGroupService.getToolGroupForViewport(viewportId);
+
+      if (!toggledState) {
+        toolGroup.setToolDisabled(ReferenceLinesTool.toolName);
+        return;
+      }
 
       toolGroup.setToolConfiguration(
         ReferenceLinesTool.toolName,
@@ -701,8 +739,15 @@ function commandsModule({
     },
     storePresentation: {
       commandFn: actions.storePresentation,
-      storeContexts: [],
-      options: {},
+    },
+    stackPrefetch: {
+      commandFn: actions.stackPrefetch,
+    },
+    toolbarSetActive: {
+      commandFn: actions.toolbarSetActive,
+    },
+    toggleReferenceLines: {
+      commandFn: actions.toggleReferenceLines,
     },
   };
 
--- a/extensions/cornerstone/src/init.tsx
+++ b/extensions/cornerstone/src/init.tsx
@@ -13,7 +13,7 @@ import {
   Settings,
   utilities as csUtilities,
 } from '@cornerstonejs/core';
-import { Enums, utilities, ReferenceLinesTool } from '@cornerstonejs/tools';
+import { Enums, } from '@cornerstonejs/tools';
 import { cornerstoneStreamingImageVolumeLoader } from '@cornerstonejs/streaming-image-volume-loader';
 
 import initWADOImageLoader from './initWADOImageLoader';
@@ -72,6 +72,7 @@ export default async function init({
     cornerstoneViewportService,
     hangingProtocolService,
     toolGroupService,
+    toolbarService,
     viewportGridService,
     stateSyncService,
   } = servicesManager.services as CornerstoneServices;
@@ -182,9 +183,45 @@ export default async function init({
     commandsManager,
   });
 
+  /**
+   * When a viewport gets a new display set, this call will go through all the
+   * active tools in the toolbar, and call any commands registered in the
+   * toolbar service with a callback to re-enable on displaying the viewport.
+   */
   const newStackCallback = evt => {
     const { element } = evt.detail;
-    utilities.stackPrefetch.enable(element);
+    const activeTools = toolbarService.getActiveTools();
+
+    activeTools.forEach(tool => {
+      const toolData = toolbarService.getButton(tool);
+      const commands = toolData?.listeners?.newStack;
+      console.log('Running', tool, commands);
+      commandsManager.run(commands, { element, evt });
+    });
+  };
+
+  const activeViewportIdChangedListener = evt => {
+    const { viewportId } = evt;
+    const toolGroup = toolGroupService.getToolGroupForViewport(viewportId);
+
+    const activeTools = toolbarService.getActiveTools();
+
+    activeTools.forEach(tool => {
+      if (!toolGroup?._toolInstances?.[tool]) {
+        return;
+      }
+
+      // check if tool is active on the new viewport
+      const toolEnabled = toolGroup._toolInstances[tool].mode === Enums.ToolModes.Enabled;
+
+      if (!toolEnabled) {
+        return;
+      }
+
+      const toolData = toolbarService.getButton(tool);
+      const commands = toolData?.listeners?.activeViewportIdChanged;
+      commandsManager.run(commands, { viewportId, evt });
+    });
   };
 
   const resetCrosshairs = evt => {
@@ -236,33 +273,7 @@ export default async function init({
 
   viewportGridService.subscribe(
     viewportGridService.EVENTS.ACTIVE_VIEWPORT_ID_CHANGED,
-    ({ viewportId }) => {
-      const toolGroup = toolGroupService.getToolGroupForViewport(viewportId);
-
-      if (!toolGroup || !toolGroup._toolInstances?.['ReferenceLines']) {
-        return;
-      }
-
-      // check if reference lines are active
-      const referenceLinesEnabled =
-        toolGroup._toolInstances['ReferenceLines'].mode === Enums.ToolModes.Enabled;
-
-      if (!referenceLinesEnabled) {
-        return;
-      }
-
-      toolGroup.setToolConfiguration(
-        ReferenceLinesTool.toolName,
-        {
-          sourceViewportId: viewportId,
-        },
-        true // overwrite
-      );
-
-      // make sure to set it to enabled again since we want to recalculate
-      // the source-target lines
-      toolGroup.setToolEnabled(ReferenceLinesTool.toolName);
-    }
+    activeViewportIdChangedListener
   );
 }
 
--- a/extensions/cornerstone/src/utils/CornerstoneViewportDownloadForm.tsx
+++ b/extensions/cornerstone/src/utils/CornerstoneViewportDownloadForm.tsx
@@ -142,12 +142,16 @@ const CornerstoneViewportDownloadForm = ({
           const properties = viewport.getProperties();
 
           downloadViewport.setStack([imageId]).then(() => {
-            downloadViewport.setProperties(properties);
+            try {
+              downloadViewport.setProperties(properties);
+              const newWidth = Math.min(width || image.width, MAX_TEXTURE_SIZE);
+              const newHeight = Math.min(height || image.height, MAX_TEXTURE_SIZE);
 
-            const newWidth = Math.min(width || image.width, MAX_TEXTURE_SIZE);
-            const newHeight = Math.min(height || image.height, MAX_TEXTURE_SIZE);
-
-            resolve({ width: newWidth, height: newHeight });
+              resolve({ width: newWidth, height: newHeight });
+            } catch (e) {
+              // Happens on clicking the cancel button
+              console.warn('Unable to set properties', e);
+            }
           });
         } else if (downloadViewport instanceof VolumeViewport) {
           const actors = viewport.getActors();
--- a/extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
+++ b/extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
@@ -1,90 +1,81 @@
-import calculateViewportRegistrations from './calculateViewportRegistrations';
+const disableSync = (syncName, servicesManager) => {
+  const { syncGroupService, viewportGridService, displaySetService, cornerstoneViewportService } =
+    servicesManager.services;
+  const viewports = getViewports(viewportGridService, displaySetService);
+  viewports.forEach(gridViewport => {
+    const { viewportId } = gridViewport.viewportOptions;
+    const viewport = cornerstoneViewportService.getCornerstoneViewport(viewportId);
+    if (!viewport) {
+      return;
+    }
+    syncGroupService.removeViewportFromSyncGroup(
+      viewport.id,
+      viewport.getRenderingEngine().id,
+      syncName
+    );
+  });
+};
 
-// [ {
-//   synchronizerId: string,
-//   viewports: [ { viewportId: string, renderingEngineId: string, index: number } , ...]
-// ]}
-let STACK_IMAGE_SYNC_GROUPS_INFO = [];
+const getViewports = (viewportGridService, displaySetService) => {
+  let { viewports } = viewportGridService.getState();
 
-export default function toggleStackImageSync({ toggledState, servicesManager, getEnabledElement }) {
+  viewports = [...viewports.values()];
+  // filter empty viewports
+  viewports = viewports.filter(
+    viewport => viewport.displaySetInstanceUIDs && viewport.displaySetInstanceUIDs.length
+  );
+
+  // filter reconstructable viewports
+  viewports = viewports.filter(viewport => {
+    const { displaySetInstanceUIDs } = viewport;
+
+    for (const displaySetInstanceUID of displaySetInstanceUIDs) {
+      const displaySet = displaySetService.getDisplaySetByUID(displaySetInstanceUID);
+
+      // TODO - add a better test than isReconstructable
+      if (displaySet && displaySet.isReconstructable) {
+        return true;
+      }
+
+      return false;
+    }
+  });
+
+  // viewports = viewports.filter(viewport => viewport.viewportOptions.viewportType === 'stack');
+
+  console.log('viewports=', viewports);
+  return viewports;
+};
+
+const STACK_SYNC_NAME = 'stackImageSync';
+
+export default function toggleStackImageSync({
+  toggledState,
+  servicesManager,
+  viewports: providedViewports,
+}) {
+  if (!toggledState) {
+    return disableSync(STACK_SYNC_NAME, servicesManager);
+  }
+
+  console.log('Toggling stack image sync');
   const { syncGroupService, viewportGridService, displaySetService, cornerstoneViewportService } =
     servicesManager.services;
 
-  if (!toggledState) {
-    STACK_IMAGE_SYNC_GROUPS_INFO.forEach(syncGroupInfo => {
-      const { viewports, synchronizerId } = syncGroupInfo;
+  const viewports = providedViewports || getViewports(viewportGridService, displaySetService);
 
-      viewports.forEach(({ viewportId, renderingEngineId }) => {
-        syncGroupService.removeViewportFromSyncGroup(viewportId, renderingEngineId, synchronizerId);
-      });
-    });
-
-    return;
-  }
-
-  STACK_IMAGE_SYNC_GROUPS_INFO = [];
-
-  // create synchronization groups and add viewports
-  const { viewports } = viewportGridService.getState();
-
-  // filter empty viewports
-  const viewportsArray = Array.from(viewports.values())
-    .filter(viewport => viewport.displaySetInstanceUIDs?.length)
-    // filter reconstructable viewports
-    .filter(viewport => {
-      const { displaySetInstanceUIDs } = viewport;
-
-      for (const displaySetInstanceUID of displaySetInstanceUIDs) {
-        const displaySet = displaySetService.getDisplaySetByUID(displaySetInstanceUID);
-
-        return !!displaySet?.isReconstructable;
-      }
-    });
-
-  const viewportsByOrientation = viewportsArray.reduce((acc, viewport) => {
-    const { viewportId, viewportType } = viewport.viewportOptions;
-
-    if (viewportType !== 'stack') {
-      console.warn('Viewport is not a stack, cannot sync images yet');
-      return acc;
+  // create synchronization group and add the viewports to it.
+  viewports.forEach(gridViewport => {
+    const { viewportId } = gridViewport.viewportOptions;
+    const viewport = cornerstoneViewportService.getCornerstoneViewport(viewportId);
+    if (!viewport) {
+      return;
     }
-
-    const { element } = cornerstoneViewportService.getViewportInfo(viewportId);
-    const { viewport: csViewport, renderingEngineId } = getEnabledElement(element);
-    const { viewPlaneNormal } = csViewport.getCamera();
-
-    // Should we round here? I guess so, but not sure how much precision we need
-    const orientation = viewPlaneNormal.map(v => Math.round(v)).join(',');
-
-    if (!acc[orientation]) {
-      acc[orientation] = [];
-    }
-
-    acc[orientation].push({ viewportId, renderingEngineId });
-
-    return acc;
-  }, {});
-
-  // create synchronizer for each group
-  Object.values(viewportsByOrientation).map(viewports => {
-    let synchronizerId = viewports.map(({ viewportId }) => viewportId).join(',');
-
-    synchronizerId = `imageSync_${synchronizerId}`;
-
-    calculateViewportRegistrations(viewports);
-
-    viewports.forEach(({ viewportId, renderingEngineId }) => {
-      syncGroupService.addViewportToSyncGroup(viewportId, renderingEngineId, {
-        type: 'stackimage',
-        id: synchronizerId,
-        source: true,
-        target: true,
-      });
-    });
-
-    STACK_IMAGE_SYNC_GROUPS_INFO.push({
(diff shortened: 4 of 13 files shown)


extensions/cornerstone/src/commandsModule.ts
wayfarer3130, line 1: This file contains the changes to re-enable stack prefetch, stack sync and reference lines.

extensions/cornerstone/src/init.tsx
wayfarer3130, line 1: Generalized the code to preserve active features in viewports.

extensions/cornerstone/src/utils/CornerstoneViewportDownloadForm.tsx
```diff
@@ -142,12 +142,16 @@ const CornerstoneViewportDownloadForm = ({
           const properties = viewport.getProperties();
 
           downloadViewport.setStack([imageId]).then(() => {
-            downloadViewport.setProperties(properties);
-
-            const newWidth = Math.min(width || image.width, MAX_TEXTURE_SIZE);
-            const newHeight = Math.min(height || image.height, MAX_TEXTURE_SIZE);
-
-            resolve({ width: newWidth, height: newHeight });
+            try {
```
wayfarer3130, line 145: Something I changed cause an exception to happen 100%, probably something around the prefetch configuration/activation.

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
```diff
@@ -1,90 +1,81 @@
-import calculateViewportRegistrations from './calculateViewportRegistrations';
-
-// [ {
-//   synchronizerId: string,
-//   viewports: [ { viewportId: string, renderingEngineId: string, index: number } , ...]
-// ]}
-let STACK_IMAGE_SYNC_GROUPS_INFO = [];
-
-export default function toggleStackImageSync({ toggledState, servicesManager, getEnabledElement }) {
+const disableSync = (syncName, servicesManager) => {
```
wayfarer3130, line 1: This code will probably get migrated into the sync service at some point as it is fairly generic now.

modes/basic-test-mode/src/toolbarButtons.js
```diff
@@ -67,6 +54,88 @@ function _createWwwcPreset(preset, title, subtitle) {
...
+  'StackImageSync',
+  'link',
+  'Stack Image Sync',
+  [
+    {
+      commandName: 'toggleStackImageSync',
+    },
+  ],
+  'Enable position synchronization on stack viewports',
+  {
```
wayfarer3130, line 87: Add however many different listener types to re-enable the tool on changes.  This allows consistency between the toolbar state and the cornerstone state, without needing to bake it into the cornerstone extension.

platform/app/cypress/integration/measurement-tracking/OHIFCornerstoneToolbar.spec.js
```diff
@@ -406,20 +406,42 @@ describe('OHIF Cornerstone Toolbar', () => {
...
+    cy.waitDicomImage();
+    cy.get('@viewportInfoMidLeft').should('contains.text', 'L');
+    cy.get('@viewportInfoMidTop').should('contains.text', 'A');
+  });
 
-    //Click on More button to close it
-    cy.get('@moreBtn').click();
-    cy.get('.tooltip-toolbar-overlay').should('not.exist');
-  });*/
+  it('checks if stack sync is preserved on new display set and uses FOR', () => {
```
wayfarer3130, line 424: Here is the primary test for this issue.

platform/app/cypress/integration/measurement-tracking/OHIFDownloadSnapshotFile.spec.js
```diff
@@ -36,6 +36,8 @@ describe('OHIF Download Snapshot File', () => {
     // Check buttons
     cy.get('[data-cy="cancel-btn"]').scrollIntoView().should('be.visible');
     cy.get('[data-cy="download-btn"]').scrollIntoView().should('be.visible');
+
+    cy.get('[data-cy="cancel-btn"]').click();
```
wayfarer3130, line 40: The close was throwing an exception, and without it, this test passes, but the next one fails.

platform/app/cypress/integration/measurement-tracking/OHIFGeneralViewer.spec.js
```diff
@@ -1,11 +1,9 @@
-describe('OHIF Study Viewer Page', function () {
-  beforeEach(function () {
-    cy.checkStudyRouteInViewer('1.2.840.113619.2.5.1762583153.215519.978957063.78');
-
-    cy.expectMinimumThumbnails(3);
-    cy.initCommonElementsAliases();
-    cy.initCornerstoneToolsAliases();
-  });
+describe('OHIF General Viewer', function () {
```
wayfarer3130, line 1: Duplicated names made it hard to debug which test was failing.

platform/app/cypress/support/commands.js
```diff
@@ -357,14 +369,9 @@ Cypress.Commands.add('percyCanvasSnapshot', (name, options = {}) => {
 });
 
 Cypress.Commands.add('setLayout', (columns = 1, rows = 1) => {
-  cy.get('[data-cy="layout"]').click();
+  cy.get('[data-cy="Layout"]').click();
```
wayfarer3130, line 372: Fix the layout command

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -172,7 +172,13 @@ export default class ToolbarService extends PubSubService {
   }
 
   getActiveTools() {
-    return [this.state.primaryToolId, ...Object.keys(this.state.toggles)];
+    const activeTools = [this.state.primaryToolId];
```
wayfarer3130, line 175: Somebody else broke this, and the generalized re-initialize of the toolbar needs this to work so fixing it here.

platform/ui/src/components/LayoutSelector/LayoutSelector.tsx
```diff
@@ -34,6 +34,7 @@ function LayoutSelector({ onSelection, rows, columns }) {
               border: '1px solid white',
               backgroundColor: isHovered(index) ? '#5acce6' : '#0b1a42',
             }}
+            data-cy={`Layout-${index % columns}-${Math.floor(index / columns)}`}
```
wayfarer3130, line 37: Need a data-cy to find the element to click.

extensions/cornerstone/src/utils/CornerstoneViewportDownloadForm.tsx
jbocce: How do you reproduce this problem?

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -76,6 +74,68 @@ function _createSetToolActiveCommands(toolName) {
...
+  }
+);
+
+const ReferenceLines = _createToggleButton(
+  'ReferenceLines',
+  'tool-referenceLines', // change this with the new icon
+  'Reference Lines',
+  [
+    {
+      commandName: 'toggleReferenceLines',
```
jbocce, line 119: We just recently removed the `toggleReferenceLines` command. Is there a way to avoid having to bring it back? Can the button instead just run the two commands and also have the listeners?

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -15,16 +15,14 @@ const { windowLevelPresets } = defaults;
...
   return {
     id,
     icon,
     label,
     type,
     commands,
-    tooltip,
-    uiType,
-    isActive,
+    ...extraOptions,
```
jbocce, line 25: Is `tooltip` a member of `extraOptions`? If so I think it should be removed from the parameter list on line 18. Furthermore, you will have to double check all the former calls to `_createButton` to be sure they are using the new form correctly. For example, I think the image overlay button is no longer active by default and it should be.

modes/basic-test-mode/src/toolbarButtons.js
```diff
@@ -67,6 +54,88 @@ function _createWwwcPreset(preset, title, subtitle) {
...
 
+const toolGroupIds = ['default', 'mpr', 'SRToolGroup'];
+
+/**
+ * Creates an array of 'setToolActive' commands for the given toolName - one for
+ * each toolGroupId specified in toolGroupIds.
+ * @param {string} toolName
+ * @returns {Array} an array of 'setToolActive' commands
+ */
+function _createSetToolActiveCommands(toolName) {
```
jbocce, line 65: Is this so that a tool will remain active when switching say to `mpr`?

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -631,6 +664,11 @@ const toolbarButtons = [
       ],
     },
   },
+
+  // Register buttons that are used for top level actions
+  StackImageSync,
```
jbocce, line 669: It is a bit unexpected to see these here, but I guess they are here so that the tools can be fetched from the ToolbarService and the listeners invoked?

jbocce (commented):
Please see my comments.

modes/basic-test-mode/src/toolbarButtons.js
```diff
@@ -67,6 +54,88 @@ function _createWwwcPreset(preset, title, subtitle) {
...
+  'tool-referenceLines', // change this with the new icon
+  'Reference Lines',
+  [
+    {
+      commandName: 'toggleReferenceLines',
+      commandOptions: {},
+      context: 'CORNERSTONE',
+    },
+  ],
+  'Show reference lines',
```
sedghi, line 124: what is this line? it is not tooltip

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -76,6 +74,68 @@ function _createSetToolActiveCommands(toolName) {
...
+      context: 'CORNERSTONE',
+    },
+  ],
+  'Show reference lines',
+  {
+    listeners: {
+      activeViewportIdChanged: {
+        commandName: 'toggleReferenceLines',
+        commandOptions: { toggledState: true },
+      },
```
sedghi, line 130: there can be more than one command to run on an event, so this should be an array

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -76,6 +74,68 @@ function _createSetToolActiveCommands(toolName) {
...
+    {
+      commandName: 'toggleReferenceLines',
+      commandOptions: {},
+      context: 'CORNERSTONE',
+    },
+  ],
+  'Show reference lines',
+  {
+    listeners: {
+      activeViewportIdChanged: {
```
sedghi, line 127: should be an enum, and OHIF should export it somewhere to get used

modes/longitudinal/src/toolbarButtons.js
```diff
@@ -76,6 +74,68 @@ function _createSetToolActiveCommands(toolName) {
...
+        commandName: 'toggleReferenceLines',
+        commandOptions: { toggledState: true },
+      },
+      newStack: {
```
sedghi, line 131: newStack should be an enum

extensions/cornerstone/src/utils/CornerstoneViewportDownloadForm.tsx
wayfarer3130: Run the integration tests with the other changes but not this one.

modes/longitudinal/src/toolbarButtons.js
wayfarer3130: What two commands would it run?

wayfarer3130: Found them - will remove this method.

modes/longitudinal/src/toolbarButtons.js
wayfarer3130: It accepts either a singleton or an array, the same as the other commands lists.

modes/longitudinal/src/toolbarButtons.js
wayfarer3130: Why?  None of the rest of this is typed.  Typing it generally would be useful, which I can do, but using an enum just makes it hard to declare things.

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -195,10 +249,25 @@ export default class ToolbarService extends PubSubService {
     }
   }
 
-  getButton(id) {
+  public getButton(id: string): Button {
     return this.buttons[id];
   }
 
+  /** Gets a nested button, found in the items/props for the children */
```
wayfarer3130, line 256: @jbocce - as you requested, do a search through the buttons to find them.

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -2,12 +2,40 @@ import merge from 'lodash.merge';
...
 import { ExtensionManager } from '../../extensions';
 import { PubSubService } from '../_shared/pubSubServiceInterface';
+import type { RunCommand, Commands } from '../../types/Command';
 
 const EVENTS = {
   TOOL_BAR_MODIFIED: 'event::toolBarService:toolBarModified',
   TOOL_BAR_STATE_MODIFIED: 'event::toolBarService:toolBarStateModified',
 };
 
+export type ButtonListeners = {
```
wayfarer3130, line 12: @sedghi - as you requested, added types for the listeners (and buttons)

platform/core/src/types/index.ts
```diff
@@ -12,12 +12,13 @@ export * from '../services/CustomizationService/types';
...
 export * from './AppConfig';
 export * from './Consumer';
-export * from './Command';
+export type * from './Command';
 export * from './DisplaySet';
 export * from './StudyMetadata';
 export * from './PanelModule';
 export * from './IPubSub';
 export * from './Color';
+export type * from '../services/ToolbarService/ToolbarService';
```
sedghi, line 21: why not others?

platform/app/cypress/support/commands.js
```diff
@@ -69,6 +69,18 @@ Cypress.Commands.add(
   }
 );
 
+Cypress.Commands.add('initViewer', (StudyInstanceUID, other = {}) => {
+  const { mode = '/basic-test', minimumThumbnails = 1, params = '' } = other;
+  cy.openStudyInViewer(StudyInstanceUID, params, mode);
+  cy.waitDicomImage();
+  // Very short wait to ensure pending updates are handled
+  cy.wait(25);
```
sedghi, line 77: we can go much higher, i don't think 100ms would hurt either

platform/app/cypress/integration/measurement-tracking/OHIFCornerstoneToolbar.spec.js
sedghi: Thanks a lot

modes/longitudinal/src/toolbarButtons.js
wayfarer3130: It wasn't supposed to get removed, so I added it back in.  It is so common I wanted to leave it.

platform/core/src/types/index.ts
wayfarer3130: Those were the ones I was touching.  The lint used to complain about it, and now it wants them exported as types.

platform/app/cypress/support/commands.js
wayfarer3130: It isn't needed - it only needs a very short wait to ensure the next render has completed.  The waitDicomImage is pretty good now about not returning until it is actually done.

wayfarer3130: Making it higher slows the tests down quite a bit.

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -2,12 +2,40 @@ import merge from 'lodash.merge';
...
+export interface Button extends Commands {
+  id: string;
+  icon?: string;
+  label?: string;
+  type?: string;
+  tooltip?: string;
+  isActive?: boolean;
+  listeners?: ButtonListeners;
+  props?: ButtonProps;
+}
```
sedghi, line 32: ButtonListeners have cornerstone logic/event (newStack) which doesn't seem right. Not really sure what/where is the best way/place to add these Enums. any thoughts? 

PS: I think later on someone can add another listener for their custom tool doing some custom event, how are we handling that?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -267,6 +267,18 @@ function commandsModule({
       toolbarService.recordInteraction(props);
     },
 
+    // Enable or disable a toggleable command, without calling the activation
+    // Used to setup already active tools from hanging protocols
+    toolbarSetActive: props => {
+      toolbarService.setActive(props.toolId, props.isActive ?? true);
+    },
```
sedghi, line 274: I think we should rename this function in the toolbarService. It is not setActive it is basically `updateToggleStatus`

```
/** Sets the toggle state of a button to the isActive state */
  public setActive(id: string, isActive: boolean): void {
    if (isActive) {
      this.state.toggles[id] = true;
    } else {
      delete this.state.toggles[id];
    }
  }
```

extensions/cornerstone/src/init.tsx
```diff
@@ -182,9 +183,44 @@ export default async function init({
...
+      // check if tool is active on the new viewport
+      const toolEnabled = toolGroup._toolInstances[tool].mode === Enums.ToolModes.Enabled;
+
+      if (!toolEnabled) {
+        return;
+      }
+
+      const toolData = toolbarService.getNestedButton(tool);
+      const commands = toolData?.listeners?.activeViewportIdChanged;
+      commandsManager.run(commands, { viewportId, evt });
```
sedghi, line 222: this seems to be a common pattern, maybe we refactor? and maybe rename toolData to buttons

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
```diff
@@ -1,90 +1,81 @@
...
   const { syncGroupService, viewportGridService, displaySetService, cornerstoneViewportService } =
```
sedghi, line 2: can we move this helpers to the bottom so that the first thing we see opening this file is the actual start of the logic?

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
```diff
@@ -1,90 +1,81 @@
...
-    });
-
-    return;
-  }
-
-  STACK_IMAGE_SYNC_GROUPS_INFO = [];
-
-  // create synchronization groups and add viewports
-  const { viewports } = viewportGridService.getState();
+const getViewports = (viewportGridService, displaySetService) => {
```
sedghi, line 19: maybe rename to `_getReconstructableViewports` or something

sedghi: why are we even filetring the volume viewports? it is stackImageSync

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
```diff
@@ -1,90 +1,81 @@
...
+  // create synchronization group and add the viewports to it.
+  viewports.forEach(gridViewport => {
+    const { viewportId } = gridViewport.viewportOptions;
+    const viewport = cornerstoneViewportService.getCornerstoneViewport(viewportId);
+    if (!viewport) {
+      return;
+    }
+    syncGroupService.addViewportToSyncGroup(viewportId, viewport.getRenderingEngine().id, {
+      type: 'stackimage',
+      id: STACK_SYNC_NAME,
```
sedghi, line 76: I liked the pattern we had before

```
  let synchronizerId = viewports.map(({ viewportId }) => viewportId).join(',');

    synchronizerId = `imageSync_${synchronizerId}`;
```

sedghi (commented):
see my comments

platform/core/src/types/index.ts
sedghi: at least put a todo that the rest should be done (it is easy just do it)

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: Renamed to setToggled to be in keeping with the existing naming.

extensions/cornerstone/src/init.tsx
wayfarer3130: Renamed but didn't factor out the common pattern.  Will think about what is worth doing there once we have one or two more examples.

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
wayfarer3130: Although I generally like that logic, it doesn't really work because you are supposed to declare functions before using them.  It isn't strictly necessary, but is preferred.

extensions/cornerstone/src/commandsModule.ts
sedghi: I don't like setToggled and don't like updateToggleStatus either, @jbocce what do you think

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
sedghi: use functions instead of const so that it get hoisted https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions#function_hoisting

extensions/cornerstone/src/commandsModule.ts
jbocce: I prefer to leave it as `setActive` because it is most explicit as such.

sedghi: But it is not actually setActive, it acts on only toggles, so toggles should be on the name at least

jbocce: In my opinion it IS `setActive`. Consider using `setActiveForToggleButton`.

modes/longitudinal/src/toolbarButtons.ts
```diff
@@ -5,32 +5,15 @@ import {
...
-    uiType,
-    isActive,
-  };
-}
 
-const _createActionButton = _createButton.bind(null, 'action');
-const _createToggleButton = _createButton.bind(null, 'toggle');
-const _createToolButton = _createButton.bind(null, 'tool');
+const _createActionButton = ToolbarService._createButton.bind(null, 'action');
+const _createToggleButton = ToolbarService._createButton.bind(null, 'toggle');
```
jbocce, line 15: With this refactoring, I noticed that the DICOM overlay button now has the wrong initial state. It requires the `isActive` to be passed as part of the extra options argument.

jbocce (commented):
Please see my comments.

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
wayfarer3130: I'd rather not because eventually it should handle other types of viewports, once the CS3D side gets fixed to handle other types.

.webpack/webpack.base.js
```diff
@@ -32,6 +32,24 @@ const COMMIT_HASH = fs.readFileSync(path.join(__dirname, '../commit.txt'), 'utf8
 //
 dotenv.config();
 
+const defineValues = {
```
sedghi, line 35: Noice

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -265,6 +265,19 @@ function commandsModule({
...
+    // Enable or disable a toggleable command, without calling the activation
+    // Used to setup already active tools from hanging protocols
+    setToolbarToggled: props => {
+      toolbarService.setToggled(props.toolId, props.isActive ?? true);
+    },
+
+    // Enable stack prefetch on the given element
+    stackPrefetch: props => {
+      const { element, prefetchType = 'stackContextPrefetch' } = props;
+      console.log('Enabling', prefetchType);
```
sedghi, line 277: remove log please

platform/app/.webpack/webpack.pwa.js
```diff
@@ -131,6 +131,7 @@ module.exports = (env, argv) => {
         maximumFileSizeToCacheInBytes: 5 * 1024 * 1024,
         // Need to exclude the theme as it is updated independently
         exclude: [/theme/],
+        maximumFileSizeToCacheInBytes: 1024 * 1024 * 50,
```
sedghi, line 134: please add a comment and why

modes/longitudinal/src/toolbarButtons.ts
```diff
@@ -5,32 +5,15 @@ import {
   // ListMenu,
   WindowLevelMenuItem,
 } from '@ohif/ui';
-import { defaults } from '@ohif/core';
+import { defaults, ToolbarService } from '@ohif/core';
+import type { Button, RunCommand } from '@ohif/core/types';
+import { EVENTS } from '@cornerstonejs/core';
```
sedghi, line 10: I beleive this means the mode package json should depend on cornerstonejs too

wayfarer3130: Yes, that is true, but it already implicitly does so.  Also, there is no reason one can't declare the values as simple string values.

extensions/cornerstone/src/utils/stackSync/toggleStackImageSync.ts
```diff
@@ -1,90 +1,80 @@
-import calculateViewportRegistrations from './calculateViewportRegistrations';
+const STACK_SYNC_NAME = 'stackImageSync';
```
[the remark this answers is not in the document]
wayfarer3130: Handled by the CS3D tool library - that way there is exactly one spot to do the calculations.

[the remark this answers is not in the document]
wayfarer3130: Also, I haven't removed calculateViewportRegistrations because it is needed for another use case which we don't yet support, allowing one to adjust hte registrations.

platform/app/.webpack/webpack.pwa.js
```diff
@@ -131,6 +131,8 @@ module.exports = (env, argv) => {
         maximumFileSizeToCacheInBytes: 5 * 1024 * 1024,
         // Need to exclude the theme as it is updated independently
         exclude: [/theme/],
+        // Cache large files for hte manifests to avoid warning messages
```
jbocce, line 134: *the

extensions/cornerstone/src/utils/initViewTiming.ts
```diff
@@ -0,0 +1,42 @@
...
+  imageTiming.viewportsWaiting += 1;
+  element.addEventListener(EVENTS.IMAGE_RENDERED, imageRenderedListener);
+}
+
+function imageRenderedListener(evt) {
+  log.timeEnd(TimingEnum.DISPLAY_SETS_TO_FIRST_IMAGE);
+  log.timeEnd(TimingEnum.STUDY_TO_FIRST_IMAGE);
+  log.timeEnd(TimingEnum.SCRIPT_TO_VIEW);
+  imageTiming.viewportsWaiting -= 1;
+  evt.detail.element.removeEventListener(EVENTS.IMAGE_RENDERED, imageRenderedListener);
```
jbocce, line 38: Should we have a fallback for removing the rendered event listener in case the image is not actually rendered. Not sure which event to listen for... maybe EVENTS.IMAGE_LOAD_FAILED?

wayfarer3130: They are removed automatically when destroyed.  I agree, if one has an image load failure and never a success, then maybe it would be na issue, but then you aren't displaying anything.

jbocce (approved):
Approved.

sedghi (changes requested):
Update: the following is not valid after upgrade of cs3d

~~stack prefetch has been broken, most likely since you assigned it to a button, although I don't think we should have it on a button, should be enabled by default BUT a button can disable it~~

https://github.com/OHIF/Viewers/assets/7490180/7ae5a619-7620-4172-8422-0798491b780d

sedghi (approved):

PR #3681, opened 2023-09-27
feat(debug): Add timing information about time to first image/all images, and query time

wayfarer3130:
Adding timing information is very low cost and provides insight into what is happening on launch or viewing a study.

### Context

The logs framework was updated to contain information on the currently timed events.  This is then used to provide information on time from clicking on a mode button to the actual display of the data, and to various intermediate points.

### Changes & Results
Add various timing information, see TimingEnum for the types.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


jbocce (approved):
Approved.

platform/core/src/log.js
```diff
@@ -4,8 +4,23 @@ const log = {
...
   trace: console.trace,
   debug: console.debug,
-  time: console.time,
-  timeEnd: console.timeEnd,
+  time: key => {
+    log.timingKeys[key] = true;
+    console.time(key);
+  },
+  timeEnd: key => {
+    if (!log.timingKeys[key]) {
```
sedghi, line 12: can we consle.warn here before retun that key does not exist?

platform/core/src/log.js
```diff
@@ -4,8 +4,23 @@ const log = {
...
+    }
+    log.timingKeys[key] = false;
+    console.timeEnd(key);
+  },
+  // Store the timing keys to allow knowing whether or not to log events
+  timingKeys: {
+    // script time values are added before log is loaded, and the log
+    // shouldn't depend on the enums, so for this case recreate the string.
+    scriptToView: true,
+  },
```
sedghi, line 23: it would be great if we can have an idea of when does script starts timing and ends

platform/core/src/types/TimingEnum.ts
```diff
@@ -0,0 +1,8 @@
+export enum TimingEnum {
+  STUDY_TO_DISPLAY_SETS = 'studyToDisplaySetsLoaded',
+  STUDY_TO_FIRST_IMAGE = 'studyToFirstImage',
+  DISPLAY_SETS_TO_FIRST_IMAGE = 'displaySetsToFirstImage',
```
sedghi, line 4: first image load? render?

platform/core/src/types/TimingEnum.ts
```diff
@@ -0,0 +1,8 @@
...
+  DISPLAY_SETS_TO_ALL_IMAGES = 'displaySetsToAllImages',
```
sedghi, line 5: to all images cached? load? decode? render?

platform/core/src/types/TimingEnum.ts
```diff
@@ -0,0 +1,8 @@
...
+  SCRIPT_TO_VIEW = 'scriptToView',
+  SEARCH_TO_LIST = 'searchToList',
```
sedghi, line 7: worklist?

extensions/cornerstone/src/utils/initViewTiming.ts
```diff
@@ -0,0 +1,45 @@
...
+export default function initViewTiming({ element }) {
+  if (!IMAGE_TIMING_KEYS.find(key => log.timingKeys[key])) {
+    return;
+  }
+  imageTiming.viewportsWaiting += 1;
+  element.addEventListener(EVENTS.IMAGE_RENDERED, imageRenderedListener);
+}
+
+function imageRenderedListener(evt) {
+  if (evt.detail.viewportStatus === 'preRender') {
```
sedghi, line 34: this has a bug regarding which image rendered, you should check the viewportId and check if it is the same id no?

sedghi (commented):
see my comments please, great initiative

extensions/cornerstone/src/utils/initViewTiming.ts
wayfarer3130: No, because we only care that the count of viewports matches - we un subscribe after the first event that is rendered from each viewport, so each viewport can only get 1 count for it, and we have the total count number.  The logic of the event handling deals with matching up viewport id's to renders.    The logging isn't for this viewport rendered for first image, it is for any cornerstone viewport rendering.

platform/core/src/log.js
wayfarer3130: That is specifically avoided - we get lots more time ends than we do time starts, and we don't care that the key doesn't exist any longer.

wayfarer3130:
Some example timing on a cached instance, with non-cache reasonably large images:
displaySetsToFirstImage: 255.36083984375 ms
studyToFirstImage: 311.17919921875 ms
scriptToView: 791.715087890625 ms
displaySetsToAllImages: 255.4677734375 ms

scriptToView on a clean browser is about 1188 ms for the same study, while the other numbers are roughly the same.

sedghi (approved):
see my comments please, great initiative

Issue #3685, opened 2023-09-29
[Bug] 3.7 QA Bug Report


james-hanks:
**Issues**

New Segmentation: Color of new segment doesn't change on first click 
- Create new segmentation
- Change color of fist segment
- Draw segment
- Color of segment drawn is the same as last selected segment
- Change color again, segment color now changes

**Feedback**
- Adjust range of mm radius slider on tool size. 0.01 is too small and 99 is too large. Also, I can't see anything drawn at 0.01 size even zoomed in

jbocce:
Swapping in an already hydrated RTSTRUCT throws an exception and results in the destination viewport of the swap becoming blank. See the video below. The exception is...

```
C:\dev\Viewers\extensions\cornerstone\src\services\CornerstoneCacheService\CornerstoneCacheService.ts:227 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '0')
    at CornerstoneCacheService._shouldRenderSegmentation (C:\dev\Viewers\extensions\cornerstone\src\services\CornerstoneCacheService\CornerstoneCacheService.ts:227:1)
    at CornerstoneCacheService.createViewportData (C:\dev\Viewers\extensions\cornerstone\src\services\CornerstoneCacheService\CornerstoneCacheService.ts:47:1)
    at loadViewportData (C:\dev\Viewers\extensions\cornerstone\src\Viewport\OHIFCornerstoneViewport.tsx:258:1)
    at C:\dev\Viewers\extensions\cornerstone\src\Viewport\OHIFCornerstoneViewport.tsx:304:1
    at invokePassiveEffectCreate (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:23487:1)
    at HTMLUnknownElement.callCallback (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:3994:1)
    at invokeGuardedCallback (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:4056:1)
    at flushPassiveEffectsImpl (C:\dev\Viewers\node_modules\react-dom\cjs\react-dom.development.js:23574:1)
    at unstable_runWithPriority (C:\dev\Viewers\node_modules\scheduler\cjs\scheduler.development.js:468:1)
```


[...]

PR #3688, opened 2023-10-02
fix(config): support more values for the useSharedArrayBuffer

sedghi:
### Context

- Enable running the app without shared array buffer
- Enable changing the segmentation color and being able to render the correct color
- Change the step size for brush size

### Testing

Comments in this https://github.com/OHIF/Viewers/issues/3685

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone/src/init.tsx
```diff
@@ -43,15 +43,15 @@ export default async function init({
...
-      cornerstone.setUseSharedArrayBuffer(csEnums.SharedArrayBufferModes.FALSE);
-      break;
-    default:
-      cornerstone.setUseSharedArrayBuffer(csEnums.SharedArrayBufferModes.TRUE);
+  // DO NOT CHANGE THE ORDER
+  const value = appConfig.useSharedArrayBuffer;
+
+  if (value === 'AUTO') {
+    cornerstone.setUseSharedArrayBuffer(csEnums.SharedArrayBufferModes.AUTO);
+  } else if (value === 'FALSE' || value === false) {
```
jbocce, line 51: Does it make sense to instead do something like `else if (value?.toString()?.toLowerCase() === 'false') {`?

jbocce:
Some issues found...

- It is not possible to enter a two digit value for the brush radius 
- It is not possible to backspace from after the decimal to correct a value for the brush radius
- A decimal value cannot be entered for the brush radius (e.g. select existing value and try to type 15.5)
- In general, the brush radius input text field is acting strange.

jbocce (changes requested):
Please have a look at my comments and issues I discovered. Thanks.

extensions/cornerstone/src/init.tsx
sedghi: I think what i have is more clear in reading if that is ok with you

jbocce: Your call...

platform/docs/docs/faq.md
```diff
@@ -29,6 +29,9 @@ If you have resources and would like to fund the development of a feature,
 please [contact us](https://www.ohif.org) or work with community members that
 offer [consulting services][commercial-support].
 
+### Cross Origin Isolation
```
jbocce, line 32: Since this is an FAQ, consider using a question instead. Perhaps, "Why do I keep seeing a Cross Origin Isolation warning?"

jbocce (commented):
Just some minor doc comments.

Issue #3764, opened 2023-11-03
[Feature Request] Create new multi-level menu for per viewport window level as per the design in the parent epic.


jbocce:
### What feature or change would you like to see made?

See the design for the menu in the epic #3493

### Why should we prioritize this feature?

It is needed for the per viewport window level changes.

sedghi:
We just release the OHIF 3.8, you can find more details here https://ohif.org/release-notes/3p8/
If you still encounter this issue in 3.8, please re-open this.

Issue #3493, opened 2023-06-23
Support for per-viewport W/L presets


james-hanks:
- [x] #3742
- [x] #3743
- [x] #3764
- [x] #3825 
- [x] #3814

Latest design and behviors - Updated Nov 2, 2023 by @dan-rukas 

- Viewport - Layout overview
- Viewport - Actions hide and show
- Viewport - Status
- Viewport - Actions and tools
- Viewport - Window Levels
- Component - Multi-level Menu

![Image](https://github.com/OHIF/Viewers/assets/2745168/0360e7c7-98b2-4d41-aef5-05842b4cda90)

![Image](https://github.com/OHIF/Viewers/assets/2745168/c87aa43c-c8ef-4612-b4e4-bbcf9fb802b3)

![Image](https://github.com/OHIF/Viewers/assets/2745168/27b1ed75-c3af-437b-bd9e-1b3242ad8ece)

![Image](https://github.com/OHIF/Viewers/assets/2745168/8a65b5f9-19a7-4d38-8c96-f4ebf4d35c2e)

![Image](https://github.com/OHIF/Viewers/assets/2745168/137b19b8-a97e-4110-a529-ce200af1f0f1)

<img width="809" alt="Component-Multi-Menu" src="https://github.com/OHIF/Viewers/assets/2745168/9624d137-f1bb-44ce-8b41-80a61f8d6d62">

sedghi:
Some sketch ideas

![image](https://github.com/OHIF/Viewers/assets/7490180/cf1d691f-e700-4307-8634-ab3e41e22a98)

sedghi:
![image](https://github.com/OHIF/Viewers/assets/7490180/76600072-abcd-4a0f-ba8a-91282f455519)

james-hanks:
@jbocce to add development tasks and link

dan-rukas:
Updated all finalized designs.

Includes:
- Viewport - Layout overview
- Viewport - Actions hide and show
- Viewport - Status
- Viewport - Actions and tools
- Viewport - Window Levels
- Component - Multi-level Menu

sedghi:
Remaining stuff
- colorLUT 
- 3d presets 
- patient info on top

dan-rukas:
Adding the last design piece here. Volume Rendering dialog is an option in the W/L menu that show an additional moveable modal window. 

Design below
3d thumbnail assets: 
[...]

PR #3774, opened 2023-11-08
feat(UI): Added new AllInOneMenu and SwitchButton components.

jbocce:
These are necessary for the new window level feature.

### Context

Addresses #3764 for the window level epic #3493

### Changes & Results

Added new components.

### Testing

Run storybook to see the new components in action.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 18.16.1
- [x] Browser: Chrome 118.0.5993.120
  


platform/ui/src/components/SwitchButton/SwitchButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import './switchButton.css';
+
+export enum SwitchLabelLocation {
+  left,
+  right,
+}
+
+export type SwitchButtonProps = {
+  isSwitchOn?: boolean;
```
sedghi, line 11: enabled? instead of isSwitchOn?

platform/ui/src/components/SwitchButton/SwitchButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  label?: string;
+  labelLocation?: SwitchLabelLocation;
+  onSwitchChange?: (isSwitchOn: boolean) => void;
```
sedghi, line 14: onChange?

platform/ui/src/components/SwitchButton/SwitchButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+          type="checkbox"
+          onChange={onChange}
+          checked={isChecked}
+        />
+        <div className="switch-button-outer border-common-bright bg-primary-dark block h-[16px] w-[30px] rounded-full border"></div>
+        <div className="switch-button-dot bg-common-bright absolute left-[4px] top-[3px] h-[10px] w-[10px] rounded-full transition duration-150 ease-in-out"></div>
+      </div>
+      {label && labelLocation === SwitchLabelLocation.right && <div>{label}</div>}
+    </label>
+  );
```
sedghi, line 48: if we are using a code from somewhere lets put a reference

platform/ui/src/components/SwitchButton/SwitchButton.tsx
jbocce: enabled to me is confusing, because the opposite of enabled is disabled which to me means the button CANNOT even be clicked.

platform/ui/src/components/index.js
```diff
@@ -79,10 +79,25 @@ import PanelSection from './PanelSection';
...
 import LegacyButtonGroup from './LegacyButtonGroup';
+import SwitchButton, { SwitchLabelLocation } from './SwitchButton';
+import {
+  AllInOneMenu,
+  AllInOneMenuDividerItem,
+  AllInOneMenuHeaderItem,
+  AllInOneMenuItem,
+  AllInOneMenuItemPanel,
+  AllInOneMenuSubMenu,
+} from './AllInOneMenu';
```
sedghi, line 90: two things

1. can we namespace this?

so `import * as AllInOneMenu` so that we can use `AllInOneMenu.AllInOneMenuDividerItem` 

2. The AllInOneMenu might be redundant and too noisy, so how about we rename things so that we can do `AllInOneMenu.DividerItem|HeaderItem|Item|ItemPanel`

platform/ui/src/index.js
```diff
@@ -37,6 +37,12 @@ export {
 export {
   AboutModal,
   AdvancedToolbox,
+  AllInOneMenu,
+  AllInOneMenuDividerItem,
+  AllInOneMenuHeaderItem,
+  AllInOneMenuItem,
+  AllInOneMenuItemPanel,
+  AllInOneMenuSubMenu,
```
sedghi, line 45: same comment regarding name spacing

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+  title: 'Components/AllInOneMenu',
+};
+
+<Meta
+  title="Components/AllInOneMenu"
+  component={AllInOneMenu}
+/>
+
+export const AllInOneMenuTemplate = args => (
+  <div className="w-80">
```
sedghi, line 24: I konw this is a test, but it shows the usage. So Let me comment here on the usage.

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+    <AllInOneMenu {...args}>
+      <AllInOneMenuItemPanel label="Panel 1">
```
sedghi, line 26: where is this label getting used? i don't see that in the UI

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+        <AllInOneMenuItem
+          label="Item 1"
+          key="0"
+          onItemClick={() => console.info('Item 1 clicked.')}
+        ></AllInOneMenuItem>
+        <AllInOneMenuItem
+          label="Item 2"
+          alternateLabel="Alt item 2"
```
sedghi, line 34: secondaryLabel?

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+          key="1"
+        ></AllInOneMenuItem>
+        <div
+          className="all-in-one-menu-item my-1 flex w-full justify-center"
+          style={{ height: '50px' }}
+        >
```
sedghi, line 40: can we extract these styles and put it as an enum, to use them elsewhere for our custom sub components?

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+            minValue={1}
+            maxValue={10}
+            value={5}
+            onChange={() => {}}
+          ></InputRange>
+        </div>
+        <AllInOneMenuDividerItem key="2"></AllInOneMenuDividerItem>
+        <AllInOneMenuSubMenu
+          key="3"
+          menuLabel="Level 2"
```
sedghi, line 52: where is this `menuLabel` used? i don't see it in the UI

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+          itemLabel="Item 3 opens a sub menu"
+          onItemClick={() => console.info('Sub menu item clicked.')}
```
sedghi, line 54: in the above you had `onItemClick` on the `AllInOneMenuItem` but here you have onItemClick on `AllInOneMenuSubMenu` which is unusual to me. Can we just say `onClick` for all?

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+          isHeaderDividerVisible={true}
```
sedghi, line 55: showHeaderDivider?

sedghi (commented):
Great stuff

1. I feel like the hover background is wrong, maybe your implementation or maybe even the design. I expected it to fill the region, or at least be symmetric with top and bottom

CleanShot 2023-11-08 at 09 23 17

2. Looks like there is a bottom layout shift, I guess it is understandable the the top would change (as of height change), but the bottom should remain still

https://github.com/OHIF/Viewers/assets/7490180/eeda8324-1504-4bf8-98d3-9f6d008169e3


3. Having said number 2, I think the menu should accept two more props regarding where to open (bottom or top) and also which direction (left or right) similar to our dropdown UI. I think right now you have implemented it so that it opens from top to bottom and to the right, but for window level we need it opens towards up and to the right. And also if we decide to put it on the top right corner, it should handle top to bottom opening and to the left

4. It says `Back to` I guess this is a bug?
CleanShot 2023-11-08 at 09 44 16

platform/ui/src/components/SwitchButton/SwitchButton.tsx
sedghi: How about checked similar to material ui

```
<Switch
  checked={checked}
  onChange={handleChange}
  inputProps={{ 'aria-label': 'controlled' }}
/>
```

sedghi: https://mui.com/material-ui/react-switch/

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
jbocce: The label is only needed if there is more than one panel and in that case it is used in the PanelSelector as the label for the tab in the selector. But in this case there is only one panel, so I will make it an optional parameter so someone does not need to use it if there is only one panel.

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+  component={AllInOneMenu}
+/>
+
+export const AllInOneMenuTemplate = args => (
+  <div className="w-80">
+    <AllInOneMenu {...args}>
+      <AllInOneMenuItemPanel label="Panel 1">
+        <AllInOneMenuItem
+          label="Item 1"
+          key="0"
```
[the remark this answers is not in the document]
jbocce: It is the react key and is needed to avoid the verbose warnings you see in the console if it is excluded. https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
jbocce: Yes, well the intent was for anyone interested to simply use the `all-in-one-menu-item` class. I believe that is better than have a const or enum of styles. I cleaned up the code so that the custom component you see here only has the one className. Of course, because it is a custom component there will always be custom styling like the height attribute you see here.

platform/ui/src/components/AllInOneMenu/__stories__/allInOneMenu.stories.mdx
jbocce: It is supposed to be used for the 'Back to' text  whenever the sub menu is actually displayed.. However, there appears to be a problem with Story Book picking up the Common.json localization file and I am not sure how to fix it. Perhaps you can give me a hand with that?

That said. I will rename this `backLabel` and maybe that will help.

jbocce: Actually, instead of the Menu and SubMenu components using `useTranslation('Commom')`, I will simply use the `backLabel` to be the "back" text so users can use whatever text they want. That will add flexibility and fix the Story Book problem.

jbocce:
> Great stuff

Thanks.

> 
> 1. I feel like the hover background is wrong, maybe your implementation or maybe even the design. I expected it to fill the region, or at least be symmetric with top and bottom

I will check the design and update accordingly.

> 2. Looks like there is a bottom layout shift, I guess it is understandable the the top would change (as of height change), but the bottom should remain still
> 3. Having said number 2, I think the menu should accept two more props regarding where to open (bottom or top) and also which direction (left or right) similar to our dropdown UI. I think right now you have implemented it so that it opens from top to bottom and to the right, but for window level we need it opens towards up and to the right. And also if we decide to put it on the top right corner, it should handle top to bottom opening and to the left

There are no and should be no implementation details in the menu about how the menu opens because it is JUST a menu and it is up to the code that opens it to position it and thus decide which way it opens. Put another way, the top, left, bottom and right position attributes that it is placed into the DOM with will determine this. This will be evident when it is used in OHIF.

> 4. It says `Back to` I guess this is a bug?

Yes this is a problem with Story Book I think, but I think I found a way around it.

platform/ui/src/components/AllInOneMenu/Menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+type MenuPathState = {
+  props: MenuProps;
+  activePanelIndex: number;
+};
+
+export const MenuContext = createContext<MenuContextProps>(null);
+
+const Menu = (props: MenuProps) => {
+  const { isVisible, onMenuVisibilityChange, activePanelIndex, preventHideMenu, className } = props;
```
sedghi, line 37: onVisibilityChange?

platform/ui/src/components/AllInOneMenu/Menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  const [isMenuVisible, setIsMenuVisible] = useState(false);
```
sedghi, line 39: I thought this should be useState(isVisible), no?

platform/ui/src/components/AllInOneMenu/Menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const [menuPath, setMenuPath] = useState<Array<MenuPathState>>([
```
sedghi, line 40: a line or two lines of doc of what is a path?

platform/ui/src/components/AllInOneMenu/Item.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  return (
+    <div
+      className="all-in-one-menu-item all-in-one-menu-item-effects"
+      onClick={onClickHandler}
+    >
+      {icon && (
+        <>
+          {icon}
+          <div className="w-2"></div>
```
sedghi, line 27: isn't this like padding left/right or something? it is odd to have empty div of w-2

platform/ui/src/components/AllInOneMenu/Item.tsx
```diff
@@ -0,0 +1,38 @@
...
+        </>
+      )}
+      <span>{label}</span>
+      {secondaryLabel != null && (
+        <span className="text-aqua-pale ml-[1ch]">{`${secondaryLabel}`}</span>
```
sedghi, line 32: what is `1ch` never saw that in tailwind

sedghi: I guess you can just say {secondaryLabel}

platform/ui/src/components/AllInOneMenu/SubMenu.tsx
```diff
@@ -0,0 +1,29 @@
...
+    showSubMenu(props);
+    props.onClick?.();
+  }, [showSubMenu, props]);
+
+  return (
+    <div
+      className="all-in-one-menu-item all-in-one-menu-item-effects flex items-center justify-between"
+      onClick={onClickHandler}
+    >
+      <div>{`${props.itemLabel}`}</div>
```
sedghi, line 23: doesn't it work with {props.itemLabel}? maybe I'm wrong

platform/ui/src/components/AllInOneMenu/Menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <div
+          className="all-in-one-menu-item all-in-one-menu-item-effects"
+          onClick={onBackClick}
+        >
+          <Icon name="content-prev"></Icon>
+          <div className="pl-2">{backLabel}</div>
+        </div>
+        <DividerItem></DividerItem>
+      </>
+    );
```
sedghi, line 102: can be moved outside I guess

platform/ui/src/components/AllInOneMenu/Menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+  addItemPanel: (index: number, label: string) => void;
+  activePanelIndex: number;
+};
+
+type MenuPathState = {
+  props: MenuProps;
+  activePanelIndex: number;
+};
+
+export const MenuContext = createContext<MenuContextProps>(null);
```
sedghi, line 34: this probably should be default context not null

platform/ui/src/components/AllInOneMenu/Menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+export interface MenuProps {
+  className?: string;
+  isVisible?: boolean;
+  preventHideMenu?: boolean;
+  backLabel: string;
+  headerComponent?: ReactNode;
+  showHeaderDivider?: boolean;
+  activePanelIndex?: number;
+  onMenuVisibilityChange?: (isVisible: boolean) => void;
+  children: unknown;
```
sedghi, line 19: isn't it always ReactNode for children?

platform/ui/src/components/AllInOneMenu/Menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+    setItemPanelLabels([]);
+    setIsMenuVisible(false);
+    onMenuVisibilityChange(false);
+  }, [preventHideMenu, onMenuVisibilityChange]);
+
+  const addItemPanel = useCallback((index, label) => {
+    setItemPanelLabels(labels => {
+      labels.splice(index, 0, label);
+      return [...labels];
+    });
```
sedghi, line 75: probably wrong 

should be creating a new `updatedLabel = []` instead of mutating the state

platform/ui/src/components/AllInOneMenu/Menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const [itemPanelLabels, setItemPanelLabels] = useState<Array<string>>([]);
+
+  useEffect(() => {
+    setIsMenuVisible(isVisible);
+    onMenuVisibilityChange?.(isVisible);
+  }, [isVisible, onMenuVisibilityChange]);
+
+  const showSubMenu = useCallback((subMenuProps: MenuProps) => {
+    setMenuPath(path => {
+      return [
```
sedghi, line 52: same here, avoid mutation here

sedghi (commented):
great update, see my comments please

platform/ui/src/components/AllInOneMenu/Item.tsx
jbocce: `1ch` is not from tailwind but a pure css unit https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units. It is used here to get a one character separation between the primary and secondary labels.

platform/ui/src/components/AllInOneMenu/Menu.tsx
jbocce: Well a default context would be weird and not meaningful in this case and according to the official React createContext docs, null is acceptable in cases like this: https://react.dev/reference/react/createContext#parameters.

platform/ui/src/components/AllInOneMenu/Menu.tsx
jbocce: If I were mutating the state OUTSIDE of the state update function (i.e. in this case setItemPanelLabels) and passing the mutated state to the update function then that would be wrong. But instead here I am passing a function to the update function that is returning a new, updated state. See here https://www.codingdeft.com/posts/react-usestate-array/#modifying-an-item-in-the-array

platform/ui/src/components/AllInOneMenu/Menu.tsx
jbocce: See https://github.com/OHIF/Viewers/pull/3774#discussion_r1389571562.

jbocce:
I must apologize to you for https://github.com/OHIF/Viewers/pull/3774#issuecomment-1802416428 because although the positioning of the menu is handled by the parent that opens the menu, things like the arrows for the back button submenus need to change depending on whether the menu tends to open left to right or right to left. I will update. Thanks and apologies again.

platform/ui/src/components/AllInOneMenu/Menu.tsx
jbocce: But I suppose I can clean this up a bit.

sedghi (approved):
Looks great, thanks

Issue #2558, opened 2021-09-24
How to load SR Dicom file using json in OHIF viewer.


pmd23:
> :hand: 

we have few dicom images which are **SR images**  I want to be load it using json. So my question is that what kind of json file I need to be generate so that I can able to view that json file into OHIF viewer?

I checked all the questions (open n close) for this as well as official site I did not find any thing. So I am posting this question. 

Need help on this.

sedghi:
Hmmm, I'm not sure what is the level of our support for SEG, SR, RT for dicom JSON. I will investigate

sedghi:
see https://github.com/OHIF/Viewers/pull/3777

sedghi:
We just release the OHIF 3.8, you can find more details here https://ohif.org/release-notes/3p8/
If you still encounter this issue in 3.8, please re-open this.

PR #3777, opened 2023-11-09
feat(dicomJSON): Add Loading Other Display Sets and JSON Metadata Generation script

sedghi:
Addresses https://github.com/OHIF/Viewers/issues/2558

### Context

- Adds couple of fixes to be able to load other displaySets (SEG, RT, SR, etc.) via dicom json data source.
- Adds the most often requested script to generate dicom json metadata (👏 huge thanks to Gradient Health (@Ouwen) that agreed to open source this script)
- 

### Testing

Run the script on the folder it should generate the metadata

If you really want to test fully you should upload studies to some server to test, but that part works I have tested.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


platform/docs/docs/configuration/url.md
```diff
@@ -62,7 +62,9 @@ WorkList by adding the `dataSources` query parameter.
 /?dataSources=orthanc
 ```
 
-Note: you should pass the `sourceName` of the data source in the configuration file (not the friendly name nor the name)
+Note1: You should pass the `sourceName` of the data source in the configuration file (not the friendly name nor the name)
+Note2: Make sure that the configuration file you are using actually include that data source. You cannot use a data source from other configuration file.
```
jbocce, line 66: *includes

*another (instead of other)

.scripts/dicom-json-generator.js
```diff
@@ -0,0 +1,264 @@
+/*
+ * This script uses nodejs to generate a JSON file from a DICOM study folder.
+ * You need to have dcmjs installed in your project.
+ * The JSON file will can be used to load the study into the OHIF Viewer. You can get more detail
```
jbocce, line 4: remove *will

.scripts/dicom-json-generator.js
```diff
@@ -0,0 +1,264 @@
...
+  };
+  return { metadata, url: instance.fileLocation };
+}
+
+function createInstanceMetaDataMultiFrame(instance) {
+  const instances = [];
+  const commonData = commonMetaData(instance);
+  const conditionalData = conditionalMetaData(instance);
+
+  for (let i = 1; i < instance.NumberOfFrames; i++) {
```
jbocce, line 256: Just double checking that this shouldn't instead be `i <= instance.NumberOfFrames` since we are starting at one, but I might be wrong.

platform/core/src/utils/addAccessors.js
```diff
@@ -0,0 +1,68 @@
...
+      return true;
+    }
+    if (prop in target) {
+      return target[prop];
+    }
+    return target[0][prop];
+  },
+
+  set: (obj, prop, value) => {
+    if (typeof prop === 'number') {
```
jbocce, line 18: Does it make sense to use an || here instead of an else if?

jbocce:
I tried running the script and found this...

If I run the script on a Windows Powershell command prompt the urls look something like this...
`"url": "dicomweb:http://localhost:3000/dicomjson/\\1-001.dcm"`

If I run using a wsl ubuntu prompt the urls look something like this...
`"url": "dicomweb:http://localhost:3000/dicomjson/mnt/c/dicom/acrin/1-001.dcm"`

I would have expected both to be...
`"url": "dicomweb:http://localhost:3000/dicomjson/1-001.dcm"`

jbocce (changes requested):
A nice addition for sure! See my comments.

jbocce (approved):
Approved.

jbocce (approved):
Approved.

PR #3788, opened 2023-11-15
Merge Data Source

igoroctaviano:
### Context

A new data source that allows merging one or more data sources at a series level.
Implementation can be expended later to allow the merge at the study level.

### Changes & Results

- Add new merge data source
- Attach new merge data source to default extension

### Testing

```
window.config = {
  ...
  dataSources: [
    {
      sourceName: 'merge',
      namespace: '@ohif/extension-default.dataSourcesModule.merge',
      configuration: {
        name: 'merge',
        friendlyName: 'Merge dicomweb-1 and dicomweb-2 data at series level',
        seriesMerge: {
          dataSourceNames: ['dicomweb-1', 'dicomweb-2'],
        },
      },
    },
    {
      sourceName: 'dicomweb-1',
      ...
    },
    {
      sourceName: 'dicomweb-2',
      ...
    },
  ],
};

```

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


platform/core/src/services/DicomMetadataStore/DicomMetadataStore.ts
```diff
@@ -186,6 +186,13 @@ const BaseImplementation = {
       madeInClient,
     });
   },
+  updateSeriesMetadata(seriesMetadata) {
+    const { StudyInstanceUID, SeriesInstanceUID } = seriesMetadata;
+    const study = _getStudy(StudyInstanceUID);
+    if (study) {
+      study.setSeriesMetadata(SeriesInstanceUID, seriesMetadata);
```
sedghi, line 193: what about if there is no series with that UID

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,140 @@
...
+          if (mergeKey && level !== -1) {
+            const dedupedData = getDedupedData({
+              globalData: globalData.flat(),
+              incomingData,
+              mergeKey,
+              level,
+            });
+            data = data.concat(dedupedData);
+          } else {
+            data = data.concat(data);
```
sedghi, line 58: should this be `data = data.concat(incomingData)`?

extensions/default/src/MergeDataSource/utils.js
```diff
@@ -0,0 +1,58 @@
...
+ * @returns {*} - The value of the key if found, otherwise undefined.
+ */
+export function getKeyByLevel(obj, key, level) {
+  if (level === 0) {
+    return obj[key];
+  }
+  const keys = Object.keys(obj);
+  for (let i = 0; i < keys.length; i++) {
+    const currentKey = keys[i];
+    if (typeof obj[currentKey] === 'object') {
```
sedghi, line 40: && obj[currentKey] !== null

extensions/default/src/MergeDataSource/utils.js
```diff
@@ -0,0 +1,58 @@
...
+ * @param {number} level - The level at which to search for the key.
+ * @returns {*} - The value of the key if found, otherwise undefined.
+ */
+export function getKeyByLevel(obj, key, level) {
+  if (level === 0) {
+    return obj[key];
+  }
+  const keys = Object.keys(obj);
+  for (let i = 0; i < keys.length; i++) {
+    const currentKey = keys[i];
```
sedghi, line 39: how about for of

extensions/default/src/MergeDataSource/index.js
igoroctaviano: Updated, thanks!

platform/core/src/services/DicomMetadataStore/DicomMetadataStore.ts
igoroctaviano: Updated, thanks!

extensions/default/src/MergeDataSource/utils.js
igoroctaviano: Updated, thanks!

extensions/default/src/MergeDataSource/utils.js
igoroctaviano: Updated, thanks!

sedghi (commented):
looks good see my comments

extensions/default/src/DicomWebDataSource/retrieveStudyMetadata.js
```diff
@@ -38,9 +39,11 @@ export function retrieveStudyMetadata(
     throw new Error(`${moduleName}: Required 'StudyInstanceUID' parameter not provided.`);
   }
 
+  const promiseId = (dicomWebConfig.name || '') + StudyInstanceUID;
```
wayfarer3130, line 42: In general making it look URI ish is nice => how about `${dicomWebConfig.name}:${StudyInstanceUID}`

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+          return series;
+        },
+      },
+    };
+
+    const { mergeKey, level, tagFunc } = mergeMap[path] || { tagFunc: x => x, level: -1 };
+
+    const dataSourceDefs = Object.values(extensionManager.dataSourceDefs);
+    const promises = [];
+    let globalData = [];
```
wayfarer3130, line 82: globalData can be const since you add to it but don't reassign it
Also, an easy way to key it for uniqueness is
const globalData = new Map<...>()
then set the values by whatever key they use, and at the end, 
return [...globalData.values()]

Also, can we call is mergedData rather than globalData?

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+
+    for (const dataSourceDef of dataSourceDefs) {
+      const { configuration, sourceName } = dataSourceDef;
+      if (configuration && dataSourceNames.includes(sourceName)) {
+        const [dataSource] = extensionManager.getDataSources(sourceName);
+        const promise = executeFunction(dataSource, path, args);
+        promises.push(promise);
+
+        promise.then(data =>
```
wayfarer3130, line 91: You need to promises.push(promise.then(data => ....
Otherwise you have introduced a very subtle bug when the allSettled can be complete, but the merge is not, and the length ends up changing after the response is returned.

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+   * @param {string} path - The path of the function to call on each data source.
+   * @param {Array} args - The arguments to pass to the function on each data source.
+   * @returns {Array} - An array of the results from each data source.
+   */
+  const callForAllDataSources = (path, args) => {
+    const dataSourceDefs = Object.values(extensionManager.dataSourceDefs);
+    const globalData = [];
+    for (const dataSourceDef of dataSourceDefs) {
+      const { configuration, sourceName } = dataSourceDef;
+      if (configuration && dataSourceNames.includes(sourceName)) {
```
wayfarer3130, line 122: This should be === as I think the includes sourceName is tricky to ensure you get it right.  
If you want a way to group calls, then add a parameter to the configuration that allows you to compare for exact equality on that parameter value.  You could make that parameter keyed so that someone could create a new/arbitrary key for this.

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+  };
+
+  /**
+   * Calls the default data source with the given path and arguments.
+   * @param {string} path - The path to the function to be executed.
+   * @param {Array} args - The arguments to be passed to the function.
+   * @returns {*} - The result of executing the function.
+   */
+  const callForDefaultDataSource = (path, args) => {
+    const defaultDataSourceName = dataSourceNames[0];
```
wayfarer3130, line 138: This is NOT the default - the default is specifically specified by name as a parameter 'defaultDataSourceName' or some such value.  Also, the default might be specific to the merge configuration, so I think you should have two levels of default to allow the merge configuration to specify a specific default.

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+   * @param {Array} args - The arguments to pass to the method.
+   * @returns {*} - The result of calling the method.
+   */
+  const callByRetrieveAETitle = (path, args) => {
+    const [displaySet] = args;
+    const seriesMetadata = DicomMetadataStore.getSeries(
+      displaySet.StudyInstanceUID,
+      displaySet.SeriesInstanceUID
+    );
+    const defaultDataSourceName = dataSourceNames[0];
```
wayfarer3130, line 155: Suggest adding a getter to get your default data source name based on the current circumstances.

wayfarer3130: Or a const that you compute up front.

extensions/default/src/MergeDataSource/utils.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * Executes a function given an object and a path to the function.
+ * @param {Object} obj - The object containing the function to be executed.
+ * @param {string} path - The path to the function, separated by dots.
+ * @param {Array} args - An array of arguments to be passed to the function.
+ * @returns {*} - The result of the executed function.
+ */
+export function executeFunction(obj, path, args) {
```
wayfarer3130, line 8: Consider if you can use the lodash get method for this.

extensions/default/src/MergeDataSource/utils.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {string} key - The key to search for.
+ * @param {number} level - The level at which to search for the key.
+ * @returns {*} - The value of the key if found, otherwise undefined.
+ */
+export function getKeyByLevel(obj, key, level) {
+  if (level === 0) {
+    return obj[key];
+  }
+  const keys = Object.keys(obj);
+  for (const currentKey of keys) {
```
wayfarer3130, line 38: The keys aren't hierarchical, they are a flat denormalized structure, so return obj[key] is sufficient.

wayfarer3130 (changes requested):
A number of smaller changes, but this looks like it is on the right track.

extensions/default/src/MergeDataSource/index.js
igoroctaviano: The includes here is running against an array, and in that case, it's the same as ===. You can test it by running ['xpto'].includes('xpt') which will return false.

extensions/default/src/DicomWebDataSource/retrieveStudyMetadata.js
igoroctaviano: Updated, thanks!

extensions/default/src/MergeDataSource/index.js
igoroctaviano: Updated, thanks!

extensions/default/src/MergeDataSource/index.js
igoroctaviano: Updated, thanks!

extensions/default/src/MergeDataSource/index.js
igoroctaviano: Updated, thanks!

extensions/default/src/MergeDataSource/utils.js
igoroctaviano: Updated, thanks!

extensions/default/src/MergeDataSource/utils.js
igoroctaviano: Updated, thanks!

platform/docs/docs/platform/extensions/modules/data-source.md
```diff
@@ -172,3 +169,44 @@ extensionManager.updateDataSourceConfiguration( "dicomweb",
   },
 );
 ```
+
+## Merge Data Source
```
sedghi, line 173: Love this thanks

platform/docs/docs/platform/extensions/modules/data-source.md
```diff
@@ -172,3 +169,44 @@ extensionManager.updateDataSourceConfiguration( "dicomweb",
...
+The built-in merge data source is a useful tool for combining results from multiple data sources.
+Currently, this data source only supports merging at the series level. This means that series from data source 'A'
+and series from data source 'B' will be retrieved under the same study. If the same series exists in both data sources,
+the first series retrieved will be used, and any other duplicated series will be ignored.
```
sedghi, line 177: this is ambigious wording, do you mean whoever arrives first?

sedghi (approved):

platform/docs/docs/platform/extensions/modules/data-source.md
igoroctaviano: Updated wording, thanks!

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
+import { DicomMetadataStore, IWebApiDataSource } from '@ohif/core';
```
wayfarer3130, line 1: Could you move this to typescript - it would make figuring out a few of the calls a bit easier, as I wasn't always sure of the types.

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+   * Calls a specified function on all data sources and merges the results.
+   *
+   * @async
+   * @function
+   * @param {string} path - The path of the function to call.
+   * @param {Object} args - The arguments to pass to the function.
+   * @returns {Promise<Array>} - A promise that resolves to an array of merged data.
+   */
+  const callForAllDataSourcesAsync = async (path, args) => {
+    const mergeMap = {
```
wayfarer3130, line 55: The merge map can be made global/static as nothing in it is specific to the function itself.

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+ * Returns deduplicated data by comparing the incomingData with the mergedData based on the mergeKey.
+ * @param {Object} options - The options object.
+ * @param {Array} options.mergedData - The global data array.
+ * @param {Array} options.incomingData - The incoming data array.
+ * @param {string} options.mergeKey - The merge key.
+ * @returns {Array} The deduplicated data array.
+ */
+const getDedupedData = ({ mergedData, incomingData, mergeKey }) => {
+  const keys = mergedData.map(r => get(r, mergeKey));
+  return incomingData.filter(r => !keys.includes(get(r, mergeKey)));
```
wayfarer3130, line 14: It is quite a bit more efficient to do
mergedData.has(get(r,mergeKey)) instead of keys.includes, which performs a linear search.  On 2000 item series, that is 2,000,000 comparisons - definitely something to avoid if possible.

wayfarer3130: What you end up doing is const key = get(r,mergeKey);  mergedData.set(key, r)  You can then also add it to a return list, OR you can choose which version to include based on other conditions (future work)

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+ * @param {Array} options.mergedData - The global data array.
+ * @param {Array} options.incomingData - The incoming data array.
+ * @param {string} options.mergeKey - The key to merge the data on.
+ * @param {Function} options.tagFunc - The function to tag the data with.
+ */
+const processDedupedData = ({ sourceName, mergedData, incomingData, mergeKey, tagFunc }) => {
+  let data = [];
+  if (mergeKey) {
+    const dedupedData = getDedupedData({
+      mergedData: mergedData.flat(),
```
wayfarer3130, line 30: You should not be calling flat here - the UID keys are always at the top level.  Anything below that is extraneous and might result in missing data if a sub-sequence has an item in it.  Also makes it REALLY slow because in DICOM JSON you will have 1000+ lists in each instance, so you can end up comparing tens of thousands of instances in a linear compare.

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+      series: {
+        search: (...args) => callForAllDataSourcesAsync('query.series.search', args),
+      },
+      instances: {
+        search: (...args) => callForAllDataSourcesAsync('query.instances.search', args),
+      },
+    },
+    retrieve: {
+      bulkDataURI: (...args) => callForAllDataSourcesAsync('retrieve.series.metadata', args),
+      directURL: (...args) => callForAllDataSourcesAsync('retrieve.directURL', args),
```
wayfarer3130, line 175: This one will only be for the specified data source.

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+      series: {
+        metadata: (...args) => callForAllDataSourcesAsync('retrieve.series.metadata', args),
```
wayfarer3130, line 177: The metadata will be for a single data source unless you merge multiple same series into one.

extensions/default/src/MergeDataSource/index.js
```diff
@@ -0,0 +1,195 @@
...
+   * @returns {Promise<Array>} - A promise that resolves to an array of merged data.
+   */
+  const callForAllDataSourcesAsync = async (path, args) => {
+    const mergeMap = {
+      'query.studies.search': {
+        mergeKey: 'studyInstanceUid',
+        /** TODO */
+        tagFunc: x => x,
+      },
+      'retrieve.series.metadata': {
```
wayfarer3130, line 61: This one had better be 'query.series.search': {....
Once you have merged the series, then you don't merge metadata instances.

extensions/default/src/MergeDataSource/index.js
igoroctaviano: I'm not sure about has, this is not part of array prototype, can you sketch some code sample?

extensions/default/src/MergeDataSource/index.js
igoroctaviano: Can you suggest where to execute flat then? this call is called once per series. I'm not sure why this will be called for every instance.

extensions/default/src/MergeDataSource/index.js
igoroctaviano: Updated, thanks.

extensions/default/src/MergeDataSource/index.js
igoroctaviano: Updated, thanks.

extensions/default/src/MergeDataSource/index.js
igoroctaviano: Updated, thanks.

extensions/default/src/MergeDataSource/index.js
igoroctaviano: This method checks for retrieveAETitle and filters out the other data sources, in the end is going to be one dataSource. This is necessary since we don't know which data source to call beforehand.

extensions/default/src/MergeDataSource/index.js
igoroctaviano: Updated, thanks.

extensions/default/src/MergeDataSource/index.test.js
wayfarer3130, line 1: Thank you for adding tests for this - that will be important as we add more functionality

sedghi (approved):

wayfarer3130 (approved):
LGTM - this looks like a solid base to build more advanced handling of different types.

Issue #3843, opened 2023-12-07
[Bug]  Green screen


narbek-abd:
### Describe the Bug

Hi.
Here is an archive.
[e5824449-2d66de43-8459fe70-9ff5359c-97655c13.zip](https://github.com/OHIF/Viewers/files/13600013/e5824449-2d66de43-8459fe70-9ff5359c-97655c13.zip)

After this commit https://github.com/OHIF/Viewers/commit/6651008fbb35dabd5991c7f61128e6ef324012df  i see green screen for this study. In local mode it works, but I I use osimis/orthanc:22.11.3 and I downloaded source code before and after this commit and it was ok before  this commit, so i don't think it is an issue with Orthanc

### Steps to Reproduce

1. use osimis/orthanc:22.11.3
2. download source code before provided commit
3. open the study

### The current behavior

green screen

### The expected behavior

proper viewer

### OS

all

### Node version

all

### Browser

all

niirdan:
same issue here , using google.js configuration latest version .
there is an estimated time for a fix  ?

sedghi:
I don't think it is related to that commit honestly. This is a long lasting bug in Orthanc that we tried to communicate to fix, but they fixed and later reverted the fix. https://discourse.orthanc-server.org/t/orthanc-convert-ybr-to-rgb-but-does-not-change-metadata/3533/16

sedghi:
I posted a comment there https://discourse.orthanc-server.org/t/orthanc-convert-ybr-to-rgb-but-does-not-change-metadata/3533/19?u=alireza

niirdan:
hey @sedghi thanks for your answer, this is very appreciated ! how does orthanc works inside ohif ? is it a remote server that the ohif sends a request or this is a static node module , because if it is a node module , maybe the temporary solution is to downgrade it to the version with the fix that they released ?

sedghi:
seems like this will help

https://discourse.orthanc-server.org/t/orthanc-convert-ybr-to-rgb-but-does-not-change-metadata/3533/20?u=alireza

niirdan:
> seems like this will help
> 
> https://discourse.orthanc-server.org/t/orthanc-convert-ybr-to-rgb-but-does-not-change-metadata/3533/20?u=alireza

this actually fix the issue !!! many thanks for mentioning! 
please considerate fixing google.js in the next branch with this fix

sedghi:
Checkout lastest master branch, it should have been fixed with the recent push

[...]

PR #3883, opened 2024-01-08
feat(transferSyntax): prefer server transcoded transfer syntax for all images

sedghi:
Fixes issues like https://github.com/OHIF/Viewers/issues/3843 and https://github.com/cornerstonejs/cornerstone3D/pull/829

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


platform/app/public/config/local_orthanc.js
```diff
@@ -24,6 +24,7 @@ window.config = {
         wadoUriRoot: 'http://localhost/dicom-web',
         qidoRoot: 'http://localhost/dicom-web',
         wadoRoot: 'http://localhost/dicom-web',
+        acceptHeader: ['multipart/related; type=application/octet-stream; transfer-syntax=*'],
```
wayfarer3130, line 27: I think this should be encoded as the default, with appropriate quotes added, rather than embedding here everywhere.

wayfarer3130: Tweak generateAcceptHeaders to have a acceptHeader.push('transfer-syntax=*')
OR set the requestTransferSyntax to be * by default, and if it is set to false then remove

salimkanoun:
Just for information there is a less verbose option to get images in original TS

requestTransferSyntaxUID: "*"

sedghi:
> Just for information there is a less verbose option to get images in original TS
> requestTransferSyntaxUID: "*"

@wayfarer3130 What do you think

wayfarer3130:
> > Just for information there is a less verbose option to get images in original TS
> > requestTransferSyntaxUID: "*"
> 
> @wayfarer3130 What do you think

Definitely prefer requestTransferSyntaxUID: "*", but I'd prefer even more not needing requestTransferSyntaxUID, and defaulting to "*" - and if it is false or empty, then not using it at all.

platform/core/src/utils/generateAcceptHeader.ts
```diff
@@ -17,6 +17,7 @@ const generateAcceptHeader = (
     acceptHeader.push('type=application/octet-stream');
   }
 
+  acceptHeader.push('transfer-syntax=*');
```
wayfarer3130, line 20: This will add a second transfer-syntax=* header.  I think you need to set requestTransferSyntaxUID to '*' when it is undefined.

wayfarer3130 (approved):

Issue #3802, opened 2023-11-23
OHIF About Info for OS is incorrect while using a Mac


tzuccher:
### Describe the Bug

OHIF About Info for OS is incorrect while using a Mac. The OS is reported as OS X 10.15.7
This Mac OS is incorrect as my MacBook Pro version is Sonoma 14.1.1

![image](https://github.com/OHIF/Viewers/assets/143200290/e28c0d19-d792-4c69-a2ca-583483b8d376)

![image](https://github.com/OHIF/Viewers/assets/143200290/1410c263-71cd-47e5-8f13-e2856f0fbd38)

### Steps to Reproduce

This was found in both of the following environments:
https://viewer.ohif.org/ - version 3.7.0
https://viewer-dev.ohif.org/ - version 3.8.0-beta.21

1. Select the wheel icon or the drop-down icon from the upper right corner of the OHIF screen(s)
2. Select the About option
3. Verify the Browser & OS info = the OS info on my Mac displays: OS X 10.15.7 however, my Mac version is Sonoma 14.1.1

### The current behavior

The About info is reporting the incorrect OS version

### The expected behavior

The About info should report the correct OS (& browser info)...or, no info at all 

### OS

MacBook Pro - Sonoma 14.1.1

### Node version

unknown

### Browser

Chrome v119

sedghi:
Seems like an intentional security approach by apple, they have stopped properly mentioning the os version 

https://discussions.apple.com/thread/253178691

https://www.otsukare.info/2021/02/15/capping-macos-user-agent

[Other detectors also have issues](https://bestiejs.github.io/platform.js/), so i'm just removing the version

sedghi:
We just release the OHIF 3.8, you can find more details here https://ohif.org/release-notes/3p8/
If you still encounter this issue in 3.8, please re-open this.

Issue #3520, opened 2023-07-05
[Bug] Image Wrong Size in the viewport


francomotta:
### Describe the Bug

When loading a study, the images are the wrong size in the viewport.

### Steps to Reproduce

Select the TCGA-QQ-A8VH study, select the 1x3 layout. Notice that the first viewport image displays the image at the wrong size (big). If you hit the "=" key, it makes the image the correct size. The same problem occurs if you double-click the image to get it to 1x1 format.

The right thing would be to put an option after opening the images to adjust the size in the viewport.
![Wrong size-2023-07-05 140722](https://github.com/OHIF/Viewers/assets/62942118/e484825c-5f8e-4bb9-817d-b0f67612a280)

### The current behavior

Note that the first image is the wrong size.
![Wrong size-2023-07-05 140722](https://github.com/OHIF/Viewers/assets/62942118/4075bdca-cef1-445e-90e8-604cc0efadb1)

### The expected behavior

The right thing would be to put an option after opening the images to adjust the size in the viewport.
![Correct size-2023-07-05 140722](https://github.com/OHIF/Viewers/assets/62942118/8a8c8f2f-7897-47eb-8096-aedfeab776a7)

### OS

Windows 11

### Node version

20.3.0

### Browser

Chrome

francomotta:
@jbocce , how are you?

Even with that tip, the error continues. Occurs most frequently in the 1x3 format with MR.

sedghi:
Bill recently merged a fix for resize, maybe this is fixed now?

sedghi:
So this is a tricky one, since we moved away from our previous setup where we re-created the viewports, and now the viewport (html div) would MOVE to the new location (for performance gain) and since there is no extra render, the camera is not changed. I don't think this is a sever problem, but if you think it is let us know

sedghi:
So in tmtv mode for instance, if we move from one stage to another, the old viewport (image) sized is preserved as of the new way of assiging Ids to viewports and not index (React just moves the element and canvas to the next stage, which is really super performant, but comes at the cost that we get wrong size, and we should trigger a resize maybe after the layout change

![CleanShot 2023-10-18 at 10 43 06](https://github.com/OHIF/Viewers/assets/7490180/83165b2f-d78f-458f-86d1-0a2f554c8cda)

Srikanthkurva:
Hi
I have successfully running this "https://github.com/OHIF/Viewers?tab=readme-ov-file". I  have one DCOM Images 
[...]

PR #3889, opened 2024-01-10
feat(resize): Optimize resizing process and maintain zoom level

sedghi:
will address this https://github.com/OHIF/Viewers/issues/3520 and https://github.com/OHIF/Viewers/issues/3802 and https://github.com/OHIF/Viewers/issues/3943

### Context

This PR will solve couple of issues regarding how viewports are resized. 

1. Opening/closing pannels previously were triggering a lot of resizes for each viewports (for tmtv 10 resize calls), with this PR we are changing the way we call resize on the grid, and only calling it once. That gives us a performance boost so that we can then smoothly resize the viewport. 

2. If the viewport is not touched by the tools (zoom) the change of the viewport size (layout) should keep the zoom 1 for the viewport, which was not happening before.

**Before**
https://github.com/OHIF/Viewers/assets/7490180/5bc8908a-237a-4195-92c7-fd11c323ccdb

**After**
https://github.com/OHIF/Viewers/assets/7490180/e218e522-d905-4aa7-b0e1-cb12350755d7

## Left to fix

We need to also fix the ugly transition for viewport size changes, but that is trickier

### Testing

Look at the videos

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -94,14 +100,84 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
+  }
+
+  private performResize() {
+    const isImmediate = false;
+
+    const viewports = this.getRenderingEngine().getViewports();
+
+    // Store the zoom scale before resizing.
+    viewports.forEach(viewport => {
+      const zoomLevel = viewport.getZoom();
```
wayfarer3130, line 153: Could you instead break up getPresentation into two child functions (or maybe more eventually)  getPositionPresentation and getLutPresentation, and then use getPositionPresentation to fetch:
zoom,
pan,
viewUp,
viewNormal,

and then use the set on that to restore the position presentation to a viewport after resize?
the pan values will have to be normalized to the total canvas size rather than being in canvas units
That will give consistent handling with restoring a viewport after it hasn't been displayed, or using a viewport to display another display set.  
That way as we improve the display positioning memory, it will get applied repeatedly in teh same way.

I think this is a better solution than remembering the camera itself

platform/app/src/components/ViewportGrid.tsx
```diff
@@ -342,13 +344,26 @@ function ViewerViewportGrid(props) {
...
+      className="h-full w-full"
     >
-      {/* {ViewportPanes} */}
-      {getViewportPanes()}
-    </ViewportGrid>
+      <ViewportGrid
+        numRows={numRows}
+        numCols={numCols}
+      >
+        <ReactResizeDetector
```
wayfarer3130, line 355: This definitely seems to be worthwhile to resize the entire grid at once.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
sedghi: Done

wayfarer3130 (approved):
I think this change is good now as it is - probably something that will be worked on further as CS3D gets better control over resize/layout

PR #3894, opened 2024-01-12
fix: Microscopy bulkdata and image retrieve

wayfarer3130:
### Context

Some backends incorrectly report the type of the bulkdata.  This PR checks the bulkdata contents to see if it is multipart, and if so, demultiparts it when doing the retrieveBulkdata
Also, for many back ends, using the generic */* accept type works fine for retrieval, and that can be faster than having a full retrieve list.  

### Changes & Results

For all back ends, unless configured otherwise, retrieve bulkdata and de-multipart it
For static wado back ends, use '*/*' as the retrieve condition.

### Testing

Configure a back end with and without staticWado true in the datasource (eg datasources='ohif' has it true)
Display a microscopy study containing bulkdata incorrectly returned as 'application/octet-stream' - the SM modality studies on 'ohif' are examples.
The SM should display and the accept header should be '*/*'

If you remove the staticWado setting, then the SM should still display, and against a system which returns uncompressed data for images, should still return correctly.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
+++ b/extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
@@ -10,7 +10,7 @@ if (!anyDicomwebClient._orig_buildMultipartAcceptHeaderFieldValue) {
   anyDicomwebClient._orig_buildMultipartAcceptHeaderFieldValue =
     anyDicomwebClient._buildMultipartAcceptHeaderFieldValue;
   anyDicomwebClient._buildMultipartAcceptHeaderFieldValue = function (mediaTypes, acceptableTypes) {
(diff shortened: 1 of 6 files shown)


sedghi:
@wayfarer3130 Can you please put a PR to dicomweb-client for these changed, I know there might be some push back but we need to at least have the PR open before we go in this direction

sedghi:
@igoroctaviano Could you do me a favor and test this branch against IDC hosted SM to see if everything is working fine? It should work

sedghi:
I tested on local orthanc and dcm4chee and works fine

sedghi (approved):

Issue #3825, opened 2023-11-29
Toolbar Overhaul for Patient and More Info


james-hanks:
As a result of the new viewport layout and for better visibility we need to move patient info from the viewport corner to the toolbar.

Requirements:

* There shall be a patient info icon in the toolbar 
* Clicking the patient info icon shall display patient information in line in the toolbar
* Clicking the patient info icon shall hide patient information

Deploy Requirements:
* There shall be two options for deployment:
* * Patient info hidden by default (this will be the default OHIF deploy setting)
* * Patient info displayed by default
* The ability for end user to click the patient info icon to display/hide patient information shall be enabled in each deploy scenario

Epic #3493 

Final design (added Feb 20, @dan-rukas )

![Image](https://github.com/OHIF/Viewers/assets/2745168/d3ae4857-a199-4951-a645-4cff002ff54a)

![Image](https://github.com/OHIF/Viewers/assets/2745168/e2525804-b884-4d59-826f-34de979b2ada)

dan-rukas:
Revised and updated specs for toolbar component layout and new patient info component:

![Image](https://github.com/OHIF/Viewers/assets/2745168/d3ae4857-a199-4951-a645-4cff002ff54a)

![Image](https://github.com/OHIF/Viewers/assets/2745168/e2525804-b884-4d59-826f-34de979b2ada)

dan-rukas:
Attached is ZIP that includes the new "multiple patients" icon along with the other navigation icons in the toolbar (no changes)

[toolbar-navigation-icons-Feb23.zip](https://github.com/OHIF/Viewers/files/14386131/toolbar-navigation-icons-Feb23.zip)

sedghi:
We just release the OHIF 3.8, you can find more details here https://ohif.org/release-notes/3p8/
If you still encounter this issue in 3.8, please re-open this.

PR #3929, opened 2024-02-12
feat(header): new patient info header

IbrahimCSAE:
### Context

Add new patient info bar to the header.

Addresses: https://github.com/OHIF/Viewers/issues/3825#issue-2017283497

### Changes & Results

- Add new HeaderPatientInfo component
- Add new configuration flag

image
image


platform/app/public/config/default.js
```diff
@@ -21,6 +21,7 @@ window.config = {
   },
   // filterQueryParam: false,
   defaultDataSourceName: 'dicomweb',
+  showPatientInfo: true,
```
sedghi, line 24: This should have mor options, should be
- `visibleCollapsed`: shows but collapsed (our default) no need to put here

Documented
- `disabled`: does not even show anything
- `visible`: visible (not collapsed)

platform/ui/src/components/HeaderPatientInfo/HeaderPatientInfo.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const updatePatientInfo = () => {
+    const { activeViewportId, viewports } = viewportGridService.getState();
+    const activeViewport = viewports.get(activeViewportId);
+    const displaySetInstanceUID = activeViewport?.displaySetInstanceUIDs[0];
+    if (!displaySetInstanceUID) {
+      return;
+    }
+    const displaySet = displaySetService.getDisplaySetByUID(displaySetInstanceUID);
+    const image0 = displaySet.images[0];
```
sedghi, line 30: not all displaySets have images, try with dicom pdf, it will fail most likely or a video (OT) one

sedghi (commented):
see my comments

platform/ui/src/components/HeaderPatientInfo/HeaderPatientInfo.tsx
salimkanoun: Hi all, I think it would be better to check that all display sets beings rendered belongs to the same patient, and warn it if not.

The common and dangerous risk is to mix patient during image comparison.
So if display sets does not all belongs to a same patient, better to not display any name (or better a "mixed patient" warning).

If you display the name of the first display set, a physician can wrongly believe are belonging to a single patient

salimkanoun: It would require to check all display set of all viewport rather than taking the active viewport (as the toolbar is common to all viewports)

sedghi: I agree, we can have a fallback in that case to show the patientId on the viewport as an overlay with two different colors

sedghi: At least we should show warning, maybe we implement overlay later

IbrahimCSAE: Just implemented this, you can test it locally

http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.3023.4024.215308722288168917637555384485&StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5

do a 3x3 grid, warning should show up, if you drop a series from the same study into all the viewports it will go away.

I also fixed the issue for OT and other modalities with no images.

If we decide to change our mind about this, there's code to also change the patient name based on the active viewport id.

IbrahimCSAE:
@sedghi I addressed the comments

platform/ui/src/components/Header/Header.tsx
```diff
@@ -58,7 +61,11 @@ function Header({
         </div>
         <div className="flex items-center">{children}</div>
         <div className="flex items-center">
-          <span className="text-common-light mr-3 text-lg">{t('INVESTIGATIONAL USE ONLY')}</span>
+          {(showPatientInfo === 'visible' || showPatientInfo === 'visibleCollapsed') && (
```
sedghi, line 64: 'visible' , 'visibleCollapsed' etc should be enums

platform/ui/src/components/HeaderPatientInfo/HeaderPatientInfo.tsx
```diff
@@ -0,0 +1,152 @@
...
+      if (!instance) {
+        return;
+      }
+
+      if (instance.PatientID !== activeViewportPatientID) {
+        isMixedPatients = true;
+      }
+    });
+    setIsMixedPatients(isMixedPatients);
+  };
```
sedghi, line 43: I guess you should just list all DSs and look if there is inconsistency, not just viewports. To be careful here

platform/ui/src/components/HeaderPatientInfo/HeaderPatientInfo.tsx
```diff
@@ -0,0 +1,152 @@
...
+      viewportGridService.subscribe(
+        viewportGridService.EVENTS.ACTIVE_VIEWPORT_ID_CHANGED,
+        event => updatePatientInfo(event.viewportId) // Update the toolbar when the active viewport changes
+      ).unsubscribe,
+      cornerstoneViewportService.subscribe(
+        cornerstoneViewportService.EVENTS.VIEWPORT_DATA_CHANGED,
+        () => updatePatientInfo() // Update the toolbar when the viewport data changes
+      ).unsubscribe,
+    ];
+
```
sedghi, line 83: this can be re-written as 

const events = [... ,...]
events.forEach(event, () => {})

sedghi (commented):
thanks, see my comments

sedghi (approved):

PR #3935, opened 2024-02-15
Enable metadata loading from images

gpasquie:
### Context

The goal of this PR is to allow the MetadataProvider to read the metadata of dicom images directly without making an extra WADO-RS Metadata request to the PACS.
Internaly it uses dataSetCacheManager of dicomImageLoader to access the DICOM dataset. Then dcmjs parse the dataset and add the metadata to the MetadataStore.
This PR allow the DicomJSON source to provide only the image UID + instance number without the need of putting all the image metadata into the final .json.

You can enable or disable this behavior with the variable fullmetadataset in the JSON datasource.

### Testing

This can be tested using a regular JSON DataSource where only SeriesInstanceUID, StudyInstanceUID, SOPClassUID and InstanceNumber must be present.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 10 
- [x] Browser: Chrome  121.0.6167.161
  

### DIFF
--- a/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
+++ b/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
@@ -356,6 +356,25 @@ const OHIFCornerstoneViewport = React.memo(props => {
     };
   }, [displaySets, elementRef, viewportId]);
 
+  useEffect(() => {
+    const SubscriptionDisplaySetsChanged = displaySetService.subscribe(
+      displaySetService.EVENTS.DISPLAY_SETS_CHANGED,
+      async (changedDisplaySets) => {
+        var displaySets = [];
+        displaySets.push(changedDisplaySets);
+        const newViewportData = await cornerstoneCacheService.createViewportData(
+          displaySets,
+          viewportOptions,
+          dataSource,
+          initialImageIndex
+        );
+
+        const keepCamera = true;
+        cornerstoneViewportService.updateViewport(viewportId, newViewportData, keepCamera);
+      }
+    );
+  }, [displaySets]);
+
   return (
     <React.Fragment>
       <div className="viewport-wrapper">
--- a/extensions/cornerstone/src/init.tsx
+++ b/extensions/cornerstone/src/init.tsx
@@ -151,6 +151,7 @@ export default async function init({
   });
 
   const metadataProvider = OHIF.classes.MetadataProvider;
+  metadataProvider.setLoadDicomMetadata(appConfig.automaticallyLoadDicomMetadata)
 
   volumeLoader.registerVolumeLoader(
     'cornerstoneStreamingImageVolume',
--- a/extensions/cornerstone/src/init.tsx
+++ b/extensions/cornerstone/src/init.tsx
@@ -151,7 +151,6 @@ export default async function init({
   });
 
   const metadataProvider = OHIF.classes.MetadataProvider;
-  metadataProvider.setLoadDicomMetadata(appConfig.automaticallyLoadDicomMetadata)
 
   volumeLoader.registerVolumeLoader(
     'cornerstoneStreamingImageVolume',
--- a/extensions/default/src/DicomJSONDataSource/index.js
+++ b/extensions/default/src/DicomJSONDataSource/index.js
@@ -219,6 +219,9 @@ function createDicomJSONApi(dicomJsonConfig) {
                 imageId: instance.url,
                 ...series,
                 ...study,
+                // We need to force this flag to false in case the client wants to load the metadata from the image
+                // directly with automaticallyLoadDicomMetadata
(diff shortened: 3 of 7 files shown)


gpasquie:
This is a new version of PR https://github.com/OHIF/Viewers/pull/3573. Sorry for the long delay. I tried to take your comments into consideration but not sure if this was exactly what you suggested.

sedghi (commented):
can you please provide a public example on how to test this? like can you provide the dicomJSON json file? or at least the structure of it?

gpasquie:
Sure. This can be tested using the following JSON datasource :
[LIDC-IDRI-0002.json](https://github.com/OHIF/Viewers/files/14308358/LIDC-IDRI-0002.json)
and using URL: http://localhost:3000/viewer/dicomjson?url=http://localhost:3000/LIDC-IDRI-0002.json

I started from this JSON file : https://ohif-dicom-json-example.s3.amazonaws.com/LIDC-IDRI-0001.json
and just removed all images metadata which are not usually present in a DICOM KOS.

sedghi (commented):
Thanks for the PR again, 

I'm trying to understand the problem that this PR is solving. I see that you are still using the DICOM JSON format to provide the Study/Series/SOP UIDs, so essentially you are providing the metadata. I am more interested in providing a manifest, and then the viewer loads the rest.

see this example 

https://storage.googleapis.com/idc-volview-pilot/idc-test.json

[From this project ](https://projectweek.na-mic.org/PW38_2023_GranCanaria/Projects/IDC_with_VolView/) which is for Volview. 

I find that really nice and believe it addresses a significant issue. 

Do you think this is something you could work on?

gpasquie:
In fact, in our project, we are working in an XDS like environment. A central registry gathers pointers (DICOM KOS) to DICOM images which are stored in different PACS. When the user wants to view the images, the first step is to retrieve the DICOM KOS and then to open OHIF. From the DICOM KOS, it's quite easy to generate the OHIF JSON datasource, except for the images metadata (rows, columns,…), which are not usually present in a KOS. This is why we submitted this PR, to be able to open OHIF with a reduced JSON datasource, and without too many changes in OHIF code.

Opening OHIF from a manifest containing only paths to DICOM images would require much more changes in OHIF in my opinion and I'm not sure we would have the bandwidth to do so. If you think this is less complex that I think, could you please provide some guidance, and we can try to have a look

sedghi:
Ok good to know, then I'll review this

gpasquie:
Any feedbacks/reviews on this PR ?

sedghi:
This was on my radar, i will get to it this week, sorry for the delay

kayone:
Question about getting the viewer to load purely based on an array of DICOM URLs like @sedghi mentioned https://storage.googleapis.com/idc-volview-pilot/idc-test.json

Couldn't it be piped down using a similar code to the local mode (https://viewer.ohif.org/local), but instead of drag and drop, just load the files from a JSON file and pass them to the FileLoaderService? 

@sedghi, if you provide some guidance, we could definitely attempt a PR that would implement this.

sedghi:
@kayone and @sebastian 

Here is what I think is a good way to implement it. I think we cannot and should not support the flat URLs since it will take a lot of time to load the whole thing to extract the metadata. So a middle ground would be like this:

```json
{
  "studies": [
    {
      "series": [
        // series 1
        {
          "instances": [
            {
              "url": "dicomweb:https://ohif-dicom-json-example.s3.amazon/1-001.dcm"
            },
            {
              "url": "dicomweb:https://ohif-dicom-json-example.s3.amazon/1-002.dcm"
            },
            {
              "url": "dicomweb:https://ohif-dicom-json-example.s3.amazon/1-003.dcm"
            },
          ]
        },
        // series 2
        {
          "instances": [
            {
              "url": "dicomweb:https://ohif-dicom-json-example.s3.amazon/1-001.dcm"
            },
            {
              "url": "dicomweb:https://ohif-dicom-json-example.s3.amazon/1-002.dcm"
            },
            {
              "url": "dicomweb:https://ohif-dicom-json-example.s3.amazon/1-003.dcm"
            },
          ]
        }
      ],
    }
  ]
}

```

The series are already grouped, which makes sense since on servers, the DICOMs for a series are typically stored in a folder.

Then there's the issue of sorting. If the instances inside the folder are sorted (by `imagePositionPatient`), we're good, and we can render one URL from each series to obtain the common metadata, and the rest should work out of the box (we can assume defaults for the `imagePositionPatient` and cornerstone would just render them properly and since they are already ordered then we just put them one after another)

However, if they are not properly sorted, we need to load all instances of a series to render them in order based on `imagePositionPatient`, which takes more time. Unfortunately, there's no better alternative.

I hope this explanation makes sense for you to proceed. Since it's not a priority for us at the moment, let me know if you have any questions, and feel free to message me on Slack so I can assist you with this new feature.

Issue #3978, opened 2024-03-05
[Bug] Create measurements on MPR, turn off MPR and cannot create new measurements on 2x2 layout viewports


tzuccher:
### Describe the Bug

Found in OHIF dev env: version 3.8.0-beta.56, Commit hash: 21e8a2bd0b7cc72f90a31e472d285d761be15d30 on a mac + chrome browser

Creating bidirectional measurements within the MPR viewports and then turning off MPR to return to 2x2 layout and try to create measurements on viewport b is not successful. The empty measure appears on the measurements panel but no measure on the image. Also note the errors in the console:
![image](https://github.com/OHIF/Viewers/assets/143200290/65a6a5f3-8224-4a1d-9711-c6c9d48aa685)

### Steps to Reproduce

1. Select CT\PT\SR study ACRIN...with 408 instances
2. Select 2x2 grid layout
3. Select viewport B + MPR tool
4. Select the bidirectional tool and measure all 3 MPR viewports (click yes to track measurements) = measurements appear in all 3 viewports but, only the axial image measure is in the right-side panel
5. Turn off MPR to go back to 2x2 layout
6. Click into viewport B, reselect the bidirectional tool and create measurements on the image = i) no measurements appear on the image, ii) empty measurements in the right-side panel and iii) errors in the dev console

### The current behavior

- no measurements appear on the image
- empty measurements in the right-side panel 
- errors in the dev console

### The expected behavior

measurements appear on the image, correct measures in the right-side panel & no errors in console

### OS

MacBook Pro - Sonoma 14.1.1

### Node version

unknown

### Browser

Chrome v122

sedghi:
We just release the OHIF 3.8, you can find more details here https://ohif.org/release-notes/3p8/
If you still encounter this issue in 3.8, please re-open this.

Issue #3947, opened 2024-02-21
[Bug] Length measurements made on MPR images persist after closing study 


tzuccher:
### Describe the Bug

Length measurements made on MPR sagittal & coronal images persist after closing study and re-open again.

Able to reproduce from dev env (https://viewer-dev.ohif.org/): 3.8.0-beta.53, Commit hash: 5bba98ed848bdf46b5ba4fc4708527cced3308b5

### Steps to Reproduce

1. From the study list select COVID-19-NY-SBU with 477 images and go to the viewer screen
2. Go into 2x2 layout and select viewport 2
3. Click on the MPR tool
4. Make a length measurement on each of the 3 viewports & open the measurements panel = you will see only the axial measurement in the panel
5. Reselect the MPR tool to go back to 2x2 layout
6. Go back to the study list screen by selecting the Open Health Imaging Foundation verbiage in the upper left corner
7. Select the same study and go to the viewer screen
8. Go to 2x2 layout - select viewport 2 and go to the MPR = see the length on the sagittal & coronal views
![image](https://github.com/OHIF/Viewers/assets/143200290/138c1ec3-2576-472d-84f3-20905f5df907)

### The current behavior

Sagittal & Coronal measurements made remain in study after going to the study list (but in same browser session)

### The expected behavior

All measurements should be removed from the study when you leave the viewer screen

### OS

MacBook Pro - Sonoma 14.1.1

### Node version

unknown

### Browser

Chrome v121

james-hanks:
Fixed https://github.com/OHIF/Viewers/pull/3983

Re-test in 3.8.0-rc1

sedghi:
We just release the OHIF 3.8, you can find more details here https://ohif.org/release-notes/3p8/
If you still encounter this issue in 3.8, please re-open this.

Issue #3927, opened 2024-02-12
[Bug] Toolbar and hotkeys bugs


sedghi:
### Describe the Bug

- [ ] https://github.com/OHIF/Viewers/issues/3831
- [ ] https://github.com/OHIF/Viewers/issues/3848
- [ ] https://github.com/OHIF/Viewers/issues/3896
- [ ] https://github.com/OHIF/Viewers/issues/3200
- [ ] https://github.com/OHIF/Viewers/issues/3947

There seem to be various bugs that are related to the toolbar, and hotkeys. 

Basically the requirements are 

There should be 3 types of tools/icons that we support in the toolbar: Toggles, Actions, Tools

There are bugs regarding toggles states and nested menu behaviour which I will explain next.

## Toggles

- Toggles (they have on and off state e.g., imageSync, referenceLines, CINE player etc.)
- should be able to run by ON or OFF state by default, e.g., should be able too choose to run the viewer with references lines on ([issue reported here](https://github.com/OHIF/Viewers/issues/3831)) and [here](https://github.com/OHIF/Viewers/issues/3848))
- They can live on the toolbar as primary or secondary (nested)
- Upon user interaction (either mouse/touch click or hotkeys) they should show a different ON state color (as we do have for reference lines for instance)

![CleanShot 2024-02-11 at 22 56 25](https://github.com/OHIF/Viewers/assets/7490180/d170dbab-1af3-4ee6-ba86-614d55d44ffc)

Current Bugs for Toggles: 
- Go to [here](https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=2.16.840.1.114362.1.11972228.22789312658.616067305.306.2)
- On the right most menu, click on the dropdown
- Click on a toggle (e.g., reference lines, CINE player works fine for some reason)
- see the window level tool is lost and the user cannot window level by mouse primary
- Correct behaviour: The reference line toggle should not affect the mouse primary as it is only toggle of an state

pretty sure this bug is in setToolActive in the commandsModule

Fixed in https://github.com/OHIF/Viewers/pull/3961

---

## Nested Menu wrong behaviour for actions

- Select rectangle on the right most menu (nested)
- Then select the rotate right
- See now the rotate right is highlighted 
- Correct behaviour: should not be highlighted, only put on the top, highlight light blue should be associated with only active tool

Fixed in https://github.com/OHIF/Viewers/pull/3961

## Nested Menu wrong behaviour for toggles

[...]

Issue #3896, opened 2024-01-15
[Bug] Tool state is buggy with measurement tools in MPR


m00n620:
### Describe the Bug

Tool state is not correctly saved & restored with measurement tools when switching hanging protocols.

### Steps to Reproduce

1. Load viewer
2. Click Zoom tool
3. Click Ellipse ROI
4. Click MPR

### The current behavior

Zoom is active although toolbar state shows Ellipse ROI.

### The expected behavior

Ellipse ROI should be active tool.

### OS

Windows 11

### Node version

18.16.0

### Browser

Chrome 120.0.6099.217

sedghi:
We can add the current active tool to the state sync service I guess to remember, but then we should have some feedbacks

sedghi:
We just release the OHIF 3.8, you can find more details here https://ohif.org/release-notes/3p8/
If you still encounter this issue in 3.8, please re-open this.

PR #3983, opened 2024-03-08
feat(toolbar): new Toolbar to enable reactive state synchronization 

sedghi:
This PR will address some fundamental issues with the current implementation of the toolbar and toolbar service.

Fixing these 
- https://github.com/OHIF/Viewers/issues/3831
- https://github.com/OHIF/Viewers/issues/3848
- https://github.com/OHIF/Viewers/issues/3896
- https://github.com/OHIF/Viewers/issues/3200
- https://github.com/OHIF/Viewers/issues/3800
- https://github.com/OHIF/Viewers/issues/3927
- https://github.com/OHIF/Viewers/issues/3987
- https://github.com/OHIF/Viewers/issues/3947
- https://github.com/OHIF/Viewers/issues/3978
- https://github.com/OHIF/Viewers/issues/3869
- https://github.com/OHIF/Viewers/issues/3820

There were so many things wrong with our toolbar, where to synchronize the button states
with the tool state you should have activated the tool via `toolbarService.recordInteraction` which
was horrible.

This PR will add a new concept of `evaluators` for each button, which will be used to determine the
state of the button based on the app context.

There are some breaking changes in this PR, but I think they are necessary to make the toolbar more
flexible and easier to use. We will talk about the breaking changes later down

## Reactive Toolbar
Addressing 
- https://github.com/OHIF/Viewers/issues/3896

By the introduction of the evaluators, the toolbar will be reactive to the viewports and tooLGroups see below movie

https://github.com/OHIF/Viewers/assets/7490180/958de3d5-a6be-46e2-9f53-5e4f1f58c0d1

## Flexible Button Sections

You can assign any button to any section you want, in your extensions you can provide a
template that accepts button section and in your modes you should be able to assign buttons
to it.

We have created a new general `ToolBox` component that you can use as many time as you want
with different section names. We are shipping it for the segmentation mode for now.
You should be able to do the following for instance

CleanShot 2024-03-08 at 14 55 27

by saying

```js
toolbarService.addButtons(toolbarButtons)
toolbarService.createButtonSection("primary", [
  "Pan",
  "Capture",
  "Layout",
  "MPR",
  "Crosshairs",
  "Zoom",
  "MoreTools",
])
toolbarService.createButtonSection("segmentationToolbox", [
  "Zoom",
  "WindowLevel",
])
```

and you should be able to modify the segmentation toolbox initial setup too via your modes by assigning buttons.

```js
{
    id: 'Shapes',
    uiType: 'ohif.radioGroup',
    props: {
      label: 'Shapes',
      evaluate: {
        name: 'evaluate.cornerstone.segmentation',
        options: { toolNames: ['CircleScissor', 'SphereScissor', 'RectangleScissor'] },
      },
      icon: 'icon-tool-shape',
      commands: _createSetToolActiveCommands('CircleScissor'),
      options: [
        {
          name: 'Shape',
          type: 'radio',
          value: 'CircleScissor',
          id: 'shape-mode',
          values: [
            { value: 'CircleScissor', label: 'Circle' },
            { value: 'SphereScissor', label: 'Sphere' },
            { value: 'RectangleScissor', label: 'Rectangle' },
          ],
          commands:'setToolActiveToolbar'
        },
      ],
    },
  },
```

above as you see we use `commands` to set the most updated option value via a command, also you could have more sophisticated option change handling such as 

```
{
      name: 'Radius (mm)',
      id: 'eraser-radius',
      type: 'range',
      min: 0.5,
      max: 99.5,
      step: 0.5,
      value: 25,
      commands: {
        commandName: 'setBrushSize',
        commandOptions: { toolNames: ['CircularEraser', 'SphereEraser'] },
      },
    },
```

You can now have toolbox (in a modal) with tools and advanced options inside and everything will be synced in terms of state

CleanShot 2024-03-25 at 14 20 28@2x

Read more in **Toolbar.md**

## initial states for toggles
Addressing 
- https://github.com/OHIF/Viewers/issues/3831
- https://github.com/OHIF/Viewers/issues/3848

Now you can set the toggles initially and start the viewer with them, e.g., start viewer by reference lines on

CleanShot 2024-03-08 at 15 09 21

You only need to active/enable them in the `initToolGroups` and everything will work

## Hotkeys synchronization with toolbar (in progress)
Will address this
- https://github.com/OHIF/Viewers/issues/3200

## Breaking Changes

1. There concept of `activeTool` and their getter and setter is removed. The active tool
   should be derived by the toolGroup and the viewport and by manually settings it.

**action need to be taken**

```js
let unsubscribe
toolbarService.setDefaultTool({
  groupId: "WindowLevel",
  itemId: "WindowLevel",
  interactionType: "tool",
  commands: [
    {
      commandName: "setToolActive",
      commandOptions: {
        toolName: "WindowLevel",
      },
      context: "CORNERSTONE",
    },
  ],
})

const activateTool = () => {
  toolbarService.recordInteraction(toolbarService.getDefaultTool())

  // We don't need to reset the active tool whenever a viewport is getting
  // added to the toolGroup.
  unsubscribe()
}

// Since we only have one viewport for the basic cs3d mode and it has
// only one hanging protocol, we can just use the first viewport
;({ unsubscribe } = toolGroupService.subscribe(
  toolGroupService.EVENTS.VIEWPORT_ADDED,
  activateTool
))
```

Since tools are setup to be (active, passive, disabled and enabled) by the toolGroup
in `initToolGroups` you shouldn't worry about this, and after this PR we will
automatically set the active tool based on the toolGroup and the viewport.

The only thing as a `Mode` writer you should do is to write what buttons you have and
where you want to put them, as simple as

```js
toolbarService.addButtons([...toolbarButtons, ...moreTools])
toolbarService.createButtonSection("primary", [
  "MeasurementTools",
  "Zoom",
  "WindowLevel",
  "Pan",
  "Capture",
  "Layout",
  "MPR",
  "Crosshairs",
  "MoreTools",
])
```

2. There is no concept of button type anymore, there is no need to decide on `toggle` , `action` or `tool`.
   Buttons now are just buttons, which are some properties (icon, label, etc.) with `commands` that will
   run (by toolbarService) after interaction. Evaluators will be used to determine the state of the button
   (color, disabled etc.)

remove the `type` property from the buttons, and instead use the uiType for the top level ui type def.

**old implementation**

```js
{
    id: 'Capture',
    type: 'ohif.action',
    props: {
      icon: 'tool-capture',
      label: 'Capture',
      type: 'action',
      commands: [
        {
          commandName: 'showDownloadViewportModal',
          commandOptions: {},
          context: 'CORNERSTONE',
        },
      ],
    },
  },

```

**New implementation**

```js
{
  id: 'Capture',
  uiType: 'ohif.radioGroup',
  props: {
    icon: 'tool-capture',
    label: 'Capture',
    commands: [
      {
        commandName: 'showDownloadViewportModal',
        context: 'CORNERSTONE',
      },
    ],
    evaluate: 'evaluate.action',
  },
},
```

(Notice the uiType)
(we will talk about the `evaluate` property later down)

In addition, the previous signature of button definition (`ToolbarService.createButton`), which relied on positional arguments, is removed and replaced by a more flexible object-based definition.

**old implementation **

```js
ToolbarService._createToolButton(
  'ArrowAnnotate',
  'tool-annotate',
  'Annotation',
  [
    {
      commandName: 'setToolActive',
      commandOptions: {
        toolName: 'ArrowAnnotate',
      },
      context: 'CORNERSTONE',
    },
    {
      commandName: 'setToolActive',
      commandOptions: {
        toolName: 'SRArrowAnnotate',
        toolGroupId: 'SRToolGroup',
      },
      context: 'CORNERSTONE',
    },
  ],
  'Arrow Annotate'
),
```

**New implementation**

```js
ToolbarService.createButton({
  id: 'ArrowAnnotate',
  icon: 'tool-annotate',
  label: 'Annotation',
  tooltip: 'Arrow Annotate',
  commands: {
      commandName: 'setToolActiveToolbar',
      commandOptions: {
          toolGroupIds: ['default', 'SRToolGroup'],
      },
   },
}),
```

(notice the object based definition in the new implementation)
(we will talk about the `evaluate` property next)

3. Add the proper `evaluate` property to the button definition, which will be used to determine the state of the button
   based on the app context.

   You can check out the `modes/longitudinal/src/toolbarButtons.ts` for the usage of the `evaluate` property.

   But in general

   - Use the `evaluate.cornerstoneTool` if you want the button to be highlighted if it is only on active primary (left mouse)
   - Use `evaluate.cornerstoneTool.toggle` if the tool is a toggle tool (like reference lines or image overlay)
   - Read more in the toolbar module documentation

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/default/src/Panels/Toolbox/ToolboxContainer.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const initialState = {
+  activeTool: null, // Tracks the current active tool
+  toolOptions: {}, // Stores the options for each tool
+};
+
+// Todo: make this configurable
+const itemsPerRow = 4;
+
+function ToolboxContainer({ servicesManager, commandsManager, buttonSectionId, title }) {
```
wayfarer3130, line 14: Docs please

extensions/default/src/Panels/Toolbox/ToolboxUI.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import { PanelSection, ToolSettings, Tooltip } from '@ohif/ui';
+import classnames from 'classnames';
+
+function ToolboxUI(props) {
```
wayfarer3130, line 5: Needs docs as to how this gets used, as this is supposed to be a generic component.

extensions/default/src/Panels/Toolbox/ToolboxUI.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
```
wayfarer3130, line 1: This belongs in the UI section, as it is a generic tool that can be used - it really isn't specific to an mode or extension.

extensions/default/src/Panels/Toolbox/ToolboxContainer.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const { stateSyncService } = servicesManager.services;
+  const { onInteraction, toolbarButtons } = useToolbar({
+    servicesManager,
+    buttonSection: buttonSectionId,
+  });
+
+  const [state, dispatch] = useReducer(toolboxReducer, initialState);
```
wayfarer3130, line 21: This feels like a different pattern from most of our patterns, but feels more connected to the interface, which I feel is a good thing.  Should this get documented somewhere as a howto/pattern example?

extensions/default/src/commandsModule.ts
```diff
@@ -243,33 +207,9 @@ const commandsModule = ({
           `${activeStudyUID || hpInfo.activeStudyUID}:activeDisplaySet:0`
         ];
         stateSyncService.store(stateSyncReduce);
-        // This is a default action applied
```
wayfarer3130, line 246: This is just so much cleaner to have the toolbar defined by the commands and state changes it makes.

extensions/default/src/getToolbarModule.tsx
```diff
@@ -1,39 +1,69 @@
...
-      defaultComponent: ToolbarDivider,
-      clickHandler: () => {},
+      name: 'ohif.radioGroup',
+      defaultComponent: ToolbarButton,
     },
     {
-      name: 'ohif.action',
-      defaultComponent: ToolbarButtonWithServices,
-      clickHandler: () => {},
+      name: 'myCustomUI',
```
wayfarer3130, line 23: This one should go into the basic test extension

extensions/default/src/getToolbarModule.tsx
```diff
@@ -1,39 +1,69 @@
...
       name: 'ohif.splitButton',
       defaultComponent: ToolbarSplitButtonWithServices,
-      clickHandler: () => {},
     },
     {
       name: 'ohif.layoutSelector',
       defaultComponent: ToolbarLayoutSelectorWithServices,
-      clickHandler: (evt, clickedBtn, btnSectionName) => {},
     },
     {
```
wayfarer3130, line 40: I really prefer the way this is going now - I think it just works a lot better

extensions/dicom-microscopy/src/getCommandsModule.ts
```diff
@@ -135,27 +135,22 @@ export default function getCommandsModule({
   const definitions = {
     deleteMeasurement: {
       commandFn: actions.deleteMeasurement,
-      storeContexts: [] as any[],
       options: {},
```
wayfarer3130, line 138: Also remove options please.

extensions/dicom-microscopy/src/index.tsx
```diff
@@ -23,7 +23,7 @@ const MicroscopyViewport = props => {
 /**
  * You can remove any of the following modules if you don't need them.
  */
-export default {
+const extension = {
```
wayfarer3130, line 26: Could you declare the type here please?  There is an Extension type exported that has required things like id, preRegistration etc.

extensions/dicom-microscopy/src/index.tsx
```diff
@@ -90,6 +90,45 @@ export default {
     ];
   },
 
+  getToolbarModule({ servicesManager }) {
+    return [
+      {
+        name: 'evaluate.microscopyTool',
```
wayfarer3130, line 96: This use here demonstrates that the toolbars are now extension independent and belong in the UI group instead of the default extension.

extensions/default/src/init.ts
```diff
@@ -38,14 +47,53 @@ export default function init({ servicesManager, configuration = {} }): void {
...
+      const supportedEvents = [
+        viewportGridService.EVENTS.ACTIVE_VIEWPORT_ID_CHANGED,
+        viewportGridService.EVENTS.VIEWPORTS_READY,
+      ];
+
+      if (supportedEvents.includes(event)) {
+        viewportGridService.subscribe(event, eventData => {
+          const viewportId = eventData?.viewportId ?? viewportGridService.getActiveViewportId();
+
+          commandsManager.run(commands, { viewportId });
```
wayfarer3130, line 71: You should be aware that the viewport grid service fires events BEFORE the viewport actually gets updated, so if your new events depend on the viewport having already changed, it won't respond with what you expect.

extensions/tmtv/src/Panels/PanelPetSUV.tsx
```diff
@@ -135,6 +118,11 @@ export default function PanelPetSUV({ servicesManager, commandsManager }) {
 
     // update the displaySets
     displaySetService.setDisplaySetMetadataInvalidated(ptDisplaySet.displaySetInstanceUID);
+
+    // seems like for some reason crosshairs should get reset
```
wayfarer3130, line 122: Can you comment on this - for example
// Crosshair position depends on the metadata values such as the positioning interaction
// between series, so when the metadata is updated, the crosshairs need to be reset.

modes/basic-dev-mode/src/index.js
```diff
@@ -89,39 +89,8 @@ function modeFactory({ modeConfiguration }) {
...
-        unsubscribe();
-      };
-
-      // Since we only have one viewport for the basic cs3d mode and it has
-      // only one hanging protocol, we can just use the first viewport
-      ({ unsubscribe } = toolGroupService.subscribe(
-        toolGroupService.EVENTS.VIEWPORT_ADDED,
-        activateTool
-      ));
+      toolGroupService.createToolGroupAndAddTools('default', tools);
```
wayfarer3130, line 92: Wow - I really like this change.

wayfarer3130: Is there some way you could register the tool sets with the customization service - will think about that, and see if we can come up with a way to have the extensions define default sets of tools by registering pieces with the customization and toolbar services, but I don't htink it belongs in this PR, jsut something to think about.

modes/basic-dev-mode/src/toolbarButtons.js
```diff
@@ -54,147 +23,91 @@ function _createWwwcPreset(preset, title, subtitle) {
...
-              toolName: 'Length',
-            },
-            context: 'CORNERSTONE',
-          },
-        ],
-        'Length'
-      ),
+      groupId: 'MeasurementToolsGroupId',
+      evaluate: 'evaluate.group.promoteToPrimaryIfCornerstoneToolNotActiveInTheList',
+      primary: ToolbarService.createButton({
```
wayfarer3130, line 44: Could you import the createButton function - eg:
const { createButton } = ToolbarService

The reason is it allows you to over-ride it for anything specific that you want to do.

modes/basic-dev-mode/src/toolbarButtons.js
```diff
@@ -54,147 +23,91 @@ function _createWwwcPreset(preset, title, subtitle) {
   };
 }
 
-const toolbarButtons = [
-  // Measurement
+function _createSetToolActiveCommands(toolName, toolGroupIds = ['default', 'mpr', ]) {
```
wayfarer3130, line 26: This feels like the options to the toolbars should include the toolbar item being selected, and that the tool group ids should belong to a set of tool groups being defined, so that you get 1 setToolActive call, and it is completely generic across all tools - that is, something like:

const commands = [{ commandName: 'setToolActive' }];

You might need to add the context here as well, but if the command caller for the tool group options automatically includes the tool group ids to default to, plus the commandName itself into the comman options, then it can automatically iterate/apply with a shared command.

modes/basic-test-mode/src/moreTools.ts
```diff
@@ -178,13 +171,14 @@ const moreTools = [
...
-          'Magnify',
-          [
+          evaluate: 'evaluate.cornerstoneTool',
+        }),
+        ToolbarService.createButton({
+          id: 'Magnify',
+          icon: 'tool-magnify',
+          label: 'Magnify',
+          tooltip: 'Magnify',
+          commands: [
```
wayfarer3130, line 181: Can you extract the commands into a constant please?

modes/basic-test-mode/src/moreTools.ts
```diff
@@ -208,21 +203,21 @@ const moreTools = [
...
-          'Dicom Tag Browser',
-          [
+          evaluate: 'evaluate.cornerstoneTool',
+        }),
+        ToolbarService.createButton({
+          id: 'TagBrowser',
+          icon: 'list-bullets',
+          label: 'Dicom Tag Browser',
+          tooltip: 'Dicom Tag Browser',
+          commands: [
```
wayfarer3130, line 213: This is one place where you need a commands list specific to the tool

wayfarer3130: However, you can use a single commands item instead of an array - eg commands: { commandName: 'openDICOMTagViewer' } - actually, we should let that default to just the string name of a command.

modes/basic-test-mode/src/toolbarButtons.ts
```diff
@@ -165,13 +151,14 @@ const toolbarButtons: Button[] = [
...
+          evaluate: 'evaluate.cornerstoneTool',
+        }),
+        ToolbarService.createButton({
+          id: 'CircleROI',
+          icon: 'tool-circle',
+          label: 'Circle',
+          tooltip: 'Circle ROI Tool',
+          commands: [
             {
               commandName: 'setToolActive',
```
wayfarer3130, line 163: If the set tool active took an array of tool group Ids, then you could do this all in one - and just use a single fixed commands list that includes the tool gorup ids.

extensions/cornerstone/src/Viewport/Overlays/CustomizableViewportOverlay.tsx
```diff
@@ -282,103 +211,101 @@ function CustomizableViewportOverlay({
...
 
   return (
     <ViewportOverlay
-      topLeft={getTopLeftContent()}
-      topRight={getTopRightContent()}
-      bottomLeft={getBottomLeftContent()}
-      bottomRight={getBottomRightContent()}
+      topLeft={getContent(
+        topLeftCustomization,
+        [
```
wayfarer3130, line 239: Why are you removing the customization capability here?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -279,35 +259,30 @@ function commandsModule({
...
+
+      // Toggle the tool's state only if the toggle is true
+      if (toggle) {
+        toolIsEnabled ? toolGroup.setToolDisabled(toolName) : toolGroup.setToolEnabled(toolName);
+      }
+
+      const renderingEngine = cornerstoneViewportService.getRenderingEngine();
+      renderingEngine.render();
+    },
+    setToolActive: ({ toolName, toolGroupId = null }) => {
```
wayfarer3130, line 285: Please allow toolGroupId to be an array here.  That is more efficient than calling multiple commands for the very common case of setting multiples.

platform/app/src/components/ViewportGrid.tsx
```diff
@@ -327,6 +340,9 @@ function ViewerViewportGrid(props) {
               viewportOptions={viewportOptions}
               displaySetOptions={displaySetOptions}
               needsRerendering={displaySetsNeedsRerendering}
+              onReady={() => {
```
wayfarer3130, line 343: That is quite nice to have, to get a callback that is timed for actually being ready.

platform/ui/src/components/ToolbarButton/ToolbarButton.tsx
```diff
@@ -73,15 +50,11 @@ const ToolbarButton = ({
...
-  type: 'action',
 };
 
 ToolbarButton.propTypes = {
   /* Influences background/hover styling */
-  type: PropTypes.oneOf(['action', 'toggle', 'tool']),
   id: PropTypes.string.isRequired,
-  isActive: PropTypes.bool,
   className: PropTypes.string,
   commands: PropTypes.arrayOf(
```
wayfarer3130, line 59: Can you update this to take the simple type or the full array, and also use the Types value exported from the commands?  That way it just becomes commands: PropTypes.instanceOf(Types.Commands)  (something like that anyways).

extensions/cornerstone/src/Viewport/Overlays/CustomizableViewportOverlay.tsx
```diff
@@ -282,103 +211,101 @@ function CustomizableViewportOverlay({
...
+    [_renderOverlayItem]
+  );
 
   return (
     <ViewportOverlay
-      topLeft={getTopLeftContent()}
-      topRight={getTopRightContent()}
-      bottomLeft={getBottomLeftContent()}
-      bottomRight={getBottomRightContent()}
+      topLeft={getContent(
```
wayfarer3130, line 237: Could you add a comment here indicating that you are inlining the default overlay items so that it looks more like a standard expansion - I like what you have done, but I missed the fact that you had done it on first read.  I might actually take THIS example and add it to the customization docs explaining that this is a good way to do things.  In fact, I plan on following this example when I tweak the search page.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -279,35 +259,30 @@ function commandsModule({
...
+      const toolIsEnabled = toolGroup.getToolOptions(toolName).mode === Enums.ToolModes.Enabled;
+
+      // Toggle the tool's state only if the toggle is true
+      if (toggle) {
+        toolIsEnabled ? toolGroup.setToolDisabled(toolName) : toolGroup.setToolEnabled(toolName);
+      }
+
+      const renderingEngine = cornerstoneViewportService.getRenderingEngine();
+      renderingEngine.render();
+    },
```
wayfarer3130, line 284: Just down below, where you set crosshairs disabled, could you choose the disabled/passive decision based on a bit of toolbar configuration?  Maybe use the cornerstone tools configuration for crosshairs, like:
toolGroup.getToolOptions(9toolName).disableOnDeactivate.  You are making almost all the rest of hte toolbar cleanly configured, I think this part should be as well.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -565,17 +522,23 @@ function commandsModule({
...
+      if (type === 'imageSlice') {
+        toggleImageSliceSync({
+          servicesManager,
+          viewports,
+        });
+      } else if (type === 'voi') {
+        toggleVOISliceSync({
+          servicesManager,
+          viewports,
+        });
```
wayfarer3130, line 535: Is it possible to call the method with syncId directly, and then call the setEnable method on the sync - that would presuppose the sync already existed, but I think that should be an option.

extensions/cornerstone/src/getToolbarModule.tsx
```diff
@@ -0,0 +1,242 @@
...
+        const toolName = getToolNameForButton(button);
+
+        if (!toolGroup || !toolGroup.hasTool(toolName)) {
+          return {
+            disabled: true,
+            className: '!text-common-bright ohif-disabled',
+          };
+        }
+
+        const isPrimaryActive = toolGroup.getActivePrimaryMouseButtonTool() === toolName;
```
wayfarer3130, line 40: After this PR gets pushed, I'm going to add additional behaviour here that allows passing in a specific binding to use for active/passive handling so that one can also define an "active right click" set of buttons - that allows using left for tools and right for voi type functionality and center for navigation type.

extensions/cornerstone/src/init.tsx
```diff
@@ -41,7 +40,6 @@ export default async function init({
   servicesManager,
   commandsManager,
   extensionManager,
```
wayfarer3130, line 42: Can we make the name of the file/function consistent - it is preRegistration in the extension vocabulary.  Since you are here and changing it anyways.

extensions/cornerstone/src/init.tsx
```diff
@@ -230,41 +237,41 @@ export default async function init({
    * active tools in the toolbar, and call any commands registered in the
    * toolbar service with a callback to re-enable on displaying the viewport.
    */
-  const toolbarEventListener = evt => {
-    const { element } = evt.detail;
-    const activeTools = toolbarService.getActiveTools();
+  // const toolbarEventListener = evt => {
```
wayfarer3130, line 240: Are you planning on leaving this commented out for now?  Would like to see it removed in the final commit.

extensions/default/src/Panels/Toolbox/ToolboxUI.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <PanelSection title={'Segmentation Tools'}>
+      <div className="flex flex-col bg-black">
+        <div className="bg-primary-dark mt-0.5 flex flex-wrap py-2">
+          {toolbarButtons.map((toolDef, index) => {
+            if (!toolDef) {
+              return null;
+            }
+
+            const { id, Component, componentProps } = toolDef;
+            const isLastRow = Math.floor(index / 4) + 1 === numRows;
```
wayfarer3130, line 18: Could you extract the 4 as a constant for now - at least that explains why the 4 is there, which I had to think about.

platform/app/pluginConfig.json
```diff
@@ -68,12 +68,10 @@
     },
     {
       "packageName": "@ohif/mode-test",
-      "default": false,
```
wayfarer3130, line 71: It should be possible to load the modes without them being defaults by configuring them in the app config - that is the preferred way to add these, rather than having them get listed all the time.

platform/core/src/hooks/useToolbar.tsx
```diff
@@ -0,0 +1,59 @@
+import { useCallback, useEffect, useState } from 'react';
+
+export function useToolbar({ servicesManager, buttonSection = 'primary' }) {
```
wayfarer3130, line 3: Docs please - also, if not already in the toolbar markdown docs, could you add the useToolbar documentation there?  This is a pretty important part of the GUI functionality.

wayfarer3130: Oh, and I really like having this here.

platform/core/src/services/CineService/CineService.ts
```diff
@@ -1,82 +1,80 @@
...
+    // reducer state does not get updated right away and if we publish the
+    // event and we use the cineService.getState() it will return the old state
+    setTimeout(() => {
+      this._broadcastEvent(this.EVENTS.CINE_STATE_CHANGED, isCineEnabled);
+    }, 0);
   }
-  if (setCineImplementation) {
-    serviceImplementation._setCine = setCineImplementation;
+
+  public playClip(element, playClipOptions) {
```
wayfarer3130, line 41: Given that you are tweaking this, could you remove the level of indirection, and make these all:
public playClip = (element,playClipOptions) => { throw new UnsupportedOperationException('playClip'); }

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -1,126 +1,169 @@
...
-  listeners?: ButtonListeners;
-  props?: ButtonProps;
+  commands?: RunCommand;
+  disabled?: boolean;
+  className?: string;
+  evaluate?: EvaluatePublic;
+  listeners?: Record<string, RunCommand>;
+}
+
+export interface NestedButtonProps {
```
wayfarer3130, line 31: Should be a type here - these are simple typed objects, not interfaces that we expect to be implemented AFAIK.

platform/docs/docs/platform/services/data/SyncGroupService.md
```diff
@@ -0,0 +1,123 @@
...
+
+
+In this example, the `ptAXIAL` viewport is part of three synchronization groups:
+
+1. `cameraPosition` group with the ID `'axialSync'`: This group synchronizes the camera position across viewports that are both source and target.
+2. `voi` (Window Level) group with the ID `'ptWLSync'`: This group synchronizes the window level settings across viewports that are both source and target.
+3. `voi` group with the ID `'ptFusionWLSync'`: This group synchronizes the window level settings, but the `ptAXIAL` viewport is only a source, not a target.
+
+
+### Manually through a button
```
wayfarer3130, line 73: Can you explain here that you can also create them through the hanging protocol, and then control the state via a toolbar button?  That is a fairly important use case as it allows defining exactly which set of viewports have the sync group turn on for.

platform/ui/src/components/SplitButton/SplitButton.tsx
```diff
@@ -54,13 +49,16 @@ const classes = {
       isExpanded
         ? 'bg-primary-light !rounded-tr-md !rounded-br-md'
         : primary.isActive
-        ? 'bg-secondary-dark'
-        : 'hover:bg-primary-dark bg-secondary-dark group-hover/button:border-primary-dark'
+          ? 'bg-secondary-dark'
+          : 'hover:bg-primary-dark bg-secondary-dark group-hover/button:border-primary-dark'
+      // Add a comma here
```
wayfarer3130, line 54: ?

platform/ui/src/components/ViewportOverlay/ViewportOverlay.tsx
```diff
@@ -10,10 +10,24 @@ const classes = {
   bottomLeft: 'bottom-viewport left-viewport',
 };
 
-const ViewportOverlay = ({ topLeft, topRight, bottomRight, bottomLeft, color }) => {
+type ViewportOverlayProps = {
```
wayfarer3130, line 13: export please

wayfarer3130 (commented):
Really like the direction this is going.  A bunch of small comments, but overall I think this is right on track.

extensions/tmtv/src/Panels/PanelPetSUV.tsx
sedghi: Thanks

platform/docs/docs/platform/extensions/modules/toolbar.md
```diff
@@ -53,236 +48,142 @@ export default function getToolbarModule({ commandsManager, servicesManager }) {
...
-export default function getToolbarModule({ commandsManager, servicesManager }) {
-  return [
-    {
-      name: 'new-tool-type',
-      defaultComponent: myToolComponent,
-      clickHandler: () => {},
+    onModeEnter: ({ servicesManager, extensionManager }) => {
+      const { toolBarService } = servicesManager.services;
+
+      toolbarService.addButtons([...toolbarButtons, ...moreTools]);
```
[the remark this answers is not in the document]
sedghi: Yes, I think this is the updated docs

platform/ui/src/components/ToolbarButton/ToolbarButton.tsx
sedghi: This one is tricky since it is in UI, not sure the best way to do this really

extensions/cornerstone/src/getToolbarModule.tsx
sedghi: Sounds good!

extensions/cornerstone/src/init.tsx
sedghi: There are 4 other init ones

CleanShot 2024-03-12 at 13 41 08

I rather do it in another PR really just to be clear about the PR commit message

platform/app/pluginConfig.json
sedghi: Yeah maybe another PR

platform/core/src/services/CineService/CineService.ts
sedghi: I'm not following this one, can you please elaborate?

platform/core/src/services/CineService/CineService.ts
```diff
@@ -1,82 +1,80 @@
...
-    serviceImplementation._stopClip = stopClipImplementation;
+  public setServiceImplementation({
+    getState: getStateImplementation,
+    setCine: setCineImplementation,
+    setIsCineEnabled: setIsCineEnabledImplementation,
+    playClip: playClipImplementation,
+    stopClip: stopClipImplementation,
+  }) {
+    if (getStateImplementation) {
+      this.serviceImplementation._getState = getStateImplementation;
```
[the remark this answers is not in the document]
sedghi: it doesn't work when I do so, maybe another PR?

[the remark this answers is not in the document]
wayfarer3130: Yes, another PR is fine.  If it had just worked immediately, I would have suggested doing it, but not if it involves more work than that.

platform/core/src/services/CineService/CineService.ts
wayfarer3130: Just that you can assign methods to the implementation directly, but given that it doesn't work, I would leave it.

extensions/cornerstone-dicom-seg/src/commandsModule.ts
```diff
@@ -397,6 +400,36 @@ const commandsModule = ({
...
+          segmentationUtils.setBrushSizeForToolGroup(toolGroupId, brushSize);
+        } else {
+          toolNames?.forEach(toolName => {
+            segmentationUtils.setBrushSizeForToolGroup(toolGroupId, brushSize, toolName);
+          });
+        }
+      });
+    },
+    setThresholdRange: ({
+      values,
```
wayfarer3130, line 417: The threshold value is a pair of numbers, but that is still the value, so I think this should be called value not values.

platform/ui/src/contextProviders/Toolbox/ToolboxContext.tsx
wayfarer3130, line 1: Can you add a few comments to this file?

platform/ui/src/components/Toolbox/ToolboxUI.tsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+import { PanelSection, ToolSettings, Tooltip } from '../../components';
+import classnames from 'classnames';
+
+const ItemsPerRow = 4;
+
+function ToolboxUI(props) {
```
wayfarer3130, line 7: Need 1-2 sentences describing what a ToolboxUI is, and why it isn't just Toolbox itself (which is also a tsx file)

platform/ui/src/components/Toolbox/Toolbox.tsx
```diff
@@ -0,0 +1,93 @@
+import React, { useEffect, useRef } from 'react';
+import { useToolbar } from '@ohif/core';
+import { ToolboxUI } from './';
+import { useToolbox } from '../../contextProviders';
+
+/**
+ * This component might seem out of place in the UI, but it plays a crucial role in
```
wayfarer3130, line 7: Need a sentence saying what it is  'A toolbox is a collection of buttons and commands that they invoke, used to provide custom control panels to users' - that is my take on what this is, but I might be completely off base.
Also, the UI is supposed to have generic UI components, and this one is generic, it is just that it is generic in that it interacts with the services/commands in a generic fashion.   There isn't anything in this that couldn't be used for a completely different type of app, so by that definition it belongs in the ui components.

platform/ui/src/components/ThumbnailTracked/ThumbnailTracked.tsx
```diff
@@ -87,12 +87,12 @@ function ThumbnailTracked ({
                 </div>
                 <div className="flex flex-1 flex-col">
                   <span>
-                    <span className="text-white">{isTracked ? t('Series is tracked') : t('Series is untracked')}</span>
+                    <span className="text-white">
```
wayfarer3130, line 90: This class to me feels like it belongs in an extension, not in platform/ui because it is dealing with mode specific components/information.  Not sure I feel like asking you to move it, but a comment about it might be appropriate.

platform/ui/src/components/SplitButtonToolbar/SplitButtonToolbar.tsx
```diff
@@ -0,0 +1,192 @@
...
+            className="h-5 w-5"
+          />
+        </span>
+      )}
+      <span className="mr-5">{t?.(label)}</span>
+    </div>
+  );
+};
+
+const SplitButtonToolbar = ({
```
wayfarer3130, line 87: How is this different from split button, or is it just the naming?  This needs at least a bit of documentation to explain what it is here.

platform/ui/src/components/SplitButtonToolbar/__stories__/splitButtonToolbar.stories.mdx
```diff
@@ -0,0 +1,80 @@
...
+  componentRelativePath="SplitButtonToolbar/SplitButtonToolbar.tsx"
+/>
+
+- [Overview](#overview)
+- [Props](#props)
+- [Contribute](#contribute)
+
+## Overview
+
+SplitButton is a component that renders the SplitButtons.
```
wayfarer3130, line 34: This is in splitBUttonToolbar - can you tweak the text to explain why it is splitButtonToolbar?

platform/core/src/types/Command.ts
```diff
@@ -4,7 +4,7 @@ export interface Command {
   context?: string;
 }
 
-export type RunCommand = Command | Command[];
+export type RunCommand = Command | Command[] | string;
```
wayfarer3130, line 7: This is quite nice to be able to just specify a simple string.

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -130,160 +136,184 @@ export default class ToolbarService extends PubSubService {
...
+        itemId,
+        interaction,
+      });
+    }
 
-        break;
+    // attach interaction to each command if it is an object {commandName, commandOptions}
+    commands.forEach(command => {
+      if (typeof command === 'object') {
+        command.commandOptions = { ...command.commandOptions, ...options, ...interaction };
```
wayfarer3130, line 170: You should construct the options outside the commands loop, not containing the commandOptions, and let commandsManager.run(command,options) combine the options as needed.  Then you can just:
this._commandsManager.run(commands, {...options, ...interaction});

platform/core/src/classes/CommandsManager.ts
```diff
@@ -167,21 +167,31 @@ export class CommandsManager {
...
+   * @param toRun - A specification of one or more commands,
+   *  typically an object of { commandName, commandOptions, context }
+   * or an array of such objects. It can also be a single commandName as string
+   * if no options are needed.
    * @param options - to include in the commands run beyond
    *   the commandOptions specified in the base.
    */
   public run(
-    toRun: Command | Commands | Command[] | undefined,
+    toRun: Command | Commands | Command[] | string | undefined,
```
wayfarer3130, line 178: If Command was string | { commandName: string, ...} and you check here, then the existing case/typedef up top works, and you can still just call it as a straight string - eg
commands: "singleCOmmandName"
or
commands: [
  "singleCOmmandName",
  { commandName: "complexCommand", commandOptions: {a:1}},
]

That also make the array conversion just work with string already.

modes/segmentation/src/segmentationButtons.ts
```diff
@@ -0,0 +1,222 @@
...
+          options: [
+            {
+              name: 'Radius (mm)',
+              id: 'brush-radius',
+              type: 'range',
+              min: 0.5,
+              max: 99.5,
+              step: 0.5,
+              value: 25,
+              onChange: executeSetBrushSize,
```
wayfarer3130, line 85: Is there any reason the value isn't simply included as 'value' in the call and then you can just execute the commands with the default properties, rather than needing an onChange?  I like having the onChange, I just don't think this particular use case is one that is needed.

modes/segmentation/src/toolbarButtons.ts
```diff
@@ -53,66 +21,45 @@ function _createWwwcPreset(preset, title, subtitle) {
...
+      }),
       secondary: {
         icon: 'chevron-down',
-        label: 'W/L Manual',
-        isActive: true,
         tooltip: 'W/L Presets',
       },
-      isAction: true, // ?
       renderer: WindowLevelMenuItem,
       items: [
```
wayfarer3130, line 64: Can these get disabled if not CT?

wayfarer3130: I mean it should be possible to have an evaluate that disables them when not CT.

sedghi: since we are re-working all of this in another PR to be modality specific, i will not touch that, but in theory yes

wayfarer3130: Lets leave it for now then.

modes/segmentation/src/segmentationButtons.ts
wayfarer3130: The issue is that we start adding executable code definitions to the toolbar definitions - which IMHO should be 99% declarative.  Yes, once in a while it needs an onChange listener, but almost entirely I think it should be defined like this:
```
 {
              name: 'Radius (mm)',
              id: 'brush-radius',
              type: 'range',
              min: 0.5,
              max: 99.5,
              step: 0.5,
              value: 25,
              // onChange: executeSetBrushSize,
              commands: { commandName: 'setBrushRadius', commandOptions: {  tools: ['CircularBrush', 'SphereBrush'] }},
            },
```

So, instead of the onChange, there is a commands inside the option, and that commands will get executed with the run option and will include the max, step etc, PLUS the actual current value.  That way it is entirely declarative and all the actual logic is inside the commands module 'setBrushRadius'
Basically - the onChange should be a commands list so we have 1 pattern for how things are done rather than 1 listener pattern, and one commands pattern.

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -130,160 +136,182 @@ export default class ToolbarService extends PubSubService {
...
-    if (groupId) {
-      this.state.groups[groupId] = itemId;
-    }
+    const commandOptions = { ...options, ...interaction };
 
-    this._broadcastEvent(this.EVENTS.TOOL_BAR_STATE_MODIFIED, { ...this.state });
-  }
+    // Loop through commands and run them with the combined options
+    commands.forEach(command => {
+      this._commandsManager.run(command, commandOptions);
```
wayfarer3130, line 171: this._commandsManager.run(commands, commandOptions) is sufficient.

sedghi: Done

wayfarer3130 (approved):
Looks wonderful - very much looking forward to having this in.

PR #3984, opened 2024-03-09
fix(viewport-sync): Enable re-sync image slices in a different position when needed

mateusfreira:
### Changes & Results
This is a specific  change [SyncGroupService.ts](https://github.com/OHIF/Viewers/compare/master...mateusfreira:Viewers:bug/unable-to-re-sync-image-slice-sync-tool?expand=1#diff-89983c288ef4490fa9c57972dd966b86e2b8c6f5ef3d8c9e95e5dbd244dbbd89) to clean up the previously registered `spatialRegistrationMetadataProvider` by cornerstone.js.

When creating the correlation matrix among the viewports cornerstone.js for performance reasons, it caches the matrix using the `spatialRegistrationMetadataProvider`,  but when the tool is disabled, it does not clean it up. This makes it impossible to re-sync the viewports to a different point if the user wants to.

This is a common use case when radiologists are comparing 2 studies (one current and one prior) and by mistake, enable the sync on a non-optimal point and want to disable it, scrool  one of the viewports to the correct point and then lock the scroll again.

[Here](https://github.com/cornerstonejs/cornerstone3D/blob/01c9f52c2a8e4263cbd9215f295e7840d3e04f9e/packages/tools/src/synchronizers/callbacks/imageSliceSyncCallback.ts#L94C1-L95C1) Cornestone calculateViewportsSpatialRegistration and it never gets cleaned up.
noted

### Testing
I Have demonstrated this Bug on a video available in youtube also show the fix

[Screenshot 2024-03-09 at 11 14 04 AM](https://www.youtube.com/live/SnnFo6TatPw?si=Z04uztvehheEc6QA&t=8450)

(Watch from 2:20:50 to 2:25:52 where I demo the Bug and the fix after my code change)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: 
macOS apple M1 
- [x] Node version: 
* v18.17.0
- [x] Browser:
  
* Arc Chromium Engine Version 119.0.6045.199
* Chrome Version 122.0.6261.112 (Official Build) (arm64)


extensions/cornerstone/src/services/SyncGroupService/SyncGroupService.ts
```diff
@@ -152,6 +154,8 @@ export default class SyncGroupService {
         return;
       }
 
+      this.unRegisterSpatialRegistration(synchronizer);
```
sedghi, line 157: can we filter and only do this for sync synchronizer?

mateusfreira: Thanks for the review, @sedghi; I sure can do it;  Must does not seem to exists a way to differentiate the synchronizer other than the name.

Is checking the name of the sincronizer enough for this situation?

sedghi: sure

mateusfreira: Thanks, @sedghi. I have updated the code, and it is working as expected now. Let me know if you think I should improve anything else.

IbrahimCSAE (approved):
LGTM

sedghi:
I still need to test this, @mateusfreira can you provide reproducible steps to follow that shows the bug?

sedghi (approved):
Tested it and worked great, Lovely fix, thanks a lot!

Issue #3960, opened 2024-02-26
Worklist UI enhancements


sedghi:
There are two new designs for the UI regarding 
- Investigational use only
- Buttons for opening sutdy

# Investigational use only

- [ ] https://github.com/OHIF/Viewers/issues/3916

# Buttons for opening sutdy (zeplin: Component - Launch button)

FYI We won't target the checkbox series for now, it will be in future

![CleanShot 2024-02-26 at 09 44 38](https://github.com/OHIF/Viewers/assets/7490180/ee7f2872-acdf-4cb8-a2fc-cc95b48e948a)

All these are in Zeplin

sedghi:
We just release the OHIF 3.8, you can find more details here https://ohif.org/release-notes/3p8/
If you still encounter this issue in 3.8, please re-open this.

PR #3989, opened 2024-03-14
feat(worklist): New worklist buttons and tooltips

IbrahimCSAE:
### Context

Address https://github.com/OHIF/Viewers/issues/3960
Also fixes issue with dialogs and modals not dismissing on mode exit.
### Changes & Results

New button design
New tooltips

Screenshot from 2024-03-14 04-19-47

Screenshot from 2024-03-14 04-19-53

### Note

Investigational use changes is a separate pull later on.


platform/ui/src/components/Button/Button.tsx
```diff
@@ -66,19 +67,20 @@ const Button = ({
   name,
   className,
   onClick,
+  tooltip = null,
```
sedghi, line 70: it's not button tooltip right? it is icon tooltip

platform/app/src/routes/WorkList/WorkList.tsx
```diff
@@ -382,16 +386,30 @@ function WorkList({
...
-                      variant={isValidMode ? 'contained' : 'disabled'}
+                    <Button
+                      type={ButtonEnums.type.primary}
+                      size={ButtonEnums.size.medium}
                       disabled={!isValidMode}
-                      endIcon={<Icon name="launch-arrow" />} // launch-arrow | launch-info
+                      tooltip={
+                        !isValidMode ? (
+                          <div className="font-inter flex w-[206px] whitespace-normal text-left text-xs font-normal text-white	">
+                            This study does not contain modalities supported by {mode.displayName}{' '}
```
sedghi, line 396: I think the mode itself should provide this text, maybe isValid on the mode should return an object a boolean and a description.

sedghi: For intstance for tmtv, we can say it is activated on studies with both ct and pt

For SEG it is a little bit more 

- on SM -> no support for seg on SM yet
- on DX, MG -> no support for non-volumetric segmentation yet

etc etc

Check wordings with Dan

sedghi (commented):
Looks good, see my comments please

sedghi:
@IbrahimCSAE Can you please add a configuration inside the app config that can people choose in their configurations like default.js and the configuration is about whether to group the enabled modes together at the front and then show the disabled ones. I think that adds a lot of value.

IbrahimCSAE:
> @IbrahimCSAE Can you please add a configuration inside the app config that can people choose in their configurations like default.js and the configuration is about whether to group the enabled modes together at the front and then show the disabled ones. I think that adds a lot of value.

Done

platform/app/src/routes/WorkList/WorkList.tsx
IbrahimCSAE: Done, wording is my own though, I didn't check with Dan on this yet.

platform/ui/src/components/Button/Button.tsx
IbrahimCSAE: Fixed

sedghi (approved):
Looks great thanks

PR #4036, opened 2024-04-15
fix(bugs): enhancements and bug fixes

IbrahimCSAE:
Issue 2: ICONs still not consistent with the design, spacing not consistent
Issue 3: There is a visible shift in the worklist when the study is expanded
Issue 4: Fix color for window level icon
Issue 5: In the M1 the 4D PT/CT is displayed after the disabled microscopy, it should be before since we have grouped the enabled first
Issue 6: Annotation edit still not centered for some reason
Issue 7: The remove tracking is destructive and remove all the annotations, We should Double confirm it with a dialog to say are you sure ?
Issue 8: There is a weird border for the prompt tracking
Issue 12: Probe does not appear in the measurements
Issue 13: DICOM Tag browser weird instance number
Issue 14: Weird precision on the segmentation panel configs
Issue 17: Segmentation panel wrong styles
Issue 20: 3D option has lost the thumb for the range, it is probably because of the new input range merges
Issue 23: The new 4D PT/CT has weird tooltip
Issue 39: Probe is persisting on study
Issue 30: Need margin top for tmtv buttons
Issue 36: display action menu even if there's no WL presets, support fusion WL presets
Issue 37: option to keep patient header always expanded
Issue 38: add enter for hydrate RT,SEG,SR
Issue 45: Add ultrasound directional support to measurement service
Issue 47: move gray scale and xray to top of colorlut list
Issue 48: cant scroll on panel
Issue 28: occupy available space for segmentations
Issue 52: rendering presets padding
Issue 56: modality text too big study list
Issue 57: For MR 3D volume we should probably apply the MR preset instead of the CT 
Issue 58: disable text selection for panel and viewport
Issue 60: action menu doesn't not appear for all modalities
Issue 64: fix modality buttons for colormap
Issue ??: disable text select for toolbar buttons


extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.tsx
```diff
@@ -340,17 +340,61 @@ function PanelStudyBrowserTracking({
...
+          });
+          const measurements = measurementService.getMeasurements();
+          measurements.forEach(m => {
+            if (m.referenceSeriesUID === displaySet.SeriesInstanceUID) {
+              measurementService.remove(m.uid);
+            }
+          });
+        };
+
+        uiDialogService.create({
```
sedghi, line 356: can you refactor this outside? it is adding too much noise

extensions/cornerstone/src/components/WindowLevelActionMenu/VolumeLighting.tsx
```diff
@@ -60,7 +60,11 @@ export function VolumeLighting({
...
             type="range"
             step={0.1}
-            style={{ background: calculateBackground(ambient) }}
+            style={{
+              background: calculateBackground(ambient),
+              '--thumb-inner-color': '#5acce6',
+              '--thumb-outer-color': '#090c29',
+            }}
           />
         )}
```
sedghi, line 69: oh you are not using our inputRange? weird, can you switch to that? hopefully it is not much work

same for all other files

VolumeLighting, VolumeRenderingQuality and VolumeShift

extensions/cornerstone/src/components/WindowLevelActionMenu/VolumeLighting.tsx
```diff
@@ -109,7 +117,11 @@ export function VolumeLighting({
             min={0}
             type="range"
             step={0.1}
-            style={{ background: calculateBackground(specular) }}
+            style={{
+              background: calculateBackground(specular),
+              '--thumb-inner-color': '#5acce6',
+              '--thumb-outer-color': '#090c29',
+            }}
```
sedghi, line 124: all of these inputs should be our inputRange component

sedghi (commented):
see my comments, thanks

extensions/cornerstone/src/components/WindowLevelActionMenu/VolumeLighting.tsx
IbrahimCSAE: The input range component doesn't work for all of them, for shift I have to block calculations when the slider is being dragged, I can modify the input range to support some of this then use it if you would like

sedghi (approved):

PR #4074, opened 2024-05-02
fix(rt): enhanced RT support, utilize SVGs for rendering.

IbrahimCSAE:
### Context

This PR changes the rendering process for RT contours, in the past we used to pass the `gemoteryIds` to our render function and create a VTK actor with the contour data.

Now, this has changed, we are still passing the `gemotryIds`, however, we take those `geomtryIds`, and call `cache.getGeomtry(geomtryId)` on each one, and use its data to create annotations that represent the contour, which gets stored internally and generate an `annotationsUIDsMap`

This helps us render the contours using SVG which is much more reliable than VTK.

From a user point of view, nothing is changing in the RT loading process, because you are still creating and caching geometries, we just handle them differently internally to generate SVGs instead of a VTK actor.

You will notice your  viewports only have 1 actor after loading an RT, which in the past would become 2 actors (1 for the scan,and 1 for RT).


extensions/cornerstone/src/init.tsx
```diff
@@ -144,14 +144,18 @@ export default async function init({
...
-
+  const contourRepresentation = cornerstoneTools.Enums.SegmentationRepresentations.Contour;
   cornerstoneTools.segmentation.config.setGlobalRepresentationConfig(labelmapRepresentation, {
     fillAlpha: 0.5,
     fillAlphaInactive: 0.2,
     outlineOpacity: 1,
     outlineOpacityInactive: 0.65,
   });
 
+  cornerstoneTools.segmentation.config.setGlobalRepresentationConfig(contourRepresentation, {
```
sedghi, line 155: remove

sedghi (changes requested):

sedghi (approved):

Issue #3714, opened 2023-10-11
[Bug] "All" in left study panel only shows "recents"


salimkanoun:
### Describe the Bug

The left study panel has 3 buttons : Primary, Recent and All.
The primary shows the opened study 
Recent shows anteriorities within 1y

I guess All should display all studies according the current but it doesn't the primary study is not listed in it.

### Steps to Reproduce

1 . go to https://viewer-dev.ohif.org/tmtv?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2403.334240657131972136850343327463
2. got to "all" tab of the study panel
3. the All tab displays only one study while it should include all patient's studies including the primary one

### The current behavior

Do not shows the primary study

### The expected behavior

Should display all studies including the primary one

### OS

Linux

### Node version

18

### Browser

Chrome

sedghi:
is this still an issue? and whether Cellian can work on it

salimkanoun:
@sedghi Celian is in school but I created a PR,
The problem here is that we have 2 panels have duplicated code : PanelStudyBrowser (used in most of modes) and PanelStudyBrowserTracking (used in basic mode)

As this panel has a significant logic in it, it would need to refactor both implementation so they can have a single code base to make it maintainable.

sedghi:
I think we can safely close this now

PR #4085, opened 2024-05-04
Fix : Differentiate recent and all in study panel

salimkanoun:
### Context
This PR tries to solve : https://github.com/OHIF/Viewers/issues/3714
It were remaining a partial implementation in which the "recent" tab and "all" tab was displaying the same content.

### Changes & Results

For Primary tab => No changes
For Recent => It now list studies prior to the oldest primary study with a limit time of 1y
For All => It now list all patient's studies without restrictions (including primary studies)

### Testing

This change does not affect Basic mode as this mode uses PanelStudyBrowserTracking, while this PR updated PanelStudyBrowser

@sedghi is there any plane to make this two sharing at least a common code base ?  There a lot of duplicated code between both. 

#### Code

  etc.)

  additions or removals.

#### Tested Environment

Chrome Latest


extensions/default/src/Panels/PanelStudyBrowser.tsx
```diff
@@ -350,15 +350,31 @@ function _createStudyBrowserTabs(primaryStudyInstanceUIDs, studyDisplayList, dis
...
+
+    //Select studies date previous to primary studies and up to one year earlier
+    recentStudies = allStudies.filter(study => {
+      if (!study.date) return false
+      const studyTimeStamp = new Date(study.date).getTime()
+      return (oldestPrimaryTimeStamp - studyTimeStamp) < oneYearInMs
+    })
+
+  }
+
```
sedghi, line 377: I believe the issue here is that it's not following the same logic as `PanelStudyBrowserTracking`. Instead of having it here, could we please export it as a utility so that both `PanelStudyBrowser` and `PanelStudyBrowserTracking` can utilize it?

sedghi: @IbrahimCSAE will do this

IbrahimCSAE:
I'm gonna open another PR for this

IbrahimCSAE:
closing in favor of https://github.com/OHIF/Viewers/pull/4242

Issue #4077, opened 2024-05-02
[Bug] Incorrect zoom value shown in the overlay


syed-qure:
### Describe the Bug

The zoom value that is calculated and shown in the ViewportsOverlay component does not match with the viewport.getZoom() value.

### Steps to Reproduce

1. Open a study and not the zoom value in the bottom left corner.
2. Go to console, get the enabledElement and get the zoom value from viewport object
```javasricpt
enabledElement.viewport.getZoom()
```

<img width="1440" alt="Screenshot 2024-05-02 at 10 04 26 AM" src="https://github.com/OHIF/Viewers/assets/109221583/9d9384cf-64e7-4b1c-97af-205a6a0090d6">

### The current behavior

I think there is something wrong in the calculation of scale inside ViewportOverlay's `updateScale` function

### The expected behavior

The zoom value in the ViewportOverlay component should be equal to `viewport.getZoom()`

### OS

macOS 13.2.1 (22D68)

### Node version

18.19.1

### Browser

Arc v1.40.1 (49250)

syed-qure:
Should I just use `viewport.getZoom()` and use its value in setScale inside ViewportOverlay and call it a day?

```typescript
( evt: Types.EventTypes.CameraModifiedEvent ) => {
  const enabledElement = cornerstone.getEnabledElement( evt.element );
  const { viewport } = enabledElement;

  setScale( viewport.getZoom() );
}
```

sedghi:
We released [OHIF 3.9](https://github.com/OHIF/Viewers/tree/release/3.9) which address this issue

Issue #4039, opened 2024-04-16
[Feature Request] Don’t show irrelevant pop-ups.


tabeascherer:
### What feature or change would you like to see made?

During user testing with radiologists, we discovered that two pop-ups are not needed. 

1. When loading a segmentation, the users see a pop-up that asks, "Do you want to open this Segmentation?". This pop-up is not needed. We can open the segmentation right away, and show it in the list to reduce friction.

2. When performing a measurement, the user is asked, "Track measurement for this series?" We can track the measurement automatically and show it in the list here. It will make a radiologist's workflow smoother. 

### Why should we prioritize this feature?

The two pop-ups are not needed, and we can reduce friction by removing them.

sedghi:
We released [OHIF 3.9](https://github.com/OHIF/Viewers/tree/release/3.9) which address this issue

Issue #4026, opened 2024-04-11
[Bug] List of measurements disappears when the user creates a report


AlexPDesigner:
### Describe the Bug

We conducted usability tests on OHIF Viewer with research and clinical radiologists and one of the findings is a behaviour that surprised our users: **when creating a report for a list of measurements, the list of measurements disappears from the interface.** Our participants got confused and frustrated that their work has disappeared and was no-where to be found in the interface at one glance.

OHIF version: Version number 3.8.0-beta.73 (assumption is previous versions as well)

### Steps to Reproduce

1. Create measurements
2. Tracks measurements
3. Open the list of measurements
4. Create report
5. Measurements list gone 😵‍💫

### The current behavior

The list of measurements disappears when the user creates a report (or a version of the measurements) without informing the users about the result of the action they took - currently, the measurement list just disappears, there is no indication why, where and what the user can do about it. The user can get confused, frustrated and scared about where they work went.

### The expected behavior

1. Create measurements
2. Track measurements by default, no banner needed
3. Open the list of measurements automatically when a measurement is created
4. Change “Create report” to “Save version”
5. User creates the report
6. Measurement list still visible in the interface :bowtie:
7. Positive feedback banner informing the user that the action they took was successful

**For the current bug, we are focusing on step 6: Measurement list still visible in the interface after the user created a report.**

### OS

macOS 14.4.1

### Node version

N/A

### Browser

Chrome Version 123.0.6312.107

sedghi:
i think it throws an error since we don't let public to push to our server and then it removes

AlexPDesigner:
It does trigger an error when the user tries to create a report (create a new SR version actually) in the open source. However, in the local version, when a user tries to save their new version of measurements or Create Report, the software behaves as described above.

sedghi:
We released [OHIF 3.9](https://github.com/OHIF/Viewers/tree/release/3.9) which address this issue

Issue #3991, opened 2024-03-15
[Bug] No support for empty label maps, loading crashes, export ignores


rubenverhack:
### Describe the Bug

I am building a labeling platform in which I want labelers to have a segmentation with three predefined segments present. This is to ensure that every labeler has the correct identifiers. Empty label maps are completely valid for our application, i.e. indicating non-presence of a label globally.

When a DICOM SEG is loaded with 3 segmentation types that are non-empty, it correctly shows the three labels.

![image](https://github.com/OHIF/Viewers/assets/28840027/7925f506-d776-4009-bfe8-ec3c0db92252)

(1) BUG => If one of segments is empy (all labels are set to zero), then the loader hangs at 100%.

![image](https://github.com/OHIF/Viewers/assets/28840027/0b6718d7-c815-47b8-8d29-100ce6190bb5)

(2) When making a Segmentation in OHIF, and an empty segment is added, then it is not exported. 

Tested on version v3.8.0-beta.59

### Steps to Reproduce

I have added a DICOM SEG file with empty label maps for reference.

[2.25.337304097621488642328905952564813011907.zip](https://github.com/OHIF/Viewers/files/14614162/2.25.337304097621488642328905952564813011907.zip)

### The current behavior

(1) loading => loader crashes
(2) exporting => empty segments are deleted

### The expected behavior

Empty label maps should just be loaded as a label with only zeros.

Empty segments should be exported as empty DICOM SEG objects (containing all layers being all set to zero)

### OS

Ubuntu 22.04

### Node version

OHIF v3.8.0-beta.59

### Browser

Chrome

sedghi:
can you provide the referencedSeiries as well? (CT, MR) the image itself

rubenverhack:
Here's the anonimized data. Thank you for your time.

https://drive.google.com/file/d/1HegTfExX9pJwxx4br42jPz5CbnBlBea_/view?usp=drive_link

sedghi:
We released [OHIF 3.9](https://github.com/OHIF/Viewers/tree/release/3.9) which address this issue

PR #4086, opened 2024-05-06
fix(bugs): enhancements and bugs in several areas

IbrahimCSAE:
### Context and Changes & Results

Fix the following issues

- https://github.com/OHIF/Viewers/issues/3955
- Resolution:  set `blurInputOnSelect` to `true` in Select.tsx

- https://github.com/OHIF/Viewers/issues/4026
- Resolution: use the new `disableConfirmationPrompts` property in your config file, this will resolve the issue because when you save a report, it will ask you to hydrate the SR after it's done, by disabling confirmation prompts, it gets automatically hydrated again and measurement list will be populated.

- https://github.com/OHIF/Viewers/issues/4039
- Resolution: added `disableConfirmationPrompts` configuration option, if set to true, there's no confirmation required for actions such as hydrating SR, RT, SEG, or tracking measurements.

- https://github.com/OHIF/Viewers/issues/3908
- Resolution: Added the tool to basic viewer mode, and test mode, under more tools section

- https://github.com/OHIF/Viewers/issues/3944
- Resolution: Added proper error handling so the tracking state machine doesn't enter `idle` state if creating the report fails, now it correctly stays in `tracking` state if this occurs.

- https://github.com/OHIF/Viewers/issues/3803
- Resolution: Restored the correct crosshair colors for MPR display.

- https://github.com/OHIF/Viewers/issues/2233
- Resolution: Added DICOM Tag Browser tool to Microscopy mode

- https://github.com/OHIF/Viewers/issues/4077
- Resolution: Utilized `viewport.getZoom()` which is more reliable than the old calculation which was wrong.

- https://github.com/OHIF/Viewers/issues/3991
- Fixed two different issues, segmentation loading hanging when centroid is not defined (could happen if labelmap is empty) and fixed issue if users create an empty label map and export the seg, the empty labelmap is ignored in the generated DICOM SEG, now the segment will still be present just empty.


extensions/cornerstone-dicom-rt/src/utils/promptHydrateRT.ts
```diff
@@ -15,23 +15,28 @@ function promptHydrateRT({
...
         callback();
       });
 
-      const isHydrated = await hydrateRTDisplaySet({
-        rtDisplaySet,
-        viewportId,
-        toolGroupId,
-        servicesManager,
-      });
+      window.setTimeout(async () => {
```
sedghi, line 30: why?

platform/app/public/config/default.js
```diff
@@ -13,6 +13,7 @@ window.config = {
   showLoadingIndicator: true,
   strictZSpacingForVolumeViewport: true,
   groupEnabledModesFirst: true,
+  disableConfirmationPrompts: false,
```
sedghi, line 16: It should be disabled by default. If someone wants it, they can enable it. However, the default setting should hide it.

sedghi (commented):
see my comments please

extensions/cornerstone-dicom-rt/src/utils/promptHydrateRT.ts
IbrahimCSAE: viewport.element is undefined otherwise, not sure what the cause is, guess it was never a problem before because the user would be too slow to click hydrate for that to happen.

platform/app/public/config/default.js
IbrahimCSAE: Sure, I thought we would keep it as it is, I’ll make confirmations hidden

sedghi (approved):

PR #4113, opened 2024-05-09
fix(crosshairs): reset angle, position, and slabthickness for crosshairs when reset viewport tool is used

IbrahimCSAE:
### Context

reset angle, position, and slabthickness for crosshairs when reset viewport tool is used

and fixes this bug https://github.com/OHIF/Viewers/issues/4158


IbrahimCSAE:
deploy will fail here until CS3D is deployed.

modes/tmtv/src/initToolGroups.js
```diff
@@ -125,12 +140,26 @@ function _initToolGroups(toolNames, Enums, toolGroupService, commandsManager, mo
...
+                colours[viewportOptions.id] ||
+                colorsByOrientation[viewportOptions.orientation] ||
+                '#0c0'
+              );
+            } else {
+              console.warn('missing viewport?', viewportId);
+              return '#0c0';
+            }
+          },
         },
```
sedghi, line 163: i don't think we should enable it in tmtv

sedghi (approved):

PR #4115, opened 2024-05-09
Updating the _validateRequiredSelectors method

Sofien-Sellami:
### Context
Certain custom attributes require additional data.

### Changes & Results
Updated the _validateRequiredSelectors method in the HangingProtocolService to pass additional props to the findMatch function, allowing for better handling of custom attributes.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: 
- [x] Node version: 
- [x] Browser:
  

### DIFF
--- a/platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
+++ b/platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
@@ -1432,7 +1432,11 @@ export default class HangingProtocolService extends PubSubService {
     // only match the required rules
     const requiredRules = seriesMatchingRules.filter(rule => rule.required);
     if (requiredRules.length) {
-      const matched = this.protocolEngine.findMatch(displaySet, requiredRules);
+      const matched = this.protocolEngine.findMatch(displaySet, requiredRules, {
+        studies: this.studies,
+        instance: displaySet.images?.[0],
+        displaySetMatchDetails: this.displaySetMatchDetails,
+      });
 
       if (!matched || matched.score === 0) {
         throw new Error(


salimkanoun:
@sedghi finally it is not a regression, just never been working before

sedghi:
How can we test it? what was not working that is now after this PR working?

salimkanoun:
Here is a sample of one of our custom rule : 

`import { utils } from '@ohif/core/src';

const { formatDate } = utils;

export default function (displaySet, options) {
  const { constraint } = this;
  const constraintValue = constraint[Object.keys(constraint)[0]];
  if (!options) {
    return getTestValue(constraintValue);
  }

  const { studies } = options;

  if (studies.length <= 1) {
    return getTestValue(constraintValue);
  }

  let result = 0;
  studies.forEach((study) => {
    if (study.StudyInstanceUID !== displaySet.StudyInstanceUID) {
      const difference =
        getTimeStamp(displaySet.instance['StudyDate']) -
        getTimeStamp(study.series[0].instances[0]['StudyDate']);
      result = difference;
    }
  });
  return result;
}

const getTimeStamp = (dateString: string) => {
  const timestamp = new Date(formatDate(dateString)).getTime() / 1000;
  return Math.floor(timestamp);
};
const getTestValue = (options) => {
  if (Array.isArray(options)) {
    return options.map((option) => option?.value ?? option);
  } else {
    return options?.value ?? options;
  }
};
`

The idea of this custom role is to have a rule in our HP saying "this viewport shall display a displayset that is earlier than the other known display set
This way we can have a HP in which we make sure the viewport in a particular location displays prior studies and other viewport display the current study.

To do so we need to get the studies candidate so we can compute the timestamp difference and then apply a rule on this timestamp difference.
Also Instance make possible to access the first instance metadata, such as special tags like CorrectedImage in PET that are not available otherwise

sedghi:
Are we moving to shadcn components? If so, can we hold off on this? Or do you need it?

sedghi:
@salimkanoun I want to test this before merging. Could you please let me know where to add your code? Is it like a custom attribute? Can you perhaps provide me with the code for the full hanging protocol that I can run with studies that have priors?

sedghi:
Do you guys still need this?

salimkanoun:
Yes but didn't get the time to give you a minimal HP to reproduce why we need it, will do ASAP

PR #4063, opened 2024-04-23
fix(toolbox): Preserve user-specified tool state and streamline command execution

sedghi:
needs this https://github.com/cornerstonejs/cornerstone3D/pull/1225

### Context

# Breaking Changes
## The Double Range option
The `values` is renamed to `value`

CleanShot 2024-04-23 at 16 51 17@2x

## Toolbox (with Options) Commmands

Now the top level `commands` is removed and each `commands` in the options automatically runs upon entry to make sure the state is preserved

CleanShot 2024-04-23 at 16 51 28@2x

previously there were two sets of commands, one at the top level and one at the options level which caused a lot of issues
```js
 {
    id: 'Shapes',
    uiType: 'ohif.radioGroup',
    props: {
      label: 'Shapes',
      evaluate: {
        name: 'evaluate.cornerstone.segmentation',
        toolNames: ['CircleScissor', 'SphereScissor', 'RectangleScissor'],
      },
      icon: 'icon-tool-shape',
      commands: _createSetToolActiveCommands('CircleScissor'),
      options: [
        {
          name: 'Shape',
          type: 'radio',
          value: 'CircleScissor',
          id: 'shape-mode',
          values: [
            { value: 'CircleScissor', label: 'Circle' },
            { value: 'SphereScissor', label: 'Sphere' },
            { value: 'RectangleScissor', label: 'Rectangle' },
          ],
          commands: 'setToolActiveToolbar',
        },
      ],
    },
  },

```

Now if you have options, you can just have the commands options, and that will make sure the state is correctly set based on the latest state

```js
{
    id: 'Shapes',
    uiType: 'ohif.radioGroup',
    props: {
      label: 'Shapes',
      evaluate: {
        name: 'evaluate.cornerstone.segmentation',
        toolNames: ['CircleScissor', 'SphereScissor', 'RectangleScissor'],
      },
      icon: 'icon-tool-shape',
      options: [
        {
          name: 'Shape',
          type: 'radio',
          value: 'CircleScissor',
          id: 'shape-mode',
          values: [
            { value: 'CircleScissor', label: 'Circle' },
            { value: 'SphereScissor', label: 'Sphere' },
            { value: 'RectangleScissor', label: 'Rectangle' },
          ],
          commands: 'setToolActiveToolbar',
        },
      ],
    },
  },
```

Toolbox was resetting the user specified tool

Previously

https://github.com/OHIF/Viewers/assets/7490180/e5b70f00-37bf-466a-ba60-5aba51f1d8b3

[...]

PR #4119, opened 2024-05-13
fix(tmtv): threshold was crashing the side panel

IbrahimCSAE:
### Context

threshold was crashing the side panel

### Changes & Results

fix the typos that were causing the threshold panel to crash


platform/ui/src/components/AdvancedToolbox/ToolSettings.tsx
```diff
@@ -103,7 +103,7 @@ const renderDoubleRangeSetting = option => {
       key={option.id}
     >
       <InputDoubleRange
-        values={option.value}
+        values={option.values}
```
sedghi, line 106: this was correct, i intentionally made it value https://github.com/OHIF/Viewers/pull/4063#issue-2259722444

sedghi (approved):

PR #4127, opened 2024-05-15
feat(window-level-region): add window level region tool

IbrahimCSAE:
### Context

Add window level region tool to the viewer in basic mode, segmentation, and test.


IbrahimCSAE:
needs an icon @dan-rukas we also need an icon for cobb angle please.

sedghi (approved):
Thanks

PR #4144, opened 2024-05-17
fix(select): utilize react portals for select component

IbrahimCSAE:
### Context

Fixes select component overflow

### DIFF
--- a/platform/ui/src/components/Select/Select.css
+++ b/platform/ui/src/components/Select/Select.css
@@ -27,16 +27,16 @@
   padding: 4px;
 }
 
-.customSelect__wrapper .customSelect__option {
-  color: #d6d6d6;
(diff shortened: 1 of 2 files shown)


platform/ui/src/components/Select/Select.css
```diff
@@ -55,26 +55,26 @@
   @apply inline-block truncate px-3 py-2;
 }
 
-.ohif-select .customSelect__menu {
+.customSelect__menu-portal .customSelect__menu {
```
sedghi, line 58: why?

sedghi:
There is some weird behaviour as well look here pleases

CleanShot 2024-05-17 at 12 14 01@2x

sedghi (approved):

PR #4152, opened 2024-05-21
fix: get direct url pixel data should be optional for video

igoroctaviano:
Refactor getDirectURL into three documented and tested functions. Also, this change allows building /rendered URLs for video instances that does not have pixel data present.

### Context

The get direct URL was a multi-purpose util function. This PR splits it into three independent functions that get called within to get the direct url.

### Changes & Results

Now getDirectURL calls getBulkDataValue and createRenderedRetrieve as fallbacks.

### Testing

- Load a video study using json loader (with the URL attribute present) and observe that it will be used as direct url and later as video src for video modality
- Load a study with DirectRetrieveURL and observe that it will be used as a direct URL
- Load a study with InlineBinary and observe that it will be used as a direct URL
- Load a study with either 'PixelData' or 'EncapsulatedDocument' and observe that /rendered URL is built and returned as getDirectURL
- Load a study with BulkDataURI and observe that it will be used as direct URL with the correct accept query param

(!singlepart || (singlepart !== true && singlepart.indexOf(fetchPart) === -1)

#### PR

#### Code

  etc.)

N/A

#### Tested Environment

- [x] "OS: 14.4.1 (23E224) 
- [x] "Node version: 16.14 
- [x] "Browser: Chrome Version 123.0.6312.59 (Official Build) (x86_64)
  


extensions/default/src/utils/getDirectURL.js
```diff
@@ -29,31 +29,31 @@ const getDirectURL = (config, params) => {
   }
 
   const value = instance[tag];
-  if (!value) {
```
wayfarer3130, line 32: I don't understand what value there is in increasing the indent level.  The recommended approach for handling things is to return quickly at the top of the function, rather than nesting a function inside, as was done originally.

wayfarer3130: Ok, I see what you are doing, github made it non obvious that there was more code - cut things off below.  What I think should be done is to separate this out into two functions, one of which constructs a URL, the other of which uses the value - and export both of them so that they can be used independently, or combined.  Then getDirectURL becomes:

```
const value = instance[tag];
return getBulkdataValue(value) || createRenderedRetrieve(instance)

Then, you can leave the return immediate as is inside getBulkdataValue.
Also, you might add a note that it is incorrect to not have a value here, or rather it means that the value itself is supposed to be empty, but that some DICOMweb systems incorrectly return empty tag values.
Then, we should handle both:

TAG: null
and
TAG: {}

I'd like to see a unit test to check a couple of the various cases, since I think this logic is getting complex enough to warrant that.  

Sorry for the extra work, but by the time things get this complex, splitting it up makes it easier to figure out what one should expect to receive back.

wayfarer3130 (approved):

wayfarer3130:
Can you update to head and check that the tests pass before committing please?  Otherwise I like how you've added the tests and separated out the functionality to make it clearly understandable.

PR #4157, opened 2024-05-22
fix(rt): dont convert to volume for RTSTRUCT

IbrahimCSAE:
### Context

Prevent viewport from going into volume when an RT is loaded, also needs this

https://github.com/cornerstonejs/cornerstone3D/pull/1269


jenny-hm-lee:
@IbrahimCSAE , sorry to bother you.
as you testing this fix for RTSTRUCT, can you check if the following scenarios would cause an error?

Different scenarios:
1. load a RTSTRUCT, then load a SR
2. load a RTSTRUCT, then load a SEG
3. load a SEG, then load a RTSTRUCT
4. load a SEG, then load a SR
5. load a SR, then load a RTSTRUCT
6. load a SR, then load a SEG

sedghi:
can you pull master and resolve

sedghi (approved):

PR #4160, opened 2024-05-22
feat(measurements): show untracked measurements in measurement panel under additional findings

IbrahimCSAE:
### Context

Update the following measurement types to be displayed in the measurement panel as “NON ACQUISITION MESUREMENTS”:

Angle
ArrowAnnotate
Bidirectional
CircleROI
EllipticalROI
Length
PlanarFreehandROI
Probe
RectangleROI

CSV export will export all the details for non acquisition measurements, SR export filters them out.

Pending design @sedghi @dan-rukas


extensions/cornerstone/src/utils/measurementServiceMappings/Angle.ts
```diff
@@ -76,11 +76,12 @@ const Angle = {
       data: data.cachedStats,
       type: getValueTypeFromToolType(toolName),
       getReport,
+      nonAcquisition: referencedImageId == null,
```
sedghi, line 79: can we just insert the referencedImageId in here, it might be useful in future

sedghi: so for nonAcquisition you can check its presence

extensions/cornerstone/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
```diff
@@ -7,12 +7,24 @@ import * as cornerstone from '@cornerstonejs/core';
...
     return _getUIDFromImageID(imageId);
   }
 
-  // Todo: implement for volume viewports and use the referencedSeriesInstanceUID
+  const { metadata } = annotation;
+  const { volumeId } = metadata;
+
+  const displaySet = displaySetService.getDisplaySetsBy(displaySet =>
+    volumeId.includes(displaySet.uid)
+  )[0];
```
sedghi, line 20: what happens for stack viewport that doesn't have volumeId, this logic doesn't look correct to me

platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -60,6 +60,7 @@ const MEASUREMENT_SCHEMA_KEYS = [
   'cachedStats',
   'selected',
   'textBox',
+  'nonAcquisition',
```
sedghi, line 63: referencedImageId

extensions/cornerstone/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
IbrahimCSAE: Stack shouldnt make it down to this code, it would have the reference imageId, only volume viewport is expected to go down

extensions/cornerstone/src/utils/measurementServiceMappings/Angle.ts
IbrahimCSAE: Sure

platform/core/src/services/MeasurementService/MeasurementService.ts
IbrahimCSAE: Sure

extensions/cornerstone/src/utils/measurementServiceMappings/utils/getSOPInstanceAttributes.js
IbrahimCSAE: resolved the rest, for stack I'm assuming referencedImageId should always be present, only volume viewport is making it down to this.

sedghi (changes requested):
Can we change these 


CleanShot 2024-05-24 at 11 45 10@2x

--------

For some reason the bidirectional, circle, rectangle don't work with untracked

CleanShot 2024-05-24 at 11 49 17@2x

------

can we round the stat for freehand

CleanShot 2024-05-24 at 11 50 17@2x

IbrahimCSAE:
https://github.com/cornerstonejs/cornerstone3D/pull/1281

Needs this then everything should work fine, I fixed button name and sub panel name, and rounding for freehand

sedghi (approved):

PR #4164, opened 2024-05-23
fix(SharedArrayBuffer/Segmentation): fix usesharedarraybuffer from config in Segmentation Service

Michael-Andersen:
### Context

When useSharedArrayBuffer was set to false in config in an environment not set up for Cross Origin Isolation and I tried to load a SEG, the viewer would crash and give an error related to CrossOriginIsolation, I noticed it was trying to use the SharedArrayBuffer.

### Changes & Results

Correctly parse useSharedArrayBuffer from the config file in the SegmentationService.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: Windows 10
- [] Node version: v16.17.0
- [] Browser: Version 124.0.6367.119 (Official Build) (64-bit)
  

### DIFF
--- a/extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
+++ b/extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
@@ -545,7 +545,7 @@ class SegmentationService extends PubSubService {
       volumeId: segmentationId,
       targetBuffer: {
         type: 'Uint8Array',
-        sharedArrayBuffer: window.SharedArrayBuffer,
+        sharedArrayBuffer: window.config?.useSharedArrayBuffer,
       },
     });
     const derivedVolumeScalarData = derivedVolume.getScalarData();
@@ -770,8 +770,8 @@ class SegmentationService extends PubSubService {
     const segmentIndices = segmentIndex
       ? [segmentIndex]
       : segmentation.segments
-          .filter(segment => segment?.segmentIndex)
-          .map(segment => segment.segmentIndex);
+        .filter(segment => segment?.segmentIndex)
+        .map(segment => segment.segmentIndex);
 
     const segmentIndicesSet = new Set(segmentIndices);
 
@@ -966,7 +966,7 @@ class SegmentationService extends PubSubService {
       volumeId: segmentationId,
       targetBuffer: {
         type: 'Uint8Array',
-        sharedArrayBuffer: window.SharedArrayBuffer,
+        sharedArrayBuffer: window.config?.useSharedArrayBuffer,
       },
     });
 
--- a/extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
+++ b/extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
@@ -539,13 +539,29 @@ class SegmentationService extends PubSubService {
       throw new Error(`No volume found for referencedVolumeId: ${referencedVolumeId}`);
     }
 
+    const useSharedArrayBufferValue = window.config?.useSharedArrayBuffer;
+    let sharedArrayBufferEnabled = true;
+
+    if (useSharedArrayBufferValue === 'AUTO') {
+      try {
+        /*eslint-disable no-constant-condition */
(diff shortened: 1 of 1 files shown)


Michael-Andersen:
@sedghi please review when you have time. I know there's some other ways to get the config values so if you don't like this approach, we can change it, let me know.

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -545,7 +545,7 @@ class SegmentationService extends PubSubService {
       volumeId: segmentationId,
       targetBuffer: {
         type: 'Uint8Array',
-        sharedArrayBuffer: window.SharedArrayBuffer,
+        sharedArrayBuffer: window.config?.useSharedArrayBuffer,
```
sedghi, line 548: I think config.useSharedArrayBuffer is a string, that can be 'FALSE', 'AUTO' , 'TRUE' so it is not a boolean

Michael-Andersen: Ah I see my bad, I have updated to handle as a string, having a getter in Cornerstone could reduce code duplication.

To be honest, I'm not 100% sure I understand this problem. Having the old code with` window.SharedArrayBuffer` definitely causes an issue. Even running OHIF master branch, connected to your PACS with no config modifications on localhost loading a seg gave me a CORS type error and crash. But using my new method of pulling from the config it seems to work with true and false values.

My team needs a change like this so I have implemented it. If you want to look further into the issue or wait for others who rely on disabling the SharedArrayBuffer to report that's ok too.

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -539,13 +539,29 @@ class SegmentationService extends PubSubService {
       throw new Error(`No volume found for referencedVolumeId: ${referencedVolumeId}`);
     }
 
+    const useSharedArrayBufferValue = window.config?.useSharedArrayBuffer;
```
wayfarer3130, line 542: You can't assume the window.config is an object - it might be a function and have gotten initialized during setup.  You will have to fetch it appropriately.  I checked and it seems that appConfig is not being passed into the constructor, so my suggestion, which @sedghi should look at is to modify ServicesManager to have a setAppConfig call that takes the application configuration, and stores it, then in the create call to the service, passes in that appConfig.  You will have to call setAppConfig in appInit, but there are other related calls, so I think that is probably the right way of doing things.

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -539,13 +539,29 @@ class SegmentationService extends PubSubService {
...
+    let sharedArrayBufferEnabled = true;
+
+    if (useSharedArrayBufferValue === 'AUTO') {
+      try {
+        /*eslint-disable no-constant-condition */
+        if (!new SharedArrayBuffer(0)) {
```
wayfarer3130, line 548: All of this code to detect the flag should be isolated to a separate function, which stores the result of the test, so that on subsequent calls it returns immediately.  That test should be in the platform/core functionality as it is shared functionality.

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -960,13 +976,29 @@ class SegmentationService extends PubSubService {
 
     const segmentationId = options?.segmentationId ?? `${csUtils.uuidv4()}`;
 
+    const useSharedArrayBufferValue = window.config?.useSharedArrayBuffer;
```
wayfarer3130, line 979: Again, don't repeat the code here, use a shared function that you import from core

sedghi:
I'm fixing this in cornerstone

Michael-Andersen:
> I'm fixing this in cornerstone

Great, thanks for the heads up.

sedghi:
https://github.com/cornerstonejs/cornerstone3D/pull/1290

PR #4178, opened 2024-05-27
feat: Add simple multi-monitor support

wayfarer3130:
> [!NOTE]
> We plan to enhance this service in the future. Currently, it offers a basic implementation of multi-monitor support, allowing you to manually open multiple windows on the same monitor. It is not yet a full multi-monitor solution!

The multi-monitor service provides detection, launch, and communication support for multiple monitors or windows/screens within a single monitor.

> [!NOTE]
> The multi-monitor service is currently applied via a configuration file.

```js
customizationService: ['@ohif/extension-default.customizationModule.multimonitor'],
```

## Configurations
The service supports two predefined configurations:

1. **Split Screen (`multimonitor=split`)**
   Splits the primary monitor into two windows.

2. **Multi-Monitor (`multimonitor=2`)**
   Opens windows across separate physical monitors.

### Launch Methods
- Specify `&screenNumber=0` to designate the first window explicitly.
- Omit `screenNumber` to let the service handle window assignments dynamically.
- Use `launchAll` in the query parameters to launch all configured screens simultaneously.

#### Example URLs:
- **Split Screen:**
  `http://viewer.ohif.org/.....&multimonitor=split`
  Splits the primary monitor into two windows when a study is viewed.

- **Multi-Monitor with All Screens:**
  `http://viewer.ohif.org/.....&multimonitor=2&screenNumber=0&launchAll`
  Launches two monitors and opens all configured screens.

---

### Refresh, Close and Open
If you refresh the base/original window, then all the other windows will also
refresh.  However, you can safely refresh any single other window, and on the next
command to the other windows, it will re-create the other window links without
losing content in the other windows.  You can also close any other window and
it will be reopened the next time you try to call to it.

## Executing Commands
The MultiMonitorService adds the ability to run commands on other specified windows.
This allows opening up a study on another window without needing to refresh
it's contents.  The command below shows an example of how this can be done:


wayfarer3130:
In the starting phase, launch a study using multimonitor=split
This will show in the current window
Select the all button to show all studies for a patient, and click on the launch study button on the right hand side of the study list:
image

wayfarer3130:
This results in a second page being displayed with the specified study:
image

wayfarer3130:
You can then navigate either window from the 'all' studies in the study browser tab.

Study navigation currently refreshes the page, which loses the position on that page, plus any markup/annotations you have created.  This is a temporary issue while the navigation is updated to use internal navigation to preserve data.

platform/core/src/services/MultiMonitorService.ts
```diff
@@ -0,0 +1,164 @@
...
+    create: ({ configuration, commandsManager }): MultiMonitorService => {
+      const service = new MultiMonitorService(configuration, commandsManager);
+      return service;
+    },
+  };
+
+  constructor(configuration, commandsManager) {
+    const params = new URLSearchParams(window.location.search);
+    const screenNumber = params.get('screenNumber');
+    const multimonitor = params.get('multimonitor');
```
salkz, line 25: Since the browser doesn't allow us to check whether a window with a specific name exists without calling `window.open` then I think another param, eg `alwaysOpen`, could prove useful. This would call `window.open` on all configured windows on init so it gets all references right away. 

The use case for this is making development easier as every code change triggers a refresh (webpack build) which loses references to previously opened windows. 

For example, if I have 2 windows opened, I refresh, call some multi window command, then it fails because it doesn't know about the other window(s) because it hasn't technically opened them yet.

platform/core/src/services/MultiMonitorService.ts
```diff
@@ -0,0 +1,164 @@
...
+        window.name = this.screenConfig.id;
+        return;
+      }
+    }
+    this.numberOfScreens = 1;
+    this.isMultimonitor = false;
+  }
+
+  public run(screenDelta = 1, commands, options) {
+    const screenNumber = (this.screenNumber + (screenDelta ?? 1)) % this.numberOfScreens;
```
salkz, line 59: Using negative delta to run commands from window 2 into window 1 feels a bit weird I think. I could see explicit screen id being more intuitive. Or what's the use case with delta

platform/core/src/services/MultiMonitorService.ts
```diff
@@ -0,0 +1,164 @@
...
+        }
+        window.name = this.screenConfig.id;
+        return;
+      }
+    }
+    this.numberOfScreens = 1;
+    this.isMultimonitor = false;
+  }
+
+  public run(screenDelta = 1, commands, options) {
```
salkz, line 58: Thoughts about a functions that would run commands and options in all windows except the origin from which it was called?

platform/core/src/services/MultiMonitorService.ts
wayfarer3130: That is a future enhancement - I'd like to proceed with a very simple multi monitor setup to start with, and then see how it works.  Right now, the first window is always window 0 and can't actually position itself correctly due to browser limitations..  Other than that, the other windows will refresh themselves when they are needed, because the system detects the missing references, and clears them if on clears another window.

platform/core/src/services/MultiMonitorService.ts
wayfarer3130: Yes, future enhancement.  This is baby steps to get started.

platform/core/src/services/MultiMonitorService.ts
wayfarer3130: I don't know of a good solution for refreshing either window and still being able to access the other monitor cleanly without sending a new open window command.  The problem is that the references are gone.  For me, if I refresh W0 then sending a command to W1 will work, it just does it by reopening the window by name.  The other way around, W1 will work if I first load a new study into W1 from W0.  That is, refresh W1, load a study from W0 into W1 and then one can run commands.

wayfarer3130: A back end service would work of course, but that requires specific browser support.

wayfarer3130: There is apparently a way to restore the references without causing a secondary reload, but it requires naming the original window correctly as well as not changing the URL of any secondary windows, so I'm going to have to redo a bit of the code to get that working.  Stay tuned.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -999,6 +1020,7 @@ export default class HangingProtocolService extends PubSubService {
 
     try {
       if (!this.protocol || this.protocol.id !== protocol.id) {
+        console.log('***** Resetting protocol', this.protocol, this.activeStudy);
```
sedghi, line 1023: remove

extensions/default/src/hangingprotocols/viewportOptions.ts
```diff
@@ -0,0 +1,18 @@
+/** A default viewport options */
+export const viewportOptions = {
```
sedghi, line 2: I think we should move this and other selectors to a utils folder to keep it organized and clarify that it is not a hanging protocol, similar to other files in this path.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -372,8 +372,26 @@ export default class HangingProtocolService extends PubSubService {
...
+    return this.studies.some(it => it.StudyInstanceUID === studyUID);
+  }
+
+  public addStudy(study) {
+    if (!this.hasStudyUID(study.StudyInstanceUID)) {
+      this.studies.push(study);
+    }
+  }
+
+  public setDisplaySets(displaySets) {
```
sedghi, line 393: I don't understand why we need this, since `.run` already has `displaySets` to work with.

```javascript
public run({ studies, displaySets, activeStudy, activeStudyUID }, protocolId, options = {}) {
```

It also seems like we're mixing things up by adding `hpservice.setDisplaySets`. How about letting `run` handle it?

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -372,8 +372,26 @@ export default class HangingProtocolService extends PubSubService {
...
+
+  public hasStudyUID(studyUID: string): boolean {
+    return this.studies.some(it => it.StudyInstanceUID === studyUID);
+  }
+
+  public addStudy(study) {
+    if (!this.hasStudyUID(study.StudyInstanceUID)) {
+      this.studies.push(study);
+    }
+  }
```
sedghi, line 391: There are too many instances of `study` versus `studyUID` in this file. I'm unsure why we have both, and this is contributing to the confusion. Can we simplify it to work with `study`, as it seems sufficient for our needs?

```typescript
export interface StudyMetadata extends Record<string, unknown> {
  readonly StudyInstanceUID?: string;
  StudyDescription?: string;
}
```

so I propose `hasStudy(study)` of `hasStudyUID`

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -393,16 +411,19 @@ export default class HangingProtocolService extends PubSubService {
    *        the studies to display in viewports.
    * @param protocol is a specific protocol to apply.
    */
-  public run({ studies, displaySets, activeStudy }, protocolId, options = {}) {
+  public run({ studies, displaySets, activeStudy, activeStudyUID }, protocolId, options = {}) {
```
sedghi, line 414: No need to provide `activeStudyUID` since we already have `activeStudy`, which includes `activeStudyUID`.

platform/core/src/classes/CommandsManager.ts
```diff
@@ -220,6 +226,47 @@ export class CommandsManager {
 
     return result;
   }
+
+  /** Like run, but await each command before continuing */
+  public async runAsync(
```
sedghi, line 231: i wanted this many times before, thanks

How about some refactoring 

```js
  private async executeCommands(
    commands: ComplexCommand[],
    options: Record<string, unknown>,
    asyncMode: boolean
  ) {
    let result: unknown;

    for (const command of commands) {
      const { commandName, commandOptions, context } = command;
      if (commandName) {
        const commandOptionsMerged = { ...options, ...commandOptions };
        if (asyncMode) {
          result = await this.runCommand(commandName, commandOptionsMerged, context);
        } else {
          result = this.runCommand(commandName, commandOptionsMerged, context);
        }
      } else if (typeof command === 'function') {
        result = asyncMode ? await command() : command();
      } else {
        console.warn('No command name supplied in', command);
      }
    }

    return result;
  }

  public run(
    toRun: Command | Commands | Command[] | string | undefined,
    options: Record<string, unknown> = {}
  ): unknown {
    if (!toRun) return;

    const commands = CommandsManager.convertCommands(toRun);
    if (commands.length === 0) {
      console.log("Command isn't runnable", toRun);
      return;
    }

    return this.executeCommands(commands, options, false);
  }

  public async runAsync(
    toRun: Command | Commands | Command[] | string | undefined,
    options: Record<string, unknown> = {}
  ): Promise<unknown> {
    if (!toRun) return;

    const commands = CommandsManager.convertCommands(toRun);
    if (commands.length === 0) {
      console.log("Command isn't runnable", toRun);
      return;
    }

    return this.executeCommands(commands, options, true);
  }
  ```

File: extensions/default/src/hangingprotocols/viewportOptions.ts (hunk 5 above)
Comment (sedghi) [reply extensions/default/src/hangingprotocols/viewportOptions.ts:2]: Maybe we also call it `defaultViewportOptions`?

File: extensions/default/src/hangingprotocols/hpMNGrid.ts [hunk 10]
```diff
@@ -309,4 +215,341 @@ const hpMN: Types.HangingProtocol.Protocol = {
...
 };
 
+/**
+ * This hanging protocol can be activated on the primary mode by directly
+ * referencing it in a URL or by directly including it within a mode, e.g.:
+ * `&hangingProtocolId=@ohif/mnGrid` added to the viewer URL
+ * It is not included in the viewer mode by default.
+ */
+export const hpMN8: Types.HangingProtocol.Protocol = {
+  ...hpMN,
```
sedghi, line 225: hmmm, why ...hpMN?

extensions/default/src/hangingprotocols/hpMNGrid.ts
```diff
@@ -309,4 +215,341 @@ const hpMN: Types.HangingProtocol.Protocol = {
   numberOfPriorsReferenced: -1,
 };
 
+/**
+ * This hanging protocol can be activated on the primary mode by directly
+ * referencing it in a URL or by directly including it within a mode, e.g.:
+ * `&hangingProtocolId=@ohif/mnGrid` added to the viewer URL
```
sedghi, line 221: @ohif/mnGrid8

extensions/default/src/hangingprotocols/hpMNGrid.ts
```diff
@@ -309,4 +215,341 @@ const hpMN: Types.HangingProtocol.Protocol = {
...
+            {
+              matchedDisplaySetsIndex: 5,
+              id: 'defaultDisplaySetId',
+            },
+          ],
+        },
+      ],
+    },
+
+    ...hpMN.stages,
```
sedghi, line 392: same here

extensions/default/src/hangingprotocols/hpMNGrid.ts
```diff
@@ -309,4 +215,341 @@ const hpMN: Types.HangingProtocol.Protocol = {
...
+
+/**
+ * This hanging protocol extends the default protocol with additional
+ * images on the second monitor.  It assumes the first four images are shown
+ * on monitor 0
+ */
+export const hpMNMonitor2: Types.HangingProtocol.Protocol = {
+  ...hpMN,
+  id: '@ohif/mnGridMonitor2',
+  description: 'A basic two monitor view',
```
sedghi, line 404: I guess the description is "second monitor hp in a two monitor config"

extensions/default/src/commandsModule.ts
```diff
@@ -141,6 +197,7 @@ const commandsModule = ({
       reset = false,
     }: HangingProtocolParams): boolean => {
       try {
+        console.log('******** Set hanging protocol', activeStudyUID);
```
sedghi, line 200: remove

extensions/default/src/Panels/requestDisplaySetCreationForStudy.js
```diff
@@ -13,7 +13,7 @@ function requestDisplaySetCreationForStudy(
     return;
   }
 
-  dataSource.retrieve.series.metadata({ StudyInstanceUID, madeInClient });
+  return dataSource.retrieve.series.metadata({ StudyInstanceUID, madeInClient });
```
sedghi, line 16: There seems to be two files named `requestDisplaySetCreationForStudy` , can we re-use one

platform/core/src/classes/CommandsManager.ts
salkz: Seeing as `CommandsManager.convertCommands(toRun)` never returns functions or strings then you can drop `typeof command === 'function'` check as it's run only with the output of the conversion function.

I would refactor in this way:

**1) Supply a new type for run input**
```typescript
export type RunInput = Command | Commands | Command[] | string | undefined;
```

**2) Separate validation and transformation into a new function**
```typescript
private validate(input: RunInput, options: Record<string, unknown> = {}): ComplexCommand[] {
  if (!input) {
    console.warn('No command to run');
    return [];
  }

  // convert commands
  const converted: ComplexCommand[] = CommandsManager.convertCommands(input);
  if (!converted.length) {
    console.warn('Command is not runnable', input);
    return [];
  }

  return converted.map(command => ({
    commandName: command.commandName,
    commandOptions: { ...options, ...command.commandOptions },
    context: command.context,
  }));
}
```

**3) Two clearly typed and different functions for running commands**
```typescript
// Executes the commands synchronously in a blocking manner
// and returns an array of results.
public run(input: RunInput, options: Record<string, unknown> = {}): unknown[] {
  const commands = this.validate(input, options);

  const results: unknown[] = [];
  for (const command of commands) {
    const { commandName, commandOptions, context } = command;
    results.push(this.runCommand(commandName, commandOptions, context));
  }

  return results;
}

// Executes the commands sequentially in an asynchronous, blocking manner
// and returns a promise that resolves to an array of results.
public async runAsync(
  input: RunInput,
  options: Record<string, unknown> = {}
): Promise<unknown[]> {
  const commands = this.validate(input, options);

  const results: unknown[] = [];
  for (const command of commands) {
    const { commandName, commandOptions, context } = command;
    results.push(await this.runCommand(commandName, commandOptions, context));
  }

  return results;
}
```
I also refactored them to return an array of `unknown` results, since currently the result variable is overwritten with every command.

extensions/default/src/Panels/StudyBrowser/PanelStudyBrowser.tsx
```diff
@@ -7,6 +7,14 @@ import { useNavigate } from 'react-router-dom';
 import { Separator } from '@ohif/ui-next';
 import { PanelStudyBrowserHeader } from './PanelStudyBrowserHeader';
 import { defaultActionIcons, defaultViewPresets } from './constants';
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuTrigger,
+  Icons,
+  Button,
+} from '@ohif/ui-next';
```
salkz, line 16: Unused imports.

sedghi: I think your version is behind, these are removed

platform/core/src/classes/CommandsManager.ts
sedghi: Applied thanks

sedghi (approved):
Thanks a lot

PR #4182, opened 2024-05-29
feat: Add CS3D WSI and Video Viewports and add annotation navigation for MPR

wayfarer3130:
### Context

Video and WSI images now have a video viewport.  This PR adds the ability to display video and wsi using the CS3D viewports instead of the OHIF ones, allowing things like annotation and segmentations to be shared.

### Changes & Results

Use video viewport for basic-test mode

- pan/zoom tools for video
- tracked annotation tools for video
- "CINE" playback using video play, with rate being relative to recording rate.
- demographics overlay

Use WSI viewport for basic-test mode

- pan/zoom tools
- needs WSI viewport loading PR to actually load images (when this is linked in/merged, the WSI testing can proceed)

Switch measurement navigation to use view references

Add tracked changes for wsi and video viewports.

MPR viewports now support annotation navigation (using setViewReference).

### Testing

Against the ohif back end, use the studies:
Horse - for video
Any SM study for microscopy
Any CT study for volume navigation

Check that video and microscopy viewports can have annotations added.
Check that on the horse study, you can add an annotation to the stack and to the video viewport, and that clicking on them in the tracked changes navigates to the specified display set and image position.
Check that on the horse study you can play video using the CINE controls
Check that on video and microscopy you can zoom/pan
Check that you can window level on all 4 viewport types

Test MPR navigation with:
1. Turn on MPR (basic-test mode is easiest as you can hard turn off MPR with a button)
2. Add an annotation to all three viewports
3. Turn off MPR
4. Click on any of the annotations and have the single viewport navigate to the specific view in the MPR AND show the correct annotation

Note how the volume viewport gets recreated if MPR is off, however it does NOT show the 3x1 MPR hanging protocol but merely shows the single viewport.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone/src/getSopClassHandlerModule.js
```diff
@@ -102,13 +108,43 @@ function _getDisplaySetsFromSeries(instances, servicesManager, extensionManager)
...
+  const instanceMap = new Map();
+  instances.forEach(instance => instanceMap.set(instance.imageId, instance));
+  if (dicomWebClient) {
+    // This is really ugly - maybe this should be stored unmodified.
+    dicomWebClient.getDICOMwebMetadata = imageId => transferDenaturalizedDataset(denaturalizeDataset(fixMultivalueKeys(instanceMap.get(imageId))));
+
+    csUtils.genericMetadataProvider.add(displaySet.imageIds[0], {
+      type: 'webClient',
+      rawMetadata: dicomWebClient,
+    });
```
sedghi, line 128: This is a bit unusual, but I somewhat understand. Do you need to pass the DICOM web client along in the other places?

maybe have the type as `dicomWebClient` then, not webClient

sedghi: I think we need to stick to the convention and call the payload `metadata` as well

extensions/cornerstone/src/getSopClassHandlerModule.js
```diff
@@ -102,13 +108,43 @@ function _getDisplaySetsFromSeries(instances, servicesManager, extensionManager)
...
-export default function getDicomMicroscopySopClassHandler({ servicesManager, extensionManager }) {
+/** Fix multi-valued keys before denaturalizing */
+export function fixMultivalueKeys(naturalData, keys = ['ImageType']) {
+  for (const key of keys) {
+    if (typeof naturalData[key] === 'string') {
+      naturalData[key] = naturalData[key].split('\\');
+    }
+  }
+  return naturalData;
+}
```
sedghi, line 145: Does it make sense to put it in the dicomUtils? `import { dicomWebUtils } from '@ohif/extension-default';`

extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
```diff
@@ -143,6 +149,25 @@ class CornerstoneCacheService {
...
+    dataSource,
+    displaySets,
+    _initialImageIndex,
+    viewportType: Enums.ViewportType
+  ): Promise<StackViewportData> {
+    // TODO - handle overlays and secondary display sets, but for now assume
+    // the 1st display set is the one of interest
+    const [displaySet] = displaySets;
+    if (!displaySet.imageIds) {
+      displaySet.imagesIds = this._getCornerstoneStackImageIds(displaySet, dataSource);
```
sedghi, line 162: I dislike this approach, though I likely added it. We can simplify it to:

```javascript
displaySet.imageIds = dataSource.getImageIdsForDisplaySet(displaySet);
```

platform/core/src/utils/sortStudy.ts
```diff
@@ -26,7 +26,7 @@ function seriesInfoSortingCriteria(firstSeries, secondSeries) {
   const bLowPriority = isLowPriorityModality(secondSeries.Modality ?? secondSeries.modality);
 
   if (aLowPriority) {
-    return bLowPriority ? defaultSeriesSort(secondSeries, firstSeries) : 1;
+    return bLowPriority ? -defaultSeriesSort(secondSeries, firstSeries) : 1;
```
sedghi, line 29: can you add a comment why

platform/core/src/services/DisplaySetService/IDisplaySet.ts
```diff
@@ -1,9 +1,12 @@
-interface IDisplaySet {
+type IDisplaySet {
```
sedghi, line 1: If this is a type, we probably don't need to add the "I" prefix to "DisplaySet".

extensions/dicom-microscopy/src/DicomMicroscopyViewport.tsx
sedghi, line 1: Do we still need any of this? Or even the entire dicom-microscopy extension? I guess not.

extensions/default/src/DicomWebDataSource/index.js
```diff
@@ -190,6 +190,13 @@ function createDicomWebApi(dicomWebConfig, servicesManager) {
           params
         );
       },
+      /**
+       * Provide direct access to the dicom web client for certain use cases
+       * where the dicom web client is used by an external library such as the
+       * microscopy viewer.
+       */
+      getWadoDicomWebClient: () => wadoDicomWebClient,
```
sedghi, line 198: getDicomWebClientInstance?

extensions/default/src/DicomWebDataSource/utils/cleanDenaturalizedDataset.ts
sedghi, line 1: some notes on this file

1. vrNumerics can be a Set for performance with `vrNumerics.has()`


2. Multiple return statements for readiblity


```js
if (Array.isArray(obj)) {
  return obj.map(o => isPrimitive(o) ? o : cleanDenaturalizedDataset(o, options));
}

if (isPrimitive(obj)) {
  return obj;
}
```

3. One line json stringify and parse so we don't assign

```js
JSON.parse(JSON.stringify(cleanDenaturalizedDataset(dataset)));
```

extensions/cornerstone/src/utils/getCornerstoneViewportType.ts
```diff
@@ -4,13 +4,26 @@ const STACK = 'stack';
...
+  const lowerViewportType =
+    displaySets?.[0]?.viewportType?.toLowerCase() || viewportType.toLowerCase();
   if (lowerViewportType === STACK) {
     return Enums.ViewportType.STACK;
   }
 
+  if (lowerViewportType === VIDEO) {
+    return Enums.ViewportType.VIDEO;
+  }
+  if (lowerViewportType === WholeSlide) {
```
sedghi, line 23: can we follow the convention? WHOLESLIDE

platform/core/src/services/DisplaySetService/DisplaySetService.ts
```diff
@@ -131,7 +131,7 @@ export default class DisplaySetService extends PubSubService {
       : [...this.getDisplaySetCache().values()];
 
     const displaySet = displaySets.find(ds => {
-      return ds.images && ds.images.some(i => i.SOPInstanceUID === sopInstanceUID);
+      return ds.instances?.some(i => i.SOPInstanceUID === sopInstanceUID);
```
sedghi, line 134: Are we just utilizing the instances instead of images, or are we deleting images too? If so, we need to add this information to the breaking change page.

extensions/dicom-video/src/getSopClassHandlerModule.js
```diff
@@ -72,20 +73,29 @@ const _getDisplaySetsFromSeries = (instances, servicesManager, extensionManager)
...
         thumbnailSrc: dataSource.retrieve.directURL({
           instance,
           defaultPath: '/thumbnail',
           defaultType: 'image/jpeg',
           tag: 'Absent',
         }),
+        imageIds: [imageId],
         isDerivedDisplaySet: true,
         isLoaded: false,
+        isClip: true,
```
sedghi, line 90: The only other place we use `isClip` is in multiframe. I'm wondering if it's correct. Perhaps a comment in the code would clarify its purpose.

extensions/default/src/getSopClassHandlerModule.js
```diff
@@ -1,12 +1,12 @@
...
-import isDisplaySetReconstructable from '@ohif/core/src/utils/isDisplaySetReconstructable';
+import { utils, classes } from '@ohif/core';
 import { id } from './id';
 import getDisplaySetMessages from './getDisplaySetMessages';
 import getDisplaySetsFromUnsupportedSeries from './getDisplaySetsFromUnsupportedSeries';
 import { chartHandler } from './SOPClassHandlers/chartSOPClassHandler';
 
+const { isImage, sopClassDictionary, isDisplaySetReconstructable } = utils;
+const { ImageSet } = classes;
+
```
sedghi, line 9: Thanks

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -428,7 +428,10 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
     // override the viewportOptions and displaySetOptions with the public ones
     // since those are the newly set ones, we set them here so that it handles defaults
     const displaySetOptions = viewportInfo.setPublicDisplaySetOptions(publicDisplaySetOptions);
-    const viewportOptions = viewportInfo.setPublicViewportOptions(publicViewportOptions);
+    const viewportOptions = viewportInfo.setPublicViewportOptions(
+      publicViewportOptions,
+      viewportData.viewportType
+    );
```
sedghi, line 434: Doesn't `publicViewportOptions` already have a viewportType?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -526,27 +529,57 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
-  public getViewportIdToJump(
-    activeViewportId: string,
-    displaySetInstanceUID: string,
-    cameraProps: unknown
-  ): string {
-    const viewportInfo = this.getViewportInfo(activeViewportId);
-
-    if (viewportInfo.getViewportType() === csEnums.ViewportType.VOLUME_3D) {
-      return null;
+  public getViewportIdToJump(activeViewportId: string, metadata): string {
```
sedghi, line 532: Could you please add extensive documentation to the function about the order in which we jump? It seems there are multiple orders and priorities that could benefit from a detailed explanation.

sedghi (commented):
Great additions indeed! I haven't tried them yet, but I've completed my first pass for the code review. Please see my comments.

extensions/dicom-microscopy/src/DicomMicroscopyViewport.tsx
wayfarer3130: Yes, at this point we need the extension.  There are things not yet implemented in the new viewport so that it isn't quite ready yet to replace the old one - such as the DICOM SR loading for the old style, certain types of microscopy images and other things that I don't know about.

extensions/default/src/DicomWebDataSource/index.js
wayfarer3130: I added commentary:

wayfarer3130: The issue is that it isn't necessarily a full dicom web client instance, but is specific to the WADO part (retrieve) as some back ends separate out the various services for different reasons.

platform/core/src/services/DisplaySetService/DisplaySetService.ts
wayfarer3130: Just utilizing instances instead of images because video display sets don't have the images, whereas instances is the canonical name for any instance.

extensions/dicom-video/src/getSopClassHandlerModule.js
wayfarer3130: Actually we don't reference it anywhere in the code, so just deleting it.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
wayfarer3130: Added the comment:

platform/app/src/components/ViewportGrid.tsx
```diff
@@ -183,7 +183,9 @@ function ViewerViewportGrid(props: withAppTypes) {
         viewport.viewportOptions.orientation = 'acquisition';
 
         const displaySet = displaySetService.getDisplaySetByUID(referencedDisplaySetInstanceUID);
+        viewport.viewportOptions.viewReference = measurement.metadata;
```
sedghi, line 186: can you please add a comment why we are doing this here?

platform/core/src/services/DisplaySetService/DisplaySet.ts
```diff
@@ -1,9 +1,12 @@
-interface IDisplaySet {
+type DisplaySet {
```
sedghi, line 1: should be added to the renames of the migraation guides

extensions/cornerstone/src/getSopClassHandlerModule.js
```diff
@@ -83,6 +89,7 @@ function _getDisplaySetsFromSeries(instances, servicesManager, extensionManager)
   const displaySet = {
     plugin: 'microscopy',
     Modality: 'SM',
+    viewportType: 'WholeSlide',
```
sedghi, line 92: is this correct? or should be wholeslide

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -2,24 +2,17 @@ import React, { useEffect, useRef, useCallback, useState } from 'react';
...
-  eventTarget,
-  getEnabledElement,
-  StackViewport,
-  utilities as csUtils,
-} from '@cornerstonejs/core';
+import { Enums, eventTarget, getEnabledElement } from '@cornerstonejs/core';
 import { MeasurementService } from '@ohif/core';
 import { Notification, useViewportDialog, AllInOneMenu } from '@ohif/ui';
-import { IStackViewport, IVolumeViewport } from '@cornerstonejs/core/dist/esm/types';
+import type { Types as CSTypes } from '@cornerstonejs/core';
```
sedghi, line 8: I think our convension is ohifTypes, csTypes etc

sedghi (commented):
This used to show microscopy but i'm not sure after this change we need a mode for microscopy even
Screenshot 2024-07-12 at 3 28 48 PM

---

i get an error opening a SM saying `Cannot read properties of undefined (reading 'ImageType')` can you try loading this 

http://localhost:3000/viewer?StudyInstanceUIDs=2.25.141277760791347900862109212450152067508

---

also the horse does not show either


http://localhost:3000/viewer?StudyInstanceUIDs=2.25.96975534054447904995905761963464388233

maybe it is the peerImport stuff?

wayfarer3130:
@sedghi - I messed up the study navigation for measurements and then returning to the study.  Let me fix that and get back to you on this PR.  There is already the presentation information, which I want to use for the view reference rather than modifying the viewport info.

wayfarer3130:
Fixed your issues and the bug I found.  There is still a remaining bug in the underlying CS3D viewport stuff where the isReferenceViewable isn't quite right, but Igor is looking at some of that, so I'd like to proceed with this PR.  The bug only affects navigation to/from same frame of reference stack only series when creating volume MPR measurements, so I think it is low enough probability that we can continue for the moment until that gets fixed properly.
The horse study shows fine, while the other reference you gave me isn't in the test dataset, so I don't expect that one to work.  However, the SM data in the test dataset works fine.

sedghi:
I still can't get the viewports showing anything

sedghi (approved):
Thanks, this is really great

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -586,47 +525,15 @@ function _jumpToMeasurement(
 
   if (enabledElement) {
     // See how the jumpToSlice() of Cornerstone3D deals with imageIdx param.
-    const viewport = enabledElement.viewport as IStackViewport | IVolumeViewport;
```
IbrahimCSAE, line 589: jump to measurements for SR is broken because of this, what should we do to fix it?

Issue #3884, opened 2024-01-08
[Bug] Incorrect slice after hydration of SR and SEG


sedghi:
### Describe the Bug

For SR: https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7695.4007.324475281161490036195179843543

For SEG: https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.256467663913010332776401703474716742458

https://github.com/OHIF/Viewers/assets/7490180/6f724350-a929-4803-84dd-fed768c3939a

### Steps to Reproduce

Links above	

### The current behavior

It jumps to start

### The expected behavior

Should keep the camera

### OS

macos

### Node version

18

### Browser

chrome

sedghi:
We released [OHIF 3.9](https://github.com/OHIF/Viewers/tree/release/3.9) which address this issue

PR #4200, opened 2024-05-31
fix(hydration): Maintain the same slice that the user was on pre hydration in post hydration for SR and SEG.

IbrahimCSAE:
### Context

Maintain the same slice that the user was on pre hydration in post hydration for SR and SEG.

Fixes https://github.com/OHIF/Viewers/issues/3884


extensions/default/src/getHangingProtocolModule.js
```diff
@@ -120,10 +120,10 @@ function getHangingProtocolModule() {
...
       protocol: hpMammography,
     },
-    // {
-    //   name: hpScale.id,
-    //   protocol: hpScale,
-    // },
+    {
+      name: hpScale.id,
+      protocol: hpScale,
+    },
```
sedghi, line 126: revert this for now since when you go to segmenttaion mode this gets matched for some reason

extensions/cornerstone-dicom-seg/src/utils/hydrationUtils.ts
```diff
@@ -61,6 +64,14 @@ async function updateViewportsForSegmentationRendering({
     };
     const viewportId = viewport.viewportId;
 
+    // maintain the prehydration slice on the target viewport only
+    if (viewportId === targetViewportId) {
+      viewport.viewportOptions.initialImageOptions = {
+        index: initialSliceIndex,
+        useOnce: true,
```
sedghi, line 71: Why do we need `useOnce`? I don't prefer this approach.

extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/TrackedMeasurementsContext.tsx
```diff
@@ -81,6 +86,29 @@ function TrackedMeasurementsContextProvider(
         },
       });
     },
+
+    jumpToSameImageInActiveViewport: (ctx, evt) => {
```
sedghi, line 90: Will this be the same logic if there is only one measurement? What I'm asking is, is `jumpToSameImage = jumpToFirstMeasurement` if there is only one measurement?

sedghi (commented):
see my comments

extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/TrackedMeasurementsContext.tsx
IbrahimCSAE: It doesn’t matter how many measurements there are, it will still be the same slice the user was on

extensions/cornerstone-dicom-seg/src/utils/hydrationUtils.ts
IbrahimCSAE: So when the seg is hydrated, the initial image index is set, but lets say the user switches to another series, then drops the referenced series for the seg again, we should not use the initial index again, we should default to first/middle or last.

extensions/default/src/getHangingProtocolModule.js
IbrahimCSAE: Sure

sedghi (approved):
Thanks

Issue #4021, opened 2024-04-10
[Bug] Information lost when downloading / exporting DICOM SEG series


nkaenzig:
### Describe the Bug

When downloading a DICOM SEG series using OHIF's download functionality, it doesn't seem to download the original series (DICOM file), but create a copy with a new SeriesInstanceUID. In addition, in the downloaded DICOM file, `SegmentAlgorithmType` is set to `MANUAL`, and additional tags such as `AlgorithmVersion` and `SegmentAlgorithmName` are lost.

A very similar behaviour occurs with the "Export DICOM SEG" functionality. While there, it makes sense to create a new series (with a new SeriesInstanceUID), additional tags such as `AlgorithmVersion` and `SegmentAlgorithmName` are also discarded. Also, there seems to be no way to backtrack exported DICOM SEG series to the series they originate from.

### Steps to Reproduce

1. Open any CT/SEG series (e.g. Stagel-Colorectal-Cancer) on https://viewer-dev.ohif.org/ using the Segmentation Viewer
2. Open the Segmentation in the viewer 
3. In right side panel click "Download DICOM SEG"
  <img width="507" alt="image" src="https://github.com/OHIF/Viewers/assets/36882833/5937f41f-21b7-4955-a8bb-c3ec66e9ba2f">

4. Compare the `SeriesInstanceUID` in the downloaded DICOM file with the DICOM SEG series opened in OHIF

The behaviour described for "Export DICOM Seg" can be reproduced in a similar way, but must be done on a deployment with write access.

### The current behavior

See description above.

### The expected behavior

From a user perspective, I'd expect the "Download DICOM SEG" functionality to download the original DICOM file without modifications and "Export DICOM SEG" to create a complete copy of the DICOM SEG series without omitting tags such as `AlgorithmVersion` and `SegmentAlgorithmName`.

### OS

OS X 10.15.7

### Node version

Whatever https://viewer-dev.ohif.org/ uses

### Browser

Chrome 123.0.0

### OHIF version

3.8.0-beta.73

sedghi:
We released [OHIF 3.9](https://github.com/OHIF/Viewers/tree/release/3.9) which address this issue

PR #4203, opened 2024-06-04
fix(seg): maintain algorithm name and algorithm type when DICOM seg is exported or downloaded

IbrahimCSAE:
### Context

This will maintain the AlgorithmName and AlgorithmType when DICOM SEG is exported, it also makes the modality selector searchable, and fixes the study date selector height to match the rest in the study list.

Fixes the https://github.com/OHIF/Viewers/issues/4021

Exported SEG: 

image


platform/ui/src/components/InputMultiSelect/InputMultiSelect.tsx
```diff
@@ -30,7 +30,7 @@ const InputMultiSelect = ({
         value={value}
         isMulti={true}
         isClearable={false}
-        isSearchable={false}
+        isSearchable={true}
```
sedghi, line 33: I think our UI for having searchable is aweful, do you have screenshot to show before and after this change? we should move to the comboBox in the shadcn

sedghi (commented):
see my comment

platform/ui/src/components/InputMultiSelect/InputMultiSelect.tsx
IbrahimCSAE: It doesn’t look bad to me, I formatted it well, I will post screenshots tomorrow

sedghi (approved):
Thanks

PR #4206, opened 2024-06-06
feat(studyPrefetcher): Study Prefetcher

lscoder:
Added support for a Study Prefetcher responsible for loading the next display sets. The following settings are available:

```javascript
  studyPrefetcher: {
    /* Enable/disable study prefetching service (default: false) */
    enabled: true,
    /* Number of displaysets to be prefetched  (default: 2)*/
    displaySetCount: 2,
    /**
     * Max number of concurrent prefetch requests (default: 10)
     * High numbers may impact on the time to load a new dropped series because
     * the browser will be busy with all prefetching requests. As soon as the
     * prefetch requests get fulfilled the new ones from the new dropped series
     * are sent to the server.
     *
     * TODO: abort all prefetch requests when a new series is loaded on a viewport.
     * (need to add support for `AbortController` on Cornerstone)
     * */
    maxNumPrefetchRequests: 10,
    /* Display sets loading order (closest (deafult), downward or upward) */
    order: 'closest',
  },
```

### Testing

- Disable the StudyPrefetcher and make sure it does not prefetch the next display sets (1).
- Enable the StudyPrefetcher and make sure the next display sets are loaded (1) based on the configuration provided.
- Load a new series, change the viewport layout (eg: 1x1 to 1x2), change the active viewport and see if the displaysets are loaded as expected based on the display set loaded on the active viewport (eg: making viewport 2 as active should stop loading the series based on viewport 1 and start loading the ones related to viewport 2).

(1) The progress is displayed while a series is being prefetched underneath its thumbnail (eg: "P: 47%") but **the design will change soon** (the text will be replaced by a progress bar).


platform/core/src/services/StudyPrefetcherService/StudyPrefetcherService.ts
```diff
@@ -0,0 +1,582 @@
...
+    Object.assign(this.config, configuration);
+    // this._displaySets = [];
+
+    // DEBUG - DELETE
+    setTimeout(() => this.start(), 1000);
+  }
+
+  // Move this method to a different place (component)
+  private start() {
+    const { displaySetService, viewportGridService } = this._servicesManager.services;
```
lscoder, line 106: @sedghi  Where would be a better place to add some listeners? (UI <-- listeners --> StudyPrefetcher)

platform/ui/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -94,6 +95,12 @@ const Thumbnail = ({
             />
             {` ${numInstances}`}
           </div>
+          {loadingProgress && loadingProgress < 1 && (
+            <div className="mr-4">
+              <span className="text-primary-main font-bold">{'P: '}</span>
```
sedghi, line 100: Let's just put the percentage, no `P`, since `S:` here means SeriesNumber. I'm afraid it would confuse people on what `P` is.

platform/core/src/services/ViewportGridService/ViewportGridService.ts
```diff
@@ -105,8 +105,8 @@ class ViewportGridService extends PubSubService {
     this.setDisplaySetsForViewports([props]);
   }
 
-  public setDisplaySetsForViewports(props) {
-    this.serviceImplementation._setDisplaySetsForViewports(props);
+  public async setDisplaySetsForViewports(props) {
+    await this.serviceImplementation._setDisplaySetsForViewports(props);
     const state = this.getState();
     const viewports = [];
```
sedghi, line 112: I didn't expect this async method here, why?

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.tsx
```diff
@@ -459,6 +489,7 @@ function _mapDataSourceStudies(studies) {
 
 function _mapDisplaySets(
   displaySets,
+  displaySetLoadingState,
```
sedghi, line 492: can you just pass in the loadingProgress value without the whole state?

platform/core/src/services/StudyPrefetcherService/StudyPrefetcherService.ts
```diff
@@ -0,0 +1,495 @@
...
+   * (need to add support for `AbortController` on Cornerstone)
+   * */
+  maxNumPrefetchRequests: number;
+  /* Display sets prefetching order (closest, downward and upward) */
+  order: StudyPrefetchOrder;
+};
+
+type DisplaySetStatus = {
+  displaySetInstanceUID: string;
+  numImages: number;
```
sedghi, line 51: displaySet does not always contain images, maybe numInstances?

sedghi (commented):
Looks very clean and nicely done. My only comment is that I see in the UI the progress shows up to 100% and then disappears. Perhaps add a database icon after completion.

https://lucide.dev/icons/database

@dan-rukas

platform/core/src/services/ViewportGridService/ViewportGridService.ts
lscoder: That is a `dispatch` method from `ViewportGridProvider` that runs asynchronously and without `await` this method was returning the old state (w/ old `displaySetInstanceUID`) instead of the updated one.

sedghi (changes requested):
Seems like the current display set does not have the cache icon, also can we make the icon aligned to the right see the misalignment below

CleanShot 2024-06-24 at 14 17 27@2x


CleanShot 2024-06-24 at 14 18 25@2x

platform/core/src/services/StudyPrefetcherService/StudyPrefetcherService.ts
```diff
@@ -0,0 +1,495 @@
+import { cache, imageLoadPoolManager, imageLoader, Enums } from '@cornerstonejs/core';
```
sedghi, line 1: This shouldn't be here, if the study prefetcher relies on cornerstone, the service should be inside the cornerstone extension. Ideally nothing inside platform/* should rely on cornerstone

lscoder:
> Seems like the current display set does not have the cache icon, also can we make the icon aligned to the right see the misalignment below
> 

Updated

platform/core/src/services/StudyPrefetcherService/StudyPrefetcherService.ts
lscoder: Updated

sedghi (approved):
Looks great, thanks!

extensions/default/src/utils/createRenderedRetrieve.js
```diff
@@ -17,12 +17,13 @@ const createRenderedRetrieve = (config, params) => {
   const { wadoRoot } = config;
   const { instance, tag = 'PixelData' } = params;
   const { StudyInstanceUID, SeriesInstanceUID, SOPInstanceUID } = instance;
-  const value = instance[tag];
+  const bulkDataURI = instance[tag]?.BulkDataURI ?? '';
```
sedghi, line 20: FYI @wayfarer3130

Issue #4083, opened 2024-05-04
[Feature] Implement custom sorting of series in study panel


sedghi:
@IbrahimCSAE Updated the descriptions on this, thanks !

### Context

The goal of this epic is to enhance the OHIF Viewer by implementing a custom sorting feature for series within the study panel. This feature will improve the user experience by allowing users to sort series based on different criteria, such as acquisition datetime, series number, or any other relevant metadata.

### Background

The current implementation of the OHIF Viewer does not support flexible sorting of series in the study panel. Users have requested the ability to sort series according to specific needs, which is essential for efficient navigation and review of medical images. This feature has been highlighted in various issues and discussions within the OHIF community.

### Related Issues and Discussions

- https://github.com/OHIF/Viewers/issues/4004
- https://github.com/OHIF/Viewers/issues/4076
- https://github.com/OHIF/Viewers/issues/3959
- https://github.com/OHIF/Viewers/issues/3933
- https://github.com/OHIF/Viewers/issues/4083
- https://github.com/OHIF/Viewers/pull/2285
- Community Discussion - [Sort instances/study by acquisition datetime](https://community.ohif.org/t/sort-instances-study-by-acquisition-datetime/1299/8)

### Objectives

Custom Sorting Implementation: Allow users to sort series based on various criteria.

- AcquisitionDate
- Series Number
- Modality
- Any other relevant DICOM metadata

User Interface Enhancements:

- Update the study panel UI to include sorting options.
- Provide an intuitive and user-friendly sorting interface.

Allow users to override the initially used default sort

- Create a customization module to replace the default sort if available on study panel load

niirdan:
@sedghi thanks for gathering all information , unfortunately all of them are unsolved / answered 
 currently I'm trying to implement a custom function which solve this topic, hope to hear from someone which already deal with this subject

niirdan:
UPDATE : 
`function getDisplaySetsFromSeries(instances) {
  // If the series has no instances, stop here
  if (!instances || !instances.length) {
    throw new Error('No instances were provided');
  }

  var displaySets = [];
  const sopClassUids = getSopClassUids(instances);

  // Search through the instances (InstanceMetadata object) of this series
  // Split Multi-frame instances and Single-image modalities
  // into their own specific display sets. Place the rest of each
  // series into another display set.
  const stackableInstances = [];
  instances.forEach(instance => {
    // All imaging modalities must have a valid value for sopClassUid (x00080016) or rows (x00280010)
[...]

Issue #4076, opened 2024-05-02
[Feature Request] sorting by Acquisition Datetime and nonClip display


niirdan:
What feature or change would you like to see made?

### sort by dicom tag (0008,002A)
In addition , disable the behavior of the system which creates a stack of images and display them each by it own.

Why should we prioritize this feature?

**This Option Will make OHIF Very useful for cardiologist's and will make much more sense when investigating studies**

sedghi:
We added many sorting features to the UI.

Can you check the latest OHIF viewer, which now includes Cornerstone 3D 2.0?

Here are the migration guides:

Try OHIF locally: [https://viewer-dev.ohif.org/localbasic](https://viewer-dev.ohif.org/localbasic)<br>Try Cornerstone3D 2.0 locally: [https://www.cornerstonejs.org/live-examples/local.html](https://www.cornerstonejs.org/live-examples/local.html)

[https://docs.ohif.org/migration-guide/3p8-to-3p9/](https://docs.ohif.org/migration-guide/3p8-to-3p9/)<br><br>----<br><br>Also see how to configure it<br><br>[https://docs.ohif.org/faq/technical/#how-can-i-change-the-sorting-of-the-thumbnail--study-panel--study-browser](https://docs.ohif.org/faq/technical/#how-can-i-change-the-sorting-of-the-thumbnail--study-panel--study-browser)

Issue #4004, opened 2024-03-25
[Feature Request] sorting by SeriesTime and SeriesDate


niirdan:
### What feature or change would you like to see made?

Hello Everyone , 
I am working with cardiologist's and their way of interpreting exams is by the exam's **sequence**.
they capture each phase of the exam (stressEcho for example) rest phase , exercise phase , and recovery phase
The Way ohif works now (sorting) is very un useful for doctors to use it ,simply because sorting isn't managed correctly for their purposes . 
I was wondering why a sorting selection does not exist - which is a really important tool for Doctors/Technician's .. 

any plans regarding that In Feature? 
I am 100% this will be useful widely across users 
Many Thanks in advance!

### Why should we prioritize this feature?

Cardiologist's cant really use OHIF , due to sorting .

Issue #3959, opened 2024-02-26
[Bug] sort by function


niirdan:
### Describe the Bug

Good afternoon ,
i tried to follow this : https://docs.ohif.org/platform/extensions/modules/sop-class-handler
implementing 'order by' acquisition DateTime. 
unfortunately sortBy function as far as i understand can handle integers only , (instanceNumber) .. 
I am trying to view Series within a study to be ordered by their acquisition DateTime .
I saw that images are converged into one series , there is a way to disable this behavior? 

### Steps to Reproduce

extensions > src > getHangingProtocolModule.js
extensions > default > src> getSopClassHandlerModule.js

### The current behavior

sorting by is currently by instanceNumber

### The expected behavior

Order a study (exam) sorted by acquisition dateTime attribute   

### OS

window 11

### Node version

v18.17.1

### Browser

latest chrome update

PR #4214, opened 2024-06-07
feat(sort): custom series sort in study panel

IbrahimCSAE:
### Context

Allows defining custom displaySet sort in the customizationModule to be used for sorting the displaySets.

We are currently redesigning the study panel and the study browser. During this process, you can enable our undesigned component via the `experimentalStudyBrowserSort` flag. This will look like:

CleanShot 2024-06-20 at 16 09 57@2x

You can  also add your own sorting functions by utilizing the `customizationService` and adding the `studyBrowser.sortFunctions` key, as shown below:

```
customizationService.addModeCustomizations([
  {
    id: 'studyBrowser.sortFunctions',
    values: [{
      label: 'Series Images',
      sortFunction: (a, b) => {
        return a?.numImageFrames - b?.numImageFrames;
      },
    }],
  },
]);
```

### Related Issues and Discussions

- https://github.com/OHIF/Viewers/issues/4004
- https://github.com/OHIF/Viewers/issues/4076
- https://github.com/OHIF/Viewers/issues/3959
- https://github.com/OHIF/Viewers/issues/3933
- https://github.com/OHIF/Viewers/issues/4083
- https://github.com/OHIF/Viewers/pull/2285
- Community Discussion - [Sort instances/study by acquisition datetime](https://community.ohif.org/t/sort-instances-study-by-acquisition-datetime/1299/8)


sedghi (commented):
The logic seems sound. 

However, we should avoid surprising users by introducing an unfinished UI feature. Therefore, please put it behind a flag that is disabled by default, perhaps named `experimentalStudyBrowserSort` or something similar. (we can just put it in default.js and netlify.js maybe) and add in the PR description and talk about it

Additionally, we need to include documentation in the FAQ explaining how users can add their custom sort functions. I'm tried of  repeatedly addressing such question, and i want to link to it

Overall great job

sedghi:
I'm having second thoughts actually.

This should work at display set service level and not at the UI level. So upon sorting it should re-order displaysets and fire modified.

Panel study browser should not have a single line of change

IbrahimCSAE:
> I'm having second thoughts actually.
> 
> This should work at display set service level and not at the UI level. So upon sorting it should re-order displaysets and fire modified.
> 
> Panel study browser should not have a single line of change

sounds good, let me sort that out monday

lscoder:
Should this file should be updated?

https://github.com/OHIF/Viewers/blob/master/extensions/default/src/commandsModule.ts#L508-L511

IbrahimCSAE:
> Should this file should be updated?
> 
> https://github.com/OHIF/Viewers/blob/master/extensions/default/src/commandsModule.ts#L508-L511

good catch

extensions/default/src/getCustomizationModule.tsx
```diff
@@ -162,6 +165,26 @@ export default function getCustomizationModule({ servicesManager, extensionManag
...
+            {
+              label: 'Series Number',
+              sortFunction: (a, b) => {
+                return a?.SeriesNumber - b?.SeriesNumber;
+              },
+            },
+            {
+              label: 'Series Date',
+              sortFunction: (a, b) => {
+                const dateA = new Date(formatDate(a?.SeriesDate));
```
wayfarer3130, line 181: If you are going to sort on series date, it HAS to include series time as well - there are cases where doing it wrong can result in patient harm.

modes/longitudinal/src/index.ts
```diff
@@ -90,6 +90,18 @@ function modeFactory({ modeConfiguration }) {
       //   },
       // ]);
 
+      customizationService.addModeCustomizations([
```
wayfarer3130, line 93: Example customizations should be included in the basic test mode, not in longitudinal.

sedghi: I was just trying to see if it appends, but it did not

sedghi (approved):
Great

niirdan:
good afternoon, the feature has been merged to master already?
 Is it available by adding a parameter in conf file?

sedghi:
@niirdan Yes, take a look https://docs.ohif.org/faq/technical#how-can-i-change-the-sorting-of-the-thumbnail--study-panel--study-browser

niirdan:
Thanks for the update!

‫בתאריך יום ד׳, 3 ביולי 2024 ב-15:56 מאת ‪Alireza‬‏ <‪
***@***.***‬‏>:‬

niirdan:
I notice that sorting takes into consideration only cines , for example if
a study contains both images and cines it will sort only the cines and the
images will remain stacked together .
As a suggestion , In order to make this feature useful for cardiologists
they are looking for the sequence(acquisition date&time) of the study
material during their diagnosis .
To implement that , the stacking behaviour in this case is not the right
Idea.


‫בתאריך יום ד׳, 3 ביולי 2024 ב-16:00 מאת ‪Dan Nir‬‏ ***@***.***‬‏>:‬

sedghi:
@niirdan Could you please elaborate on this? What do you mean by "stacking behavior"? Are you suggesting that the grouping is not done correctly? Which study are you using for your example? It would be helpful if you could use one of our studies from the demo server to explain the current behavior and the desired behavior.

niirdan:
@sedghi for sure!
Cardiologists diagnosis for example stress-echo exams by comparing different breathing cases (In stress/rest etc..)
they acquire during the exam both single instance Img's and cine loops .
during the diagnosis they want to see the Chronological order of both img's and cines they took from the US because the sequence has medical reasons which are Important for them .
This is why the stacking of single img's (In This Case) is not an Idea presentation way. 
I managed to disable the stacking behavior - looks like that
list-with-sepimg
In this image you can see both a cine and image's , Which are not in the chronological ordered . 
The Dicom attribute which define the chronological order is acquisition date time , (0008,002A)

Issue #4205, opened 2024-06-05
[Bug] async await.. SRs are not visible properly (I can create PR).


Joeycho:
### Describe the Bug

https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2403.334240657131972136850343327463

Click PT SR and then CT SR. It failed to visualize CT SR (the second one).

### Steps to Reproduce

1. Click PT SR
2. Click CT SR -> you will see no annotation there. Only image
3. click PT SR again
4. click CT SR again -> you will see annotation this time.

### The current behavior

https://github.com/OHIF/Viewers/blob/14221eb78ede2048392e352811a5f98a080c049f/extensions/cornerstone-dicom-sr/src/viewports/OHIFCornerstoneSRViewport.tsx#L266

https://github.com/OHIF/Viewers/blob/14221eb78ede2048392e352811a5f98a080c049f/extensions/cornerstone-dicom-sr/src/viewports/OHIFCornerstoneSRViewport.tsx#L280

While waiting load() (L266), the next useEffect() (L280).. can run. 

### The expected behavior

L266 should be finished before L280 code blocks. Therefore, srDisplaySet.isLoaded should be true.

### OS

mac OS 14.4

### Node version

16.18

### Browser

Chrome  125.0.6422.113

Joeycho:
https://github.com/OHIF/Viewers/pull/4226 @sedghi

sedghi:
I merged your PR, does it solve your issue?

Joeycho:
Yes it is :)

PR #4226, opened 2024-06-11
fix(sr): rendering issue by running loadSR before updateSR

Joeycho:
Fix the following issue.
https://github.com/OHIF/Viewers/issues/4205

loadSR should run at least one time, before updateSR is running.


sedghi (approved):
Thanks a lot!

PR #4235, opened 2024-06-16
feat(HP): Frame View HP

IbrahimCSAE:
### Context

Source of idea: https://santesoft.com/win/sante-dicom-viewer-lite/sante-dicom-viewer-lite.html

image

image

image

image


IbrahimCSAE:
needs an icon if we want this in the layout selector

sedghi (approved):
Thanks

deathknight0718:
@IbrahimCSAE protocolMatchingRules is undefined. so, When using segmentation mode, the following code generates an error.
```
  findMatchByStudy(study, options) {
    const matched = [];

    this.protocols.forEach(protocol => {
      // Clone the protocol's protocolMatchingRules array
      // We clone it so that we don't accidentally add the
      // numberOfPriorsReferenced rule to the Protocol itself.
      console.log(protocol);
      let rules = protocol.protocolMatchingRules.slice(); // FAILED
```

PR #4254, opened 2024-06-22
feat(studies-panel): New OHIF study panel - under experimental flag

IbrahimCSAE:
# New OHIF Studies Panel

This PR introduces a redesigned OHIF study panel as part of an ongoing effort to enhance the UI and support mobile devices. The panel now uses the `@ohif/ui-next` components built with `shadcn`.

## View presets

The new panel includes two view presets: a thumbnail view and a list view. By default, the thumbnail view is selected, but users can customize the default using the `customizationService`. Example customization:

```js

export default function getCustomizationModule({ servicesManager, extensionManager }) {
  return [
    {
      name: 'default',
      value: [
        {
          id: 'studyBrowser.viewPresets',
          // change your default selected preset here
          value: [
            {
              id: 'list',
              iconName: 'ListView',
              selected: true, // -> defaults to this on initial load
            },
            {
              id: 'thumbnails',
              iconName: 'ThumbnailView',
              selected: false,
            },
          ],
        },
      ],
    },
  ];
}

```

Users can also customize icons and create new presets, although adding custom render functions is not yet available via the `customizationService`, if you want to add a custom render function you can do so in the `Thumbnail.tsx` file.

### List and Thumbnail Presets

<p align="center">

</p>
<p align="center">The panel enables users to switch between list and thumbnail views to match their preference.</p>

## Panel options

The settings menu allows users to filter and sort studies. A dropdown lets users choose which studies to display (primary, recent, all) and sort them by series number, date, etc. Sorting is customizable through `customizationService`:

```js
export default function getCustomizationModule({ servicesManager, extensionManager }) {
  return [
    {
      name: 'default',
      value: [
        {
          id: 'studyBrowser.sortFunctions',
          values: [
            {
              label: 'Series Number',
              sortFunction: (a, b) => {
                return a?.SeriesNumber - b?.SeriesNumber;
              },
            },
            {
              label: 'Series Date',
              sortFunction: (a, b) => {
                const dateA = new Date(formatDate(a?.SeriesDate));
                const dateB = new Date(formatDate(b?.SeriesDate));
                return dateB.getTime() - dateA.getTime();
              },
            },
          ],
        },
      ],
    },
  ];
  ```
  
### Settings menu

<p align="center">

</p>
<p align="center">The dropdown offers sorting and study filtering options.</p>

## Opting Into the New UI

To enable the experimental UI, set the `appConfig` option `useExperimentalUI` to `true`. This will activate the new study panel and other updated components. If users prefer the older design, they should leave this option as `false`.

## Demo

CleanShot 2024-09-07 at 22 50 47

This redesign is a work-in-progress, and feedback is welcome for future improvements.

## Deploy preview

You can access a deploy preview [here](https://66dd21a6aa07a9893b6f12ad--inspiring-salamander-51e88d.netlify.app/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.21255249241959598781018667405790)

### CODE REVIEW

File: platform/ui/src/components/SidePanel/SidePanel.tsx [hunk 1]
```diff
@@ -5,6 +5,7 @@ import { useTranslation } from 'react-i18next';
 
 import Icon from '../Icon';
 import Tooltip from '../Tooltip';
+import { Separator, ToggleGroup, ToggleGroupItem } from '@ohif/ui-next';
```
sedghi, line 8: This seems unusual. All of this should move to ui-next. There shouldn't be any imports here. Ideally, this PR shouldn't touch platform/ui at all. Any necessary code should be copied and pasted into ui-next, as we'll eventually delete platform/ui.

IbrahimCSAE: good callout, This flew over my head at the start, now it makes sense

extensions/cornerstone-dicom-seg/src/getPanelModule.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
 
 import { useAppConfig } from '@state';
-import { Toolbox } from '@ohif/ui';
+import { Toolbox as NewToolbox } from '@ohif/ui-next';
```
wayfarer3130, line 4: Add a comment to the use of this that it is activated by the use experimental flag and that the use of OldToolBox is going away once the new one is fully functional.

extensions/cornerstone-dicom-seg/src/panels/PanelSegmentation.tsx
```diff
@@ -319,52 +323,76 @@ export default function PanelSegmentation({
...
+            thickness="2px"
+          />
+        </>
+      )}
+      <SegmentationGroupTableComponent
+        title={t('Segmentations')}
+        segmentations={segmentations}
+        disableEditing={configuration.disableEditing}
+        activeSegmentationId={selectedSegmentationId || ''}
+        onSegmentationAdd={onSegmentationAddWrapper}
```
wayfarer3130, line 350: The right way to add these is to provide the commands manager and just have the child component execute commands.  Then, to customize it, add a group priority in the commands manager so that the right command gets executed.  Having the UI decide where to get customizations from isn't a good idea.  The problem is that you are exposing volatility in multiple places, rather than letting a single component in the architecture manage the volatility of hte mapping between command invocations and which actual command to run.

sedghi: Great point, for now since this is experimental I will merge it

sedghi (approved):
Thanks this is really great stuff

Kudos to @dan-rukas for the amazing design too

PR #4265, opened 2024-06-26
fix: webpack import bugs showing warnings on import

wayfarer3130:
relies on https://github.com/OHIF/Viewers/pull/4182 to be merged

### Context

This change moves the dynamic import of modules into the plugin config so that they can be configured as desired.  Then, this is used to import peer dependencies for cornerstone.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/dicom-microscopy/src/services/MicroscopyService.ts
```diff
@@ -69,6 +70,10 @@ export default class MicroscopyService extends PubSubService {
     });
   }
 
+  public importDicomMicroscopyViewer(): Promise<any> {
```
wayfarer3130, line 73: This seems like a reasonable way to access the imports, regardless of whether they are dynamic or statically loaded.

platform/app/pluginConfig.json
```diff
@@ -88,6 +88,12 @@
   "public": [
     {
       "directory": "./platform/public"
+    },
+    {
```
wayfarer3130, line 92: This example shows how to declare a package and to copy a node modules result to the output.

platform/app/public/html-templates/index.html
```diff
@@ -214,6 +214,12 @@
       rel="preload"
       as="style"
     />
+    <script>
```
wayfarer3130, line 217: This is the work around for the dynamic import - just use the browser native import.

sedghi (approved):

PR #4274, opened 2024-07-02
feat: Add hanging protocol other configuration values

wayfarer3130:
### Context

Hanging protocols have a number of other settings they need to be able to apply, in addition to the basic viewport type settings.  These are things like panel ordering for the right panel, primary menu name, tool group id, callbacks, image load strategy etc.  These are currently dumped at the top level of the protocol, but should really be grouped appropriately so that specific configurations could be created for specific types.

### Changes & Results

Added a new type/value to the hanging protocol 'otherDisplayConfiguration' to enable setting display specific configuration.

### Testing

Run the unit tests

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/platform/core/src/services/HangingProtocolService/HangingProtocolService.test.js
+++ b/platform/core/src/services/HangingProtocolService/HangingProtocolService.test.js
@@ -11,6 +11,10 @@ const testProtocol = {
       },
     },
   ],
+  otherDisplayConfiguration: {
+    panel1: { name: 'Panel1' },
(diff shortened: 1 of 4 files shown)


sedghi (commented):
look into workflow manager and see if that satisfies the use case  modes/preclinical-4d/src/getWorkflowSettings.ts

Issue #4210, opened 2024-06-06
[Bug] US modality with single slice data shows wrong orientation marker on viewport, it shows H & A on top & left


HemantPatel30:
### Describe the Bug

US modality with single slice data shows wrong orientation marker on viewport, it shows H & A on top & left.

Data does not contains rowCosines & ColumCosine but it still shows wrong information on viewport.

Issue is in our code, where it tries to read WADO_IMAGE_LOADER.imagePlaneModule & rowCosine & ColumnCosine is undefined we are returning hardcoded value [0, 1, 0] & [0, 0, -1].

### Steps to Reproduce

https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.1188.2803.137585363493444318569098508293

Load last series with 2 images into it from series navigator.

### The current behavior

it shows orientation marker H & A on viewport when image is loading

### The expected behavior

It should not show orientation marker

### OS

macOS

### Node version

18.16.2

### Browser

Chrome

sedghi:
Is this fixed? Can I close it? Reopen if the issue persists.

PR #4275, opened 2024-07-02
fix(image-orientation): Prevent incorrect orientation marker display for single-slice US images

HemantPatel30:
… not have value (ImageOrientationPatient) in header

This PR provide a Fix for verified Bug #4210 

https://github.com/OHIF/Viewers/issues/4210

Issue wa, if ImageOrientationPatient - Metadata does not exists then it's returning fix - hard coded value, which was used to calculate orientation marker - so for any single slice US image, it always shows orientation marker, which is wrong.

as Image Plane module was used at different places, fix provide a flag if default value is being set, & if default value is being set then do not display orientation marker

It will not display orientation marker on UI, if ImageOrientationMaker is not available into metadata

Done testing with DICOMWEB & DICOMLOCAL data source

https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.1188.2803.137585363493444318569098508293

Load last series with 2 images into it from series navigator.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


sedghi (approved):
Thanks for your contribution

PR #4284, opened 2024-07-10
feat(pmap): added support for parametric map

lscoder:
Adds support for DICOM Parametric Maps

1. Added a new extension `cornerstone-dicom-pmap` that has the `sopClassHandler` responsible for handling and loading parametric maps SOP Class UIDs
2. The new extension is added to longitudinal mode and it also uses the new adapter responsible for loading parametric maps on CS3D side

https://github.com/user-attachments/assets/6ec94a95-4ae0-4964-9f75-94df71ba9e59

Comparison between OHIF and Slicer to make sure the data is loaded properly

https://github.com/user-attachments/assets/7f7b1181-3e71-4964-962d-59e00a88c70b


extensions/cornerstone-dicom-pmap/src/utils/initPMAPToolGroup.ts
```diff
@@ -0,0 +1,6 @@
+function createPMAPToolGroupAndAddTools(ToolGroupService, customizationService, toolGroupId) {
+  const { tools } = customizationService.get('cornerstone.overlayViewportTools') ?? {};
```
sedghi, line 2: I'm not sure what overlayViewportTools is, but it seems like someone else added it rather than you.

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -304,6 +304,11 @@ const OHIFCornerstoneViewport = React.memo((props: withAppTypes) => {
         initialImageIndex
       );
 
+      // Make sure displaySetOptions has one object per displaySet
+      while (displaySetOptions.length < displaySets.length) {
+        displaySetOptions.push({});
+      }
+
```
sedghi, line 311: Does it make sense to put it outside this function?

extensions/cornerstone/src/services/ParametricMapService/index.ts
```diff
@@ -0,0 +1 @@
+export { ParametricMapService } from './ParametricMapService'
```
sedghi, line 1: seems like you forgot to commit the service? or maybe you don't need the service anymore?

extensions/cornerstone-dicom-pmap/src/getSopClassHandlerModule.ts
```diff
@@ -0,0 +1,224 @@
...
+  extensionManager,
+  servicesManager,
+  displaySet,
+  headers,
+}: withAppTypes) {
+  const utilityModule = extensionManager.getModuleEntry(
+    '@ohif/extension-cornerstone.utilityModule.common'
+  );
+
+  const { dicomLoaderService } = utilityModule.exports;
```
sedghi, line 171: We can now import components from other extensions, eliminating the need for the utility module. Check if the `dicomLoaderService` is exported from the Cornerstone extension. If it is, import it at the top of the file. If not, export it from the Cornerstone extension.

extensions/cornerstone-dicom-pmap/src/getSopClassHandlerModule.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    const referencedDisplaySet = displaySet.getReferenceDisplaySet();
+    const referencedVolumeURI = referencedDisplaySet.displaySetInstanceUID;
+    const referencedVolumeId = `${VOLUME_LOADER_SCHEME}:${referencedVolumeURI}`;
+
+    displaySet.referencedVolumeURI = referencedVolumeURI;
+    displaySet.referencedVolumeId = referencedVolumeId;
+
+    return referencedVolumeId;
+  };
```
sedghi, line 111: There's nothing asynchronous about this. I'm wondering why we're creating a method for it. Does it need to be a method, or can we simply derive the values?

extensions/cornerstone-dicom-pmap/src/getSopClassHandlerModule.ts
```diff
@@ -0,0 +1,224 @@
...
+  const arrayBuffer = await dicomLoaderService.findDicomDataPromise(displaySet, null, headers);
+  const referencedVolumeId = displaySet.getReferencedVolumeId();
+  const cachedReferencedVolume = cache.getVolume(referencedVolumeId);
+
+  if (!cachedReferencedVolume) {
+    throw new Error(
+      'Referenced Volume is missing for the SEG, and stack viewport SEG is not supported yet'
+    );
+  }
+
```
sedghi, line 181: Add a comment indicating that this `_load` function must run after the base volume is mounted; otherwise, it will not work.

extensions/cornerstone-dicom-pmap/src/getSopClassHandlerModule.ts
```diff
@@ -0,0 +1,224 @@
...
+    metaData
+  );
+  const { pixelData } = results;
+  const TypedArrayConstructor = pixelData.constructor;
+  const paramMapId = displaySet.displaySetInstanceUID;
+
+  const derivedVolume = await volumeLoader.createAndCacheDerivedVolume(referencedVolumeId, {
+    volumeId: paramMapId,
+    targetBuffer: {
+      type: TypedArrayConstructor.name, // 'Int16Array'
```
sedghi, line 195: remove Int16Array comment

extensions/cornerstone-dicom-pmap/src/getSopClassHandlerModule.ts
```diff
@@ -0,0 +1,224 @@
...
+    },
+  });
+
+  derivedVolume.getScalarData().set(pixelData);
```
sedghi, line 199: I believe the `createAndCacheDerivedVolume` method can accept `pixelData` as an argument, eliminating the need to handle it separately afterwards. Take a look at the parameters for `createAndCacheDerived` to confirm this.

extensions/cornerstone-dicom-pmap/src/viewports/OHIFCornerstonePMAPViewport.tsx
```diff
@@ -0,0 +1,162 @@
...
+          { value: 0, opacity: 0.5 },
+          { value: 1, opacity: 1 },
+        ],
+      },
+    });
+
+    return (
+      <Component
+        {...props}
+        displaySets={[referencedDisplaySet, pmapDisplaySet]}
```
sedghi, line 60: I think the order here is crucial. The reverse sequence would not be effective. Please add a comment explaining this.

extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
```diff
@@ -200,6 +200,8 @@ class CornerstoneCacheService {
     const volumeData = [];
 
     for (const displaySet of displaySets) {
+      const { hasVolumeData = false } = displaySet;
```
sedghi, line 203: I believe it should be `isOverlayVolume`, right? It's not about whether it has volume data or not, but rather about whether to overlay or not, I guess.

sedghi (commented):
Thank you. It looks good. Please see my comments.

extensions/cornerstone/src/services/CornerstoneCacheService/CornerstoneCacheService.ts
lscoder: Updated

sedghi (approved):
Thanks

PR #4287, opened 2024-07-12
feat(SR): text structured report (TEXT, CODE, NUM, PNAME, DATE, TIME and DATETIME)

lscoder:
### Context

Added support for text Structure Reports that can contain TEXT, CODE, NUM, PNAME, DATE, TIME and DATETIME content items, nested containers and separate/continous ContinuityOfContent. The value types IMAGE, WAVEFORM, COMPOSITE, SCOORD and TCOORD are still not supported. The user shall see "[VALUE_TYPE is not supported]" for each of the unsupported value types.

IMAGE value type support will be added soon and SCOORD/TCOORD later.

### Testing

Some SR files can be downloaded from here:
https://www.dropbox.com/scl/fi/mmq2ofh88c9ytn7fmkx7z/Structured-Reports.zip?rlkey=46szt3w974jwlgs0paw617t2l&st=0sglct4o&dl=0

### DIFF
--- a/extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRContainer.tsx
+++ b/extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRContainer.tsx
@@ -1,0 +1,78 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { OHIFCornerstoneSRContentItem } from './OHIFCornerstoneSRContentItem';
+
+export function OHIFCornerstoneSRContainer(props) {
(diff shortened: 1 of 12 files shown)


sedghi (commented):
Very nice addition!

The unit test is failing because you added something to utils in the core. You need to add that to the unit test as well and export it.

@wayfarer3130, I dislike this test.

wayfarer3130:
> Very nice addition!
> 
> The unit test is failing because you added something to utils in the core. You need to add that to the unit test as well and export it.
> 
> @wayfarer3130, I dislike this test.

Yes, I hate that unit test too.

sedghi (approved):

Issue #4232, opened 2024-06-13
The instances of a 4D MR series are not loading completely.


rghgit1:
### Describe the Bug

A series has both 4D and 3D instances. Only 3D instance are getting displayed in viewport. The 4D instance are not displayed. Attached is the series which is having issue.
[d64b9ef7-0434b983-b6fad63e-94c7c0f0-b64c7464.zip](https://github.com/user-attachments/files/15826807/d64b9ef7-0434b983-b6fad63e-94c7c0f0-b64c7464.zip)

The attached series has 36 instances. 1-18 are 4D and 19-36 are 3D. Only 19-36 instances are getting displayed in viewport.

This is happening with v3.8.0. However, with v3.8.0_beta.18 all the scans were getting loaded.

### Steps to Reproduce

Load the attached series to Orthanc.
View the series using Ohif v 3.8.0

### The current behavior

Only instance 19-36 are getting displayed 
![image](https://github.com/OHIF/Viewers/assets/133641462/69f34398-1744-497d-91e2-724392e996a0)

### The expected behavior

Load all 36 instance and display a warning message similar to how it was in v3.8.0_beta.18 version
![image](https://github.com/OHIF/Viewers/assets/133641462/3b115f07-0dca-4ddd-8020-728c66573560)

### OS

Window11

### Node version

21.6.1

### Browser

Edge 125.0.2535.92

rghgit1:
extensions\default\src\getSopClassHandlerModule.js  - Looks like the DYNAMIC_VOLUME_LOADER_SCHEME is causing this behavior for timepoint data. If I use DEFAULT_VOLUME_LOADER_SCHEME all the instances are displayed (the order is still not correct though)

rghgit1:
Made below changes in extensions\default\src\getSopClassHandlerModule.js to get it working for now. Only use DYNAMIC_VOLUME_LOADER_SCHEME for CT/PT. Do suggest if there are better solutions. 

`function getDisplaySetInfo(instances) {
  const dynamicVolumeInfo = getDynamicVolumeInfo(instances);
  const { isDynamicVolume, timePoints } = dynamicVolumeInfo;
  let displaySetInfo;

  const { appConfig } = appContext;

  // if (isDynamicVolume) {
  if (isDynamicVolume && (instances.Modality === 'CT' || instances.Modality === 'PT')) {
    const timePoint = timePoints[0];
    const instancesMap = new Map();`

adriangaliana:
[...]

PR #4294, opened 2024-07-17
fix: 4D series loaded in basic mode #4232

lscoder:
### Context

4D series was not loading in basic mode because the image ids were being split into time points even though all images should be loaded in a stack viewport ("normal mode").

### Changes & Results

4D mode is disabled until preclinical 4D mode is started when it set 4D mode to `true`. Once 

### Testing

Load the study shared [here](https://github.com/OHIF/Viewers/issues/4232) and make sure it loads in basic mode.

### DIFF
--- a/modes/preclinical-4d/src/index.tsx
+++ b/modes/preclinical-4d/src/index.tsx
@@ -1,9 +1,12 @@
 import { id } from './id';
 import { hotkeys } from '@ohif/core';
+import * as cornerstoneExtension from '@ohif/extension-cornerstone';
 import initWorkflowSteps from './initWorkflowSteps';
 import initToolGroups from './initToolGroups';
 import toolbarButtons from './toolbarButtons';
 import segmentationButtons from './segmentationButtons';
+
+const { volumeLoaderUtil } = cornerstoneExtension.utils;
 
 const extensionDependencies = {
   '@ohif/extension-default': '3.7.0-beta.76',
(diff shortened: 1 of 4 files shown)


modes/preclinical-4d/src/index.tsx
```diff
@@ -48,6 +48,13 @@ function modeFactory({ modeConfiguration }) {
         viewportGridService,
       } = servicesManager.services;
 
+      const { exports: volumeLoaderUtil } = extensionManager.getModuleEntry(
```
sedghi, line 51: use import instead

lscoder: Updated

modes/preclinical-4d/src/index.tsx
```diff
@@ -91,14 +97,17 @@ function modeFactory({ modeConfiguration }) {
...
+    onModeExit: ({ servicesManager, extensionManager }: withAppTypes) => {
       const {
         toolGroupService,
         syncGroupService,
         segmentationService,
         cornerstoneViewportService,
       } = servicesManager.services;
 
+      // Turn 4D mode off to move back to its initial state.
+      volumeLoaderUtil.disable4DMode();
```
sedghi, line 109: This situation is somewhat concerning. What's the reason for it? If it's truly necessary, we should implement a service with an `onModeExit` method to ensure it always runs.

sedghi (approved):
looks great

sedghi (changes requested):
Wrong PR aproved

modes/preclinical-4d/src/index.tsx
lscoder: There is a call to `volumeLoaderUtil.enable4DMode()` on line 55 in `onModeEnter` and this call to `disable4DMode()` here in `onModeExit` because only in this mode 4D will be enabled otherwise the viewer will try to split any series into a 4D series even in basic mode.

sedghi:
I don't think we need this anymore

PR #4315, opened 2024-07-26
feat(SR): SCOORD3D point annotations support for stack viewports

igoroctaviano:
### Context

- Support for SCOORD3D point annotations (just rendering / no hydration)
- Support for SCT finding site (rendering labels to point and panel)
- Update to untracked findings so the user can interact (jump to measurement) 
- Issue: https://github.com/OHIF/Viewers/issues/1215
- https://github.com/cornerstonejs/cornerstone3D/pull/1446

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


igoroctaviano:
@wayfarer3130 could you please review it and test using this SCOORD3D point? https://drive.google.com/drive/folders/1Zwk4KcbTbylilmk6U0byrxsDP1fO7UGx?usp=drive_link

extensions/cornerstone-dicom-sr/src/enums.js
```diff
@@ -0,0 +1,39 @@
+import { adaptersSR } from '@cornerstonejs/adapters';
+
+const { CodeScheme: Cornerstone3DCodeScheme } = adaptersSR.Cornerstone3D;
+
+export const CodeNameCodeSequenceValues = {
```
wayfarer3130, line 5: It is definitely useful to have the enums provided here - could they be in one file per enum?  I generally find that easier to organize, but will let you decide if you want to do that.

extensions/cornerstone-dicom-sr/src/enums.js
wayfarer3130, line 1: Please export these enums from the top level of this module as they get redefined several places in other modules and should be defined once rather than multiple times.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -1,27 +1,30 @@
...
+ * - [ ] Get stacks from referenced displayInstanceUID and load into wrapped CornerStone viewport
+ */
 
 const sopClassUids = [
-  '1.2.840.10008.5.1.4.1.1.88.11', // BASIC TEXT SR
-  '1.2.840.10008.5.1.4.1.1.88.22', // ENHANCED SR
-  '1.2.840.10008.5.1.4.1.1.88.33', // COMPREHENSIVE SR
-  '1.2.840.10008.5.1.4.1.1.88.34', // Comprehensive 3D SR
-  // '1.2.840.10008.5.1.4.1.1.88.50', // Mammography CAD SR
+  '1.2.840.10008.5.1.4.1.1.88.11' /** BASIC_TEXT_SR */,
```
wayfarer3130, line 23: Please import these from @ohif/core sopClassDictionary (you will need to export it at the top level I believe)

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -132,6 +123,12 @@ function _getDisplaySetsFromSeries(
   return [displaySet];
 }
 
+/**
+ * Loads the display set with the given services and extension manager.
+ * @param {Object} displaySet - The display set to load.
```
wayfarer3130, line 128: You don't need to include the type prefix if you area defining the types within the function call itself.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -132,6 +123,12 @@ function _getDisplaySetsFromSeries(
...
+ * @param {Object} servicesManager - The services manager containing displaySetService and measurementService.
+ * @param {Object} extensionManager - The extension manager containing data sources.
+ */
 async function _load(displaySet, servicesManager: AppTypes.ServicesManager, extensionManager) {
```
wayfarer3130, line 132: There is a type definition for display set so that you can define both displaySet and extensionManager as the appropriate types.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -179,7 +176,7 @@ async function _load(displaySet, servicesManager: AppTypes.ServicesManager, exte
   displaySet.isRehydratable = isRehydratable(displaySet, mappings);
   displaySet.isLoaded = true;
 
-  // Check currently added displaySets and add measurements if the sources exist.
+  /** Check currently added displaySets and add measurements if the sources exist */
```
wayfarer3130, line 179: This comment should be // as it is a single line comment, not a function/class comment.

igoroctaviano: This is an opinion. I find /** format to be way more cleaner than // which are often use for temporary comments or commenting blocks of code. I use them for "permanent comments" or docs.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
igoroctaviano: Sounds good, I'll update it.

extensions/cornerstone-dicom-sr/src/enums.js
igoroctaviano: I believe having one file per enum is overkill (too many imports and files).

extensions/cornerstone/src/initMeasurementService.ts
```diff
@@ -230,7 +238,7 @@ const connectToolsToMeasurementService = (servicesManager: AppTypes.ServicesMana
         annotationToMeasurement(toolName, annotationAddedEventDetail);
       }
     } catch (error) {
-      console.warn('Failed to update measurement:', error);
+      console.error('Failed to update measurement:', error);
```
wayfarer3130, line 241: This had better not be error unless you ensure that annotationToMeasurement is actually found/doesn't throw an exception when the measurement is simply not registered.

extensions/cornerstone/src/utils/measurementServiceMappings/constants/supportedTools.js
```diff
@@ -11,6 +11,8 @@ export default [
   'PlanarFreehandROI',
   'SplineROI',
   'LivewireContour',
-  'Probe',
```
wayfarer3130, line 14: Is there a reason not to support probe?

modes/longitudinal/src/initToolGroups.js
```diff
@@ -68,6 +68,7 @@ function initDefaultToolGroup(
       { toolName: toolNames.Bidirectional },
       { toolName: toolNames.DragProbe },
       { toolName: toolNames.Probe },
+      { toolName: 'SCOORD3DPoint' },
```
wayfarer3130, line 71: Can you export this name from toolNames please.

extensions/cornerstone-dicom-sr/src/utils/getRenderableData.ts
```diff
@@ -145,13 +100,19 @@ function _getRenderableData(GraphicType, GraphicData, imageId) {
...
 
       const pointsWorld: csTypes.Point3[] = [];
-      for (let i = 0; i < GraphicData.length; i += 2) {
-        const worldPos = utilities.imageToWorldCoords(imageId, [
-          GraphicData[i],
-          GraphicData[i + 1],
-        ]);
 
-        pointsWorld.push(worldPos);
+      if (ValueType === 'SCOORD3D') {
```
wayfarer3130, line 104: This looks like the right way to convert this data - it will work for 3d length and other 3d coordinates.

extensions/cornerstone-dicom-sr/src/utils/addSRAnnotation.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    measurementData.renderableData[GraphicType].push(
+      getRenderableData(GraphicType, GraphicData, ValueType, imageId)
+    );
+  });
+
+  const { ValueType, GraphicType } = measurement.coords[0];
+
+  /** TODO: Add support for other SCOORD3D graphic types */
```
wayfarer3130, line 31: Add to the comment something like: 
Read the tool name from the DICOM SR identification type in the future.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -561,51 +648,56 @@ function _processNonGeometricallyDefinedMeasurement(mergedContentSequence) {
...
   if (ValueType === 'SCOORD') {
-    const { ReferencedSOPSequence } = item.ContentSequence;
-
+    const { ReferencedSOPSequence } = graphicItem.ContentSequence;
     coords.ReferencedSOPSequence = ReferencedSOPSequence;
   } else if (ValueType === 'SCOORD3D') {
-    const { ReferencedFrameOfReferenceSequence } = item.ContentSequence;
-
-    coords.ReferencedFrameOfReferenceSequence = ReferencedFrameOfReferenceSequence;
+    if (graphicItem.ReferencedFrameOfReferenceUID) {
```
wayfarer3130, line 664: This is easier written as:
coords.ReferencedFrameOfReferenceSequence = graphicItem.ReferencedFrameOfReferenceUID || graphicItem.ContenxtSequence?.ReferencedFrameOfReferenceSequence

wayfarer3130, line 664: Can you assign the referenced sop sequence regardless if it is 3d or 2d coordinates?  It is perfectly value to have an annotation reference an image only, and NOT reference the frame of reference uid, for example, when applying a length on a specific image as opposed to on every image in that frame of reference.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -465,20 +539,36 @@ function _processTID1410Measurement(mergedContentSequence) {
...
     if (MeasuredValueSequence) {
       measurement.labels.push(
         _getLabelFromMeasuredValueSequence(ConceptNameCodeSequence, MeasuredValueSequence)
       );
     }
   });
 
+  const findingSites = mergedContentSequence.filter(
+    item =>
+      item.ConceptNameCodeSequence.CodingSchemeDesignator === CodingSchemeDesignators.SCT &&
```
wayfarer3130, line 551: This is inconsistent with other uses of the labels/finding sites.  Labels is a bit artificial - I'd rather see it use the finding sites directly.  I can show you an example of how that works in the basic test view.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -205,69 +204,100 @@ async function _load(displaySet, servicesManager: AppTypes.ServicesManager, exte
...
+  const { sopClassUids, images } = newDisplaySet;
 
-  // Check if any have the newDisplaySet is the correct SOPClass.
+  /** Check correct SOPClassUID */
   unloadedMeasurements = unloadedMeasurements.filter(measurement =>
-    measurement.coords.some(coord =>
-      sopClassUids.includes(coord.ReferencedSOPSequence.ReferencedSOPClassUID)
-    )
+    measurement.coords.some(coord => {
+      for (let i = 0; i < images.length; ++i) {
```
wayfarer3130, line 245: This is a fairly long loop going over both the measurements and coordinates.  Is there a way to only do it for coordinates containing a frame of reference, and then figure out which image inside a volume should apply using the distance calculations from the volume display set?

extensions/cornerstone/src/initMeasurementService.ts
igoroctaviano: It's catching an error, no?

extensions/cornerstone/src/utils/measurementServiceMappings/constants/supportedTools.js
igoroctaviano: Probe its still there.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
igoroctaviano: I'm using the same code today but I extended it to support SCT instead of just SRT.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
igoroctaviano: Updated!

extensions/cornerstone-dicom-sr/src/utils/getRenderableData.ts
```diff
@@ -1,94 +1,49 @@
...
     case SCOORD_TYPES.CIRCLE: {
       const pointsWorld = [];
-      for (let i = 0; i < GraphicData.length; i += 2) {
-        const worldPos = utilities.imageToWorldCoords(imageId, [
-          GraphicData[i],
-          GraphicData[i + 1],
-        ]);
 
-        pointsWorld.push(worldPos);
+      if (ValueType === 'SCOORD3D') {
```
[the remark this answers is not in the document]
igoroctaviano: Updated!

extensions/cornerstone/src/initMeasurementService.ts
wayfarer3130: It is catching an error, but it is one that happens at least semi-regularly.  I agree it SHOULD be an error, but that requires fixing the code to handle a few situations that aren't well handled.  You can convert it, but then you need to go through the various unhandled conversion types and make sure that they don't throw errors any longer, but just warn with a message about unsupported conversion types.  That is, the flow should be that all normal situations are handled without exceptions, and only if the code is incorrect should it throw an exception.

igoroctaviano: Updated.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -205,69 +210,101 @@ async function _load(displaySet, servicesManager: AppTypes.ServicesManager, exte
...
+    })
   );
 
   if (unloadedMeasurements.length === 0) {
-    // New displaySet isn't the correct SOPClass, so can't contain the referenced images.
+    /** New displaySet isn't the correct SOPClassso can't contain the referenced images */
     return;
   }
 
   const SOPInstanceUIDs = [];
```
wayfarer3130, line 295: Could you change this to a Set<string> as that turns the algorithm from O(n^2) to O(n) because it doesn't need to check every instance for equality.  It should be only a few lines changed.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -205,69 +210,101 @@ async function _load(displaySet, servicesManager: AppTypes.ServicesManager, exte
...
 
-  const imageIdsForDisplaySet = dataSource.getImageIdsForDisplaySet(newDisplaySet);
-
-  for (const imageId of imageIdsForDisplaySet) {
-    if (!unloadedMeasurements.length) {
-      // All measurements loaded.
-      return;
-    }
-
+  const imageIds = dataSource.getImageIdsForDisplaySet(newDisplaySet);
```
wayfarer3130, line 305: If you create a Map<SOP Instance, ImageId[]> up front then, you can iterate over the relevant image ids.  You could even find the exact one of interest by including the frame# in the image id map.  Otherwise very large image id maps will get quite slow, even for a small number of measurements.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
igoroctaviano: Updated!

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
igoroctaviano: Updated!

extensions/measurement-tracking/src/getPanelModule.tsx
```diff
@@ -24,7 +24,6 @@ function getPanelModule({ commandsManager, extensionManager, servicesManager }):
         />
       ),
     },
```
sedghi, line 26: revert please

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -611,4 +611,4 @@ OHIFCornerstoneViewport.propTypes = {
   initialImageIdOrIndex: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
 };
 
-export default OHIFCornerstoneViewport;
+export default OHIFCornerstoneViewport;
```
sedghi, line 614: revert please

extensions/cornerstone-dicom-sr/src/utils/getLabelFromDCMJSImportedToolData.js
```diff
@@ -9,13 +13,15 @@
...
 
   if (freeTextLabel) {
     return freeTextLabel.CodeMeaning;
   }
 
-  if (finding && finding.CodeValue === 'CORNERSTONEFREETEXT') {
+  if (finding && finding.CodeValue === Cornerstone3DCodeScheme.codeValues.CORNERSTONEFREETEXT) {
     return finding.CodeMeaning;
   }
 }
```
sedghi, line 27: maybe extract above this doc, like other files

extensions/cornerstone-dicom-sr/src/utils/addToolInstance.ts
```diff
@@ -1,8 +1,16 @@
...
-export default function addToolInstance(name: string, toolClass, configuration?): void {
+export default function addToolInstance(name: string, toolClass, configuration = {}): void {
   class InstanceClass extends toolClass {
     static toolName = name;
+    constructor(toolProps, defaultToolProps) {
+      if (toolProps.configuration) {
+        toolProps.configuration = { ...toolProps.configuration, ...configuration };
+      } else {
+        toolProps.configuration = configuration;
+      }
```
sedghi, line 11: toolProps.configuration = toolProps.configuration
? { ...toolProps.configuration, ...configuration }
: configuration;

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -205,115 +207,153 @@ async function _load(displaySet, servicesManager: AppTypes.ServicesManager, exte
...
+  }
 
-  // Check if any have the newDisplaySet is the correct SOPClass.
+  // Filter unloaded measurements based on SOPClassUID
   unloadedMeasurements = unloadedMeasurements.filter(measurement =>
-    measurement.coords.some(coord =>
-      sopClassUids.includes(coord.ReferencedSOPSequence.ReferencedSOPClassUID)
-    )
+    measurement.coords.some(coord => {
+      if (coord.ReferencedSOPSequence === undefined && coord.ReferencedFrameOfReferenceSequence) {
```
sedghi, line 255: can we reduce indentation by returning early here instead?

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -205,115 +207,153 @@ async function _load(displaySet, servicesManager: AppTypes.ServicesManager, exte
...
+          sliceNormal[0] = orientation[1] * orientation[5] - orientation[2] * orientation[4];
+          sliceNormal[1] = orientation[2] * orientation[3] - orientation[0] * orientation[5];
+          sliceNormal[2] = orientation[0] * orientation[4] - orientation[1] * orientation[3];
+
+          let distanceAlongNormal = 0;
+          for (let j = 0; j < 3; ++j) {
+            distanceAlongNormal += sliceNormal[j] * imageMetadata.ImagePositionPatient[j];
+          }
+
+          /** Assuming 2 mm tolerance */
```
sedghi, line 273: where is this assumption coming from? can we introduce a epsilon on top?

sedghi (commented):
see my comments please, great work

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
fedorov: Great catch!

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
igoroctaviano: I don't think it's going to make it better though, I'll have to run two checks instead of one e.g. 

```js
// Filter unloaded measurements based on SOPClassUID
  unloadedMeasurements = unloadedMeasurements.filter(measurement =>
    measurement.coords.some(coord => {
      if (coord.ReferencedSOPSequence !== undefined || !coord.ReferencedFrameOfReferenceSequence) {
        return coord.ReferencedSOPSequence &&
          sopClassUidSet.has(coord.ReferencedSOPSequence.ReferencedSOPClassUID);
      }
      
      if (coord.ReferencedSOPSequence === undefined && coord.ReferencedFrameOfReferenceSequence) {
        for (const [key, imageId] of imageIdMap) {
          const imageMetadata = metadataProvider.get('instance', imageId);
          if (imageMetadata.FrameOfReferenceUID !== coord.ReferencedFrameOfReferenceSequence) {
            continue;
          }

          const sliceNormal = [0, 0, 0];
          const orientation = imageMetadata.ImageOrientationPatient;
          sliceNormal[0] = orientation[1] * orientation[5] - orientation[2] * orientation[4];
          sliceNormal[1] = orientation[2] * orientation[3] - orientation[0] * orientation[5];
          sliceNormal[2] = orientation[0] * orientation[4] - orientation[1] * orientation[3];

          let distanceAlongNormal = 0;
          for (let j = 0; j < 3; ++j) {
            distanceAlongNormal += sliceNormal[j] * imageMetadata.ImagePositionPatient[j];
          }

          // Add this constant near the top of the file, after the imports
          const POSITION_EPSILON = 2; // 2 mm tolerance

          // Use POSITION_EPSILON instead of hardcoded value
          if (Math.abs(distanceAlongNormal - coord.GraphicData[2]) > POSITION_EPSILON) {
            continue;
          }

          const [SOPInstanceUID, frameNumber] = key.split(':');
          coord.ReferencedSOPSequence = {
            ReferencedSOPClassUID: imageMetadata.SOPClassUID,
            ReferencedSOPInstanceUID: SOPInstanceUID,
            ReferencedFrameNumber: frameNumber !== '1' ? parseInt(frameNumber, 10) : undefined,
          };

          break;
        }
      }

      return (
        coord.ReferencedSOPSequence &&
        sopClassUidSet.has(coord.ReferencedSOPSequence.ReferencedSOPClassUID)
      );
    })
  );
```

Any suggestions?

igoroctaviano:
@sedghi, could you please review this? If any changes or clean-ups are needed, I'll handle them based on your final feedback.

platform/app/src/components/ViewportGrid.tsx
```diff
@@ -173,6 +183,26 @@ function ViewerViewportGrid(props: withAppTypes) {
...
+              .getImageIds()
+              .indexOf(referencedImageId);
+            csToolsUtils.jumpToSlice(currentViewport.element, { imageIndex });
+          } catch (error) {
+            console.warn(
+              'ViewportGrid::Unable to jump to image based on measurement coordinate',
+              error
+            );
+          }
+
```
sedghi, line 205: This shoouldn't be here, cornerstoneviewport service should not get called in the platform/**/*

modes/longitudinal/src/index.ts
```diff
@@ -199,17 +200,14 @@ function modeFactory({ modeConfiguration }) {
...
                   ],
                 },
                 {
                   namespace: dicomsr.viewport,
                   displaySetsToDisplay: [dicomsr.sopClassHandler],
                 },
-                // {
-                //   namespace: dicomvideo.viewport,
-                //   displaySetsToDisplay: [dicomvideo.sopClassHandler],
-                // },
```
sedghi, line 212: please bring back this one

platform/app/src/components/ViewportGrid.tsx
igoroctaviano: Updated

modes/longitudinal/src/index.ts
igoroctaviano: Updated

sedghi (approved):

fedorov:
Really happy to see this effort completed! Thank you @igoroctaviano and @sedghi for bringing it over the finish line! 🏆

PR #4322, opened 2024-07-31
fix(dicomjson): Update getUIDsFromImageID to work with json data source + update getDisplaySetImageUIDs to work with mixed sop class json  

igoroctaviano:
### Context
- MetadataProvider's getUIDsFromImageID should look first for previously added image id's UIDs. 
- getDiplaySetImageIds should work for JSON files that have mixed sop class series. 

Example: The DICOM JSON data source already calls addImageIdToUIDs when loading the JSON data during initialization so when using JSON, we might not be able to get UIDs from the image IDs. Also, if the image ID was generated from the URL attribute of the JSON file and the URL does not have any UIDs as part of the URL (e.g. proprietary URLs) then it still works since the JSON already provided all those UIDs during initialization.

### Testing

Using any multiframe json (multiple instances with URL attributes and wadors loader to retrieve frames/pixeldata)
```
{
  "studies": [
    {
      "StudyInstanceUID": "1.2.300.0.7230010.3.1.2.2137700.1796.3",
      "StudyDescription": "Test",
      "StudyDate": "20240613",
      "StudyTime": "181423.930000",
      "PatientName": [
        {
          "Alphabetic": "Test^Test"
        }
      ],
      "PatientID": "Test",
      "AccessionNumber": "Test",
      "PatientAge": "Test",
      "PatientSex": "F",
      "series": [
        {
          "SeriesInstanceUID": "1.2.300.0.7230010.3.1.3.321.7700.1718304996.2",
          "SeriesDescription": "Test Series",
          "SeriesNumber": 7,
          "SeriesTime": "111813.948",
          "Modality": "XA",
          "instances": [
            {
              "metadata": {
                ...
                "ImageType": [
                  "ORIGINAL",
                  "PRIMARY",
                  "SINGLE PLANE",
                  "SINGLE A"
                ],
                "Modality": "XA",
                "SOPInstanceUID": "1.2.300.0.7230010.3.1.4.123.7700.1718304996.1.1",
                "SeriesInstanceUID": "1.2.300.0.7230010.3.1.3.4123831.7700.1718304996.2",
                "StudyInstanceUID": "1.2.300.0.7230010.3.1.2.4331.7700.211718304996.3",
                "WindowCenter": 127.75,
                "WindowWidth": 255.75,
                "SeriesDate": "20201104",
                "AcquisitionDate": "20170116",
                "AcquisitionTime": "111813.948",
                "NumberOfFrames": 48,
                "FrameTime": 66.667
              },
              "url": "wadors:http://localhost:3001?frame=1"
            },
```

Use this node service that returns frames from a dicom file from file system
```
const express = require("express");
const fs = require("fs");
const dicomParser = require("dicom-parser");
const cors = require("cors");

const app = express();
const PORT = process.env.PORT || 3001;

app.use(cors());

function getDicomFrame(filePath, frameNumber) {
  const dicomFile = fs.readFileSync(filePath);
  const dataSet = dicomParser.parseDicom(dicomFile);

  const numberOfFrames = dataSet.intString("x00280008");
  const pixelDataElement = dataSet.elements.x7fe00010;
  const frameSize = pixelDataElement.length / numberOfFrames;
  const frameOffset = (frameNumber - 1) * frameSize;

  return dicomFile.slice(
    pixelDataElement.dataOffset + frameOffset,
    pixelDataElement.dataOffset + frameOffset + frameSize
  );
}

app.get("/", (req, res) => {
  const frameNumber = parseInt(req.query.frame);
  console.debug('Retrieving frame number:', frameNumber);

  try {
    const dicomFilePath = "./multiframe.dcm";
    const frame = getDicomFrame(dicomFilePath, frameNumber);
    res.setHeader("Content-Type", "application/octet-stream");
    res.send(frame);
  } catch (error) {
    res.status(400).send({ error: error.message });
  }
});

app.listen(PORT, () => {
  console.log(`Server is running on port ${PORT}`);
});

```

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/default/src/DicomJSONDataSource/index.js
```diff
@@ -258,19 +262,19 @@ function createDicomJSONApi(dicomJsonConfig) {
```
wayfarer3130, line 262: On the line just above const images = displaySet.images, do you mean to fetch images, or instances?  They aren't necessarily the same as the instances is defined for other display set types, while images is only defined for image containing types.  Also, instances can include the overall sop instance, without frame number, while images are supposed to be individual images with frame number.

extensions/default/src/DicomJSONDataSource/index.js
```diff
@@ -258,19 +262,19 @@ function createDicomJSONApi(dicomJsonConfig) {
...
 
-      instances.forEach(instance => {
+      displaySet.images.forEach(instance => {
         const NumberOfFrames = instance.NumberOfFrames;
-
         if (NumberOfFrames > 1) {
           for (let i = 0; i < NumberOfFrames; i++) {
+            /** In case there are multiple sop class we filter sop instances */
+            const images = series.instances
+              .filter(i => i && i.metadata && i.metadata.SOPInstanceUID === instance.SOPInstanceUID)
```
wayfarer3130, line 273: Please rename the filter variable i to not override the loop variable i.
Also, 
.filter(testInstance => testInstance?.metadata?.SOPInstanceUID === instance.SOPInstanceUID
is easier to read than the && version.

platform/core/src/classes/MetadataProvider.ts
```diff
@@ -22,6 +22,12 @@ class MetadataProvider {
       writable: false,
       value: new Map(),
     });
+    Object.defineProperty(this, 'imageUIDsByImageId', {
```
wayfarer3130, line 25: This seems equivalent to 
private readonly imageUIDSByImageId = new Map<string,...>()
This code was originally javascript based, but now that it is typescript, using the behaviour in typescript makes sense.

extensions/default/src/DicomJSONDataSource/index.js
```diff
@@ -258,19 +262,19 @@ function createDicomJSONApi(dicomJsonConfig) {
...
-      }
+      const series = study.series.find(s => s.SeriesInstanceUID === SeriesInstanceUID) || [];
 
-      instances.forEach(instance => {
+      displaySet.images.forEach(instance => {
         const NumberOfFrames = instance.NumberOfFrames;
-
         if (NumberOfFrames > 1) {
           for (let i = 0; i < NumberOfFrames; i++) {
+            /** In case there are multiple sop class we filter sop instances */
```
wayfarer3130, line 271: This comment doesn't make sense - you are referring to sop class, but looking at sop instance uid.

extensions/default/src/DicomJSONDataSource/index.js
```diff
@@ -258,19 +262,19 @@ function createDicomJSONApi(dicomJsonConfig) {
...
-        instances = series.instances;
-      }
+      const series = study.series.find(s => s.SeriesInstanceUID === SeriesInstanceUID) || [];
 
-      instances.forEach(instance => {
+      displaySet.images.forEach(instance => {
         const NumberOfFrames = instance.NumberOfFrames;
-
         if (NumberOfFrames > 1) {
           for (let i = 0; i < NumberOfFrames; i++) {
```
wayfarer3130, line 270: Can you re-implement this to deal with large number of frames and/or large number of sop instances in a series?  THe JSON data should be able to just reference an instance with NumberOfFrames in it, and the URL not having a per-frame value in it, and be able to find that easily.
That is, allowing either the representation where every frame is present in the JSON or just a single multiframe is present in the JSON should both be allowed, and should match up correctly, without needing to go over every single frame and match up to every single image instance by looking at every JSON image instance for each frame.

extensions/default/src/DicomJSONDataSource/index.js
```diff
@@ -258,19 +262,19 @@ function createDicomJSONApi(dicomJsonConfig) {
...
+            /** In case there are multiple sop class we filter sop instances */
+            const images = series.instances
+              .filter(i => i && i.metadata && i.metadata.SOPInstanceUID === instance.SOPInstanceUID)
+              .map(i => ({ ...(i.metadata || {}), url: i.url }));
```
wayfarer3130, line 274: This will copy the metadata multiple times, which can get very expensive for large multiframes.  I have examples with 340,000 frames in it, and those should still work successfully.

extensions/default/src/DicomJSONDataSource/index.js
igoroctaviano: This implementation should work for both. This was a change for an old project, do you think its a quick change to deal with large number of frames?

extensions/default/src/DicomJSONDataSource/index.js
igoroctaviano: If you have a multiframe series with mixed sop class you might have different sop instance uid. I'm filtering here.

extensions/default/src/DicomJSONDataSource/index.js
igoroctaviano: I wanted all instances from a multiframe json file instead of just one instance representing the multiframe. My use case had a json file with a single multiframe series but It had individual URL attributes per frame.

extensions/default/src/DicomJSONDataSource/index.js
wayfarer3130: If you have a multiframe series without mixed sop class you might have different sop instance uid - actually you WILL have different sop instance uid, one per instance object, so the whole thing is independent of sop class.

extensions/default/src/DicomJSONDataSource/index.js
```diff
@@ -258,19 +262,19 @@ function createDicomJSONApi(dicomJsonConfig) {
...
       const study = findStudies('StudyInstanceUID', StudyInstanceUID)[0];
-      const series = study.series.find(s => s.SeriesInstanceUID === SeriesInstanceUID);
-      let instances = displaySet.images;
-      if (series.instances.length > displaySet.images.length) {
-        instances = series.instances;
-      }
+      const series = study.series.find(s => s.SeriesInstanceUID === SeriesInstanceUID) || [];
 
-      instances.forEach(instance => {
+      displaySet.images.forEach(instance => {
```
wayfarer3130, line 267: This algorithm is basically finding the right series.instances value, according to sop  instance uid and possibly frame number, so if you just create a map from  sop instance uid to the series instance that is relevant for that sop instance, then you can go through the loop without doing the series.instances filter query on every image, which reduces it from O(n^2) to O(n)
Something like
const seriesSopMap = new Map()
series.instances.forEach(   instance => addToSeriesSopMap(instance.sopInstanceUID, instance) ) 

Then just use seriesSopMap instead of sereis.instances.filter - that is, 
const images = seriesSopMap.get(instance.SOPInstanceUID)
const imageId = getImageId( {   instance: { ...(images.length > 0 ? images[i] : instance), frame: i, config: dicomJsonConfig })

igoroctaviano: Updated!

extensions/default/src/DicomJSONDataSource/index.js
igoroctaviano: Updated!

extensions/default/src/DicomJSONDataSource/index.js
igoroctaviano: Updated!

extensions/default/src/DicomJSONDataSource/index.js
igoroctaviano: Updated!

extensions/default/src/DicomJSONDataSource/index.js
igoroctaviano: Code updated!

platform/core/src/classes/MetadataProvider.ts
igoroctaviano: Updated!

igoroctaviano:
@wayfarer3130 could you please review this? If any changes or clean-ups are needed, I'll take care of them based on your final feedback.

sedghi (approved):

PR #4338, opened 2024-08-16
fix(tools): check if seriesNumber is an undefined properly

IbrahimCSAE:
### Context

some tools were not rendering displayText if series number is 0, now we check properly for undefined.

### DIFF
--- a/extensions/cornerstone/src/utils/measurementServiceMappings/LivewireContour.ts
+++ b/extensions/cornerstone/src/utils/measurementServiceMappings/LivewireContour.ts
@@ -142,9 +142,7 @@ function getDisplayText(annotation, displaySet, customizationService) {
   const frameText = displaySet.isMultiFrame ? ` F: ${frameNumber}` : '';
 
   const { SeriesNumber } = displaySet;
-  if (SeriesNumber !== undefined) {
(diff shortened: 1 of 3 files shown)


extensions/cornerstone/src/utils/measurementServiceMappings/LivewireContour.ts
```diff
@@ -142,7 +142,7 @@ function getDisplayText(annotation, displaySet, customizationService) {
   const frameText = displaySet.isMultiFrame ? ` F: ${frameNumber}` : '';
 
   const { SeriesNumber } = displaySet;
-  if (SeriesNumber) {
+  if (SeriesNumber !== undefined) {
```
sedghi, line 145: How about

```
displayText.push(`S: ${SeriesNumber ?? ''}${instanceText}${frameText}`);
```

IbrahimCSAE: i like that, thats good

IbrahimCSAE: actually I tried it, doesn't work too, maybe if its 0 its also a problem

PR #4346, opened 2024-08-26
fix: 🐛 SeriesInstanceUID fallback + update retrieve metadata filtered to check for lazy

igoroctaviano:
- SeriesInstanceUID query param fallback to avoid breaking changes
- Update retrieve metadata filtered func to check enableStudyLazyLoad before merging results

### Context

The retrieveMetadataFiltered function logic was specific to lazy study load and would not work if it was disabled. If you had a configuration with disabled lazy study load then this function would not work properly because of the data format of lazy loaded studies.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


igoroctaviano:
@sedghi PR context added.

PR #4347, opened 2024-08-26
OHIF with Cornerstone3D 2.0

sedghi:
- **was not that hard migration at least for basic render nice**
- **try cs3d 2.0 in master**

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone-dicom-pmap/src/getSopClassHandlerModule.ts
```diff
@@ -119,6 +119,22 @@ function _getDisplaySetsFromSeries(
   return [displaySet];
 }
 
+const getRangeFromPixelData = (pixelData: Float32Array) => {
```
wayfarer3130, line 122: Does this need to include pixel padding range, and then exclude that from the range calculations?

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
```diff
@@ -72,22 +72,27 @@ function _getDisplaySetsFromSeries(
...
-  displaySet.getReferenceDisplaySet = () => {
-    const { displaySetService } = servicesManager.services;
-    const referencedDisplaySets = displaySetService.getDisplaySetsForSeries(
-      displaySet.referencedSeriesInstanceUID
+  const { displaySetService } = servicesManager.services;
+  const referencedDisplaySets = displaySetService.getDisplaySetsForSeries(
+    displaySet.referencedSeriesInstanceUID
+  );
+
+  if (!referencedDisplaySets || referencedDisplaySets.length === 0) {
```
wayfarer3130, line 80: Does this introduce a bug if one display set is already referenced, but a second one becomes available which would have been referenced had it been loaded?

extensions/cornerstone-dicom-rt/src/viewports/OHIFCornerstoneRTViewport.tsx
```diff
@@ -91,12 +97,31 @@ function OHIFCornerstoneRTViewport(props: withAppTypes) {
     });
   }, [viewportGrid]);
 
-  const hydrateRTDisplaySet = ({ rtDisplaySet, viewportId }) => {
-    commandsManager.runCommand('loadSegmentationDisplaySetsForViewport', {
-      displaySets: [rtDisplaySet],
-      viewportId,
-    });
-  };
+  const hydrateRTDisplaySet = useCallback(
```
wayfarer3130, line 100: Should this full function be a commands manager run command?  It feels like hydrating an RT display set might be done via a command/external function.

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -28,428 +29,401 @@ const STACK = 'stack';
...
-    // check if their imageIds are the same
-    if (foundDisplaySet.images?.length) {
-      for (let j = 0; j < foundDisplaySet.images.length; j++) {
-        if (foundDisplaySet.images[j].imageId !== prevDisplaySet.images[j].imageId) {
-          return false;
-        }
+    // Since we only have support for dynamic data in volume viewports, we should
+    // handle this case here and set the viewportType to volume if any of the
+    // displaySets are dynamic volumes
+    viewportOptions.viewportType = displaySets.some(
```
wayfarer3130, line 82: Can you encapsulate this into a function which can be replaced in the future?  Something which can handle WSI, video etc in the future.

wayfarer3130: What about letting the display set itself indicate the allowed viewport types.  That way the sop class handler can define exactly what is permitted, since it actually knows that information.

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
```diff
@@ -28,428 +29,401 @@ const STACK = 'stack';
...
+        unsubscribe();
+      };
+    }, [viewportId]);
 
-      if (onElementDisabled) {
-        onElementDisabled(viewportInfo);
+    useEffect(() => {
+      // handle the default viewportType to be stack
+      if (!viewportOptions.viewportType) {
+        viewportOptions.viewportType = STACK;
```
wayfarer3130, line 248: Grab the default type from the display set first, and then default to stack after that.

extensions/cornerstone/src/Viewport/Overlays/CustomizableViewportOverlay.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { vec3 } from 'gl-matrix';
 import PropTypes from 'prop-types';
 import { metaData, Enums, utilities } from '@cornerstonejs/core';
-import { ImageSliceData } from '@cornerstonejs/core/dist/esm/types';
+import { ImageSliceData } from '@cornerstonejs/core/types';
```
wayfarer3130, line 5: If this is used as a type, can you import as a type so it doesn't depend on the implementation?

extensions/cornerstone/src/components/CinePlayer/CinePlayer.tsx
```diff
@@ -111,7 +110,10 @@ function WrappedCinePlayer({
     return () => {
       cineService.setCine({ id: viewportId, isPlaying: false });
 
-      eventTarget.removeEventListener(Enums.Events.STACK_VIEWPORT_NEW_STACK, newDisplaySetHandler);
+      enabledVPElement.removeEventListener(
+        Enums.Events.VIEWPORT_NEW_IMAGE_SET,
+        newDisplaySetHandler
+      );
       enabledVPElement.removeEventListener(
```
wayfarer3130, line 117: Does it still need the new volume listener?  Seems to me it should have the new image set triggered for new volume as well.

extensions/cornerstone/src/components/ViewportWindowLevel/ViewportWindowLevel.tsx
```diff
@@ -17,8 +17,9 @@ const ViewportWindowLevel = ({
   const { cornerstoneViewportService } = servicesManager.services;
   const [windowLevels, setWindowLevels] = useState([]);
   const [cachedHistograms, setCachedHistograms] = useState({});
+  const [isLoading, setIsLoading] = useState(true);
 
-  /**
+  /**c
```
wayfarer3130, line 22: Not sure what the c is

extensions/cornerstone/src/components/ViewportWindowLevel/getViewportVolumeHistogram.ts
```diff
@@ -32,12 +28,13 @@ const getViewportVolumeHistogram = async (viewport, volume, options?) => {
...
-    scalarData = volume.getScalarData(middleTimePoint);
+    const targetTimePoint = volume.numTimePoints - 1; // or any other time point you need
+    scalarData = volume.voxelManager.getTimePointScalarData(targetTimePoint);
+  }
+
+  if (!scalarData?.length) {
+    return undefined;
   }
 
   const { dimensions, origin, direction, spacing } = volume;
```
wayfarer3130, line 40: Just below the range should exclude pixel padding values, for which you should pass in the pixel padding range data.

extensions/cornerstone/src/components/WindowLevelActionMenu/getWindowLevelActionMenu.tsx
```diff
@@ -24,6 +30,12 @@ export function getWindowLevelActionMenu({
       return { [displaySet.Modality]: presets[displaySet.Modality] };
     });
 
+  const modalities = displaySets.map(displaySet => displaySet.Modality);
```
wayfarer3130, line 33: This should reduce the set of modalities by excluding the non-WL modalities, and then if the modalities set still contains any element, then window level is permitted.  That is,
SEG => not permitted
SEG, SM => not permitted
SEG, CR => permitted
SEG, CR, OT => permitted
You can do this with an every statement rather than a some.

extensions/cornerstone/src/contextProviders/ViewportActionCornersProvider.tsx
```diff
@@ -35,8 +35,8 @@ export const ViewportActionCornersContext = createContext(DEFAULT_STATE);
 export function ViewportActionCornersProvider({ children, service }) {
   const viewportActionCornersReducer = (state, action) => {
     switch (action.type) {
-      case 'SET_ACTION_COMPONENT': {
-        const { viewportId, id, component, location, indexPriority = 0 } = action.payload;
+      case 'ADD_ACTION_COMPONENT': {
+        const { viewportId, id, component, location, indexPriority } = action.payload;
```
wayfarer3130, line 39: The location should come from the component and NOT the add command.  The reason is that one can then update the component to work in a different location if desired, and it will just flow through properly.  For example, a window level component which is down at the bottom of the screen is perfectly fine, rather than one at the right.

extensions/cornerstone/src/getToolbarModule.tsx
```diff
@@ -20,9 +45,32 @@ export default function getToolbarModule({ commandsManager, servicesManager }: w
   return [
     // functions/helpers to be used by the toolbar buttons to decide if they should
     // enabled or not
+    {
+      name: 'evaluate.not.sm',
```
wayfarer3130, line 49: This is quite ugly - how about
evaluate.viewport.supported
evaluate({ viewportId, button }) => {
      const viewport = ...
      if( button.unsupportedTypes.includes(viewport?.type) ) return getDisabledState().

Then, you can get rid of tools not supported by video viewport as well as the evaluate actions.

extensions/cornerstone/src/getToolbarModule.tsx
```diff
@@ -176,14 +220,7 @@ export default function getToolbarModule({ commandsManager, servicesManager }: w
...
         const viewport = cornerstoneViewportService.getCornerstoneViewport(viewportId);
-
-        if (viewport?.type === 'volume3d') {
-          return {
-            disabled: true,
-            className: '!text-common-bright ohif-disabled',
-            disabledText: disabledText ?? 'Not available on the current viewport',
-          };
-        }
+        return viewport?.type === 'volume3d' ? getDisabledState(disabledText) : undefined;
```
wayfarer3130, line 223: This one can be done just like the evaluate viewport not supported or viewport supported.

extensions/cornerstone-dicom-pmap/src/getSopClassHandlerModule.ts
sedghi: is there any pixel padding in pmap?

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
sedghi: i don't think so since it is filtering based on `if (addedDisplaySet.SeriesInstanceUID === displaySet.referencedSeriesInstanceUID) {` ?

extensions/cornerstone-dicom-rt/src/viewports/OHIFCornerstoneRTViewport.tsx
sedghi: Done

extensions/cornerstone/src/components/CinePlayer/CinePlayer.tsx
sedghi: Seems like i forgot to migrate all viewports to this event, so cornerstone3D 3.0 candidate

extensions/cornerstone/src/components/ViewportWindowLevel/getViewportVolumeHistogram.ts
sedghi: Seems like we need a proper fix for padding values here and there

extensions/cornerstone/src/contextProviders/ViewportActionCornersProvider.tsx
sedghi: I'm not sure about this, let's talk

extensions/cornerstone/src/getToolbarModule.tsx
sedghi: I thought i had it like this, hmm weird

extensions/cornerstone/src/hooks/useActiveViewportSegmentationRepresentations.ts
```diff
@@ -0,0 +1,219 @@
+import { useState, useEffect } from 'react';
+import debounce from 'lodash.debounce';
+import { roundNumber } from '@ohif/core/src/utils';
+import {
+  SegmentationData,
+  SegmentationRepresentation,
+} from '../services/SegmentationService/SegmentationService';
+
+const excludedModalities = ['SM', 'OT', 'DOC', 'ECG'];
```
wayfarer3130, line 9: Can you define this somewhere else so we use a consistent list?  It seems to vary all over the place - no idea why OT is excluded.

extensions/cornerstone/src/hps/mpr.ts
```diff
@@ -108,6 +116,15 @@ export const mpr: Types.HangingProtocol.Protocol = {
                   syncColormap: true,
                 },
               },
+              {
```
wayfarer3130, line 119: Can you define this as a const and then reference it?

extensions/cornerstone/src/panels/PanelSegmentation.tsx
```diff
@@ -0,0 +1,286 @@
...
+      segmentationService.addSegment(segmentationId);
+    },
+
+    onSegmentClick: (segmentationId, segmentIndex) => {
+      segmentationService.setActiveSegment(segmentationId, segmentIndex);
+      segmentationService.jumpToSegmentCenter(segmentationId, segmentIndex);
+    },
+
+    onSegmentEdit: (segmentationId, segmentIndex) => {
+      const segmentation = segmentationService.getSegmentation(segmentationId);
```
wayfarer3130, line 43: This bit should be in a commands module so it can be attached to a keyboard shortcut or the like.

extensions/cornerstone/src/panels/PanelSegmentation.tsx
```diff
@@ -0,0 +1,286 @@
...
+      callInputDialog(uiDialogService, label, (label, actionId) => {
+        if (label === '') {
+          return;
+        }
+
+        segmentationService.setSegmentLabel(segmentationId, segmentIndex, label);
+      });
+    },
+
+    onSegmentationEdit: segmentationId => {
```
wayfarer3130, line 61: If you are going to have two very similar names, you need a comment to explain that one is segmentation overall which contains segments, and the other is an individual segment.  It took me a moment to remember that.
Also, this one should also be in a commands module.

extensions/cornerstone/src/panels/PanelSegmentation.tsx
```diff
@@ -0,0 +1,286 @@
...
+      callInputDialog(uiDialogService, label, (label, actionId) => {
+        if (label === '') {
+          return;
+        }
+
+        segmentationService.addOrUpdateSegmentation(segmentationId, { label: label });
+      });
+    },
+
+    onSegmentColorClick: (segmentationId, segmentIndex) => {
```
wayfarer3130, line 79: Commands module.

extensions/cornerstone/src/panels/PanelSegmentation.tsx
```diff
@@ -0,0 +1,286 @@
...
+          return;
+        }
+
+        const color = [newRgbaColor.r, newRgbaColor.g, newRgbaColor.b, newRgbaColor.a * 255.0];
+        segmentationService.setSegmentColor(viewportId, segmentationId, segmentIndex, color);
+      });
+    },
+
+    onSegmentDelete: (segmentationId, segmentIndex) => {
+      segmentationService.removeSegment(segmentationId, segmentIndex);
```
wayfarer3130, line 100: Ideally even really small ones like this would have their own commands module item so that it is possible to redirect/replace them

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -836,22 +806,52 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
       return { properties, volumeId };
     });
 
+    // For SEG and RT viewports
+    this._processOverlaysForViewport(viewport);
+
+    // Always load the reference volume if the displaySet is an overlay
```
wayfarer3130, line 812: If this is still true, then do you need to do something here?  If not, can you remove

extensions/cornerstone/src/hooks/useActiveViewportSegmentationRepresentations.ts
sedghi: OT Excluded for now

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -863,7 +863,7 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
     });
   }
 
-  private _handleOverlays(viewport: Types.IStackViewport | Types.IVolumeViewport) {
+  private _processOverlaysForViewport(viewport: Types.IStackViewport | Types.IVolumeViewport) {
```
wayfarer3130, line 866: Overlays has an awful lot of meetings - is this really about RT/Seg - if there is a better name for that sort of thing it would be helpful to use it.

wayfarer3130: What about processExtraDIsplaySetsForViewport

extensions/cornerstone/src/services/ViewportService/Viewport.ts
wayfarer3130, line 1: Since You are touching this, is there any way you could rename it to OhifViewport or something like that which distinguishes it from CS3D viewports?  I really had using a very common name like that overloaded.

extensions/cornerstone/src/services/ViewportActionCornersService/ViewportActionCornersService.ts
```diff
@@ -33,34 +33,34 @@ class ViewportActionCornersService extends PubSubService {
...
+    addComponents: addComponentsImplementation,
+    clear: clearComponentsImplementation,
   }): void {
     if (getStateImplementation) {
       this.serviceImplementation._getState = getStateImplementation;
     }
-    if (setComponentImplementation) {
-      this.serviceImplementation._setComponent = setComponentImplementation;
+    if (addComponentImplementation) {
+      this.serviceImplementation._addComponent = addComponentImplementation;
```
wayfarer3130, line 44: Since you are tweaking this anyways, can you just make it this.setComponent = setComponentImplementation and skip the extra level of indirection.  Having that doesn't really help anything, and not doing it means the service implementation can directly call this.X for any of the other services.

extensions/cornerstone/src/stores/presentationUtils.ts
```diff
@@ -0,0 +1,42 @@
+const JOIN_STR = '&';
+
+// The default lut presentation id if none defined
+const DEFAULT_STR = 'default';
+
+// This code finds the first unique index to add to the presentation id so that
```
wayfarer3130, line 6: Can you make this an official doc so it gets included as documentation?

extensions/cornerstone/src/stores/usePositionPresentationStore.ts
wayfarer3130, line 1: This is quite a bit cleaner - thank you for doing that.

platform/ui/src/contextProviders/ViewportGridProvider.tsx
```diff
@@ -513,4 +524,6 @@ ViewportGridProvider.propTypes = {
   service: PropTypes.instanceOf(ViewportGridService).isRequired,
 };
 
-export const useViewportGrid = () => useContext(ViewportGridContext);
+// Update the useViewportGrid hook
+export const useViewportGrid = (): [AppTypes.ViewportGrid.State, ViewportGridApi] =>
```
wayfarer3130, line 528: Thank you for adding types on this - it was not great before not having that.
In the previous version, the viewport grid set state methods weren't done immediately - that is,
viewportGrid.setState({valueKey: newValue} )
state.valueKey
didn't necessarily return the same value as newValue - eventualy that will be true, but it isn't necessarily immediate.  That should be a warning in the docs that indicate that the state may not be reflected immediately - I've seen several bugs in the code caused by that issue.  Instead, the way to handle it is to make all the changes at once OR to react to changes.

platform/ui-next/src/components/Sonner/Sonner.tsx
```diff
@@ -18,16 +18,20 @@ const Toaster = ({ ...props }: ToasterProps) => {
...
-          toast:
-            'group toast group-[.toaster]:bg-popover group-[.toaster]:text-foreground group-[.toaster]:border-background group-[.toaster]:shadow-lg',
-          info: 'group-[.toast] group-[.toaster]:!bg-popover group-[.toaster]:!text-foreground group-[.toaster]:!border-background',
-          error: 'group-[.toaster]:!border-background',
-          success: 'group-[.toaster]:!border-background',
-          warning: 'group-[.toaster]:!border-background',
-          description: 'group-[.toast]:text-secondary-foreground',
-          actionButton: 'group-[.toast]:bg-primary group-[.toast]:text-primary-foreground',
-          cancelButton: 'group-[.toast]:bg-muted group-[.toast]:text-muted-foreground',
+        //   classNames: {
```
wayfarer3130, line 21: Can you remove this now?

extensions/cornerstone-dicom-pmap/src/getSopClassHandlerModule.ts
sedghi: Seems like pmap does not have pixel padding

extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
sedghi: Added issue https://github.com/OHIF/Viewers/issues/4448

extensions/cornerstone/src/utils/measurementServiceMappings/Length.ts
```diff
@@ -18,9 +21,16 @@ const Length = {
...
-    const { annotation, viewportId } = csToolsEventDetail;
+    const { annotation } = csToolsEventDetail;
     const { metadata, data, annotationUID } = annotation;
 
+    const isLocked = getIsLocked(annotationUID);
+    const isVisible = getIsVisible(annotationUID);
+    const colorString = config.style.getStyleProperty('color', { annotationUID });
+
+    // color string is like 'rgb(255, 255, 255)' we need them to be in RGBA array [255, 255, 255, 255]
+    const color = colorString.replace('rgb(', '').replace(')', '').split(',').map(Number);
```
wayfarer3130, line 32: Even though this is trivial, could you define this in a utility?  It just feels like something which will be needed elsewhere.

extensions/cornerstone/src/utils/measurementServiceMappings/LivewireContour.ts
```diff
@@ -142,19 +147,26 @@ function getDisplayText(annotation, displaySet, customizationService) {
   const frameText = displaySet.isMultiFrame ? ` F: ${frameNumber}` : '';
```
wayfarer3130, line 147: This bit of code for the series text seems like it is repeated over and over again.  Can you add it to a parent class?

extensions/cornerstone/src/utils/measurementServiceMappings/RectangleROI.ts
```diff
@@ -13,8 +15,10 @@ const RectangleROI = {
     getValueTypeFromToolType,
     customizationService
   ) => {
-    const { annotation, viewportId } = csToolsEventDetail;
+    const { annotation } = csToolsEventDetail;
```
wayfarer3130, line 18: It is a code smell that a modification to an external library has the same code modified in the same way across multiple files.  Generally that means that a super class should be created that has most of the shared code, with only the small/specific bit which is needed to be changed modified.  That would also make it much easier to create new annotation types.  I'm not asking for that to be done now, but please have an issue for it.

extensions/cornerstone/src/utils/measurementServiceMappings/utils/getIsVisible.ts
wayfarer3130, line 1: I like encapsulating some of these small bits of code which occasionally change as functions so that one doesn't need to parse what is happening outside that, and one can change everywhere it happens in a single function change.

extensions/default/src/hangingprotocols/hpMNGrid.ts
```diff
@@ -99,6 +122,17 @@ const hpMN: Types.HangingProtocol.Protocol = {
           viewportOptions: {
             toolGroupId: 'default',
             allowUnmatchedView: true,
+            syncGroups: [
```
wayfarer3130, line 125: Please extract a const that is shared between hanging protocols.

extensions/tmtv/src/utils/hpViewports.ts
```diff
@@ -24,6 +24,15 @@ const ctAXIAL = {
           syncColormap: true,
         },
       },
+      {
```
wayfarer3130, line 27: Use const value imported from a shared location for this sync value.

modes/longitudinal/src/index.ts
```diff
@@ -192,7 +177,7 @@ function modeFactory({ modeConfiguration }) {
             id: ohif.layout,
             props: {
               leftPanels: [tracked.thumbnailList],
-              rightPanels: [dicomSeg.panel, tracked.measurements],
+              rightPanels: [cornerstone.segmentation, tracked.measurements],
```
wayfarer3130, line 180: Can the order be swapped here?  It seems to me that measurements are more common that segmentations, so starting showing the measurements first would be better.

modes/segmentation/src/customizations.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { Button, useSegmentationTableContext } from '@ohif/ui-next';
+
+export const performCustomizations = customizationService => {
+  // customizationService.addModeCustomizations([
```
wayfarer3130, line 5: Is this still items to add back in?

wayfarer3130:
I think we are going to have to revisit the handling of viewport type coming from the display set - not including that change breaks the display of video and sm in the CS3D viewports.

platform/core/src/services/DisplaySetService/DisplaySetService.ts
```diff
@@ -204,7 +204,7 @@ export default class DisplaySetService extends PubSubService {
     }
 
     // If array of instances => One instance.
-    const displaySetsAdded = [];
+    const displaySetsAdded = [] as DisplaySet[];
```
wayfarer3130, line 207: I generally prefer the following:
new Array<DisplaySet>()
to the casting.  Not sure which form we should use.

platform/ui-next/src/components/DataRow/DataRow.tsx
```diff
@@ -7,193 +7,274 @@ import {
...
+  title: string;
+  onRename: () => void;
+  //
+  onDelete: () => void;
+  //
+  colorHex?: string;
+  onColor: () => void;
 }
 
 const DataRow: React.FC<DataRowProps> = ({
```
wayfarer3130, line 37: Can you add documentation as to what this is?  DataRow is pretty general, but this class seems pretty specific as to what it displays.  I certainly would have no idea as to how to use DataRow without reading the code and figuring that out is fairly hard.

platform/ui-next/src/components/ScrollArea/ScrollArea.tsx
```diff
@@ -2,23 +2,63 @@ import * as React from 'react';
 import * as ScrollAreaPrimitive from '@radix-ui/react-scroll-area';
 
 import { cn } from '../../lib/utils';
+import { Icons } from '../Icons';
+
+interface ScrollAreaProps extends React.ComponentPropsWithoutRef<typeof ScrollAreaPrimitive.Root> {
+  showArrows?: boolean;
+}
 
 const ScrollArea = React.forwardRef<
```
wayfarer3130, line 11: Can you add docs as to what this does?

platform/ui-next/src/components/SegmentationTable/AddSegmentRow.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { Button, Icons } from '@ohif/ui-next';
+import { useSegmentationTableContext } from './SegmentationTableContext';
+
+export const AddSegmentRow: React.FC<{
```
wayfarer3130, line 5: Docs

wayfarer3130: Really all I'm looking for is just a comment indicating  that this is used to add a segment row to the segmentation table - I know I can get most of that from the path, so perhaps a tiny bit more.

platform/ui-next/src/components/DataRow/types.ts
```diff
@@ -0,0 +1,13 @@
+export interface DataItem {
```
wayfarer3130, line 1: type probably - this looks like a simple structure, not an interface for a class.

platform/ui-next/src/components/DataRow/types.ts
```diff
@@ -0,0 +1,13 @@
...
+  id: number;
+  title: string;
+  description: string;
+  optionalField?: string;
+  colorHex?: string;
+  details?: string;
+}
+
+export interface ListGroup {
```
wayfarer3130, line 10: Ditto

extensions/cornerstone/src/contextProviders/ViewportActionCornersProvider.tsx
sedghi: I decided to do this in next version not this one https://github.com/OHIF/Viewers/issues/4457

extensions/cornerstone/src/services/ViewportService/Viewport.ts
sedghi: I think it should be called ViewportInfo instead of OHIF Viewport, since we have OHIFCornerstoneViewport too

extensions/cornerstone/src/services/ViewportActionCornersService/ViewportActionCornersService.ts
sedghi: Let's plan to clean this up in the next version

platform/ui/src/contextProviders/ViewportGridProvider.tsx
sedghi: I haven't seen anything delay wise for these Zustand stores yet. Let's open an issue when we encounter it.

extensions/cornerstone/src/utils/measurementServiceMappings/Length.ts
sedghi: It is not implemented so i'm removing it for now

extensions/cornerstone/src/utils/measurementServiceMappings/LivewireContour.ts
sedghi: I initially planned to create a parent class, but things got hectic. Let's do it in the next release.https://github.com/OHIF/Viewers/issues/4458

modes/longitudinal/src/index.ts
sedghi: We recently reordered it. I think we should ensure that panels switch their active status if they are not in the active panel context. For example, if the segmentation panel is open and a measurement is added, we should automatically switch to the measurement panel, and vice versa. https://github.com/OHIF/Viewers/issues/4459

PR #4356, opened 2024-09-02
fix(sr): load existing point, if there is 2nd point in renderableData (Fix rotation in arrow annotation)

Joeycho:
This is related with PR in cornerstone3D:  https://github.com/cornerstonejs/cornerstone3D/pull/1450/files


Joeycho:
@sedghi I think I did it :)

Joeycho:
https://github.com/cornerstonejs/cornerstone3D/pull/1450

@sedghi thanks for approval, I think the above PR should be merged as well in cornerstone.

PR #4365, opened 2024-09-10
fix(hp-presets): select the active displaySet when toggling an HP preset, not a random one from the series panel

IbrahimCSAE:
### Context

This makes it so when a preset is selected it chooses the active displayset.


sedghi (approved):
Nice

PR #4367, opened 2024-09-11
feat(toolbar): enable extensions to change toolbar button sections

pedrokohler:
### Context

Currently, it's not possible to modify the primary toolbar button section with an extension. There are a few scenarios in which one could want to do that. For example:

1. It's possible you want to add a button to all modes at once by using your extension.
2. You might not have a custom mode in your project, and the only thing you want to do is add a button to your primary toolbar button section

The reason it's not possible to do that today is the following:

The mode's toolbar operations override the extensions', because the `onModeEnter` method of the mode runs after the extensions' one:

image

Since OHIF's pre-existing modes, such as longitudinal, tmtv, etc, create button sections from scratch, instead of adding buttons to a button section, it's not possible to add anything to the primary button section from the extension, since it gets overwritten later on.

### Changes & Results

With these changes, one can now add buttons to every mode by using the extensions `onModeEnter` to alter the button section. See an example:

Before (extension has no bearing in the toolbar):
image

After (extension can add buttons to the toolbar):
image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/modes/basic-dev-mode/src/index.ts
+++ b/modes/basic-dev-mode/src/index.ts
@@ -91,17 +91,14 @@ function modeFactory({ modeConfiguration }) {
 
       toolGroupService.createToolGroupAndAddTools('default', tools);
 
-      const buttonSectionId = 'primary';
-      const extensionsToolbarSection = toolbarService.getButtonSection(buttonSectionId) ?? [];
       toolbarService.addButtons(toolbarButtons);
-      toolbarService.createButtonSection(buttonSectionId, [
+      toolbarService.createButtonSection('primary', [
         'MeasurementTools',
         'Zoom',
         'WindowLevel',
         'Pan',
         'Layout',
         'MoreTools',
-        ...extensionsToolbarSection.map(button => button.id),
       ]);
     },
     onModeExit: ({ servicesManager }: withAppTypes) => {
--- a/modes/basic-dev-mode/src/index.ts
+++ b/modes/basic-dev-mode/src/index.ts
@@ -91,14 +91,17 @@ function modeFactory({ modeConfiguration }) {
 
       toolGroupService.createToolGroupAndAddTools('default', tools);
 
+      const buttonSectionId = 'primary';
+      const extensionsToolbarSection = toolbarService.getButtonSection(buttonSectionId) ?? [];
       toolbarService.addButtons(toolbarButtons);
-      toolbarService.createButtonSection('primary', [
+      toolbarService.createButtonSection(buttonSectionId, [
         'MeasurementTools',
         'Zoom',
         'WindowLevel',
         'Pan',
         'Layout',
         'MoreTools',
+        ...extensionsToolbarSection.map(button => button.id),
       ]);
     },
     onModeExit: ({ servicesManager }: withAppTypes) => {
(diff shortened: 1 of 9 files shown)


IbrahimCSAE (approved):
This PR makes sense to me, unless i'm missing something. It's a nice fix. I also tried locally and for example, if in your extensions 'onModeEnter' you define something like:

```js
   toolbarService.createButtonSection('primary', ['CornerstoneButton']);
    toolbarService.addButtons([
      {
        id: 'CornerstoneButton',
        uiType: 'ohif.radioGroup',
        props: {
          icon: 'tool-zoom',
          label: 'Zoom',
          commands: [
            {
              commandName: 'setToolActiveToolbar',
              commandOptions: {
                toolGroupIds: ['default', 'mpr', 'SRToolGroup', 'volume3d'],
              },
            },
          ],
        },
      },
    ]);
```

It seems that later on when `mode.OnModeEnter()` is called, that also calls the `toolbarServices.createButtonSection()`, which does set the buttons but also overrides the existing section at the same time

```js
  createButtonSection(key, buttons) {
    this.state.buttonSections[key] = buttons;
    this._broadcastEvent(this.EVENTS.TOOL_BAR_MODIFIED, { ...this.state });
  }
```
I thought we could try using a different `buttonSectionId` for our section like `cornerstoneSection`, but the top toolbar uses `primary` by default so it won't appear up there but will be in the state,

however, do you think this is a cleaner fix? I think this way no one needs to modify their modes or add anything extra.

```
  createButtonSection(key, buttons) {
    if (this.state.buttonSections[key]) {
      this.state.buttonSections[key].push(...buttons);
    } else {
      this.state.buttonSections[key] = buttons;
    }
    this._broadcastEvent(this.EVENTS.TOOL_BAR_MODIFIED, { ...this.state });
  }
```
This appends to the existing section if it exists, so if a mode finds that primary exists, it will add on top of it, and it should work fine since the toolbar is always reset when you enter a brand new mode.

modes/basic-dev-mode/src/index.ts
```diff
@@ -91,14 +91,17 @@ function modeFactory({ modeConfiguration }) {
...
       toolbarService.addButtons(toolbarButtons);
-      toolbarService.createButtonSection('primary', [
+      toolbarService.createButtonSection(buttonSectionId, [
         'MeasurementTools',
         'Zoom',
         'WindowLevel',
         'Pan',
         'Layout',
         'MoreTools',
+        ...extensionsToolbarSection.map(button => button.id),
```
sedghi, line 104: i'm not sure what it means to have toolbar section added to "tools"

pedrokohler: This line adds the button ids that were added previously by the extension but that will be removed by the mode if we don't re-add them.

IbrahimCSAE: Related to https://github.com/ImagingDataCommons/ViewersV3/pull/67

sedghi: Do it in toolbarservice, not in the mode

sedghi (approved):

PR #4372, opened 2024-09-13
[WIP] feat(ui): interactive onboarding guides

IbrahimCSAE:
### Context

Wip

### DIFF
--- a/platform/app/src/App.tsx
+++ b/platform/app/src/App.tsx
@@ -24,7 +24,7 @@ import {
   UserAuthenticationProvider,
   ToolboxProvider,
 } from '@ohif/ui';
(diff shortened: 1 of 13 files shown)


salimkanoun:
Hi there, be carefull you are linking intro.js which is AGPL licence,
It will make all OHIF AGPL : https://opensource.stackexchange.com/questions/1640/if-im-using-a-gpl-3-library-in-my-project-can-i-license-my-project-under-mit-l

IbrahimCSAE:
> Hi there, be carefull you are linking intro.js which is AGPL licence, It will make all OHIF AGPL : https://opensource.stackexchange.com/questions/1640/if-im-using-a-gpl-3-library-in-my-project-can-i-license-my-project-under-mit-l

Thanks for pointing that out, we will look into it @sedghi

PR #4373, opened 2024-09-14
fix(tmtv):Change volume unit from mm³ to cm³ in ROI thresholding display

JeffreyWW:
### Context
In the getLesionStats function of extensions/tmtv/src/commandsModule.ts, the volume is calculated as voxelCount * spacing[0] * spacing[1] * spacing[2] * 1e-3. This 1e-3 means that the final unit should be cubic centimeters rather than millimeters. Additionally, the final TMTV unit is in mL, which corresponds correctly.

### DIFF
--- a/extensions/tmtv/src/commandsModule.ts
+++ b/extensions/tmtv/src/commandsModule.ts
@@ -381,7 +381,7 @@ const commandsModule = ({ servicesManager, commandsManager, extensionManager }: 
           segmentationValues
             .map((k) => (k - mean) ** 2)
             .reduce((acc, curr) => acc + curr, 0) / voxelCount),
-        volume: voxelCount * spacing[0] * spacing[1] * spacing[2] * 1e-3,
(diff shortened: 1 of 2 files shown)


sedghi:
@salimkanoun

JeffreyWW:
> mhhh
test222
I agree with your point, but what I was addressing is not the unit of TMTV, but the unit for the segment’s volume. I noticed that in the calculation, it multiplies by 1e-3 at the end, so I changed it to cubic centimeters.

salimkanoun:
Yes sure but maybe it is better to remove the 10-3 rather than changing the unit

JeffreyWW:
> Yes sure but maybe it is better to remove the 10-3 rather than changing the unit

Now I have done this. Thank you for your suggestion.

salimkanoun:
Thanks ! Sounds good !

sedghi:
So i guess we close this?

PR #4410, opened 2024-10-10
fix(auth): oidc-react-issue

JamesAPetts:
### Context

In the OpenIdConnectRoutes there was a component which was defined as asynchronous, which causes issues in react.

This caused the following error before redirect: https://react.dev/errors/31?invariant=31&args%5B%5D=%5Bobject%20Promise%5D

### Changes & Results

To prevent this issue we instead don't await the redirect promise and just return null. Redirect works as expected, but React doesn't complain about its input

### Testing

Try an OpenIdConnect setup and you will not get a React error prior to redirect after this commit.


sedghi (approved):
Thanks

PR #4420, opened 2024-10-14
fix(tmtv): prevent fusion row in tmtv from getting inverted unexpectedly 

IbrahimCSAE:
### context

fixes this bug:

CleanShot 2024-10-14 at 14 10 59@2x


extensions/tmtv/src/utils/hpViewports.ts
```diff
@@ -128,6 +128,7 @@ const ptAXIAL = {
         target: false,
         options: {
           syncColormap: false,
+          syncInvertState: false,
```
sedghi, line 131: I remember this caused something else to get broken, can you see git blame who removd it and why?

IbrahimCSAE: ok, but for some reason if its not set, it defaults to true, why causes that stupid bug that has been there for few months now in tmtv

sedghi (approved):

PR #4501, opened 2024-11-14
feat: Start using group filtering to define measurements table layout

wayfarer3130:
### Context

The measurements service already has a few groups of measurement types, it would be nice to have more groups of types and allow them to be listed in various ways in the measurements panel.

This PR is a start of how the grouping might work, which includes some changes in the core grouping code.  However, it doesn't include any enhanced features yet.

### Changes & Results

* Add measurements filter option to getMeasurements and clearMeasurements
* Move commands for measurements into a command module
* Use the filtering for listing studies etc in the measurements panel
* Added an "Untracked" measurements section
* Fixed probe etc non-measurements to go into a separate section as designed

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


platform/core/src/utils/measurementFilters.ts
```diff
@@ -0,0 +1,54 @@
+export function filterTracked(trackedStudy: string, trackedSeries) {
```
sedghi, line 1: some are obvious some are not, can you add docs for non obvious ones at least? filterOr, filterUnless,

extensions/cornerstone-dicom-sr/src/commandsModule.ts
```diff
@@ -42,8 +43,57 @@ const _generateReport = (measurementData, additionalFindingTypes, options = {})
 
 const commandsModule = (props: withAppTypes) => {
   const { servicesManager } = props;
-  const { customizationService } = servicesManager.services;
+  const { customizationService, measurementService, viewportGridService, uiDialogService } =
+    servicesManager.services;
   const actions = {
+    jumpToMeasurement: ({ uid }) => {
```
sedghi, line 49: These don't belong here; they are generic measurement manipulations (jump, remove, rename) and should be in the cornerstone extension.

extensions/cornerstone-dicom-sr/src/commandsModule.ts
```diff
@@ -42,8 +43,57 @@ const _generateReport = (measurementData, additionalFindingTypes, options = {})
...
+
+    changeColorMeasurement: ({ uid }) => {
+      const { color } = measurementService.getMeasurement(uid);
+      const rgbaColor = {
+        r: color[0],
+        g: color[1],
+        b: color[2],
+        a: color[3] / 255.0,
+      };
+      colorPickerDialog(uiDialogService, rgbaColor, (newRgbaColor, actionId) => {
```
sedghi, line 79: This is misleading because we don't support re-color. We should consider throwing an error to indicate it's not supported yet. I think this was my mistake.

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
```diff
@@ -29,9 +31,18 @@ function PanelMeasurementTableTracking({
   const [displayStudySummary, setDisplayStudySummary] = useState(
     DISPLAY_STUDY_SUMMARY_INITIAL_VALUE
   );
+  const initialTrackedFilter = trackedStudy
+    ? filterTracked(trackedStudy, trackedSeries)
+    : filterAny;
+  const [measurementFilters, setMeasurementFilters] = useState({
+    measurementFilter: initialTrackedFilter,
+    untrackedFilter: filterNot('measurementFilter'),
```
sedghi, line 39: filterNot('measurementFilter') is confusing, i'm not sure what is it

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -129,15 +117,31 @@ export default function PanelMeasurementTable({
...
-            onToggleVisibility={toggleVisibilityMeasurement}
-            onToggleLocked={toggleLockMeasurement}
-            onRename={renameMeasurement}
-            // onColor={changeColorMeasurement}
+            {...onArgs}
+          >
+            <MeasurementTable.Header>
+              <>
+                <div>Hello World</div>
+              </>
```
sedghi, line 128: typo?

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -129,15 +117,31 @@ export default function PanelMeasurementTable({
...
+          <MeasurementTable
+            key="untracked"
+            data={untrackedFindings}
+            title="Untracked Findings"
+            {...onArgs}
           >
+            <MeasurementTable.Header>
+              <>
+                <span>Untracked</span>
+              </>
```
sedghi, line 143: do you need <> </> ?

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -32,69 +42,50 @@ export default function PanelMeasurementTable({
...
+    setMeasurements(
+      displayMeasurements.filter(item => !additionalFilter(item) && trackedFilter(item))
+    );
+    setAdditionalFindings(
+      displayMeasurements.filter(item => additionalFilter(item) && trackedFilter(item))
+    );
+    setUntrackedFindings(displayMeasurements.filter(untrackedFilter.bind(measurementFilters)));
+  }, [measurementFilters, displayMeasurements]);
+
+  const onArgs = {
```
sedghi, line 82: this looks like a nice pattern to me, cool

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -32,69 +42,50 @@ export default function PanelMeasurementTable({
...
-    });
+      if (callJump) {
+        commandsManager.runCommand('jumpToMeasurement');
+      }
+      commandsManager.runCommand(name, { uid });
+      if (callJump === null) {
+        onMeasurementItemClickHandler(uid, true);
+      }
+    };
   };
```
sedghi, line 62: We could implement a pipe operator in the `commandsManager` to allow chaining like this:

`commandsManager.run('toggleLockMeasurement|activateMeasurement|whatever')`

Otherwise, we would need to implement `bindCommand` everywhere, which involves custom logic for handling `callJump` and other parameters.

What do you think?

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -32,69 +42,50 @@ export default function PanelMeasurementTable({
...
+  useEffect(() => {
+    const { measurementFilter: trackedFilter, untrackedFilter } = measurementFilters;
+    setMeasurements(
+      displayMeasurements.filter(item => !additionalFilter(item) && trackedFilter(item))
+    );
+    setAdditionalFindings(
+      displayMeasurements.filter(item => additionalFilter(item) && trackedFilter(item))
+    );
+    setUntrackedFindings(displayMeasurements.filter(untrackedFilter.bind(measurementFilters)));
+  }, [measurementFilters, displayMeasurements]);
```
sedghi, line 80: You don't need a state for these; they can all be computed during the render call. If you remove `useEffect` and those `useState` hooks, it should work. The `useMeasurements` hook already abstracts the reactivity of the measurement service, so everything else can be computed states (not actual states).

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -129,15 +117,31 @@ export default function PanelMeasurementTable({
...
+          >
+            <MeasurementTable.Header>
+              <>
+                <div>Hello World</div>
+              </>
+            </MeasurementTable.Header>
+            <MeasurementTable.Body />
+          </MeasurementTable>
+        )}
+        {untrackedFindings.length > 0 && (
```
sedghi, line 133: I think this might upset some people because we're imposing an untracked view on them. I was hoping people could choose what to see based on the filters they set through the customization service. Here, we simply loop over those filters and create the view, regardless of the number.

<MeasurementTable.Header>

</MeasurementTable.Header>

extensions/cornerstone/src/panels/PanelMeasurement.tsx
wayfarer3130: When we need it, I'm going to implement a runCommands command which runs the specified commands.  That will allow chaining appropriately - in fact, I will do that now, since we need it.

extensions/cornerstone/src/panels/PanelMeasurement.tsx
wayfarer3130: Actually, that is what I wanted to have, I just wasn't quite there yet.  Let me develop this a bit mroe and will pass it by you.

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
wayfarer3130: Added docs not it.

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -27,75 +32,39 @@ export default function PanelMeasurementTable({
...
 
-  const toggleLockMeasurement = (uid: string) => {
-    measurementService.toggleLockMeasurement(uid);
-  };
+  const jumpToImage = bindCommand('jumpToMeasurement', { displayMeasurements });
+  const removeMeasurement = bindCommand('removeMeasurement');
+  const renameMeasurement = bindCommand('run', {
+    commands: ['jumpToMeasurement', 'renameMeasurement'],
+    displayMeasurements,
+  });
```
sedghi, line 46: I did not expec this honestly, which is basically this 

```
commandsManager.run('run', {commands, displayMeasurements})
```

can't we just edit the commandsManager to accepts string[]? instead of just string?

extensions/cornerstone-dicom-sr/src/commandsModule.ts
```diff
@@ -61,6 +82,13 @@ const commandsModule = (props: withAppTypes) => {
       window.location.assign(objectUrl);
     },
 
+    /**
+     * Download the CSV report for the measurements.
+     */
+    downloadCSVMeasurementsReport: ({ measurementFilter }) => {
```
sedghi, line 88: since this PR is going through another iteration, can you move this to also cornerstone commands too?

extensions/cornerstone/src/panels/PanelMeasurement.tsx
wayfarer3130: Done.

platform/core/src/classes/CommandsManager.ts
```diff
@@ -169,7 +169,7 @@ export class CommandsManager {
    *   the commandOptions specified in the base.
    */
   public run(
-    toRun: Command | Commands | Command[] | string | undefined,
+    toRun: Command | Commands | (Command | string)[] | string | undefined,
```
sedghi, line 172: could we please update the docsring above and give example of what we can handle 

something like 

run ('updateMeasurement') 
run (['updateMeasurement', 'displayWhatever'])
// other scnearios

extensions/cornerstone/src/getPanelModule.tsx
```diff
@@ -81,7 +81,7 @@ const getPanelModule = ({ commandsManager, servicesManager, extensionManager }:
       name: 'panelMeasurement',
       iconName: 'tab-linear',
       iconLabel: 'Measure',
-      label: 'Measurement',
+      label: 'All Measurements',
```
sedghi, line 84: I think changing this might disrupt other people's tests if they're selecting based on the label. Can we leave this one unchanged?

platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -168,7 +171,10 @@ class MeasurementService extends PubSubService {
    *
    * @return {Measurement[]} Array of measurements
    */
-  getMeasurements() {
+  public getMeasurements(filter?: MeasurementFilter) {
+    if (filter) {
+      return [...this.measurements.values()].filter(measurement => filter.call(this, measurement));
+    }
     return [...this.measurements.values()];
   }
```
sedghi, line 179: how about this 

```js
public getMeasurements(filter?: MeasurementFilter) {
return [...this.measurements.values()].filter(measurement => filter?.call(this, measurement) ?? true);
}
```

platform/core/src/utils/measurementFilters.ts
```diff
@@ -0,0 +1,108 @@
+/**
+ * Returns a filter function which filters for measurements belonging to both
+ * the study and series.
+ */
+export function filterTracked(trackedStudy: string, trackedSeries) {
+  return measurement => {
```
sedghi, line 6: does it matter tracked? can't it be `filterByStudyAndSeries` and still does the job?

platform/core/src/utils/measurementFilters.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+    return false;
+  };
+}
+
+/**
+ * Filters for additional findings, that is, measurements with
+ * a value of type point, and having a referenced image
+ */
+export function filterAdditionalFinding(measurementService) {
```
sedghi, line 49: filterAdditionalFindings? (s)

extensions/cornerstone/src/components/StudySummaryFromMetadata.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { DicomMetadataStore, utils } from '@ohif/core';
+import { StudySummary } from '@ohif/ui-next';
+
+const { formatDate } = utils;
+
+export function StudySummaryFromMetadata({ studyInstanceUID }) {
```
sedghi, line 7: Can we use dcmjs style namings? so `StudyInstanceUID` for the prop? below you are actually doing `const { StudyDate, StudyDescription } = instanceMeta;`

sedghi (commented):
see my comments thanks

sedghi (approved):
I was worried this part is removed

```js
useEffect(() => {
    const updateDisplayStudySummary = async () => {
      if (trackedMeasurements.matches('tracking') && trackedStudy) {
        const studyMeta = DicomMetadataStore.getStudy(trackedStudy);
        if (!studyMeta || !studyMeta.series || studyMeta.series.length === 0) {
          console.debug('Study metadata not available');
          return;
        }

        const instanceMeta = studyMeta.series[0].instances[0];
        const { StudyDate, StudyDescription } = instanceMeta;

        const modalities = new Set();
        studyMeta.series.forEach(series => {
          if (trackedSeries.includes(series.SeriesInstanceUID)) {
            modalities.add(series.instances[0].Modality);
          }
        });
        const modality = Array.from(modalities).join('/');

        setDisplayStudySummary(prevSummary => {
          if (prevSummary.key !== trackedStudy) {
            return {
              key: trackedStudy,
              date: StudyDate,
              modality,
              description: StudyDescription,
            };
          }
          return prevSummary;
        });
      } else if (!trackedStudy) {
        setDisplayStudySummary(DISPLAY_STUDY_SUMMARY_INITIAL_VALUE);
      }
    };

    updateDisplayStudySummary();
  }, [trackedMeasurements, trackedStudy, trackedSeries]);
```

but during my testing i didn't see any regression, so i guess it is fine

Issue #4201, opened 2024-06-02
[Bug] cannot recover camera values after resizing (release v3.8 and master)


Joeycho:
### Describe the Bug

It happens to release and master (latest). 

https://github.com/OHIF/Viewers/assets/29337166/2b97ee60-9d56-4ca2-b02c-b55c78cb6398

### Steps to Reproduce

1. Open the browser as Developer mode to reduce viewport size less than the minimum (I tried with Chrome)
2. Open the images
3. Reduce the width extremely
4. Recover the width
5. Image won't be rendered 

### The current behavior

The image won't be rendered after viewport size is recovered.

The camera values are not recovered.

![Screenshot 2024-06-02 at 22 52 55](https://github.com/OHIF/Viewers/assets/29337166/0cc72a17-695b-45d1-b5b5-ad2abbc5280f)

### The expected behavior

Image should be visible if the viewport size is recovered. (v3.7 can be recovered, by zoom in again)

### OS

mac OS 14.4

### Node version

16.18

### Browser

Chrome  125.0.6422.113

sedghi:
We should set a minimum width to like 5 % or something

Joeycho:
Agree, or.. how about re-rendering?

sedghi:
we can't re-render that small i guess

Issue #3159, opened 2023-02-10
Viewports in OHIF V3 can be reduced to zero and unusable small widths


jbocce:
### Describe the Bug

OHIF V3 can be resized such that the viewports can be reduced to zero and/or unusably small widths. Consideration should be made as to what is a tolerable small width. Furthermore consideration should be made as to how the OHIF app page behaves once a minimum width is decided upon. For example, should a horizontal scroll bar appear for the entire page once the minimum viewport width is reached and the page contents are too wide for the browser window? Or perhaps maybe just a horizontal scroll bar for the viewport area?

### Steps to Reproduce:

1. Launch OHIF.
2. Resize the browser window such that its width becomes smaller (i.e. resize to a thinner window).
3. Notice that the viewports can become an unusable small size and can be reduced to zero width.

### The current behavior

See the attached video.

https://user-images.githubusercontent.com/109477394/218104748-862504c0-bd35-4d36-9e2f-4257b3e1bed7.mp4

### The expected behavior

Nothing proposed - this issue is being logged so that consideration is made for this.

> :warning: Reports we cannot reproduce are at risk of being marked stale and
> closed. The more information you can provide, the more likely we are to look
> into and address your issue.

james-hanks:
Let's revisit when we redo mobile design @dan-rukas

james-hanks:
@dan-rukas assigned to take a look at this when we do mobile design and make sure this isn't an issue after the new design

PR #4517, opened 2024-11-15
[OHI-865] [OHI-646] fix(viewport): set a minimum width of 5px on viewports to prevent them from turning black/ going into an unrecoverable state.

IbrahimCSAE:
fixes https://github.com/OHIF/Viewers/issues/3159 
fixes https://github.com/OHIF/Viewers/issues/4201

## Context

Set a minimum width of 5px on viewports to prevent them from turning black/ going into an unrecoverable state.

before:

CleanShot 2024-11-15 at 02 01 45

after:

CleanShot 2024-11-15 at 02 00 49


sedghi (approved):
Thanks

PR #4555, opened 2024-11-29
fix(4d): disable 4d in any mode that is not preclincial 4D.

IbrahimCSAE:
### Context

There was some complaints that some series look like they are missing images when loader into the viewer because the number of images don't match the instance count. This was due to `getDynamicVolumeInfo` being called.

This has now been disabled and will only run in 4D mode.

Fixes: [OHI-829](https://linear.app/ohif/issue/OHI-829/[bug]-there-is-incomplete-display-of-series-images-in-latest-version)

thanks @lscoder


sedghi (commented):
I don't think this is the right fix, each displaySet has the info whether it is dynamic or not, we don't need an extra service for it

IbrahimCSAE:
No idea whats the alternative then if people want to view non 4D

lscoder:
> I don't think this is the right fix, each displaySet has the info whether it is dynamic or not, we don't need an extra service for it

`getDisplaySetInfo` changed by @IbrahimCSAE is called and `isDynamicVolume` returned is used by [makeDisplaySet](https://github.com/OHIF/Viewers/blob/master/extensions/default/src/getSopClassHandlerModule.js#L100). Since the displaySet is not available at the time `getDisplaySetInfo` is called IMO this is a viable solution and it is a bit better than my first one (#4294) because it reset the flag in a better way instead of relying on the mode to do it.

PR #4558, opened 2024-11-29
fix(SR): set view reference on hydrated measurements, don't jump to displaySets that are excluded from thumbnail browser

IbrahimCSAE:
### Context

Adds a view reference on hydrated measurements, since otherwise the viewport can't jump to them since there is not sufficient metadata.

Also ignores displaySets that are not in the thumbnail browser even if they have the made in client flag, it will use the actual SR from the panel, not the OHIF generated reference

### DIFF
--- a/extensions/cornerstone-dicom-sr/src/utils/hydrateStructuredReport.ts
+++ b/extensions/cornerstone-dicom-sr/src/utils/hydrateStructuredReport.ts
@@ -46,7 +46,13 @@ export default function hydrateStructuredReport(
 ) {
   const annotationManager = CsAnnotation.state.getAnnotationManager();
   const dataSource = extensionManager.getActiveDataSource()[0];
-  const { measurementService, displaySetService, customizationService } = servicesManager.services;
+  const {
+    measurementService,
+    displaySetService,
+    customizationService,
+    viewportGridService,
+    cornerstoneViewportService,
+  } = servicesManager.services;
 
   const codingValues = customizationService.getCustomization('codingValues', {});
 
@@ -191,13 +197,18 @@ export default function hydrateStructuredReport(
         // StudyInstanceUID,
       } = instance;
 
+      const activeViewportId = viewportGridService.getActiveViewportId();
+      const cornerstoneViewport =
+        cornerstoneViewportService.getCornerstoneViewport(activeViewportId);
+      const viewRef = cornerstoneViewport.getViewReference({ referenceImageId: imageId });
+
       const annotation = {
         annotationUID: toolData.annotation.annotationUID,
         data: toolData.annotation.data,
         metadata: {
           toolName: annotationType,
-          referencedImageId: imageId,
           FrameOfReferenceUID,
+          ...viewRef,
         },
       };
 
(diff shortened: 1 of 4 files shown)


IbrahimCSAE:
@wayfarer3130 Bill since you added this here https://github.com/OHIF/Viewers/pull/4182/files#r1863657682 , let me know if this fix is ok to you, thanks

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.tsx
```diff
@@ -282,14 +282,14 @@ export default function PanelStudyBrowserTracking({
...
-
+          const isExcludedFromThumbnailBrowser = dSet.excludeFromThumbnailBrowser;
           const newImageSrcEntry = {};
           const displaySet = displaySetService.getDisplaySetByUID(displaySetInstanceUID);
           if (displaySet?.unsupported) {
             return;
           }
 
-          if (options.madeInClient) {
+          if (options.madeInClient && !isExcludedFromThumbnailBrowser) {
```
sedghi, line 292: can you just do !dSet.excludeFromThumbnailBrowser here directly?

extensions/cornerstone/src/utils/measurementServiceMappings/Length.ts
```diff
@@ -6,7 +6,7 @@ import { utils } from '@ohif/core';
 import { config } from '@cornerstonejs/tools/annotation';
 
 const Length = {
-  toAnnotation: measurement => {},
+  toAnnotation: measurement => { },
```
sedghi, line 9: can you revert this, it adds noise to history

extensions/cornerstone-dicom-sr/src/utils/hydrateStructuredReport.ts
```diff
@@ -191,13 +197,21 @@ export default function hydrateStructuredReport(
         // StudyInstanceUID,
       } = instance;
 
+      const activeViewportId = viewportGridService.getActiveViewportId();
+      const cornerstoneViewport =
+        cornerstoneViewportService.getCornerstoneViewport(activeViewportId);
+      const imageIds = cornerstoneViewport.getImageIds();
+      const viewRef = cornerstoneViewport.getViewReference({
+        sliceIndex: imageIds.indexOf(imageId),
+      });
```
sedghi, line 206: This is tricky, there are some weird scenarios that we don't want this here, did you get this logic from somewhere else?

IbrahimCSAE: no, I just wrote it, so if it's not the best then I'm not sure what alternative we have

wayfarer3130: The problem is that the slice index may not be the image id index - what should be passed in is the image id itself, and let the different scenarios be computed inside.

sedghi: Agree

IbrahimCSAE: That sounds good to me too, I tried that initially however I think the method doesn’t take imageId unless I did something wrong.

I think I need to update it to work on cs3d side, it doesn’t let you get a viewref from an imageId

sedghi:
we don't need this, closing in favor of https://github.com/OHIF/Viewers/pull/4576

Issue #4180, opened 2024-05-28
[Feature Request] Publish docker images for ARM


jorianvo:
### What feature or change would you like to see made?

Currently only AMD64 is supported, it would be very helpful if the build also publishes a native ARM image.

### Why should we prioritize this feature?

It enables rapid prototyping against ARM-based clusters, without the need to rebuild images.

jorianvo:
Happy to contribute, let me know if this is something you would like to support.

sedghi:
I would love to do this as well, I think circleCI added the ARM option as well so i will check it

PR #4559, opened 2024-11-29
fix: Docker build time was very slow on a tiny change

wayfarer3130:
fixes https://github.com/OHIF/Viewers/issues/4180

### Context

The docker build setup was quite slow due to downloading and re-installing all the dependencies on every build.  Now it just caches those in a docker image and uses that until the package files change.

### Changes & Results
Dockerfile changes to improve build time
Use bun to build, improving the build time
Use pre-compressed files with gunzip to serve already zipped files
Fix bugs in setting the environment variables for app config
Allow the `--build-arg` for APP_CONFIG default and PUBLIC_URL
Preserve the build arg PUBLIC_URL as an environment variable when running

### Testing

Run docker build .
Make a tiny change in a typescript file - eg a console message in Mode.tsx or the like
Run docker build .
The second build should be far faster

Check that the build output for js/svg/map files is all compressed

Check that you can actually use an environment variable for APP_CONFIG - this was broken previously.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


Dockerfile
```diff
@@ -19,45 +20,51 @@
...
+RUN yarn install
 
-ENV PATH /usr/src/app/node_modules/.bin:$PATH
+# Build here
+# After install it should hopefully be stable until the local directory changes
 ENV QUICK_BUILD true
 # ENV GENERATE_SOURCEMAP=false
-# ENV REACT_APP_CONFIG=config/default.js
-
+ENV REACT_APP_CONFIG=config/default.js
```
sedghi, line 67: I think this was intentional, why uncommenting it?

platform/app/src/routes/index.tsx
```diff
@@ -112,6 +113,7 @@ const createRoutes = ({
   const allRoutes = [
     ...routes,
     ...(showStudyList ? [WorkListRoute] : []),
+    ...(publicUrl !== '/' && showStudyList ? [{ ...WorkListRoute, path: publicUrl }] : []),
```
sedghi, line 116: Do we need the line before? since that has also `...(showStudyList ? [WorkListRoute] : []),`

.docker/Viewer-v3.x/entrypoint.sh
```diff
@@ -7,9 +7,15 @@ if [ -n "$SSL_PORT" ]
     envsubst '${PORT}' < /usr/src/default.conf.template > /etc/nginx/conf.d/default.conf
 fi
 
-if [ -n "$APP_CONFIG" ]
+if [ -n "$APP_CONFIG" ] ; then
+  echo "$APP_CONFIG" > /usr/share/nginx/html${PUBLIC_URL}app-config.js
+  rm /usr/share/nginx/html${PUBLIC_URL}app-config.js.gz
+  gzip /usr/share/nginx/html${PUBLIC_URL}app-config.js
+  touch /usr/share/nginx/html${PUBLIC_URL}app-config.js
```
sedghi, line 14: can you add a comment why? so we put the APP_Config first via echo, then we remove and gzip again, so far makes sense. Why we are again creating an empty appConfig.js via `touch`?

platform/app/src/routes/index.tsx
wayfarer3130: Yes, the first one is on the / and the second on the public url.  I swapped those so the second is on the public url.

.docker/Viewer-v3.x/entrypoint.sh
wayfarer3130: Because you have to ensure it gets compressed for serving it up.  That is done by having an empty file named app_config.js and a compressed file named app_config.js.gz

IbrahimCSAE (approved):
I tested a bunch of the build commands locally with build-args and also env variables, everything works fine.

Issue #4551, opened 2024-11-27
[Feature Request] Improve Multiframe performance - metadata


sedghi:
### What feature or change would you like to see made?

We currently parse the metadata by creating the frame metadata from shared and per-frame sequences each time, which is inefficient. We should explore a better approach.

One suggestion by @wayfarer3130 is to use JavaScript objects with inheritance. We can start with an object containing all the instance data we have, then use:

```javascript
const shared = Object.create(instance, ...sharedFunctionalGroups);
SharedFunctionalGroupsSequence = shared;
```

For the functional groups, use a hidden value per frame with:

```javascript
Object.create(shared, ...perFrameValues);
```

This method will be more memory-efficient and will avoid constantly recreating objects. We can also implement a caching strategy for it.

### Why should we prioritize this feature?

Faster multiframe

PR #4560, opened 2024-11-29
fix: Having sop instance in a per-frame or shared attribute breaks load

wayfarer3130:
### Context

requires https://github.com/cornerstonejs/cornerstone3D/pull/1752

Some PACS systems combine single frame into multi-frame and then record the original SOPInstanceUID in the per frame attribute sequence.  This PR simply omits such per frame instances.

### Changes & Results

Uses a shared object for the parent series/instance data rather than merging, and avoid over-riding anything containing a SOP instance UID. 
This addresses https://github.com/OHIF/Viewers/issues/4551
It also addresses the issue from Jagan Vijayaraghavan in slack, for which he provided the study below.
[StudyWithSOPInFrameSequence.zip](https://github.com/user-attachments/files/18363130/Anonymized.-.202411528.1.zip)

The issue is that a shared functional groups sequence contains an original SOP instance UID for another object, and that SOP ends up replacing the correct SOP/Series UIDs, causing the wrong UIDs to be used.

### Testing

Load the study with the per-frame data from the cornerstone chat

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -1058,11 +1058,12 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
       // Store the current position presentations for each viewport.
       viewports.forEach(({ id: viewportId }) => {
         const presentation = this._getPositionPresentation(viewportId);
-
         // During a resize, the slice index should remain unchanged. This is a temporary fix for
         // a larger issue regarding the definition of slice index with slab thickness.
         // We need to revisit this to make it more robust and understandable.
-        delete presentation.viewReference.sliceIndex;
+        if (presentation?.viewReference) {
```
wayfarer3130, line 1064: This is still an issue entering MPR mode, but at least it doesn't throw

platform/core/src/utils/combineFrameInstance.ts
```diff
@@ -75,22 +75,24 @@ const combineFrameInstance = (frame, instance) => {
     }
     console.debug('🚀 ~ ImagePositionPatientToUse:', ImagePositionPatientToUse);
 
-    const newInstance = Object.assign(instance, { frameNumber: frameNumber });
+    const newInstance = Object.assign(Object.create(instance), { frameNumber: frameNumber });
```
wayfarer3130, line 78: This seems to be fine and is far faster

sedghi: can you remove my console debug above it?

sedghi (changes requested):
This is broken now after this PR, it does not go MPR, it was working before

[datscan upside down.zip](https://github.com/user-attachments/files/17962620/datscan.upside.down.zip)

sedghi:
@wayfarer3130 can you upload an anonymized study that does not work right now or bug report issue number

wayfarer3130:
@sedghi - could you check the uploaded study?  That study works correctly now, but I don't understand why your MPR issue breaks.  I tried looking at that, but if you could help debug that with me sometime I'd appreciate it.

platform/core/src/utils/combineFrameInstance.ts
```diff
@@ -73,26 +67,65 @@ const combineFrameInstance = (frame, instance) => {
...
+    );
 
-    const newInstance = Object.assign(instance, { frameNumber: frameNumber });
-
-    // merge the shared first then the per frame to override
-    [...shared, ...perFrame].forEach(item => {
-      Object.entries(item).forEach(([key, value]) => {
-        newInstance[key] = value;
-      });
+    // Note: do not assign directly to newInstance.ImagePositionPatient
```
wayfarer3130, line 76: This note doesn't seem to apply - I tried with:
objPar = { a: 3, b:4 }
objChild = Object.create(objPar)
objChild.a = 'ay'
objChild.c = 'c'
console.log("obj=", objChild.a, objChild.b, objChild.c, objPar.a)
And I see ay, 4, 'c', 3
which is what I expect.

wayfarer3130: I wonder if the instance is a proxy object and what we are seeing is a proxy assignment?

platform/core/src/utils/combineFrameInstance.ts
```diff
@@ -73,26 +67,65 @@ const combineFrameInstance = (frame, instance) => {
...
+  const shared = functionalGroups
+    ? Object.values(functionalGroups)
+        .filter(Boolean)
+        .map(it => it[0])
+        .filter(it => typeof it === 'object')
+    : [];
+
+  // merge the shared first then the per frame to override
+  [...shared].forEach(item => {
+    if (item.SOPInstanceUID) {
```
wayfarer3130, line 115: Should we simply avoid overriding any values in the child?  I know this tests for the SOPInstanceUID, but I would guess other things could get overridden.

wayfarer3130 (commented):
Somehow this won't let me approve today - other than the comment about the proxy I think it is fine.

platform/core/src/utils/combineFrameInstance.ts
sedghi: I don't know what is happening but it is the issue. The `instance.ImagePositionPatient` should not change

sedghi (approved):

PR #4586, opened 2024-12-10
feat(measurements): Provide for the Load (SR) measurements button to optionally clear existing measurements prior to loading the SR.

jbocce:
### Context

Prior to these changes, the load of SR measurements was confined to the `OHIFCornerstoneSRMeasurementViewport`. This meant that actions such as clearing the measurements prior to load were not easy to customize. Now the command(s) to run for the load are defined in a command and associated with a toolbar button. 

### Changes & Results

Refactored the load measurements code in `OHIFCornerstoneSRMeasurementViewport.tsx` to a new `loadMeasurements` command. The SR measurements viewport status bar now extracts the Load button to use from the 'loadMeasurements' toolbar. The Load button added to the toolbar defines the commands to be run. Typically this includes the (new) `loadMeasurements` command and optionally the `clearMeasurements` command.

Moved the load measurements toolbar buttons and section out of the modes and into the `cornerstone-dicom-sr` extension `onModeEnter` hook.

Refactored the HTML for the Load button out of `OHIFCornerstoneSRMeasurementViewport.tsx` into its own component named `ViewportActionButton`. The load measurements toolbar button uses this new component.
RT struct and SEG viewport statuses now use `ViewportActionButton` as well.

#### Test A
The load of measurements, SEG and RT struct should work as before.

#### Test B
1. To test the customization, add a new mode with no measurement tracking.
2. Edit the `LoadMeasurements` button in the mode to not clear the measurements prior to load. Alternatively, create a custom list of commands to execute instead.
3. The customization should work as whatever was intended in step 2 (e.g. the measurements should not clear, etc.).

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 Pro
- [x] Node version: 18.16.1
- [x] Browser: Chrome 131.0.6778.110
  


extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRMeasurementViewport.tsx
jbocce, line 1: There is some reformatting that was just done automagically. Hopefully it is all cool.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRMeasurementViewport.tsx
```diff
@@ -479,23 +438,52 @@ function _getStatusComponent({
...
-      </div>
-      {state === 3 && (
-        <div
-          className="bg-primary-main hover:bg-primary-light ml-1 cursor-pointer rounded px-1.5 hover:text-black"
-          // Using onMouseUp here because onClick is not working when the viewport is not active and is styled with pointer-events:none
-          onMouseUp={handleMouseUp}
-        >
-          {loadStr}
+  const StatusArea = () => {
+    const { toolbarButtons: loadMeasurementsButtons, onInteraction } = useToolbar({
```
wayfarer3130, line 442: What I was thinking is that a default definition of a toolbar could be provided here - or alternatively grabbed from the customization service, so that the mode doesn't need to touch it unless it is being customized.

modes/basic-test-mode/src/toolbarButtons.ts
```diff
@@ -248,6 +251,19 @@ const toolbarButtons: Button[] = [
       evaluate: 'evaluate.cornerstoneTool',
     },
   },
+  {
```
wayfarer3130, line 254: We need to figure out a way for the extensions to define default buttons too, so that it isn't necesary for every mode to know about all the button definitions unless they are creating new, custom ones.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRMeasurementViewport.tsx
```diff
@@ -479,23 +438,52 @@ function _getStatusComponent({
...
+                    displaySetInstanceUID: srDisplaySet.displaySetInstanceUID,
+                    viewportId,
+                  },
+                };
+              });
+              const tool = (
+                <Component
+                  key={id}
+                  id={id}
+                  onInteraction={onInteraction}
```
wayfarer3130, line 475: onInteraction={ (args) => onInteraction(args,commandOptions) }

extensions/cornerstone-dicom-rt/src/viewports/_getStatusComponent.tsx
```diff
@@ -34,13 +34,10 @@ export default function _getStatusComponent({ isHydrated, onStatusClick }) {
...
         </div>
         {!isHydrated && (
-          <div
-            className="bg-primary-main hover:bg-primary-light ml-1 cursor-pointer rounded px-1.5 hover:text-black"
-            // Using onMouseUp here because onClick is not working when the viewport is not active and is styled with pointer-events:none
-            onMouseUp={onStatusClick}
-          >
-            {loadStr}
-          </div>
+          <ViewportActionButton
```
wayfarer3130, line 37: Isn't this definition supposed to come from the toolbar definition?  That way the toolbar can have additional viewport action buttons - one for load, and a second for add

extensions/cornerstone-dicom-seg/src/viewports/_getStatusComponent.tsx
```diff
@@ -34,13 +34,10 @@ export default function _getStatusComponent({ isHydrated, onStatusClick }) {
...
         </div>
         {!isHydrated && (
-          <div
-            className="bg-primary-main hover:bg-primary-light ml-1 cursor-pointer rounded px-1.5 hover:text-black"
-            // Using onMouseUp here because onClick is not working when the viewport is not active and is styled with pointer-events:none
-            onMouseUp={onStatusClick}
-          >
-            {loadStr}
-          </div>
+          <ViewportActionButton
```
wayfarer3130, line 37: Is there any reason that the names/values can't be common and the getStatusComponent be moved up a layer?  That way the same status component would be used for both seg and sr, but with differing actions on the click.  That was kind of the point of using the toolbar.  I was not suggestion changing the SEG version to be common yet, although definitely using the same component for the button is nice to do.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRMeasurementViewport.tsx
```diff
@@ -479,23 +435,50 @@ function _getStatusComponent({
...
-        <span className="ml-1">SR</span>
-      </div>
-      {state === 3 && (
-        <div
-          className="bg-primary-main hover:bg-primary-light ml-1 cursor-pointer rounded px-1.5 hover:text-black"
-          // Using onMouseUp here because onClick is not working when the viewport is not active and is styled with pointer-events:none
-          onMouseUp={handleMouseUp}
-        >
-          {loadStr}
+  const StatusArea = () => {
```
wayfarer3130, line 438: Sorry about the earlier comments - those were not SR changes, but merely using a common component.

extensions/cornerstone-dicom-seg/src/viewports/_getStatusComponent.tsx
wayfarer3130: Sorry, I wasn't paying enough attention - please disregard.

extensions/cornerstone-dicom-rt/src/viewports/_getStatusComponent.tsx
wayfarer3130: Please disregard - I thought it was SR, not RT

extensions/cornerstone-dicom-sr/src/index.tsx
```diff
@@ -30,7 +32,27 @@ const dicomSRExtension = {
    * Only required property. Should be a unique value across all extensions.
    */
   id,
-  onModeEnter,
+
+  onModeEnter({ servicesManager }) {
+    const { toolbarService } = servicesManager.services;
+
+    toolbarService.addButtons([
```
wayfarer3130, line 39: Perhaps add a comment indicating these are default values to be used by the toolbars, and that they can of course be changed in the mode itself.

platform/ui/src/components/ViewportActionButton/ViewportActionButton.tsx
```diff
@@ -0,0 +1,28 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+function ViewportActionButton({ onInteraction, label, commands, id }) {
+  return (
+    <div
+      className="bg-primary-main hover:bg-primary-light ml-1 cursor-pointer rounded px-1.5 hover:text-black"
+      // Using onMouseUp here because onClick is not working when the viewport is not active and is styled with pointer-events:none
```
sedghi, line 8: nice comment

extensions/cornerstone-dicom-rt/src/viewports/_getStatusComponent.tsx
```diff
@@ -34,13 +34,10 @@ export default function _getStatusComponent({ isHydrated, onStatusClick }) {
...
+            label={loadStr}
```
sedghi, line 38: ditto

extensions/cornerstone-dicom-sr/src/index.tsx
```diff
@@ -30,7 +32,34 @@ const dicomSRExtension = {
...
+        // For example, the button label can be changed and/or command props to
+        // not clear the measurements can be passed to the command.
+        id: 'LoadMeasurements',
+        component: ViewportActionButton,
+        props: {
+          label: i18n.t('Common:LOAD'),
+          commands: [
+            {
+              commandName: 'loadMeasurements',
+              context: 'CORNERSTONE_STRUCTURED_REPORT',
```
sedghi, line 52: i think we don't need context anymore, we removed that, ask Bill please

extensions/cornerstone-dicom-sr/src/commandsModule.ts
```diff
@@ -123,6 +124,47 @@ const commandsModule = (props: withAppTypes) => {
...
+     * Loads measurements using either the given function or if the function is not present
+     * it hydrates and loads the SR for the given display set instance UID and displays
+     * it in the active viewport.
+     * @param loadMeasurementsEventFn the function to call to load the measurements
+     * @param loadMeasurementsEventName the name of the event to fire to load the measurements
+     * @param loadMeasurementsEventArgs arguments for the event; in particular the display set instance UID
+     * @param clearMeasurements optionally clear the measurements prior to hydrating the SR when loadMeasurementsEventFn is not present
+     * @returns
+     */
+    loadMeasurements: ({
```
sedghi, line 138: This is more like `loadStructuredReport` since sometimes those SRs are not measurements and just text right? like this study https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.2155604110.4180.1021041295.21

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRMeasurementViewport.tsx
```diff
@@ -73,28 +58,6 @@ function OHIFCornerstoneSRMeasurementViewport(props: withAppTypes) {
     sendTrackedMeasurementsEvent = tracked?.[1];
   }
 
-  if (!sendTrackedMeasurementsEvent) {
```
sedghi, line 76: I don't understand why `sendTrackedMeasurementsEvent` is removed it is used in the state machine right?

extensions/cornerstone-dicom-sr/src/index.tsx
```diff
@@ -30,7 +32,34 @@ const dicomSRExtension = {
...
+            },
+          ],
+        },
+      },
+    ]);
+
+    // The toolbar used in the viewport's status bar. The mode can further customize
+    // it to optionally add other buttons.
+    toolbarService.createButtonSection('loadMeasurements', ['LoadMeasurements']);
```
sedghi, line 61: since this is global, i like this to be more specific, like loadSRMeasurements or something

extensions/cornerstone-dicom-sr/src/commandsModule.ts
```diff
@@ -123,6 +124,47 @@ const commandsModule = (props: withAppTypes) => {
...
+      loadMeasurementsEventFn,
+      loadMeasurementsEventName,
+      loadMeasurementsEventArgs,
+      clearMeasurements = true,
+    }) => {
+      if (loadMeasurementsEventFn) {
+        loadMeasurementsEventFn(loadMeasurementsEventName, loadMeasurementsEventArgs);
+        return;
+      }
```
sedghi, line 147: I don't understand this pattern. Why are we passing all this information just to call it immediately? Can't this be done outside in the `OHIFCornerstoneSRMeasurementViewport`?

sedghi (commented):
great addition, see my comments please

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRMeasurementViewport.tsx
jbocce: This code is now in the `loadMeasurements` command. It is not removed, just moved. 😊

extensions/cornerstone-dicom-sr/src/commandsModule.ts
jbocce: I don't think so, but I am not the expert here. I believe text SRs are ultimately parsed and handled by `OHIFCornerstoneSRContentItem`. This command is strictly for loading (SR) measurements and was extracted from the `OHIFCornerstoneSRMeasurementViewport`

extensions/cornerstone-dicom-sr/src/index.tsx
jbocce: @wayfarer3130, comments?

platform/ui/src/components/ViewportActionButton/ViewportActionButton.tsx
jbocce: Thanks.

extensions/cornerstone-dicom-sr/src/index.tsx
jbocce: Sure.

extensions/cornerstone-dicom-seg/src/viewports/_getStatusComponent.tsx
```diff
@@ -34,13 +34,10 @@ export default function _getStatusComponent({ isHydrated, onStatusClick }) {
...
+            onInteraction={onStatusClick}
+            label={loadStr}
```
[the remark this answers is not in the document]
jbocce: certainly

extensions/cornerstone-dicom-sr/src/commandsModule.ts
jbocce: The idea was to move this code out of `OHIFCornerstoneSRMeasurementViewport` so that it can be more easily customizable in particular for those OHIF users that are NOT using tracked measurements. When not using tracked measurements, `loadMeasurementsEventFn` is undefined and the rest of the command is executed. Does this make sense?

wayfarer3130: Even for those using tracked measurements, the ability to execute load or other measurements options tied to commands or other buttons means one can then avoid having to display the SR panel, and go directly to loading the SR data.  That means one could, for example, double click to load an SR series.

extensions/cornerstone-dicom-sr/src/index.tsx
wayfarer3130: No, the context isn't required, and leaving it out means one can over-ride the command in the tracked measurements extension if we decide to replace it.

extensions/cornerstone-dicom-sr/src/commandsModule.ts
sedghi: It is just a strange pattern. I'm pretty sure there are better ways to do this, but I need to think about it. Maybe chop it into two things: one that is strange like this and the other that has the general logic regarding hydrateSR.

wayfarer3130: The right way of doing it is to register a command of the same name in the tracked extension, and then just call it without context.  In a tracked setting, it would just call the right command automatically because of priority of calls.
In a non-tracked setting, it would call the baseline.   The command itself would directly invoke the useTracked updates to perform the load.

jbocce: @wayfarer3130 , I tried adding the same command (name) in both the `measurement-tracking` and the `cornerstone-dicom-sr` extensions. In the mode I was using BOTH extensions are loaded and used (i.e. I was simply using the `longitudinal` mode). However, I noticed that the command in `cornserstone-dicom-sr` was called and NOT the one in `measurement-tracking`. Could it be so because the `cornerstone-dicom-sr` extension is registered AFTER the `measurement-tracking` extension? Please advise.

wayfarer3130: It is supposed to be called consistently in terms of ordering, in the order specified in the load order for the longitudinal mode.  Maybe that isn't happening correctly.  Hmm, the order is specified incorrectly in the load modules:
const extensionDependencies = {
  // Can derive the versions at least process.env.from npm_package_version
  '@ohif/extension-default': '^3.0.0',
  '@ohif/extension-cornerstone': '^3.0.0',
  '@ohif/extension-measurement-tracking': '^3.0.0',
  '@ohif/extension-cornerstone-dicom-sr': '^3.0.0',
  '@ohif/extension-cornerstone-dicom-seg': '^3.0.0',
  '@ohif/extension-cornerstone-dicom-pmap': '^3.0.0',
  '@ohif/extension-cornerstone-dicom-rt': '^3.0.0',
  '@ohif/extension-dicom-pdf': '^3.0.1',
  '@ohif/extension-dicom-video': '^3.0.1',
};


See how the sr is after the measurement tracking extension.  If you swap them it should fix things, and the tracking does depend on the dicom-sr module, so they should be loaded int he correct order.

extensions/cornerstone-dicom-sr/src/index.tsx
```diff
@@ -30,7 +31,37 @@ const dicomSRExtension = {
...
+        props: {
+          commands: [
+            {
+              commandName: 'clearMeasurements',
+            },
+            {
+              commandName: 'loadSRMeasurements',
+            },
+          ],
+        },
```
sedghi, line 57: we can just put strings i guess now, no? commands: ['clearMeasurements', 'loadSRMeasurements']

extensions/measurement-tracking/src/index.tsx
```diff
@@ -12,6 +16,32 @@ const measurementTrackingExtension = {
...
+        {
+          // A button for loading tracked, SR measurements.
+          // Note that the command run is registered in TrackedMeasurementsContext
+          // because it must be bound to a React context's data.
+          id: 'loadSRMeasurements',
+          component: props => (
+            <ViewportActionButton {...props}>{i18n.t('Common:LOAD')}</ViewportActionButton>
+          ),
+          props: {
+            commands: [
```
sedghi, line 34: same here, i think you can just say commands: ['loadTrackedSRMeasurements']

sedghi (commented):
looks much better, thanks minor comment

jbocce:
> looks much better, thanks minor comment

forget about it 😉

sedghi (approved):
Great addition

Issue #4580, opened 2024-12-08
[Bug] Patient list - selected patient changes randomly when clicked. (OHIF v3.10 beta/DCM4CHEE in Docker)


andylinton:
### Describe the Bug

The patient list shows abnormal behaviour when displaying a list of studies (in this case XR studies sent to a DCM4CHEE server by Voyance acquisition software).
When clicking on a patient in the list, the order suddenly changes and sometimes the name of the selected patient will be replaced by another patient's name.
This happens only when the patient list is sorted by date, usually when most recent studies listed first

This is a critical issue as it could lead to inadvertently opening the wrong patient’s study, risking incorrect diagnoses.

Issue noticed with various versions of OHIF, including 3.7, 3.9, latest, 3.10.0-beta.18. Able to reproduce with either DCM4CHEE or Orthanc archives
No obvious issues found with DICOM data

### Steps to Reproduce

1. Install OHIF in Docker with a minimal dcm4chee-arc. Using app-config.js example from https://github.com/OHIF/Viewers/blob/master/platform/app/public/config/local_dcm4chee.js 
2. Upload DICOM studies generated by the Voyance XR acquisition system to the PACS.
3. Open OHIF Viewer, click on patient names, observe study sort order

I've attached my app-config.js, docker-compose.yml, and 3 test DICOM files that are causing this behaviour via Dropbox here: https://www.dropbox.com/scl/fo/seack4lrw8q7uepyk24fn/ALzrmWxhu-WqzLPQWydWX-Y?rlkey=cd4puvmm0ilqo6uluhevfdt2i&st=imh29d3l&dl=0

### The current behavior

Clicking on a patient name causes the list to reorder unexpectedly, and the wrong patient may be highlighted.
See video of this behaviour at https://youtu.be/4CJ9Ano9O2Q

### The expected behavior

- The patient list should sort correctly by patient name or study date
- Clicking on a patient name should open that patient’s studies without reordering the list or selecting a different patient.
- If this is an issue with DICOM files or acquisition software - an error should be displayed rather than allowing the patient list to randomly change order.

### OS

Docker running on Ubuntu server 24.04, Windows 10

### Node version

Node.js v22.12.0 (LTS)

### Browser

Chrome / Firefox

sedghi:
I tried your data against the latest master branch commit, and can't reproduce it. 
I tried Orthanc

https://github.com/user-attachments/assets/0386581b-92c0-4f53-934e-5b4daa4030de

[...]

PR #4592, opened 2024-12-11
fix(worklist): selected patient changes randomly when clicked. 

IbrahimCSAE:
### Context

This PR fixes bug #4580, where the Worklist component was mutating the original studies array from the DatasourceWrapper component. This mutation caused state inconsistencies when the parent component updated the studies value, leading to incorrect behavior (e.g., patient switching) during sorting and row expansion.

Before (buggy):

https://github.com/user-attachments/assets/c9e2a0ed-5714-4f60-a4ac-f710695b0136

After (fixed):

https://github.com/user-attachments/assets/01de8a9f-686f-4788-9f90-3b82df449c73


sedghi (approved):
Thanks

PR #4599, opened 2024-12-12
fix: Inconsistencies and update the style setting on load for embedded styles from codingValues

wayfarer3130:
### Context

When the codingValues has a site/style value, and an SR is loaded containing annotations labelled with finding/findingSites values, there are inconsistencies with how the styling is applied, or whether it is applied at all.  
This also occurred with styling applied in different orders.

### Changes & Results

Added some extra styling for the test codingValues provided so that this can be seen.
Changed the updateMeasurement to apply the provided style first, but otherwise use the first style found from:  finding, the first findingSite found.
Added support for multiple findingSites names.  This allows differentiation between site data for different uses, but doesn't really deal well with multiple sites yet.

### Testing

Draw some annotations in basic test mode
Right click annotations, and use the finding/site sub-menus to apply finding or site values.
Note how the new version is order independent - the finding style always takes precedence if there is one, while the previous version last change to be applied got done.
Save an SR
Load the SR again in basic mode

You should see the annotation styles being applied.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/test-extension/src/custom-context-menu/codingValues.ts
```diff
@@ -23,28 +23,44 @@ const codingValues = {
   'SCT:69536005': {
     text: 'Head',
     type: 'site',
+    style: {
+      color: 'red',
```
sedghi, line 27: Can we normalize this to RGB or hex instead of text? What shade of red are we referring to, and what is its brightness?

sedghi (commented):
see my only comment please

sedghi (approved):

extensions/test-extension/src/custom-context-menu/codingValues.ts
sedghi: This seems to be a CSS style

PR #4602, opened 2024-12-13
[OHI-1479] fix(segmentation): black preview when loading a seg, and crash on opening panel

IbrahimCSAE:
### Context

- Adds a check for both SOPClassUID spellings: SOPClassUID, SopClassUID.
- Adds some metadata to the derived segmentation images

Fixes OHI-1479. this problem is only in local, when I tried with Orthanc there was no issues

### DIFF
--- a/extensions/cornerstone/src/panels/PanelSegmentation.tsx
+++ b/extensions/cornerstone/src/panels/PanelSegmentation.tsx
@@ -243,13 +243,15 @@ export default function PanelSegmentation({
     const firstImageId = referencedImageIds[0];
 
     const instance = metaData.get('instance', firstImageId);
-    const { SOPInstanceUID, SeriesInstanceUID } = instance;
+    const SOPInstanceUID = instance.SOPInstanceUID || instance.SopInstanceUID;
+    const SeriesInstanceUID = instance.SeriesInstanceUID;
 
     const displaySet = displaySetService.getDisplaySetForSOPInstanceUID(
       SOPInstanceUID,
       SeriesInstanceUID
     );
-    const isExportable = displaySet.isReconstructable;
+
+    const isExportable = displaySet?.isReconstructable;
 
     return {
       segmentationId,
--- a/extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
+++ b/extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
@@ -415,15 +415,10 @@ class SegmentationService extends PubSubService {
       imageIds as string[]
     );
 
-    segDisplaySet.images = derivedSegmentationImages;
-
-    segDisplaySet.images = segDisplaySet.images.map(image => {
-      const instance = metaData.get('instance', image.referencedImageId);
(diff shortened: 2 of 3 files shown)


extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -416,6 +417,14 @@ class SegmentationService extends PubSubService {
 
     segDisplaySet.images = derivedSegmentationImages;
 
+    segDisplaySet.images = segDisplaySet.images.map(image => {
```
IbrahimCSAE, line 420: let me know if this is ok.

extensions/cornerstone/src/panels/PanelSegmentation.tsx
```diff
@@ -243,13 +243,15 @@ export default function PanelSegmentation({
     const firstImageId = referencedImageIds[0];
 
     const instance = metaData.get('instance', firstImageId);
-    const { SOPInstanceUID, SeriesInstanceUID } = instance;
+    const SOPInstanceUID = instance.SOPInstanceUID || instance.SopInstanceUID;
```
IbrahimCSAE, line 246: fixes seg Panel

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
sedghi: Are you proposing deleting the previous line then?

extensions/default/src/DicomLocalDataSource/index.js
```diff
@@ -221,7 +221,8 @@ function createDicomLocalApi(dicomLocalConfig) {
       //   return instance.imageId;
       // }
 
-      const { StudyInstanceUID, SeriesInstanceUID, SOPInstanceUID } = instance;
+      const { StudyInstanceUID, SeriesInstanceUID } = instance;
+      const SOPInstanceUID = instance.SOPInstanceUID || instance.SopInstanceUID;
```
sedghi, line 225: this is very odd, i did not expect this to be an issue, is there really two things on the instance?

sedghi (commented):
see my comments

extensions/default/src/DicomLocalDataSource/index.js
IbrahimCSAE: it can be seen here too, so I guess its a thing https://github.com/cornerstonejs/cornerstone3D/blob/61c7d8beff128eed837feacdda495c9ec21780c2/packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts#L45

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
IbrahimCSAE: yes i'll remove

sedghi (approved):
I'm gonna merge this, but we should fix this better

PR #4603, opened 2024-12-13
[OHI-1357] feat(static-wado): add support for case-insensitive searching

### LINKED ISSUE #4522: [Bug] How to modify a configuration file to make case insensitive when searching？
### Describe the Bug

V 3.9.1
I want to search by name, but it seems to be case sensitive when searching. How can I modify the configuration file to ignore case when searching?
I tried to modify the configuration file, but it didn't work。I changed qidoSupportsIncludeField and supportsFuzzyMatching to true. It didn't work

      sourceName: 'dicomweb',
      configuration: {
        friendlyName: 'AWS S3 Static wado server',
        name: 'aws',
        wadoUriRoot: '****',
        qidoRoot: '*****',
        wadoRoot: '*****',
        imageRendering: 'wadors',
        thumbnailRendering: 'wadors',
        enableStudyLazyLoad: true,
        supportsFuzzyMatching: true,
        supportsWildcard: true,
        staticWado: true,
        singlepart: 'bulkdata,video',
        bulkDataURI: {
          enabled: true,
          relativeResolution: 'studies',
          transform: url => url.replace('/pixeldata.mp4', '/rendered'),
        },
        omitQuotationForMultipartRequest: true,

### Steps to Reproduce

1、Enter the all lowercase name in the search box
2、The result will not display the same patient name with uppercase letters

### The current behavior

The result will not display the same patient name with uppercase letters

### The expected behavior

Distinguish case when searching

### OS

Win10

### Node version

v20.15.1

### Browser

Chrome

IbrahimCSAE:
### Context

This adds the option to filter static-wado results without being case-sensitive. You need to set the `caseSensitive` property on your static wado `dataSource` to `false`, the default is `true`.

Orthanc and other sources support case insensitive searching already.

Fixes OHI-1357
Close #4522


platform/app/public/config/default.js
```diff
@@ -52,6 +52,7 @@ window.config = {
         supportsFuzzyMatching: false,
         supportsWildcard: true,
         staticWado: true,
+        caseSensitive: false,
```
sedghi, line 55: Can you add it to the docs as well?

https://docs.ohif.org/configuration/configurationFiles

It would be useful to have this per query, so that the patient name search is case-insensitive, but other searches are not. Can you make this an object with keys?

sedghi (commented):
we should make it per qido call and not global

sedghi:
Can you try it in Orthanc too please?

IbrahimCSAE:
> Can you try it in Orthanc too please?

I did, it works fine

sedghi (approved):

sedghi:
@wayfarer3130 can you take a look at this please

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -156,19 +164,28 @@ export default class StaticWadoClient extends api.DICOMwebClient {
...
     }
     if (actual?.Alphabetic) {
       actual = actual.Alphabetic;
     }
     if (typeof actual == 'string') {
+      if (!caseSensitive) {
+        if (typeof desired === 'string') {
+          desired = desired.toLowerCase();
+        }
+        if (typeof actual === 'string') {
```
wayfarer3130, line 185: actual.toLowerCase is sufficient, and resolves one bug where === 'string' doesn't work correctly.

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -227,7 +245,7 @@ export default class StaticWadoClient extends api.DICOMwebClient {
       return this.compareDateRange(testValue, valueElem.Value[0]);
     }
     const value = valueElem.Value;
```
wayfarer3130, line 247: You can do the testValue.toLowerCase here instead.

platform/docs/docs/configuration/configurationFiles.md
```diff
@@ -193,6 +193,12 @@ if auth headers are used, a preflight request is required.
 - `activateViewportBeforeInteraction`: (default to true), if set to false, tools can be used directly without the need to click and activate the viewport.
 - `autoPlayCine`: (default to false), if set to true, data sets with the DICOM frame time tag (i.e. (0018,1063)) will auto play when displayed
 - `addWindowLevelActionMenu`: (default to true), if set to false, the window level action menu item is NOT added to the viewport action corners
+- `caseSensitive`: an object with filter keys that should be case sensitive. By default, all filters are case sensitive. If you want to set a filter to be case insensitive, you can set it to false. Example:
```
wayfarer3130, line 196: You need to add a note that this only works for static DICOMweb query back ends

wayfarer3130 (changes requested):
Just a couple of small changes.

wayfarer3130:
Also, please change the default.js and e2e.js configurations to set this on by default - you may need to modify an integration test to support that.

IbrahimCSAE:
> Just a couple of small changes.

thanks Bill, I'll do them as soon as I can

sedghi (commented):
wait a minute does this only work on static wado backends? what happened to orthanc testing? did you do it?

IbrahimCSAE:
> wait a minute does this only work on static wado backends? what happened to orthanc testing? did you do it?

orthanc case sensitivity is server side, OHIF doesn’t control it

sedghi:
I thought the filtering happens client side when we looked at the code

IbrahimCSAE:
> I thought the filtering happens client side when we looked at the code

It only happens if it's a static wado datasource, otherwise it calls the super method, we don't control or filter anything there, so it's all up to backend (orthanc in this case) what to return

See here

https://github.com/dan-rukas/viewers/blob/a926b82a6231e54fc097335f9ccf6fd3079f1231/extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts#L102

```js
  async searchForStudies(options) {
    if (!this.staticWado) {
      return super.searchForStudies(options);
    }

    const searchResult = await super.searchForStudies(options);
    const { queryParams } = options;

    if (!queryParams) {
      return searchResult;
    }

    const lowerParams = this.toLowerParams(queryParams);
    const filtered = searchResult.filter(study => {
      for (const key of Object.keys(StaticWadoClient.studyFilterKeys)) {
        if (!this.filterItem(key, lowerParams, study, StaticWadoClient.studyFilterKeys)) {
          return false;
        }
      }
      return true;
    });
    return filtered;
  }
  ```

**Comment by sedghi** [comment]
I think the way forward is to add `fuzzyMatching` to our static wado, since that seems to be wildly supported by lots of dicom servers, see https://cloud.google.com/healthcare-api/docs/dicom#search_transaction `fuzzyMatching` section. 

Right now it is set in the config file as `supportsFuzzyMatching`:false 


**From the link** 
fuzzymatching: If set to true, fuzzy matching will be applied to the PatientName tag. Fuzzy matching will perform tokenization and normalization of both the value of PatientName in the query and the stored value. It will match if any search token is a prefix of any stored token. For example, if PatientName is "John^Doe", then "jo", "Do" and "John Doe" will all match. However "ohn" will not match.

**Comment by IbrahimCSAE** [comment]
> I think the way forward is to add `fuzzyMatching` to our static wado, since that seems to be wildly supported by lots of dicom servers, see https://cloud.google.com/healthcare-api/docs/dicom#search_transaction `fuzzyMatching` section.
> 
> Right now it is set in the config file as `supportsFuzzyMatching`:false
> 
> **From the link** fuzzymatching: If set to true, fuzzy matching will be applied to the PatientName tag. Fuzzy matching will perform tokenization and normalization of both the value of PatientName in the query and the stored value. It will match if any search token is a prefix of any stored token. For example, if PatientName is "John^Doe", then "jo", "Do" and "John Doe" will all match. However "ohn" will not match.

I implemented it according to the definition from GCP. please take another look now, thanks

File: extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts [hunk 5]
```diff
@@ -156,18 +156,36 @@
...
     if (Array.isArray(actual)) {
-      return actual.find(actualItem => this.compareValues(desired, actualItem));
+      return actual.find(actualItem => this.compareValues(desired, actualItem, fuzzyMatching));
     }
     if (actual?.Alphabetic) {
       actual = actual.Alphabetic;
     }
+
+    if (fuzzyMatching && typeof actual === 'string' && typeof desired === 'string') {
+      const normalizeValue = str => str.replace(/^\*+|\*+$/g, '').toLowerCase();
```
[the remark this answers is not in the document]
sedghi: @IbrahimCSAE can you check this as well?

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -156,18 +156,44 @@ export default class StaticWadoClient extends api.DICOMwebClient {
...
+        if (str.endsWith('*')) {
+          str = str.slice(0, -1);
+        }
+        return str.toLowerCase();
+      };
+
+      const normalizedDesired = normalizeValue(desired);
+      const normalizedActual = normalizeValue(actual);
+
+      const tokenizeAndNormalize = str => str.replace(/\^/g, ' ').split(/\s+/).filter(Boolean);
```
wayfarer3130, line 187: This can be written:
 const tokenizeAndNormalize = str => str.split(/(\s|\^]+/).filter(Boolean);

sedghi (approved):
See Bill's comment and it is good to go thanks

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -156,18 +156,44 @@ export default class StaticWadoClient extends api.DICOMwebClient {
...
+
+      const desiredTokens = tokenizeAndNormalize(normalizedDesired);
+      const actualTokens = tokenizeAndNormalize(normalizedActual);
+
+      return desiredTokens.some(desiredToken =>
```
wayfarer3130, line 192: This should be desiredTokens.every - I know that the GCP definition has it with any, but that doesn't make any sense because it should be possible to provide a list of values to support "any", while it is important to allow multiple values to be done using every - think of patients whose first name is "John", and we are looking for "John Smith"
If it is done with John Smith as the query field, then some will find all the johns AND all the smiths, not the john smiths.

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -156,18 +156,44 @@ export default class StaticWadoClient extends api.DICOMwebClient {
...
-      return actual.find(actualItem => this.compareValues(desired, actualItem));
+      return actual.find(actualItem => this.compareValues(desired, actualItem, fuzzyMatching));
     }
     if (actual?.Alphabetic) {
       actual = actual.Alphabetic;
     }
+
+    if (fuzzyMatching && typeof actual === 'string' && typeof desired === 'string') {
+      const normalizeValue = str => {
+        if (str.startsWith('*')) {
```
wayfarer3130, line 175: If you supported wildcard matching here too, with prefix/suffix, then you can support fuzzy matching for names, and wildcard start/ends with for other fields.  I would suggest changing the support of supportsWildcard to `suffix`, to automatically have the wildcard added only at the end, although doing a full contains works as well for client side matching.

platform/app/public/config/default.js
```diff
@@ -50,7 +50,7 @@ window.config = {
         imageRendering: 'wadors',
         thumbnailRendering: 'wadors',
         enableStudyLazyLoad: true,
-        supportsFuzzyMatching: false,
+        supportsFuzzyMatching: true,
```
wayfarer3130, line 53: Suggest changing the next line for supportsWildcard to false, or else implement the logic for wildcard matching too.

IbrahimCSAE: done

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
IbrahimCSAE: done

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -156,18 +156,44 @@ export default class StaticWadoClient extends api.DICOMwebClient {
...
     if (Array.isArray(actual)) {
-      return actual.find(actualItem => this.compareValues(desired, actualItem));
+      return actual.find(actualItem => this.compareValues(desired, actualItem, fuzzyMatching));
     }
     if (actual?.Alphabetic) {
       actual = actual.Alphabetic;
     }
+
+    if (fuzzyMatching && typeof actual === 'string' && typeof desired === 'string') {
+      const normalizeValue = str => {
```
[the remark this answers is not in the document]
IbrahimCSAE: done

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
IbrahimCSAE: I think this implementation is sufficient for now. We can dedicate more time to implementing proper server-side filtering logic for static wado using lamdas or something similar then we can avoid all of this in the frontend.

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -211,6 +231,9 @@ export default class StaticWadoClient extends api.DICOMwebClient {
    * @returns
    */
   filterItem(key: string, queryParams, study, sourceFilterMap) {
+    const { supportsFuzzyMatching = false } = this.config;
+    const isPatientName = key === 'patientname';
```
wayfarer3130, line 235: Could you make it key.indexof('name')!==-1 so that it works for other names as well?

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -227,7 +250,8 @@ export default class StaticWadoClient extends api.DICOMwebClient {
       return this.compareDateRange(testValue, valueElem.Value[0]);
     }
     const value = valueElem.Value;
-    return this.compareValues(testValue, value);
+
+    return this.compareValues(testValue, value, supportsFuzzyMatching && isPatientName);
```
wayfarer3130, line 254: Can you pass in an options object instead of a single key - that way we can add wildcard matching in the future without changing the API.

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
IbrahimCSAE: done

wayfarer3130 (approved):

andylinton:
This change seems to have broken patient search for me. Using DCM4CHEE PACS with supportsFuzzyMatching: true, and supportsWildcard: true. I tried adding caseSensitive: false but this didn't help. Whatever I try, searching the patient list by name gives no results.

andylinton:
I set staticWado to false and the search issue was resolved, I can now search by patient name.

Issue #4593, opened 2024-12-11
[Bug] Segmentation not loaded by default with initialSeriesInstanceUID


xbankov:
### Describe the Bug

Parameter `initialSeriesInstanceUID` is not loading the segmentation by default.

### Steps to Reproduce

1. Go to [https://viewer.ohif.org/viewer?StudyInstanceUIDs=61.7.93273854116647800470730671671118421206&initialSeriesInstanceUID=61.7.280176113695462316231376981119115690926](https://viewer.ohif.org/viewer?StudyInstanceUIDs=61.7.93273854116647800470730671671118421206&initialSeriesInstanceUID=61.7.280176113695462316231376981119115690926)

### The current behavior

Segmentation present but not loaded by default.

### The expected behavior

Segmentation loads by default.

### OS

Not applicable

### Node version

Not applicable

### Browser

Mozilla Firefox 133.0, Google Chrome 131.0.6778.139

sedghi:
Hmmmm

xbankov:
I have similar problems with the RT modality. I tried to look into it, and I realize in 3.7.0 it works as indented, and 3.8.0 is already not loading segmentations by default. 

I was also surprised to see some segmentations not going through segmentation Viewport when opened, as I setup logging and some segmentation printed logs, and some not. 

I cannot really see the difference for the segmentations files itself, only that they also have associated CT ref AND RTSTRUCT.

Is there some inherent logic that loads every associated files that could prevent Segmentation to load?

xbankov:
Thanks for the fix!

PR #4605, opened 2024-12-16
fix(seg): jump to the first slice in SEG and RT that has data

sedghi:
fixes https://github.com/OHIF/Viewers/issues/4593

and it fixes the initial slice to load for SEG and RTSTRUCT to be the slice that has data and is not empty

### DIFF
--- a/extensions/cornerstone/src/hps/frameView.ts
+++ b/extensions/cornerstone/src/hps/frameView.ts
@@ -20,7 +20,7 @@ const frameView: Types.HangingProtocol.Protocol = {
         },
         {
           attribute: 'isDisplaySetFromUrl',
-          weight: 10,
+          weight: 20,
           constraint: {
             equals: true,
           },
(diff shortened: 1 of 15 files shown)


extensions/cornerstone/src/hps/frameView.ts
```diff
@@ -20,7 +20,7 @@ const frameView: Types.HangingProtocol.Protocol = {
         },
         {
           attribute: 'isDisplaySetFromUrl',
-          weight: 10,
+          weight: 20,
```
wayfarer3130, line 23: Why is this change being included?

wayfarer3130 (approved):
Only the question on why the change to weight: 20 in the isDisplaySetFromUrl

extensions/cornerstone/src/hps/frameView.ts
sedghi: The weight was the same as `numImages > 0`, which didn't make the RT or SEG appear as the starting series, so I'm giving more weight to the URL.

PR #4611, opened 2024-12-18
fix ohif icons in Header

Ceciliadrc:
Fix breaking change in Header legacy, <svg> component was not imported (Cecilia and Salim)


sedghi (approved):
Thanks a lot!

PR #4617, opened 2024-12-19
WIP - feat: group measurements by study

pedrokohler:
### Context

Building on top of: https://github.com/OHIF/Viewers/pull/4501

### Changes & Results

These changes group the measurements of the all-measurements tab by study.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -49,9 +52,22 @@ export default function PanelMeasurementTable({
...
+
+      if (!groupedMeasurements.has(key)) {
+        groupedMeasurements.set(key, [item]);
+        return groupedMeasurements;
+      }
+
+      const oldValues = groupedMeasurements.get(key);
+      oldValues.push(item);
+      return groupedMeasurements;
+    }, new Map()) as Map<string, object[]>;
```
wayfarer3130, line 69: Cant you do new Map<string,object[]> to get the typing right?

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -71,27 +87,31 @@ export default function PanelMeasurementTable({
...
-                  : customHeader}
-              </>
-            )}
-          </MeasurementTable.Header>
-          <MeasurementTable.Body />
-        </MeasurementTable>
+        {Array.from(measurements).map(([key, value]) => {
+          return (
+            <MeasurementTable
+              key={`tracked-${key}`}
```
wayfarer3130, line 93: entThis isn't necessarily tracked on this version of it, so you shouldn't be referencing tracked here, but should just get something passed in from the parent about what you are naming this.

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -69,29 +85,33 @@ export default function PanelMeasurementTable({
       <div
         className="invisible-scrollbar overflow-y-auto overflow-x-hidden"
         ref={measurementsPanelRef}
-        data-cy={'trackedMeasurements-panel'}
+        data-cy={'measurements-panel'}
```
wayfarer3130, line 88: Does this change cause any of the tests to fail?

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -48,12 +51,25 @@ export default function PanelMeasurementTable({
 
   const additionalFilter = filterAdditionalFinding(measurementService);
 
-  const { measurementFilter: trackedFilter } = measurementFilters;
-  const measurements = displayMeasurements.filter(
-    item => !additionalFilter(item) && trackedFilter(item)
-  );
+  const { measurementFilter } = measurementFilters;
+  const measurements = displayMeasurements
```
wayfarer3130, line 55: This grouping function is basically what should be done, but the point is to do it generically with a passed in grouping method so that one can group by study, but also in the future use other grouping methods.   Also, it should be recursive, so that the child grouping items are also using the same group.

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -69,29 +85,33 @@ export default function PanelMeasurementTable({
...
+                        })
+                      : customHeader}
+                  </>
+                )}
+              </MeasurementTable.Header>
+              <MeasurementTable.Body />
+            </MeasurementTable>
+          );
+        })}
         {additionalFindings.length > 0 && (
```
wayfarer3130, line 115: The additional findings are a nested/recursive application of the grouping function - so that this is a sub-group of hte study grouping.

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
```diff
@@ -53,6 +53,7 @@ function PanelMeasurementTableTracking({
         extensionManager={extensionManager}
         commandsManager={commandsManager}
         measurementFilters={measurementFilters}
+        title="Measurements"
```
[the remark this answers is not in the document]
wayfarer3130: Probably the default grouping function for PanelMeasurements is fine, just passing in the top level selector which selects only tracked measurements, and that should ONLY use series instance UID so that when we add multiple studies to PanelMeasurement it just automatically creates multiple sub-groups.

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -1,24 +1,27 @@
...
 import { MeasurementTable } from '@ohif/ui-next';
 import debounce from 'lodash.debounce';
 import { useMeasurements } from '../hooks/useMeasurements';
 
-const { filterAdditionalFinding, filterOr, filterAny } = utils.MeasurementFilters;
+const { filterAdditionalFinding, filterAny } = utils.MeasurementFilters;
 
 export type withAppAndFilters = withAppTypes & {
   measurementFilters: Record<string, (item) => boolean>;
+  groupBy: string;
```
wayfarer3130, line 11: Add a method here which creates an array of group items, and assign it the default groupMeasurementsByStudyUID

wayfarer3130:
Could you add two additional breakdown to the PanelMeasurements, and change the configuration of PanelMeasurements to use that.  The classes/functions we want are:

groupStudyMeasurement(props, measurements)
   - this is a configurable function that takes the list of measurements and produces a set of grouping items for them, using the props to define any of the grouping methods.
   - the return type is an array of object containing an optional header and a mandatory body, maybe a collapsed body
   - the return instance value creates one array element per study UID found in the measurements, and has an instance of PanelMeasurementGroup as it's default body instance if not configured otherwise in the props
  - For example, it might have   [ {key: studyInstanceUID, body: PanelMeasurementGroup, items: filtered measurements }, ... same sort of item for every study]

Then split the existing primary items/additional items into a separate component that just shows the two items, along with the primary header.

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -48,12 +51,25 @@ export default function PanelMeasurementTable({
...
-  const { measurementFilter: trackedFilter } = measurementFilters;
-  const measurements = displayMeasurements.filter(
-    item => !additionalFilter(item) && trackedFilter(item)
-  );
+  const { measurementFilter } = measurementFilters;
+  const measurements = displayMeasurements
+    .filter(item => !additionalFilter(item) && measurementFilter(item))
+    .reduce((groupedMeasurements, item) => {
+      const displaySet = displaySetService.getDisplaySetByUID(item.displaySetInstanceUID);
+      const key = displaySet.instances[0].StudyDescription;
```
wayfarer3130, line 59: Better use the StudyInstanceUID here to make it unique

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -48,12 +51,25 @@ export default function PanelMeasurementTable({
...
+
+      if (!groupedMeasurements.has(key)) {
+        groupedMeasurements.set(key, [item]);
+        return groupedMeasurements;
```
wayfarer3130, line 63: Create a grouped measurement object that defines the sub-component to render this on rather than making it a simple list.

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -48,12 +52,29 @@ export default function PanelMeasurementTable({
...
   const additionalFilter = filterAdditionalFinding(measurementService);
 
-  const { measurementFilter: trackedFilter } = measurementFilters;
-  const measurements = displayMeasurements.filter(
-    item => !additionalFilter(item) && trackedFilter(item)
-  );
+  const { measurementFilter } = measurementFilters;
+  const defaultGroupingFunction = (groupedMeasurements, item) => {
+    const displaySet = displaySetService.getDisplaySetByUID(item.displaySetInstanceUID);
+    const key = displaySet.instances[0].StudyDescription;
```
wayfarer3130, line 58: Use StudyInstanceUID

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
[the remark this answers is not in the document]
pedrokohler: Filtering measurements to include only tracked ones is understandable—that makes sense.

However, I don’t understand why, or even what it means, to 'group' measurements by 'tracked measurements.' Could you clarify?

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -48,13 +54,34 @@ export default function PanelMeasurementTable({
...
-  const { measurementFilter: trackedFilter } = measurementFilters;
-  const measurements = displayMeasurements.filter(
-    item => !additionalFilter(item) && trackedFilter(item)
-  );
-  const additionalFindings = displayMeasurements.filter(
-    item => additionalFilter(item) && trackedFilter(item)
-  );
+  const { measurementFilter } = measurementFilters;
+
+  const defaultGroupingFunction = groupByStudy(displaySetService);
```
wayfarer3130, line 59: I think you should assume that any other grouping function might also need services and/or commands managers.  If you do it this way, then you can setup the grouping function appropriately:
const groupingFunction = (groupingFunctionConstructor || groupByStudyConstructor)(props)
where props is the props to the panel measurement table.

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -1,24 +1,30 @@
...
   measurementFilters: Record<string, (item) => boolean>;
+  groupingFunction?: (groupedMeasurements: Map<string, object[]>, item) => Map<string, object[]>;
+  title: string;
 };
 
 export default function PanelMeasurementTable({
   servicesManager,
   commandsManager,
   customHeader,
   measurementFilters = { measurementFilter: filterAny },
```
wayfarer3130, line 21: Could you fix this type - make it just pass in measurementFilter, it doesn't need to be multiple instances here.

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
```diff
@@ -28,7 +28,7 @@ function PanelMeasurementTableTracking({
   });
 
   useEffect(() => {
-    let updatedMeasurementFilters = { ...measurementFilters };
+    const updatedMeasurementFilters = { ...measurementFilters };
     if (trackedMeasurements.matches('tracking') && trackedStudy) {
       updatedMeasurementFilters.measurementFilter = filterTracked(trackedStudy, trackedSeries);
```
wayfarer3130, line 33: Can you fix the measurementFilter to be just a single filter, and make it independent of the tracked study.

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
```diff
@@ -46,65 +46,63 @@ function PanelMeasurementTableTracking({
...
-              </div>
+        return (
+          <div className="bg-background flex h-9 w-full items-center rounded pr-0.5">
+            <div className="flex space-x-1">
+              <Button
+                size="sm"
+                variant="ghost"
+                className="pl-1.5"
+                onClick={() => {
+                  commandsManager.runCommand('clearMeasurements', measurementFilters);
```
wayfarer3130, line 70: Ouch - this one is running clear measurements instead of saveCSV - can you fix that so it runs saveCSV (or whatever the command name is to save the CSV)
Also, can you pass in the nested measurement filter instead of the top level measurement filter for all 3 commands - that is the new filter I (just) asked to be created in the PanelMeasurement table - that way the command applies to the correct set of data.

platform/core/src/utils/measurementGroupings.ts
```diff
@@ -0,0 +1,13 @@
+export const groupByStudy = displaySetService => (groupedMeasurements, item) => {
```
wayfarer3130, line 1: Just add a docs to indicate that this groups measurements by study in order to allow display and saving by study.

wayfarer3130 (changes requested):
very close now, I did notice one bug that had been previously introduced (probably by me) that needs fixing, and another issue that would only occur if we enabled multiple study tracking.

sedghi:
Can you please revisit this PR since the parent PR is merged? Thanks

extensions/cornerstone/src/getPanelModule.tsx
```diff
@@ -81,7 +81,7 @@ const getPanelModule = ({ commandsManager, servicesManager, extensionManager }:
       name: 'panelMeasurement',
       iconName: 'tab-linear',
       iconLabel: 'Measure',
-      label: 'Measurement',
+      label: 'All Measurements',
```
wayfarer3130, line 84: Can you revert this change - it was somethign Alireza didn't want.

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -68,39 +103,60 @@ export default function PanelMeasurement({
...
+          </div>
+        ) : (
+          <></>
         )}
+        {items.map(item => {
+          return (
+            <div key={`${item.study}`}>
+              <StudySummaryFromMetadata StudyInstanceUID={item.study} />
+              {item.measurements.length === 0 ? (
+                <></>
```
wayfarer3130, line 120: null can returned here

wayfarer3130 (approved):
These changes are a good base for the grouping of measurements

wayfarer3130 (approved):
It looks nice to be able to have the additional studies grouping.

PR #4657, opened 2025-01-08
chore: delete unused and duplicate function, it exists under /utils

Joeycho:
`extensions/default/src/utils/_shared/getNextSRSeriesNumber.ts` is not used anymore

`extensions/default/src/utils/getNextSRSeriesNumber.ts` is used.


sedghi (approved):
Thanks

PR #4668, opened 2025-01-09
study item dropdown menu customizations

sedghi:
some ideas

CleanShot 2025-01-09 at 13 55 46@2x


extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.tsx
```diff
@@ -24,6 +30,35 @@ const thumbnailNoImageModalities = [
   'OT',
   'PMAP',
 ];
+
+const getMenuItems = ({ commandsManager, items, servicesManager, ...props }) => {
```
wayfarer3130, line 34: Can the menu items come from the base study browser?  Is there a good reason to use different menus/setups between the tracked and non-tracked study browser?

platform/ui-next/src/components/StudyBrowser/StudyBrowser.tsx
wayfarer3130, line 1: This file should also work fine to display non-tracked views, and still get the menu etc.

wayfarer3130 (changes requested):
It LGTM except that the StudyBrowser base panel should also be able to have the context menu.  We really need to add a test mode for non-tracked OHIF so that we make sure it all works.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -413,7 +409,7 @@ export default class HangingProtocolService extends PubSubService {
    */
   public run({ studies, displaySets, activeStudy, activeStudyUID }, protocolId, options = {}) {
     this.studies = [...(studies || this.studies)];
-    this.displaySets = displaySets || this.displaySets;
+    this.displaySets = displaySets;
```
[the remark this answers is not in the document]
sedghi: We can add that change in a future PR when needed, if that's okay.

extensions/default/src/Panels/StudyBrowser/PanelStudyBrowser.tsx
```diff
@@ -10,43 +10,38 @@ import { defaultActionIcons, defaultViewPresets } from './constants';
...
 } from '@ohif/ui-next';
 
 const { sortStudyInstances, formatDate, createStudyBrowserTabs } = utils;
 
-const StudyMenuItems = ({ StudyInstanceUID, commandsManager }) => {
+const getMenuItems = ({ commandsManager, items, servicesManager, ...props }) => {
+  const { customizationService } = servicesManager.services;
+
+  // Todo: this feels odd, but maybe my feelings are wrong
+  const menuContent = customizationService.getCustomization('ohif.menuContent');
```
wayfarer3130, line 24: I think this is generally a good idea, but the menu content should be named something like ohif.menuContent.PanelStudyBrowser  and the name should come from a parameter so that the panel props can change the entire menu for an instance, or just change the menu directly.

extensions/default/src/Panels/StudyBrowser/PanelStudyBrowser.tsx
```diff
@@ -356,10 +346,51 @@ function PanelStudyBrowser({
         showSettings={actionIcons.find(icon => icon.id === 'settings').value}
         viewPresets={viewPresets}
         ThumbnailMenuItems={props => (
-          <CustomizedThumbnailMenuItems
-            {...props}
-            commandsManager={commandsManager}
-          />
+          <DropdownMenu>
```
wayfarer3130, line 349: Can this be externalized as a shared class?  It feels like 
<CustomizedDropdownMenu customizationService={customizationService} menuId="studyBrowser.thumbnailMenuItems" />
would allow both menus to use the same component.

extensions/default/src/getCustomizationModule.tsx
```diff
@@ -312,7 +312,7 @@ export default function getCustomizationModule({ servicesManager, extensionManag
             },
             {
               id: 'showInOtherMonitor',
-              label: 'Launch On Other Monitor',
+              label: 'Launch On Second Monitor',
```
wayfarer3130, line 315: It is actually supposed to be other monitor since there might be three monitors, and it might be executed on the second monitor already, causing it to launch on the first monitor.

PR #4673, opened 2025-01-10
[OHI-1547] fix(docker): run compression regardless of APP_CONFIG being present ( in cases such as volume mount)

### LINKED ISSUE #4669: [Bug] Unable to load app-config.js via docker volume mount since PR#4559 / v3.10.0-beta.34 - demo data showing
### Describe the Bug

After updating to the latest beta version (to correct the 2025 date filter issue), OHIF is displaying demo data instead of the data from my DCM4CHEE server. 
I am using the OHIF docker image v3.10.0-beta.49 but the issue is noted for all versions after v3.10.0-beta.34
In my setup, I pass my customised app-config.js to OHIF via a Docker volume mount: 
```
    volumes:
      - ./config/app-config.js:/usr/share/nginx/html/app-config.js
```

It seems that since PR#4559 / v3.10.0-beta.34 this is no longer passing the app-config.js correctly and OHIF is just displaying demo data. 

### Steps to Reproduce

Sample docker compose (with OHIF and DCM4CHEE) - https://pastebin.com/v72eyL6m 
Sample app-config.js - https://pastebin.com/GuDGjaVm
Put both in the same folder, adjust IP address in app-config.js
```docker compose up -d```
http://<ip_address>:3002 - will show demo data
http://<ip_address>:3003 - will show black list as it connects properly to DCM4CHEE using the app-config.js

### The current behavior

After v3.10.0-beta.34 it is no longer working to pass app-config.js as a docker volume mount to OHIF Docker container

Note: executing ```docker exec -it ohif-test1 /bin/sh``` and ```cat /usr/share/local/nginx/html/app-config.js``` shows that the correct file is being passed into the OHIF container. It is just not using the file correctly

### The expected behavior

Passing app-config.js as a docker volume mount should mean that OHIF uses it correctly.

### OS

Docker

### Node version

current

### Browser

Chrome / Firefox

### PR DESCRIPTION
**Pull request by IbrahimCSAE**
### Context

Docker compression step was not being triggered if APP_CONFIG is not present, but it should still do if a user mounts a file using volumes.

Fixes OHIF-1547
Closes #4669

### DIFF
--- a/.docker/Viewer-v3.x/entrypoint.sh
+++ b/.docker/Viewer-v3.x/entrypoint.sh
@@ -7,19 +7,25 @@ if [ -n "$SSL_PORT" ]
     envsubst '${PORT}' < /usr/src/default.conf.template > /etc/nginx/conf.d/default.conf
 fi
 
-if [ -n "$APP_CONFIG" ] ; then
+
+if [ -n "$APP_CONFIG" ]; then
   echo "$APP_CONFIG" > /usr/share/nginx/html${PUBLIC_URL}app-config.js
-  # Removes the old compressed app-config file, then compresses the replacement
-  # and finally creates a new empty file so that gunzip works correctly.
-  # This code is correct despite the AI warning otherwise about order of create/delete
-  rm /usr/share/nginx/html${PUBLIC_URL}app-config.js.gz
+  echo "Using custom APP_CONFIG environment variable"
+else
+  echo "Not using custom APP_CONFIG"
+fi
+
+if [ -f /usr/share/nginx/html${PUBLIC_URL}app-config.js ]; then
+  echo "Detected app-config.js. Ensuring .gz file is updated..."
+  rm -f /usr/share/nginx/html${PUBLIC_URL}app-config.js.gz
   gzip /usr/share/nginx/html${PUBLIC_URL}app-config.js
   touch /usr/share/nginx/html${PUBLIC_URL}app-config.js
+  echo "Compressed app-config.js to app-config.js.gz"
+else
+  echo "No app-config.js file found. Skipping compression."
 fi
-if [ ! -n "$APP_CONFIG" ]
-  then
-    echo "Not using custom app config"
-fi
+
+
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .docker/Viewer-v3.x/entrypoint.sh [hunk 1]
```diff
@@ -7,20 +7,30 @@ if [ -n "$SSL_PORT" ]
...
+  echo "Using custom APP_CONFIG environment variable"
+else
+  echo "Not using custom APP_CONFIG"
 fi
-if [ ! -n "$APP_CONFIG" ]
-  then
-    echo "Not using custom app config"
+
+if [ -f /usr/share/nginx/html${PUBLIC_URL}app-config.js ]; then
+  if [ -s /usr/share/nginx/html${PUBLIC_URL}app-config.js ]; then
```
IbrahimCSAE, line 19: check if empty

.docker/Viewer-v3.x/entrypoint.sh
```diff
@@ -7,20 +7,30 @@ if [ -n "$SSL_PORT" ]
...
-  touch /usr/share/nginx/html${PUBLIC_URL}app-config.js
+  echo "Using custom APP_CONFIG environment variable"
+else
+  echo "Not using custom APP_CONFIG"
 fi
-if [ ! -n "$APP_CONFIG" ]
-  then
-    echo "Not using custom app config"
+
+if [ -f /usr/share/nginx/html${PUBLIC_URL}app-config.js ]; then
```
IbrahimCSAE, line 18: check if exists

wayfarer3130:
Thank you for fixing that.  
Could you update the documentation for the PUBLIC_URL to indicate that it is no longer necessary or desirable to update the routerBasename variable?    That was also changed as part of the base issue here, and would be good to address.

IbrahimCSAE:
> Thank you for fixing that. Could you update the documentation for the PUBLIC_URL to indicate that it is no longer necessary or desirable to update the routerBasename variable? That was also changed as part of the base issue here, and would be good to address.

Yes

sedghi (approved):
Thanks

sedghi:
I merged too quick sorry

PR #4688, opened 2025-01-14
feat/new customization service fix ss

sedghi:
https://deploy-preview-4688--ohif-platform-docs.netlify.app/platform/services/customization-service/customizationService


extensions/default/src/getCustomizationModule.tsx
```diff
@@ -29,315 +30,267 @@ export default function getCustomizationModule({ servicesManager, extensionManag
...
     {
       name: 'datasources',
-      merge: 'Append',
       value: {
-        id: 'customRoutes',
-        routes: [
-          {
-            path: '/datasources',
-            children: DataSourceSelector,
+        customRoutes: {
```
wayfarer3130, line 50: Could we get better naming on this - even if we don't do it now, at least it should be documented that in the next major release/version that this will be being renamed to the standardized naming.

extensions/default/src/getCustomizationModule.tsx
```diff
@@ -29,315 +30,267 @@ export default function getCustomizationModule({ servicesManager, extensionManag
...
-              {
-                id: 'applyDefaultProtocol',
-                label: 'Default',
-                commands: [
-                  'loadStudy',
-                  {
-                    commandName: 'setHangingProtocol',
-                    commandOptions: {
-                      protocolId: 'default',
+        'studyBrowser.studyMenuItems': {
```
wayfarer3130, line 65: Could we import complex items like this one from a study browser import so that they are defined alongside other study browser content?  I don't like having a mish mash of different things in this file.  It does need to bring things together, but that could be done effectively by using imports and treating the customizations basically like an export set of stuff.

extensions/default/src/getCustomizationModule.tsx
wayfarer3130, line 1: If the customization modules all use imports of the longer/complex definitions, then the  file naming can be left as getCustomizationModule.ts, and it means the file basically looks like an import/export module rather than having a mix of a whole bunch of different things.  It also allows localizing the definitions for things like menus etc into their own directory/set of files.

platform/docs/docs/platform/services/customization-service/customizationService.md
wayfarer3130, line 1: Could there be a section that explains the recommended naming?  And the migration path to get there?
I'd like to suggestion:

HighLevelComponent(.SubComponent)?.customizationItem

so that we would have, for example:
Routes.PrivateRoutes.customRoutes = [... list of routes to add as private routes]
or
OhifCornerstoneViewport.ContextMenu.rightClickMenu

I know we aren't there right now, but I think with this PR we should at least have a plan as to how we are going to get there.

Then, if we have a way to alias things, then we can start working towards getting there.

platform/core/src/services/CustomizationService/CustomizationService.ts
wayfarer3130, line 1: The reason that the right click context menus no longer were is that the new setup doesn't have a way to provide the final transforms for customizations.  For the context menus this takes a variety of different components and generates a function out of it.  These transforms are cached as they might take a while to apply, but they are done very late, just before returning values in order to have all the changes applied.  I think we need to discuss how to define that

extensions/default/src/getCustomizationModule.tsx
sedghi: As discussed "Routes.CustomRoutes"

modes/preclinical-4d/src/index.tsx
```diff
@@ -62,36 +62,19 @@ function modeFactory({ modeConfiguration }) {
...
-          mode: 'expanded',
+      customizationService.setCustomizations({
+        'panelSegmentation.tableMode': {
+          $set: 'expanded',
         },
-        {
-          id: 'PanelSegmentation.onSegmentationAdd',
-          onSegmentationAdd: () => {
+        'panelSegmentation.onSegmentationAdd': {
+          $set: () => {
```
wayfarer3130, line 70: A simple assignment as a function should require the $set

platform/core/src/services/CustomizationService/CustomizationService.test.js
wayfarer3130, line 1: Could you add a test that performs customizations using a second default, simulating extensions that enhance the previous extension value, for example:
extension 1 - default value
   simpleList: [1,2,3],

extension 2 - default value
    simpleList: { $push: [4,5] }

platform/core/src/services/CustomizationService/CustomizationService.test.js
```diff
@@ -1,313 +1,399 @@
-import CustomizationService, { CustomizationType, MergeEnum } from './CustomizationService';
-import log from '../../log';
-
-jest.mock('../../log.js', () => ({
-  info: jest.fn(),
-  warn: jest.fn(),
-  error: jest.fn(),
-}));
+// File: CustomizationService.registrationAndOperations.test.js
+import CustomizationService, { CustomizationScope } from './CustomizationService';
```
wayfarer3130, line 2: Can you add a test for:
* inheritance of values
* transformed customization
* register of set of values in default, where both sets gets applied - this can be used to apply different initial configuration sets

platform/core/src/services/CustomizationService/CustomizationService.ts
```diff
@@ -235,292 +173,333 @@ export default class CustomizationService extends PubSubService {
...
-          }
-        } else {
-          newList.splice(position, 0, value);
-        }
-      }
-      return newList;
-    }
-    return obj.concat(src);
+/** Add custom $filter command */
+extend('$filter', (query, original) => {
```
wayfarer3130, line 418: Is there some way to use the customization service itself to extend the customization service $ matches?

platform/core/src/services/CustomizationService/CustomizationService.ts
```diff
@@ -235,292 +173,333 @@ export default class CustomizationService extends PubSubService {
...
-   * type into the new type, allowing default behaviour to be configured.
+   * type into the new type, allowing default behavior to be configured.
    */
   public transform(customization: Customization): Customization {
     if (!customization) {
       return customization;
     }
-    const { customizationType } = customization;
-    if (!customizationType) {
+    const { inheritsFrom } = customization;
```
wayfarer3130, line 273: Could we make it $inheritsFrom to agree with the other transforms?  I know that means that the hasDollar needs to change, but that one should really check whether that particular operation exists anyways, so that would work generally.

platform/core/src/services/CustomizationService/CustomizationService.ts
```diff
@@ -235,292 +173,333 @@ export default class CustomizationService extends PubSubService {
...
+    if (!inheritsFrom) {
       return customization;
     }
-    const parent = this.getCustomization(customizationType);
+    const parent = this.getCustomization(inheritsFrom);
     const result = parent ? Object.assign({}, parent, customization) : customization;
     // Execute an nested type information
     return result.transform?.(this) || result;
```
wayfarer3130, line 280: Can we make it $transform to agree with the updating naming?

platform/docs/docs/platform/services/customization-service/sampleCustomizations.tsx
```diff
@@ -0,0 +1,1154 @@
...
+import thumbnailMenuItemsImage from '../../../assets/img/thumbnailMenuItemsImage.png';
+import studyMenuItemsImage from '../../../assets/img/studyMenuItemsImage.png';
+
+export const viewportOverlayCustomizations = [
+  {
+    id: 'viewportOverlay.topRight',
+    description: 'Defines the items displayed in the top-right overlay of the viewport.',
+    default: [],
+    configuration: `
+window.config = {
```
wayfarer3130, line 24: Shouldn't these be window.config.customizationService = [...]
Otherwise you replace the entire window.config, which seems a bit excessive.

platform/docs/docs/platform/services/customization-service/customizationService.md
```diff
@@ -0,0 +1,467 @@
...
+         name: 'default',
+         value: {
+           defaultList: ['Item A', 'Item B'],
+         },
+       },
+     ];
+   }
+   ```
+
+   By naming it `default`, it is automatically registered.
```
wayfarer3130, line 58: Also explain that by using other names, that set of customizations can be registered as an entire set by refererring to the entire name - that is the point of the register commands, so that something like:
{
  name: 'myCustomizationSet',
  value: {
    ... customizations here
  }
}
could be applied by registering the customizations ['myCustomizationSet']
That is also intended to allow nested named sets to be registered, eg by referring to another set within that set.

platform/docs/docs/platform/services/customization-service/customizationService.md
```diff
@@ -0,0 +1,467 @@
...
+
+### Priority of Scopes
+When a customization is retrieved:
+1. **Global Scope**: Takes precedence if defined.
+2. **Mode Scope**: Used if no global customization is defined.
+3. **Default Scope**: Fallback when neither global nor mode-specific values are available.
+
+
+As you have guessed the `.setCustomizations` accept a second argument which is the scope. By default it is set to `mode`.
+
```
wayfarer3130, line 223: Can you add a section in here explaining how to build up customization modules across multiple extensions?  For example, to add sop class handlers to a set of sop class handlers, or to add buttons to a list of buttons.
Then, show how the mode can "register" those during the mode startup, or use the compound registrations to do things.

platform/docs/docs/platform/services/customization-service/customizationService.md
```diff
@@ -0,0 +1,467 @@
...
+
+## Customization Syntax
+
+
+The customization syntax is designed to offer **flexibility** when modifying configurations. Instead of simply replacing values, you can perform granular updates like appending items to arrays, inserting at specific indices, updating deeply nested fields, or applying filters. This flexibility ensures that updates are efficient, targeted, and suitable for complex data structures.
```
wayfarer3130, line 228: As part of the customization syntax, can you explain inherit and transform?

platform/ui-next/src/components/DataRow/DataRow.tsx
```diff
@@ -261,25 +261,28 @@ const DataRow: React.FC<DataRowProps> = ({
           </Button>
 
           {/* Lock Icon (if needed) */}
```
wayfarer3130, line 263: I'm working on replacing the entire DataRow with customizeable pieces to allow customizing the menu, appearance etc.

wayfarer3130 (changes requested):
Mostly documentation and unit tests.

platform/docs/docs/platform/services/customization-service/customizationService.md
sedghi: I thought i did, but i will expand on it

modes/preclinical-4d/src/index.tsx
sedghi: I know it involves more boilerplate code, but this approach consistently handles both simple cases and really complex deep nestings. I'd rather not have a long explanation of different setup methods and instead stick to one consistent way to `set` (replace). 

Basically the message to the user is if you want to customize use a $ operator.

platform/core/src/services/CustomizationService/CustomizationService.ts
sedghi: I'm not sure if that's an operator like others or not. What makes you think it is an operator? To me, it looks like a referenced id, not something that performs an operation. I changed the $transform though

wayfarer3130 (commented):
I think I'm happy with this now, once hte conflicts are resolved and the tests complete.

PR #4709, opened 2025-01-18
feat(panels): responsive thumbnails based on panel size

IbrahimCSAE:
### Context

Building on top of @jbocce Joe's PR to make thumbnails respond to panel resizing

### Thumbnail Items only

https://github.com/user-attachments/assets/bcbbb292-ba83-4070-a500-0464214fa580

### List Items only

https://github.com/user-attachments/assets/096bcc61-0e5b-435d-83cb-21fbdd8066c5

### Mix of thumbnail items and list items

https://github.com/user-attachments/assets/f8511de6-10fe-4cf6-9307-1a792d64ad20


dan-rukas:
@IbrahimCSAE These behaviors work great especially for the mixed view. They align with each other well as the panel resizes and components reflow in the best way possible (best compromise overall). 

Thanks for taking a detailed look!

PR #4711, opened 2025-01-20
WIP - feat: group measurements component

pedrokohler:
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone/src/getPanelModule.tsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
-
 import { Toolbox } from '@ohif/ui-next';
+
 import PanelSegmentation from './panels/PanelSegmentation';
 import ActiveViewportWindowLevel from './components/ActiveViewportWindowLevel';
-import PanelMeasurement from './panels/PanelMeasurement';
+import PanelMeasurement from './panels/PanelMeasurementWithNodeComponent';
```
wayfarer3130, line 6: Can we avoid renaming this?  I agree the new name is more descriptive, but avoiding the renaming makes merges cleaner overall, and people get used to specific names.

extensions/cornerstone/src/panels/NodeComponent.tsx
wayfarer3130, line 1: Lets talk about where to put this, but I believe it belongs in ui-next as it is a generic component.  It is also going to need documentation on how to use it, particularly as it is a slightly higher level component than most which are used.

extensions/cornerstone/src/components/CollapsibleStudySummaryFromMetadata.tsx
wayfarer3130, line 1: The file name/export name don't match, and this looks like something that could be in ui-next as a generic component.

platform/core/src/utils/measurementGroupings.ts
```diff
@@ -0,0 +1,43 @@
...
+      items.filter(item => {
+        const itemStudyInstanceUID = getItemStudyInstanceUID(item);
+        return itemStudyInstanceUID === studyInstanceUID;
+      })
+    );
+  };
+
+export const groupIntoSingleGroup = () => items => [items];
+
+export const groupByStudyDeprecated =
```
wayfarer3130, line 28: Why is there a deprecated version?

platform/app/cypress/support/aliases.js
```diff
@@ -26,7 +26,7 @@ export function initCommonElementsAliases(skipMarkers) {
   cy.get('[data-cy="trackedMeasurements-btn"]').click();
 
   // TODO: Panels are not in DOM when closed, move this somewhere else
-  cy.get('[data-cy="trackedMeasurements-panel"]').as('measurementsPanel');
```
wayfarer3130, line 29: This change should test the trackedMeasurements-panel here.  If that means passing down an extra parameter that indicates the panel name, then it will have to be done.  That also allows different panels to have different names.

platform/core/src/utils/measurementGroupings.ts
pedrokohler: This is because I had to change this groupByStudy from the preivou PR... so I renamed it depecrated. But in the end this will be removed, since the changes in the original panel from last PR don't really make sense anymore.

So this won't be merged. I'll remove this in the end.

platform/core/src/utils/measurementGroupings.ts
wayfarer3130, line 1: Can you create a function that groups by measurements/additional findings, always returning 2 groups with those two items in it.
Also, a function which groups by series instance uid, which does not return empty groups

extensions/cornerstone/src/panels/PanelMeasurementWithNodeComponent.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+export default function PanelMeasurement({
+  servicesManager,
+  commandsManager,
+  measurementFilter = filterAny,
+}: withAppAndFilters): React.ReactNode {
+  const { measurementService } = servicesManager.services;
+
+  const measurementsPanelRef = useRef(null);
+  const additionalFilter = filterAdditionalFindings(measurementService);
```
wayfarer3130, line 26: The additional filter should be part of the grouping function for the nested child element.

PR #4712, opened 2025-01-20
[WIP] feat: Start of being able to group findings table

wayfarer3130:
### Context

When there are lots of measurements/contours, a better organization is required than simply dumping all the items into a list.  This PR is the start of being able to group items in the measurements table.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/extensions/cornerstone/src/getPanelModule.tsx
+++ b/extensions/cornerstone/src/getPanelModule.tsx
@@ -3,7 +3,7 @@ import React from 'react';
 import { Toolbox } from '@ohif/ui-next';
 import PanelSegmentation from './panels/PanelSegmentation';
 import ActiveViewportWindowLevel from './components/ActiveViewportWindowLevel';
-import PanelMeasurement from './panels/PanelMeasurement';
+import PanelMeasurement from './panels/PanelMeasurementConfigurable';
 
 const getPanelModule = ({ commandsManager, servicesManager, extensionManager }: withAppTypes) => {
   const wrappedPanelSegmentation = ({ configuration }) => {
--- a/extensions/cornerstone/src/panels/AccordionGroup.tsx
+++ b/extensions/cornerstone/src/panels/AccordionGroup.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import ShowItem from './ShowItem';
+
+export default function AccordionGroup(props) {
+  const { grouping, component: Component, componentProps, items } = props;
+  const groups = grouping.groupingFunction(items);
+  console.log('Groups=', groups);
+  return [...groups.entries()].map(([key, group]) => (
+    <>
+      <ShowItem
+        {...props}
+        {...componentProps}
+        items={group.items}
+        key={key}
+        group={group}
+      />
+    </>
+  ));
+}
--- a/extensions/cornerstone/src/panels/PanelMeasurementConfigurable.tsx
+++ b/extensions/cornerstone/src/panels/PanelMeasurementConfigurable.tsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import { useMeasurements } from '../hooks/useMeasurements';
+import StudyMeasurements from './StudyMeasurements';
+
+export default function PanelMeasurement(props): React.ReactNode {
+  const {
+    servicesManager,
+    measurementFilter,
+    content: Content = StudyMeasurements,
+    contentProps,
+  } = props;
+
+  const displayMeasurements = useMeasurements(servicesManager, {
+    measurementFilter,
+  });
+
+  // Need to merge defaults on the content props to ensure they get passed to hcildren
+  return (
+    <Content
+      {...props}
+      {...contentProps}
+      items={displayMeasurements}
+    />
+  );
+}
--- a/extensions/cornerstone/src/panels/ShowItem.tsx
+++ b/extensions/cornerstone/src/panels/ShowItem.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import { MeasurementTable } from '@ohif/ui-next';
+
+export default function ShowItem(props) {
+  const { title, items, group } = props;
+
+  return (
+    <MeasurementTable
+      title={title ? title : `Measurements`}
+      data={items}
+    >
+      <MeasurementTable.Header />
+      <MeasurementTable.Body />
+    </MeasurementTable>
+  );
+}
--- a/extensions/cornerstone/src/panels/StudyMeasurements.tsx
+++ b/extensions/cornerstone/src/panels/StudyMeasurements.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import AccordionGroup from './AccordionGroup';
+import ShowItem from './ShowItem';
+/**
+ * Groups measurements by study in order to allow display and saving by study
+ * @param {Object} servicesManager
+ */
+export const groupByStudy = ({ servicesManager }) => {
+  const { displaySetService } = servicesManager.services;
+
+  const getItemStudyInstanceUID = item => {
+    const displaySet = displaySetService.getDisplaySetByUID(item.displaySetInstanceUID);
+    return displaySet.instances[0].StudyInstanceUID;
+  };
+
+  return items => {
+    const groups = new Map();
+    items.forEach(item => {
+      const studyUID = getItemStudyInstanceUID(item);
+      if (!groups.has(studyUID)) {
+        groups.set(studyUID, {
+          items: [],
+        });
+      }
+      groups.get(studyUID).items.push(item);
+    });
+
+    return groups;
+  };
+};
+
+export default function StudyMeasurements(props): React.ReactNode {
+  const { items, title = 'Measurements', grouping = {} } = props;
+
+  // Need to merge defaults on the content props to ensure they get passed to hcildren
+  return (
+    <AccordionGroup
+      grouping={{
+        groupingFunction: grouping.groupingFunction || groupByStudy(props),
+        component: ShowItem,
+        componentProps: { ...props, ...grouping.componentProps },
+      }}
+      items={items}
+    />
+  );
+}


extensions/cornerstone-dicom-sr/src/utils/hydrateStructuredReport.ts
```diff
@@ -215,10 +215,14 @@ export default function hydrateStructuredReport(
         dataSource
       );
 
-      commandsManager.runCommand('updateMeasurement', {
-        uid: newAnnotationUID,
-        code: annotation.data.finding,
-      });
+      const style = annotation.data.finding?.style || annotation.data.site?.style;
```
wayfarer3130, line 218: This is fixing a bug on rehydrating in the basic-test mode that was being seen when testing the rest of the layout.

extensions/cornerstone/src/components/ColorCircle.tsx
wayfarer3130, line 1: Should this go in ui-next?

extensions/default/src/utils/promptSaveReport.js
wayfarer3130, line 1: Uses the new functionality for both tracked and untracked saving of measurements.

platform/app/public/config/kheops.js
wayfarer3130, line 1: Just fixing this so it starts up by default and can be used as a good test for dev:fast

platform/core/src/hooks/useActiveViewportDisplaySets.ts
wayfarer3130, line 1: This was completely broken previously and wouldn't update the view correclty because it ended up with multiple calls with undefined.

platform/core/src/services/MeasurementService/MeasurementService.ts
wayfarer3130, line 1: Allow array of strings for specifying measurements instead of a single string.

platform/core/src/services/ViewportGridService/ViewportGridService.ts
wayfarer3130, line 1: This was causing excessive re-renders on load.

platform/ui-next/src/index.ts
wayfarer3130, line 1: This change means we don't need to add the components in multiple places, which often seems to get missed.

platform/core/src/services/MeasurementService/MeasurementService.ts
```diff
@@ -570,7 +573,10 @@ class MeasurementService extends PubSubService {
    *
    * @param {string} measurementUID The measurement uid
    */
-  remove(measurementUID: string): void {
+  remove(measurementUID: string | string[]): void {
```
sedghi, line 576: can you add a new method removeMany instead?

sedghi: I see you've also included that below for visibility. I think this could really confuse the API and potentially cause issues when the function name doesn't match the parameter. I'd prefer the panel, toolbar, or command to simply loop over the UIDs and run them, or perhaps use a concrete `removeMany` or `toggleVisibilityMany` definition for new methods.

extensions/cornerstone/src/utils/getActiveViewportEnabledElement.ts
```diff
@@ -2,8 +2,15 @@ import { getEnabledElement } from '@cornerstonejs/core';
...
 
 import { getEnabledElement as OHIFgetEnabledElement } from '../state';
+import { ViewportGridService } from '@ohif/core';
+
+export default function getActiveViewportEnabledElement(viewportGridService: ViewportGridService, viewportId: string): IEnabledElement {
+  if (viewportId) {
+    const { element } = OHIFgetEnabledElement(viewportId) || {};
+    const enabledElement = getEnabledElement(element);
+    return enabledElement;
+  }
```
sedghi, line 13: I am not a fan of this either, for the same reason: the function name doesn't really match what it does. If `viewportId` is known, a simple `cornerstone.getEnabledElementByViewportId(viewportId)` does the job.
we don't need to edit this function

extensions/cornerstone/src/getPanelModule.tsx
```diff
@@ -84,6 +81,13 @@ const getPanelModule = ({ commandsManager, servicesManager, extensionManager }:
       label: 'Measurement',
       component: wrappedPanelMeasurement,
     },
+    {
+      name: 'panelMeasurementSeries',
+      iconName: 'tab-linear',
+      iconLabel: 'Measure',
+      label: 'Measurement Series',
+      component: wrappedPanelMeasurement,
+    },
```
sedghi, line 90: this seems to be duplicate of the item beore

platform/ui-next/src/components/DataRow/DataRow.tsx
```diff
@@ -178,25 +144,20 @@ const DataRow: React.FC<DataRowProps> = ({
...
-      ref={rowRef}
-      className={`flex flex-col ${isVisible ? '' : 'opacity-60'}`}
-    >
+    <div className={`flex flex-col ${isVisible ? '' : 'opacity-60'}`}>
       <div
         className={`flex items-center ${
           isSelected ? 'bg-popover' : 'bg-muted'
         } group relative cursor-pointer`}
-        onClick={onSelect}
+        onClick={e => onAction(e, 'jumpToMeasurement')}
```
sedghi, line 160: I don't think these changes make sense. `DataRow` is a general row-wise visualization that we use for measurements *and* segmentations, and potentially contours in the future. So, it really shouldn't mention `jumpToMeasurement` or `toggleVisibilityMeasurement`, or anything like that.

Also you have removed the `rowRef` which we need

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -1403,9 +1421,6 @@ function commandsModule({
     updateMeasurement: {
       commandFn: actions.updateMeasurement,
     },
-    clearMeasurements: {
-      commandFn: actions.clearMeasurements,
-    },
     jumpToMeasurement: {
       commandFn: actions.jumpToMeasurement,
```
sedghi, line 1425: why `clearMeasurements` is removed, i see it is being used

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -397,6 +395,26 @@ function commandsModule({
       viewports.forEach((_, index) => cineService.setCine({ id: index, isPlaying: false }));
     },
 
+    navigateToMeasurementItems: props => {
```
sedghi, line 398: Could you provide the documentation, please? I'm not really sure what `measurementItems` are. Also, what does `navigate` mean in this context? Does it mean to jump to something?

extensions/cornerstone/src/panels/PanelMeasurement.tsx
```diff
@@ -1,107 +1,58 @@
...
+import React from 'react';
+import { useSystem } from '@ohif/core';
 
-  const { measurementService } = servicesManager.services;
+import { useMeasurements } from '../hooks/useMeasurements';
+import StudyMeasurements from '../components/StudyMeasurements';
 
-  const displayMeasurements = useMeasurements(servicesManager, {
+export default function PanelMeasurement(props): React.ReactNode {
+  const {
```
sedghi, line 8: Use `useSystem` to get the managers, etc.

extensions/cornerstone/src/components/StudySummaryMenu.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { StudySummaryFromMetadata } from './StudySummaryFromMetadata';
+import StudyMeasurementMenu from './StudyMeasurementMenu';
+
+export default function StudySummaryMenu(props) {
```
sedghi, line 5: What does "Menu" mean in this `StudySummaryMenu`? Shouldn't it be called StudySummaryWithExports or something like that?

extensions/cornerstone/src/components/StudyMeasurementMenu.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { Button, Icons } from '@ohif/ui-next';
+import { useSystem } from '@ohif/core';
+
+export default function StudyMeasurementMenu({ items, StudyInstanceUID, measurementFilter }) {
```
sedghi, line 5: StudyMeasurementsActions?

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
```diff
@@ -23,67 +22,92 @@ function PanelMeasurementTableTracking({
...
-                  Delete All
-                </Button>
+        emptyComponent={EmptyComponent}
+        componentProps={{
+          grouping: {
+            header: props => (
+              <div>
+                <StudySummaryFromMetadata {...props} />
+                <CustomMenu {...props} />
               </div>
```
sedghi, line 106: This seems to be the same as StudySummaryWithActions (currently named StudySummaryMenu).

sedghi (commented):
I appreciate the ability to change the grouping in the measurement panels. I think you're definitely on the right track with giving the component this capability.

If I understand correctly (since there's no PR description—please update it with usage instructions), one needs to do this:

```jsx
 <PanelMeasurement
        measurementFilter={measurementFilter}
        component={MyGroupingComponent}
        componentProps={{
          grouping: {
            header: props => (
              <div>
                <StudySummaryFromMetadata {...props} />
                <CustomMenu {...props} />
              </div>
            ),
          },
        }}
      />
```

This defaults to `StudyMeasurements`;  looking at the `StudyMeasurements`, I was expecting to see the previous implementation with the `MeasurementTable` containing a Header and Body. Instead, I noticed a newly implemented `AccodionGroup` with `Accodions` nested within `Accordions`, and so on.

I'm not particularly concerned about other groupings and their UI. In fact, perhaps we should extract them out to the extension-test so we're not responsible for designing and maintaining them.

The current implementation has some UI regressions that should be fixed.


CleanShot 2025-03-06 at 22 55 09@2x


---

The new components getting added should follow our other component designs (code) and use composition with context and providers over props. 


The following code is 

```jsx
<PanelMeasurement
  servicesManager={servicesManager}
  measurementFilter={measurementFilter}
  emptyComponent={EmptyComponent}
  componentProps={{
    grouping: {
      header: props => (
        <div>
          <StudySummaryFromMetadata {...props} />
          <CustomMenu {...props} />
        </div>
      ),
    },
  }}
/>
```


Should in fact be something like 



```jsx
<PanelMeasurement
  servicesManager={servicesManager}
  measurementFilter={measurementFilter}
  emptyComponent={EmptyComponent}
>
  <PanelMeasurement.Header>
    <StudySummaryFromMetadata />
    <CustomMenu />
  </PanelMeasurement.Header>
  <PanelMeasurement.Body />
</PanelMeasurement>
```




and also 

```js
   <AccordionGroup
      grouping={{
        name: 'groupByStudy',
        groupingFunction: groupByStudy,
        header: StudySummaryMenu,
        activeStudyUID,
        ...grouping,
      }}
      items={items}
      value={[activeStudyUID]}
    >
      <StudyMeasurementItem
        activeStudyUID={activeStudyUID}
        children={children}
      />
    </AccordionGroup>
```

Having `header` inside `grouping` is strange, and `grouping = {{ ...grouping }}` is also strange.

I think the new SegmentationTable tries to follow the new pattern

platform/core/src/services/MeasurementService/MeasurementService.ts
wayfarer3130: Definitely want to have a method to do it in the measurement service since it might send notifications in different ways based on how many things are getting changed and that needs to be known about in that method.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: It is duplicated here and in the sr module.  I don't really care which one we leave, but there should only be one.

extensions/cornerstone/src/components/AccordionGroup/AccordionGroup.tsx
wayfarer3130, line 1: @sedghi - I've moved to contained elements for including the children.  Can you take a look at this file and StudyMeasurements to see how it gets used and whether that is the right direction?  Hoping to complete this this week.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -496,16 +501,18 @@ function commandsModule({
...
       const labelConfig = customizationService.getCustomization('measurementLabels');
       const renderContent = customizationService.getCustomization('ui.labellingComponent');
 
-      callInputDialogAutoComplete({
+      const value = await callInputDialogAutoComplete({
         uiDialogService,
         labelConfig,
         renderContent,
       });
+      callback?.(value);
```
wayfarer3130, line 513: These changes allow the arrow annotation to work.

extensions/cornerstone/src/components/AccordionGroup/AccordionGroup.tsx
```diff
@@ -0,0 +1,192 @@
...
+  if (!group) {
+    return null;
+  }
+  if (asChild) {
+    return React.cloneElement(children, childProps);
+  }
+  return (
+    <AccordionTrigger
+      value={group.value}
+      asChild={true}
```
wayfarer3130, line 178: The asChild prevents the button within button exception.

extensions/cornerstone/src/components/MeasurementTableNested.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <MeasurementTable
+      title={title ? title : `Measurements`}
+      data={items}
+      onAction={onAction}
+      {...group}
+      key={group.key}
+    >
+      <MeasurementTable.Header key="measurementTableHeader">
+        {customHeader && group.isFirst && customHeader({ ...props, items: props.allItems })}
```
wayfarer3130, line 25: This bit shows the overall measurement menu, and fixes the bug you asked me to fix.

sedghi (approved):
Awesome addition

PR #4723, opened 2025-01-22
feat(panels): responsive thumbnails based on panel size 

IbrahimCSAE:
### Context

Building on top of @jbocce Joe's PR to make thumbnails respond to panel resizing

### Thumbnail Items only

https://github.com/user-attachments/assets/bcbbb292-ba83-4070-a500-0464214fa580

### List Items only

https://github.com/user-attachments/assets/096bcc61-0e5b-435d-83cb-21fbdd8066c5

### Mix of thumbnail items and list items

https://github.com/user-attachments/assets/f8511de6-10fe-4cf6-9307-1a792d64ad20

### DIFF
--- a/extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
+++ b/extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
@@ -238,13 +238,19 @@ const useResizablePanels = (
   }, [setLeftPanelClosed]);
 
   const onLeftPanelOpen = useCallback(() => {
-    resizableLeftPanelAPIRef?.current?.expand();
-    if (!isResizableHandleDraggingRef.current) {
-      setMinMaxWidth(
-        resizableLeftPanelElemRef.current,
-        leftPanelExpandedWidth + panelGroupDefinition.shared.expandedInsideBorderSize
-      );
-    }
+    const panelGroupElem = resizablePanelGroupElemRef.current;
+    const { width: panelGroupWidth } = panelGroupElem.getBoundingClientRect();
+
+    const expandedSize = Math.max(
+      leftPanelExpandedWidth + panelGroupDefinition.shared.expandedInsideBorderSize,
+      panelGroupDefinition.left.minimumExpandedOffsetWidth
+    );
+
+    const expandedPercentage = (expandedSize / panelGroupWidth) * 100;
+
+    resizableLeftPanelAPIRef?.current?.expand(expandedPercentage);
+    setMinMaxWidth(resizableLeftPanelElemRef.current, expandedSize);
+
     setLeftPanelClosed(false);
   }, [leftPanelExpandedWidth, setLeftPanelClosed]);
 
@@ -254,9 +260,14 @@ const useResizablePanels = (
     }
 
     const { width: panelGroupWidth } = resizablePanelGroupElemRef.current.getBoundingClientRect();
-    setLeftPanelExpandedWidth(
-      (size / 100) * panelGroupWidth - panelGroupDefinition.shared.expandedInsideBorderSize
-    );
+    const newExpandedWidth =
+      (size / 100) * panelGroupWidth - panelGroupDefinition.shared.expandedInsideBorderSize;
+
+    setLeftPanelExpandedWidth(newExpandedWidth);
+
+    if (isResizableHandleDraggingRef.current) {
+      setMinMaxWidth(resizableLeftPanelElemRef.current, newExpandedWidth);
+    }
   }, []);
 
   const onRightPanelClose = useCallback(() => {
@@ -266,13 +277,19 @@ const useResizablePanels = (
   }, [setRightPanelClosed]);
 
   const onRightPanelOpen = useCallback(() => {
-    resizableRightPanelAPIRef?.current?.expand();
-    if (!isResizableHandleDraggingRef.current) {
-      setMinMaxWidth(
-        resizableRightPanelElemRef.current,
-        rightPanelExpandedWidth + panelGroupDefinition.shared.expandedInsideBorderSize
-      );
-    }
+    const panelGroupElem = resizablePanelGroupElemRef.current;
+    const { width: panelGroupWidth } = panelGroupElem.getBoundingClientRect();
+
+    const expandedSize = Math.max(
+      rightPanelExpandedWidth + panelGroupDefinition.shared.expandedInsideBorderSize,
+      panelGroupDefinition.right.minimumExpandedOffsetWidth
+    );
+
+    const expandedPercentage = (expandedSize / panelGroupWidth) * 100;
+
+    resizableRightPanelAPIRef?.current?.expand(expandedPercentage);
+    setMinMaxWidth(resizableRightPanelElemRef.current, expandedSize);
+
     setRightPanelClosed(false);
   }, [rightPanelExpandedWidth, setRightPanelClosed]);
 
@@ -280,10 +297,16 @@ const useResizablePanels = (
     if (resizableRightPanelAPIRef?.current?.isCollapsed()) {
       return;
     }
+
     const { width: panelGroupWidth } = resizablePanelGroupElemRef.current.getBoundingClientRect();
-    setRightPanelExpandedWidth(
-      (size / 100) * panelGroupWidth - panelGroupDefinition.shared.expandedInsideBorderSize
-    );
+    const newExpandedWidth =
+      (size / 100) * panelGroupWidth - panelGroupDefinition.shared.expandedInsideBorderSize;
+
+    setRightPanelExpandedWidth(newExpandedWidth);
+
+    if (isResizableHandleDraggingRef.current) {
+      setMinMaxWidth(resizableRightPanelElemRef.current, newExpandedWidth);
+    }
   }, []);
 
   return [
--- a/extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
+++ b/extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
@@ -34,7 +34,7 @@ const useResizablePanels = (
   );
   const [leftResizablePanelMinimumSize, setLeftResizablePanelMinimumSize] = useState(0);
   const [rightResizablePanelMinimumSize, setRightResizablePanelMinimumSize] = useState(0);
-  const [leftRizeablePanelCollapsedSize, setLeftResizePanelCollapsedSize] = useState(0);
+  const [leftResizeablePanelCollapsedSize, setLeftResizePanelCollapsedSize] = useState(0);
   const [rightResizePanelCollapsedSize, setRightResizePanelCollapsedSize] = useState(0);
 
   const resizablePanelGroupElemRef = useRef(null);
@@ -311,7 +311,7 @@ const useResizablePanels = (
       minSize: leftResizablePanelMinimumSize,
       onResize: onLeftPanelResize,
       collapsible: true,
-      collapsedSize: leftRizeablePanelCollapsedSize,
+      collapsedSize: leftResizeablePanelCollapsedSize,
       onCollapse: () => setLeftPanelClosed(true),
       onExpand: () => setLeftPanelClosed(false),
       ref: resizableLeftPanelAPIRef,
--- a/extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
+++ b/extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
@@ -60,20 +60,20 @@ const useResizablePanels = (
     const rightPanelElem = getPanelElement(panelGroupDefinition.right.panelId);
     resizableRightPanelElemRef.current = rightPanelElem;
 
-    // we use the initialExpandedOffsetWidth on the first render incase the panel has min width but we want the initial state to be larger than that
-    const leftResizablePanelExpandedSize =
-      (panelGroupDefinition.left.initialExpandedOffsetWidth / panelGroupWidth) * 100;
-    const rightResizablePanelExpandedSize =
-      (panelGroupDefinition.right.initialExpandedOffsetWidth / panelGroupWidth) * 100;
-
     // Since both resizable panels are collapsed by default (i.e. their default size is zero),
     // on the very first render check if either/both side panels should be expanded.
+    // we use the initialExpandedOffsetWidth on the first render incase the panel has min width but we want the initial state to be larger than that
+
     if (!leftPanelClosed) {
+      const leftResizablePanelExpandedSize =
+        (panelGroupDefinition.left.initialExpandedOffsetWidth / panelGroupWidth) * 100;
       resizableLeftPanelAPIRef?.current?.expand(leftResizablePanelExpandedSize);
       setMinMaxWidth(leftPanelElem, panelGroupDefinition.left.initialExpandedOffsetWidth);
     }
 
     if (!rightPanelClosed) {
+      const rightResizablePanelExpandedSize =
+        (panelGroupDefinition.right.initialExpandedOffsetWidth / panelGroupWidth) * 100;
       resizableRightPanelAPIRef?.current?.expand(rightResizablePanelExpandedSize);
       setMinMaxWidth(rightPanelElem, panelGroupDefinition.right.initialExpandedOffsetWidth);
     }
--- a/extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
+++ b/extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
@@ -1,6 +1,6 @@
 import { useState, useCallback, useLayoutEffect, useRef } from 'react';
 import { getPanelElement, getPanelGroupElement } from 'react-resizable-panels';
-import { panelGroupDefination } from './constants/panels';
+import { panelGroupDefinition } from './constants/panels';
 
 /**
  * Set the minimum and maximum css style width attributes for the given element.
@@ -27,13 +27,13 @@ const useResizablePanels = (
   setRightPanelClosed
 ) => {
   const [leftPanelExpandedWidth, setLeftPanelExpandedWidth] = useState(
-    panelGroupDefination.left.expandedDefaultWidth
+    panelGroupDefinition.left.initialExpandedWidth
   );
   const [rightPanelExpandedWidth, setRightPanelExpandedWidth] = useState(
-    panelGroupDefination.right.expandedDefaultWidth
+    panelGroupDefinition.right.initialExpandedWidth
   );
-  const [leftResizablePanelDefaultSize, setLeftResizePanelDefaultSize] = useState(0);
-  const [rightResizePanelDefaultSize, setRightResizePanelDefaultSize] = useState(0);
+  const [leftResizablePanelMinimumSize, setLeftResizablePanelMinimumSize] = useState(0);
+  const [rightResizablePanelMinimumSize, setRightResizablePanelMinimumSize] = useState(0);
   const [leftRizeablePanelCollapsedSize, setLeftResizePanelCollapsedSize] = useState(0);
   const [rightResizePanelCollapsedSize, setRightResizePanelCollapsedSize] = useState(0);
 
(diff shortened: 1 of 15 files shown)


modes/segmentation/src/index.tsx
```diff
@@ -132,7 +132,9 @@ function modeFactory({ modeConfiguration }) {
             id: ohif.layout,
```
IbrahimCSAE, line 132: I think this is all that is needed to enable it in any mode, you can set it in the layout template @wayfarer3130

platform/ui-next/src/components/ThumbnailList/ThumbnailList.tsx
```diff
@@ -11,19 +11,85 @@ const ThumbnailList = ({
...
       style={{
         '--radix-accordion-content-height': '350px',
       }}
     >
+      {/* Thumbnail Items */}
+      <div
+        id="ohif-thumbnail-list"
+        className="ohif-scrollbar bg-bkg-low grid grid-cols-[repeat(auto-fit,_minmax(0,135px))] place-items-start gap-[4px] overflow-y-hidden pt-[4px] pr-[2.5px] pl-[5px]"
+      >
+        {thumbnailItems.map(
```
wayfarer3130, line 36: Is there a reason not to do:
{thumbnailItems.map(item => (<Thumbnail {...item} ... any extra calculations here like onClick />

platform/ui-next/src/components/ThumbnailList/ThumbnailList.tsx
```diff
@@ -11,19 +11,85 @@ const ThumbnailList = ({
...
+                imageSrc={imageSrc}
+                imageAltText={imageAltText}
+                messages={messages}
+                isActive={isActive}
+                canReject={canReject}
+                onReject={onReject}
+                modality={modality}
+                viewPreset="thumbnails"
+                thumbnailType={componentType}
+                onClick={() => onThumbnailClick(displaySetInstanceUID)}
```
wayfarer3130, line 74: Consider onClick={onThumbnailClick.bind(null,displaySetInstanceUID)}
I think that is slightly more efficient than creating the extra function.

modes/segmentation/src/index.tsx
wayfarer3130: I just tested it in ohif.thumbnailList and yes, just using that now works.

extensions/default/src/ViewerLayout/constants/panels.ts
```diff
@@ -0,0 +1,30 @@
+const panelGroupDefination = {
```
jbocce, line 1: Typo... panelGroupDefinition

extensions/default/src/ViewerLayout/constants/panels.ts
```diff
@@ -0,0 +1,30 @@
...
+    collapsedOutsideBorderSize: 8,
+    collapsedWidth: 25,
+  },
+  left: {
+    // id
+    panelId: 'viewerLayoutResizableLeftPanel',
+    // expanded width
+    expandedDefaultWidth: 282,
+    // expanded width + expanded inside border
+    expandedDefaultOffsetWidth: 145 + 4,
```
jbocce, line 15: I don't understand why we have an `expandedDefaultOffsetWidth` and an `initialExpandedDefaultOffsetWidth`. Furthermore why is `expandedDefaultOffsetWidth` so small?

extensions/default/src/ViewerLayout/constants/panels.ts
```diff
@@ -0,0 +1,30 @@
...
+    // initial expanded width
+    initialExpandedDefaultOffsetWidth: 282 + 4,
```
jbocce, line 17: We should use declared constants instead of hardcoded values. If ever the border sizes change, we will have to change multiple lines of code. Not cool.

modes/segmentation/src/index.tsx
```diff
@@ -132,7 +132,9 @@ function modeFactory({ modeConfiguration }) {
...
             props: {
               leftPanels: [ohif.leftPanel],
+              leftPanelResizable: true,
```
jbocce, line 135: Do we want to make the default resizable? If so consider changing the corresponding ViewerLayout props to have a default value of true.

extensions/default/src/ViewerLayout/constants/panels.ts
jbocce: Ok. I think I understand. We have an initial expanded width and then a minimum width. Right? Let's rename them. Maybe something like...

`expandedDefaultWidth` rename to `initialExpandedWidth`
`expandedDefaultOffsetWidth` rename to `minimumExpandedOffsetWidth`
`initialExpandedDefaultOffsetWidth` rename to `initialExpandedOffsetWidth`

To me default and initial kinda mean the same thing. So initial and default in the same variable name was driving me crazy. 🤣

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
```diff
@@ -270,53 +281,55 @@ const useResizablePanels = (
...
+      expandedInsideBorderSize: panelGroupDefination.shared.expandedInsideBorderSize,
       onClose: onRightPanelClose,
       onOpen: onRightPanelOpen,
     },
-    { direction: 'horizontal', id: viewerLayoutResizablePanelGroupId },
+    { direction: 'horizontal', id: panelGroupDefination.groupId },
     {
-      defaultSize: resizablePanelDefaultSize,
-      minSize: resizablePanelDefaultSize,
+      defaultSize: leftResizablePanelDefaultSize,
```
jbocce, line 310: Set this to zero. It really is a useless prop for us.

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
```diff
@@ -270,53 +281,55 @@ const useResizablePanels = (
...
       ref: resizableLeftPanelAPIRef,
       order: 0,
-      id: viewerLayoutResizableLeftPanelId,
+      id: panelGroupDefination.left.panelId,
     },
     { order: 1, id: 'viewerLayoutResizableViewportGridPanel' },
     {
-      defaultSize: resizablePanelDefaultSize,
-      minSize: resizablePanelDefaultSize,
+      defaultSize: rightResizePanelDefaultSize,
```
jbocce, line 323: Set this to zero like the left one above.

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
```diff
@@ -270,53 +281,55 @@ const useResizablePanels = (
...
   }, []);
 
   return [
     {
       expandedWidth: leftPanelExpandedWidth,
-      collapsedWidth: sidePanelCollapsedWidth,
-      collapsedInsideBorderSize: sidePanelCollapsedInsideBorderSize,
-      collapsedOutsideBorderSize: sidePanelCollapsedOutsideBorderSize,
-      expandedInsideBorderSize: sidePanelExpandedInsideBorderSize,
+      collapsedWidth: panelGroupDefination.left.collapsedOffsetWidth,
```
jbocce, line 292: This was not an offset width before so I am surprised it is now. I think it should just be collapsedWidth.

jbocce:
I did a bit of testing and found some things...

1. Whenever there is enough space in the panel, we probably shouldn't have an ellipsis in the study description. Right?
image
2. When the left panel is collapsed, the thumbnails tab icon looks like it is too far to the right. This doesn't seem to happen for the right side panel.
image

jbocce:
I am a bit surprised that when the thumbnail panel has more room studies don't shift up. Instead there is empty space...

image

platform/ui-next/src/components/ThumbnailList/ThumbnailList.tsx
IbrahimCSAE: i like that, let me do it

platform/ui-next/src/components/ThumbnailList/ThumbnailList.tsx
IbrahimCSAE: like this too, will add it

extensions/default/src/ViewerLayout/constants/panels.ts
IbrahimCSAE: good catch thanks

extensions/default/src/ViewerLayout/constants/panels.ts
IbrahimCSAE: Haha i agree, the naming was confusing, I'll fix it now.

IbrahimCSAE:
> I did a bit of testing and found some things...
> 
> 1. Whenever there is enough space in the panel, we probably shouldn't have an ellipsis in the study description. Right?
>    image
> 2. When the left panel is collapsed, the thumbnails tab icon looks like it is too far to the right. This doesn't seem to happen for the right side panel.
>    image

fixed, thanks for calling those out

modes/segmentation/src/index.tsx
IbrahimCSAE: It's up to Alireza, I think the default should be true and you can just set it to false here, so we don't run into confusing scenarios like Bill did. @sedghi

extensions/default/src/ViewerLayout/constants/panels.ts
IbrahimCSAE: fixed

IbrahimCSAE:
> I am a bit surprised that when the thumbnail panel has more room studies don't shift up. Instead there is empty space...
> 
> image

This is also fixed

https://github.com/user-attachments/assets/fd57848d-42eb-4cfe-a22f-83d22323a0bf

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
IbrahimCSAE: Setting this to 0 throws errors in the console, so I'm going to set it to be the same as `minSize`

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
IbrahimCSAE: Setting this to 0 throws errors in the console, so I'm going to set it to be the same as `minSize`

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
```diff
@@ -270,53 +281,55 @@ const useResizablePanels = (
...
+      expandedInsideBorderSize: panelGroupDefinition.shared.expandedInsideBorderSize,
       onClose: onRightPanelClose,
       onOpen: onRightPanelOpen,
     },
-    { direction: 'horizontal', id: viewerLayoutResizablePanelGroupId },
+    { direction: 'horizontal', id: panelGroupDefinition.groupId },
     {
-      defaultSize: resizablePanelDefaultSize,
-      minSize: resizablePanelDefaultSize,
+      defaultSize: leftResizablePanelMinimumSize,
```
jbocce, line 310: I'm not sure this is what we want here. Please try the following scenario...

- initially (on start up) have the left panel collapsed (i.e. closed can be set in the mode definition)
- expand the left panel and check which size is it and is it the one we want; I personally would expect it to be the larger of the two widths (i.e. the initial expanded width)

I can help you with this if you want.

extensions/default/src/ViewerLayout/constants/panels.ts
```diff
@@ -0,0 +1,34 @@
...
+    // expanded width + expanded inside border
+    minimumExpandedOffsetWidth: 145 + expandedInsideBorderSize,
+    // initial expanded width
+    initialExpandedOffsetWidth: 282 + expandedInsideBorderSize,
+    // collapsed width + collapsed inside border + collapsed outside border
+    collapsedOffsetWidth: 25 + collapsedInsideBorderSize + collapsedOutsideBorderSize,
+  },
+  right: {
+    panelId: 'viewerLayoutResizableRightPanel',
+    initialExpandedWidth: 280,
```
jbocce, line 27: declare 280 as a constant too

modes/segmentation/src/index.tsx
sedghi: Let's make it resizable, these are our mode, we can do whatever we want. We should also apply similar changes to longitudinal and all other modes, thanks

extensions/default/src/ViewerLayout/constants/panels.ts
```diff
@@ -0,0 +1,34 @@
+const expandedInsideBorderSize = 4;
+const collapsedInsideBorderSize = 4;
+const collapsedOutsideBorderSize = 8;
+
+const panelGroupDefinition = {
```
sedghi, line 5: Is there really a need for this file, or can we simplify things by putting this code directly in ResiablePanelsHook? If it's used in multiple places, that's a pretty good reason to keep it as a separate file, but if not, it's kind of redundant.

platform/ui-next/src/components/StudyItem/StudyItem.tsx
```diff
@@ -4,6 +4,8 @@ import classnames from 'classnames';
 import { ThumbnailList } from '../ThumbnailList';
 
 import { Accordion, AccordionContent, AccordionItem, AccordionTrigger } from '../Accordion';
+import { Tooltip, TooltipContent } from '../Tooltip';
+import { TooltipTrigger } from '@radix-ui/react-tooltip';
```
sedghi, line 8: I think this is wrong, you should grab the TooltipTrigger from our UI not from radix, no?

platform/ui-next/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -246,11 +255,11 @@ const Thumbnail = ({
         className,
         'bg-muted hover:bg-primary/30 group flex cursor-pointer select-none flex-col rounded outline-none',
         viewPreset === 'thumbnails' && 'h-[170px] w-[135px]',
-        viewPreset === 'list' && 'col-span-2 h-[40px] w-[275px]'
+        viewPreset === 'list' && 'h-[40px] w-full'
       )}
       id={`thumbnail-${displaySetInstanceUID}`}
       data-cy={
-        thumbnailType === 'thumbnailNoImage'
+        componentType === 'thumbnailNoImage'
```
sedghi, line 262: is there really a need for this rename? it will be a breaking change that we need to document

platform/ui-next/src/components/ThumbnailList/ThumbnailList.tsx
```diff
@@ -89,15 +92,7 @@ ThumbnailList.propTypes = {
       description: PropTypes.string,
       componentType: PropTypes.any,
       isTracked: PropTypes.bool,
```
sedghi, line 94: please add the comment back

platform/ui-next/src/components/ThumbnailList/ThumbnailList.tsx
```diff
@@ -11,68 +11,71 @@ const ThumbnailList = ({
   activeDisplaySetInstanceUIDs = [],
   viewPreset,
   ThumbnailMenuItems,
-}: withAppTypes) => {
+}) => {
+  // Filter thumbnails into list items and thumbnail items
+  const listItems = thumbnails.filter(
```
sedghi, line 16: thumbnails?.filter

platform/ui-next/src/components/ThumbnailList/ThumbnailList.tsx
```diff
@@ -11,68 +11,71 @@ const ThumbnailList = ({
...
+    ({ componentType }) => componentType === 'thumbnailNoImage' || viewPreset === 'list'
```
sedghi, line 17: `thumbnailNoImage` and `list` makes sense to be CONSTs in a separate file

platform/ui-next/src/components/StudyBrowserSort/StudyBrowserSort.tsx
```diff
@@ -50,11 +51,17 @@ export function StudyBrowserSort({ servicesManager }: withAppTypes) {
...
   return (
-    <div className="flex items-center gap-1">
+    <div className="flex w-[50%] items-center gap-1">
       <DropdownMenu>
-        <DropdownMenuTrigger className="border-inputfield-main focus:border-inputfield-main flex h-[26px] w-[125px] items-center justify-start rounded border bg-black p-2 text-base text-white">
-          {selectedSort.label}
-        </DropdownMenuTrigger>
+        <Tooltip>
+          <TooltipTrigger className="w-full overflow-hidden">
+            {' '}
```
sedghi, line 58: what is this { ' ' }

platform/ui-next/src/components/StudyBrowserSort/StudyBrowserSort.tsx
```diff
@@ -50,11 +51,17 @@ export function StudyBrowserSort({ servicesManager }: withAppTypes) {
...
+            <DropdownMenuTrigger className="border-inputfield-main focus:border-inputfield-main flex h-[26px] w-full items-center justify-start overflow-hidden whitespace-nowrap rounded border bg-black p-2 text-base text-white">
+              {selectedSort.label}
+            </DropdownMenuTrigger>{' '}
```
sedghi, line 61: another {' ' }

platform/ui-next/src/components/StudyBrowserViewOptions/StudyBrowserViewOptions.tsx
```diff
@@ -5,6 +5,8 @@ import {
   DropdownMenuContent,
   DropdownMenuItem,
 } from '../DropdownMenu/DropdownMenu';
+import { Tooltip, TooltipContent } from '../Tooltip';
+import { TooltipTrigger } from '@radix-ui/react-tooltip';
```
sedghi, line 9: same

sedghi (commented):
Great stuff, see my comments thanks

extensions/default/src/ViewerLayout/constants/panels.ts
IbrahimCSAE: we discussed this on slack and we decided to keep it. I personally think it's too many lines and large enough to be separate.

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
```diff
@@ -43,15 +27,15 @@ const useResizablePanels = (
...
-    sidePanelExpandedDefaultWidth
+    panelGroupDefinition.right.initialExpandedWidth
   );
-
-  // Percentage sizes.
-  const [resizablePanelCollapsedSize, setResizablePanelCollapsedSize] = useState(0);
-  const [resizablePanelDefaultSize, setResizablePanelDefaultSize] = useState(0);
+  const [leftResizablePanelMinimumSize, setLeftResizablePanelMinimumSize] = useState(0);
+  const [rightResizablePanelMinimumSize, setRightResizablePanelMinimumSize] = useState(0);
+  const [leftRizeablePanelCollapsedSize, setLeftResizePanelCollapsedSize] = useState(0);
```
jbocce, line 37: *typo this should be leftResizablePanelCollapsedSize. I believe there is an extension for VS code that will help for spell checking. It is called Code Spell Checker. Comes in handy.

IbrahimCSAE: thanks spelling police

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
```diff
@@ -65,30 +49,33 @@ const useResizablePanels = (
...
     resizableLeftPanelElemRef.current = leftPanelElem;
 
-    const rightPanelElem = getPanelElement(viewerLayoutResizableRightPanelId);
+    const rightPanelElem = getPanelElement(panelGroupDefinition.right.panelId);
     resizableRightPanelElemRef.current = rightPanelElem;
 
-    const resizablePanelExpandedSize =
-      (sidePanelExpandedDefaultOffsetWidth / panelGroupWidth) * 100;
+    // we use the initialExpandedOffsetWidth on the first render incase the panel has min width but we want the initial state to be larger than that
+    const leftResizablePanelExpandedSize =
```
jbocce, line 64: This should be inside the first if statement below.

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
IbrahimCSAE: I see the problem, let me see if I can figure it out, thanks Joe

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
```diff
@@ -65,30 +49,33 @@ const useResizablePanels = (
...
+      (panelGroupDefinition.left.initialExpandedOffsetWidth / panelGroupWidth) * 100;
+    const rightResizablePanelExpandedSize =
```
jbocce, line 66: This should be inside the second if statement below.

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
jbocce: Actually I can cover this in a subsequent PR if you would like. We will have a PR for the resize handle and I can take care of it there.

IbrahimCSAE: I have something worked that i just pushed, if it doesn't work for you I can revert it and you can take this.

sedghi (approved):
Looks great thanks!

extensions/default/src/ViewerLayout/ResizablePanelsHook.tsx
```diff
@@ -245,7 +260,14 @@ const useResizablePanels = (
     }
 
     const { width: panelGroupWidth } = resizablePanelGroupElemRef.current.getBoundingClientRect();
-    setLeftPanelExpandedWidth((size / 100) * panelGroupWidth - sidePanelExpandedInsideBorderSize);
+    const newExpandedWidth =
+      (size / 100) * panelGroupWidth - panelGroupDefinition.shared.expandedInsideBorderSize;
+
+    setLeftPanelExpandedWidth(newExpandedWidth);
+
+    if (isResizableHandleDraggingRef.current) {
```
jbocce, line 268: I'm not sure this is right. We should be setting the min max during drag.

PR #4745, opened 2025-01-29
feat: Remove non-functional Tailwind class for SegmentationPanel

Celian-abd:
### Context
We removed the logic where the panel should take 200px by number of segments due to the fact that it can't work with Tailwind.

Writing something like `h-[${segmentCount * 200}px]` does not work because Tailwind does not know which number of segments to generate in advance.

### Changes & Results
Removed non-functional logic for dynamic height of the segmentation panel

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/platform/ui-next/src/components/SegmentationTable/SegmentationSegments.tsx
+++ b/platform/ui-next/src/components/SegmentationTable/SegmentationSegments.tsx
@@ -33,12 +33,9 @@ export const SegmentationSegments: React.FC<{
     return null;
   }
 
-  const segmentCount = Object.keys(representationToUse.segments).length;
-  const height = mode === 'collapsed' ? 'h-[900px]' : `h-[${segmentCount * 200}px]`;
-
   return (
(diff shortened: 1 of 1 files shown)


sedghi:
Can you show the before and after? what you were trying to achieve that didn't work?

Celian-abd:
Hey @sedghi,
In the following screenshot we can see that the panel has an height of 900 pixels but if we follow the code it should be 3*200px = 600px because we have 3 segments.
image

We can find again the same classNames but with the `h-[900px]` always there.
image

That's why we removed it because we thought that this part cannot work.

sedghi (approved):

salimkanoun:
Just to clarify, Tailwind builds the CSS by analyzing the static code, so all concatenation that would generate a talwind class name at execution time will not be included in the build css

PR #4727, opened 2025-01-23
fix(context menu): Context menus for edge-proximate measurements are partially obscured.

arul-trenser:
### Context
The issue that context menu is not fully visible near the edges of the viewport has be incorporated by [FlyWheel.io](https://flywheel.io/).
Closes #4726 

### Changes & Results
The context menu for edge-proximate measurements experiences clipping. A solution involves dynamically calculating the menu's position relative to the window's inner height and width to ensure full visibility.

#### PR

- [x] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [x] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version:  18.19.0
- [x] Browser: Chrome 131.0.6778.265
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #4760, opened 2025-02-03
fix(context-menu):  Fixing regression introduced by PR #4727

jbocce:
Fixing regression introduced by PR https://github.com/OHIF/Viewers/issues/4727

### Context

Longer/taller context menus are incorrectly opening from bottom to top and are getting clipped.

In most cases, taller (custom) context menus are opening "bottom-to-top" from the click point even though there is enough space to open them (the desired) "top-to-bottom". Furthermore, the menus are also getting clipped at the top of the browser window. See the screen shot below...

In this screen shot there is plenty of space to open the menu in its traditional sense. Yet, look at how it opens "bottom-to-top" and is clipped at the top.

image

### Changes & Results

The logic in ContextMenu.tsx was incorrectly adding the height of the menu to the bottom max point to test if the menu were off the screen. This was changed to just test the bottom max point. A similar calculation was altered for the width of the menu.

### Testing

Create a custom context menu with about twelve or more items. Right anywhere. If the menu were to be clipped at the bottom then it will open "bottom-to-top".

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 20.9.0 
- [x] Browser: Chrome
  

### DIFF
--- a/platform/ui/src/components/ContextMenu/ContextMenu.tsx
+++ b/platform/ui/src/components/ContextMenu/ContextMenu.tsx
@@ -6,17 +6,17 @@ import { Icons } from '@ohif/ui-next';
 const ContextMenu = ({ items, ...props }) => {
   const contextMenuRef = useRef<HTMLDivElement>(null);
   useEffect(() => {
-    if(!contextMenuRef?.current) {
+    if (!contextMenuRef?.current) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


sedghi (approved):
Thanks

PR #4777, opened 2025-02-10
feat(hotkeys): Migrate hotkeys to customization service and fix issues with overrides

sedghi:
# Hotkeys

## Key Changes:

* Hotkeys are no longer defined in mode factory via `hotkeys: [...hotkeys.defaults.hotkeyBindings]`
* Hotkeys are now managed through the `customizationService` under the key `ohif.hotkeyBindings`
* Default hotkeys are set automatically and can be customized using the customization service
* User-defined hotkey preferences are now stored in a new format in localStorage
* The `HotkeysManager` has undergone significant updates including better handling of defaults, key persistence, and cleanup

### 1. Remove hotkeys array from mode factory definition

**Before:**
```diff
- function modeFactory({ modeConfiguration }) {
-   return {
-     id: 'basic',
-     // ... other configuration
-     hotkeys: [...hotkeys.defaults.hotkeyBindings],
-   };
- }
```

**After:**
```diff
+ function modeFactory({ modeConfiguration }) {
+   return {
+     id: 'basic',
+     // ... other configuration
+     // No hotkeys array necessary
+   };
+ }
```

### 2. Set custom hotkeys using the customization service

There are several methods to modify hotkeys using the customization service:

#### a. Completely replace all hotkeys using `$set`:

```diff
+ onModeEnter: function ({ servicesManager }) {
+   const { customizationService } = servicesManager.services;
+   customizationService.setCustomizations({
+     'ohif.hotkeyBindings': {
+       $set: [
+         {
+           commandName: 'setToolActive',
+           commandOptions: { toolName: 'Zoom' },
+           label: 'Zoom',
+           keys: ['z'],
+           isEditable: true,
+         },
+       ],
+     },
+   });
```

#### b. Add new hotkeys using `$push`:

```diff
+ onModeEnter: function ({ servicesManager }) {
+   const { customizationService } = servicesManager.services;
+   customizationService.setCustomizations({
+     'ohif.hotkeyBindings': {
+       $push: [
+         {
+           commandName: 'myCustomCommand',
+           label: 'My Custom Function',
+           keys: ['ctrl+m'],
+           isEditable: true,
+         },
+       ],
+     },
+   });
+}
```

### 4. Update configuration file if you were setting window.config.hotkeys

If you were previously defining hotkeys in your window.config.js file, it was not really
taken into account. So you can safely remove it now.

**Before:**
```diff
- window.config = {
-   // ...other config
-   hotkeys: [
-     {
-       commandName: 'incrementActiveViewport',
-       label: 'Next Viewport',
-       keys: ['right'],
-     },
-     // ...more hotkeys
-   ],
- };
```

**After:**
```diff
+ window.config = {
+   // ...other config
+ };
```

### 5. Be aware that user preferences are now handled differently

The new system automatically handles user-preferred hotkey mappings:

- User hotkey preferences are stored in `localStorage` under the key `user-preferred-keys`
- The format is a hash-based mapping rather than a full array of definitions
- There's a migration utility that converts old preferences to the new format
- You don't need to manually handle this, but be aware of it if you're accessing localStorage directly

---

# Tours 

### 1. Update any direct references to window.config.tours

If you have any code that directly references window.config.tours, update it to use the customization service:

```diff
- const tours = window.config.tours;
+ const tours = customizationService.getCustomization('ohif.tours');
```

### 2. Use config update patterns for configuring tours

**Before:**
```diff
- window.config = {
-   tours: [
-     {
-       id: 'basicViewerTour',
-       route: '/viewer',
-       steps: [
-         // tour steps...
-       ],
-       tourOptions: {
-         // tour options...
-       },
-     },
-   ],
- };
```

**After:**
```javascript
window.config = {
  customizationService: {
    'ohif.tours': {
      $set: [
        {
          id: 'basicViewerTour',
          route: '/viewer',
          steps: [
            // Your tour steps
          ],
        },
      ],
    },
  },
};
```


platform/core/src/classes/HotkeysManager.ts
```diff
@@ -12,20 +11,19 @@ import Hotkey from './Hotkey';
...
+  private isEnabled: boolean = true;
+  public hotkeyDefinitions: Record<string, any>;
+  public hotkeyDefaults: any[];
 
-  constructor(commandsManager, servicesManager: AppTypes.ServicesManager) {
+  constructor(
+    commandsManager: AppTypes.CommandsManager,
+    servicesManager: AppTypes.ServicesManager
+  ) {
     this.hotkeyDefinitions = {};
```
wayfarer3130, line 25: Please assign the value on line 18/19 since you declare them there.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -430,6 +430,37 @@ function commandsModule({
 
       actions.setViewportWindowLevel({ ...props, viewportId });
     },
+    setWindowLevelPreset: ({ presetName, presetIndex }) => {
```
wayfarer3130, line 433: Having these as commands is really nice for allowing different bindings.

platform/app/public/config/default.js
```diff
@@ -293,305 +293,4 @@ window.config = {
   //       ))
   //   },
   // },
-  hotkeys: [
```
wayfarer3130, line 296: Do you need an example somewhere of how to add a custom binding?  Probably something in e2e.js that demonstrates customizing the configuration bindings by adding a specific customization.

platform/ui-next/src/components/Onboarding/Onboarding.tsx
```diff
@@ -7,15 +7,18 @@ import './Onboarding.css';
 
 import { hasTourBeenShown, markTourAsShown, defaultShowHandler, middleware } from './utilities';
 
-const Onboarding = () => {
-  const Shepherd = useShepherd();
-  const location = useLocation();
-  const tours = window.config.tours as Array<{
+const Onboarding = ({
+  tours,
```
wayfarer3130, line 11: Could you assign an empty array by default so it doesn't have to have the tours assigned?  That should be allowed to work to define no tours.

wayfarer3130:
Can you add the changes:
1. Define the bindings for the extension in the extension
2. Define an array of customizations in the mode that combined define the hotkeys extensions to use
3. Define the default array of customizations in the extensions, and apply that as the default customizations to apply for a mode (so that modes do not need to define hotkey bindings customizations, but get something to apply by default)
4. Add a customization to the extensions in the e2e config that over-writes at least one hotkey to demonstrate how to use customization to add hotkey bindings.
5. Add at least one hotkey binding which has no keys, but is editable to demonstrate the ability to list hotkey commands which could be customized for the user should they choose to do so.
Generally I really like this change - I think it is a significant improvement

wayfarer3130 (changes requested):
Can you add the changes:

Define the bindings for the extension in the extension
Define an array of customizations in the mode that combined define the hotkeys extensions to use
Define the default array of customizations in the extensions, and apply that as the default customizations to apply for a mode (so that modes do not need to define hotkey bindings customizations, but get something to apply by default)
Add a customization to the extensions in the e2e config that over-writes at least one hotkey to demonstrate how to use customization to add hotkey bindings.
Add at least one hotkey binding which has no keys, but is editable to demonstrate the ability to list hotkey commands which could be customized for the user should they choose to do so.
Generally I really like this change - I think it is a significant improvement

extensions/cornerstone/src/commandsModule.ts
sedghi: Thanks

sedghi:
As we discussed,
- I'll create issues for the requested scopes, specifically the extension-level hotkey customization, so we can mix and match them.
- I'll also ensure the case for `isEditable: true` with an empty key works properly.

Issue #4613, opened 2024-12-18
[Bug] Segmentation Cycle End Alert Message


xbankov:
### Describe the Bug

When accessing previous segment before 1st and next segment after the last one, error message is not explaining what is the problem. Show Details button does not work.

### Steps to Reproduce

1. Go to [https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.5962.99.1.2968617883.1314880426.1493322302363.3.0](https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.5962.99.1.2968617883.1314880426.1493322302363.3.0)
2. Open RTSTRUCT (or any other segmentation)
3. Don't load segmentation (See #4612) click left arrow for previous segment. 

### The current behavior

Error message appears, when clicking on details nothing happens.

### The expected behavior

Either iterate to the last segment, or show message explaining that there is no next or previous segmentation.

### OS

Not applicable

### Node version

Not applicable

### Browser

Mozilla Firefox 133.0, Google Chrome 131.0.6778.139

xbankov:
![Image](https://github.com/user-attachments/assets/1b69ce81-7e6d-4ff7-bf67-5101fa47fc6b)

YiannisTheocharakis:
Greetings to all. 

The error is indeed reproducible and I will try to detect where the problem is and return with some feedback.

Yiannis

PR #4783, opened 2025-02-14
fix(Segments Looping) : Fixed looping through Segments of RTSTRUCT using Left/Right viewport buttons

YiannisTheocharakis:
### Context
Fixes [[Bug] Segmentation Cycle End Alert Message #4613](https://github.com/OHIF/Viewers/issues/4613)
Working on the issue realized that onSegmentChange callback function loops through object array indexes while the functions responsible on selecting the segments work with segmentIndex attribute.  Some RTStructures may start from random indexes or may missing some indexes in between. 
example: 

- LUNG1-008 case has 6 Segments with IDs [7, 8, 9, 10, 11, 16]
- When using the arrows on top of the viewport the indexes loops as [0, 1, 2, 3, 4, 5,]
- The values from the looping where fed to the jumpToSegmentCenter() which need the prior indexes. 

### Changes & Results

Issue resolved by:
a) Converting the object array index of navigation cycle to segmentIndex and feed the functions with the correct values.
b) Used also setActiveSegment() function from segmentationService to highlight the focused Segment on the Segment List on the right Panel. 

https://github.com/user-attachments/assets/aa576915-2ba7-4ea7-9299-edef7f68152d

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version:  v23.6.1
- [x] Browser: Vivaldi 7.1.3570.50 (Stable channel) (64-bit)
  


sedghi (commented):
can you apply it for SEG viewport as well? maybe refactor the logic to be used by both

sedghi (approved):
Ok thanks, looking forward for the other PR

Issue #4763, opened 2025-02-04
[Bug] Dependency Imports


Blackglade:
### Describe the Bug

I noticed some strange imports in specific utility functions / files that were not labeled as a direct dependency in the parent `package.json` file. For example take a look at this import in the cornerstone extension:

https://github.com/OHIF/Viewers/blob/ac5ae4a29ee2b91d0dd2eb207dbcee649f94877e/extensions/cornerstone/src/utils/interleaveTopToBottom.ts#L2

This sounds like something that is a crucial utility library. It is, however, labeled as a dev dependency in the parent `package.json`:

https://github.com/OHIF/Viewers/blob/ac5ae4a29ee2b91d0dd2eb207dbcee649f94877e/extensions/cornerstone/package.json#L54

Why does this occur?

### Steps to Reproduce

Steps defined as above.

### The current behavior

Imports are in dev dependency. 

### The expected behavior

Should be moved to dependency 

### OS

macOS

### Node version

v22

### Browser

Chrome

sedghi:
I don't know

PR #4791, opened 2025-02-18
fix: lodash dependencies

pedrokohler:
### Context

Fixes: https://github.com/OHIF/Viewers/issues/4763

The previous dependency setup "worked" due to implicit version resolution by package managers, cached dependencies in node_modules or lock files, and transitive dependencies satisfying API requirements. However, this was unreliable because fresh installs could resolve different versions, leading to inconsistencies or silent failures. The project’s functionality depended on indirect compatibility rather than explicitly defined, correct versions.

This PR ensures explicit dependency control, preventing unintended upgrades and reducing the risk of regressions. It improves reproducibility across environments, future-proofs the project against breaking changes, and enhances security by avoiding untracked vulnerabilities. By properly defining dependencies, we align with OHIF Viewer's intended versioning, ensuring long-term stability and maintainability. This is not just a cleanup—it’s a necessary fix for correctness.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


platform/app/package.json
```diff
@@ -87,7 +87,7 @@
     "history": "^5.3.0",
     "i18next": "^17.0.3",
     "i18next-browser-languagedetector": "^3.0.1",
-    "lodash.isequal": "4.5.0",
+    "lodash": "^4.17.21",
```
sedghi, line 90: why we are bringing the whole lodash in ohif while we just need isEqual? if we don't need the whole lodash i prefer to not have it and remove it

pedrokohler: I'll double check and let you know. I think I made a mistake there.

pedrokohler: Yes, indeed. I assumed there were other lodash functions being used because we had a "lodash" dependency in the devDependencies. But I checked and we really don't. I fixed it now. Thanks for catching that.

sedghi (approved):
awesome!

Issue #4614, opened 2024-12-18
[Bug] Selected Segment is not used when iterating over


xbankov:
### Describe the Bug

The iterating over the segments and selecting a segment are not aligned. I can select the segment from the menu, and then iterate, but it iterates from the segment it iterated over before, not the one selected.

### Steps to Reproduce

1. Go to [https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.5962.99.1.2968617883.1314880426.1493322302363.3.0](https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.5962.99.1.2968617883.1314880426.1493322302363.3.0)
2. Open RTSTRUCT
3. Do not load the segmentation.
4. Open right panel menu with segments/labels shown.
5. Click on any - Segmentation highlight
6. Iterate over the next segment using arrow.

### The current behavior

It starts iterating from 0 even though you have just selected different segment.

### The expected behavior

Iterates over the next segment in the menu.

### OS

Not applicable

### Node version

Not applicable

### Browser

Mozilla Firefox 133.0, Google Chrome 131.0.6778.139

sedghi:
I think this would be a nice addition

YiannisTheocharakis:
Hi all! Started working on the issue. Will return asap with a proper solution.

xbankov:
Awesome! Thanks for the great work!

PR #4793, opened 2025-02-20
feat(Segments Looping) : Fixed looping through Segments of SEG

YiannisTheocharakis:
### Context
Implements/fixes [[feat] Selected Segment is not used when iterating over #4614](https://github.com/OHIF/Viewers/issues/4614).
The navigation between segments when using viewport arrows was inconsistent when selecting manually a SEG and then using arrow. 

### Changes & Results
The functionality was improved by
a) Synchronizing index of arrow looping with the selected segment. If the SEG is changed manually to any position the next time an arrow is pressed the selected segment is retrieved and used to continue navigation from there and not the previous position.
b) I noticed that Segments (and RTS segments also) can be so many that the list must be scrolled to detect the selected segment. Previously when arrows were used the selected segment was not visible on the list. Now it is but if you press right arrow multiple times the selection disappears as it is not visible anymore. This new issue was resolved by editing the dataRow component to ensure visibility always. (see attached video)

https://github.com/user-attachments/assets/a7591024-d8ee-4d8c-87ee-55aaf1178988

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 
- [x] Node version: Node version: v23.6.1 
- [x] Browser: Vivaldi 7.1.3570.50 (Stable channel) (64-bit)
  

### DIFF
--- a/extensions/cornerstone-dicom-rt/src/viewports/OHIFCornerstoneRTViewport.tsx
+++ b/extensions/cornerstone-dicom-rt/src/viewports/OHIFCornerstoneRTViewport.tsx
@@ -2,6 +2,7 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import { ViewportActionArrows } from '@ohif/ui';
 import { useViewportGrid } from '@ohif/ui-next';
+import { onSegmentChange as handleSegmentChange } from '../../../cornerstone/src/utils/segmentUtils';
 
 import promptHydrateRT from '../utils/promptHydrateRT';
 import _getStatusComponent from './_getStatusComponent';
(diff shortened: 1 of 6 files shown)


extensions/cornerstone-dicom-seg/src/viewports/OHIFCornerstoneSEGViewport.tsx
```diff
@@ -138,20 +138,33 @@ function OHIFCornerstoneSEGViewport(props: withAppTypes) {
 
       const numberOfSegments = Object.keys(segments).length;
 
-      let newSelectedSegmentIndex = selectedSegment + direction;
-
-      // Segment 0 is always background
+      //Get activeSegment each time because the user can select any segment from the list and thus the index should be updated
```
sedghi, line 141: i guess now is the time to refactor the logic here that is shared with RTViewport, i think there are duplications

YiannisTheocharakis: Hi @sedghi ! Just updated my PR with some refactoring if you want to check the code. 

Regards
Yiannis

extensions/cornerstone-dicom-rt/src/viewports/OHIFCornerstoneRTViewport.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import { ViewportActionArrows } from '@ohif/ui';
 import { useViewportGrid } from '@ohif/ui-next';
+import { onSegmentChange as handleSegmentChange } from '../../../cornerstone/src/utils/segmentUtils';
```
sedghi, line 5: we shouldn't import like this, export it at the cornerstone as a utils, and then here do 

import { utils } from '@ohif/extension-cornerstone'

extensions/cornerstone-dicom-seg/src/viewports/OHIFCornerstoneSEGViewport.tsx
```diff
@@ -7,6 +7,7 @@ import promptHydrateSEG from '../utils/promptHydrateSEG';
 import _getStatusComponent from './_getStatusComponent';
 import { usePositionPresentationStore } from '@ohif/extension-cornerstone';
 import { SegmentationRepresentations } from '@cornerstonejs/tools/enums';
+import { onSegmentChange as handleSegmentChange } from '../../../cornerstone/src/utils/segmentUtils';
```
sedghi, line 10: same

extensions/cornerstone/src/utils/segmentUtils.ts
```diff
@@ -0,0 +1,45 @@
+import { servicesManager } from '@ohif/app/src/App';
+
+const {
+  segmentationService,
+} = servicesManager.services;
```
sedghi, line 5: We shouldn't import like this as far as i know, pass in the segmentationService from the component itself like onSegmentChange(options) and options has those + segmentationService

sedghi (commented):
see my comments

YiannisTheocharakis:
> see my comments

Ok @sedghi thanks for the review. Will fix my PR and return ASAP

sedghi (approved):
Looks great, thanks a lot

PR #4811, opened 2025-02-26
Update index.tsx with contentF

mbellehumeur:
Ticket: 8562338482  

Support for both 'value:'  or 'contentF' when defining microscopy overlays.

### DIFF
--- a/extensions/dicom-microscopy/src/components/ViewportOverlay/index.tsx
+++ b/extensions/dicom-microscopy/src/components/ViewportOverlay/index.tsx
@@ -88,7 +88,7 @@ export const generateFromConfig = ({ config, overlayData, ...props }) => {
 
 const itemGenerator = (props: any) => {
   const { item } = props;
-  const { title, value: valueFunc, condition, contents } = item;
+  const { title, value: valueFunc, contentF: contentFunc, condition, contents } = item;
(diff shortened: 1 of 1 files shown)


extensions/dicom-microscopy/src/components/ViewportOverlay/index.tsx
```diff
@@ -88,7 +88,7 @@ export const generateFromConfig = ({ config, overlayData, ...props }) => {
 
 const itemGenerator = (props: any) => {
   const { item } = props;
-  const { title, value: valueFunc, condition, contents } = item;
+  const { title, value: valueFunc, contentF: contentFunc, condition, contents } = item;
```
wayfarer3130, line 91: const { title, value: valueFunc, contentF: contentFunc = valueFunc, condition, contents } = item;
will assign to contentFunc directly

sedghi:
@mbellehumeur 
What's the problem, how can we reproduce it, and what steps will confirm the fix?

mbellehumeur (commented):
I tested that it works with value as well

wayfarer3130:
> @mbellehumeur What's the problem, how can we reproduce it, and what steps will confirm the fix?

The problem is that the configuration looks different for overlays for microscopy versus cornerstone viewports, whereas they really should be the same.

PR #4813, opened 2025-02-26
separate checks for addWindowLevelActionMenu and addSegmentationOverlay

Sofien-Sellami:
### Context
This PR adds the `addSegmentationOverlay` option, allowing users to disable the segmentation overlay if set to false. By default, both the segmentation overlay and window-level action menu are enabled.

### Changes & Results

- Now checks `appConfig.addWindowLevelActionMenu !== false` and `appConfig.addSegmentationOverlay !== false`.
- Ensures that both overlays are added by default unless explicitly disabled.

### Testing

- When `appConfig.addWindowLevelActionMenu` is `undefined` or `true`, the window-level action menu is added.
- When `appConfig.addSegmentationOverlay` is `undefined` or `true`, the segmentation overlay is added.
- When either config is set to `false`, the respective component is not added.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: 
- [x] Node version: 
- [x] Browser:
  


sedghi (changes requested):
I'd rather use the `customizationService` for this.

Could you register a default customization template like this:

```javascript
export default {
  'windowLevelActionMenu' : {
     enabled: true
  }
}
```

This would let you set it to `false` in your mode or globally.

The same goes for `segLayer`.


See  https://docs.ohif.org/platform/services/customization-service/customizationService

sedghi (commented):
can you add them to sampleCusotmizations along with screenshots too? thanks

platform/docs/docs/platform/services/customization-service/sampleCustomizations.tsx
```diff
@@ -784,6 +785,45 @@ window.config = {
...
+    description:
+        'Configures the display and location of the window level action menu in the viewport.',
+    image: windowLevelActionMenu,
+    default: null,
+    configuration: `
+      window.config = {
+        // rest of window config
+        customizationService: [
+          {
+            'viewportActionMenu.windowLevelActionMenu': {
```
sedghi, line 799: this is wrong, this would not take effect if not set with $ operators. Can you try each of these to make sure it works?

Sofien-Sellami: Yes, it doesn't work with the previous approach. I just fixed it by using the $set operator and now it works as expected

sedghi:
sorry last thing, can you add a migration guide as well? that we have moved it to customizationService, people might have had the configuration set. Update your PR with the migration too thanks

PR #4821, opened 2025-02-28
[OHI-1681] feat(fonts): bundle fonts in OHIF and dont use google cdn

### LINKED ISSUE #4805: [Feature Request] With or Without Google fonts

I noticed Google Fonts is loading by the OHIF viewer, but is there a way to disable this font, or to provide it directly in the application?
I mean, if there is any issue with the fonts loading, a timeout is waiting to continue the OHIF viewer launch. Is there a solution to avoid this?

Low priority

IbrahimCSAE:
### Context

Bundling fonts

Fixes #4805


sedghi (approved):
Thanks

PR #4825, opened 2025-03-04
feat/threshold brush

sedghi:
- **upgrade cs3d**
- **feat(segmentation): Enhance threshold tools with dynamic and shape options**
- **refactor: Consolidate Toolbox import and remove redundant components**
- **feat: Add Toolbox utility component for dynamic tool configuration**
- **refactor: Enhance toolbar interaction and command handling**
- **refactor: Improve toolbar service and tool settings interaction**
- **feat: Add Shapes tool and update segmentation mode toolbar**
- **refactor: Remove ToolboxProvider and related context**
- **refactor: Simplify toolbar button command and option handling**
- **refactor: Centralize toolbar command processing in ToolbarService**
- **refactor: Simplify toolbar command processing and remove processCommands method**
- **s**
- **dd**

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone-dicom-seg/src/commandsModule.ts
```diff
@@ -329,9 +329,19 @@ const commandsModule = ({
     },
     setThresholdRange: ({
       value,
-      toolNames = ['ThresholdCircularBrush', 'ThresholdSphereBrush'],
+      toolNames = [
```
wayfarer3130, line 332: Can you get a customization for this list?

extensions/default/src/Toolbar/ToolBoxWrapper.tsx
```diff
@@ -20,24 +20,34 @@ export function ToolBoxButtonGroupWrapper({ groupId, items, onInteraction, ...pr
           key={item.id}
           size="small"
           className={props.disabled && 'text-primary'}
-          onInteraction={() =>
-            onInteraction?.({ groupId, itemId: item.id, commands: item.commands })
-          }
+          onInteraction={() => {
```
wayfarer3130, line 23: Can you preserve the event?  Otherwise it is difficult for some interactions to popup a localized menu etc.

extensions/default/src/utils/index.ts
```diff
@@ -1 +1,2 @@
 export { addIcon } from './addIcon';
+export { Toolbox } from './Toolbox';
```
wayfarer3130, line 2: Could we consistently use export * from './Toolbox'
so that it becomes easy to add new exports.  That also means any types defined will get exported.

platform/core/src/hooks/useToolbar.tsx
```diff
@@ -12,14 +14,52 @@ export function useToolbar({ servicesManager, buttonSection = 'primary' }: withA
   const onInteraction = useCallback(
```
wayfarer3130, line 14: If there is an event, could it stop propagation?

platform/core/src/services/ToolBarService/ToolbarService.ts
wayfarer3130, line 1: If the ToolbarService is going to get modified like this, could we please use it to define the hierarchical toolbars?
That is, items such as "MeasurementTools" item should itself be defined as a button, but also use a nested toolbar to define it's contents with a toolbar called "MeasurementTools"

That is, the mode would have to register the "primary" toolbar but also "More Tools" toolbar, and the content of that would be the id's of the buttons included.  That then makes the toolbar button definitions reuseable in multiple places, and allows defining alternate menu setups just be re-using the button definitions.  Those button definitions should then all go into the extensions unless there are specific ones needing to be added in the mode that weren't already defined.  I'm not asking for that second part to be done, just to make the toolbars actually recursive.

The MeasurementTools button parent class should then use `useToolbar` itself to get the list of tools to use at the time it is displayed.

platform/ui-next/src/components/OHIFToolbox/index.ts
```diff
@@ -1,3 +1,2 @@
 import { ToolboxUI } from './ToolboxUI';
```
wayfarer3130, line 1: Please use export * from './ToolboxUI'

modes/basic-dev-mode/src/index.ts
```diff
@@ -96,7 +96,7 @@ function modeFactory({ modeConfiguration }) {
 
       toolbarService.addButtons(toolbarButtons);
       toolbarService.createButtonSection('primary', [
-        'MeasurementTools',
```
wayfarer3130, line 99: I think this should stay MeasurementTools - what it means is that you want to put the "MeasurementTools" button there, which happens to useToolbar('measurementTools'), but that is an internal detail of the MeasurementTools button.

modes/basic-dev-mode/src/toolbarButtons.ts
```diff
@@ -9,74 +9,112 @@ function _createWwwcPreset(preset, title, subtitle) {
...
+    toolGroupIds: ['default', 'mpr'],
+  },
+  context: 'CORNERSTONE',
+};
 
 const toolbarButtons: Button[] = [
+  // sections
   {
-    id: 'MeasurementTools',
+    id: 'measurementSection',
```
wayfarer3130, line 33: As suggested, this is still a button, so it uses the same casing as the other buttons.
The id of it is 'MeasurementTools' as a button.
The name of the button section to use is 'measurementSection' 
They are different objects and shouldn't use the same names.

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -390,16 +398,38 @@ export default class ToolbarService extends PubSubService {
...
-        ...buttons.filter(
-          button => !this.state.buttonSections[key].find(sectionButton => sectionButton === button)
-        )
-      );
-    } else {
+  createButtonSection(key: string, buttons: string[]) {
+    // Create or update the button section
+    if (!this.state.buttonSections[key]) {
       this.state.buttonSections[key] = buttons;
+    } else {
```
wayfarer3130, line 405: I don't think you should have to know anything about the nested button section - that update should be handled by an entirely different useToolbar, and should get the appropriate enhanced information when the toolbar is actually used.

platform/core/src/services/ToolBarService/ToolbarService.ts
wayfarer3130, line 1: I believe the way to make this order independent is to disconnect the sub items generation/fetch from the ToolbarService.  Instead, I think it belongs in ToolButtonListWrapper as a simple useToolbar directive on the contained buttonSection id.   That way the id of the toolbar item is completely independent of the actual button section id.  In fact, one doesn't even know if it USES a nested list.  All the logic for that is contained within the wrapper itself, and it becomes possible to write new wrappers that do other types of behaviours with the included items.

wayfarer3130 (approved):

PR #4835, opened 2025-03-09
feat(viewport-gap): remove viewport gap and update border behavior

IbrahimCSAE:
### Context

This removes the viewport gap and updates the border behavior.

I weighted different options to achieve this, and I believe this is the most reasonable. I wanted to go for a grid approach, but that would break how hanging protocols work and their custom viewport positioning, so we need to stay on the current approach of absolute positioning with a calculated width and a height. For this reason I added a `getBorderStyle` function that can help us get the right borders without needing a display grid parent.

https://github.com/user-attachments/assets/f723a49c-f8e6-48e1-a94b-01c3d3de515d

Fixes: https://linear.app/ohif/issue/OHI-1640

@sedghi @dan-rukas

### DIFF
--- a/platform/app/src/components/ViewportGrid.tsx
+++ b/platform/app/src/components/ViewportGrid.tsx
@@ -284,19 +284,26 @@ function ViewerViewportGrid(props: withAppTypes) {
         viewportGridService.setActiveViewportId(viewportId);
       };
 
-      const getBorderStyle = index => {
+      const getBorderStyle = viewportIndex => {
         const style = {} as any;
-        if (numViewportPanes === 1) {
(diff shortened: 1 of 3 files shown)


sedghi:
I see some weird artifacts in tmtv, probably because of css % 
CleanShot 2025-03-10 at 09 15 20@2x

IbrahimCSAE:
> I see some weird artifacts in tmtv, probably because of css % CleanShot 2025-03-10 at 09 15 20@2x

I will check it out

IbrahimCSAE:
CleanShot 2025-03-10 at 10 10 01@2x

fixed

sedghi (approved):
Thanks

PR #4853, opened 2025-03-14
fix(OIDC): Use new access_token after refresh

elawende:
### Context

When using OIDC authorisation with `automaticSilentRenew: true`, OHIF keeps using the old OIDC access_token after the token has been refreshed. Requests with an expired access_token fail with a 401 or 403 response.

See https://community.ohif.org/t/fix-user-state-not-updating-during-token-refresh-in-oidc-with-keycloak-and-automaticsilentrenew-true-configuration/1065

### Changes & Results

In OpenIdConnectRoutes.tsx, set the user in the userAuthenticationService after a new token is available.

### Testing

Use a setup with OIDC and `automaticSilentRenew: true`. When starting OHIF, a token will be requested and the response of the token endpoint will look like this:

```json
{
    "access_token": "eyJ...VOT3",
    "token_type": "bearer",
    "expires_in": 300,
    "refresh_token": "eyJh...g1xB",
    "id_token": "eyJh...gq5X"
}
```

Wait until the access token has expired (based on `expires_in`, which are seconds) and then perform a search. The request should have an authorization header with the new acccess token (from the last refresh) instead of the expired one.

Please note that the value of `expires_in` varies among OIDC providers. It may be 5 minutes (as required by IHE's IUA profile, this is also the default value in Keycloak), but it may also be hours or even days.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Debian 12 (server), Windows 11 (client)
- [x] Node version: 20
- [x] Browser: Chrome 134.0.6998.89, Firefox 136.0.1
  


andylinton:
Any way I can implement this?

elawende:
@andylinton I'm currently using a fork with this fix, and ran my own build.
@sedghi Is the merge blocked because of the failing deploy/netlify? I did not experience build issues on my local env and can't think of a way how the "Can't resolve '@radix-ui/react-hover-card'" error is related to my change. Is there something I can do to push this PR further?

sedghi:
I will merge today, sorry

sedghi (approved):

PR #4866, opened 2025-03-17
fix(DicomWebDataSource) : Fix a missing null check in `addRetrieveBulkDataNaturalized`

nss10:
### Context
* When the function `addRetrieveBulkDataNaturalized` was called recursively, there is a chance of `naturalized` being null causing Object.keys(naturalized) to error out. 
* Javascript returns `true` for `typeof value[0] === 'object'` even if `value[0]` is `null`, which causes the series data not to continue processing further. 

### Changes & Results
* Added a simple null check to prevent the function from breaking. 

* This error prevented some of the DICOM images from loading when certain elements in the nested structure coming from the backend have any `null` values. 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/default/src/DicomWebDataSource/index.ts
```diff
@@ -431,6 +431,9 @@ function createDicomWebApi(dicomWebConfig: DicomWebConfig, servicesManager) {
        * This is done recursively, for sub-sequences.
        */
       const addRetrieveBulkDataNaturalized = (naturalized, instance = naturalized) => {
+        if (!naturalized) {
```
sedghi, line 434: Hmmm, shouldn't we not return undefined here?

sedghi (approved):
Thanks

PR #4872, opened 2025-03-20
feat: customize cornerstone-SR-extension load toolbar

pedrokohler:
### Context

Currently, there is no way to customize the buttons of the "load toolbar" of the SR viewport with the customization service.

This PR adds this ability.

### Changes & Results

Before:
Uncustomizable toolbar
image

After:
Customized toolbar
image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


pedrokohler:
> I don't think this should be treated as a customization but rather a toolbarSection. So just assign a toolbarSectionId to it, register buttons in the toolbarButton, and make it appear in the UI in the mode.

@sedghi I guess the question then is how are the developers supposed to customize this, since there are buttons already being registered in the upstream code.

Suppose they want to replace the buttons entirely. Are they supposed to clear with the current toolbar values and add new ones?

I believe that what you are saying is already possible. I was using this on the modeEnter callback:

```
toolbarService.addButtons(
    [
      // button config here
    ],
    true
  );

  toolbarService.clearButtonSection('loadSRMeasurements');
  toolbarService.createButtonSection('loadSRMeasurements', [
    'loadSRMeasurements',
    'replaceSRMeasurements',
  ]);
  ```
  
 And it was working, but @wayfarer3130 recommended we added an customization instead.

sedghi:
@wayfarer3130 Do you think it should be a csutomization or a button section with toolbarbuttons?

PR #4902, opened 2025-03-28
Import Path Dicom Microscopy viewer

salimkanoun:
### Context
Following https://github.com/OHIF/Viewers/pull/4891 trying to solve the import of dicom microscopy viewer if served by non root path.

When using publicURL, dicom-microscopy-viewer retrieve from OHIF still miss the basename prefix,

### Changes & Results

In importPath of pluginConfig was settings "/dicom-microscopy-viewer/dicomMicroscopyViewer.min.js " so was evaluated as absolute path and thus PUBLIC_URL was not applyed in write pluginImportFile.js

Removing the first '/' to make it relative solve the issue for retrieving the assets but OHIF crashes on another issue by calling
"/dicom-microscopy-viewer/index.worker.min.worker.js" from with an absolute path without adding the PUBLIC_URL in this http request (but didn't found where it comes from)

### Testing

Use a build with PUBLIC_URL, with this fix retrieve of dicomMicroscopyViewer.min.js is OK but fails for index.worker.min.worker.js

#### Code

  etc.)

  additions or removals.

### DIFF
--- a/platform/app/.webpack/writePluginImportsFile.js
+++ b/platform/app/.webpack/writePluginImportsFile.js
@@ -10,10 +10,6 @@ const extractName = val => (typeof val === 'string' ? val : val.packageName);
 
 const publicURL = process.env.PUBLIC_URL || '/';
 
-function isAbsolutePath(path) {
-  return path.startsWith('http') || path.startsWith('/');
-}
-
 function constructLines(input, categoryName) {
   let pluginCount = 0;
 
@@ -77,7 +73,7 @@ function getRuntimeLoadModesExtensions(modules) {
     if (module.importPath) {
       dynamicLoad.push(
         `  if( module==="${packageName}") {`,
-        `    const imported = await window.browserImportFunction('${isAbsolutePath(module.importPath) ? '' : publicURL}${module.importPath}');`,
+        `    const imported = await window.browserImportFunction('${publicURL}${module.importPath}');`,
         '    return ' +
           (module.globalName
             ? `window["${module.globalName}"];`
--- a/platform/app/pluginConfig.json
+++ b/platform/app/pluginConfig.json
@@ -96,7 +96,7 @@
     },
     {
       "packageName": "dicom-microscopy-viewer",
-      "importPath": "/dicom-microscopy-viewer/dicomMicroscopyViewer.min.js",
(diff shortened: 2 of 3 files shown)


platform/app/.webpack/writePluginImportsFile.js
```diff
@@ -77,7 +73,7 @@ function getRuntimeLoadModesExtensions(modules) {
     if (module.importPath) {
       dynamicLoad.push(
         `  if( module==="${packageName}") {`,
-        `    const imported = await window.browserImportFunction('${isAbsolutePath(module.importPath) ? '' : publicURL}${module.importPath}');`,
+        `    const imported = await window.browserImportFunction('${publicURL}${module.importPath}');`,
```
wayfarer3130, line 76: Please don't undo the change - if you want to prefix it, then use a relative path, and if you don't, then use an absolute path.  This change will break absolute paths to external modules.

platform/app/pluginConfig.json
```diff
@@ -96,7 +96,7 @@
     },
     {
       "packageName": "dicom-microscopy-viewer",
-      "importPath": "/dicom-microscopy-viewer/dicomMicroscopyViewer.min.js",
+      "importPath": "dicom-microscopy-viewer/dicomMicroscopyViewer.min.js",
```
wayfarer3130, line 99: This change by itself is sufficient to load the dicom microscopy viewer inside the publicURL path, without the change to writePluginImportsFile change.  However, the library itself can't handle being loaded at a relative path.

wayfarer3130:
>Removing the first '/' to make it relative solves this issue for retrieving the assets but OHIF crashes on another issue by calling
> "/dicom-microscopy-viewer/index.worker.min.worker.js" from an absolute path without adding the PUBLIC_URL in this http 
> request (but didn't found where it comes from)

The problem is an internal issue in dicom-microscopy-viewer libraries that has to be solved in that library.  Without fixing that internal issue, changing to a relative path is insufficient.

wayfarer3130:
To get the dicom microscopy viewer loading with a public url, you will have to publish it on the root path, OR create a custom version of the microscopy viewer that knows about the public path.  This didn't work in 3.9 either with a public path and not serving it on the root path as it is a limitation of the internal library.

Issue #4903, opened 2025-03-30
[Bug] BulkDataURI Structure Clone Error


salimkanoun:
### Describe the Bug

When OHIF tries to retrieve data from BulkDataURI, there is an error with 

Error Boundary DOMException: Failed to execute 'structuredClone' on 'Window': #<Promise> could not be cloned.

The problem is in getDirectURL that can return a string or a promise of a string because of the
`value.retrieveBulkData(options).then(arr => {
          value.DirectRetrieveURL = URL.createObjectURL(new Blob([arr], { type: defaultType }));
          return value.DirectRetrieveURL;
        })`

Thus the promise propagate from OHIF to cornerstone and the structuredClone does not handle cloning of a Promise

I'm trying to update the code to await the DirectRetrieveURL but this make putting  a lot of fuctions async and changing getDisplaySetsFromSeries to a async function in getSopClassHandlerModule

### Steps to Reproduce

Use a video dicom which will retrieve data from bulkdatauri

### The current behavior

Thrown error Failed to execute 'structuredClone' on 'Window': #<Promise> 

### The expected behavior

Should work

### OS

Linux

### Node version

20

### Browser

Chrome

biskyt:
I was hit with this issue recently. With the help of Claude, I managed to track down the issue and have a **deterministic, file-attached reproduction** — it turns out you don't need an actual video to hit this: an ordinary **multi-frame Secondary Capture _image_** triggers the same `structuredClone(Promise)` crash.

**Attached** 

[ohif-4903-repro.zip](https://github.com/user-attachments/files/29411443/ohif-4903-repro.zip) — GitHub won't take bare `.dcm`s), two anonymised samples:

- **`multiframe-sc-image-176frames.dcm`** — the deterministic crash repro: a multi-frame **Secondary Capture** image, SOP Class `1.2.840.10008.5.1.4.1.1.7`, **176 frames**, `MONOCHROME2`, transfer syntax `1.2.840.10008.1.2.4.70` (JPEG Lossless — *not* a video syntax). A real-world export shape rather than a contrived file: a CT was opened in **MicroDicom** and exported, which merged the localizer/scout views and the axial slice stack into a **single multi-frame SC instance**. (For contrast, the equivalent study in your own demo data keeps the scouts and the slice stack as separate display sets, so it doesn't hit this path.)
[...]

PR #4904, opened 2025-03-30
trying to get bulk data uri without promise

salimkanoun:
See https://github.com/OHIF/Viewers/issues/4903

This PR seems dangerous to me as it changes SOPClass handler and make get display sets from series promised based,

However I dont see which would be the correct way to have a non promised directURL in case of Video dicom

### DIFF
--- a/extensions/default/src/DicomWebProxyDataSource/index.ts
+++ b/extensions/default/src/DicomWebProxyDataSource/index.ts
@@ -46,7 +46,7 @@ function createDicomWebProxyApi(dicomWebProxyConfig, servicesManager: AppTypes.S
       },
     },
     retrieve: {
-      directURL: (...args) => dicomWebDelegate.retrieve.directURL(...args),
+      directURL: async (...args) => dicomWebDelegate.retrieve.directURL(...args),
       series: {
         metadata: async (...args) => dicomWebDelegate.retrieve.series.metadata(...args),
       },
--- a/extensions/default/src/utils/getDirectURL.test.js
+++ b/extensions/default/src/utils/getDirectURL.test.js
@@ -24,10 +24,10 @@ describe('getDirectURL', () => {
     },
   };
 
-  it('should return the provided URL if it exists', () => {
+  it('should return the provided URL if it exists', async () => {
     const url = 'https://example.com/direct-retrieve';
 
-    const result = getDirectURL(config, {
+    const result = await getDirectURL(config, {
       ...params,
       url: 'https://example.com/direct-retrieve',
     });
@@ -35,12 +35,12 @@ describe('getDirectURL', () => {
     expect(result).toBe(url);
   });
 
-  it('should return the DirectRetrieveURL if it exists', () => {
+  it('should return the DirectRetrieveURL if it exists', async () => {
     const value = {
       DirectRetrieveURL: 'https://example.com/direct-retrieve',
     };
 
-    const result = getDirectURL(config, {
+    const result = await getDirectURL(config, {
       ...params,
       tag: 'PixelData',
       instance: {
@@ -52,12 +52,12 @@ describe('getDirectURL', () => {
     expect(result).toBe(value.DirectRetrieveURL);
   });
 
-  it('should return the URL for InlineBinary', () => {
+  it('should return the URL for InlineBinary', async () => {
     const value = {
       InlineBinary: 'base64-encoded-data',
     };
 
-    const result = getDirectURL(config, {
+    const result = await getDirectURL(config, {
       ...params,
       tag: 'PixelData',
       instance: {
@@ -69,12 +69,12 @@ describe('getDirectURL', () => {
     expect(result).toContain('blob:');
   });
 
-  it('should return the BulkDataURI with defaultType if singlepart is false and there is no retrieveBulkData', () => {
+  it('should return the BulkDataURI with defaultType if singlepart is false and there is no retrieveBulkData', async () => {
     const value = {
       BulkDataURI: 'https://example.com/bulkdata',
     };
 
-    const result = getDirectURL(
+    const result = await getDirectURL(
       {
         ...config,
         singlepart: false,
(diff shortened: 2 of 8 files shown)


extensions/dicom-video/src/getSopClassHandlerModule.js
```diff
@@ -50,11 +50,12 @@ const _getDisplaySetsFromSeries = (instances, servicesManager, extensionManager)
...
       );
     })
-    .map(instance => {
+    const displaySets = [];
+    for(const instance of videoInstances) {
       const { Modality, SOPInstanceUID, SeriesDescription = 'VIDEO', imageId } = instance;
       const { SeriesNumber, SeriesDate, SeriesInstanceUID, StudyInstanceUID, NumberOfFrames, url } =
         instance;
-      const videoUrl = dataSource.retrieve.directURL({
+      const videoUrl = await dataSource.retrieve.directURL({
```
wayfarer3130, line 58: This is going to cause real rendering delays and issues for large videos.  
The getDirectUrl is supposed to return nearly immediately if it can directly compute a URL that can be used to display the given data.  There SHOULD be a separate query that returns an async result something like getDataUrl, and then both videoUrl and thumbnailSrc should provide a separate method to fetch the data url object, something like:

const videoUrl = dataSource.retrieve.directURL(...)
const getVideoUrl = () => dataSource.retrieve.getDataUrl(...)

Then, the video viewport should check videoUrl and if not defined, should call the getVideoUrl and assign it to the videoUrl attribute once it comes back.  That has to happen outside the sop class handler module as it should only start fetching once requested to view.

extensions/dicom-video/src/getSopClassHandlerModule.js
```diff
@@ -78,7 +79,7 @@ const _getDisplaySetsFromSeries = (instances, servicesManager, extensionManager)
         videoUrl,
         renderedUrl: videoUrl,
         instances: [instance],
-        thumbnailSrc: dataSource.retrieve.directURL({
+        thumbnailSrc: await dataSource.retrieve.directURL({
```
wayfarer3130, line 82: The thumbnailSrc should not retrieve the async result here as that means display sets don't load until thumbnails are loaded.  That will massively slow down initial display.  Instead, implement the getThumbnailSrc method that calls the dataSoruce.retrieve.getDataUrl.
See the uses of getThumbnailSrc in the current OHIF branch.  The intent is to use a method like:
```
function getThumbnailSrc(instance) {
  const { extensionManager } = appContext;
  const dataSource = extensionManager.getActiveDataSource()[0];
  const dataSourceConfig = dataSource.getConfig();
  if (dataSourceConfig.thumbnailRendering === 'rendered') {
    return;
  }
  if (
    dataSourceConfig.thumbnailRendering === 'directURL' ||
    dataSourceConfig.thumbnailRendering === 'thumbnail'
  ) {
    const directURL = dataSource.retrieve.directURL({
      instance,
      defaultPath: '/thumbnail',
      defaultType: 'image/jpeg',
      singlepart: true,
      tag: 'Absent',
    });
    return directURL;
  }
  if (dataSourceConfig.thumbnailRendering === 'wadors') {
    let imageIdForThumbnail = dataSource.getImageIdsForInstance({
      instance: instance,
      thumbnail: true,
    });
    imageIdForThumbnail = imageIdForThumbnail.replace('wadors:', ''); //  could be a better way
    return function getThumbnailSrc() {
      return dataSource.retrieve.bulkDataURI({
        BulkDataURI: imageIdForThumbnail,
        //      StudyInstanceUID: instance.StudyInstanceUID, // maybe this is not needed
        //     SeriesInstanceUID: instance.SeriesInstanceUID,
        //     SOPInstanceUID: instance.SOPInstanceUID,
      });
    };
  }
  return;
}

```

extensions/dicom-video/src/getSopClassHandlerModule.js
wayfarer3130: There is an existing method which you can use to assign the thumbnailSrc called load - you have to be careful not to load the video multiple times if it takes a while and several view events call load - each one should return the same async load promise, and it should assign the videoUrl when done.

extensions/default/src/utils/getDirectURL.ts
```diff
@@ -49,10 +49,11 @@ const getDirectURL = (config, params) => {
         const options = {
           mediaType: defaultType,
         };
-        return value.retrieveBulkData(options).then(arr => {
+        const directRetrieveURL = await value.retrieveBulkData(options).then(arr => {
```
wayfarer3130, line 52: Please remove this code rather than returning it like this.  The call to retrieveBulkData was a bad idea (of mine unfortunately).  There should be a separate method retrieveDataUrl instead containing the retrieve object logic.  This method should be fully synchronous all the time (no async AND no call returning a promise)

IF the bulkdata uri can directly return the URL without needing to include authorization, then it should return the URL.

extensions/default/src/utils/getDirectURL.test.js
```diff
@@ -52,12 +52,12 @@ describe('getDirectURL', () => {
     expect(result).toBe(value.DirectRetrieveURL);
   });
 
-  it('should return the URL for InlineBinary', () => {
+  it('should return the URL for InlineBinary', async () => {
```
wayfarer3130, line 55: Can you leave the tests here, but change them to use the getDataURL instead

extensions/default/src/DicomWebProxyDataSource/index.ts
```diff
@@ -46,7 +46,7 @@ function createDicomWebProxyApi(dicomWebProxyConfig, servicesManager: AppTypes.S
       },
     },
     retrieve: {
-      directURL: (...args) => dicomWebDelegate.retrieve.directURL(...args),
```
wayfarer3130, line 49: Please revert this change, and add getDataURL.  There is no need to make it "async" as the return from the nested instance is already a promise and that is all async means (plus the handlign to allow await)  Doing it unnecessarily adds quite a bit of overhead.

extensions/default/src/utils/getDirectURL.ts
salimkanoun: Sorry i'm lost with this PR as I don't catch the fetching logic with all these endpoint,

However for this poins I don't see how we can guess if the authorization header is needed or not, if an authorization system is in the dicom web it is likely to affect all routes including direct urls endpoint, otherwise it sounds like a security breach.

is it possible / expected to have a public endpoint on a dicom webserver with an authorization policy ?

Issue #4908, opened 2025-04-01
[Bug] Cine "pilles up" when playing and video viewport is updated


pedrokohler:
### Describe the Bug

You have a video viewport that has a cine playing and then you update it with a different displaySet it will incur into a faulty state where: 

1- Both cines are playing one on top of the other simultaneously
2- You may have an inconsistent cine player state

### Steps to Reproduce

1. Open https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=2.25.96975534054447904995905761963464388233
2. Open the video series
3. Start cine
4. Either drag or double click the other series, to load it into the same viewport as the video that is currently playing
5. Watch as hell breaks loose with a mixture of both series playing together

### The current behavior

https://github.com/user-attachments/assets/868243ac-1fb1-41f5-aae8-87303db4c33b

### The expected behavior

https://github.com/user-attachments/assets/09f668a5-5ca7-44cd-a134-f608e30e6b68

### OS

MacOS

### Node version

20ˆ

### Browser

Chrome

PR #4909, opened 2025-04-01
fix(videoViewport): Resolve Cine Player Faults on DisplaySet Update

pedrokohler:
### Context

Fixes: https://github.com/OHIF/Viewers/issues/4908

### Changes & Results

Before:

https://github.com/user-attachments/assets/868243ac-1fb1-41f5-aae8-87303db4c33b

After:

https://github.com/user-attachments/assets/09f668a5-5ca7-44cd-a134-f608e30e6b68

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


platform/core/src/services/ViewportGridService/ViewportGridService.ts
```diff
@@ -184,14 +184,27 @@ class ViewportGridService extends PubSubService {
...
+    await this.serviceImplementation._setDisplaySetsForViewports(viewportsToUpdate);
     const state = this.getState();
     const updatedViewports = [];
 
     const removedViewportIds = [];
 
-    for (const viewport of props) {
+    // Stopping the cine of modified viewports before changing the viewports to
+    // avoid inconsistent state and lost references
+    viewportsToUpdate.forEach(viewport => {
```
sedghi, line 197: This doesn't feel right. 

Using the `cine` service inside `viewportGridService` is tight coupling, which is really not ideal.

pedrokohler: Yes. I agree it is not ideal. However it must be done BEFORE the update happens. 

I tried adding a listener to the CineService for the `ViewportGridService.EVENTS.GRID_STATE_CHANGED` event, but it doesn't help, since it happens AFTER this function has done its job and because of that trying to turn off the cine does not work at this point.

One alternative would be find every place where this ViewportGridService function is called and also do it there.

But maybe you have a better suggestion than this one.

pedrokohler: I could create a command that does that and runs the original function and call it in place of the `viewportGridService.setDisplaySetsForViewports`. And substitute everywhere where it is called.

This way only the commandsModule will know about the cineService and the function will not have to know about it.

pedrokohler: > I could create a command

@sedghi I did it the way I just described above so you can take a look and see if you like it.

PR #4973, opened 2025-04-18
fix(dicomWebClient): use public getter for activeDataSourceName instead of accessing private property

pedrokohler:
### Context

The `dicomWebClient.ts` file was directly accessing a private property (named `activeDataSource` in the past, but recently renamed to `activeDataSourceName`) from the `ExtensionsManager`. This private access caused issues in certain features, such as microscopy mode which was broken.

This PR resolves the issue by introducing a public getter for `activeDataSourceName` in `ExtensionsManager` and updates `dicomWebClient` to use the getter, ensuring proper encapsulation and avoiding direct access to internal properties.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/extensions/dicom-microscopy/src/utils/dicomWebClient.ts
+++ b/extensions/dicom-microscopy/src/utils/dicomWebClient.ts
@@ -11,7 +11,7 @@ import { StaticWadoClient } from '@ohif/extension-default';
  */
 export default function getDicomWebClient({ extensionManager, servicesManager }: withAppTypes) {
   const dataSourceConfig = window.config.dataSources.find(
-    ds => ds.sourceName === extensionManager.activeDataSource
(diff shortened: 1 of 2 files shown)


sedghi (commented):
Can you create a migration guide outlining any changes we're making?
is there anything publically changing?

pedrokohler:
> Can you create a migration guide outlining any changes we're making? is there anything publically changing?

This is not supposed to affect anyone that is using the properties properly. If they are not accessing forbidden private properties of the classes, they should not have any problems.

This change should only affect internal `ExtensionMode` code which was already updated with this change.

The problem is that the `dicomWebClient.ts` was accessing a private variable from another class, but now that is fixed.

sedghi (approved):

Issue #4759, opened 2025-02-03
[Bug] OHIF V3.9 Fails to Render DX Modality Images Properly When RescaleIntercept (0028,1052) and RescaleSlope (0028,1053) Tags Are Missing, While OHIF V2 Displays Them Correctly


komalreddy3:
### Describe the Bug

[Anonymised DX modality.zip](https://github.com/user-attachments/files/18639952/Anonymised.DX.modality.zip)

### Steps to Reproduce

Use attached DX modality anonymized DICOM file to render images in OHIF V3.9.
Load the DICOM file into OHIF V3.9 using DICOM JSON as the display method.

### The current behavior

In OHIF V2, the images from the same study are rendered correctly, even when the Rescale Intercept (0028,1052) and Rescale Slope (0028,1053) tags are missing or empty. However, in OHIF V3.9, the images appear blank when these tags are not present.

### The expected behavior

The images should render consistently in both OHIF V2 and V3.9, regardless of whether the Rescale Intercept (0028,1052) and Rescale Slope (0028,1053) tags are missing or empty

### OS

Windows 10

### Node version

20.0.0

### Browser

Chrome

PR #5003, opened 2025-04-29
Fix: issue #4759 Enable proper DICOM image rendering when Rescale parameters are set to null

tctco:
### Context

This pr aims to solve https://github.com/OHIF/Viewers/issues/4759

The current version of OHIF fails to render DICOM image data when RescaleIntercept (0028,1052) and RescaleSlope (0028,1053) Tags are set to null. This pr simply enables early return when these parameters are set to None.

```typescript
// platform\core\src\classes\MetadataProvider.ts
      case WADO_IMAGE_LOADER_TAGS.MODALITY_LUT_MODULE:
        const { RescaleIntercept, RescaleSlope } = instance;
        if (
          RescaleIntercept === undefined ||
          RescaleIntercept === null || // this line
          RescaleSlope === undefined ||
          RescaleSlope === null // and this line
        ) {
          return;
        }
```

### Changes & Results

This is an easy fix, and [komalreddy3](https://github.com/komalreddy3) already provided a [testing case](https://github.com/user-attachments/files/18639952/Anonymised.DX.modality.zip) for this. If the ohif team needs the example from our center, I can also upload another example.

**before**

https://github.com/user-attachments/assets/14e08279-0e49-42c3-9407-11ea17942676

**after**

https://github.com/user-attachments/assets/20350235-35ab-421c-b7f1-4c320cbf7f3a

### Testing

Please refer to this [testing case](https://github.com/user-attachments/files/18639952/Anonymised.DX.modality.zip).

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: WSL2 Ubuntu 22.04.5 LTS
- [x] Node version: v20.8.0
- [x] Browser: MS Edge 135.0.3179.98


platform/core/src/classes/MetadataProvider.ts
```diff
@@ -215,7 +215,12 @@ class MetadataProvider {
         break;
       case WADO_IMAGE_LOADER_TAGS.MODALITY_LUT_MODULE:
         const { RescaleIntercept, RescaleSlope } = instance;
-        if (RescaleIntercept === undefined || RescaleSlope === undefined) {
+        if (
```
sedghi, line 218: I think we can say:

```js
if (RescaleIntercept == null || RescaleSlope == null) {
  return;
}
```

And add a note that this covers both `null` and `undefined`. What do you think?

tctco: I have updated this pr to use loose equality and added a short comment addressing the issue. How do you like this update?

Please tell me if any other amendments are required :)

sedghi (approved):

PR #5010, opened 2025-05-02
fix/remove legacy ui

sedghi:
This pull request removes the old UI, keeping only the worklist-related UI. We're also migrating the worklist to the new UI in `@ohif/ui-next`.


sedghi:
All tests pass locally

CleanShot 2025-05-02 at 14 03 14@2x

Issue #4954, opened 2025-04-14
[Bug] Incorrect units for Median and Standard Deviation in Sidebar Segment Statistics in TMTV


dxlin:
### Describe the Bug

Median and Standard Deviation showing units in mm rather than SUV in latest viewer-dev code.

### Steps to Reproduce

1. Open TMTV mode
https://viewer-dev.ohif.org/tmtv?StudyInstanceUIDs=1.2.840.113619.2.290.3.3767434740.226.1600859119.501
2. Add Segment
3. Select brush and create segment
4. Mouse over "Segment 1" on side bar (or in console get the segments cachedStats via SegmentationService)

### The current behavior

Median and STD are in mm

### The expected behavior

Median and STD units should be same as Max, Min, etc. (SUV).

### OS

Windows 11

### Node version

22.12.0

### Browser

Chrome 135.0.7049.85

sedghi:
good catch

PR #5014, opened 2025-05-06
chore: recovery 2 

sedghi:
Addresses these issues: 

- [x] https://github.com/OHIF/Viewers/pull/4957
- [x] https://github.com/OHIF/Viewers/issues/4954
- [x] https://github.com/OHIF/Viewers/pull/5003
- [x] https://github.com/OHIF/Viewers/pull/5005
- [x] https://github.com/OHIF/Viewers/issues/5084
- [x] https://github.com/OHIF/Viewers/pull/4974#pullrequestreview-2784072165


extensions/cornerstone/src/utils/CornerstoneViewportDownloadForm.tsx
```diff
@@ -180,15 +180,12 @@ const CornerstoneViewportDownloadForm = ({
...
-        } catch (error) {
-          console.debug('Error enabling tool:', error);
-        }
-      } else {
-        toolGroup.setToolDisabled(toolName);
+    const toolInstances = toolGroup.getToolInstances();
+    const toolInstancesArray = Object.values(toolInstances);
+
+    toolInstancesArray.forEach(toolInstance => {
+      if (toolInstance.constructor.isAnnotation !== false) {
```
jbocce: I suspect you still want the `show` check. I believe this is what determines if the annotations check box is checked  or not. Please double check this.

extensions/cornerstone/src/init.tsx
```diff
@@ -310,37 +316,73 @@ export default async function init({
...
+            error: 'Web Worker failed',
+          },
+        });
+      } else {
+        // Already tracking this type of task, just let it continue
+        console.debug(`Already tracking a "${type}" task, skipping duplicate notification`);
+      }
+    }
+    // Task completed
+    else if (progress === 100) {
```
jbocce: stupid question... what happens if the worker errors out? Will the task get removed from the `activeWorkerTasks`? Maybe we don't care, but I wanted for us to be sure. Furthermore, does the promise need to be rejected in that case too?

extensions/cornerstone/src/utils/CornerstoneViewportDownloadForm.tsx
sedghi: Good catch

extensions/cornerstone/src/init.tsx
```diff
@@ -310,37 +316,100 @@ export default async function init({
...
+              deduplicationInterval: 60000, // 60 seconds - prevent frequent notifications of same type
+              promise: progressPromise,
+              promiseMessages: {
+                loading: `Computing...`,
+                success: `Completed successfully`,
+                error: 'Web Worker failed',
+              },
+            });
+          } catch (error) {
+            console.error('Error showing web worker notification:', error);
```
jbocce: I could be wrong, but don't we need to (re)throw the error and/or call `reject` here so that the outer `finally` gets called. Apologies if I am missing something here.

extensions/cornerstone/src/init.tsx
```diff
@@ -310,37 +316,100 @@ export default async function init({
...
+  const getNormalizedTaskKey = type => {
+    return `worker-task-${type.toLowerCase().replace(/\s+/g, '-')}`;
+  };
+
+  eventTarget.addEventListener(EVENTS.WEB_WORKER_PROGRESS, ({ detail }) => {
+    let normalizedKey;
+    let shouldCleanup = false;
+    
+    try {
+      const { progress, type, id } = detail;
```
jbocce: Is there any `detail` from the web worker progress that we might cause us to want to `reject` the promise? For example, could `detail` ever indicate an error occured in the worker?

jbocce (changes requested):
Just some more comments.

extensions/cornerstone/src/utils/initWebWorkerProgressHandler.ts
```diff
@@ -0,0 +1,109 @@
...
+              deduplicationInterval: 60000, // 60 seconds - prevent frequent notifications of same type
+              promise: progressPromise,
+              promiseMessages: {
+                loading: `Computing...`,
+                success: `Completed successfully`,
+                error: 'Web Worker failed',
+              },
+            });
+          } catch (error) {
+            console.error(`Error showing web worker notification for type "${type}":`, error);
```
jbocce, line 68: I think we need `reject` and `throw` here too.

jbocce (changes requested):
One comment.

extensions/cornerstone/src/utils/initWebWorkerProgressHandler.ts
sedghi: added a throw there is no promise to reject there

jbocce (approved):
Approved.

sedghi:
All tests pass locally

PR #5022, opened 2025-05-08
fix: image slices are displayed in reverse order, if we switch from common layout to 2D MPR

### LINKED ISSUE #5021: [Bug] Image slices are displayed in reverse order, if we switch from common layout to 2D MPR for some specific data
### Describe the Bug

For certain datasets https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.2.826.0.1.3680043.2.1125.1.11608962641993666019702920539307840, switching from the common layout to the 2D MPR view results in image slices being displayed in reverse order.

### Steps to Reproduce

1. Launch the viewer with https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.2.826.0.1.3680043.2.1125.1.11608962641993666019702920539307840
2. Observe the slice order / Draw an annotation
3. Switch to 2D MPR mode
4. Now check the slice order / Click the annotation check the slice no in the viewer and in the measurement panel

### The current behavior

The slice order is observed to be reversed when 

https://github.com/user-attachments/assets/6d2821d6-a179-44d8-8c8b-4cddda46dd0a

### The expected behavior

The slice order is expected to be maintained without inversion when switching from the common layout to the 2D MPR view.

### OS

Windows 11

### Node version

v18.19.0

### Browser

Chrome  v136.0.7103.93

arul-trenser:
…like ImagePositionPatient/ImageOrientationPatient is available and sorted stack images by instance number when necessary metadata is not present

### Context

- Fixes #5021 

### Changes & Results

This issue is fixed by sorting slices according to spatial position when necessary metadata like `ImagePositionPatient`/`ImageOrientationPatient` is available and sorted stack images by instance number when necessary metadata is not present.

https://github.com/user-attachments/assets/dd249c2a-c42b-402b-8af4-1c8bc7a7cb09

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS:   Windows 11
- [x] Node version:  18.19.0
- [x] Browser: Chrome 131.0.6778.265
  


extensions/default/src/utils/validations/areAllImagePositionsEqual.ts
```diff
@@ -1,7 +1,7 @@
 import { vec3 } from 'gl-matrix';
 import toNumber from '@ohif/core/src/utils/toNumber';
 import { _getPerpendicularDistance } from '@ohif/core/src/utils/isDisplaySetReconstructable';
-import calculateScanAxisNormal from '../calculateScanAxisNormal';
+import calculateScanAxisNormal from '@ohif/core/src/utils/calculateScanAxisNormal';
```
wayfarer3130, line 4: You can't access src directory from @ohif/core - I know it mostly builds, but not consistently.  You will have to export it from the utils module in @ohif/core and then import it from @ohif/core and then extract from utils.

platform/core/src/classes/ImageSet.ts
```diff
@@ -79,63 +75,96 @@ class ImageSet {
...
+   * 1. Image position (if ImagePositionPatient and ImageOrientationPatient are defined)
+   * 2. InstanceNumber
+   * Note: Images are sorted in-place and a reference to the sorted image array is returned.
+   *
+   * @returns images - reference to images after sorting
+   */
+  sort(sortingCallback: (a: Image, b: Image) => number): Image[] {
+    try {
+      return this.sortByImagePositionPatient();
+    } catch (error) {
```
wayfarer3130, line 89: Catching an error for a normal flow isn't the right way to handle this - what you should do is something like:
const sortImagePositionCompare = createSortByImagePositionPatient(images);
this.images.sort(sortImagePositionCompare || defaultImageSort)

platform/core/src/classes/ImageSet.ts
```diff
@@ -79,63 +75,96 @@ class ImageSet {
...
 
-function _getImagePositionPatient(image) {
-  return image.getData().metadata.ImagePositionPatient;
-}
+    // Calculate the scan axis normal using the cross product
+    const scanAxisNormal = calculateScanAxisNormal(ImageOrientationPatient);
+
+    // Compute distances from each image to the reference image
+    const distanceInstancePairs = images.map(image => {
+      const imagePositionPatient = image.ImagePositionPatient;
```
wayfarer3130, line 151: Please use the 
const deltaVector = vec3.create();
...
vec3.dot(scanAxisNormal, vec3.subtract(deltaVector,imagePositionPatient, referenceImagePositionPatient)) to calculate distances rather than re-implementing it dot product using reduce.  It makes it clearer what is going on.

platform/core/src/classes/ImageSet.ts
```diff
@@ -1,15 +1,11 @@
...
 type Attributes = Record<string, unknown>;
 type Image = {
   StudyInstanceUID?: string;
-  getData(): {
-    metadata: {
-      ImagePositionPatient: number[];
-      ImageOrientationPatient: number[];
-    };
-  };
+  ImagePositionPatient?: string;
```
wayfarer3130, line 7: This shouldn't be a string ever - it is a Point3 if defined, same with ImageOrientationPatient.

platform/core/src/classes/ImageSet.ts
```diff
@@ -79,63 +75,96 @@ class ImageSet {
...
-        distance,
-        image,
-      };
+  /**
+   * Sort by InstanceNumber
+   * Note: Images are sorted in-place and a reference to the sorted image array is returned.
+   *
+   * @returns images - reference to images after sorting
+   */
+  sortByInstanceNumber(): Image[] {
```
wayfarer3130, line 111: This should use the public sort instances in the sortStudy utility - you will probably have to export the instancesSortCriteria, although the thumbnail PR adds that already, and will probably merge before your PR does.

platform/core/src/classes/ImageSet.ts
```diff
@@ -79,63 +75,96 @@ class ImageSet {
...
-    distanceImagePairs.sort(function (a, b) {
-      return b.distance - a.distance;
-    });
+  /**
+   * Sort by image position, calculated using ImageOrientationPatient and ImagePositionPatient
+   * Note: Images are sorted in-place and a reference to the sorted image array is returned.
+   *
+   * @returns images - reference to images after sorting
+   */
+  sortByImagePositionPatient(): Image[] {
```
wayfarer3130, line 125: With the remaining fixes for making this a comparator, the createImagePositionPatientCompare should get moved into sortStudy so that it can be re-used consistently.

platform/core/src/classes/ImageSet.ts
arul-trenser: @wayfarer3130  Modified as suggested.

arul-trenser:
@wayfarer3130  I have made the updates to the PR based on your review comments. Could you please take a look when you have a moment?

Thanks!

platform/core/src/utils/sortStudy.ts
```diff
@@ -128,11 +130,60 @@ export default function sortStudy(
...
 }
 
+/**
+ * Sort by image position, calculated using imageOrientationPatient and ImagePositionPatient
+ * If imageOrientationPatient or ImagePositionPatient is not available, Images will be sorted by the provided fallbackSort function
+ * Note: Images are sorted in-place and a reference to the sorted image array is returned.
+ *
+ * @returns images - reference to images after sorting
+ */
+const sortStudyByImagePositionPatient = (images, fallbackSort) => {
```
wayfarer3130, line 140: Can you pass a comparator in rather than a full sort function?

wayfarer3130 (commented):
Almost there - the only critical fix is to change the final sort order.  Using indexOf is an O(n) function, so doing it the way you have would make the sort O(n*n*log n) which is too slow.

arul-trenser:
@wayfarer3130  I have updated as per the suggestions. Please review at your convenience. 
Thanks!

sedghi (changes requested):
1) We need migration guide in the https://docs.ohif.org/migration-guide/3p10-to-3p11/ explains our sorting has changed, and how they can switch back to the old way (we should provide away to say add this configuration and it will use the old way

Otherwise looks great

arul-trenser:
@sedghi  I have added a customization to switch back to the old way also updated the documentation.
Could you please review it at you convenience?

platform/core/src/classes/ImageSet.ts
```diff
@@ -1,15 +1,16 @@
...
 import guid from '../utils/guid.js';
-import { Vector3 } from 'cornerstone-math';
+import {
+  instancesSortCriteria,
+  isValidForPositionSort,
+  sortImagesByPatientPosition,
+} from '../utils/sortStudy';
 
 type Attributes = Record<string, unknown>;
 type Image = {
```
wayfarer3130: export type Image here so it is accessible.

platform/core/src/classes/ImageSet.ts
```diff
@@ -79,63 +80,46 @@ class ImageSet {
...
-      )
+  /**
+   * Default image sorting. Sorts by the following (in order of priority)
+   * 1. Image position (if ImagePositionPatient and ImageOrientationPatient are defined)
+   * 2. Sort by a provided sortingCallback Criteria
+   * Note: Images are sorted in-place and a reference to the sorted image array is returned.
+   *
+   * @returns images - reference to images after sorting
+   */
+  sort(customizationService): Image[] {
```
wayfarer3130, line 91: @sedghi - do you think the customization service should be passed in here, or do you think the instance sorting criteria should be passed - if it is the instance sorting criteria, then it can be used to generate different sort types for different image sets or at different times?

wayfarer3130: Hmm, this is the image set sort function, so it is already the spot looking up the sorting.

wayfarer3130 (approved):
Thinking about where the customization goes, I think the current behaviour is good for now, but eventually having more control over the comparison, eg to have an init method or a factor method available to provide context would be useful, and then using straight comparators.  That is something that can be added later and I think this is a good change as it is now.

wayfarer3130:
Can you look at the failing comparisons?

platform/core/src/classes/ImageSet.ts
arul-trenser: changed as per your suggestion

IbrahimCSAE:
The changes in the PR broke some of the sorting on my data, I will see if I can get permission to post an anonymized version, but after I reverted the changes from this PR, the sorting worked again as expected

arul-trenser:
> Can you look at the failing comparisons?

@wayfarer3130 The Playwright test `jumpToMeasurementMPR` initialDraw is failing, likely because the screenshot is being captured before the image has fully loaded. This issue appears to be recurring in other pull requests as well.

arul-trenser:
@sedghi  @wayfarer3130  Could you please approve and merge this PR if everything looks good?

Thanks!!

platform/core/src/classes/ImageSet.ts
arul-trenser: @sedghi Could you please share your thoughts on this?

Thanks!

jbocce:
@arul-trenser , since I have been working on the e2e tests I figured I would have a look for you. There are two failing tests:

```
  2 failed
    [chromium] › tests/JumpToMeasurementMPR.spec.ts:10:5 › should hydrate in MPR correctly ─────────
    [chromium] › tests/MPRThenRTOverlayNoHydration.spec.ts:10:5 › should launch MPR with unhydrated RTSTRUCT chosen from the data overlay menu 
```

The `MPRThenRTOverlayNoHydration.spec.ts` you can ignore. I will be improving those tests like this one where there are pop ups soon.

However `JumpToMeasurementMPR.spec.ts` is something you should look into - be sure to look at the screen captures of the actual and expected to help you out. I can see the problem reported by testing manually as well. One of the first things that the test does is navigate to image 21. However, this PR seems to have changed what image 21 actually is. So either change the screenshot (be sure to add a 5 second wait prior to taking the screen shot) OR fix the problem.

sedghi:
We're wrapping up a release, so if you'd like to include this, please address all comments by the end of the week.

arul-trenser:
@jbocce Previously, the study appeared reversed in the common layout, although it was correct in the MPR layout. This PR addresses that reversal. As a result, the slice order in the common layout will now be different, which is the expected behavior and resolves the underlying issue. This change caused a mismatch with existing screenshots, leading to test failures, so I've updated them accordingly.

However, I've observed that the display set order in the left side panel is inconsistent when running Playwright tests. This issue is difficult to reproduce manually, occurring only about once in 50 attempts, and typically during the initial un-cached load. It's also not consistently reproducible even under those conditions (in master). However, the order becomes correct if we reload the page or navigate back to the study list and re-select the study. This order mismatch is causing a test failure in the master branch, which I've also confirmed locally.

cc: @sedghi

jbocce:
> @jbocce Previously, the study appeared reversed in the common layout, although it was correct in the MPR layout. This PR addresses that reversal. As a result, the slice order in the common layout will now be different, which is the expected behavior and resolves the underlying issue. This change caused a mismatch with existing screenshots, leading to test failures, so I've updated them accordingly.
> 
> However, I've observed that the display set order in the left side panel is inconsistent when running Playwright tests. This issue is difficult to reproduce manually, occurring only about once in 50 attempts, and typically during the initial un-cached load. It's also not consistently reproducible even under those conditions (in master). However, the order becomes correct if we reload the page or navigate back to the study list and re-select the study. This order mismatch is causing a test failure in the master branch, which I've also confirmed locally.
> 
> cc: @sedghi

Yes I understand. I think it might be best to simply close the left panel prior to taking screenshots and open as necessary with `await page.getByTestId('side-panel-header-left').click();`.

wayfarer3130:
> > @jbocce Previously, the study appeared reversed in the common layout, although it was correct in the MPR layout. This PR addresses that reversal. As a result, the slice order in the common layout will now be different, which is the expected behavior and resolves the underlying issue. This change caused a mismatch with existing screenshots, leading to test failures, so I've updated them accordingly.
> > However, I've observed that the display set order in the left side panel is inconsistent when running Playwright tests. This issue is difficult to reproduce manually, occurring only about once in 50 attempts, and typically during the initial un-cached load. It's also not consistently reproducible even under those conditions (in master). However, the order becomes correct if we reload the page or navigate back to the study list and re-select the study. This order mismatch is causing a test failure in the master branch, which I've also confirmed locally.
> > cc: @sedghi
> 
> Yes I understand. I think it might be best to simply close the left panel prior to taking screenshots and open as necessary with `await page.getByTestId('side-panel-header-left').click();`.

Someone removed the sort order from the display sets on startup.  That has to get added back in again.  Adding it in will make the display consistent in the left hand panel.  That should be a one line change.

jbocce:
> > > @jbocce Previously, the study appeared reversed in the common layout, although it was correct in the MPR layout. This PR addresses that reversal. As a result, the slice order in the common layout will now be different, which is the expected behavior and resolves the underlying issue. This change caused a mismatch with existing screenshots, leading to test failures, so I've updated them accordingly.
> > > However, I've observed that the display set order in the left side panel is inconsistent when running Playwright tests. This issue is difficult to reproduce manually, occurring only about once in 50 attempts, and typically during the initial un-cached load. It's also not consistently reproducible even under those conditions (in master). However, the order becomes correct if we reload the page or navigate back to the study list and re-select the study. This order mismatch is causing a test failure in the master branch, which I've also confirmed locally.
> > > cc: @sedghi
> > 
> > 
> > Yes I understand. I think it might be best to simply close the left panel prior to taking screenshots and open as necessary with `await page.getByTestId('side-panel-header-left').click();`.
> 
> Someone removed the sort order from the display sets on startup. That has to get added back in again. Adding it in will make the display consistent in the left hand panel. That should be a one line change.

Ok so disregard the comment about closing the left panel. In [createStudyBrowserTabs](https://github.com/OHIF/Viewers/blob/a7892c0d59dc77ef94a34a922bbdea338fe1ca52/platform/core/src/utils/createStudyBrowserTabs.ts#L26) there is a check to consistently sort by series UID. So I suspect somewhere in this PR the sorting of the series is breaking because I have had several PRs over the last 24 hours where the tests were not breaking for the thumbnail ordering. @arul-trenser please have a look, but I could be wrong.

jbocce:
@arul-trenser another thing to ensure is that the `TEST_ENV` environment variable is actually being set. So be sure to run and update screen shots with one of the predefined scripts like `bun test:e2e:ui` or `bun test:e2e:update`.

arul-trenser:
> @arul-trenser another thing to ensure is that the `TEST_ENV` environment variable is actually being set. So be sure to run and update screen shots with one of the predefined scripts like `bun test:e2e:ui` or `bun test:e2e:update`.

@jbocce  Thanks for the heads-up! I've updated the screenshots using scripts.

platform/core/src/classes/ImageSet.ts
arul-trenser: > @sedghi Could you please share your thoughts on this?
> 
> Thanks!

@sedghi  Could you confirm if there's anything I should be doing?

sedghi (changes requested):
please add a migration guides too, to tell people about the new sorting, and how they can switch back to the old sorting (if you have not this fallback we should)

arul-trenser:
> please add a migration guides too, to tell people about the new sorting, and how they can switch back to the old sorting (if you have not this fallback we should)

@sedghi  I've already updated that here. https://github.com/OHIF/Viewers/pull/5022/files#diff-576361e4d812c6714e3a50d6eaa728844da7022a97238ba754e6a7da84a6135e

sedghi (approved):

sedghi:
can you check cypress why it fails and fix it

arul-trenser:
> can you check cypress why it fails and fix it

@sedghi The tests for OHIF Cornerstone Hotkeys and OHIF Cornerstone Toolbar both failed, exhibiting timeout errors.

jbocce:
> > can you check cypress why it fails and fix it
> 
> @sedghi The tests for OHIF Cornerstone Hotkeys and OHIF Cornerstone Toolbar both failed, exhibiting timeout errors.

@arul-trenser , I took a quick look at these failing tests and I would say that similar to the playwright tests, these failing cypress tests are due to the images now being in reverse order than what they were previously. Please ensure that the new ordering is correct and update the tests accordingly. Reach out if you need a hand. Thanks!

arul-trenser:
> > > can you check cypress why it fails and fix it
> > 
> > 
> > @sedghi The tests for OHIF Cornerstone Hotkeys and OHIF Cornerstone Toolbar both failed, exhibiting timeout errors.
> 
> @arul-trenser , I took a quick look at these failing tests and I would say that similar to the playwright tests, these failing cypress tests are due to the images now being in reverse order than what they were previously. Please ensure that the new ordering is correct and update the tests accordingly. Reach out if you need a hand. Thanks!

@jbocce Cypress issues fixed, thanks to your guidance!

IbrahimCSAE:
FYI. This has broken the sort on some series, they became completely out of order. I can't share the data but the only information I can share is that it was a sagittal series with an axial cover slice.

Its sorted correctly on viewer.ohif.org, but on viewer-dev.ohif.org is completely messed up @sedghi

sen-trenser:
@IbrahimCSAE Thank you for the feedback!

However, there is a way to switch the sorting behavior to previous version.  Please check if [this](https://github.com/OHIF/Viewers/pull/5022/files#diff-576361e4d812c6714e3a50d6eaa728844da7022a97238ba754e6a7da84a6135eR34-R42) is feasible for your case.

@arul-trenser We should add a check for any cover/localizer slice is there in any of the image by comparing all images' IOP [here](https://github.com/OHIF/Viewers/pull/5022/files#diff-ecbe693c303bf4a2b1398c45ddc3d8f0558893a47c47b3dcba909a077948701fR136-R149) with first image's IOP. 
Notice first axial slice is sorted to last in this series: [v3.10](https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.26968527900428638961173806140069&SeriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.99.1071.19276989507392553891297154495191) vs [v3.11](https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.26968527900428638961173806140069&SeriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.99.1071.19276989507392553891297154495191) 

Also it should default to previous sort fn for secondary captured image series having multiple orientation images within the series. 
I have also noticed an issue with this MRCP MIP data. Notice the instance number text overlay in viewport bottom-right corner. [v3.10](https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.21255249241959598781018667405790&SeriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.99.1071.18655460124733602726552772434272) vs [v3.11](https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.21255249241959598781018667405790&SeriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.99.1071.18655460124733602726552772434272)

cc: @sedghi

arul-trenser:
> FYI. This has broken the sort on some series, they became completely out of order. I can't share the data but the only information I can share is that it was a sagittal series with an axial cover slice.
> 
> Its sorted correctly on viewer.ohif.org, but on viewer-dev.ohif.org is completely messed up @sedghi

The issue mentioned above has been addressed. I've implemented a fix in PR #5223.
 
@IbrahimCSAE  Could you please confirm if this resolves your data sorting problem?

IbrahimCSAE:
Its resolved now, Thanks!

arul-trenser:
@sedghi Could you please review #5223 and approve tha PR

> > FYI. This has broken the sort on some series, they became completely out of order. I can't share the data but the only information I can share is that it was a sagittal series with an axial cover slice.
> > Its sorted correctly on viewer.ohif.org, but on viewer-dev.ohif.org is completely messed up @sedghi
> 
> The issue mentioned above has been addressed. I've implemented a fix in PR #5223.
> 
> @IbrahimCSAE Could you please confirm if this resolves your data sorting problem?

@sedghi Could you please review the PR #5223, I've fixed the above mentioned issue.

PR #5023, opened 2025-05-08
Feat:  Jump to measurement moves the camera if it is not in the viewport

mbellehumeur:
### Context
In microscopy/wsi, it happens frequently that the jump to measurement feature does not work because the measurement is outside the viewport. 

This PR adds a check in jumptomeasurement handler to see if the measurement is outside the viewport.
Only if the measurement is not visible, then the camera is moved to the center of the measurement.
Then if the measurement is larger then the viewport, the camera is zoomed out.
  

### Changes & Results

Added a function to check if measurements are within the current viewport.
- Implemented camera adjustments to center on measurements that are not visible.
- Adjusted zoom level based on measurement size to ensure visibility.

### Testing
- Make measurements of different sizes.
- Zoom in until you do not see them.
- Jump to measurement by clicking in the measurement panel.

### Checklist

#### Code

  etc.)

### DIFF
--- a/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
+++ b/extensions/cornerstone/src/Viewport/OHIFCornerstoneViewport.tsx
@@ -4,6 +4,8 @@ import { Enums, eventTarget, getEnabledElement } from '@cornerstonejs/core';
 import { MeasurementService } from '@ohif/core';
 import { useViewportDialog } from '@ohif/ui-next';
 import type { Types as csTypes } from '@cornerstonejs/core';
+import { vec3 } from 'gl-matrix';
+import { Point3 } from '@cornerstonejs/core/types';
 
 import { setEnabledElement } from '../state';
 
@@ -380,6 +382,28 @@ const OHIFCornerstoneViewport = React.memo(
   areEqual
 );
 
+function isMeasurementWithinViewport(viewport, measurement) {
+  const camera = viewport.getCamera();
+  const { focalPoint, parallelScale } = camera;
+  // Check if the measurement points are inside the extent
+  for (const point of measurement.points) {
+    const [x, y, z] = point;
+    // Calculate the distance from the focal point
+    const dx = x - focalPoint[0];
+    const dy = y - focalPoint[1];
+    const dz = z - focalPoint[2];
+    // Check if the point is within the extent
+    if (
+      Math.abs(dx) > parallelScale ||
+      Math.abs(dy) > parallelScale ||
+      Math.abs(dz) > parallelScale
+    ) {
+      return false; // Point is outside the extent
+    }
+  }
+  return true; // All points are inside the extent
+}
+
 // Helper function to handle jumping to measurements
 function handleJumpToMeasurement(event, elementRef, viewportId, cornerstoneViewportService) {
   const { measurement, isConsumed } = event;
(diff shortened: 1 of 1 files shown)


extensions/cornerstone/src/commandsModule.ts
```diff
@@ -161,6 +162,34 @@ function commandsModule({
...
         viewport.render();
+
+        /**
+         * If the measurement is not visible inside the current viewport,
+         * we need to move the camera to the measurement.
+         */
+        if (!isMeasurementWithinViewport(viewport, measurement)) {
+          const camera = viewport.getCamera();
+          const { focalPoint: cameraFocalPoint, position: cameraPosition } = camera;
+          const focalPoint: Point3 = [
```
wayfarer3130: This only works for axial planar views.  The right way to do this is something like:
const { center, extent } = getCenterExtent(measurement)

where the calculation of center is the average x,y,z point, and the extent is the min/max of the x,y,z points

Then, measurement size is vec3.dist(extent.min, extent.max)

wayfarer3130: We should get a general utility getCenterExtent for measurements, as it is quite common to use that.

extensions/cornerstone/src/utils/isMeasurementWithinViewport.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Checks if all measurement points fall within the visible area defined by the camera's
+ * focal point and parallel scale (extent boundaries).
+ *
+ * @param {Object} params - The parameters object
+ * @param {Object} params.viewport - The viewport object containing camera information
+ * @param {Object} params.measurement - The measurement object containing points to check
+ * @param {Array<Array<number>>} params.measurement.points - Array of 3D points [x, y, z]
+ * @returns {boolean} True if all measurement points are within the viewport extent, false otherwise
+ */
+export const isMeasurementWithinViewport = (viewport, measurement) => {
```
wayfarer3130, line 12: This could use the getMeasurementExtent, and then just test on the extent function.

extensions/cornerstone/src/commandsModule.ts
igoroctaviano: Updated

wayfarer3130: I still see the old code?

extensions/cornerstone/src/utils/isMeasurementWithinViewport.ts
igoroctaviano: Pushed

extensions/cornerstone/src/utils/getCenterExtent.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  // Initialize min and max with the first point
+  const min: [number, number, number] = [...points[0]] as [number, number, number];
+  const max: [number, number, number] = [...points[0]] as [number, number, number];
+
+  // Find the bounding box by iterating through all points
+  for (let i = 1; i < points.length; i++) {
+    const point = points[i];
+    for (let j = 0; j < 3; j++) {
```
wayfarer3130, line 29: vec3.min(min,min,point) is sufficient.

igoroctaviano:
@wayfarer3130

PR #5051, opened 2025-05-16
feat: create memo when manually adding and removing annotations

pedrokohler:
### Context

Currently if you add or delete an annotation through the `measurementService`'s api, the changes are not commited to the recently introduced cs3d history api. This PR adds this functionality so that it works in both cases.

### Changes & Results

Before:

https://github.com/user-attachments/assets/2a09d7de-7240-408f-8098-cda03ba6d5de

After:

https://github.com/user-attachments/assets/bac9fe60-2754-4df6-b6b2-49a62ea15001

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/extensions/cornerstone/src/initMeasurementService.ts
+++ b/extensions/cornerstone/src/initMeasurementService.ts
@@ -12,6 +12,7 @@ import {
   triggerAnnotationRenderForViewportIds,
 } from '@cornerstonejs/tools/utilities';
 import getActiveViewportEnabledElement from './utils/getActiveViewportEnabledElement';
+import { triggerCreateAnnotationMemo } from './utils/triggerCreateMemo';
 
 const { CORNERSTONE_3D_TOOLS_SOURCE_NAME, CORNERSTONE_3D_TOOLS_SOURCE_VERSION } = CSExtensionEnums;
 const { removeAnnotation } = annotation.state;
@@ -350,7 +351,13 @@ const connectMeasurementServiceToTools = (
       if (source.name !== CORNERSTONE_3D_TOOLS_SOURCE_NAME) {
         continue;
       }
+      const removedAnnotation = annotation.state.getAnnotation(uid);
       removeAnnotation(uid);
+      triggerCreateAnnotationMemo({
+        annotation: removedAnnotation,
+        FrameOfReferenceUID: removedAnnotation.metadata.FrameOfReferenceUID,
+        options: { deleting: true },
+      });
     }
 
     // trigger a render
@@ -441,7 +448,7 @@ const connectMeasurementServiceToTools = (
        * This is not the read-only annotation rendered by the SR viewport.
        */
       const annotationManager = annotation.state.getAnnotationManager();
-      annotationManager.addAnnotation({
+      const newAnnotation = {
         annotationUID: measurement.uid,
         highlighted: false,
         isLocked: false,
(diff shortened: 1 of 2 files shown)


extensions/cornerstone/src/initMeasurementService.ts
```diff
@@ -350,7 +351,13 @@ const connectMeasurementServiceToTools = (
       if (source.name !== CORNERSTONE_3D_TOOLS_SOURCE_NAME) {
         continue;
       }
+      const removedAnnotation = annotation.state.getAnnotation(uid);
       removeAnnotation(uid);
+      triggerCreateAnnotationMemo({
```
sedghi, line 356: how about a command in commandModule and more generic that annotation maybe? or it has modes, annotation, segmentation etc in future

pedrokohler: I added it to the command modules but I think it's best if we leave separate methods for different types. Even the API of the labelmap segmentation createMemo is different, it takes different arguments such as a voxelManager.

extensions/cornerstone/src/init.tsx
```diff
@@ -179,7 +179,10 @@ export default async function init({
   initWADOImageLoader(userAuthenticationService, appConfig, extensionManager);
 
   /* Measurement Service */
-  this.measurementServiceSource = connectToolsToMeasurementService(servicesManager);
+  this.measurementServiceSource = connectToolsToMeasurementService({
```
wayfarer3130, line 182: This will have to be documented in the migration guide.
Also, my preference is to just pass in the services object so that anyone needing any of the core services just gets the services, rather than having anyone "know" about what needs to be passed in.

extensions/cornerstone/src/initMeasurementService.ts
```diff
@@ -12,6 +12,7 @@ import {
   triggerAnnotationRenderForViewportIds,
 } from '@cornerstonejs/tools/utilities';
 import getActiveViewportEnabledElement from './utils/getActiveViewportEnabledElement';
+import { servicesManager } from 'platform/app/src/App';
```
wayfarer3130, line 15: You can't import things this way - you need to import from a package itself, as the extensions should never try to directly reference a source file in the platform.

extensions/cornerstone/src/initMeasurementService.ts
```diff
@@ -188,25 +189,26 @@ const initMeasurementService = (
   return csTools3DVer1MeasurementSource;
 };
 
-const connectToolsToMeasurementService = (servicesManager: AppTypes.ServicesManager) => {
+const connectToolsToMeasurementService = ({
+  commandsManager,
+  servicesManager,
+}: {
```
wayfarer3130, line 195: Please just pass the props through to children calls, so that if this gets updated to need the extension manager at some point or some other new manager, it can just be included withotu further change.

extensions/cornerstone/src/initMeasurementService.ts
```diff
@@ -350,7 +350,13 @@ const connectMeasurementServiceToTools = (
       if (source.name !== CORNERSTONE_3D_TOOLS_SOURCE_NAME) {
         continue;
       }
+      const removedAnnotation = annotation.state.getAnnotation(uid);
       removeAnnotation(uid);
+      commandsManager.run('triggerCreateAnnotationMemo', {
```
wayfarer3130, line 355: Is there a way to bulk add them as a single memo?  The undo you undo as a single operation every cleared item all at once.

wayfarer3130 (changes requested):
Please update the migration guide to include the change in parameters for the connect tools call, and use the props for the function argument to pass  all the props to the children when you are passing servicesManager/commandsManager/... to children.

wayfarer3130:
Don't worry about the image consistency check - I'm going to try looking at that as a separate PR to fix it.

wayfarer3130 (approved):

PR #5056, opened 2025-05-20
feat(PanelStudyBrowser): Added a customizable studyBrowser.thumbnailClickCallback handler

rinslow:
### Context

Modern web application users generally expect to interact with UI components via single clicks, whereas double-click interactions are increasingly considered non-intuitive and outdated.

This PR introduces support for single-click interactions on study thumbnails within the viewer, enhancing usability and aligning with standard UX practices. The change is implemented in a customizable manner to allow flexibility. While single-click behavior is introduced as the new default, I am open to making this configurable-only if there is a compelling reason to retain double-click as the default behavior.

### Changes & Results

- Bugfix: Prevented a null error when the double click handler is removed 
- Feat: Added a customizable studyBrowser.thumbnailClickCallback handler, complementary to studyBrowser.thumbnailDoubleClickCallback. 
- Changed the default behavior of clicking thumbnails to be a single click
- Bugfix: Cypress artifacts were not stored because of incorrect paths 

<video width="1192" src="https://github.com/user-attachments/assets/d9ae99dc-25a6-46f7-9642-8935e1afa0c6"></video>

image

### Testing

- Test with/without on-click handler, configuration can be changed at `extensions/default/src/customizations/studyBrowserCustomization.ts` 

#### PR

#### Code

  etc.)

  additions or removals. 

** Didn't find documentation about the customization options ** 

#### Tested Environment

- [x] OS: MacOS 
- [x] Node version: v20.18.2
  

### DIFF
--- a/extensions/default/src/customizations/studyBrowserCustomization.ts
+++ b/extensions/default/src/customizations/studyBrowserCustomization.ts
@@ -66,33 +66,24 @@ export default {
   'studyBrowser.studyMode': 'all',
   'studyBrowser.thumbnailDoubleClickCallback': {
     callbacks: [
-      ({ activeViewportId, servicesManager, commandsManager, isHangingProtocolLayout }) =>
+      ({ activeViewportId, commandsManager, servicesManager }) =>
         async displaySetInstanceUID => {
-          const { hangingProtocolService, uiNotificationService } = servicesManager.services;
-          let updatedViewports = [];
-          const viewportId = activeViewportId;
-
           try {
-            updatedViewports = hangingProtocolService.getViewportsRequireUpdate(
-              viewportId,
-              displaySetInstanceUID,
-              isHangingProtocolLayout
-            );
+            commandsManager.run('loadSegmentationDisplaySetsForViewport', {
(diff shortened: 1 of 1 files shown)


extensions/default/src/customizations/studyBrowserCustomization.ts
```diff
@@ -88,6 +88,36 @@ export default {
...
 
+          commandsManager.run('setDisplaySetsForViewports', {
+            viewportsToUpdate: updatedViewports,
+          });
+        },
+    ],
+  },
+  'studyBrowser.thumbnailClickCallback': {
+    callbacks: [
+      ({ activeViewportId, servicesManager, commandsManager, isHangingProtocolLayout }) =>
```
wayfarer3130, line 99: This action should be added to a commands module, and the customization should simply indicate which command should be called.  That way the code doesn't get repeated, and it is as easy as just specifying the command name to use.
If you do that for double click and single click, then either one can be bound to load the display set without duplicating how it is handled, and you can also do custom behaviour like loading the DICOM SR on double click, and displaying it with single click.

Issue #4241, opened 2024-06-19
[Bug] navigateHistory command is not available on startup


mbellehumeur:
### Describe the Bug

The navigateHistory command of commandsManager is initialized in Mode.tsx therefore it is not available right after the client is started; before a study is opened.  Adding the initialization in Worklist.tsx (in addition to Mode.tsx) fixed the problem for me:
  // Expose the react router dom navigation
  history.navigate = useNavigate();

### Steps to Reproduce

Call: 
commandsManager.runCommand('navigateHistory',{to:'/viewer?StudyInstanceUIDs='+studyUID}); }
after client startup.  The error is 'navigate function does not exist'.

### The current behavior

The navigateHistory command fails and the study does not open.

### The expected behavior

The navigateHistory command runs and the study opens.

### OS

macOS

### Node version

18

### Browser

Chrome

mbellehumeur:
This is needed for the FHIRCast feature - to open studies on receiving command from hub.

PR #5062, opened 2025-05-22
fix(router): Expose the react router from the worklist too.

jbocce:
### Context
See #4241 

### Changes & Results
The worklist now sets `history.navigate` so that it is available on the worklist page too.

### Testing

#4241  should no longer be a bug.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 23.9.0
- [x] Browser: Chrome
  


sedghi:
is this related to this? which approach is better?

https://github.com/OHIF/Viewers/pull/4899/files

jbocce:
> is this related to this? which approach is better?
> 
> https://github.com/OHIF/Viewers/pull/4899/files

@sedghi yes it is related. Let me start my apologizing as I did not scroll down far enough when looking at the issue to realize that a PR was already started.

As to which is better, well I'm not sure. Doing the conditional assignment as I did makes sense to me. That said, considering the `useNavigate()` is called on every render regardless and likely returns an equivalent if not the same reference each time means that either approach is fine IMHO. Let me know. I can close this PR. Sorry again and thanks for pointing this out.

jbocce:
@sedghi I pushed updates that take a slightly different approach.

Instead of multiple (i.e. two locations) to set the history navigate function, do it in one location in the RouteWithErrorBoundary.
To avoid conditional warnings/errors for calling `useNavigate` I changed the history to have a getter and setter.

Anyway I am open to other alternatives too. Let me know what you think.

sedghi (approved):
Looks great thanks

PR #5064, opened 2025-05-25
feat(PanelStudyBrowser): Added a customizable studyBrowser.thumbnailClickCallback handler 

rinslow:
### Context

Modern web application users generally expect to interact with UI components via single clicks, whereas double-click interactions are increasingly considered non-intuitive and outdated.

This PR introduces support for single-click interactions on study thumbnails within the viewer, enhancing usability and aligning with standard UX practices. The change is implemented in a customizable manner to allow flexibility. While single-click behavior is introduced as the new default, I am open to making this configurable-only if there is a compelling reason to retain double-click as the default behavior.

### Changes & Results

- Bugfix: Prevented a null error when the double click handler is removed 
- Feat: Added a customizable studyBrowser.thumbnailClickCallback handler, complementary to studyBrowser.thumbnailDoubleClickCallback. 
- Changed the default behavior of clicking thumbnails to be a single click
- Bugfix: Cypress artifacts were not stored because of incorrect paths 

<video width="1192" src="https://github.com/user-attachments/assets/d9ae99dc-25a6-46f7-9642-8935e1afa0c6"></video>

image

### Testing

- Test with/without on-click handler, configuration can be changed at `extensions/default/src/customizations/studyBrowserCustomization.ts` 

#### PR

#### Code

  etc.)

  additions or removals. 

** Didn't find documentation about the customization options ** 

#### Tested Environment

- [x] OS: MacOS 
- [x] Node version: v20.18.2
  


extensions/default/src/customizations/studyBrowserCustomization.ts
```diff
@@ -66,32 +66,24 @@ export default {
...
-          let updatedViewports = [];
-          const viewportId = activeViewportId;
-
           try {
-            updatedViewports = hangingProtocolService.getViewportsRequireUpdate(
-              viewportId,
-              displaySetInstanceUID,
-              isHangingProtocolLayout
-            );
+            commandsManager.run('loadSegmentationDisplaySetsForViewport', {
```
sedghi, line 72: Hmmm i don't follow this logic, why we are saying the default is loadSegmentation?

rinslow: Looking at the implementation of loadSegementationDisplaySetsForViewport it seems exactly the same. Am i missing something? still a bit new to this repo

sedghi: there is nothing segmentation wise for the study browser

rinslow: ok, my bad! undone

rinslow:
@sedghi kindly, can you take another look at this one?

sedghi:
@wayfarer3130 could you review this please

extensions/default/src/Panels/StudyBrowser/PanelStudyBrowser.tsx
```diff
@@ -102,6 +107,40 @@ function PanelStudyBrowser({
...
+      if (!customHandler) {
+        return;
+      }
+
+      const setupArgs = {
+        activeViewportId,
+        commandsManager,
+        servicesManager,
+        isHangingProtocolLayout,
+        appConfig: extensionManager._appConfig,
```
wayfarer3130, line 125: Does the app config actually need to be passed here?  Would rather pass the extension manager than extracting specific parts.

extensions/default/src/Panels/StudyBrowser/PanelStudyBrowser.tsx
wayfarer3130, line 1: Does this work with a different binding for double click from single click?  That is, can I bind double click to one command, and single click to a different one, and have both of them execute correctly?

extensions/default/src/customizations/studyBrowserCustomization.ts
```diff
@@ -94,4 +94,5 @@ export default {
         },
     ],
   },
+  'studyBrowser.thumbnailClickCallback': {},
```
wayfarer3130, line 97: Do you have an example of actually enabling the customization - that can go into the test extension examples if you want, or you can leave it here as a non-default customization and then demo it in the test mode.
The best sort of customization for this would be a pair of command names to execute on single/double click.  It could be extended in the future to add things like right click as well (not in this PR)

platform/app/cypress.config.ts
```diff
@@ -33,5 +33,9 @@ export default defineConfig({
       mochaFile: 'cypress/results/test-output.xml',
       toConsole: true,
     },
+
+    // We go back two levels up to the repository root folder
+    videosFolder: '../../cypress/videos',
```
jbocce, line 38: I think it might be best  to either not commit this OR place the videos and screenshots in the ../cypress/results directory and add the appropriate .gitignore entries since we don't want those types of things in git. Thanks.

extensions/default/src/Panels/StudyBrowser/PanelStudyBrowser.tsx
```diff
@@ -102,6 +107,40 @@ function PanelStudyBrowser({
     ]
   );
 
+  const onClickThumbnailHandler = useCallback(
```
jbocce, line 110: There is a bit of code duplication here. Please refactor this and the callback above to one method that takes the customization as a parameter.

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/PanelStudyBrowserTracking.tsx
```diff
@@ -135,6 +135,7 @@ export default function PanelStudyBrowserTracking({
       customMapDisplaySets={mapDisplaySetsWithTracking}
       onClickUntrack={onClickUntrack}
       onDoubleClickThumbnailHandlerCallBack={checkDirtyMeasurements}
+      onClickThumbnailHandlerCallBack={checkDirtyMeasurements}
```
jbocce, line 138: I don't believe there is a need for two callbacks here. Perhaps just refactor to one callback and call it something appropriate, say `onThumbnailActivatedHandler1 or somethinig.

extensions/default/src/customizations/studyBrowserCustomization.ts
```diff
@@ -94,4 +94,5 @@ export default {
         },
     ],
   },
```
jbocce, line 96: Here there is the potential for code duplication. In fact, it might be best to refactor these two customization points and make the callbacks commands. Please consider this. Thanks.

jbocce (changes requested):
We are planning a code freeze at the end of this week. So please make updates soon to have this include in the release. Thanks for you contribution.

Issue #5058, opened 2025-05-22
[Bug] The first series is loaded into the MPR mode when selecting the MPR mode after selecting the multi layout(1x1) for the second series.


abhijith-trenser:
### Describe the Bug

If the data is a multi series study and the user load the second series into the MPR mode, then switches to the 1x1 layout and again selects the MPR mode. At that time, the first series is loaded into the MPR mode.

https://github.com/user-attachments/assets/af7112b8-bb88-4921-bb7f-30c27d6f91e6

### Steps to Reproduce

1. Launch a multi-series study.
2. Select the MPR mode.
3. Load the second series into the MPR mode.
4. Select 1x1 layout.
5. Again select the MPR mode.
6. Observe the viewport.

### The current behavior

The first series is loaded into the MPR mode when selecting the MPR mode after selecting the 1x1 layout for the second series.

### The expected behavior

The second series should be loaded into the MPR mode when selecting MPR mode from the second series.

### OS

Windows 11

### Node version

22

### Browser

136.0.7103.114

jbocce:
Fixed with PR #5074

Issue #4982, opened 2025-04-21
[Bug] Frame View loads series with lowest series number


rghgit1:
### Describe the Bug

If there are multiple series with series number 1, 5, 10, 15, 20. Most of the times Frame View loads series 1, even though I try to do Frame View for series 20.

### Steps to Reproduce

1. Open a study with multiple series having different series number like 1, 5, 10, 15, 20
2. Load Series with series number 20 to viewport
3. Click on layout selector and select Frame View
4. Loads series with series number 1, instead of series number 20

### The current behavior

Frame View loads series with lowest series number

### The expected behavior

Frame View should load the series selected in the Viewport and not the series with lowest series number

### OS

Windows 11

### Node version

18.18.1

### Browser

Microsoft Edge Version 135.0.3179.85 (Official build) (64-bit)

jbocce:
This issue is the same problem as #5058

jbocce:
Fixed with PR #5074

PR #5074, opened 2025-05-27
fix(hp): When reapplying an hp the display set selectors should be considered.

jbocce:
### Context
#4982 and #5058 

### Changes & Results
When reapplying an hp the display set selectors should be considered.
The frameView HP now uses the activeDisplaySet display set selector id.

### Testing
Both #4982 and #5058 should work as intended.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 20.9.0
- [x] Browser: Chrome 137.0.7151.41
  


extensions/cornerstone/src/hps/frameView.ts
jbocce, line 1: Just like various other hps (e.g. mpr), use the activeDisplaySet display set selector id.

extensions/cornerstone/src/synchronizers/frameViewSynchronizer.ts
```diff
@@ -43,7 +43,8 @@ const createFrameViewSynchronizer = (synchronizerName: string): Synchronizer =>
   const synchronizer = SynchronizerManager.createSynchronizer(
     synchronizerName,
     EVENTS.STACK_VIEWPORT_SCROLL,
-    frameViewSyncCallback
+    frameViewSyncCallback,
+    { auxiliaryEvents: [{ name: EVENTS.CAMERA_MODIFIED, source: 'element' }] }
```
jbocce, line 47: I kinda messed up on a previous PR. The camera modified event is still needed when the HP is initially opened.

extensions/default/src/commandsModule.ts
```diff
@@ -395,6 +395,7 @@ const commandsModule = ({
           hangingProtocolService.setProtocol(protocolId, {
             stageId,
             stageIndex: useStageIdx,
+            displaySetSelectorMap,
```
jbocce, line 398: I think this is the correct thing here because since we are reapplying the hp (and stage) we should also apply the same display set that it might be displaying.

extensions/cornerstone/src/hps/frameView.ts
```diff
@@ -9,7 +9,7 @@ const frameView: Types.HangingProtocol.Protocol = {
   toolGroupIds: ['default'],
   protocolMatchingRules: [],
   displaySetSelectors: {
-    defaultDisplaySetId: {
+    activeDisplaySet: {
```
wayfarer3130, line 12: Why are you renaming this to active display set?  There is a difference between the default display set and the active one.

jbocce: It is being renamed so that if the HP is reapplied it will use whatever display set was swapped in. For instance take this scenario.., 
1. start with the frame view HP 
2. double click to one-up
3. drag in a different series
4. reapply (i.e. choose) the frame view HP
5. The display set dragged in from 3 should display. Without this, the original display set from 1 will show.

jbocce: I think this is similar to what the MPR HP does.

extensions/default/src/utils/reuseCachedLayouts.ts
```diff
@@ -61,7 +61,8 @@ const reuseCachedLayout = (state, hangingProtocolService: HangingProtocolService
       if (viewportId === activeViewportId && i === 0) {
         setDisplaySetSelector(`${activeStudyUID}:activeDisplaySet:0`, displaySetUID);
       }
-      if (displaySetOptions[i]?.id) {
+      // The activeDisplaySet selector should only be set once (i.e. for the actual active display set)
+      if (displaySetOptions[i]?.id && displaySetOptions[i].id !== 'activeDisplaySet') {
```
wayfarer3130, line 65: Can you remove the -1 from the frame view for the offset index?  I believe that will mean you don't need this change here.  Otherwise I think the display is in error.

extensions/cornerstone/src/hps/frameView.ts
wayfarer3130: Yes, I understand now what it is doing - I was confused with the matchedDisplaySetsIndex value, which is causing an issue.

wayfarer3130 (changes requested):
Can you retest with the removed attribute - I pushed that, and if it is causing issues, you will have to restore your other change, but regardless i don't believe the -1 matched frame is correct.  I think without that it will not call the reuse display set in the context that it gets the "next" one - the matchedDisplaySetsIndex: -1 means to take the next display set after the ones which are already in view, and I think that is why the other call is happening.

jbocce:
> Can you retest with the removed attribute - I pushed that, and if it is causing issues, you will have to restore your other change, but regardless i don't believe the -1 matched frame is correct. I think without that it will not call the reuse display set in the context that it gets the "next" one - the matchedDisplaySetsIndex: -1 means to take the next display set after the ones which are already in view, and I think that is why the other call is happening.

I had to restore the check in `reuseCachedLayouts` for two reasons...
1. The `activeDisplaySet` entry should ONLY be set for the active viewport - doing this multiple times is extra work which is of no value.
2. The following scenario would break otherwise...
    - Open a study using the `frameView` hp.
    - Drag in a different series into a "middle" viewport.
    - With the newly dragged in series as the active viewport, relaunch the frame view hp.
    - That series should "fill" the view.

So the second scenario could be averted by setting `allowUnmatchedViews` to false for the default viewport in the HP. However, for the frame view HP,  dragging in a different series appears to cause OHIF to "kinda loose its mind". That said, the check I added makes those HPs with `allowUnmatchedViews` set to `true` behave consistently.

wayfarer3130 (approved):

PR #5079, opened 2025-05-29
feat(hooks): Added useCustomization hook

rinslow:
### Context

Added a `useCustomization` hook, which allows React components to read customization values, and re-render when they change. (e.g. via `setCustomizations()`) 

I've added an example customization for thumbnails. 
My use case was highlighting the borders via an AI tool. 

### Changes & Results

* Created useCustomization hook, with public documentation.
* Added customizations of type CSS classes and CSS styles. 
* Added a usage of useCustomization in `Thumbnail.tsx`, which uses modified styles for the displayset. 

https://github.com/user-attachments/assets/d0925e2b-98e1-4904-ab6d-e25e933cc9e6

### Testing

Can be tested with these console commands:
image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: MacOS 
- [x] Node version: v20.18.2
  

### DIFF
--- a/platform/ui-next/src/components/Thumbnail/Thumbnail.tsx
+++ b/platform/ui-next/src/components/Thumbnail/Thumbnail.tsx
@@ -5,6 +5,8 @@ import { useDrag } from 'react-dnd';
 import { Icons } from '../Icons';
 import { DisplaySetMessageListTooltip } from '../DisplaySetMessageListTooltip';
 import { TooltipTrigger, TooltipContent, Tooltip } from '../Tooltip';
+import { useCustomization } from '@ohif/core';
+import { CSSStylesCustomization } from '@ohif/core/src/services/CustomizationService/types';
 
 /**
  * Display a thumbnail for a display set.
@@ -47,6 +49,10 @@ const Thumbnail = ({
 
   const [lastTap, setLastTap] = useState(0);
 
+  const thumbnailStylesCustomization = useCustomization(`ui.thumbnail#${displaySetInstanceUID}`) as
+    | CSSStylesCustomization
+    | undefined;
+
   const handleTouchEnd = e => {
     const currentTime = new Date().getTime();
     const tapLength = currentTime - lastTap;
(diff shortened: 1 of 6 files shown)


platform/ui-next/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -5,6 +5,8 @@ import { useDrag } from 'react-dnd';
 import { Icons } from '../Icons';
 import { DisplaySetMessageListTooltip } from '../DisplaySetMessageListTooltip';
 import { TooltipTrigger, TooltipContent, Tooltip } from '../Tooltip';
+import { useCustomization } from '@ohif/core';
+import { CSSStylesCustomization } from '@ohif/core/src/services/CustomizationService/types';
```
sedghi, line 9: I'm not really sure about this. I like the idea of a `useCustomization` hook, but importing from `ohif/core` isn't ideal. If you search for `ohif/core` imports in the `ui-next` folder, you'll probably find none, or if there are any, they're really necessary.

Also, I don't think you need this customization to apply this logic. You could have listeners for the `useCustomization` changes and just target the DOM elements to apply your styles. Instead of that, we're introducing a whole other naming schema for customizing bits and pieces of the UI. I mean, you can use `querySelector` to select the element you want to customize.

rinslow: 1. where should I place useCustomization then? 
2. The issue is that some elements do not have unique classnames or ids, so using the query selector will be prone to breaking by ui changes (for example, changing structure or tailwind classes)

In my use case, we're wrapping OHIF with an AI tool that finds pathologies in X-Rays. 
When a user asks about a pathology, we want to show them in which images the pathology exists, so we highlight it.

sedghi: I'd rather add a unique ID to each element through the app. I don't mind you submitting pull requests to add proper IDs to each of our UI elements. Then, you can place the `useCustomization` hook in your extension/mode, allowing you to target each element and apply your style.

Basically, we have two options:

1.  Add `useCustomization` to each UI element and then target it using each ID, which isn't documented anywhere.
2.  Add an ID to each UI element.

I favor the second method because it doesn't tie the `useCustomization` implementation to the UI.

Issue #4785, opened 2025-02-15
[Bug] After the crosstools is rotated in the mpr view, the measurement panel will report an error when clicking to jump


shaoding:
### Describe the Bug

After performing a length measurement on the axial plane, attempting to rotate the crosshair tool on the sagittal plane and clicking the measurement to jump results in an error.

### Steps to Reproduce

1.Enter into a [study](https://viewer.ohif.org/viewer?StudyInstanceUIDs=2.16.840.1.114362.1.11972228.22789312658.616067305.306.2)
2.Switch to the mpr layout
3.Perform length measurements on the axial plane
4.Open the crosshair tool and rotate the crosshair in the sagittal plane
5.Click on the previous length measurement data in the measurement panel

### The current behavior

<img width="1907" alt="Image" src="https://github.com/user-attachments/assets/d7f865cf-5d95-4cbc-afad-90f5d36068cb" />

Experience lag on the page.The crosshair function is abnormal.

### The expected behavior

Can jump normally, the crosshair function works properly, and no errors are reported.

### OS

macOS 15.1.1

### Node version

20.15.1

### Browser

Chrome 133.0.6943.54

shaoding:
Calling the **viewport.setViewReference** method to switch from orthographic slice to oblique slice, or from oblique slice to orthographic slice, will result in the above error, leading to an infinite loop.

shaoding:
Regardless of whether the annotations are on the orthogonal plane or on the oblique section, after rotating the camera using the crosshair and clicking the measurement panel, I hope the camera can rotate back to the plane where the annotations are located, activate the viewport containing the annotations, and highlight the annotations. In the absence of the crosshair rotation of the camera, the interaction logic is as I described. Now it may be necessary to add camera rotation and correct the operations of other viewport cameras.

nithin-trenser:
[...]

PR #5090, opened 2025-06-02
 Fix: Jump to measurement from measurement panel after rotating viewport

nithin-trenser:
### Context

- Fix for https://github.com/OHIF/Viewers/issues/4785

### Changes & Results

The issue was resolved by resetting the viewport to the plane where the measurement is located and highlighting it when a measurement is selected from the measurement panel.

https://github.com/user-attachments/assets/b5e5556e-d248-4dd5-9b37-1c034a3c2f55

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
+++ b/extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
@@ -577,21 +577,7 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
 
     // Compute view-plane alignment scores for all viewports to prefer the one
     // requiring the least orientation change when navigation-only is not possible.
-    const viewportAlignmentData = [];
-
-    for (const id of this.viewportsById.keys()) {
-      const viewport = this.getCornerstoneViewport(id);
-      const { viewPlaneNormal } = viewport.getCamera();
-      const { viewPlaneNormal: refViewPlaneNormal } = metadata;
-
-      if (viewPlaneNormal && refViewPlaneNormal) {
-        const alignmentScore = Math.abs(vec3.dot(viewPlaneNormal, refViewPlaneNormal));
-        viewportAlignmentData.push({ viewportId: id, alignmentScore });
-      }
-    }
-
-    // Try best-aligned viewports first
-    viewportAlignmentData.sort((a, b) => b.alignmentScore - a.alignmentScore);
+    const viewportAlignmentData = this.getViewportAlignmentData(metadata);
 
     // See if any viewport could show this with an orientation change
     for (const { viewportId: id } of viewportAlignmentData) {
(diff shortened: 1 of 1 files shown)


nithin-trenser:
@sedghi Could you please review and merge this PR?

jbocce:
@nithin-trenser, We are planning a code freeze at the end of this week. So please make updates soon to have this include in the release. Thanks for you contribution.

nithin-trenser:
@jbocce Conflicts resolved and committed. Could you please re-review and approve?

platform/app/src/components/ViewportGrid.tsx
```diff
@@ -190,6 +192,17 @@ function ViewerViewportGrid(props: withAppTypes) {
             })
           ) {
             canAnyViewportDisplayMeasurement = true;
+          } else {
+            const isSameMeasurementOrientation =
+              getClosestOrientation(measurement.metadata.viewPlaneNormal) ===
+              viewport?.viewportOptions?.orientation;
+            if (isSameMeasurementOrientation) {
+              const { cornerstoneViewportService } = servicesManager.services;
```
sedghi: we should not use the cornerstoneViewportService in platofrm/app

sedghi (changes requested):
platform/app should not use cornerstoneViewportService

nithin-trenser:
Upon revisiting the issue, it was observed that the implementation of `jumpToMeasurement` has changed. Therefore, the fix provided in this PR is no longer applicable. As the issue still persists in the latest version (3.12.0), further analysis will be conducted later to determine an appropriate solution.

arul-trenser:
On further analysis found the below issues
**Reproduction Steps**

- Draw an annotation in the axial viewport.
- Enable the crosshair.
- Rotate the crosshair in the sagittal or coronal viewport.
- Click on the measurement.

**Observed Behavior**

1. A “Maximum call stack exceeded” error occurs.
2. After a short delay, the measurement appears active in both the previously active viewport and the axial viewport, both shown in axial orientation.

https://github.com/user-attachments/assets/02da99dc-864e-412b-9898-4f0e79251ae1

**Note**
This works correctly in the following scenarios:

- When the active viewport has the same orientation as the measurement.
- When the crosshair is disabled

https://github.com/user-attachments/assets/922ac34a-d318-4f27-af52-0fe255ca768a

**Issue 1:** Measurement jump occurs in the active viewport

**Root Cause:**
In `findNavigationCompatibleViewportId`, the viewport selection logic prioritizes viewports based on the following order:

1. Active viewport that can navigate to the image without orientation change
2. Other viewport that can navigate to the image without orientation change
4. Active viewport that can navigate to the image with orientation change
5. Other viewport that can navigate to the image with orientation change

When the viewport is rotated using the crosshair, navigation without orientation change is no longer possible. As a result, the logic falls back to selecting the active viewport that can change orientation, which is incorrectly chosen as the best fit.

**Fix**
Before evaluating viewports that require orientation changes, all viewports are first sorted by least orientation difference using a computed view-plane alignment score. Navigation with orientation change is then evaluated in this order, ensuring the most suitable viewport is selected.

**Issue 2:** Maximum call stack exceeded error when crosshair is enabled

**Root Cause**
After selecting the target viewport, navigation to the measurement triggers `setViewReference`, which calls `setOrientation`.
This leads to the following call sequence:

- `setOrientation` → `applyViewOrientation`
- `applyViewOrientation` → `resetCamera({ resetOrientation: false, resetRotation: false })`
- `resetCamera` triggers `CAMERA_RESET`
- `CAMERA_RESET` resets crosshairs
- Crosshair reset triggers another camera reset with `resetRotation: true`

After applying the orientation, `setViewReference` continues under the assumption that the viewport is aligned with the measurement’s orientation. However, the repeated camera resets triggered by the crosshair revert the orientation. This causes `setViewReference` to continuously reapply the orientation, resulting in an infinite loop and eventually a maximum call stack exceeded error.

**Fix**
Suppress camera reset events during `setOrientation` to prevent recursive camera and crosshair resets.

I have:

Created a PR in CS3D https://github.com/cornerstonejs/cornerstone3D/pull/2519 to fix the maximum call stack exceeded issue.
Updated this PR to address the viewport selection logic for measurement navigation.

@sedghi  Could you please review both this PR and the corresponding CS3D PR?

wayfarer3130:
@wayfarer3130 is reviewing these changes, will check and then merge if it looks good.  I was noticing this hang as well, so I'm glad to have this in for the next release.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -562,33 +562,41 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
+    // Compute view-plane alignment scores for all viewports to prefer the one
+    // requiring the least orientation change when navigation-only is not possible.
+    const viewportAlignmentData = [];
+
+    for (const id of this.viewportsById.keys()) {
+      const viewport = this.getCornerstoneViewport(id);
+      const { viewPlaneNormal } = viewport.getCamera();
+      const { viewPlaneNormal: refViewPlaneNormal } = metadata;
+
+      if (viewPlaneNormal && refViewPlaneNormal) {
```
wayfarer3130: Ideally this should use the sum of the dot products of the two in-plane vectors rather than the view plane normal.  Otherwise on restore you will see the same hanging issue for tools which do not have the ability to restore the refViewPlaneNormal.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -562,33 +562,41 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
-    // the active viewport could change orientations to show this
-    if (
-      activeViewport?.isReferenceViewable(metadata, { withNavigation: true, withOrientation: true })
-    ) {
-      return activeViewportId;
+    // Compute view-plane alignment scores for all viewports to prefer the one
+    // requiring the least orientation change when navigation-only is not possible.
+    const viewportAlignmentData = [];
+
+    for (const id of this.viewportsById.keys()) {
```
wayfarer3130: Could you extract this logic into a separate function so it can be shared with the logic where there is no viewport showing that display set and we want to pick the viewport with the closest orientation?

wayfarer3130: That PR is https://github.com/OHIF/Viewers/pull/5630

wayfarer3130: Just trying to do the changes here locally so we can get this in.

PR #5091, opened 2025-06-02
fix(measurement-group): all measurements were appearing under a single study

IbrahimCSAE:


sedghi:
What are the results before and after?

IbrahimCSAE:
> What are the results before and after?

CleanShot 2025-06-05 at 01 36 22@2x

this is after the fix, before this, if you have measurements on a current/prior the table headers would both show the same study (active viewport study), which is not correct, It should not be based on the active viewport study, but the measurement data itself, and pull the correct information from there.

Issue #5087, opened 2025-05-30
[Bug] Add Segmentation gives error in Route


carjohnson:
### Describe the Bug

Using the Segmentation mode, after adding a segment, I get a route error when I return to the Study List.
This is happening on a fresh download of OHIF (tested on versions 3.11.0-beta.32, 3.11.0-beta.42, 3.10.1) and run with no customization.

------------------------------------------------------------------
NOTE: 
This error was consistent when using FireFox. I did not see the error in Chrome or Edge.
As I kept testing this in the various browsers, the error became inconsistent.

### Steps to Reproduce

1. Download zip and extract files
2. >> yarn install and >> yarn run
3. Select Anonymous study (as an example) > Segmentation Mode
4. Add Segment
5. Using Brush tool - create a segment
6. Return to Study List

This is when I would see the pop-up error for the ./ route

### The current behavior

When returning to Study List page, I get the route error. The first line in the error stack is:

normalizeSegmentationInput@http://localhost:3000/ohif/vendors-node_modules_cornerstonejs_tools_dist_esm_utilities_segmentation_index_js.js:5944:39

![Image](https://github.com/user-attachments/assets/1aeea4d5-0e30-4c8a-b7b1-0e02ac0c076b)

I also saw this in the development tools:

![Image](https://github.com/user-attachments/assets/45f5dcbd-b127-4b8e-98a0-f4c26797bced)

---------------------------------------------------
I don't know if this is related, but, I started testing this because I am creating a customized mode that has segmentation tools available. 
I was consistently getting this error in FireFox until  I changed the hangingProtocol from '@ohif/mnGrid' to 'default' and now the error in my customized mode seems to be gone.

### The expected behavior

No route error should appear when exiting the Segmentation mode back to the Study List

### OS

Windows 10

### Node version

20.12.2

### Browser

Firefox v138.0.4 (64bit))

jbocce:
Fixed with PR #5092

PR #5092, opened 2025-06-02
fix(segmentation): Allow for debounced pub/sub events to be reliably cancelled when unsubscribed to.

jbocce:
### Context
See #5087.

### Changes & Results
The approach to fixing this bug was done in stages per se.

At first it was noticed that the callbacks where the exception was thrown were not being unsubscribed to in a timely manner. So the first logical step was to ensure they were unsubscribed to as the very first thing in the `onModeExit` call. This did NOTHING to fix the problem, however the code was left as in practice unsubscribing "earlier" is the right thing to do.

The next item noticed was that the `SegmentationService` that the callbacks listen to was blindly clearing all the listeners in its `destroy` method prior to the `unsubscribe` calls. In practice this  clearing of the listeners is equivalent to `unsubscribe` calls except of course if the callbacks are debounced and just so happen are queued when the listeners are cleared. This resulted in the `SegmentationService` calling `reset` on the pub/sub service. The `reset` was coded to now not only clear the listeners, but also cancel any pending/debounced callbacks.

For the most part this proved fruitful. However there was the odd/occasional callback still sneaking through and causing an exception. This was because once of the debounced callbacks was also doing some asynchronous work and then calling the `SegmentationService`. So in this case, the `reset`/`unsubscribe` was called during the asynchronous work. A guard/check was added after the asynchronous work in case the callback was unsubscribed during the asynchronous call.

### Testing

See #5087 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 20.9.0
- [x] Browser: Chrome 137.0.7151.56
  

### DIFF
--- a/extensions/cornerstone/src/index.tsx
+++ b/extensions/cornerstone/src/index.tsx
@@ -132,6 +132,10 @@ const cornerstoneExtension: Types.Extensions.Extension = {
   },
   getPanelModule,
   onModeExit: ({ servicesManager }: withAppTypes): void => {
+    unsubscriptions.forEach(unsubscribe => unsubscribe());
+    // Clear the unsubscriptions
+    unsubscriptions.length = 0;
+
     const { cineService, segmentationService } = servicesManager.services;
     // Empty out the image load and retrieval pools to prevent memory leaks
     // on the mode exits
@@ -150,10 +154,6 @@ const cornerstoneExtension: Types.Extensions.Extension = {
     useToggleOneUpViewportGridStore.getState().clearToggleOneUpViewportGridStore();
     useSegmentationPresentationStore.getState().clearSegmentationPresentationStore();
     segmentationService.removeAllSegmentations();
-
-    unsubscriptions.forEach(unsubscribe => unsubscribe());
-    // Clear the unsubscriptions
-    unsubscriptions.length = 0;
   },
 
   /**
--- a/extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
+++ b/extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
@@ -255,7 +255,7 @@ class SegmentationService extends PubSubService {
       this._onSegmentationAddedFromSource
     );
 
-    this.listeners = {};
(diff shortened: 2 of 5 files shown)


extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -255,7 +255,7 @@ class SegmentationService extends PubSubService {
       this._onSegmentationAddedFromSource
     );
 
-    this.listeners = {};
+    this.reset();
```
jbocce, line 258: debounced listeners need some special TLC.

extensions/cornerstone/src/utils/segmentationHandlers.ts
```diff
@@ -42,7 +46,7 @@ export function setupSegmentationDataModifiedHandler({
         readableText,
       });
 
-      if (updatedSegmentation) {
+      if (!isUnsubscribed && updatedSegmentation) {
```
jbocce, line 49: The unsubscribe might have occurred during the await, so don't do stuff if that is the case.

extensions/cornerstone/src/index.tsx
```diff
@@ -132,6 +132,10 @@ const cornerstoneExtension: Types.Extensions.Extension = {
   },
   getPanelModule,
   onModeExit: ({ servicesManager }: withAppTypes): void => {
+    unsubscriptions.forEach(unsubscribe => unsubscribe());
```
jbocce, line 135: Doing the unsubscribe first is a good idea eventhough for this case it did nothing to help the problem.

platform/core/src/services/_shared/pubSubServiceInterface.ts
```diff
@@ -134,6 +137,13 @@ export class PubSubService {
   reset() {
     this.unsubscriptions.forEach(unsub => unsub());
```
jbocce, line 138: These `unsubscriptions` are kinda useless. Should we remove them?

PR #5101, opened 2025-06-05
fix(markerlabelmap): accept preview/reject preview not working when the marker annotation tools are active

IbrahimCSAE:
### Context

The preview accept/reject functionality wasn't working properly for MarkerLabelmap tools. When MarkerInclude or MarkerExclude was the active tool, the corresponding MarkerLabelmap tool would be in passive mode and get
filtered out from preview actions.

Added logic to include passive MarkerLabelmap tools when their associated marker tools (MarkerInclude/MarkerExclude) are active in the same tool group. This ensures preview actions reach all relevant tools regardless
of their current mode state.

### DIFF
--- a/extensions/cornerstone/src/commandsModule.ts
+++ b/extensions/cornerstone/src/commandsModule.ts
@@ -65,6 +65,33 @@ const getLabelmapTools = ({ toolGroupService }) => {
   return labelmapTools;
 };
 
+const MARKER_TOOLS = ['MarkerInclude', 'MarkerExclude'];
+
+const getActiveLabelmapTools = ({ toolGroupService }) => {
+  const labelmapTools = getLabelmapTools({ toolGroupService });
+  const activeTools = labelmapTools.filter(
+    tool => tool.mode === 'Active' || tool.mode === 'Enabled'
+  );
+
+  // Include passive MarkerLabelmap if corresponding marker tools are active
+  labelmapTools
+    .filter(tool => tool.getToolName?.() === 'MarkerLabelmap' && tool.mode === 'Passive')
+    .forEach(passiveTool => {
+      const toolGroup = cornerstoneTools.ToolGroupManager.getToolGroup(passiveTool.toolGroupId);
+      const hasActiveMarkerTool = Object.keys(toolGroup.getToolInstances()).some(
+        toolName =>
+          MARKER_TOOLS.includes(toolName) &&
+          toolGroup.getToolInstances()[toolName].mode === 'Active'
+      );
+
+      if (hasActiveMarkerTool && !activeTools.includes(passiveTool)) {
+        activeTools.push(passiveTool);
+      }
+    });
+
+  return activeTools;
+};
+
 const segmentAI = new ONNXSegmentationController({
   autoSegmentMode: true,
   models: {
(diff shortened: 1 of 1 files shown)


IbrahimCSAE:
This is not the cleanest fix in the world, but we can remove once there's a better system for this in place

wayfarer3130:
@IbrahimCSAE - could you add a description of how to test this?  I think I see it, but would like to know exactly what you are seeing.

IbrahimCSAE:
> @IbrahimCSAE - could you add a description of how to test this? I think I see it, but would like to know exactly what you are seeing.

- Visit: https://viewer-dev.ohif.org/segmentation?StudyInstanceUIDs=2.16.840.1.114362.1.11972228.22789312658.616067305.306.2
- Add a new segmentation
- Select marker labelmap tool
- Add a point
- Seg will appear
- Press enter to accept, nothing happens, it stays forever in preview mode.
- Press esc to reject, nothing happens, preview still visible


On the other hand, it works fine here: https://www.cornerstonejs.org/live-examples/samclientside

If you press enter it will accept the preview as expected, the reason for this is that its called more directly here.

@wayfarer3130

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -1709,11 +1736,8 @@ function commandsModule({
...
     },
     _handlePreviewAction: action => {
-      const labelmapTools = getLabelmapTools({ toolGroupService });
       const { viewport } = _getActiveViewportEnabledElement();
-      const activeTools = labelmapTools.filter(
-        tool => tool.mode === 'Active' || tool.mode === 'Enabled'
-      );
+      const activeTools = getActiveLabelmapTools({ toolGroupService });
 
       activeTools.forEach(tool => {
```
IbrahimCSAE: The bug happens here, if `MarkerLabelMap tool instance` is not pulled into the `activeTools` array, it's `acceptPreview` is never called

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -65,6 +65,33 @@ const getLabelmapTools = ({ toolGroupService }) => {
   return labelmapTools;
 };
 
+const MARKER_TOOLS = ['MarkerInclude', 'MarkerExclude'];
+
+const getActiveLabelmapTools = ({ toolGroupService }) => {
```
wayfarer3130: Can you call this
getPreviewTools
and then instead of having named tools, get the actual tool instance, and filter for .accept containing tools which are not disabled?  That way we can add more preview types in the future and have them accept/reject it.  
Alternatively, make the accept/reject functionality a general labelmap static instance method which can get called on any labelmap tool.

wayfarer3130:
@IbrahimCSAE  - are you able to update the PR as requested?

IbrahimCSAE:
> @IbrahimCSAE - are you able to update the PR as requested?

Yes I haven’t had free time yet go get back to this. I’ll try today.

extensions/cornerstone/src/commandsModule.ts
IbrahimCSAE: Done

IbrahimCSAE:
> @IbrahimCSAE - are you able to update the PR as requested?

It's done, should be good now.

wayfarer3130 (approved):

PR #5135, opened 2025-06-16
fix(ViewportOrientationMarkers): round rotation to fix floating point issue

pwespi:
### Context

After rotate right 3 times and flip horizontally the orientation markers are wrong. This is because the rotation is `90.00000000000001` instead of `90`.

### Changes & Results

Round rotation to fix floating point issue.

Before:
before

After:
after

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone/src/Viewport/Overlays/ViewportOrientationMarkers.tsx
```diff
@@ -198,6 +198,9 @@ function _getOrientationMarkers(rowCosines, columnCosines, rotation, flipVertica
   const oppositeRowString = invertOrientationStringLPS(rowString);
   const oppositeColumnString = invertOrientationStringLPS(columnString);
 
+  // After rotate right 3 times and flip horizontally -> rotation is 90.00000000000001
+  rotation = Math.round(rotation);
```
sedghi: hmmm, how much round we want? @wayfarer3130

wayfarer3130: I'm assuming it is in degrees, otherwise Math.round doesn't make sense.  I think degrees is sufficient - having to high a precision here is undesirable as it doesn't compare very nicely.  However, I would also try using the rotation tool and make sure it is still smooth/works.

pwespi: Yes, it's in degrees here. The problem is the equality checks further below (e.g. `rotation === 90`) don't work correctly without rounding.

Another solution could be to replace the equality checks with something like `Math.abs(rotation - 90) < epsilon`. If you prefer I can also do this.

pwespi: @wayfarer3130 Let me know in case you prefer another implementation.

wayfarer3130: It needs to be rounded so that a single pixel at the top of hte screen is moved by 1 pixel lcoation so that the control works.  I think that should be 1 part in 4000, so call it rounding to 4 decimal places.

wayfarer3130: I'm concerned about the free rotate tool working

pwespi: Changed code to round to 4 decimal places.

jbocce (approved):
Approved.

@wayfarer3130 please have another look. Thanks.

PR #5139, opened 2025-06-17
fix(segmentation): Changes to permit non hydrated/loaded segmentations to be viewable in hanging protocols

jbocce:
### Changes & Results
Adding segmentation via data portal in MPR mode.

Use volumeLoader.createAndCacheVolume to create the volume for a SEG display set. This produces a StreamingImageVolume which is essential for switching to hanging protocols like MPR. RTSTRUCT display sets are now marked as reconstructable.

For when multiple display sets occupy the active viewport, cache and restore each appropriately.

Added tests:
- launch MPR with unhydrated segmentation.
- launch MPR and then add (but not load) a segmentation from the overlay menu
- add (but not load) a segmentation from the overlay menu and then launch MPR

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
```diff
@@ -31,7 +31,7 @@ function _getDisplaySetsFromSeries(
   const displaySet = {
     Modality: 'RTSTRUCT',
     loading: false,
-    isReconstructable: false, // by default for now since it is a volumetric SEG currently
+    isReconstructable: true, // by default for now since it is a volumetric SEG currently
```
jbocce: Allow RTSTRUCT to be loaded into volumetric hanging protocols. There looks to be some problems with RTSTRUCT and volumetric hanging protocols, but we can address those in another PR.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -772,7 +773,11 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
 
       const displaySet = displaySetService.getDisplaySetByUID(displaySetInstanceUID);
       if (!volume && displaySet.images) {
-        volume = csToolsUtils.getOrCreateImageVolume(displaySet.images.map(image => image.imageId));
+        const volumeLoaderSchema = displaySet.volumeLoaderSchema ?? VOLUME_LOADER_SCHEME;
+        const volumeId = `${volumeLoaderSchema}:${displaySet.displaySetInstanceUID}`;
+        volume = await volumeLoader.createAndCacheVolume(volumeId, {
```
jbocce: Create the volume this way for a SEG or RTSTRUCT because it appears the `nthLoader` (used by "most" HPs) is expecting a `StreamingImageVolume` instead of an `ImageVolume` as it was before.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -871,29 +870,9 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
-        throw new Error('Background display set not found');
-      }
-
-      await viewport.setVolumes([
-        { volumeId: `${VOLUME_LOADER_SCHEME}:${backgroundDisplaySet[0].displaySetInstanceUID}` },
-      ]);
-    } else {
-      await viewport.setVolumes(filteredVolumeInputArray);
-    }
+    await viewport.setVolumes(volumeInputArray);
```
jbocce, line 895: No longer is there need to set the overlay display set volumes separately.

extensions/default/src/utils/reuseCachedLayouts.ts
```diff
@@ -58,8 +58,10 @@ const reuseCachedLayout = (state, hangingProtocolService: HangingProtocolService
       if (!displaySetUID) {
         continue;
       }
-      if (viewportId === activeViewportId && i === 0) {
-        setDisplaySetSelector(`${activeStudyUID}:activeDisplaySet:0`, displaySetUID);
+      if (viewportId === activeViewportId) {
+        // For the active viewport, overload the `matchedDisplaySetsIndex` position in the cache key
```
jbocce: @wayfarer3130, please have a look at this and let me know what you think. This is one way to cache the various overlay display sets for the active viewport. The other thing I can think of is to add overlays to hanging protocols which does not make much sense because in essence multiple display sets can already be defined for a viewport in a hanging protocol. The overlays appear to be a "view only"/"in app only" feature. One other thing could be to expand the cache key here to include the overlay index separately from the `matchedDisplaySetsIndex`.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -1374,6 +1374,25 @@ export default class HangingProtocolService extends PubSubService {
         };
 
         displaySetsInfo.push(displaySetInfo);
+
+        if (id === 'activeDisplaySet') {
```
jbocce: @wayfarer3130, here is where the various overlay display sets are added for the "active display set" of a hanging protocol.

tests/utils/screenShotPaths.ts
```diff
@@ -89,6 +89,18 @@ const screenShotPaths = {
...
+    segNoHydrationPostMPR: 'segNoHydrationPostMpr.png',
+  },
+  segDataOverlayNoHydrationThenMPR: {
+    segDataOverlayNoHydrationPreMPR: 'segDataOverlayNoHydrationPreMpr.png',
+    segDataOverlayNoHydrationPostMPR: 'segDataOverlayNoHydrationPostMpr.png',
+  },
+  mprThenSEGOverlayNoHydration: {
+    mprPreSEGOverlayNoHydration: 'mprPreSEGOverlayNoHydration.png',
+    mprPostSEGOverlayNoHydration: 'mprPostSEGOverlayNoHydration.png',
+  },
```
sedghi, line 103: I thought we already have these covered no? what are these then?

segHydrationThenMPR: {
    segPostHydration: 'segPostHydration.png',
    segPostHydrationMPRAxialPrimary: 'segPostHydrationMPRAxialPrimary.png',
  },
  segHydrationFromMPR: {
    mprBeforeSEG: 'mprBeforeSEG.png',
    mprAfterSEG: 'mprAfterSEG.png',
    mprAfterSegHydrated: 'mprAfterSegHydrated.png',
    mprAfterSegHydratedAfterLayoutChange: 'mprAfterSegHydratedAfterLayoutChange.png',
  },

  
  -----

  Can we get rtstruct tests please

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -772,7 +773,11 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
 
       const displaySet = displaySetService.getDisplaySetByUID(displaySetInstanceUID);
       if (!volume && displaySet.images) {
-        volume = csToolsUtils.getOrCreateImageVolume(displaySet.images.map(image => image.imageId));
+        const volumeLoaderSchema = displaySet.volumeLoaderSchema ?? VOLUME_LOADER_SCHEME;
+        const volumeId = `${volumeLoaderSchema}:${displaySet.displaySetInstanceUID}`;
```
sedghi: we don't need volume loader scheme anymore, you can try removing the left part of the `:` i think it should work

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -772,7 +773,11 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
+        volume = await volumeLoader.createAndCacheVolume(volumeId, {
+          imageIds: displaySet.images.map(image => image.imageId),
+        });
       }
```
sedghi: I'm surprised we don't need the old implementation, since this is the old implementation 

```js
function getOrCreateImageVolume(
  referencedImageIds: string[]
): Types.IImageVolume | undefined {
  if (!referencedImageIds || referencedImageIds.length <= 1) {
    return;
  }

  const isValidVolume = csUtils.isValidVolume(referencedImageIds);

  if (!isValidVolume) {
    return;
  }

  const volumeId = cache.generateVolumeId(referencedImageIds);

  // Check if volume already exists in cache
  let imageVolume = cache.getVolume(volumeId);

  if (imageVolume) {
    return imageVolume;
  }

  // Create and cache the volume if it doesn't exist
  imageVolume = volumeLoader.createAndCacheVolumeFromImagesSync(
    volumeId,
    referencedImageIds
  );

  return imageVolume;
}

```

do you know why it doesn't work? seems really similar

tests/utils/screenShotPaths.ts
jbocce: The existing tests and screen shots are for hydrated (i.e. LOADed) SEGs. The problems I was fixing deal with non-hydrated SEGs. Note the difference in name of the tests and screen shots (i.e. NoHydration).

With regards to RTSTRUCT, the very next issue I am working on deals with RTSTRUCT. Can we wait until then for tests or should I just continue with this PR dealing with the other scenarios regarding RTSTRUCT and SEG? There appear to many scenarios that never worked or are broken.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
jbocce: They are indeed different. The call to `createAndCacheVolumeFromImagesSync` returns an `ImageVolume` but we need a `StreamingImageVolume` for the `nthLoader`. `createAndCacheVolume` returns a `StreamingImageVolume`.

jbocce:
> Thanks
> 
> Here are the things i found
> 
> 1. Go here http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.1706.8374.643249677828306008300337414785
> 
> Do the data overlay, see does not work
> 
> see https://share.cleanshot.com/MGnJgpFZ

This issue is related to the next issue I will be looking at regarding RTSTRUCT. I think the problem is that we get an exception when we try to overlay a SEG or RTSTRUCT on a series that is NOT referenced by the SEG or RTSTRUCT series. Can we leave this for now or should continue fixing all these scenarios in this PR?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
jbocce: You are correct.

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
wayfarer3130: This should only be true when the spacing is regular, same as normal volumes.  The problem with irregular volumes is that vtkjs will treat it as regular and may get the wrong positioning on images.

extensions/cornerstone/src/components/SelectItemWithModality.tsx
```diff
@@ -4,11 +4,16 @@ import React from 'react';
 const SelectItemWithModality = ({
   displaySet,
   showModality = true,
+  dataCY = '',
```
wayfarer3130: Can you have a default name other than '' for identification?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
wayfarer3130: Likely we should reduce the use of the nthLoader in favour of the interleaved loading definition in CS3D, as that one deals with more cases for multiple views.

extensions/default/src/utils/reuseCachedLayouts.ts
wayfarer3130: What we really want is a way to match a list of display sets at a given position - that is, the matched display sets could return more than a single match for a single position.  This is relevant for more than just the active viewport id I suspect, as we should be able to get secondary matches from other match keys.

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
jbocce: @wayfarer3130 , does your comment also refer to SEG display sets as well? Should I revert this?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
sedghi: Okay, I think the Nth loader shouldn't need `StreamingImageVolume`. It should work fine if it's just an `ImageVolume`. If the images are already cached, there's really no need for an async call to have `StreamingImageVolume`.

Could you fix this, so the Nth loader can also work with `createAndCacheVolumeFromImagesSync`? Thanks!

tests/utils/screenShotPaths.ts
sedghi: Good note

sedghi (approved):

wayfarer3130:
I'm getting exceptions in tmtv mode when I double click various series in the initial view and drag/drop different series in, and then return to the original view.  Sometimes it works to change views and sometimes it doesnt

jbocce:
> I'm getting exceptions in tmtv mode when I double click various series in the initial view and drag/drop different series in, and then return to the original view. Sometimes it works to change views and sometimes it doesnt

Thanks @wayfarer3130. But could you please give me some exact steps? It appears to work the same as in https://viewer-dev.ohif.org/

jbocce:
> > I'm getting exceptions in tmtv mode when I double click various series in the initial view and drag/drop different series in, and then return to the original view. Sometimes it works to change views and sometimes it doesnt
> 
> Thanks @wayfarer3130. But could you please give me some exact steps? It appears to work the same as in https://viewer-dev.ohif.org/

Further on this... I have spent a bit of time double clicking series and dragging and dropping series in the TMTV HP. I discovered no specific exceptions that were bugs. I had https://viewer-dev.ohif.org/ side-by-side with this PR branch. Both behaved similarly.

The exceptions I did manage to receive were the following and were ALSO present in master (i.e. viewer-dev)

So @sedghi and @wayfarer3130  unless there are more specific steps and/or exceptions that need to be handled that are CAUSED by this PR, I will not look further.
```
ViewportGrid.tsx:118 Error: The displaySetInstanceUID 0e19c131-2724-3a1e-e8b9-487df67ed738 does not satisfy the required seriesMatching criteria for the protocol
    at HangingProtocolService._validateRequiredSelectors (HangingProtocolService.ts:1471:1)
    at HangingProtocolService.ts:1438:1
    at HangingProtocolService._validateViewportSpecificMatch (HangingProtocolService.ts:1427:1)
    at HangingProtocolService.getViewportsRequireUpdate (HangingProtocolService.ts:713:1)
    at ViewportGrid.tsx:112:1
    at ViewportGrid.tsx:262:1
```

```
Route tmtv Error Boundary Error: Volumes being added to viewport fusionCoronal do not share the same FrameOfReferenceUID. This is not yet supported
    at VolumeViewport._isValidVolumeInputArray (BaseVolumeViewport.js:818:1)
    at VolumeViewport.setVolumes (BaseVolumeViewport.js:714:1)
    at VolumeViewport.setVolumes (VolumeViewport.js:101:1)
    at CornerstoneViewportService.setVolumesForViewport (CornerstoneViewportService.ts:895:1)
    at CornerstoneViewportService._setVolumeViewport (CornerstoneViewportService.ts:825:1)
    at CornerstoneViewportService._setDisplaySets (CornerstoneViewportService.ts:1030:1)
    at CornerstoneViewportService.setViewportData (CornerstoneViewportService.ts:468:1)
    at loadViewportData (OHIFCornerstoneViewport.tsx:296:1) null Route tmtv
```

```
ViewportGrid.tsx:118 Error: Cannot update viewport with multiple displaySets yet
    at HangingProtocolService.getViewportsRequireUpdate (HangingProtocolService.ts:685:1)
    at ViewportGrid.tsx:112:1
    at ViewportGrid.tsx:262:1
```

sedghi:
Let's merge this and then if Bill got some time can report what he is seeing.

PR #5148, opened 2025-06-19
fix(UnitTests): Fixing "Cannot find module @cornerstonejs/core" error while running unit tests

ViniciusResende:
### Context

The TypeScript tests were not being run, they were added to the test execution now. Also, there were some tests failing with an error `Cannot find module @cornerstonejs/core`, this is now fixed. To perform this, those were the commits:

- 1f1940d2493d494d2a7a3a46fbc22c70b907e06a - feat: add transform-for-of babel plugin to test env
  - Due to the need of loading cornerstonejs in jest unit tests, since the files are ESM, there is the need to handle with the transpiling of the newest features that are used on those files, including the for of on class this.

- d9802f904b7f3ad75f3184fcb6ddf0ffef4511f3 - fix: added right module mapper to jest config for cornerstone
  - Since the extensions/default folder has files that import modules from @cornerstonejs, it was needed to configure jest to know where to search for those modules. (First version of it was wrong, fixed in 752543a98aea29ca035817eb019738a46c53c597)

- fe3f955eaf86b643ed641b272c3c63fe37b62db2 - feat: add ts tests and babel transpiling to jest config
  - Since now we have TS test files, they needed to be added to the jest.config in order to be ran. Also, some tests import ESM modules, thus it is needed to have the transpiling of those files in order to the tests to properly run.

- e72b32cfbc1403e3a53d553dc333150739b2bca9 - fix: changed expectancy of getBulkDataValue to be not called
  - For this test, specially considering TS pre-processing now that it was migrated to TS, is that the second part of the || condition will not be executed if the first one is executed. Refer to:https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR#short-circuit_evaluation

- 6071c4973bc4bef6e4a1232a5ae03deabff8b193 - fix: change areAllImageDimensionsEqual util import
  - This is needed so the Jest unit test can properly handle the importing of the module.
  
- f5a85ba09fbaefaac9b7864e22a5bd2e11bbe7c3 - feat:  exposing toNumber util function on ohif core export
  - This was needed so areAllImageDimensionsEqual could import this utility without needing to directly search it inside the folder. Preventing Jest to properly map it.
  
- 4bf966d10829e5e4d96b51648058ea0d2faa92b9 - fix: @ohif/extension-default was not being included in lerna
  - During the test:unit:ci execution, the @ohif/extension-default was not being included. So the fixes on this commit were necessary in order to it to run and have it's tests passing.

### Changes & Results

Now all the JS and TS tests are executed

Before:
image

After:
image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Ubuntu 22.04.4 LTS
- [x] Node version: v20.9.0 & v22.12.0
- [x] Browser: Not applicable


sedghi:
hmmm, unit tests failing still :(

ViniciusResende:
This is still work in progress, changed it to draft. End next week I'll have availability to finalize it. You can close it in the meantime.

I do think we need it, since we're not executing all tests that are in OHIF currently. It's a sort of silent skip of them, since lerna isn't able to find and execute them.

Also, every test that has a module that tries to import any @cornerstonejs/* module will fail.

This is exactly what is being fixed in this PR, locally the unit tests are running, but not in CircleCI. This is why I put this in draft, to have the time to debug why it's failing on CircleCI.

sedghi:
rebase

ViniciusResende:
@sedghi Just rebased.

Have re-created the `bun.lock` to minimize changes.

Issue #4600, opened 2024-12-12
[Feature Request] Display the label for the segment at the mouse location


fedorov:
### What feature or change would you like to see made?

It would be helpful to be able to show the name of the segment at the pixel location pointed by the mouse in a tooltip, or in some "probe" area of the interface.

### Why should we prioritize this feature?

Once the number of segments reaches couple of dozens (and it can easily be ~100 with the modern AI segmentation tools), it is quite difficult to know what is the label for a given segmentation in the viewer.

Especially for someone who does not know anatomy well - like me (and at least some AI segmentation tools developers, I dare say)! :-D

pedrokohler:
@fedorov I'm just waiting for a green light to start working on this one.

fedorov:
I deployed using this recipe we have been using in IDC https://github.com/ImagingDataCommons/idc-viewers-sandbox-gha-testing/blob/main/.github/workflows/deploy-ohif-v3-upstream-with-add-ons.yml, and I do not see this feature. Do I need to anything special to be able to use it?

pedrokohler:
@fedorov you have to activate it on the segmentation panel here:

<img width="2914" height="1768" alt="Image" src="https://github.com/user-attachments/assets/9df1db9d-a299-47a9-ba25-3a850ed96d9a" />

(this is a screenshot from the version deployed by you)

fedorov:
Ok, I found it in my deploy. So now I have the following questions:

1. Why is it in the Segmentation mode? This is a feature most valuable for the basic viewer, when we need static review of **existing** segmentations.
2. I still cannot make it to work, even though it is enabled, see demo here: https://app.screencast.com/QqWys9ShkGRxK
3. Is this setting exposed via the config file, so it can be enabled by default on deploy?

pedrokohler:
> 1. Why is it in the Segmentation mode? This is a feature most valuable for the basic viewer, when we need static review of existing segmentations.

I've also added it to the basic viewer (longitudinal mode). It should be available on the same menu. Please let me know if it is not.

> 2. I still cannot make it to work, even though it is enabled, see demo here: https://app.screencast.com/QqWys9ShkGRxK

That is weird. Did you get any errors in the console?

> 3. Is this setting exposed via the config file, so it can be enabled by default on deploy?

No. I'll add that for you.

pedrokohler:
@fedorov can you share the url of the study you were looking at in the video?

fedorov:
[...]

PR #5164, opened 2025-06-27
feat: add segment label tool

pedrokohler:
### Context

Fixes https://github.com/OHIF/Viewers/issues/4600

Adding the new cornerstone tool, segmentLabel.

### Changes & Results

Now you can toggle an option to see the name of the segment on mouse hover:

image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


sedghi:
Can we make them same color as the image?

pedrokohler:
> Can we make them same color as the image?

I'm not sure what you mean. The text that shows is the same color as any of the other annotation labels. 

In case you're talking about the red box, I just added it around the text to highlight the change in the screenshot for the PR, it's not something that will be there when the user interacts with it.

Also, if we were to change the colors, wouldn't that have to be done on the cornerstone side?

sedghi:
So like this, if you look the label that appears on the right side of the mouse pointer has the same color as the segment
This is Siemens, i think we should not treat it as the same as annotation colors. This is a segmentatation tool

CleanShot 2025-06-27 at 08 47 31@2x

CleanShot 2025-06-27 at 08 50 38@2x

I don't think we need the linked box as well a label on top of the mouse is enough

pedrokohler:
Alright. Sure, I can do that. 

Shouldn't this change be made on the cornerstone side though?

sedghi:
Yes it is in cornerstone.

immahi79:
@sedghi  @pedrokohler  when this feature be available for release ?  
Can this be extended to always show the label, instead of only on mouse hover? The goal is to display all marked lesions with their IDs so that comparisons with baseline and follow-up Lesion tracking with IDs are easily traceable and identifiable directly on the image, without requiring additional mouse clicks or navigation

fedorov:
> Can this be extended to always show the label, instead of only on mouse hover?

Not always, but having a toggle to enable showing all labels would be fine with me!

pedrokohler:
@sedghi I've opened the PR for the change you requested: https://github.com/cornerstonejs/cornerstone3D/pull/2169

> > Can this be extended to always show the label, instead of only on mouse hover?
> 
> Not always, but having a toggle to enable showing all labels would be fine with me!

@fedorov I think this could be done as another effort, since it would require a different approach from the one right now, which uses a tool that relies on mouse hovering.

It could be yet another tool, but it would work very differently internally.

And I suspect one challenge would be deciding where to show the label for each segment since there's no mouse pointer as a reference and the same segment could span to different slices.

Let me know if you want me to pursue that.

fedorov:
I agree - let's keep that separate! This feature is not of priority for our project, so no need to pursue it.

sedghi:
now that we merged this can we set the timeout to 1 in OHIF

pedrokohler:
@sedghi done.

sedghi (approved):

sedghi:
Test not passing, but not your fault, this one https://github.com/cornerstonejs/cornerstone3D/pull/2163 has caused some prefetch issues, that is being fixed, gonna say request change so that no one merges yet

sedghi:
can you update again in 10 mins that it publishes to npm

pedrokohler:
@sedghi done

sedghi (approved):

Issue #5182, opened 2025-07-08
[Bug] Adding multiple unhydrated segmentation display set overlays does not work


jbocce:
### Describe the Bug

The viewport display set overlay menu allows for multiple segmentation overlays to be added. However, adding more than one unhydrated segmentation does not work. The first unhydrated segmentation can be added and displayed, but adding subsequent segmentation display sets does NOT actually display those added. Note that if the previous overlays are all hydrated, then the next subsequently added display set is displayed.

### Steps to Reproduce

1. Load https://viewer-dev.ohif.org/segmentation?StudyInstanceUIDs=1.3.6.1.4.1.32722.99.99.239341353911714368772597187099978969331
2. Add the `Segmentation` overlay.
3. It displays.
4. Add the `2D-TTA_NNU-NET_SEGMENTATION` overlay.

### The current behavior

The `2D-TTA_NNU-NET_SEGMENTATION` does NOT display.

### The expected behavior

Both the `Segmentation` and the `2D-TTA_NNU-NET_SEGMENTATION` display.

### OS

Windows 11

### Node version

20.9.0

### Browser

Chrome 138.0.7204.97

PR #5189, opened 2025-07-09
fix(segmentation overlays): Allow for the addition of multiple segmentation overlays.

jbocce:
### Context
See #5182 

### Changes & Results
Changed the code to detect and add multiple segmentation overlays. Only one segmentation overlay was being added previously.

Added a test.

### Testing
See #5182 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 20.9.0
- [x] Browser: Chrome 138.0.7204.101
  


extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -651,8 +651,9 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
     });
 
     let imageIdsToSet = imageIds;
-    const overlayProcessingResult = this._processExtraDisplaySetsForViewport(viewport);
-    imageIdsToSet = overlayProcessingResult?.imageIds ?? imageIdsToSet;
+    const overlayProcessingResults = this._processExtraDisplaySetsForViewport(viewport);
+    // Since we have a stack viewport, all of the imageIds should be the same for each overlay, no?
```
jbocce: @sedghi Is this assumption about the imageIds correct for a stack viewport? If not, what should I do instead?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -874,28 +879,35 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
-        displaySet =>
-          !displaySet.isOverlayDisplaySet &&
-          displaySet.images.some(image => image.imageId === sampleImageId)
-      );
-
-      if (backgroundDisplaySet.length !== 1) {
-        throw new Error('Background display set not found');
-      }
+    const overlayProcessingResults = this._processExtraDisplaySetsForViewport(viewport) || [];
+    if (!filteredVolumeInputArray.length && overlayProcessingResults?.length) {
```
jbocce, line 880: Does this make sense for volume viewports?

playwright.config.ts
```diff
@@ -21,7 +21,7 @@ export default defineConfig({
   use: {
     baseURL: 'http://localhost:3335',
     trace: 'on-first-retry',
-    video: 'on',
+    video: 'on-first-retry',
```
jbocce, line 24: Less noise. Instead of saving a video for each test (regardless of pass or fail), only save videos when a test is retried and furthermore only save the video once.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -934,30 +946,31 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
     const displaySetInstanceUIDs = this.viewportsDisplaySets.get(viewport.id);
 
     // Find overlay display sets (e.g. SEG, RTSTRUCT)
-    const overlayDisplaySet = displaySetInstanceUIDs
+    const overlayDisplaySets = displaySetInstanceUIDs
```
jbocce, line 946: Prior to this we were only ever processing ONE overlay display set. But now we do them ALL.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
sedghi: each segmentation (labelmap) in stack settings will have its own sets of imagesIds. But for instance if we are on slice 150 in CT , only 1 or more (overlapping) of those imageIds can get rendererd, but that is done automatically by cornerstone. So here you should just add the segmentation representation with those imageIds and add them

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
sedghi: sure

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
sedghi: good

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
jbocce: @sedghi  So are you suggesting that we don't even need this line? I don't think I completely understand.

Issue #5141, opened 2025-06-17
[Bug] Multimonitor support (currently working?)


Joeycho:
### Describe the Bug

Hello OHIF Team,

Thank you for supporting more features in 3.10. I'm currently testing Multimonitor feature, and I cannot see any change from my side. Does it work currently? Either `viewer-dev.ohif.org` or `viewer.ohif.org` website?

I checked the feature like the below. Attach more params in URL: `&multimonitor=2&screenNumber=0&launchAll`
OR `&multimonitor=split`

https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.2155604110.4180.1021041295.21&multimonitor=2&screenNumber=0&launchAll

https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.2155604110.4180.1021041295.21&multimonitor=split

I also includes multimonitor via customizationservice locally, like the below:
```js
  customizationService: [
    '@ohif/extension-default.customizationModule.multimonitor',]
```

But I cannot see the change. What is the current status?

### Steps to Reproduce

Access the following websites:

https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.2155604110.4180.1021041295.21&multimonitor=2&screenNumber=0&launchAll

https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.2155604110.4180.1021041295.21&multimonitor=split

https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.2155604110.4180.1021041295.21&multimonitor=2&screenNumber=0&launchAll

https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.2155604110.4180.1021041295.21&multimonitor=split

### The current behavior

No change

### The expected behavior

Show OHIF through multi-monitors

### OS

macOS 15.5

### Node version

18

### Browser

Chrome  137.0.7151.120

Joeycho:
@sedghi I think you reproduced this bug, right?

sedghi:
yes it is a bug

jbocce:
So according to #4178, there are two configurations...

1. Split Screen (multimonitor=split) 
Splits the primary monitor into two windows.

2. Multi-Monitor (multimonitor=2)
[...]

PR #5208, opened 2025-07-15
fix(multimonitor): Study browser more menu fixes

jbocce:
### Context
See https://github.com/OHIF/Viewers/issues/5141#issuecomment-3071617294

### Changes & Results

Only add config, global customization references once at startup.

Override the display:table for the ScrollArea component used in the study browser.

### Testing

https://github.com/OHIF/Viewers/issues/5141#issuecomment-3071617294 should be addressed

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 20.9.0
- [x] Browser: Chrome 138.0.7204.101
  


platform/ui-next/src/components/StudyBrowser/studyBrowserScrollArea.css
```diff
@@ -0,0 +1,3 @@
+div.StudyBrowserScrollArea > div[data-radix-scroll-area-viewport] > div {
+  display: block !important; /* Scroll area is using a display:table which really messes things up */
```
jbocce: The ScrollArea component has a div with `display:table`. I think I saw that once before among dinosaur bones. 🤣

platform/ui-next/src/components/StudyBrowser/StudyBrowser.tsx
```diff
@@ -60,7 +61,7 @@ const StudyBrowser = ({
   };
 
   return (
-    <ScrollArea>
+    <ScrollArea className="StudyBrowserScrollArea">
```
sedghi: is there any way we can avoid css ? and just write tailwind?

platform/core/src/services/CustomizationService/CustomizationService.ts
```diff
@@ -131,13 +131,16 @@ export default class CustomizationService extends PubSubService {
       }
     });
 
-    this.addReferences(this.configuration);
+    if (addReferences) {
```
sedghi: why we don't want to add references on init?

platform/ui-next/src/components/StudyBrowser/StudyBrowser.tsx
jbocce: I wish there were. But it is a div that `ScrollArea` adds and does not appear to inject any (tailwind) class names into it.

platform/core/src/services/CustomizationService/CustomizationService.ts
jbocce: We only want to add references on the very first init (i.e. from App). If we do it for EVERY `modeEnter` it will add to the global customizations. In particular those global customizations that have `$push` will add duplicate entries each time the mode is entered. See https://github.com/OHIF/Viewers/issues/5141#issuecomment-3071617294 for the problematic result.

sedghi: please add a lot of comment on this 
FYI @wayfarer3130

platform/core/src/services/CustomizationService/CustomizationService.ts
```diff
@@ -131,13 +131,18 @@ export default class CustomizationService extends PubSubService {
       }
     });
 
-    this.addReferences(this.configuration);
+    // This allows us to control when configuration references are added.
```
jbocce: @wayfarer3130 could you please comment on this? Thanks. See also line 136 below.

wayfarer3130: I'd really rather not have the init method control whether the references get added or not.  
The right solution is only to allow the update to occur for a given customization item once.  
What about defining a property on this.configuration indicating that references have been added, so the code becomes
if (!this.configuration?._hasBeenAdded ) {
   this.addReferences(this.configuration);
   Object.defineProperty(this.configuration, "_hasBeenAdded", { value: true });
}

It might be desirable to do that on global/default customizations as well.

jbocce: @wayfarer3130 , that is a good suggestion. I don't think we have to do that for global/default customizations because they are cleared at the beginning of this `init` method. Does that make sense?

platform/core/src/services/CustomizationService/CustomizationService.ts
```diff
@@ -109,7 +109,7 @@ export default class CustomizationService extends PubSubService {
     this.commandsManager = commandsManager;
   }
 
-  public init(extensionManager: ExtensionManager): void {
+  public init(extensionManager: ExtensionManager, addConfigReferences = true): void {
```
wayfarer3130: Can you remove your addConfigReferences?

jbocce:
> Just remove the unneeded parameter.

Done. May I merge after that?

PR #5223, opened 2025-07-18
fix: issue introduced from PR 5022

arul-trenser:
### Context

- Fix for the issue mentioned by @IbrahimCSAE in #5022 

> FYI. This has broken the sort on some series, they became completely out of order. I can't share the data but the only information I can share is that it was a sagittal series with an axial cover slice.
 
### Changes & Results
The issue is fixed by validating whether the `areAllImageOrientationsEqual` before sorting slices with spatial position

### Testing

1. Launch viewer with data /viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.26968527900428638961173806140069&SeriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.99.1071.19276989507392553891297154495191 & /viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.99.1071.21255249241959598781018667405790&SeriesInstanceUID=1.3.6.1.4.1.14519.5.2.1.99.1071.18655460124733602726552772434272
2. Verify that slices are in order

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS:   Windows 11
- [x] Node version:  18.19.0
- [x] Browser: Chrome 131.0.6778.265
  

### DIFF
--- a/platform/core/src/utils/sortStudy.ts
+++ b/platform/core/src/utils/sortStudy.ts
@@ -1,6 +1,7 @@
 import { vec3 } from 'gl-matrix';
 import isLowPriorityModality from './isLowPriorityModality';
 import calculateScanAxisNormal from './calculateScanAxisNormal';
+import areAllImageOrientationsEqual from './areAllImageOrientationsEqual';
 
 const compareSeriesDateTime = (a, b) => {
   const seriesDateA = Date.parse(`${a.seriesDate ?? a.SeriesDate} ${a.seriesTime ?? a.SeriesTime}`);
@@ -145,6 +146,11 @@ function isValidForPositionSort(images): boolean {
   if (!referenceImagePositionPatient || !imageOrientationPatient) {
     return false;
   }
+
+  if (!areAllImageOrientationsEqual(images)) {
+    return false;
+  }
+
   return true;
 }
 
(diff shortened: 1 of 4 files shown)


IbrahimCSAE:
The issue with my data is fixed now. Thanks!

platform/core/src/utils/sortStudy.ts
```diff
@@ -145,6 +146,11 @@ function isValidForPositionSort(images): boolean {
   if (!referenceImagePositionPatient || !imageOrientationPatient) {
     return false;
   }
+
```
wayfarer3130, line 149: Usually this is sorted on display sets and not series - could you check if this is your use case, and if so, then instead of using areAllImageOrientationsEqual, which is actually a sub-set of what is required, and just check isReconstructable?

It is actually the display sets which should be ordered, and NOT the study series instances, since those may well not sort the same way.

wayfarer3130: In fact if it is done after the display set is created, you can have the display set choose the sort method directly.

arul-trenser: @wayfarer3130  The sorting is happen here https://github.com/OHIF/Viewers/blob/master/extensions/default/src/getSopClassHandlerModule.js#L133, to avoid the sorting twice the customization is provided here itself.

sedghi:
@wayfarer3130 do we need this AND yours? or one of them?

arul-trenser:
> @wayfarer3130 do we need this AND yours? or one of them?

@sedghi  Both PRs appear to be necessary.

cc: @wayfarer3130

PR #4922, opened 2025-04-04
fix: x-ray pixel spacing calibration

pedrokohler:
### Context

Fixes: https://github.com/OHIF/Viewers/issues/4921

But depends on https://github.com/cornerstonejs/cornerstone3D/pull/1966 being merged first. Also need to update packages after it's merged.

### Changes & Results

Before:
image

After:
image

#### PR

- [] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #5231, opened 2025-07-20
fix(calibration): do not override user set calibration

IbrahimCSAE:
### Context

calibration line tool was broken after this PR https://github.com/OHIF/Viewers/pull/4922, it was a much needed one but it also needed to not override user set calibrations.


sedghi:
Add a test so that we don't break it again

IbrahimCSAE:
okay will do

platform/core/src/classes/MetadataProvider.ts
```diff
@@ -535,11 +535,17 @@ const WADO_IMAGE_LOADER = {
     let imageOrientationPatient;
     if (PixelSpacing) {
       [rowPixelSpacing, columnPixelSpacing] = PixelSpacing;
-      calibratedPixelSpacingMetadataProvider.add(instance.imageId, {
-        rowPixelSpacing: parseFloat(PixelSpacing[0]),
-        columnPixelSpacing: parseFloat(PixelSpacing[1]),
-        type,
-      });
+      const calibratedPixelSpacing = utilities.calibratedPixelSpacingMetadataProvider.get(
```
wayfarer3130, line 538: Can you update the values when they change?

wayfarer3130: Never mind, that happens elsewhere.   Just testing this now.

PR #5232, opened 2025-07-21
Fix DefferedPromise implementation in retrieveMetadataLoaderAsync.js

elazarcoh:
Current implementation of `DefferedPromise` has some issues. I encountered it while changing the `enableStudyLazyLoad` from false to true in my app. Investigation revealed that the issue is originated because of the issues in the `DefferedPromise`. Just look at the differences between the implementations and you'll understand.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


elazarcoh:
The issue did occur when I opened the viewer directly with a studyinstaceuid in the url, instead of going through the study browser.
In that case, when setting the enableStudyLazyLoad to true, the viewer just hanged loading. It seemed like some promises were never awaited. The fix for it was to fix the chain. It might be the queue is not necessary, only to fix the returned promise when doing "then".

PR #5234, opened 2025-07-21
fix(security): Pin eslint-config-prettier and eslint-plugin-prettier versions.

jbocce:


addOns/externals/devDependencies/package.json
```diff
@@ -39,14 +39,14 @@
     "css-loader": "^6.8.1",
     "dotenv": "^8.1.0",
     "eslint": "^8.39.0",
-    "eslint-config-prettier": "^7.2.0",
+    "eslint-config-prettier": "7.2.0",
```
sedghi, line 42: can we fix all of our dependencies to fixed from now? replace "^ to " in the package json

jbocce: A straight removal of "^" is easy but, if we want to pin based on the actual version we were using then it is a bit of a chore. For example, `"eslint-plugin-prettier": "^5.0.0"` changed to `"eslint-plugin-prettier": "5.5.1"` to be consistent with what version we were actually using.

So which would you like?

PR #5250, opened 2025-07-23
fix(security): Removed stylus package due to potential security vulnerability and we don't use it.

jbocce:


wayfarer3130 (approved):
Approved as soon as the build passes.

sedghi:
same PR to 3.10 please

PR #5262, opened 2025-07-25
test(SegmentationPanel): add test for jump to center of segments from PACS and basic add, rename, delete from panel

dxlin:
-test segmentation from PACS jumps to center when clicked on segmentation panel

### Context

Create test for 
1. Basic add, rename, and delete
2. Jump to center slice index when segment loaded from PACS (note that it seems this only works for segments loaded from PACS and not new segments or newly modified segments)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/tests/SegmentationPanel.spec.ts
+++ b/tests/SegmentationPanel.spec.ts
@@ -1,0 +1,97 @@
+import { test, expect } from 'playwright-test-coverage';
+import { visitStudy } from './utils';
+
+test.beforeEach(async ({ page }) => {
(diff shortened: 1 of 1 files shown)


jbocce (approved):
Thanks for these tests! I like the various asserts/expects that you used. Code approved!

If you could please log an issue for the jump to slice not working when clicking a segment in the side panel then that would be greatly appreciated.

PR #5285, opened 2025-07-31
fix local ModalitiesInStudy check

rodrigobasilio2022:
### Context

When using /local path, the ModalitiesInStudy is not read properly as /local does not communicate with a server. Also only the  SeriesInstanceUID is read correctly, sothis PR uses the instances[0] metadata to read the Modality correctly

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -77,15 +77,25 @@ export default class HangingProtocolService extends PubSubService {
...
-        metadata.ModalitiesInStudy ??
-        (metadata.series || []).reduce((prev, curr) => {
-          const { Modality } = curr;
-          if (Modality && prev.indexOf(Modality) == -1) {
-            prev.push(Modality);
-          }
-          return prev;
-        }, []),
+      callback: metadata => {
+        if (Array.isArray(metadata.ModalitiesInStudy) && metadata.ModalitiesInStudy.length > 0) {
```
wayfarer3130: if (metadata.ModalitiesInStudy?.length)   is a sufficient check as it will be undefined/null or an array

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -77,15 +77,25 @@ export default class HangingProtocolService extends PubSubService {
...
+          return metadata.ModalitiesInStudy;
+        }
+        if (Array.isArray(metadata.series)) {
```
wayfarer3130, line 85: Can you extract the logic for modalities in study calculation to a utility?  Really, no one cares about it beyond it beyond the name, so it reduces the complexity of HangingProtocolService to not have it calculated inline.
Also,
metadata.series.map(ser => ser.Modality).reduce( ... ) makes it a bit clearer.

platform/core/src/services/HangingProtocolService/HangingProtocolService.ts
```diff
@@ -19,6 +19,26 @@ const DEFAULT_VIEWPORT_OPTIONS: HangingProtocol.ViewportOptions = {
...
 };
 
+/**
+ * Returns an array of unique values for the given attribute from a series array.
+ * If the attribute is not present on the series, attempts to get it from the first instance.
+ * @param {Array} series - The series array to extract attributes from.
+ * @param {string} attribute - The attribute name to extract.
+ * @returns {Array} Array of unique attribute values.
+ */
+function getUniqueAttributeFromSeries(series, attribute) {
```
wayfarer3130: Can you move this to the utilities, can call it getUniqueAttributeFromList
This is a generally useful function and it shouldn't stay embedded in the hanging protocol service.

wayfarer3130 (approved):

Issue #5291, opened 2025-08-04
[Feature Request] Add Proxy-Based Metadata Fallback for Multiframe DICOM Instances


Devu-trenser:
### What feature or change would you like to see made?

In multiframe DICOM datasets, key image plane attributes such as ImagePositionPatient, ImageOrientationPatient, and PixelSpacing may not be directly present in each frame's metadata. Instead, they are often stored at a parent or shared level.

### Why should we prioritize this feature?

Currently, few utility functions assume that these fields exist directly on the instance object.
Examples:

- https://github.com/OHIF/Viewers/blob/master/platform/core/src/utils/getClosestOrientationFromIOP.ts#L12
- https://github.com/OHIF/Viewers/blob/master/platform/core/src/utils/isDisplaySetReconstructable.js#L55

In this case, if the value isn't present directly on instance, the logic silently fails or behaves incorrectly, even though the data exists at a higher level in the metadata hierarchy.

This feature introduces a reusable metadata proxy (wrapWithMetadataProxy) that transparently resolves these fields using fallback logic in the following order:

- From the instance itself
- From _parentInstance
- From _parentInstance._shared

This ensures utilities and renderers relying on these fields continue to work correctly without requiring manual fallback code in every usage site.

Related issues: [OHIF/Viewers#4804 (comment)](https://github.com/OHIF/Viewers/issues/4804#issuecomment-2701056580)

PR #5292, opened 2025-08-04
feat(metadata): metadata access improvement

Devu-trenser:
### Context

Feature request: https://github.com/OHIF/Viewers/issues/5291
In multiframe DICOM datasets, metadata such as `ImagePositionPatient`, `ImageOrientationPatient`, and `PixelSpacing` are often defined at the shared or parent level rather than in each frame.

Currently, utility functions like [`getClosestOrientationFromIOP`](https://github.com/OHIF/Viewers/blob/master/platform/core/src/utils/getClosestOrientationFromIOP.ts#L12) assume these fields exist directly on the instance. This can lead to undefined values and incorrect behavior.
This PR introduces a metadata proxy that automatically resolves these fields through a fallback mechanism (`instance -> _parentInstance -> _parentInstance._shared`), ensuring consistency and preventing silent failures across the codebase.

This PR is incorporated by [FlyWheel.io](https://flywheel.io/)

### Changes & Results

-  **Added**: `wrapWithMetadataProxy()` utility with symbol-based double-wrapping prevention.
- **Logic Update**: In `addInstances()`, automatically wraps multiframe instances with the proxy.
- Now fields like `ImagePositionPatient` can be directly accessed.
- Proxy logic is scoped only to a defined set of metadata fields (`ImagePositionPatient`, `ImageOrientationPatient`, `PixelSpacing`) to minimize risk.

### Testing

- Load a multiframe DICOM study with metadata directly not accessible.
- Check with the measurement jump feature to check metadata value in getClosestOrientationFromIOP function.
- Check the PixelSpacing in isDisplaySetReconstructable function

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


Devu-trenser:
@wayfarer3130  Could you please review it when you get a chance?

wayfarer3130:
There is already a proxy based metadata multi-frame accessor, which uses a bunch of custom logic to get the correct positioning based on how it gets defined.  I'd much rather that gets used consistently than trying to re-implement some of it in different places.  See combineFrameIntance.ts for the logic, and MetadataProvider for usage of it.
You should be able to get the per-frame overall data using the get instance call, and the data should be available there.  That should be available from the shared data, the per-frame data or the overall root data.

Longer term, the right way to do it is to create custom getters that take a frame and attribute and get the right value directly, as that prevents needing to use a proxy object.  Those should take something like  module name from the DICOM module set, the base image id and the frame number/tag of interest, and return the value.  Having variants on that to allow detecting if the value differs across the series or multiframe object allows for some types of checks to be extremely quick.

wayfarer3130:
The problem with this change is that it introduces another layer of indirection to requesting values, and the parsing/setup for metadata in OHIF is already one of the performance issues.  Instead, I've moved the retrieval such that it is only added for the specific fields of interest.

wayfarer3130 (approved):

platform/core/src/utils/addProxyFields.js
```diff
@@ -0,0 +1,57 @@
...
+
+  for(const fieldProxy of fieldProxies) {
+    if( fieldProxy in instance) {
+      continue;
+    }
+    Object.defineProperty(instance,fieldProxy, {
+      configurable: true,
+      enumerable: true,
+      get: () => {
+        return instance._parentInstance?.[fieldProxy] ?? instance._parentInstance?._shared?.[fieldProxy];
```
sedghi, line 46: @wayfarer3130 
who is adding this metadata to _parentInstance and _parentInstance._shared? I don't see anywhere in our code base these variables

Devu-trenser: @sedghi
The metadata is added inside the `combineFrameInstance` function. `_parentInstance` and `_parentInstance._shared` are initialized within that function to store instance-specific and shared data. They aren't declared elsewhere, they're created as part of combining frame data when the function runs.

You can view the code here: https://github.com/OHIF/Viewers/blob/master/platform/core/src/utils/combineFrameInstance.ts#L84

cc: @wayfarer3130

Devu-trenser:
@wayfarer3130 @sedghi Are we good to merge this MR?

wayfarer3130:
@Devu-trenser We are looking at a general alternative to resolve the issues in both OHIF and CS3D 
@sedghi - I'm ok merging this for now to fix the immediate issue, even if we do replace the entire class in the future.

wayfarer3130 (approved):
@sedghi - as noted separately, I think this change is worthwhile for now to resolve the immediate need as it doesn't add or remove API, simply making additional fields iterable in multi-frame.

sedghi (approved):

PR #5293, opened 2025-08-04
fix 5288 removal of accept header options in retrieve metadata

### LINKED ISSUE #5288: [Feature Request] Removal of accept header config options inclusion during study metadata request.

# Introduction

Some VNAs are picky about accept header handling. Working out this issue with my VNA, I noticed that omitting the accept header makes more sense than its inclusion during series discovery. What I mean is that asking for `application/dicom+json` should be sufficient to allow the server to decide how to handle the request. It got me thinking of when do we really want the full slice instead of the JSON when requesting a study's metadata from the WADO interface? Do data sources realistically transfer raw DICOM instead of JSON via `/metadata`? I think it is more likely for developers to gravitate to `JSON` for `/metadata` and elsewhere for actual DICOM retrieval.

The current default behavior leads to request rejection in my VNA. Although, the VNA is partly to blame, it got me thinking about the semantics of `/metadata`.

# Plan

- Remove the `generateWadoHeader` for the WADO interface when fetching from `_retrieveSeriesMetadataAsync` and `_retrieveSeriesMetadataSync`.

It's a low priority request, but I will be working on the application for some time now and I will be providing the PR for the change.
I will need this change to continue testing the application and developing other patches.

luissantosHCIT:
### Context

Fixes #5288.

My VNA does not accept the default extra options in the `Accept` header for a `WADO /metadata` request. It expects either an `application/json` or `application/dicom+json`. The only returned data is purely in `JSON`.

It got me thinking that rarely would you expect anything but `JSON` in a `/metadata` request response (specially at the study and series levels). Otherwise, the server would have to respond with all of the slices in a single multipart stream, which would be very wasteful!!!

For these reasons, it makes sense to me to omit the `Accept` header options during the `/metadata` request only. Other requests are not affected. I am open to counter points and I have full attention on this project for the foreseeable future, so I can address potential concerns.

### Changes & Results

- Added flag in `generateWadoHeader` to allow `_retrieveSeriesMetadataSync` and `_retrieveSeriesMetadataAsync` to skip accept header generation.
- Added a few TypeScript interfaces to improve function signatures and field expectations in `getAuthorizationHeader` and `generateAcceptHeader`.
- As a result, made stylistic refactors to accommodate the new interfaces.
- Upgraded `user` and `getAuthorizationHeader` modules to TypeScript modules.
- Added `RequestHeaders` types module to house the new interfaces.

#### Before

image

#### After

image

### Testing

This is purely a refactor type of change so it does not contain new features to test.
It also lacks behavior changes other than omitting the Accept header options.
For those reasons, I relied on already defined tests to guard against regressions.

#### Unit Tests

image

#### E2E

image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [Linux/Ubuntu] OS: 
- [20.18.0] Node version: 
- [Firefox 141, Chromium 138.0.7204.157] Browser:
  


extensions/default/src/DicomWebDataSource/index.ts
```diff
@@ -410,7 +416,9 @@ function createDicomWebApi(dicomWebConfig: DicomWebConfig, servicesManager) {
       madeInClient
     ) => {
       const enableStudyLazyLoad = false;
-      wadoDicomWebClient.headers = generateWadoHeader();
+      // Skip inclusion of Accept Header options other than the request type of `application/dicom+json`
+      // See issue #5288
+      wadoDicomWebClient.headers = generateWadoHeader(true);
```
wayfarer3130: According to https://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_10.4.4 the default accept header in the absence of anything else for the metadata is indeed the JSON encoding.  However, at least some web servers REQUIRE the accept header.  I suggest passing in the request type "metadata", and the default inclusion type, and if not specifically configured to require the accept header, then omit it only for metadata.
Thus, the call would be:
generateWadoHeader('metadata', null)
and search of a configuration in dicomWebConfig.acceptHeaderSkip${queryType}

Probably skipping it by default for bulkdata requests would also work.

luissantosHCIT: The default behavior is to generate => `Accept: application/dicom+json` omitting the default options such as multipart and transfer syntaxes. Are you saying that some web servers reject requests with Accept header => `Accept: application/dicom+json`? Does that mean that some web servers are not conformant with this part of the standard?

To be clear, the current change simply leads to including `Accept: application/dicom+json` instead of `Accept: application/dicom+json, multipart/related; type=application/octet-stream; transfer-syntax=*`

I would like to avoid adding a new configuration item because configuration options are easy to add but very hard to remove. In my opinion, it would be best if `generateWadoHeader` could be driven purely with `dicomWebConfig.acceptHeader`.

For example
```
acceptHeader: {
          '/metadata': 'application/dicom+json...',
},
```
But I suspect that would break current configurations and require larger changes, so I refrained from refactoring the current behavior. 

After these clarifications, I will think about your suggestion.

Thank you for the review!

sedghi: I agree with Bill

extensions/default/src/DicomWebDataSource/index.ts
```diff
@@ -128,27 +129,32 @@ function createDicomWebApi(dicomWebConfig: DicomWebConfig, servicesManager) {
...
+        const xhrRequestHeaders: HeadersInterface = {};
         const authHeaders = userAuthenticationService.getAuthorizationHeader();
         if (authHeaders && authHeaders.Authorization) {
           xhrRequestHeaders.Authorization = authHeaders.Authorization;
         }
         return xhrRequestHeaders;
       };
 
-      generateWadoHeader = () => {
+      generateWadoHeader = (skipAccept: boolean = false): HeadersInterface => {
```
wayfarer3130: Please rename to "includeTsuid" and set it to true only when the full generated accept header is required (or you can default it to true, and set it to false when required).  That way it is actually clear what is occurring.

extensions/default/src/DicomWebDataSource/index.ts
```diff
@@ -128,27 +129,32 @@ function createDicomWebApi(dicomWebConfig: DicomWebConfig, servicesManager) {
...
+            dicomWebConfig.acceptHeader,
+            dicomWebConfig.requestTransferSyntaxUID,
+            dicomWebConfig.omitQuotationForMultipartRequest
+          );
+          return {
+            ...authorizationHeader,
+            Accept: formattedAcceptHeader,
+          };
+        } else {
+          return {
```
wayfarer3130, line 177: Please add a comment here that the base/default header from dicomweb-client will get included as appropriate for this request.

extensions/default/src/DicomWebDataSource/index.ts
wayfarer3130: I agree it is acceptable to always just use the baseline header for non-image bulkdata retrieves, so please reverse the parameter type, calling it "includeTsuid", still default it to true, but set the value to false here to indicate NOT to include the tsuid specifically.

wayfarer3130:
Once the requested changes are included, and the branch updated to current/passes the tests, then I agree this change is acceptable and I can merge it.

wayfarer3130:
It was suggested to use includeTransferSyntax as an option in the call, rather than includeTsuid so that it is clearer naming/more descriptive.

luissantosHCIT:
I made the changes as requested.

wayfarer3130:
Just waiting for the build/tests to complete.

platform/core/src/DICOMWeb/getAuthorizationHeader.ts
```diff
@@ -1,17 +1,23 @@
...
+ * @param {Object} [user]
+ * @param {function} [user.getAccessToken]
  * @returns {Object} { Authorization }
  */
-export default function getAuthorizationHeader({ requestOptions } = {}, user) {
-  const headers = {};
+export default function getAuthorizationHeader(
+  {requestOptions}: RequestOptions,
+  user: UserAccountInterface = {}): HeadersInterface
+{
```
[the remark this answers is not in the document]
sedghi: @wayfarer3130 please check this out

PR #5306, opened 2025-08-06
feat: Segmentation stats calculation customizable

Celian-abd:
### Context

see issue : https://github.com/OHIF/Viewers/issues/5301

### Changes & Results
Computation cost of labelmap statistics can be avoided if the project is ready to handle statistics calculation on labelmap by custom code (or even not need calculation at all for example labeling purpose for deep learning datasets)

It would need to add this doc :
panelSegmentation.disableUpdateSegmentationStats | Control labelmap statistics refresh on each labelmap update (default true) 

### Testing
set 'panelSegmentation.disableUpdateSegmentationStats': true to disabled stats computation

#### PR

- [] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #5324, opened 2025-08-11
Segmentation stats customization module

salimkanoun:
Replacing https://github.com/OHIF/Viewers/pull/5306 as Celian is away,

@sedghi could you tell me where is the hover you want to remove ?


salimkanoun:
@sedghi I think tooltip stats display is already covered by a check if the segments has namedStats see https://github.com/salimkanoun/Viewers/blob/33b2900cb614ac0d5a0c35f2104c4e62a5f46fed/platform/ui-next/src/components/SegmentationTable/SegmentationSegments.tsx#L105

PR #5327, opened 2025-08-12
fix 5287 optimize series discovery on load time

### LINKED ISSUE #5287: [Feature Request] Optimize WADO metadata retrieval.

# Introduction

I am new to this project and I appreciate the work you have put. 

While testing the viewer in my development environment, I noticed that metadata retrieval is large. Based on my current level of familiarity with the code base, I believe the main contribution of the metadata retrieval is for series discovery. In my experience, the more efficient route for series discovery is to break down the process into QIDO queries and only ask for metadata at the lowest level possible. Unfortunately, DICOMWeb's WADO interface does not allow for QIDO like refinement of target search and thus you are left with pulling megabytes of metadata for a few bits of information.

If you think I misconfigured my dev environment or misunderstood the application behavior, feel free to help me understand.

This is the first of several patches I want to contribute to the project. I am just breaking all of these in as small a unit as possible for your review benefit. I will be creating PRs soon after setting up all of the tickets I want to start with.

# Example Retrieval

Image

# Sample Data Source

```
{
      sourceName: 'dicomweb',
      configuration: {
        friendlyName: '[redacted] (Test)',
        name: 'VNA',
        wadoUriRoot: '[redacted]/wado-rs/[redacted]',
        qidoRoot: '[redacted]/qido-rs/[redacted]',
        wadoRoot: [redacted]wado-rs/[redacted]',
        imageRendering: 'wadors',
        enableStudyLazyLoad: false,
        supportsFuzzyMatching: true,
        supportsWildcard: true,
        thumbnailRendering: 'wadors',
        acceptHeader: [
          'multipart/related; type=application/octet-stream; transfer-syntax=1.2.840.10008.1.2.1.99',
        ],
        requestOptions: {
          auth: '',
        },
        dicomUploadEnabled: true,
        singlepart: 'video,thumbnail,pdf',
        bulkDataURI: {
          enabled: false,
          relativeResolution: 'series',
        },
        omitQuotationForMultipartRequest: true,
      },
    },
```

# Plan

- [QIDO] Get list of studies for patient ID.
- [QIDO] Get selected study's series list using the given Study UID.
- [WADO] Fetch first slice for the thumbnail.

- Large data transfers waste bandwidth.
- Large data transfers make the application appear slower than it is.
- Large data transfers put more load on the data source host than necessary.

luissantosHCIT:
### Context

The synchronous load method pulls too much metadata during the loading phase. This wastes resources such as bandwidth, server connections, and user patience. In my test case, I was downloading up to 29 MB of JSON metadata which could take up to 10 seconds. The vast majority of the metadata is not used at all when setting up the UI/Hanging Protocol.

We can have a snappy synchronous experience by requesting and making use of as much of the `QIDO` metadata as possible.
We can also skip downloading the full study metadata and simply download two slices worth of metadata per series. With a little `IPP` patching, the Viewer can still enable the 3D tools.

Breaking the process in smaller QIDO and WADO queries allows for parallelization of the request against the server thus minimizing time spent requesting data.

Fixes #5287 

#### Note

I apologize for the number of changes but I saw opportunities to improve code quality while chasing a few defects prior to PR release. I took time to add more type definitions, reorganize some of the files, and testing both the lazy load and sync load features.

Also, note that this only takes care of the metadata portion itself but not the image load. I will tackle the image load optimization in a separate PR as an opt in setting. For now, the UI should update pretty fast although the thumbnail retrieval will prevent it from showing the images which can take a second in my environment.

Ignore commits from August 4th since those are being reviewed in #5293. They are only showing here because I had to base my new branch off the one associated with that PR so I could test against my live environment as well.

Finally, I will be adding a few more commits but those will be unit tests. I decided to open this PR to allow initial reviews. I should be adding the rest of the unit tests soon (tomorrow).

### Changes & Results

- Broke down metadata retrieval into study and series level QIDO queries.
- Added logic for retrieving the first and last slice for each series discovered in study.
- Added logic for preparing the metadata for viewer consumption, which includes updating the IPP and instance number of reconstructed missing items.
- Refactored the organization of the `DicomWebDataSource` module/component.
- Added several new TypeScript modules and upgraded a few source files to be in TypeScript so I could apply type annotations.
- Added a few basic tests (I will add more tomorrow while awaiting for a reviewer).
- Added function and module docstrings.
- Removed no longer needed code.
- Moved a few functions into their own TypeScript module.

#### Before

image

#### After

image

### Testing

- Run a webserver or dev node script.
- Open a web browser and make sure to show the network tab from dev tools.
- Navigate to page and observe the burst of queries and fetches each taking a few milliseconds.
- Confirm that only a few hundred KB are downloaded for a given study in the worst case scenario.

#### PR

- [x] Added new unit tests
- [x] Ran unit tests
- [x] Ran end to end tests

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [Ubuntu Linux] OS: 
- [20.18.0] Node version: 
- [Firefox 141, Chromium 138.0.7204.157] Browser:
  


luissantosHCIT:
Added the other tests I wanted to add.
Feel free to request other tests I should include.

extensions/default/src/DicomJSONDataSource/index.js
```diff
@@ -280,21 +288,18 @@ function createDicomJSONApi(dicomJsonConfig) {
...
         }
       });
 
       return imageIds;
     },
-    getImageIdsForInstance({ instance, frame }) {
-      const imageIds = getImageId({ instance, frame });
-      return imageIds;
-    },
+    getImageIdsForInstance: getImageIdsForInstance,
```
wayfarer3130: This can be just getImageIdsForInstance,
as your assignment variable name is identical to what you are assigning it to.

wayfarer3130:
It is indeed interesting to be able to query for studies with qido and display them, fetching first/last etc, but it definitely needs to be something that can be turned on/off with a flag because the performance of this can be either much better or much worse than the existing behaviour.  The reason is that not all systems store all the metadata as individual pieces, but if they store it like static dicomweb or other stream optimized variants, then the fastest way to get the data is actually to fetch metadata for the entire series at once - that allows the best compression to be applied, and works equally well for both  single and multiframe.
On the OHIF side, these changes break unknown items throughout the viewer.  For example, have you tried this at all with whole slide microscopy viewing in the CS3D mode?  Another example is CINE playback rates - the tags that end up being needed differ based on modalities.  Custom hanging protocols can depend on any tag, and so can custom split/sort rules.
You will need really clear documentation on exactly what queries get made under what circumstances and how to decide whether that will work for various display conditions.

wayfarer3130:
To make this more configurable/manageable without introducing bugs, could you create a new data source with these changes? 
Something like @ohif/extension.dataSourcesModule.dicomweb_minimal
and document exactly what it requires differently, and what sorts of things that will affect.
This may need to make changes to the existing data sources to make them more extensible, but that is something that has been intended all along.

I would set it up so that it is easy to fall back to the basic data source retrieval options on conditions like only have <= N sop instances - that way you can use N~=10 and just fallback to retrieving full metadata, and only use the qido options when they actually make a significant difference.  It might also need to have some way to customize when that occurs, and almost certainly needs to have a way to go back and reload everything using the full metadata set on a per-series basis.

luissantosHCIT:
> It is indeed interesting to be able to query for studies with qido and display them, fetching first/last etc, but it definitely needs to be something that can be turned on/off with a flag because the performance of this can be either much better or much worse than the existing behaviour. The reason is that not all systems store all the metadata as individual pieces, but if they store it like static dicomweb or other stream optimized variants, then the fastest way to get the data is actually to fetch metadata for the entire series at once - that allows the best compression to be applied, and works equally well for both single and multiframe. On the OHIF side, these changes break unknown items throughout the viewer. For example, have you tried this at all with whole slide microscopy viewing in the CS3D mode? Another example is CINE playback rates - the tags that end up being needed differ based on modalities. Custom hanging protocols can depend on any tag, and so can custom split/sort rules. You will need really clear documentation on exactly what queries get made under what circumstances and how to decide whether that will work for various display conditions.

Let me make it a flag for now. You bring excellent points that I need to consider.

In terms of the WADO/QIDO part of the standard, it's a REST API, so I think a standards compliant system has to comply with expected metadata. 

WADO example,

- `GET {s}/studies/{study}/series/{series}/metadata` => has to return the series metadata
- `GET {s}/studies/{study}/series/{series}/instances/{instance}/metadata` => has to return the instance metadata

https://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_10.4.1.1.2

QIDO example,

-  `GET {s}/studies/{study}` => has to return the series list for the study
- `GET {s}/studies/{study}/series/{series}` => has to return the instance list for the series

https://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_10.6.3.3.1
https://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_10.6.3.3.2
https://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_10.6.3.3.3

I do understand helping gzip or zstd achieve better compression, but I am not convinced that it is necessarily faster to retrieve MB of data in most cases. In my particular case, the response is already gzip compressed. This is something that would have to be extensively benchmarked across vendor implementations and it is well outside my scope and resources at the moment. 

With that said, if we stick strictly to the standard, QIDO as deep into the tree as possible and then WADO at the leaf should include much fewer metadata than WADO at the higher levels (series) despite having multiple separate requests and potentially poor compression per request. Also, we get to benefit from request parallelization and keeping them close to less than 14KB to avoid the TCP slow start tax per connection. Meaning, it should be near instantaneous which is what I found in my crude testing.

This is not to discredit your observations. I am thinking that most studies should have a lot of metadata because that is the nature of DICOM (CT, MRI, PET, DWI, Perfusion, etc...) and DICOMweb lacks a GraphQL specification like FHIR has.

**PS.....**

**I just read your follow up comment and I fully agree. That is a better way to approach this. I was avoiding adding to the project since I am new here, but I do like that suggestion the more I think about this. One thing we (more like I) might need to revisit the correct threshold for falling back to full retrieval. Starting at 10 sounds good to me for now.**

I will check CINE and CS3D if possible. Expect changes starting tomorrow since I need to take care of something this afternoon at my institution.

Thank you. This is a good discussion! 

Also, I already saw a few other emails about style and I will look at those as well for the other PRs, but I wanted to share that TypeScript is not my primary language so occasionally you might see more of a C or Rust or Pythony style when writing things. I am going to take this opportunity to learn from you!

luissantosHCIT:
**Moved my changes to its own module as requested.**
**Do not review yet. I am missing the implementation of the default behavior if slices < 10.**

luissantosHCIT:
TODO:
- [x] Organizing the Data Source directories.
- [x] Refactoring imports and logic out of the individual DicomWeb DataSource items.
- [x] Created the new Minimal Data Source
- [x] Added default full retrieval to Minimal Data Source.
- [x] Added a few docstrings.
- [ ] Documentation of new Minimal DicomWeb Data Source.
- [x] Review all tests.
- [ ] Add new tests.

wayfarer3130:
The different queries do in fact need to be implemented, but the performance of different types of queries and what is available to actually return differs between systems.  As well, the actual parsing performance differs between systems.  
During testing of different PACS systems, the performance of the back end PACS system was always the determining factor for how fast things could display, and not the transfer speed.  Only when there was a metadata gzip cache in between the PACS and the front end did transfer speed become an issue, and that only occurs for very large single frame studies.  Those ones get pretty decent performance when switching to multi-frame.  I was able to test that on tens of thousands of frames and get time to first image of a couple of seconds from clicking on the image in the display area.

Given that, to figure out how to better deal with slower PACS systems or ones not supporting multiframes:
1. Reduce the number of query/retrieves used
2. Reduce the expense of the query/retrieves done

For the first one, try the hanging protocols which don't load the entire study, but only load the required series.  That means that it has to be determinable which series are displayed first by things available in the series query, without needing to load all the metadata.  That can be a problem for some study types.

For the second, you idea of using the instances query is probably the best option - however, you still need to be able to determine the split/sort rules that are applicable.  There are ways to "guess" that, but the query data returned by the initial query isn't really good enough to guess that - something like:
a. Load the instance query for the series determined to be of interest
b. Sort the series by the appropriate criteria
c. Let the split/sort rules determine which instances need full metadata up front (this can use the first/last series, number of frames and other common attributes to guess if this instance needs full metadata) - this may determine it needs all metadata for this series.
d. Load instance metadata in batches as required, or if all metadata is needed, load it all up front

You can start seeing why a general solution to not load all the metadata up front becomes a problem because you need to deal with a lot of different use cases.  If the hanging protocol used to determine initial load can also determine whether it is safe to do a query instance metadata for some of the instances, then it would help you to switch between full metadata modes and partial, even on a single series.  Those hanging protocols probably need to be tuned to the exact back end response data you can get on the instance level query and series level queries.

wayfarer3130:
I merged the previous PR, but there are conflicts in the current one that need updating.
Let me know when you have moved this to a separate data source and I can look at it again.

Have you tried caching/gzipping the metadata object?  With it gzip encoded so that the browser does the ungzip, and with it cached, I find the performance of the retrieve/display is quite good.  There are more efficient formats which could be used - a relatively easy one is to use a tag/value encoded pairs where the value can indicate exceptions, computed or single value items - something like:

ImagePatientPosition: { type: 'perFrame', value: [ [0,0,1], [0,0,2], ...] }
ImagePatientPosition: { type: 'nested', value: [0,0, { type: 'computed', value: '${frameIndex}+1'}] }

Rows: 512,
Columns: 512,
PhotometricInterpretation: { type: 'exception', value: 'RGB', exceptions: { '5': 'HSV' } },

This compresses about 100:1 for not that large CT type series, and even better, it parses at about the same performance advantage.  

The problem with not getting all the image data is that you can falsely assume the data is regular, and that can cause defects in the display such as showing a volume when there are slices missing.  To alleviate that, I'd suggest always loading the full metadata, and popping up a warning if it is significantly different from what is expected.  The problem is how to tell what is significant and what isn't because more complex hanging protocols can use a lot more fields to display the data correctly.  As long as you know the potential problems, and can configure your data source as a separate data source, I'm ok with having it separate, but I just want you to know some of the perils - this should NOT be used for clinical workflows as there isn't a good fallback position to know when to get all the data.

luissantosHCIT:
> he full metadata, and popping up a warning if it is significantly different from what is expected.

I agree about having it as a separate data source and I believe I did so in one of the PRs but I may not have retrospectively pushed changes to other branches. I had to keep a linear dependency so I would continue meeting customer demands while doing my best to meet project demands. Now that I might have a chance to be more involved with the remainder of my work, I am going to revisit the changes and make sure the branches are correct.

Thank you for your patience!

PR #5328, opened 2025-08-12
feat(orientationMarkers): improved the orientation markers with a simplified version

lscoder:
A simplified orientation markers version that does not need to check if the series is flipped or rotated because it gets the row/column cosines vectors directly from the viewport.


sedghi (approved):
I like it, thanks

Issue #5082, opened 2025-05-29
[Bug] TID1500 SR SCOORD bounding box annotations do not show


fedorov:
### Describe the Bug

According to @wayfarer3130 and @pedrokohler, per discussion in https://github.com/OHIF/Viewers/issues/1215#issuecomment-2912606769, "[SCOORD and SCOORD3D TID1500 annotations will] only be visible, but they won't be hydrated (meaning they won't be editable or interactable at all)."

There is a regression apparently, and they are in fact not visible.

### Steps to Reproduce

1. Download sample from [here](https://github.com/user-attachments/files/18790628/prostate_point_02_13_25.zip)
2. Load folder in the viewer via https://viewer.ohif.org/local
3. Drag and drop SR into viewport
4. Try finding the point

### The current behavior

The point is not visible

### The expected behavior

The point, according to @wayfarer3130 and @pedrokohler, should be visible.

### OS

macOS

### Node version

n/a

### Browser

Chrome

### Steps to Reproduce

1. Download sample from [here](https://www.dropbox.com/scl/fi/zxikh4kp53xeg0y7go2i6/OHIF_bboxes_02_13_25.zip?rlkey=koce906ii9js8pg2ewuzc6snu&st=jszu9i08&dl=0)
2. Load folder in the viewer via https://viewer.ohif.org/local
3. Drag and drop SR into viewport
4. Try finding the bounding box

### The current behavior

Bounding box does not show up

### The expected behavior

Bounding box should show up

### OS

macOS

### Node version

n/a

### Browser

Chrome

fedorov:
I should deploy 3.11 and then ask Deepa to test. Data is in `/projects/idc-tcia/locations/us-central1/datasets/tcia-idc-datareviewcoordination/dicomStores/planar_annotations/study/1.3.6.1.4.1.14519.5.2.1.7310.5101.860473186348887719777907797922`.

fedorov:
@igoroctaviano unable to test with 3.11.0 - I am getting a crash with `?gcp=` parameter.

<img width="559" height="520" alt="Image" src="https://github.com/user-attachments/assets/2572361a-76fa-431f-96c4-7333542dcc3f" />

fedorov:
@igoroctaviano @deepakri201 I am testing with a sybil bounding box sample. The great news is I see the bounding box. The bad part it is in the wrong place. I do not know if the issue is in the data or in OHIF...

Deepa, you should be able to access the URL below.

[...]

PR #5335, opened 2025-08-14
fix: load 3dsr

wayfarer3130:
### Context

The 3d sr loading got broken due to removing a sop class uid from a list of supported classes.
Additionally, this PR adds the view reference to the unloaded SOP class to help with navigation.

### Testing

Use the two point example SR from  https://github.com/OHIF/Viewers/issues/5082
Hydrate the SR, and you should see two lesion annotations, displayed as probe instances.
The points are not visible until after hydration - there is an issue with needing an image id to navigate to in the SR display set,
AND with needing an image id to be visible in the stack viewport.  This isn't an issue after rehydration.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/modes/longitudinal/src/initToolGroups.js
+++ b/modes/longitudinal/src/initToolGroups.js
@@ -87,9 +87,6 @@ function initDefaultToolGroup(extensionManager, toolGroupService, commandsManage
     enabled: [
       { toolName: toolNames.ImageOverlayViewer },
       { toolName: toolNames.ReferenceLines },
-      {
-        toolName: SRToolNames.SRSCOORD3DPoint,
-      },
     ],
     disabled: [
       {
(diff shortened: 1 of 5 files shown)


modes/longitudinal/src/initToolGroups.js
```diff
@@ -87,9 +87,6 @@ function initDefaultToolGroup(extensionManager, toolGroupService, commandsManage
     enabled: [
       { toolName: toolNames.ImageOverlayViewer },
       { toolName: toolNames.ReferenceLines },
-      {
-        toolName: SRToolNames.SRSCOORD3DPoint,
```
igoroctaviano, line 91: Why remove this tool? IDC uses it

wayfarer3130: It doesn't load data properly and causes a second instance of the tool to appear using the default probe tool.  It should be possible to add this back in and rehydrate it properly, replacing the Probe tool as the SR Coord 3d.

igoroctaviano (approved):

PR #5337, opened 2025-08-14
fix 5323 unexpected series list filter change from all to recent

luissantosHCIT:
### Context

See issue #5323.

Basically, under an specific condition, selecting a same day (similar) study was forcing a change in view filter from the `All` mode I was in to the `Recent` mode which was not expected. It startled me enough to think that the UI had dropped the series list via a filter.

The issue here is that there is a search step performed by the StudyBrowser component looking for the subgroup in which the study lives. If the study is present in multiple subgroups, the first one in the list of subgroups is returned. There is a further hook called to then set the browser's current subgroup to the one returned in the search.

### Changes & Results

- Added Playwright test to simulate issue (note that the AWS store from which Playwright fetches lacks the offending pair of cases so you will not see any before and after during local testing; the offending data is in the local datastore I had from following the docs).
- Updated the logic such that it places the current tab group first in the search to bias the results towards remaining in the current study list until a user selects a different filter.

#### Before

[...]

PR #5345, opened 2025-08-19
feat 5340 addition of html and pdf report rendering of encapsulated SR reports.

### LINKED ISSUE #5340: [Feature Request] Add HTML Report display widget

# Introduction

Currently, some SR encapsulated reports are not properly shown in the viewer. I am referring to Philips/CareStream Final reports which are SR objects with HTML renderings of the final report inside. Something cool about OHIF is that it is able to show the report contents without issue but fails to render the contents as a page.

I have a set of React components from prior work that I would like to adapt and donate to OHIF. The components allow for rendering the HTML, Markdown, and PDF reports. I saw pdf related source files in the OHIF codebase so I might skip the PDF portion, but HTML and Markdown rendering would be great to share. 

I might look into adding XML rendering to show scanner generated reports such as GE US Logiq scanners.

# Unrendered HTML

Image

Image

Enhancing report presentation adds options for users to look at results. In our research institution, sometimes it is helpful to be able to provide anonymized reports alongside the anonymized images in a convenient package. This feature would also make this viewer one step closer to having feature parity with commercial viewers. Finally, HTML rendering will allow folks at SIIM Hackathon to experiment with the idea of interactive reports.

luissantosHCIT:
### Context

Fixes #5340 
Requires #5293, #5327, and #5337 before merging.
Commits start August 15th.

This PR brings a React component that can handle SR encapsulated contents.
Currently, the viewer assumes all data in the SR tree are pieces of formatted text.
However, in my environment, I come across SR series that are encapsulated reports.
The encapsulated report is an HTML report in my case.
As part of the component bundle, I included a component for handling PDF if the encountered data is PDF.
The new components default to text like before.

Also, because of the danger of rendering HTML, I added a dependency on `sanitize-html` and take steps to clean the input HTML as much as possible. I make sure no scripts are allowed to execute when viewing HTML reports.

A follow up PR or two will be needed to introduce:

- Refactor the UTF-8 encoding logic to use a more updated version of `dcmjs` library once my PR (455) gets merged.
- Introduce test data and unit tests.

### Changes & Results

- Added HTML rendering component.
- Added PDF rendering component
- Preserved Text Rendering behavior as default.
- Added dependency on `html-sanitize` to help clean up document.
- Added utility function to make sure DICOM inputs are encoded in utf-8.
- Added regexes to make sure we isolate HTML in the payload and discard anything else.
- Corrected logic bug with obtaining the `CodeMeaning` elements in SR nodes.

#### Before

image

#### After

image

### Testing

E2E tests and test dataset will need to be introduced in a later PR because I need to figure out how to propose test data.
Also, the PR is getting much bigger than I feel comfortable and I think it will create more burden to the reviewer, so I am breaking this into a few chunks.

TODO: Finalize before and after testing with current test suites in the morning.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Ubuntu Linux
- [x] Node version: 20.18.0
- [x] Browser: Firefox 141, Chromium 138.0.7204.157
  

<h3>Greptile Summary</h3>

This PR adds HTML and PDF rendering support for encapsulated SR reports, replacing the flat text-only `OHIFCornerstoneSRContentItem` with a new component stack (`OHIFCornerstoneSRContainerItem` → `OHIFCornerstoneSREncapsulatedReport` → PDF/HTML/text sub-renderers). It also centralises Blob/MIME handling into `platform/core/src/utils/toBlob.ts` and adds `sanitize-html` for XSS mitigation. The bulk of the previously identified bugs (missing `useEffect` deps, `Buffer` browser crash, `JPEG` undefined variable, null guards, memory leaks, wrong PDF detection, stale `data` closure, and the `measurements` null guard ordering) have all been addressed in the current iteration.

Safe to merge; all previously identified P0/P1 bugs have been resolved and only minor style/cleanup findings remain.

The prior round of review surfaced a substantial number of correctness issues. All have been addressed in this iteration. The three remaining comments are P2 style suggestions that do not affect correctness on the happy path.

No files require special attention for merge; OHIFCornerstoneSREncapsulatedReport.tsx and payload.ts are worth a follow-up pass when the test dataset PR lands.

<details open><summary><h3>Vulnerabilities</h3></summary>

- HTML sanitization via `sanitize-html` is in place; `allowVulnerableTags` has been removed and `script` is not in the allowlist.
- `data:` URIs are restricted to `img[src]` only via `allowedSchemesByTag`, mitigating the most common data-URI XSS vector.
- `dangerouslySetInnerHTML` is guarded by `sanitizeHtml()` which calls `extractHTMLFromPayload` + `sanitize-html` before injection.
- The allowlist for tags and attributes is very broad; future tightening is advisable but not a blocking concern.
- No secrets or credentials exposed in the diff.
</details>

| Filename | Overview |
|----------|----------|
| extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSREncapsulatedReport.tsx | New component handling HTML/PDF/text SR content; previous bugs fixed (useEffect deps, null guard, loading spinner), but stale `mime` closure in the effect is a minor edge-case issue |
| extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSREncapsulatedPDFReport.tsx | New PDF viewer component; object URL lifecycle is correctly managed; height=500rem is an invalid HTML attribute value that may cause zero/default height in some browsers |
| extensions/cornerstone-dicom-sr/src/utils/payload.ts | Core payload utilities (MIME detection, HTML sanitization, UTF-8 decode); sanitizer config is verbose with many duplicate entries, but security posture is reasonable; data: URIs restricted to img tag only |
| extensions/cornerstone-dicom-sr/src/utils/srInspection.ts | New utility for inspecting SR tree nodes; null-safety added via optional chaining on ConceptNameCodeSequence |
| extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRContainerItem.tsx | New container item renderer replacing OHIFCornerstoneSRContentItem; encoding fallback chain (content item to root to default) is correct |
| extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRMeasurementViewport.tsx | Null guard moved before measurements[measurementSelected] access; encapsulated SR fallback to text viewport added |
| platform/core/src/utils/toBlob.ts | New consolidated blob/MIME utility; b64toBlob backward-compat alias preserved; MimeOptions enum centralises MIME string constants |
| extensions/default/src/DicomWebDataSource/index.ts | Refactored thumbnail src generation to use utils.toBlob and MimeOptions; Jpeg/AcceptJpeg destructured correctly |
| extensions/default/src/utils/getDirectURL.ts | Switches from new Blob(...) to utils.toBlob and from b64toBlob to b64ToBlob; functionally equivalent refactor |
| extensions/cornerstone-dicom-sr/src/utils/constants.ts | Encoding map adds ISO_IR 192 to utf-8; covers the key UTF-8 DICOM encoding; defaults are consistent |

</details>

</details>

`````markdown
Path: extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSREncapsulatedReport.tsx
Line: 32

Comment:
**Stale `mime` closure in `useEffect` dependency array**

`mime` is used as the fallback argument to `getPayloadType(decoded, mime)` inside the effect, but it is not listed in the dependency array `[props.content, props.expectB64, props.encoding]`. If `props.content` changes between renders and the previous content caused `mime` to be set to a non-default value (e.g. `text/html`), the new content will be evaluated with that stale fallback. In the rare case where the new content cannot be auto-detected, `getPayloadType` would return the stale `mime` instead of the intended default. Adding `mime` to the dependency array (or using the initial `utils.MimeOptions.Default` constant directly as the second argument) prevents this drift.

```suggestion
        const correctMime = getPayloadType(decoded, utils.MimeOptions.Default);
```

---

Path: extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSREncapsulatedPDFReport.tsx
Line: 19

Comment:
**Invalid CSS unit in HTML `height` attribute**

The `height` attribute on `<object>` (and most HTML elements) expects a plain integer (pixels) or a percentage string — not a CSS length with a unit like `rem`. `"500rem"` will be treated as an invalid value by browsers; the effective height will fall back to the browser default, likely making the PDF viewer very short or zero-height. Use a pixel value or control the height via CSS instead.

```suggestion
        <object data={fileURL} width="100%" height="500" type="application/pdf">
```

---

Path: extensions/cornerstone-dicom-sr/src/utils/payload.ts
Line: 35-45

Comment:
**Duplicate entries in `allowedTags` and `nonBooleanAttributes` / `allowedAttributes`**

Many tags (e.g. `h1`–`h4`, `li`, `ul`, `ol`, `table`, `tr`, `td`, `tbody`, `th`, `a`, `img`, `br`, `em`, `strong`, `i`, `sub`, `sup`, `p`) appear two or three times in `allowedTags`. Likewise `style`, `href`, `height`, `width`, `class`, `id`, `src`, `name`, `content`, `colspan`, `rowspan`, `http-equiv` all appear twice in both `nonBooleanAttributes` and the `allowedAttributes['*']` array. While `sanitize-html` deduplicates internally, the redundancy makes this configuration harder to audit for security correctness. Removing duplicates would reduce the risk of unintended entries being missed during future review.

`````

</details>

<sub>Reviews (12): Last reviewed commit: ["fix(dicom-sr) fixed a potential React re..."](https://github.com/ohif/viewers/commit/7395ffd5beb9668c1f63b6d7cfa4d98a309e441f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=23872630)</sub>

### CODE REVIEW

**Comment by luissantosHCIT** [comment]
I am not sure what Cypress is complaining about.
The test suite runs without errors in my local environment and I have not changed the tests at all.

image

**Comment by luissantosHCIT** [comment]
I see lots of Segmentation related errors in Playwright which does not look related to anything I have done so far.

I am thinking the last set of commits I synced from upstream might be faulty or incomplete.

Removing the [WIP] designation since I do not see anything else to do for this PR until further guidance.

**Comment by sedghi** [comment]
Same comment I had in other PR, which is seems like the diff includes other changes. This makes merging your PRs slower, if you can only include changes for SR for this we can review,test faster and merge while waiting for the others to get reivewed

**Comment by wayfarer3130** [comment]
Can you provide an anonymized example of both pdf and html reports which can be added to the automated test data set?  I can get those uploaded.
You must include a notification in that which indicates there is no PHI data in them, and that OHIF has permission to include and distribute them in the viewer test dataset.

File: extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSREncapsulatedReport.tsx [hunk 1]
```diff
@@ -0,0 +1,64 @@
+import React, { useEffect } from 'react';
+import Markdown from 'marked-react';
```
wayfarer3130: Can you do lazy inclusion of the marked-react?  This component isn't used elsewhere so loading it using a dynamice import statement, causing the separation into a separate load container will keep the overall OHIF size a bit smaller.

sedghi: why are you not using react-markdown which is far more utilized in other projects?

https://npmtrends.com/marked-react-vs-react-markdown

extensions/cornerstone-dicom-sr/src/utils/payload.ts
```diff
@@ -0,0 +1,492 @@
...
+ *
+ * This is helpful in deciding how to render and clean up the payload later on.
+ *
+ * @param {string} payload
+ * @param {string} suggested_mime Default MIME to use if we cannot identify the content's MIME
+ * @return string
+ */
+export function getPayloadType(payload: string, suggested_mime: string = 'text/plain') {
+  // PDF
+  if (!payload.indexOf('%PDF-')) {
```
wayfarer3130: Don't you want payload.indexOf('%PDF-') to test if it IS PDF?

extensions/cornerstone-dicom-sr/src/utils/payload.ts
```diff
@@ -0,0 +1,492 @@
...
+ * Given a string payload, attempt to find out the likely MIME associated with it.
+ * This function focuses on the MIMEs we care about for document handling.
+ *
+ * This is helpful in deciding how to render and clean up the payload later on.
+ *
+ * @param {string} payload
+ * @param {string} suggested_mime Default MIME to use if we cannot identify the content's MIME
+ * @return string
+ */
+export function getPayloadType(payload: string, suggested_mime: string = 'text/plain') {
```
wayfarer3130: Please add an options containing the suggestedMime (note lower camel case), and whether it should try scanning for html/pdf.  It shouldn't automatically do it, since the mime type is supposed to be provided in the DICOM itself for encapsulated objects.  You could then set a flag to only do the test if a certain customization was present.  That customization could be a simple test function for the given type.  That way other types can be recognized as needed, and no special types would need to be provided by default.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
```diff
@@ -0,0 +1,492 @@
...
+
+/**
+ * Given a string payload, encapsulate it into a Blob object.
+ * This is used mostly to interface with components expecting blobs from different sources.
+ *
+ * @param {string} data
+ * @param {string} mime MIME to add to Blob so other components can know how to handle contents.
+ * @return Blob
+ */
+export function stringToBlob(data: string, mime: string = payloadMIMEOptions.DEFAULT): Blob {
```
wayfarer3130: This should be a shared external function.  It gets repeated several times in the code so sharing it means any bugs in it could be fixed.

extensions/cornerstone-dicom-sr/src/utils/srInspection.ts
```diff
@@ -0,0 +1,35 @@
...
+  ConceptNameCodeSequenceItem,
+  ContentSequence,
+  ContentSequenceItem,
+  DICOMStandardReport,
+} from './srTypes';
+
+export function getCodeMeaningFromConceptNameCodeSequence(
+  conceptNameCodeSequence: ConceptNameCodeSequence
+): string {
+  let item: ConceptNameCodeSequenceItem = conceptNameCodeSequence[0];
```
wayfarer3130: const

extensions/cornerstone-dicom-sr/src/utils/srInspection.ts
```diff
@@ -0,0 +1,35 @@
...
+  const { CodeMeaning } = item;
+  return CodeMeaning ?? "";
```
wayfarer3130: Why not 
{
   return conceptNameCodeSequence?.[0]?.CodeMeaning || ""
}

Really, all you are doing is extracting path values, and a single extractor is easier to manage.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: Sorry, you are probably asking about a TypeScript specific thing. I am not understanding this request. The function is already exported from the payload module in utils. I thought I had all areas pointing to this function as an import. Is it that I left places accidentally duplicating code?

My brain cleans the code in passes and sometimes I forget to do one last pass.

extensions/cornerstone-dicom-sr/src/utils/srInspection.ts
luissantosHCIT: I just assumed I was done with cleaning up the code after some debugging and before PR which clearly it is not true. I will correct that.

Question: would the "compiler" elide the extra lines? In other words, do TS compilers optimize code (not just minimize) like a C or Rust compiler? I know TS is mostly linters on top of JS, but I am curious how far the compiler goes through and if the extra lines actually have effects on final package. Thank you for educating me on this one.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: I am using the Magic number approach I would do in C specially if avoiding deep reads into data contents for performance reasons. Since we are talking JS here which is many abstraction layers above bare metal, I tend to be even less liberal with potentially costly checks. I am primarily a systems programmer so hopefully you can see a pattern here. 

With that said, perhaps it is best I regex the input if PDF has a magic pattern at end of file so we ignore any potential malicious payloads at the boundaries and then verify here the payload starts with the magic number otherwise ignore it.
There's arguments in either direction, but I would rather detect the payload is a potentially well formed PDF (starts with the correct signature) and then let the browser object actually test if the PDF is valid. That way, I should have guarantees that the browser implementors have better ways to test this while being potentially more performant (the payload would be on the binary side of the JS engine at that point). 

That's my thought process at least.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSREncapsulatedReport.tsx
luissantosHCIT: Thank you. I will read more about this and educate myself and then submit a commit.

luissantosHCIT:
> Can you provide an anonymized example of both pdf and html reports which can be added to the automated test data set? I can get those uploaded. You must include a notification in that which indicates there is no PHI data in them, and that OHIF has permission to include and distribute them in the viewer test dataset.

Let me get back to you on this. I believe I have access to purely fictitious examples that have not patient involvement. How should the notice be provided? What's the expected format so I can ask for proper permission? 

Thank you!

luissantosHCIT:
@wayfarer3130 I apologize for not being able to follow up on my PRs these past few months. Between finding myself in a situation in which I am no longer working under my previous employer and the US government shutdown, I have not been able to revisit these PRs. It does not help that some of the complexity was based on the fact I could not test other features without creating a dependency between branches at the time.

The dust is settling a little bit for me (although the ongoing shutdown does affect me directly), so I hope to return to more involvement in this project.

In terms of this PR, I no longer can request a sample so I will need to set up a prototype PACS in the form of Orthanc + OHIF like in your documentation example and generate a sample that I can vouch for. This will take some time.

Thank you for your patience!

wayfarer3130:
> @wayfarer3130 I apologize for not being able to follow up on my PRs these past few months. Between finding myself in a situation in which I am no longer working under my previous employer and the US government shutdown, I have not been able to revisit these PRs. It does not help that some of the complexity was based on the fact I could not test other features without creating a dependency between branches at the time.
> 
> The dust is settling a little bit for me (although the ongoing shutdown does affect me directly), so I hope to return to more involvement in this project.
> 
> In terms of this PR, I no longer can request a sample so I will need to set up a prototype PACS in the form of Orthanc + OHIF like in your documentation example and generate a sample that I can vouch for. This will take some time.
> 
> Thank you for your patience!

Sorry to hear about your difficulties.  The government shutdown is quite painful.  Could you send me a message on the cornerstone slack?  wayfarer3130 is my account id there.

For this PR, creating a sample could be as easy as using dcmtk to edit a DICOM SR example and just updating one of the text fields with the appropriate data.  If you did that, and send it to me with a statement that you have created it with no PHI and that you give permission to OHIF to distribute, then I can add it to the E2E test dataset and the rendering can be tested with an integration test.

luissantosHCIT:
> > @wayfarer3130 I apologize for not being able to follow up on my PRs these past few months. Between finding myself in a situation in which I am no longer working under my previous employer and the US government shutdown, I have not been able to revisit these PRs. It does not help that some of the complexity was based on the fact I could not test other features without creating a dependency between branches at the time.
> > The dust is settling a little bit for me (although the ongoing shutdown does affect me directly), so I hope to return to more involvement in this project.
> > In terms of this PR, I no longer can request a sample so I will need to set up a prototype PACS in the form of Orthanc + OHIF like in your documentation example and generate a sample that I can vouch for. This will take some time.
> > Thank you for your patience!
> 
> Sorry to hear about your difficulties. The government shutdown is quite painful. Could you send me a message on the cornerstone slack? wayfarer3130 is my account id there.
> 
> For this PR, creating a sample could be as easy as using dcmtk to edit a DICOM SR example and just updating one of the text fields with the appropriate data. If you did that, and send it to me with a statement that you have created it with no PHI and that you give permission to OHIF to distribute, then I can add it to the E2E test dataset and the rendering can be tested with an integration test.

You are right. I will manually craft an example after making a quick python script.

luissantosHCIT:
@wayfarer3130 Thank you for your patience. I finally got around to compiling a few sample reports, which I am adding here. However, since my PR branch is greatly out of sync, I will have to figure out how to reconcile the ~500 commits with my changes here.

[sample_reports.zip](https://github.com/user-attachments/files/25727555/sample_reports.zip)

wayfarer3130:
Did you mean to close it?  You can usually merge using AI pretty effectively without much intervention if any.

luissantosHCIT:
@wayfarer3130 Sorry, I accidentally nuked my copy of the repository but recovered from reflog. Would you be kind to reopen? Thank you!

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: Ok, so re-reading this after a year I think I understand what you are saying. 

This is one of those things in which I am no longer in my ideal test environment. I believe that the encapsulated report is technically treated as a text string (value in `TextValue` SR node). In my use case, I only found html and xml in such node but I added PDF in the event you could include a binary blob in one of these SR nodes that include the PDF. I do not remember the DICOM including an explicit clue as to what the mime is supposed to be so I test the blob here.

I will need to revisit if I find other interesting usages of the SR tree.

For now, this should work against the examples I provided. I will need to test once you include them in the `testdata` before the PR gets merged to make sure everything works as I expect.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: Hmm, I still do not understand this one. Are you asking me to move this function somewhere else? Where were you thinking this function should live?

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: I should have tested against `-1`. The AI analyzer also pointed this one out as well.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSREncapsulatedReport.tsx
luissantosHCIT: I switched to `react-markdown`. The logic here came from another project in which I had created a unified SR report rendering component. I am not sure why I did not see react-markdown over marked-react. 

You are right though, so I switched over.

luissantosHCIT:
Ok, so I added fixes to requests from reviewers including the findings from the AI review.

It's pretty cool, by the way. I like its potential as a basic linter/ pattern reviewer.

luissantosHCIT:
> Ok, so I added fixes to requests from reviewers including the findings from the AI review.
> 
> It's pretty cool, by the way. I like its potential as a basic linter/ pattern reviewer.

P.S. I am not sure how to resolve the Playwright dependency issue since it is related to the lockfile being frozen.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSREncapsulatedReport.tsx
luissantosHCIT: I had to revert my change because `react-markdown`  was causing compilation errors and I could not figure out how to resolve it. Something about `Module "." not found` when react-markdown imported micromark?

I might play with it some more locally but made the current markdown component lazy loaded. Let me know if you encounter the same issue if trying react-markdown.

luissantosHCIT:
@wayfarer3130 Where does the information on `CodeNameCodeSequenceValues` comes from? I am curious because I never understood where an SR node's `CodeValue` comes from and I assumed that it was relative to a manufacturer's internal SR generator tool. Clearly, I was wrong.

For background, it came up when testing this branch manually since it has been a hot minute. I saw that now the SR component was going to the Measurement SR component instead of my Text SR component. I found a minor issue with that code path that I corrected in a commit but it was interesting that an example radiology report was getting flagged  as a `Measurement SR Report` which it isn't. I believe the handling in this project is correct and I would like to learn more about why. I have a feeling a certain manufacturer I mentioned in the `dcmjs` pr was misusing the `CodeValue` which I am not surprised about.

Thank you for any info or pointing me in the correct direction!

extensions/cornerstone-dicom-sr/src/components/OHIFLazyMarkdown.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { lazy, Suspense } from 'react';
+
+// Since I was not sure how to approach lazy loading of components, I consulted with Google Search/ Gemini.
```
wayfarer3130: You can remove this comment - Just the implementation is ok for now, and that is self explanatory.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
```diff
@@ -0,0 +1,495 @@
...
+/**
+ * Set of options to pass through to sanitize-html.
+ *
+ * Over time, we should optimize this structure towards maximum security possible. The idea is to
+ * allow some level of document fidelity as generated by vendors but help guard users from any potential
+ * malicious payloads that might be living in the user's data repository (e.g. VNA).
+ */
+export const htmlSanitizerOptions = {
+  allowVulnerableTags: true, //We block most styles below, but we should allow some styles through for now so documents can display.
+  allowedTags: [
```
wayfarer3130, line 21: Consider a static string for allowedTags with it comma separated, and then a getter method for allowedTags

get allowedTags() {
   return this._allowedTags || (this._allowedTags = allowedTagsComma.split(','));
}

That will be a bit slower first time access, but is much more efficient for startup time.  The number of elements is just about big enough to be worthwhile.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
```diff
@@ -0,0 +1,495 @@
...
+ * This function focuses on the MIMEs we care about for document handling.
+ *
+ * This is helpful in deciding how to render and clean up the payload later on.
+ *
+ * @param {string} payload
+ * @param {string} suggestedMime Default MIME to use if we cannot identify the content's MIME
+ * @return string
+ */
+export function getPayloadType(payload: string, suggestedMime: string = 'text/plain') {
+  // PDF
```
wayfarer3130, line 405: Since you are working with a custom function, it is worth splitting the payload into the first 2k characters, or the whole string if less than, say 32k.
const prefix = payload.length < 25k ? payload : payload.substring(0,2048);

and then test on prefix - otherwise for a large payload, this gets really slow.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
```diff
@@ -0,0 +1,495 @@
...
+  if (payload.indexOf('%PDF-') != -1) {
+    return 'application/pdf';
+  }
+  // HTML.
+  // Credit for validation regex goes to CSᵠ (https://stackoverflow.com/questions/15458876/check-if-a-string-is-html-or-not)
+  if (HTML_REGEX.test(payload)) {
```
wayfarer3130: This would need a test for hte first n characters, and the last n characters.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
wayfarer3130: I'm asking you to move that function into platform/core/src/utilities  - in fact I'm almost sure there already is one in OHIF that you can re-use, so just delete it and re-use the existing one.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
```diff
@@ -0,0 +1,495 @@
...
+ *
+ * @param {string} data HTML input.
+ * @return string
+ */
+export function extractHTMLFromPayload(data: string): string {
+  const results = HTML_EXTRACTION_REGEX.exec(data);
+  if (results && results.length) {
+    return results.shift();
+  }
+  return data;
```
wayfarer3130: The should throw an exception here, as you have already asserted it is html - just return results.shift() without the if condition and let it throw if it isn't passing.

extensions/cornerstone-dicom-sr/src/utils/srInspection.ts
wayfarer3130, line 1: This stuff all belongs in the new @cornerstonejs/metadata module, but that won't get merged until after this does.  Can you add a comment that this should move there once it is available?

luissantosHCIT: Added a comment and also created a ticket to remind me about making this move.

#5906

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: Yeah, reviewing this again, you are right. We probably don't want to pass-through potentially dangerous contents anyways. Thank you for catching that!

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: I see what you mean.

Based on my search, I see all other places where a blob is generated. I will go ahead and consolidate them.

Image

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: Do you know why a JavaScript regex might sporadically fail?

I think I had the guard because of such an issue and I probably forgot (I was working on multiple projects back then).

Image

I know React runs renders multiple times (at least in dev mode). However, I do not know how that interacts with the RegEx engine since the input is the same and ran against it 3 times in my example above and the 2nd attempt is the one yielding a `null` object. 

For now, what I am going to do is attempt the shift(), catch the exception, and then default to an empty string. Let me see if that allows us to block legitimately bad inputs while still being able to display the report (in dev mode at least).

luissantosHCIT: Hmm, defaulting to nothing makes it such that between two similar report entries for my test case only one gets displayed. It's very strange. That issue was not happening with the pass-through. The pass-through was clearly incorrect, but I think this surfaces some weird conflict I do not currently understand.

luissantosHCIT: Nvm, I figured it out.

Because my extraction regex had the `g` flag and React renders twice (in dev mode), calls to `exec` past the first one will make it "fail" by returning null because it creates internal state between calls to track the offset from the last match...

### From Google

```
Global Flag (g) State Retention: When a regex has the global (g) flag, it retains an internal state via its lastIndex property.
If you call exec() repeatedly on the same regex object and input string, it starts searching from the lastIndex of the previous match. Once it has exhausted all matches, the next call will return null and reset the lastIndex to 0.
Solution:
Remove the global (g) flag if you only need the first match.
Declare the RegExp inside a function or loop so a new object is created each time, preventing state retention.
Manually reset myRegex.lastIndex = 0; after each use.
```

Removing the `g` flag is sufficient.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: Did some refactoring of those blobs to point to a new family of functions (`toBlob`, `csvToBlob`, and `b64ToBlob`). Tests passed locally so the refactored is tentatively stabled.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: Built a series of unit tests for the `HTML_REGEX`, `extractHTMLFromPayload`, and `sanitizeHTML`.

In that process, I uncovered many situations in which it failed to work properly including discriminating HTML from PDF tags. Using the tests, I updated the regex.

I will be uploading my changes soon.

luissantosHCIT: Oh, and I uncovered a bug with the sanitizeHTML project in how it handles whitespacing inside a valid tag.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
luissantosHCIT: Would you be kind to elaborate? This is a struct following the sanitizeHTML package expected object syntax. How would the struct have the array initialized by the time it gets passed to the sanitization function?

Thank you!

luissantosHCIT:
@wayfarer3130 I will check tomorrow for whatever issues come up from the PR_CHECKS job.

By any chance, is it important to run `prettier` in this project?

Running it found 180 files with style issues. It's mostly extra commas and spacing, but let me know if I should include a commit fixing them.

:)

luissantosHCIT:
Should I update dependencies outside the context of my changes here (and run tests) or is that something you would prefer doing? What is the process for when the security audit presents multiple packages as needing updates?

Thank you!

luissantosHCIT:
I fixed the other issues that were reported.

Are the sample reports in this PR added to the test data? Anything that I can do to help with that? I would like to add a Playwright test before merging if the test data could be loaded.

luissantosHCIT:
Hey, just letting you know that I tried to resolve the security advisories by updating dependencies locally and I think some of the dependencies still force the installation of vulnerable sub-dependencies like `js-yaml`. I guess I should try again at a later date and pray to the TS gods no more advisories are available and no updates break anything.

If you have a better idea for how to deal with these advisories and pass the `PR_CHECKS`, please let me know. 

P.S. >> I looked at other PRs and the only way for me to resolve this is by restoring the bun lockfile to a prior version so the security audit is `skipped`. Did you intend for the `PR_CHECKS` step to look for high security audits on every PR or only when a new dependency is proposed? Right now it looks to be working like the rather without descriminating whether the offending dependency was one already present in prior "safe" PRs or from the newly added dependency. It feels like it will penalize any attempts at updating the dependencies while allowing other PRs go through and missing out on an opportunity to make a decision on whether attempting to update the dependencies.

Looking forward to understanding what the process is so I can more effectively help manage the security state of the project.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
```diff
@@ -0,0 +1,468 @@
...
+
+/**
+ * Set of options to pass through to sanitize-html.
+ *
+ * Over time, we should optimize this structure towards maximum security possible. The idea is to
+ * allow some level of document fidelity as generated by vendors but help guard users from any potential
+ * malicious payloads that might be living in the user's data repository (e.g. VNA).
+ */
```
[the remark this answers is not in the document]
luissantosHCIT: Hmm, that is an interesting suggestion. I am going to look further into this to figure out the full range of styles I can passthrough. 

A little background here. The changes I originally proposed enables the rendering of encapsulated HTML (primarily) reports as Philips had them in the CareStream PACS. I was hoping to eliminate vectors of vulnerability from arbitrary payloads with `sanitize-html`. However, a full sanitization of the html would strip styling and a few urls which would break document fidelity. At the time, the only way I knew without manually having to specify each style and edge case was by allowing vulnerable tags through. 

I agree with greptile here because I didn't mean for script tags to make it through, but I was prioritizing proper rendering of the document.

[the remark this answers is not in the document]
luissantosHCIT: I ended up removing the `allowVulnerableTags` option but kept the `style` tag as one of the whitelisted tags because some documents do need styles present in the global style tag.

extensions/cornerstone-dicom-sr/src/utils/payload.ts
```diff
@@ -0,0 +1,469 @@
...
+    'th',
+    'h1',
+    'h2',
+    'h3',
+    'h4',
+    'a',
+    'img',
+    'sup',
+    'sub',
+    'style',
```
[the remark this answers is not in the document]
luissantosHCIT: The viewer uses version `2.17.0` and it does not appear to silently remove the style tag. I confirmed by loading the website with no-cache enabled and opened up a report that contains the `<style>` tag without issues. I inspected the injected HTML and the style tag was present in the `<head>` section.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSREncapsulatedReport.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <blockquote className="invert [&_img]:invert">
+          <div dangerouslySetInnerHTML={{ __html: sanitizeHtml(textContent) }} />
+        </blockquote>
+      );
+    case utils.MimeOptions.Pdf:
+      return (
+        <blockquote>
+          <OHIFCornerstoneSREncapsulatedPDFReport content={data} />
+        </blockquote>
+      );
```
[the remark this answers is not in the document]
luissantosHCIT: It is best to make this change at the time we identify the correct mime during the previous `useEffect`. That's what I did. Otherwise, good suggestion.

PR #5378, opened 2025-09-01
feat(ErrorBoundary): Fixing console errors and Improving error logging for better developer experience

ViniciusResende:
### Context

I was having a bad experience while trying to find the source of errors I was facing, not being able to see the errors at the console, and being misguided by errors in the console that had no relation with the issue itself. The changes on this PR solved both of this problems.

### Changes & Results

- a0f9bc8203a2d029df4ad20e4e7aa825f38af1c5 - This fix removed the following error to be shown in the console while rendering the error dialog, which was misguiding me while searching for the actual error.

**Previous:**
image

**After:**
(No error)

- 01e5c43f607bb002e5019f0a3e4b82fdf16db338 - This fix removed the following error to be shown while opening the error dialog after clicking the "Show details" button, also was misguiding since it is not related to the issue itself.

**Previous:**
image

**After:**
(No error)

- d5b7a04e944dfd59d9d39f3145c290c4e56ed223 - This removes the prevent default behavior that was making the error that generated the error screen itself, to not be printed to the console.

**Previous:**
(No error)

**After (example):**
image

> Note that there are two huge wins on having this error being displayed at the console here. 
The first one one of them is the usage of the code source maps, I couldn't find a way to use a JavaScript API to display the mapped files in the Dialog UI without a huge change. But, by having the error in the console, the browser automatically uses the source maps and we can see the original file and line where the error happened, in this example is "ToolbarService.ts:664:1". Previously I would have just the Dialog UI file, that says that the file where the error happened is "http://localhost:3000/static/js/index.js:18727:40", which is not of much help since it points to the "compiled" file, and not the original one.
The second one is the integration with the browser's DevTools, with the error in the console I can click on the file and go directly to the file in DevTools, putting a breakpoint on it to further investigate what was the issue, allowing me to go through the error stack in real time. 

### Testing

Intentionally create a issue that will throw an unhandled error in OHIF, e.g. add a not registered tool at modes/longitudinal/src/index.ts:133 into the `toolbarService.updateSection` second parameter array, such as:
```ts
toolbarService.updateSection('MeasurementTools', [
    'Length',
    'Bidirectional',
    'ArrowAnnotate',
    'NotRegistered', // <--- will cause error
    'EllipticalROI',
    'RectangleROI',
    'CircleROI',
    'PlanarFreehandROI',
    'SplineROI',
    'LivewireContour',
]);
```

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] Node version: v20.9.0

### DIFF
--- a/platform/ui-next/src/components/Button/Button.tsx
+++ b/platform/ui-next/src/components/Button/Button.tsx
@@ -39,13 +39,13 @@ export interface ButtonProps
 }
 
 const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
-  ({ className, variant, size, asChild = false, ...props }, forwardRef) => {
+  ({ className, variant, size, asChild = false, dataCY, ...props }, forwardRef) => {
     const Comp = asChild ? Slot : 'button';
-    const dataCY = props.dataCY || `${props.name}-btn`;
+    const testId = dataCY || `${props.name}-btn`;
 
     return (
       <Comp
-        data-cy={dataCY}
+        data-cy={testId}
         className={cn(buttonVariants({ variant, size }), className)}
         ref={forwardRef}
         {...props}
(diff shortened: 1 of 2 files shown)


sedghi (commented):
can you update the images in pr description, they are not showing up

ViniciusResende:
@sedghi 

> can you update the images in pr description, they are not showing up

Can you see if they're showing now? I've reset them and checked on an anonymous window and it worked.

sedghi (approved):

PR #5381, opened 2025-09-03
feat: Add command to load derived display sets

igoroctaviano:
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -41,7 +41,7 @@
     },
     "addOns/externals/devDependencies": {
       "name": "@externals/devDependencies",
-      "version": "3.12.0-beta.18",
+      "version": "3.12.0-beta.22",
       "dependencies": {
         "@babel/runtime": "^7.20.13",
         "@kitware/vtk.js": "32.12.0",
@@ -126,14 +126,14 @@
     },
     "addOns/externals/dicom-microscopy-viewer": {
       "name": "@externals/dicom-microscopy-viewer",
-      "version": "3.12.0-beta.18",
+      "version": "3.12.0-beta.22",
       "dependencies": {
         "dicom-microscopy-viewer": "^0.48.6",
       },
     },
     "extensions/cornerstone": {
       "name": "@ohif/extension-cornerstone",
-      "version": "3.12.0-beta.18",
+      "version": "3.12.0-beta.22",
       "dependencies": {
         "@babel/runtime": "^7.20.13",
         "@cornerstonejs/adapters": "^4.0.0",
@@ -159,8 +159,8 @@
         "@cornerstonejs/codec-openjpeg": "^1.2.4",
         "@cornerstonejs/codec-openjph": "^2.4.5",
         "@cornerstonejs/dicom-image-loader": "^4.0.0",
-        "@ohif/core": "3.12.0-beta.18",
-        "@ohif/ui": "3.12.0-beta.18",
+        "@ohif/core": "3.12.0-beta.22",
+        "@ohif/ui": "3.12.0-beta.22",
         "dcmjs": "0.43.1",
         "dicom-parser": "^1.8.21",
         "hammerjs": "^2.0.8",
@@ -172,7 +172,7 @@
     },
     "extensions/cornerstone-dicom-pmap": {
       "name": "@ohif/extension-cornerstone-dicom-pmap",
-      "version": "3.12.0-beta.18",
+      "version": "3.12.0-beta.22",
       "dependencies": {
         "@babel/runtime": "^7.20.13",
         "@cornerstonejs/adapters": "^4.0.0",
@@ -195,7 +195,7 @@
     },
     "extensions/cornerstone-dicom-rt": {
       "name": "@ohif/extension-cornerstone-dicom-rt",
-      "version": "3.12.0-beta.18",
+      "version": "3.12.0-beta.22",
       "dependencies": {
         "@babel/runtime": "^7.20.13",
         "react-color": "^2.19.3",
@@ -215,7 +215,7 @@
     },
     "extensions/cornerstone-dicom-seg": {
       "name": "@ohif/extension-cornerstone-dicom-seg",
-      "version": "3.12.0-beta.18",
+      "version": "3.12.0-beta.22",
       "dependencies": {
         "@babel/runtime": "^7.20.13",
         "@cornerstonejs/adapters": "^4.0.0",
@@ -238,7 +238,7 @@
     },
     "extensions/cornerstone-dicom-sr": {
       "name": "@ohif/extension-cornerstone-dicom-sr",
-      "version": "3.12.0-beta.18",
+      "version": "3.12.0-beta.22",
       "dependencies": {
         "@babel/runtime": "^7.20.13",
         "@cornerstonejs/adapters": "^4.0.0",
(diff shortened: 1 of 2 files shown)


extensions/cornerstone/src/commandsModule.ts
```diff
@@ -145,6 +146,45 @@ function commandsModule({
...
+        const currentDs = displaySetService.getDisplaySetByUID(displaySetUID);
+        const displaySetCache = displaySetService.getDisplaySetCache();
+        return Array.from(displaySetCache.values()).filter(
+          (ds: any): ds is DisplaySet =>
+            (ds?.referencedDisplaySetInstanceUID === displaySetUID ||
+              ds.referencedSeriesInstanceUID === currentDs.SeriesInstanceUID) &&
+            !ds?.isHydrated
+        );
+      };
+
```
sedghi: I don't see any logic to check if this is derived. I think derived DS have isDerived on them or similar

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -145,6 +146,45 @@ function commandsModule({
...
+        const displaySet = displaySetService.getDisplaySetByUID(displaySetInstanceUID);
+        if (displaySet.Modality === 'SEG' || displaySet.Modality === 'RTSTRUCT') {
+          const headers = userAuthenticationService.getAuthorizationHeader();
+          await displaySet.load({ headers }).catch(err => {
+            console.warn(`Failed to load display set ${displaySet.displaySetInstanceUID}:`, err);
+          });
+          segmentationService.addSegmentationRepresentation(activeViewportId, {
+            segmentationId: displaySet.displaySetInstanceUID,
+            type:
+              displaySet.Modality === 'SEG'
```
sedghi: we like to think about Segmentation as a general term that includes Contour, Labelmap or Meshes. And each of them are segmentation representation. 
I think this function should be called `loadAvailableLabelmapsForActiveViewport` maybe?

bun.lock
```diff
@@ -126,14 +126,14 @@
...
-      "version": "3.12.0-beta.18",
+      "version": "3.12.0-beta.23",
       "dependencies": {
         "dicom-microscopy-viewer": "^0.48.6",
       },
     },
     "extensions/cornerstone": {
       "name": "@ohif/extension-cornerstone",
-      "version": "3.12.0-beta.18",
+      "version": "3.12.0-beta.23",
```
sedghi: no sure why bun lock is changing but not yarn lock

extensions/cornerstone/src/commandsModule.ts
igoroctaviano: But is mesh a representation today? I thought it was just contour and labelmap. Surface generation occurs even without explicit requests, and the mesh is loaded through geometry calls (no mesh/STL code in ohif today). 
I agree to changing the name. Maybe loadLabelmapsAndContoursForActiveViewport?

bun.lock
igoroctaviano: Are we using both? I find it very confusing that we use yarn and bun today. Maintaining two lock files seems error-prone.

extensions/cornerstone/src/commandsModule.ts
igoroctaviano: @sedghi i'll need some input about this question above

bun.lock
igoroctaviano: @sedghi

sedghi: Ask @jbocce please

jbocce: @igoroctaviano 

So yes we are currently maintaining both `bun.lock` and `yarn.lock`. I don't know why - it predates my current stint on OHIF.

In this case I'm not sure why the OHIF dependencies were updated. This sometimes occurs when someone branches off `master` prior to the  `Update package versions [skip ci]` step running. In this particular case, I think simply updating from master again should remove this from the PR's list of changed files. If there is a conflict be sure to take the `bun.lock` file from `master`. Please merge from master again and see what happens.

In any case the changes I see here to the `bun.lock` file are benign.

Does this make sense (i.e. capisce?) 😜

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -145,7 +147,86 @@ function commandsModule({
...
+        viewportGridService.getDisplaySetsUIDsForViewport(activeViewportId);
+      if (!displaySetInstanceUIDs?.length) {
+        console.warn('No display sets found for active viewport');
+        return;
+      }
+
+      const primaryDisplaySetUID = displaySetInstanceUIDs[0];
+      const derivedDisplaySets = _getDerivedSegmentations(primaryDisplaySetUID);
+      if (!derivedDisplaySets.length) {
+        return;
```
jbocce: Should we do a `console.warn` here too?

jbocce (approved):
Minor comment

fedorov:
Can someone clarify what needs to be happen to merge this? @igoroctaviano

platform/docs/versioned_docs/version-3.11/platform/extensions/modules/hpModule.md
```diff
@@ -600,6 +600,17 @@ The `onLayoutChange` callback is executed before the layout change is started. Y
 
 The `onViewportDataInitialized` callback is executed after the initial viewport grid data is set and all viewport data includes a designated display set. This callback runs during the initial layout setup for each stage. You can use it to perform actions or apply settings to the viewports at the start.
 
+### `onViewportDataChanged`
```
sedghi: no need to touch versioned docs

sedghi:
I understand your idea, but I'm not convinced it's the best way forward anymore. We've got new plans for the `viewportGrid` state. It starts with this PR: https://github.com/OHIF/Viewers/pull/5624, which aims to fix timing issues where the state updates but the event doesn't reflect it. You should check out @wayfarer3130's comments on that PR; it shows the right approach. We should refactor the `viewportGrid` to separate the layout logic from the individual viewport data. This will give us the hooks needed to run callbacks assigned by the Hanging Protocol. Let me know if you wanna meet with me and Bill to go over this

igoroctaviano:
> I understand your idea, but I'm not convinced it's the best way forward anymore. We've got new plans for the `viewportGrid` state. It starts with this PR: #5624, which aims to fix timing issues where the state updates but the event doesn't reflect it. You should check out @wayfarer3130's comments on that PR; it shows the right approach. We should refactor the `viewportGrid` to separate the layout logic from the individual viewport data. This will give us the hooks needed to run callbacks assigned by the Hanging Protocol. Let me know if you wanna meet with me and Bill to go over this

This is not just about the callback. I'm happy to remove the callback part of it.

wayfarer3130:
> > I understand your idea, but I'm not convinced it's the best way forward anymore. We've got new plans for the `viewportGrid` state. It starts with this PR: #5624, which aims to fix timing issues where the state updates but the event doesn't reflect it. You should check out @wayfarer3130's comments on that PR; it shows the right approach. We should refactor the `viewportGrid` to separate the layout logic from the individual viewport data. This will give us the hooks needed to run callbacks assigned by the Hanging Protocol. Let me know if you wanna meet with me and Bill to go over this
> 
> This is not just about the callback. I'm happy to remove the callback part of it.

The fact that there are load commands for various things actually fits into the new framework fairly well because it means the load action has already gotten isolated to the command module.  That will make things more reliable and break fewer things going forward.  I'm hoping that the hanging protocol methods will be backwards compatible as well for some length of time as we switch to a better design around the viewport grid.  That way people can continue using the commands to allow loading/specifying applicability, and the commands can be enhanced in the future to add more options to allow finer control over viewport setup related to loading of data.  Ideally the existing sr/seg viewports should be switched to use the commands, and then once the viewport grid is ready, we only have to fix the commands in one spot to fix loading generally.

lisacico4-hash (approved):

Issue #5192, opened 2025-07-10
[Bug] Incorrect PROXY_TARGET in dev:orthanc Script Causes Proxy Failure


AdilZahid:
### Describe the Bug

When running the OHIF Viewer in development mode using the default dev:orthanc script, the app fails to connect to the Orthanc DICOMweb data source. The UI displays a "Data Source Connection Error." and cannot retrieve studies from Orthanc, even though Orthanc is running and accessible directly.
The root cause is that the PROXY_TARGET in the dev:orthanc script is set to a full URL (http://localhost:3000/pacs/dicom-web), which does not match the intended proxy context and results in the dev server proxy not forwarding requests to Orthanc.

Checked this too:
https://community.ohif.org/t/how-to-solve-this-error-data-source-connection-error-error-request-failed-please-ensure-the-following-data-source-is-configured-correctly-or-is-running-orthanc-server/1750/2

### Steps to Reproduce

Followed this guide.
https://docs.ohif.org/development/getting-started

1. Start Orthanc `yarn run orthanc:up` and ensure it is accessible at http://localhost:8042/dicom-web/studies.
2. Run the dev server: `yarn run dev:orthanc`
3. Open the OHIF Viewer in your browser at http://localhost:3000.
    Observe the error message: "Data Source Connection Error."

### The current behavior

1. The OHIF Viewer displays "Data Source Connection Error."
2. Requests to /pacs/dicom-web/studies are not proxied to Orthanc and do not return DICOM data.

### The expected behavior

1. The OHIF Viewer should connect to Orthanc and display available studies.
2. Requests to /pacs/dicom-web/studies should be proxied to http://localhost:8042/dicom-web/studies and return DICOM study data.
3. The dev server should use a path for PROXY_TARGET (e.g., /pacs/dicom-web), not a full URL.

Code from webpack.pwa.js:
if (hasProxy) {
  mergedConfig.devServer.proxy = [
    {
      context: [PROXY_TARGET],  // This should be a path, not a full URL
      target: PROXY_DOMAIN,     // This is the target server
      changeOrigin: true,
      pathRewrite: {
        [`^${PROXY_PATH_REWRITE_FROM}`]: PROXY_PATH_REWRITE_TO,
      },
    },
  ];
}

### OS

Ubuntu

### Node version

18.18.0

### Browser

Chrome

AdilZahid:
Solution:
Edit your platform/app/package.json file and replace the dev:orthanc script with the following line:

[...]

PR #5389, opened 2025-09-05
Fixed OHIF integration.

michaelonken:
PR for the issue reported in https://github.com/OHIF/Viewers/issues/5192 and using the proposed solution. This also fixed the problem for me.

Note that I also adapted the entry for "dev:orthanc:no:cache" but that "target" does not work for me (or I used it incorrectly since I'm not familiar with how everything plays together).


sedghi:
@igoroctaviano

PR #5391, opened 2025-09-08
[DRAFT] POC multi and single tab segmentation panel

jbocce:
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone-dicom-seg/src/getToolbarModule.ts
```diff
@@ -33,6 +65,21 @@ export function getToolbarModule({ servicesManager }: withAppTypes) {
           };
         }
 
+        const activeRepresentations = segmentationService.getSegmentationRepresentations(
```
jbocce, line 50: Disable various segmentation tools and utilities based on the type of the active segmentation.

extensions/cornerstone-dicom-seg/src/getToolbarModule.ts
```diff
@@ -33,6 +65,21 @@ export function getToolbarModule({ servicesManager }: withAppTypes) {
...
+          {
+            segmentationId: activeSegmentation.segmentationId,
+            type: segmentationRepresentationType,
+          }
+        );
+        if (!activeRepresentations?.length) {
+          return {
+            disabled: true,
+            disabledText: `Active segmentation is not a ${segmentationRepresentationType} segmentation`,
+            visible: customizationService.getCustomization('panelSegmentation.isMultiTab'),
```
jbocce, line 64: This little trick allows us to show ALL the tools and utilities disabled in multi-tab mode whenever the active segmentation is for the "other" tab type. In single tab mode, we only want to show the tools and utiliies for the active type.

extensions/cornerstone-dicom-seg/src/init.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ *
+ * @param {Object} servicesManager
+ * @param {Object} configuration
+ */
+export default function init({ servicesManager }: withAppTypes): void {
+  const { segmentationService, toolbarService } = servicesManager.services;
+
+  toolbarService.registerEventForToolbarUpdate(segmentationService, [
```
jbocce: The toolbars need updating now when segmentations change.

extensions/cornerstone-dicom-seg/src/getToolbarModule.ts
```diff
@@ -10,9 +11,40 @@ export function getToolbarModule({ servicesManager }: withAppTypes) {
         };
       },
     },
+    {
+      name: 'evaluate.cornerstone.hasSegmentationOfType',
```
jbocce, line 15: I am using this for the AI assist utility since I think it only applies to lable maps. Did I do the right thing?

extensions/default/src/Components/SidePanelWithServices.tsx
```diff
@@ -37,9 +37,18 @@ const SidePanelWithServices = ({
...
 
-  const handleActiveTabIndexChange = useCallback(({ activeTabIndex }) => {
-    setActiveTabIndex(activeTabIndex);
-  }, []);
+  const handleActiveTabIndexChange = useCallback(
+    ({ activeTabIndex }) => {
+      // This is likely breaking an encapsulation rule. However, the PanelService
+      // does NOT keep track of the active tab index for each side panel, so
+      // the event is broadcast here.
+      panelService._broadcastEvent(panelService.EVENTS.PANEL_ACTIVATED, {
```
jbocce: Am I going to go to hell for this? 🤣

platform/i18n/src/locales/en-US/SegmentationTable.json
```diff
@@ -3,16 +3,24 @@
...
+  "Contour tools": "Contour tools",
+  "Contour Segmentations": "Contour segmentations",
   "Delete": "Delete",
   "Display inactive segmentations": "Display inactive segmentations",
   "Export DICOM SEG": "Export DICOM SEG",
   "Download DICOM SEG": "Download DICOM SEG",
   "Download DICOM RTSTRUCT": "Download DICOM RTSTRUCT",
   "Fill": "Fill",
   "Inactive segmentations": "Inactive segmentations",
+  "Labelmap": "Label map",
```
sedghi, line 16: I think labelmap is used as one word, I might be wrong

extensions/cornerstone-dicom-seg/src/init.ts
sedghi: maybe we put it at cornerstone extension not dicom seg?

extensions/default/src/Components/SidePanelWithServices.tsx
sedghi: i see you are creating a new even in panel service just for here to broadcast and elsewhere to subscribe for toolbar service change. my question is why don't we here just update toolbar service directly

modes/segmentation/src/index.tsx
```diff
@@ -118,18 +122,42 @@ function modeFactory({ modeConfiguration }) {
...
+        'ContourTools',
+      ]);
+
       toolbarService.updateSection('SegmentationUtilities', [
         'LabelmapSlicePropagation',
         'InterpolateLabelmap',
         'SegmentBidirectional',
         'SegmentLabelTool',
       ]);
+      toolbarService.updateSection('LabelMapUtilities', [
```
sedghi, line 162: please add a comment here about what we have in mind here in long run, since one might ask why we have SegmentationUtilities and Labelmaputils and contour utils etc etc

modes/segmentation/src/index.tsx
```diff
@@ -118,18 +122,42 @@ function modeFactory({ modeConfiguration }) {
...
+      ]);
+      toolbarService.updateSection('ContourUtilities', ['SegmentLabelTool']);
+
       toolbarService.updateSection('SegmentationTools', [
         'BrushTools',
         'MarkerLabelmap',
         'RegionSegmentPlus',
         'Shapes',
       ]);
+      toolbarService.updateSection('LabelMapTools', [
```
sedghi: same here

modes/segmentation/src/toolbarButtons.ts
```diff
@@ -172,14 +173,42 @@ const toolbarButtons: Button[] = [
...
   },
   {
     id: 'SegmentationTools',
-    uiType: 'ohif.toolBoxButton',
+    uiType: 'ohif.toolBoxButtonGroup',
+    props: {
+      buttonSection: true,
+    },
+  },
+  {
```
sedghi, line 202: i don't think we need separate sections for labelmap/contour tools and labelmap/contour utils. Both can use the segmentation tools/utils and make themselves visible or not based on the segmentatino type no? I'm confused why we need new sections while we can dynamically update tools/utils

modes/segmentation/src/toolbarButtons.ts
```diff
@@ -404,6 +433,7 @@ const toolbarButtons: Button[] = [
       evaluate: {
         name: 'evaluate.cornerstone.segmentation',
         toolNames: ['CircularBrush', 'SphereBrush'],
+        segmentationRepresentationType: SegmentationRepresentations.Labelmap,
```
sedghi: this should be in a new evaluator right? the hasSegmentationOfType, since evals are chained together, i think it should be its own eval

extensions/cornerstone-dicom-seg/src/getToolbarModule.ts
```diff
@@ -33,6 +65,24 @@ export function getToolbarModule({ servicesManager }: withAppTypes) {
...
         }
 
+        const activeRepresentations = segmentationService.getSegmentationRepresentations(
+          viewportId,
+          {
+            segmentationId: activeSegmentation.segmentationId,
+            type: segmentationRepresentationType,
+          }
+        );
+        if (!activeRepresentations?.length) {
```
sedghi, line 57: i prefer separate eval for segmentation type (the one you have in the top)

extensions/cornerstone/src/panels/PanelSegmentation.tsx
```diff
@@ -145,17 +168,41 @@ export default function PanelSegmentation({ children }: withAppTypes) {
...
     onSegmentationAdd,
     showAddSegment,
     renderInactiveSegmentations: handlers.getRenderInactiveSegmentations(),
+    segmentationRepresentationType,
+    selectedSegmentationIdForType:
+      segmentationRepresentationType !== undefined
+        ? selectedSegmentationIdByViewportAndType
+            .get(activeViewportId)
+            ?.get(segmentationRepresentationType)
+        : activeSegmentationInfo?.segmentation.segmentationId,
```
sedghi: why do we need selectedSegmentationIdForType, shouldn't it always show the acgtive segmentation for the viewport? This is really tricky, whenever the seg is shown in the multi tab means we should actually change the active segmentation as well when we switch tab (that is another reason why i hate the multi tab design)

sedghi (commented):
not sure about selectedSegmentationIdForType
i think rest looks good see my comments

extensions/default/src/utils/Toolbox.tsx
```diff
@@ -106,7 +106,11 @@ export function Toolbox({ buttonSectionId, title }: { buttonSectionId: string; t
               className="bg-muted flex flex-wrap space-x-2 py-2 px-1"
             >
               {buttons.map(tool => {
-                if (!tool) {
+                // Skip over tools that are not visible. The visible flag is typically set to
+                // false as a result of the evaluator function. The evaluator might explicitly
+                // set visible to false. Alternatively, the ToolbarService will set the visible flag to
+                // false when the evaluator sets disabled to true and the tool has the hideWhenDisabled flag set to true.
+                if (!tool || !tool.componentProps.visible) {
```
jbocce, line 113: FYI... this also fixes an issue whereby setting the `hideWhenDisabled` property for a button was being ignored for buttons in an embedded section of the Toolbox.

extensions/cornerstone-dicom-seg/src/init.ts
jbocce: Sure NP.

extensions/default/src/Components/SidePanelWithServices.tsx
jbocce: @sedghi do you mean do something like this instead in the callback?

```
      const { activeViewportId: viewportId } = viewportGridService.getState();
      toolbarService.refreshToolbarState({ viewportId });
```

sedghi: something like that

platform/i18n/src/locales/en-US/SegmentationTable.json
jbocce: Interesting... DICOM uses [label map](https://dicom.nema.org/dicom/news/June2024/docs/sups/sup243-slides.pdf). However something like 3D Slicer refers to them as [labelmap](https://slicer.readthedocs.io/en/latest/user_guide/modules/segmentations.html). @sedghi I will let you choose 😉

sedghi: I didn't know that in dicom, let's use label map

extensions/cornerstone/src/panels/PanelSegmentation.tsx
jbocce: The `selectedSegmentationIdForType` is precisely for the fact that the current implementation (in this PR) does NOT automatically activate the segmentation of that tab's type when switching to that tab. From my understanding with @dan-rukas, this behaviour is still being discussed. So until it is decided I will leave this here, but we can revisit it later.

modes/segmentation/src/toolbarButtons.ts
jbocce: It can be separate. However it is NOT the `hasSegmentationOfType`. That one checks if there is at least one segmentation of the given type. It would have to be a new one that checks if the active segmentation is of the given type (i.e. `isActiveSegmentationOfType`. I'm not sure if it deserves its own evaluator, but if you think so I will do it.

extensions/cornerstone-dicom-seg/src/getToolbarModule.ts
jbocce: Sure.

modes/segmentation/src/toolbarButtons.ts
jbocce: There are two reasons why there are separate sections for the label maps and contours.

1. The first reason again concerns the issue of whether or not to activate a segmentation automatically when changing tabs. If we don't activate when changing tabs then the wrong tools will be shown on a tab if they are combined (i.e. the tools of the other/active tab will be shown)
2. The second reason is that when there are zero segmentations (i.e. on initial load), ALL the tools/utils for both types would be combined and shown as disabled in each tab (i.e. like it currently does for the single tab). I would say this is NOT desirable. I suppose the alternative behaviour here would be to show NO/ZERO tools/utils when there are no segmentations - I personally don't like that.

There is something I just thought of that would suggest we might NOT want to activate automatically when switching tabs. I noticed in OHIF master when using the `panelSegmentation.tableMode` set to `expanded`, expanding and collapsing segmentations does NOT automatically change the active segmentation. See the below screen shot as an example. The contour is the active segmentation even though it is collapsed. Note also how misleading it is having the other segmentations indicating a selected segment! So likewise I would NOT expect changing tabs to change the active segmentation.

image

jbocce:
> not sure about selectedSegmentationIdForType i think rest looks good see my comments

Thanks. See my updates and responses.

PR #5396, opened 2025-09-10
fix(SegmentationService): _getSegmentCenter wrongly handling seg center availability

ViniciusResende:
### Context

The PR #5198 introduced some small changes that were needed to prevent an error to be thrown when creating new segmentations. The center wasn't available and the incorrect handling of it was causing an error to be experienced according to Rodrigo Basilio([rodrigo.basilio@radicalimaging.com](mailto:rodrigo.basilio@radicalimaging.com)), which I've personally reached to get better context of how to reproduce the issue.

The problem is that the logic implementation had a small flaw on it, the `!cachedStats?.center || !cachedStats?.namedStats?.center` piece. Step by step the problem was that, the `||` makes it so that, if one of the sides is true, the if condition will be matched and therefore the function will exit without returning the center. The problem is that, since we have the `!` we're verifying that the `cachedStats?.center` or `cachedStats?.namedStats?.center` is undefined, which means that if one of them is undefined the if condition will be matched. But, the truth is that if any of them is present, we want to return it, and **`only if both are undefined`** then we return undefined.

### Changes & Results

The logic that was previously implemented would require the SEG to have the center of it present both at `cachedStats.center` and `cachedStats.namedStats.center`, if not it would return undefined and prevent functions that depend on it to work properly. For exemple, the jumpToSegmentCenter one.

This PR changes the logic of it to return the center if any of the both is present, I also took the opportunity to add some better typing to the function without making much changes on interfaces files. 

### Testing

I wasn't able to load the [1.3.6.1.4.1.32722.99.99.239341353911714368772597187099978969331](http://localhost:3000/viewer?StudyInstanceUIDs=1.3.6.1.4.1.32722.99.99.239341353911714368772597187099978969331) and click on it to be navigated to the center of the segmentation on the latest OHIF local version. With the changes I'm now able to perform it normally.

I've reached to Rodrigo to understand the issue at #5198 and guarantee that they wouldn't be reproducible after my PR changes, I wasn't able to see any errors while creating new segmentations and navigating through them. Also, my PR keeps the logic added by Rodrigo, just re-structure it and fix the small logic flaw.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] Node version: v20.9.0


jbocce (commented):
@ViniciusResende the changes look great and make sense. Please update your branch with master and let's see if the tests all pass. They should with this change!

ViniciusResende:
@jbocce I see that the same playwright test is still failing.

I'd guess that this could be due to a difference in the timing of the animation, looking at the actual and the expected I saw that the animation was in its peak on the expected and in the actual it is not.

Although that with these changes I'm able to see the animation normally when interacting with the same study of the test. Don't know if the code changed the timing of the animation. How should we proceed?

jbocce:
> @jbocce I see that the same playwright test is still failing.
> 
> I'd guess that this could be due to a difference in the timing of the animation, looking at the actual and the expected I saw that the animation was in its peak on the expected and in the actual it is not.
> 
> Although that with these changes I'm able to see the animation normally when interacting with the same study of the test. Don't know if the code changed the timing of the animation. How should we proceed?

@ViniciusResende I will have a look now.

jbocce:
> > @jbocce I see that the same playwright test is still failing.
> > I'd guess that this could be due to a difference in the timing of the animation, looking at the actual and the expected I saw that the animation was in its peak on the expected and in the actual it is not.
> > Although that with these changes I'm able to see the animation normally when interacting with the same study of the test. Don't know if the code changed the timing of the animation. How should we proceed?
> 
> @ViniciusResende I will have a look now.

@ViniciusResende I believe the failure is now a result of the changes made in CS3D regarding camera FOV. So the thing to do here is to update the screenshot for that test. Could you please do that? If you need a hand with that just reach out to me!

Thanks so much for your efforts on this.

PR #5433, opened 2025-09-25
 International support methods - 国际化支持方式

hns1971:
# Description
This pull request aims to improve the project's internationalization (i18n) support, ensuring that all previously unsupported components are now internationalized.

# Changes
- Component Internationalization: Internationalize static text in core components, replacing hard-coded text with the $t('key') method.
 
# 描述

本 PR 旨在修改项目的国际化 (i18n) 支持，使未支持国际化的部分，都能支持国际化。

# 修改内容
​- ​组件国际化​：对核心组件中的静态文本进行了国际化改造，使用 $t('key')方法替换硬编码文本。

### DIFF
--- a/extensions/default/src/Toolbar/ToolButtonListWrapper.tsx
+++ b/extensions/default/src/Toolbar/ToolButtonListWrapper.tsx
@@ -8,6 +8,7 @@ import {
   ToolButtonListDivider,
 } from '@ohif/ui-next';
 import { useToolbar } from '@ohif/core/src';
+import { useTranslation } from 'react-i18next';
 
 interface ToolButtonListWrapperProps {
   buttonSection: string;
@@ -22,6 +23,7 @@ interface ToolButtonListWrapperProps {
  * // test
  */
 export default function ToolButtonListWrapper({ buttonSection, id }: ToolButtonListWrapperProps) {
+  const { t } = useTranslation();
   const { onInteraction, toolbarButtons } = useToolbar({
     buttonSection,
   });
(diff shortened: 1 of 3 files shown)


jbocce:
Looks like there are conflicts on this branch. Please fix them. Thanks.

modes/segmentation/src/toolbarButtons.ts
```diff
@@ -210,7 +210,7 @@ const toolbarButtons: Button[] = [
     uiType: 'ohif.toolButton',
     props: {
       icon: 'tool-move',
-      label: 'Pan',
+      label: 'Buttons:Pan',
```
jbocce, line 213: This is not a bad approach, however we currently are using i18n right in the various toolbarButton files. I think once you update your branch you will see what I mean. Or see this: https://github.com/OHIF/Viewers/blob/f0e9865ca19fe2e744f294f9cc49b62b65d87b95/modes/segmentation/src/toolbarButtons.ts#L34

extensions/default/src/Toolbar/ToolButtonListWrapper.tsx
jbocce, line 1: These changes will no longer be needed once you switch to the standard way we do i18n. Thanks.

pedrokohler:
@jbocce not my business, but shouldn’t commit messages be in English for readability?

Also, having the PR description in English would be great so anyone can reference it later. I’ve lost count of how many times I had to dig through old PRs just to understand what was going on while fixing or investigating an issue.

PR #5435, opened 2025-09-25
feat: Add inherit from basic/longitudinal modes

wayfarer3130:
### Context

The goal for this feature is to make the modes more extensible.  To demonstrate how to do this, I created a basic mode (non-longitudinal), and then updated the longitudinal mode to depend on it.  This reduces the size of the longitudinal mode significantly, and makes re-use possible much more consistently.

### Changes & Results

Move functions/data for modes into exported functionality which can be re-used.
Create a basic mode with the basic functionality
Re-use the basic mode to define the longitudinal mode.

### Testing

Basic integration tests should cover this fairly well

Also, test that the "hidden" mode "@ohif/mode-basic" is defined and available by going to a longitudinal viewer mode,
and replacing /viewer with /basic, seeing that the annotations no longer use the tracked mode.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


modes/longitudinal/src/index.ts
wayfarer3130, line 1: See how much one can restrict the changes to just what is unique in the given mode - the file is far shorter, and doesn't redeclare methods that can be shared.

platform/app/src/routes/WorkList/WorkList.tsx
```diff
@@ -377,7 +377,7 @@ function WorkList({
         >
           <div className="flex flex-row gap-2">
             {(appConfig.groupEnabledModesFirst
-              ? appConfig.loadedModes.sort((a, b) => {
+              ? appConfig.loadedModes.filter(m => m.hide!==true).sort((a, b) => {
```
wayfarer3130: This allows including modes, but not listing them, so that things like test modes can be includes but hidden.

modes/basic/src/index.tsx
wayfarer3130, line 1: This file is more or less the longitudinal mode, but extracted as re-useable functions.

modes/basic/src/index.tsx
```diff
@@ -0,0 +1,347 @@
...
+      dicomsr.sopClassHandler,
+      dicomRT.sopClassHandler,
+    ];
+
+export function isValidMode({ modalities }) {
+      const modalities_list = modalities.split('\\');
+
+      // Exclude non-image modalities
+      return {
+        valid: !!modalities_list.filter(modality => this.nonModeModalities.indexOf(modality) === -1)
```
wayfarer3130: See how this function uses the "this.nonModeModalities" in order to allow a much more generic mode to be created.

modes/basic/src/index.tsx
```diff
@@ -0,0 +1,347 @@
...
+        servicesManager.services;
+
+      measurementService.clearMeasurements();
+
+      // Init Default and SR ToolGroups
+      initToolGroups(extensionManager, toolGroupService, commandsManager);
+
+      toolbarService.register(this.toolbarButtons);
+
+      for(const [key,section] of Object.entries(this.toolbarSections)) {
```
wayfarer3130: Here, the this.toolbarSections are all registered, so it becomes easy to define new sections.

modes/basic/src/index.tsx
```diff
@@ -0,0 +1,347 @@
...
+        customizationService.setCustomizations({
+          'panelSegmentation.disableEditing': {
+            $set: true,
+          },
+        });
+      }
+
+      // // ActivatePanel event trigger for when a segmentation or measurement is added.
+      // // Do not force activation so as to respect the state the user may have left the UI in.
+      if( this.activatePanelTrigger ) {
```
wayfarer3130: The previously commented out code is now wrapped in an if condition to allow it to be included or excluded.

modes/basic/src/index.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+      uiDialogService.hideAll();
+      uiModalService.hide();
+      toolGroupService.destroy();
+      syncGroupService.destroy();
+      segmentationService.destroy();
+      cornerstoneViewportService.destroy();
+    };
+
+export const toolbarSections = {
```
wayfarer3130, line 201: The toolbar sections are exports to allow re-use as required by different applications.
The plan is to actually move these defaults to extensions, and register them by default, referring to them 
inside the customization service.  That is phase 2.

wayfarer3130:
@sedghi - this is an initial example of how we might make the modes more extensible.  I'm demonstrated it by creating a basic mode (non-longitudinal), and then extending it in the longitudinal version.  I think it makes a lot of sense, and I would like to do the other modes as well, excluding perhaps the WSI one, but I wanted you to see how it works as it is now.

modes/basic/src/index.tsx
```diff
@@ -0,0 +1,347 @@
+import initToolGroups from './initToolGroups';
+import toolbarButtons from './toolbarButtons';
+import { id } from './id';
+import { ToolbarService } from '@ohif/core';
+
+const { TOOLBAR_SECTIONS } = ToolbarService;
+
+// Allow this mode by excluding non-imaging modalities such as SR, SEG
+// Also, SM is not a simple imaging modalities, so exclude it.
```
rleisti: This comment does not seem to match the list of excluded modalities: neither SR nor SM are in the list

modes/basic/src/index.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+export function isValidMode({ modalities }) {
+      const modalities_list = modalities.split('\\');
+
+      // Exclude non-image modalities
+      return {
+        valid: !!modalities_list.filter(modality => this.nonModeModalities.indexOf(modality) === -1)
+          .length,
+        description:
+          'The mode does not support studies that ONLY include the following modalities: SM, ECG, SEG, RTSTRUCT',
```
rleisti: Could do `'The mode does not support studies that ONLY include the following modalities:' + this.nonModeModalities.join(', ')` to reduce duplication (and chance of the lists getting out of sync)

platform/app/src/routes/WorkList/WorkList.tsx
rleisti: The hide property should be added to the documentation: [platform/docs/docs/platform/modes/index.md](https://github.com/OHIF/Viewers/blob/master/platform/docs/docs/platform/modes/index.md)

modes/basic/src/index.tsx
```diff
@@ -0,0 +1,347 @@
...
+      dicomPmap.sopClassHandler,
+      ohif.sopClassHandler,
+      ohif.wsiSopClassHandler,
+      dicompdf.sopClassHandler,
+      dicomsr.sopClassHandler3D,
+      dicomsr.sopClassHandler,
+      dicomRT.sopClassHandler,
+    ];
+
+export function isValidMode({ modalities }) {
```
rleisti, line 96: It might helpful to reference this function in [platform/docs/docs/platform/modes/validity.md](https://github.com/OHIF/Viewers/blob/master/platform/docs/docs/platform/modes/validity.md) for new developers to re-use.

rleisti (approved):

platform/app/src/routes/WorkList/WorkList.tsx
```diff
@@ -446,7 +455,7 @@ function WorkList({
                       }
                       onClick={() => {}}
                       dataCY={`mode-${mode.routeName}-${studyInstanceUid}`}
-                      className={isValidMode ? 'text-[13px]' : 'bg-[#222d44] text-[13px]'}
+                      className={isValidMode && 'bg-[#222d44]'}
```
wayfarer3130: Something in the overall change modified the evaluation of css ordering, and caused the text-[13px] to stop working as an override if the original text-[14px] from button is also present.

addOns/externals/devDependencies/package.json
```diff
@@ -35,7 +35,7 @@
     "babel-plugin-module-resolver": "^5.0.0",
     "clean-webpack-plugin": "^3.0.0",
     "copy-webpack-plugin": "^9.0.1",
-    "cross-env": "^5.2.0",
+    "cross-env": "7.0.3",
```
wayfarer3130: Using 7.0.3 consistently as the 10.x version doesn't work in older node

extensions/cornerstone-dicom-pmap/package.json
```diff
@@ -33,10 +33,10 @@
     "start": "yarn run dev"
   },
   "peerDependencies": {
-    "@ohif/core": "3.12.0-beta.35",
-    "@ohif/extension-cornerstone": "3.12.0-beta.35",
-    "@ohif/extension-default": "3.12.0-beta.35",
-    "@ohif/i18n": "3.12.0-beta.35",
+    "@ohif/core": "workspace:*",
```
wayfarer3130: Using workspace reduces need to update numbers

extensions/measurement-tracking/package.json
```diff
@@ -49,7 +49,6 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.20.13",
-    "@ohif/ui": "3.12.0-beta.35",
```
wayfarer3130: These should be peerDependencies so we can link them with workspace:*

package.json
```diff
@@ -55,8 +55,8 @@
     "test:unit:ci": "lerna run test:unit:ci --parallel --stream",
     "test:e2e": "lerna run test:e2e --stream",
     "test:e2e:coverage": "nyc --reporter=html yarn run test:e2e:ci",
-    "test:e2e:ci": "TEST_ENV=true npx playwright test",
-    "test:e2e:ui": "TEST_ENV=true npx playwright test --ui",
+    "test:e2e:ci": "cross-env TEST_ENV=true npx playwright test",
```
wayfarer3130: This allows running on Windows

platform/docs/versioned_docs/version-3.11/migration-guide/3p11-to-3p12/index.md
```diff
@@ -6,3 +6,11 @@ sidebar_label: 3.11 -> 3.12 beta
 # Migration Guide
 
 This guide provides information about migrating from OHIF version 3.11 to version 3.12 beta
+
+## Optional: Migrate modes to extend `modes/basic`
```
sedghi: no need to change this versioned docs, this should be in migration docs of 3.12 beta

platform/ui/src/components/Button/Button.tsx
```diff
@@ -6,6 +6,7 @@ import Tooltip from '../Tooltip/Tooltip';
 
 const sizeClasses = {
   [ButtonEnums.size.small]: 'h-[26px] text-[13px]',
+  [ButtonEnums.size.smallTall]: 'h-[32px] text-[13px]',
```
sedghi, line 9: hmmm, where are you using this new variant? i did not expect this PR to change ui

package.json
```diff
@@ -83,23 +83,24 @@
   },
   "devDependencies": {
     "@babel/core": "7.28.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
```
sedghi, line 86: ask Joe, since i expected these to be fixed in versioning

bun.lock
```diff
@@ -25,6 +25,7 @@
         "@pmmmwh/react-refresh-webpack-plugin": "0.5.17",
         "babel-loader": "8.4.1",
         "babel-plugin-istanbul": "^7.0.0",
+        "cross-env": "7.0.3",
```
jbocce: Did you check if this version is clear of any security flags? Be sure to check meta dependencies too. Let me know - I may check later. Thanks.

jbocce:
@wayfarer3130, I started looking at the lock files and I think based on the fact that there are plenty of conflicts you are best to first resolve the conflicts. However please refrain from manually resolving the lock file conflicts by hand. Instead, revert the lock files back to what they are in master and then run `bun install` and `yarn install` from there. This will give us a better idea of what changed and what new packages and versions are actually being brought in by `bun` and `yarn`. Thanks.

platform/docs/versioned_docs/version-3.11/migration-guide/3p11-to-3p12/index.md
wayfarer3130: This is the 3.12 beta migration document - look at the file name:  3p11-to-3p12/index.md   If that isn't the right location for this document, that is a separate issue, and I can move this to version-3.12 as an entire directory.

platform/ui/src/components/Button/Button.tsx
wayfarer3130: It is used in Worklist.tsx - we were previously using
text-[14px]  ... text-[13px]
which is compiler ordering dependent as to how the css files get generated.  It works if 13px is defined after 14px, but not vice-versa.  This PR changed the compile order just enough to hit that case, although I believe it could have been hit previously randomly.

addOns/externals/devDependencies/package.json
jbocce: Ok so I found that cross-env has a potential vulnerability with one of its metadependencies - namely cross-spawn with versions <7.0.5 (CVE-2024-21538).

So we have two choices: 
- add a cross-spawn resolution of 7.0.6 OR
- use the latest cross-env version of 10.1.0

I will let you decide!

jbocce:
Are we purposely changing the styles of the mode button on the Worklist page?

Now:
image
Before:
image

platform/app/src/routes/WorkList/WorkList.tsx
jbocce: I am not sure I understand. So what does removing the text-[13px] do here?

wayfarer3130:
> Are we purposely changing the styles of the mode button on the Worklist page?
> 
> Now: image Before: image

I removed the text-[13px] incorrectly, just pushed an update for that.  Thank you for noticing that.

modes/basic/README.md
```diff
@@ -0,0 +1,7 @@
+# basic 
+## Description 
+A basic mode used to build other modes on top of
```
jbocce, line 3: Stupid question: can this basic mode be used on its own or is it more of a base mode to build other modes on top of? Two reasons for asking both concern its name (i.e. basic):
1. if it can be used on its own then it MIGHT be confusing because we unfortunately use longitudinal to mean basic.
2. if it cannot be used on its own then base MIGHT be a better name

Actually either way I like base better. But what do I know? 🤣

wayfarer3130:
> I am continuing to review, but definitely the lock files need an update. Again start with the lock files in master and run `bun install` and `yarn install` anew.

That is how I created the updated lock files.  Let me redo that just to make absolutely sure I got them correctly

platform/docs/versioned_docs/version-3.11/migration-guide/3p11-to-3p12/index.md
wayfarer3130: Duh, I see, you mean edit the original version here.  Just updating

platform/app/src/routes/WorkList/WorkList.tsx
```diff
@@ -446,7 +455,7 @@ function WorkList({
                       }
                       onClick={() => {}}
                       dataCY={`mode-${mode.routeName}-${studyInstanceUid}`}
-                      className={isValidMode ? 'text-[13px]' : 'bg-[#222d44] text-[13px]'}
+                      className={!isValidMode && 'bg-[#222d44]'}
```
wayfarer3130, line 458: @sedghi - this over-ride for the text size is why the smallTall above gets used - it wasn't taking effect correctly depending on compile order.
@jbocce - this check got rewritten incorrectly previously to apply the wrong background colour.  This should not really use an absolute colour here, but there doesn't seem to be one defined elsewhere to use instead.  I've corrected this now so the buttons should be ok.

platform/docs/docs/migration-guide/3p11-to-3p12/index.md
wayfarer3130, line 1: @sedghi - moved the documentation to the right file.

addOns/externals/devDependencies/package.json
wayfarer3130: I just removed this change entirely.

modes/basic/README.md
wayfarer3130: I know the naming is confusing - longitudinal should be called longitudinal and not basic.
This mode is actually what should be basic, and yes, it can be used on it's own.  You can see it with:
/viewer?...
replacing with:
/basic?...

sedghi: call yours template, mode-template
so many links are used in demos etc that uses /viewer so i vote against that

platform/docs/docs/migration-guide/3p11-to-3p12/index.md
sedghi: hmmm it showed me in different place, weird

platform/docs/docs/platform/modes/validity.md
```diff
@@ -11,6 +11,8 @@ summary: Documentation for OHIF Mode validity checks, which determine when speci
 There are two mechanism for checking the validity of a mode for a study.
 
 - `isValidMode`: which is called on a selected study in the workList.
+   - There is an example function in modes/basic which returns true if there
```
jbocce: I looked at that function and what you have here is a bit confusing.

addOns/externals/devDependencies/package.json
jbocce: Ok. I checked the old bun.lock file. And we should lock it using exact version 5.2.1. Just FYI

platform/ui/src/components/Button/Button.tsx
jbocce: I am a bit confused. If there is an ordering issue, can't we just use !text-[13px] if we want a text height of 13px? What am I missing?

platform/docs/docs/migration-guide/3p11-to-3p12/index.md
```diff
@@ -6,3 +6,18 @@ sidebar_label: 3.11 -> 3.12 beta
 # Migration Guide
 
 This guide provides information about migrating from OHIF version 3.11 to version 3.12 beta
+
+## Optional: Migrate modes to extend `modes/basic`
+
+There is a lot of support for the basic mode definition contained in the
+`modes/basic` module.  Using this framework will allow your mode to avoid
+creating a lot of boilerplate code that may not upgrade very well.
+However, the existing mode definitions can just be re-used without be based
```
jbocce: I'm not sure I understand this setence that starts with "However...". Please reword. Thanks.

platform/docs/docs/migration-guide/3p11-to-3p12/index.md
```diff
@@ -6,3 +6,18 @@ sidebar_label: 3.11 -> 3.12 beta
...
+on the modes import.
+
+## ui button with text size
+
+Using the class text size with the ui-button is inconsistent as to whether
```
jbocce, line 23: This is confusing to me.

jbocce:
Not sure if we care or not, but do we need to check if the various CLI functions still make new modes correctly?

jbocce (changes requested):
The changes overall look good and will definitely simplify modes.

I reviewed the lock files and no new packages and/or versions were added other than the new cross-env version which I think you said will be reverted. If not then please consider adding the cross-spawn resolution.

I think there are a couple of other comments to address as well.

Thanks so much!

addOns/externals/devDependencies/package.json
wayfarer3130: This is just agreeing with what is used elsewhere in OHIF - it doesn't change what is actually used in OHIF.

platform/app/src/routes/WorkList/WorkList.tsx
wayfarer3130: The problem is that it was random depending on build order whether the text- line would take affect or not.  Hence, periodic e2e test failures.  This has now been incorporated into smallTall size up above.

platform/ui/src/components/Button/Button.tsx
wayfarer3130: I suppose we could but I didn't think of that.  Really, specifying size and then overriding it isn't a great idea and the new code is cleaner.  It is also going away whenever the worklist is redone in favour of ui-next components, so I don't think we should worry about it very much.

platform/app/src/routes/WorkList/WorkList.tsx
jbocce: Not sure I love the name `smallTall`, but if it is ok with @sedghi then it is ok with me. I would rather do the !important !text-[13px] I mentioned.

modes/basic/README.md
wayfarer3130: Calling it template is really ugly if people want to use it as an actual mode - for that, /basic makes sense, and using /template is really ugly.

jbocce (changes requested):
Sorry there are still some docs wording that we should consider updating.

modes/basic/README.md
jbocce: Agreed template is brutto (Italian for ugly). We need something bello.

platform/app/src/routes/WorkList/WorkList.tsx
wayfarer3130: This is code that is going away once the worklist gets revamped.  I don't like either design because it mixes several things, but I like the !important !text-[13px]  less because the class lists are so long it is easy to miss something.

PR #5447, opened 2025-09-30
fix(segmentation): Added missing tool config for labelmap and contour segmentation toolboxes.

jbocce:
### Context
Title says it all.

### Changes & Results
Changed the segmentation configuration to add the config for each added toolbox.

### Testing
Run OHIF. The config section should be available for each segmentation (type) toolbox.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 23.9.0
- [x] Browser: Chrome 140.0.7339.208 
  

### DIFF
--- a/extensions/cornerstone/src/customizations/segmentationPanelCustomization.tsx
+++ b/extensions/cornerstone/src/customizations/segmentationPanelCustomization.tsx
@@ -1,7 +1,7 @@
 import { CustomDropdownMenuContent } from './CustomDropdownMenuContent';
 import { CustomSegmentStatisticsHeader } from './CustomSegmentStatisticsHeader';
-import React, { useState } from 'react';
-import { Switch } from '@ohif/ui-next';
+import SegmentationToolConfig from '../components/SegmentationToolConfig';
+import React from 'react';
 import { SegmentationRepresentations } from '@cornerstonejs/tools/enums';
 
 export default function getSegmentationPanelCustomization({ commandsManager, servicesManager }) {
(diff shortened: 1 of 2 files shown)


extensions/cornerstone/src/customizations/segmentationPanelCustomization.tsx
```diff
@@ -42,66 +42,13 @@ export default function getSegmentationPanelCustomization({ commandsManager, ser
       center: 'Center',
     },
     'segmentationToolbox.config': () => {
```
sedghi, line 44: if they all same return, why we have different key name for this then?

jbocce: @sedghi good question. The Toolbox component looks up the key for the config and we have three different toolboxes. I don't see any other way. Suggestions?

sedghi: can't we just have one key and toolbox look that? I assume inside the Configs it is finding the correct representation type right?

jbocce: @sedghi 

The Toolbox knows nothing about the representation type.

We need three toolboxes because we have different tools in each.

sedghi: alright

jbocce: Maybe this is something we can revisit if (before) we merge to master.

PR #5454, opened 2025-10-02
test(CornerstoneUtils): Adding unit tests for most of Cornerstone Extension utilities

ViniciusResende:
### Context

This PR introduces some small fixes across some of the Cornerstone Extension utility functions and a great amount of unit tests for these utility functions module. This is seen as a improvement in the code quality, enhancing the capability of detecting defects early, also these tests should be kept with care, evolving together with the files.

### Changes & Results

- cd19093f4505f03c1fe35868f18a93b0885d634c, 5566da705bfd18b3a9cb43d98c61b4171e0ce28e, bddda958f0985ba432e5742682125da374b7fbf9, 4d07800936a68d3b78f43c48ed951a362391b165, 7a420c10e6e0fff43e9f2c3b76e183bc13ddb0f3, ec04bc5e2b178d76e962afccafacf303b5857e59 - These are commits for small fixes that were done during the confection of the tests, they're quite self contained, and the commit itself has a message explaining each one individually.
- fcaa079ed998547340a185247c25a3f620a40574, c81e2d58bea503b9a76ee2ad998c46b1627127c3, e5ae00e55ff9cca371876d3f8e0e47d79761fa59 - Moving JS files to TS.
- The other commits are all related to addition of unit tests to the Cornerstone Extension utility functions.

#### Results
The tests are all passing and for most of them 100% of code coverage was reached. For those that do not have this value, it was evaluated that the effort to cover these lines was not compatible with the value got from covering each individual line on those specific files.

image

For some of the functions inside the folder tests were not added now intentionally, because they deserve their own PRs due to it's complexity.

### Testing

To test the specific tests that were added one could run: `yarn test extensions/cornerstone/src/utils`

This will also allow you to check the coverage.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] Node version: v20.9.0

### DIFF
--- a/extensions/cornerstone/src/utils/findNearbyToolData.test.ts
+++ b/extensions/cornerstone/src/utils/findNearbyToolData.test.ts
@@ -1,0 +1,125 @@
+import { findNearbyToolData } from './findNearbyToolData';
+
+describe('findNearbyToolData', () => {
+  const mockCommandsManager = {
+    runCommand: jest.fn(),
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should return undefined when event is null', () => {
+    const result = findNearbyToolData(mockCommandsManager, null);
+
+    expect(result).toBeUndefined();
+    expect(mockCommandsManager.runCommand).not.toHaveBeenCalled();
+  });
+
+  it('should return undefined when event is undefined', () => {
+    const result = findNearbyToolData(mockCommandsManager, undefined);
+
+    expect(result).toBeUndefined();
+    expect(mockCommandsManager.runCommand).not.toHaveBeenCalled();
+  });
+
+  it('should return undefined when event has no detail', () => {
+    const event = {};
+    const result = findNearbyToolData(mockCommandsManager, event);
+
+    expect(result).toBeUndefined();
+    expect(mockCommandsManager.runCommand).not.toHaveBeenCalled();
+  });
+
+  it('should return undefined when event detail is null', () => {
+    const event = { detail: null };
+    const result = findNearbyToolData(mockCommandsManager, event);
+
+    expect(result).toBeUndefined();
+    expect(mockCommandsManager.runCommand).not.toHaveBeenCalled();
+  });
+
+  it('should call runCommand with correct parameters when event has valid detail', () => {
+    const mockElement = document.createElement('div');
+    const mockCurrentPoints = { canvas: { x: 100, y: 200 } };
+    const mockToolData = { id: 'annotation-1' };
+    const event = {
+      detail: {
+        element: mockElement,
+        currentPoints: mockCurrentPoints,
+      },
+    };
+
+    mockCommandsManager.runCommand.mockReturnValue(mockToolData);
+
+    const result = findNearbyToolData(mockCommandsManager, event);
+
+    expect(mockCommandsManager.runCommand).toHaveBeenCalledWith(
+      'getNearbyAnnotation',
+      {
+        element: mockElement,
+        canvasCoordinates: mockCurrentPoints.canvas,
+      },
+      'CORNERSTONE'
+    );
+    expect(result).toBe(mockToolData);
+  });
+
+  it('should handle event with element but no currentPoints', () => {
+    const mockElement = document.createElement('div');
+    const event = {
+      detail: {
+        element: mockElement,
+      },
+    };
+
+    const result = findNearbyToolData(mockCommandsManager, event);
+
+    expect(mockCommandsManager.runCommand).toHaveBeenCalledWith(
+      'getNearbyAnnotation',
+      {
+        element: mockElement,
+        canvasCoordinates: undefined,
+      },
+      'CORNERSTONE'
+    );
+  });
+
+  it('should handle event with currentPoints but no canvas coordinates', () => {
+    const mockElement = document.createElement('div');
+    const mockCurrentPoints = {};
+    const event = {
+      detail: {
+        element: mockElement,
+        currentPoints: mockCurrentPoints,
+      },
+    };
+
+    findNearbyToolData(mockCommandsManager, event);
+
+    expect(mockCommandsManager.runCommand).toHaveBeenCalledWith(
+      'getNearbyAnnotation',
+      {
+        element: mockElement,
+        canvasCoordinates: undefined,
+      },
+      'CORNERSTONE'
+    );
+  });
+
+  it('should return result from runCommand', () => {
+    const mockToolData = { id: 'test-annotation', data: 'test-data' };
+    const event = {
+      detail: {
+        element: document.createElement('div'),
+        currentPoints: { canvas: { x: 50, y: 75 } },
+      },
+    };
+
+    mockCommandsManager.runCommand.mockReturnValue(mockToolData);
+
+    const result = findNearbyToolData(mockCommandsManager, event);
+
+    expect(result).toBe(mockToolData);
+  });
+});
--- a/extensions/cornerstone/src/utils/getInterleavedFrames.test.ts
+++ b/extensions/cornerstone/src/utils/getInterleavedFrames.test.ts
@@ -1,0 +1,85 @@
+import getInterleavedFrames from './getInterleavedFrames';
+
+describe('getInterleavedFrames', () => {
+  it('should return single element when input has one element', () => {
+    const imageIds = ['image-1'];
+    const result = getInterleavedFrames(imageIds);
+
+    expect(result).toEqual([{ imageId: 'image-1', imageIdIndex: 0 }]);
+  });
+
+  it('should return correct order for three elements', () => {
+    const imageIds = ['image-1', 'image-2', 'image-3'];
+    const result = getInterleavedFrames(imageIds);
+
+    expect(result).toEqual([
+      { imageId: 'image-2', imageIdIndex: 1 },
+      { imageId: 'image-1', imageIdIndex: 0 },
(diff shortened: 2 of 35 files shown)


extensions/cornerstone/src/utils/segmentationHandlers.ts
```diff
@@ -85,27 +86,21 @@ export function setupSegmentationModifiedHandler({ segmentationService }) {
...
-        // does not exists anymore we need to remove the bidirectional data
-        const bidirectionalAnnotationsToRemove = bidirectionalAnnotations.filter(
-          annotation =>
-            annotation.metadata.segmentationId === segmentationId &&
-            !segmentIndices.includes(annotation.metadata.segmentIndex)
-        );
-
-        toRemoveUIDs = bidirectionalAnnotationsToRemove.map(annotation => annotation.annotationUID);
-      }
+      const segmentIndices = Object.keys(segmentation.segments)
```
jbocce, line 89: Should we not at the very least still check for a `null` `segmentation` and simply return?

ViniciusResende: There is indeed a check for it at line 25, this is why I removed the second if condition, because it was unreachable.

extensions/cornerstone/src/utils/interleave.ts
```diff
@@ -3,28 +3,30 @@
...
     return [];
   }
   if (lists.length === 1) {
     return lists[0];
   }
-  console.time('interleave');
+
   const useLists = [...lists];
   const ret = [];
   for (let i = 0; useLists.length > 0; i++) {
```
jbocce, line 19: It is unfortunate that the length (and contents) of `useLists` changes with each iteration. This code is just asking/screaming 🤣 for more problems in the future. But it was like that before you improved it, so we leave it, but boy oh boy.

extensions/cornerstone/src/utils/interleave.test.ts
```diff
@@ -0,0 +1,304 @@
...
+    const startTime = performance.now();
+    const result = interleave(input);
+    const endTime = performance.now();
+
+    expect(result).toHaveLength(2000);
+    expect(result[0]).toBe(0);
+    expect(result[1]).toBe(1);
+    expect(result[2]).toBe(2);
+    expect(result[3]).toBe(3);
+    expect(endTime - startTime).toBeLessThan(100);
```
jbocce, line 170: This one bothers me a little. I would assume the actual time is MUCH less than 100 right? If not, then I would consider choosing something a little higher.

extensions/cornerstone/src/utils/interleaveTopToBottom.ts
```diff
@@ -39,31 +39,6 @@ export default function interleaveTopToBottom({
...
-  const filteredMatchDetails = [];
-  const displaySetsToLoad = new Set();
-
-  // Check all viewports that have a displaySet to be loaded. In some cases
-  // (eg: line chart viewports which is not a Cornerstone viewport) the
-  // displaySet is created on the client and there are no instances to be
-  // downloaded. For those viewports the displaySet may have the `skipLoading`
-  // option set to true otherwise it may block the download of all other
-  // instances resulting in blank viewports.
-  Array.from(matchDetails.values()).forEach(curMatchDetails => {
```
jbocce, line 51: Why was this removed? Sorry I am not overly familiar with the code so a little explanation for me would be appreciated. Thanks.

extensions/cornerstone/src/utils/getInterleavedFrames.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    expect(result).toEqual([
+      { imageId: 'image-2', imageIdIndex: 1 },
+      { imageId: 'image-1', imageIdIndex: 0 },
+      { imageId: 'image-3', imageIdIndex: 2 },
+    ]);
+  });
+
+  it('should start with middle element', () => {
+    const imageIds = ['image-1', 'image-2', 'image-3', 'image-4', 'image-5'];
```
jbocce, line 23: Is this really any different than the check for odd length array?

extensions/cornerstone/src/utils/getInterleavedFrames.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    ]);
+  });
+
+  it('should handle large array correctly', () => {
+    const imageIds = Array.from({ length: 10 }, (_, i) => `image-${i + 1}`);
+    const result = getInterleavedFrames(imageIds);
+
+    expect(result).toHaveLength(10);
+    expect(result[0]).toEqual({ imageId: 'image-6', imageIdIndex: 5 });
+    expect(result).toContainEqual({ imageId: 'image-1', imageIdIndex: 0 });
```
jbocce: Instead of `toContainEqual` would it not be better to check `result[8]` and `result[9]`?

extensions/cornerstone/src/utils/findNearbyToolData.test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  it('should handle event with element but no currentPoints', () => {
+    const mockElement = document.createElement('div');
+    const event = {
+      detail: {
+        element: mockElement,
+      },
+    };
+
+    const result = findNearbyToolData(mockCommandsManager, event);
```
jbocce: Do we want to assert this result?

extensions/cornerstone/src/utils/findNearbyToolData.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    const mockElement = document.createElement('div');
+    const mockCurrentPoints = {};
+    const event = {
+      detail: {
+        element: mockElement,
+        currentPoints: mockCurrentPoints,
+      },
+    };
+
+    findNearbyToolData(mockCommandsManager, event);
```
jbocce, line 98: We don't care about the result?

extensions/cornerstone/src/utils/interleave.test.ts
ViniciusResende: Yeah, it is way lower. Actually for my tests it took less that 1ms. My idea with this test was to have a validation that would prevent changes that would drastically lower the performance of the algorithm. I could lower the bound to 10 so it would be tighter, but I think this way is fine, as I said, the idea is to get performance issues that would severely impact the algorithm.

extensions/cornerstone/src/utils/interleaveTopToBottom.ts
ViniciusResende: So the message I added with the commit was: 
>This piece of code was removed due to the fact that it was not being
used anymore by the function after its latest changes, the changes
modified the way that the viewports are verified, thus removing the need
for this logic. Since it was not being used anymore, I opted for
removing it so it would not increase the complexity of a new reader
unnecessarily.

The thing is that with these changes from Alireza: 28e724853b027619fdde570fc73316447d3e37c9, the piece of code that I removed was basically orphan, because it created and filled the values of two variables named `filteredMatchDetails` and `displaySetsToLoad` that weren't being used throughout the code. The problem that this code that I removed solved, seemed to not be reproducible anymore after Alireza changes. Talked with him about that and he agreed that the code wasn't needed, since there was more than a year that it was orphan and there ir no complaints about any loading issue.

Therefore, I opted to remove the unneeded code so there was no complexity overhead for a new reader.

extensions/cornerstone/src/utils/getInterleavedFrames.test.ts
ViniciusResende: They're the same in practical terms but from my view they test different concepts. If you think about the unit tests being a source of documentation of a function/module. Knowing that there is an assertion for the result starting with the middle element and it working correctly for odd length array, seem like different concepts that need to be fulfilled in order to the function to properly work.

extensions/cornerstone/src/utils/findNearbyToolData.test.ts
ViniciusResende: Not actually, this test has the intention to test the correct branching of the `runCommand` call depending on the input parameter

extensions/cornerstone/src/utils/getInterleavedFrames.test.ts
ViniciusResende: You're right, changed that at f12b20b7b18b6cd17320e15395f73da0fafc2c99 to keep consistency.

extensions/cornerstone/src/utils/findNearbyToolData.test.ts
ViniciusResende: We do not, removed it at feda2fc7126c577046c3646e1507c389189e18d2. This due to the same reason of the test below:
> this test has the intention to test the correct branching of the `runCommand` call depending on the input parameter

To validate the result again would be quite redundant due to the other tests above it.

ViniciusResende: We do not, removed it at feda2fc7126c577046c3646e1507c389189e18d2. This due to the same reason of the test below:
> this test has the intention to test the correct branching of the `runCommand` call depending on the input parameter

To validate the result again would be quite redundant due to the other tests above it.

extensions/cornerstone/src/utils/interleave.test.ts
jbocce: No. Please do not decrease it. We want stable, reproducible tests. Decreasing the value, increases the chances of a flakey test.

PR #5455, opened 2025-10-02
fix(segmentation): Activate the segmentation when adding a new segment.

jbocce:
### Context
Whenever a segment is added for a segmentation that is not already active (because a segmentation of the opposite type is active), the segmentation for which the segment was added is NOT activated.

#### Steps to Reproduce
1. Launch OHIF
2. Add a contour segmentation
3. Add a label map segmentation. Note that the label map segmentation is active.
4. Return to the contour segmentation tab and a (contour) segment.
5. Notice that the label map continues to be active

### Changes & Results
Activate the segmentation when adding a new segment.
Folded all segmentation toolboxes into one and hide sections based on where the toolbox is used. Likewise all the segmentation utility sections/toolbars are in one toolbar and sections are hidden based on where the toolbar is used.

### Testing
Perform the steps to reproduce the issue. With the fix, step 5 should now have the contour segmentation activated.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 23.9.0
- [x] Browser: Chrome 140.0.7339.208
  

### DIFF
--- a/extensions/default/src/Toolbar/Toolbar.tsx
+++ b/extensions/default/src/Toolbar/Toolbar.tsx
@@ -66,10 +66,14 @@ export function Toolbar({
 
         const { id, Component, componentProps } = toolDef;
 
-        if (
-          componentProps.buttonSection &&
-          !componentProps?.isSectionVisible?.(subSectionVisibilityProps)
-        ) {
+        // This sub tool/component is visible if it is NOT a button section or
+        // it is a button section and has an isSectionVisible function that returns true.
+        const isSubComponentVisible =
+          !componentProps.buttonSection ||
+          typeof componentProps.isSectionVisible !== 'function' ||
+          componentProps.isSectionVisible(subSectionVisibilityProps);
+
+        if (!isSubComponentVisible) {
           return null;
         }
 
--- a/extensions/default/src/utils/Toolbox.tsx
+++ b/extensions/default/src/utils/Toolbox.tsx
@@ -95,7 +95,11 @@ export function Toolbox({ buttonSectionId, title, subSectionVisibilityProps }: T
           const sectionId = section.componentProps.buttonSection;
           const buttons = toolbarService.getButtonSection(sectionId) as any[];
 
-          if (!section.componentProps?.isSectionVisible?.(subSectionVisibilityProps)) {
+          const isSectionVisible =
+            typeof section.componentProps?.isSectionVisible !== 'function' ||
+            section.componentProps.isSectionVisible(subSectionVisibilityProps);
+
+          if (!isSectionVisible) {
             return null;
           }
 


extensions/default/src/Toolbar/Toolbar.tsx
```diff
@@ -60,7 +101,17 @@ export function Toolbar({ buttonSection = 'primary', viewportId, location }: Too
           />
         );
 
-        return <div key={id}>{tool}</div>;
+        return (
+          <div
```
jbocce, line 108: This div was messing me up.

extensions/cornerstone/src/panels/PanelSegmentation.tsx
```diff
@@ -45,10 +44,8 @@ export default function PanelSegmentation({
     ? selectedSegmentationsForViewportMap?.get(segmentationRepresentationType)
     : undefined;
 
-  const buttonSection = utilitiesSectionMap[segmentationRepresentationType ?? 'Segmentation'];
-
   const { activeToolOptions: activeUtilityOptions } = useActiveToolOptions({
-    buttonSectionId: buttonSection,
+    buttonSectionId: 'SegmentationUtilities',
```
sedghi: can you grab it from ToolbarService enum

sedghi (approved):

extensions/cornerstone/src/panels/PanelSegmentation.tsx
jbocce: ToolbarService does not define "SegmentationUtilities" because "SegmentationUtilities" is NOT a toolbox. However, I did notice a place where I was hardcoding 'segmentationToolbox' so I will use a constant there.

jbocce: I suppose I could add them there too. Ok I will do that.

platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -582,7 +581,7 @@ export default class ToolbarService extends PubSubService {
       btn.component = buttonType.defaultComponent;
     }
 
-    if (!buttonType) {
+    if (!buttonType && !btn.component) {
```
wayfarer3130, line 584: Add a warning/error message here please.

extensions/default/src/Toolbar/Toolbar.tsx
```diff
@@ -33,6 +66,17 @@ export function Toolbar({ buttonSection = 'primary', viewportId, location }: Too
 
         const { id, Component, componentProps } = toolDef;
 
+        // This sub tool/component is visible if it is NOT a button section or
```
wayfarer3130, line 69: It isn't just that it isn't visible, but also that it is removed from the DOM.

PR #5456, opened 2025-10-02
fix(toolbar): prevent duplicate command execution on repeated interactions

pedrokohler:
### Context

Make toolbar interactions idempotent and pure

* **Bug**: Each click mutated `buttonProps.commands` and re-added option commands, so commands duplicated across interactions. 
* **Fix**: Do not mutate. Build a fresh `commands` array per interaction via `buildButtonCommands`, then pass computed props to `recordInteraction`. 

Also:
* adding tests so there are no regressions
* fixing and exporting a few types

### Changes & Results

Before:

https://github.com/user-attachments/assets/fe9bb73b-5595-437a-a8a1-1ab9545fc0de

After:

https://github.com/user-attachments/assets/c36ad371-a6ec-45b4-ad61-0c7dd15d0d05

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


platform/core/src/services/ToolBarService/types.ts
```diff
@@ -40,10 +40,11 @@ export type ButtonOptions = {
   max?: number;
   step?: number;
   value?: number | number[] | string;
-  commands?: (value: unknown) => void;
+  commands?: RunCommand;
   condition?: (props: Record<string, unknown>) => boolean;
   children?: React.ReactNode | (() => React.ReactNode);
   options?: Array<{ value: string; label: string }>;
+  ignoreOnToolbarInteraction?: boolean;
```
sedghi: what is this? why we are adding it?

platform/core/src/utils/buildButtonCommands.ts
```diff
@@ -0,0 +1,45 @@
...
+  for (const command of toArray(buttonProps.commands as RunCommand)) {
+    const key = JSON.stringify({ cmd: command, v: (baseArgs as any)?.value ?? null, src: 'item' });
+    if (seenCommands.has(key)) continue;
+    seenCommands.add(key);
+    allCommands.push(() => commandsManager.run(command, baseArgs));
+  }
+
+  // 2) normalize option-level commands
+  for (const option of toArray(buttonProps.options)) {
+    if (!option || option.ignoreOnToolbarInteraction || !option.commands) continue;
```
sedghi: our formatting rules disallows one liners. Use carely braces.

platform/core/src/utils/buildButtonCommands.ts
```diff
@@ -0,0 +1,45 @@
...
+      };
+      const key = JSON.stringify({ cmd: command, v: valueToUse, src: 'option' });
+
+      if (seenCommands.has(key)) continue;
+
+      seenCommands.add(key);
+      allCommands.push(() => commandsManager.run(command, commandOptions));
+    }
+  }
+
```
sedghi, line 38: I'm kind of confused by this overal piece of code. What are you trying to do?

pedrokohler: We're aggregating the commands into a single commands array while making sure that there are no duplicating any of them.

platform/core/src/services/ToolBarService/types.ts
pedrokohler: This is just a flag we have been using. It's useful if for whatever reason we don't want the options commands to be added to the parent item too.

If you want I can remove it, but it's been very useful in our context to create more complex toolbars.

platform/core/src/utils/buildButtonCommands.ts
pedrokohler: Which is already exactly what was done before, except for the fact that it was duplicating commands because of mutations on the buttonProps, and thus generating the error that I showed in the videos attached.

sedghi: I don't understand this pattern 

JSON.stringify({ cmd: command, v: (baseArgs as any)?.value ?? null, src: 'item' })

can you simplify it somehow else?

pedrokohler: I just removed them as they are not required to solve the issue.

But what it was doing was: it would generate a snapshot of the command, its arguments and its origin as text and compare it. If for whatever reason there was another command with the exact same snapshot, it would simply ignore and skip it. But as I said, this is not necessary, it's was just being extra careful.

platform/core/src/utils/buildButtonCommands.ts
```diff
@@ -0,0 +1,40 @@
...
+        servicesManager,
+        commandsManager,
+      };
+
+      allCommands.push(() => commandsManager.run(command, commandOptions));
+    }
+  }
+
+  return allCommands;
+};
```
sedghi, line 40: Looks much more readable, my last comment is what is ignoreOnToolbarInteraction , we didn't have it before?

pedrokohler: No, I answered above:

This is just a flag we have been using. It's useful if for whatever reason we don't want the options commands to be added to the parent item too.

If you want I can remove it, but it's been very useful in our context to create more complex toolbars.

pedrokohler: But we didn't have that on open source before.

jbocce: For me it is fine to leave `ignoreOnToolbarInteraction`. Perhaps you could give us an example of where you found it useful? It is ok if you can't.

jbocce (approved):

platform/core/src/utils/buildButtonCommands.ts
sedghi: It would be great if you don't resolve comments that include questions and answers. I typically resolve comments after applying reviewer feedback, but I leave them open if there's information the reviewer needs to see or note. Let me read

sedghi: I read the comment, please remove it until we have a real use case for it. 
@pedrokohler

pedrokohler: We do have real use cases. 

In a case like this one below, we don't want to run every command when we click the parent element. Instead, we want to leave the commands to when you interact with the children. 

```
{
      id: 'SAM',
      uiType: 'ohif.toolBoxButton',
      props: {
        icon: 'icon-smart-tool',
        label: 'Smart SAM',
        value: 'SmartSAM',
        evaluate: {
          name: 'evaluate.cornerstoneTool',
          toolNames: [
            'SmartSAM',
            'BoxPrompt',
            'MarkerInclude',
            'MarkerExclude',
          ],
        },
        commands: 'setToolActiveToolbar',
        options: [
          {
            name: 'SAM Mode',
            type: 'radio',
            id: 'sam-marker-mode',
            value: 'BoxPrompt',
            values: [
              { value: 'BoxPrompt', label: 'Box' },
              { value: 'MarkerInclude', label: 'Include' },
              { value: 'MarkerExclude', label: 'Exclude' },
            ],
            commands: 'setToolActiveToolbar',
          },
          {
            ignoreOnToolbarInteraction: true,
            name: 'EnablePropagation',
            type: 'checkbox',
            id: 'enable-propagation',
            label: 'Propagate',
            value: false,
            commands: {
              commandName: 'togglePropagation',
              commandOptions: {
                toolNames: ['SmartSAM'],
              },
            },
          },
          {
            ignoreOnToolbarInteraction: true,
            name: 'PropagationRange',
            id: 'propagate-range',
            type: 'custom-double-range',
            commands: {
              commandName: 'setPropagationRange',
              commandOptions: {
                toolNames: ['SmartSAM'],
              },
            },
          },
          {
            ignoreOnToolbarInteraction: true,
            name: 'Submit',
            type: 'button',
            id: 'submit-button',
            variant: 'default',
            commands: 'invokeSAM',
          },
          {
            ignoreOnToolbarInteraction: true,
            name: 'Clear Markers',
            type: 'button',
            id: 'clear-markers',
            variant: 'default',
            label: 'Clear Markers',
            commands: 'clearMarkers',
          },
        ],
      },
    },
    ```

sedghi: In that case, we should call it explicitRunOnly or independent or similar

However, the correct way tp implement this is to have your command actual skip running and returning early based on who initiated the sequence. Right now your flag (ignore...) is exposing implementation detail of our toolbar service which runs everything for parent and children. I am fine with this new flag for now but we should move towards book keeping who initiated the sequence soon

pedrokohler: Cool. I've changed the name.

PR #5494, opened 2025-10-14
fix(security): Use exact versioning for dependencies in package.json files.

jbocce:
### Context
There have been many supply chain attacks. One level of mitigating the risk of these types of attacks is to use exact versioning in the package.json files for a project. Note however that this does NOT eliminate the thread as meta/transitive dependencies (i.e. the dependencies of the dependencies) do not necessarily use exact versioning.
 

### Changes & Results
Use exact versioning for dependencies in package.json files.
Added `bun audit` to build process.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 23.9.0
- [x] Browser: Chrome  141.0.7390.77
  

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,13 +18,6 @@ commands:
       - run:
           name: Install Bun
           command: |
-            # Uninstall existing Bun if present
-            if [ -d "$HOME/.bun" ]; then
-              echo "Removing existing Bun installation..."
-              rm -rf "$HOME/.bun"
-            fi
-
-            # Install Bun
             curl -fsSL https://bun.sh/install | bash -s "bun-v1.2.23"
             echo 'export BUN_INSTALL="$HOME/.bun"' >> $BASH_ENV
             echo 'export PATH="$BUN_INSTALL/bin:$PATH"' >> $BASH_ENV


.circleci/config.yml
```diff
@@ -124,6 +124,34 @@ jobs:
...
           name: Install Dependencies
           command: bun install --frozen-lockfile
+      # SECURITY AUDIT
+      - run:
+          name: 'Security Audit - High Risk Vulnerabilities'
+          command: |
+            echo "🔍 Running bun audit for security vulnerabilities..."
+            echo "Checking for HIGH-RISK vulnerabilities..."
+
+            if bun audit --audit-level high; then
```
jbocce, line 125: bun audit existed in pre/before bun version 1.3.

bun.lock
jbocce, line 1: The state of this diff demonstrates that we have the correct exact versions in the package.json files. Why? Because the "bottom" portion of the `bun.lock` file (i.e. the packages section) shows no new and/or different package versions.

.circleci/config.yml
```diff
@@ -21,9 +21,14 @@ commands:
       - run:
           name: Install Bun
           command: |
-            if [ ! -d "$HOME/.bun" ]; then
-              curl -fsSL https://bun.sh/install | bash -s "bun-v1.2.23"
+            # Uninstall existing Bun if present
+            if [ -d "$HOME/.bun" ]; then
```
jbocce: I was having trouble without this. So now we uninstall bun first. That said, I don't think we need the restore_cache and save_cache steps anymore. Thoughts? Note that this has NOTHING to do with the added audit step. I needed this regardless of that step.

PR #5495, opened 2025-10-14
fix(segmentationService): prevent no representation crash

pedrokohler:
### Context

Prevent crash when no representation exists: `representation.type` could read `type` of `undefined`. Since the function already receives `type` getting the representation first is unnecessary; also the “no representation” path is already handled by `_setSegmentationRepresentationVisibility`.

### Changes & Results

* Use the passed `type` directly.
* Delegate "no representations" case to `_setSegmentationRepresentationVisibility` with the provided `type`.
* Tests: adjust call counts; and actually test early return of `_setSegmentationRepresentationVisibility`.
* Behavior unchanged except the possible crash is gone.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


extensions/cornerstone/src/services/SegmentationService/SegmentationService.test.ts
```diff
@@ -2511,20 +2511,20 @@ describe('SegmentationService', () => {
...
-      service.toggleSegmentationRepresentationVisibility(viewportId, {
-        segmentationId: representations[0].segmentationId,
-        type: representations[0].type,
-      });
+      service['_setSegmentationRepresentationVisibility'](
+        viewportId,
+        representations[0].segmentationId,
+        representations[0].type,
+        true
+      );
```
pedrokohler, line 2527: There was never an "early return" in the toggle function, so I'm just directly testing the function that actually returns early here, which is `setSegmentationRepresentationVisbility` and not the previous function.

extensions/cornerstone/src/services/SegmentationService/SegmentationService.test.ts
```diff
@@ -2482,7 +2482,7 @@ describe('SegmentationService', () => {
         type: representations[0].type,
       });
 
-      expect(service.getSegmentationRepresentations).toHaveBeenCalledTimes(2);
+      expect(service.getSegmentationRepresentations).toHaveBeenCalledTimes(1);
```
pedrokohler, line 2485: Since we only get the segmentation representation once now, this had to change.

jbocce (approved):

Issue #5395, opened 2025-09-09
[Bug] RTSTRUCT contours are editable by default


fedorov:
### Describe the Bug

When opening and loading RTSTRUCT series, it is opened in editable mode where contour can be (unintentionally) modified.

### Steps to Reproduce

1. Get the sample from IDC (you can see it with the older release of OHIF here: https://viewer.imaging.datacommons.cancer.gov/v3/viewer/?StudyInstanceUIDs=1.3.6.1.4.1.32722.99.99.282967364651788470277412461462049836277) (note that RTSTRUCT would not open there, due to another regression in OHIF documented in https://github.com/ImagingDataCommons/ViewersV3/issues/95)
  * `pip install --upgrade idc-index`
  * `idc download 1.3.6.1.4.1.32722.99.99.282967364651788470277412461462049836277`
2. Open RTSTRUCT series

### The current behavior

Clicking on a contour and dragging the mouse will change the contour.

### The expected behavior

Contours are viewed by default in read-only mode.

### System Information

Chrome browser, but this issue is most likely independent from the browser or environment.

sedghi:
That is bad

fedorov:
@igoroctaviano it is ok to do this under IDC, even though it is a regression in OHIF, as long as this does not evolve into a significant effort.

sedghi:
I'm not sure if this was working before? we never tested RT against Google maybe? But great that it is fixed now

fedorov:
Are you saying that the fact that it is editable is somehow caused by the DICOM backend? That would be super strange.

sedghi:
Oh not the editing not, but the PR was making it work with bulkData, i think we never tested the RT against servers that return bulkData

fedorov:
Right, but the PR you are talking about https://github.com/OHIF/Viewers/pull/5415 is for a different issue https://github.com/ImagingDataCommons/ViewersV3/issues/95. I was talking about **this** issue! 😄

sedghi:
@fedorov we are fixing it in our 3.11 recovery 1

fedorov:
@sedghi we need better coordination - @igoroctaviano already started working on this one (and he was assigned to the ticket).

sedghi:
That sounds good, we will hold

igoroctaviano:
@jbocce 

### **Description**

When `disableEditing: true` is set in the `appConfig`, the behavior of the measurement and segmentation panels does not fully respect the intended read-only state.

[...]

PR #5503, opened 2025-10-16
fix(segmentation): Lock all rehydrated segmentation segments when panelSegmentation.disableEditing is true.

jbocce:
### Context
See #5395 

### Changes & Results
In CS3D, lock all annotations associated with the contour segment being locked.
In OHIF, when segmentations are locked via `panelSegmentation.disableEditing` lock the hydrated segmentations and segments. Prior to this PR, even hydrated label maps were NOT being locked properly.

### Testing
Test A
1. Launch a study with segmentations into OHIF using basic mode.
2. Hydrate the segmentations.
3. The segmentations should not be editable.

Test B
1. Launch a study with segmentations into OHIF using segmentation mode.
2. Hydrate the segmentations.
3. The segmentations should be editable by default.
4. Lock a segment. That segment should no longer be editable.

Test C
1. Set `panelSegmentation.disableEditing` to  `true` in the config file.
2. Open a study with segmentations in OHIF.
3. Hydrate the segmentations.
4. The segmentations should not be editable.

Test D
1. Set `panelSegmentation.disableEditing` to  `false` in the config file. Note that this is a global configuration setting and will override any configuration set by a mode.
2. Open a study with segmentations in OHIF.
3. Hydrate the segmentations.
4. The segmentations will be editable by default.
5. Lock a segment. That segment should no longer be editable.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Wiindows 11
- [x] Node version: 23.9.0
- [x] Browser: Chrome 141.0.7390.108
  

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -52,7 +52,7 @@
     },
     "addOns/externals/devDependencies": {
       "name": "@externals/devDependencies",
-      "version": "3.12.0-beta.64",
+      "version": "3.12.0-beta.67",
       "dependencies": {
         "@babel/runtime": "7.28.2",
         "@kitware/vtk.js": "32.12.0",
@@ -137,14 +137,14 @@
     },
     "addOns/externals/dicom-microscopy-viewer": {
       "name": "@externals/dicom-microscopy-viewer",
-      "version": "3.12.0-beta.64",
+      "version": "3.12.0-beta.67",
       "dependencies": {
         "dicom-microscopy-viewer": "0.48.6",
       },
     },
     "extensions/cornerstone": {
       "name": "@ohif/extension-cornerstone",
-      "version": "3.12.0-beta.64",
+      "version": "3.12.0-beta.67",
       "dependencies": {
         "@babel/runtime": "7.28.2",
         "@cornerstonejs/adapters": "4.5.13",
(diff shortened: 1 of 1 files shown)


extensions/cornerstone/src/commandsModule.ts
```diff
@@ -266,6 +266,36 @@ function commandsModule({
...
         storePositionPresentation(referencedDisplaySet);
+
+        const disableEditing = customizationService.getCustomization(
+          'panelSegmentation.disableEditing'
+        );
+        if (disableEditing) {
+          // The segmentations are hydrated asynchronously, so we need to subscribe
+          // to the grid state changed event to ensure the segmentations are first
+          // hydrated before we lock the segments.
+          const { unsubscribe } = viewportGridService.subscribe(
```
sedghi: What if the event happens sooner? You'll never catch it then. Maybe you should perform the operation you want first, and if it doesn't succeed (check some state), then subscribe for the next update.

jbocce: Hmmm, but then it might be too late to subscribe to the event.

Maybe I need to listen for a segmentation event here instead.

sedghi: I'm not sure, that sounds more correct to listen to seg event, but still don't like that we are doing it here. Maybe we do it in segmentation service just before/after we add the seg?

jbocce: But checking customization/configuration in the SegmentationService does not seem right either, no?

sedghi: We're already using services within services. I'm not a huge fan of it either, but it honestly seems more stable than listening to them here

jbocce: Ok. I have something. Once https://github.com/cornerstonejs/cornerstone3D/pull/2399 is merged, I update the dependencies and I push some new code here there will be no event to wait for and everything will be nice (in my opinion).

PR #5508, opened 2025-10-20
fix(security): For bun, enforce frozen lockfile. For yarn, strongly suggest using frozen lockfile.

jbocce:
### Context
One way to help minimize supply chain attacks is to always use a frozen lockfile. `bun` provides a mechanism for doing so. Modern `yarn`  has a mechanism as well, but OHIF uses classic `yarn`.

### Changes & Results
Added a `bunfig.toml` file to enforce `--frozen-lockfile` for `bun` installs.  This means that changes to the package.json files requires performing a `bun install --config=./bunfig.no-frozen-lockfile.toml` in order to update the `bun.lock` file. Of course, care should be taken as to what is being added to the package.json files - ensuring no vulnerable packages and/or versions are introduced.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 23.9.0
- [x] Browser: Chrome 141.0.7390.108
  


bunfig.update-lockfile.toml
jbocce, line 1: What do you think about this file and its name?

jbocce:
A request was made to bun to allow for easier override on the `frozenLockFile` setting in the `bunfig.toml` file. See https://github.com/oven-sh/bun/issues/23913

bunfig.update-lockfile.toml
wayfarer3130: It looks like we need a second toml file to do this, so I think the naming is reasonable.

package.json
wayfarer3130, line 1: Can you create a task `install:unfrozen`
that renames the bunfig from the frozen bunfig, calls bun install, and then renames it back?
That way:
bun run install:unfrozen 
just works to update the bun.lock.

bunfig.toml
wayfarer3130, line 1: Can you call this one bunfig.froze-lockfile.toml
and copy it in the postinstall script?

jbocce: @wayfarer3130  But why? As the PR stands we have...

1. `bun install` that because of this bunfig.toml file it will automatically enforce a frozen lockfile. We really, really want this.
2. `bun install --config=./bunfig.no-frozen-lockfile.toml` when someone wants to update dependencies

If anything I could add equivalent tasks for 1 and 2 above so that we ALSO have
1a. `bun run install:frozen`
2a. `bun run install:unfrozen`

But to be honest I see no reason for renaming and copying files. But maybe I am missing something.

package.json
jbocce: If I were to add such a task it would simply run `bun install --config=./bunfig.no-frozen-lockfile.toml`. No need for renaming.

jbocce:
> Could you add the command I suggested and the postinstall script to copy the bunfig.frozen... to bunfig?

See my comments.

package.json
wayfarer3130: Oh, yes, that would work.  Can you create such a script in the root level?  That is even easier, I didn't realize there was a config setting for it.

bunfig.toml
wayfarer3130: Just have the script in the root package.json to install with the no-frozen.  Even if it is just for reference, it makes it easier to find the right option, and I like discoverability.

sedghi (approved):

PR #5519, opened 2025-10-23
feat(jpeg2000): Add 16-bit RGB support to JPEG2000 decoder 

IbrahimCSAE:
### Context

Implements missing 16-bit RGB pixel copy logic in the OpenJPEG decoder. Previously, 16-bit RGB images would decode with zero pixel data.

### Issue
The decoder handled single-component and 8-bit RGB images correctly, but 16-bit RGB images had their pixel copy code commented out. This caused the buffer to be allocated but remain empty, resulting in black output.

### Changes
**`src/J2KDecoder.hpp`** (lines 884-901)
- Implemented 16-bit RGB pixel copy for both signed and unsigned images
- Handles all three color components with proper bounds checking

### Testing
Verified with 16-bit RGB ultrasound DICOM images that previously decoded as black. Images now display correctly.

### Impact
Low risk, adds missing functionality without affecting existing 8-bit or grayscale decoding paths.

### Before / After

CleanShot 2025-10-21 at 23 03 50@2x

CleanShot 2025-10-21 at 23 03 21@2x

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -979,6 +979,7 @@
     "rollup": "2.79.2",
     "commander": "8.3.0",
     "tapable": "2.2.2",
+    "@cornerstonejs/codec-openjpeg": "1.3.0",
     "axios": "1.12.0",
   },
   "packages": {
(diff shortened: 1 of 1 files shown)


IbrahimCSAE:
test error is 502 gateway, not actual

jbocce:
> test error is 502 gateway, not actual

I am rerunning the tests.

package.json
```diff
@@ -148,7 +148,8 @@
     "axios": "1.12.0",
     "core-js": "3.45.1",
     "@babel/runtime-corejs2": "7.26.10",
-    "tapable": "2.2.2"
+    "tapable": "2.2.2",
+    "@cornerstonejs/codec-openjpeg": "1.3.0"
```
jbocce, line 152: Oh yeah... a resolution. Nice.

bun.lock
jbocce, line 1: lock file looks bellissimo

yarn.lock
jbocce, line 1: nice yarn lock file 😉

PR #5531, opened 2025-10-28
fix(dev): update webpack proxy config syntax for Orthanc dev mode

Omnipotent-coder0:
### Context

Running `yarn run dev:orthanc` failed due to an outdated proxy configuration syntax in `.webpack/webpack.pwa.js`.

The error message from the dev server was:

```
[webpack-cli] Invalid options object. Dev Server has been initialized using an options object that does not match the API schema.
 - options.proxy should be an array:
   [object { … } | function, ...]

```

This happens because recent versions of `webpack-dev-server` (v4+) require the `proxy` configuration to be defined as an **array**, but the OHIF Viewer still uses the older object-based format.

As a result, the Orthanc development mode fails to start.

---

### Changes & Results

**Before:**

```jsx
if (hasProxy) {
  mergedConfig.devServer.proxy = mergedConfig.devServer.proxy || {};
  mergedConfig.devServer.proxy = {
    [PROXY_TARGET]: {
      target: PROXY_DOMAIN,
      changeOrigin: true,
      pathRewrite: {
        [`^${PROXY_PATH_REWRITE_FROM}`]: PROXY_PATH_REWRITE_TO,
      },
    },
  };
}

```

**After:**

```jsx
if (hasProxy) {
  mergedConfig.devServer.proxy = mergedConfig.devServer.proxy || {};
  mergedConfig.devServer.proxy = [
    {
      context: [PROXY_PATH_REWRITE_FROM || '/dicomweb'],
      target: PROXY_DOMAIN,
      changeOrigin: true,
      pathRewrite: {
        [`^${PROXY_PATH_REWRITE_FROM}`]: PROXY_PATH_REWRITE_TO,
      },
    },
  ];
}

```

✅ After this change, running `yarn run dev:orthanc` successfully starts the development environment and connects to the Orthanc DICOM server without errors.

No regressions were observed for other dev modes.

---

### Testing

**Steps to verify the fix:**

1. Checkout this branch
2. Run:
    
    ```bash
    yarn install
    yarn run dev:orthanc
    
    ```
    
3. Confirm that:
    - The development server starts successfully
    - The OHIF Viewer loads in the browser
    - DICOM studies from Orthanc are accessible at `/dicomweb`

**Tested environment:**

- **OS:** Windows 10
- **Node:** v22.16.0
- **Browser:** Microsoft Edge 141.0.3537.99 (64-bit)

### DIFF
--- a/platform/app/public/config/docker-nginx-orthanc.js
+++ b/platform/app/public/config/docker-nginx-orthanc.js
@@ -16,11 +16,11 @@ window.config = {
     maxNumPrefetchRequests: 10,
     order: 'closest',
   },
-  defaultDataSourceName: 'dicomweb',
+  defaultDataSourceName: 'orthancProxy',
   dataSources: [
     {
       namespace: '@ohif/extension-default.dataSourcesModule.dicomweb',
-      sourceName: 'dicomweb',
+      sourceName: 'orthancProxy',
       configuration: {
         friendlyName: 'Orthanc Server',
         name: 'Orthanc',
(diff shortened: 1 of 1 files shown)


wayfarer3130:
Will merge once the integration test passes.
I added two datasources for testing as I don't have an orthanc setup to directly test against

PR #5535, opened 2025-10-29
fix(SegmentationTools): [Bug] Changes of brush/eraser radius with hotkey do not reflect on segmentation tool

### LINKED ISSUE #5528: [Bug] Changes of brush/eraser radius with hotkey do not reflect on segmentation tool
### Describe the Bug

Changes of brush/eraser radius with hotkey [] do not reflect on segmentation tool

### Steps to Reproduce

When i use hotkey [] to change the size of the brush or eraser while in segmentation tool, it changes correctly in the viewport but the segmentation tool 'Radius (mm)' doesn't reflect the changes. 

### The current behavior

Changes of brush/eraser radius with hotkey [] do not reflect on segmentation tool

### The expected behavior

Changes of brush/eraser radius with hotkey [] reflect on segmentation tool

### System Information

  System:
    OS: Linux 6.8 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (16) x64 AMD EPYC-Rome Processor
    Memory: 28.41 GB / 31.34 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 20.19.5 - /usr/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 10.8.2 - /usr/bin/npm

and it also happen to https://viewer.ohif.org/ 
I am using Ohif 3.12.0 beta 54. Tested in both chrome and edge.

ViniciusResende:
### Context
This PR contains fix and enhancements related to the Segmentation Brush Tools, being them the Brush, Eraser or Threshold. The issue is better described at #5528. Also, previously the hotkey value change was not being kept inside the allowed range, causing weird behaviors.

Fixes #5528

### Changes & Results

- 2995cf5388d9c73a2e6ac3c39dff03830e113333 - adds two new methods to `ToolbarService` to allow external change of button option value.

- fe587742b1383e2c92105dada35927c2852e0a61 - uses new method to enhance increase and decrease brush sizes commands to reflect change across the Segmentation Panel buttons.

#### Previous behavior
https://github.com/user-attachments/assets/f335cc72-6f40-4adc-9854-309e4368e464
> Note that previously, if I kept decreasing with the hotkey, the brush size would suddenly grow again

#### Current behavior
https://github.com/user-attachments/assets/6d217117-a238-4323-b629-febee2751e2a
> Values are now kept between allowed range

### Testing

Please see the videos above or follow the steps described at #5528

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] Node version: 20.9.0


platform/core/src/services/ToolBarService/ToolbarService.ts
```diff
@@ -833,4 +833,63 @@ export default class ToolbarService extends PubSubService {
...
+    const button = this.getButton(buttonId);
+    const option = button?.props?.options?.find(option => option.id === optionId);
+
+    if (!option) {
+      console.warn(`Option ${optionId} not found for button ${buttonId}`);
+      return;
+    }
+
+    let newValue = value;
+    if (option.max && option.min) {
```
jbocce: It looks like min and max capping are going on in two different places - here and in the commandsModule. We really should only have one place.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -2082,7 +2092,16 @@ function commandsModule({
...
         const brushSize = segmentationUtils.getBrushSizeForToolGroup(toolGroupId);
-        segmentationUtils.setBrushSizeForToolGroup(toolGroupId, brushSize - 3);
+
+        const newBrushSize = Math.max(brushSize - 3, 0.5);
+
+        if (brushSize) {
+          segmentationUtils.setBrushSizeForToolGroup(toolGroupId, newBrushSize);
+
+          const activeButton = toolbarService.getActiveButton();
+          const optionId = activeButton.props.options[0].id;
```
jbocce: Why the first option? If the order changes this would be problematic, no?

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -2082,7 +2092,16 @@ function commandsModule({
 
       for (const toolGroupId of toolGroupIds) {
         const brushSize = segmentationUtils.getBrushSizeForToolGroup(toolGroupId);
-        segmentationUtils.setBrushSizeForToolGroup(toolGroupId, brushSize - 3);
+
```
jbocce: This and the above are almost exactly alike so we should probably refactor them to a utility method.

extensions/cornerstone/src/commandsModule.ts
jbocce: Hmm... because this code is in an extension, but the buttons are defined in the mode we have a bit of a problem here I think. This code is assuming it knows something about how the button and its properties are defined (e.g. the ordering). I think an alternative approach that keeps the responsibilities separate would be to have an `evaluate` that keeps the brush size in sync. Instead of updating the button state itself, these commands would simply invoke `toolbarService.refreshToolbarState(...)`. The evaluator would then kick in and update the button state. I think we should at least try that approach.

jbocce:
We should also add a playwright test for this.

sedghi:
> We should also add a playwright test for this.

I agree, but instead of screenshots, we should do it with hitting keyboard keys and programmatically and checking changing element values.

platform/core/src/services/ToolBarService/ToolbarService.ts
ViniciusResende: For this solution specifically, the two checks aren't related. One of them is inside a ToolbarService method, that is a general purpose method that works on top of Tool Button options specifications, so it is not purpose specific and can be reused in different scenarios where this method might be called in the future.

The other place that I check this is inside the `commandsModule`, there I have something that is context specific, and I do that check not only for the call of the `setButtonValue` but also for the `setBrushSizeForToolGroup`. The lack of this check was causing a bug:
> Note that previously, if I kept decreasing with the hotkey, the brush size would suddenly grow again

extensions/cornerstone/src/commandsModule.ts
ViniciusResende: That is indeed a good approach that I haven't tried, I'll try it out. Thanks for the suggestion.

extensions/cornerstone/src/commandsModule.ts
ViniciusResende: Yeah, it was already like that previously so I thought it was intentional. I'll extract that logic.

ViniciusResende:
@jbocce @sedghi I think that a playwright test specifically for that seems like a ad-hoc test scenario.
I propose we add a `SegmentationDrawingToolsResizing` integration test. So the issue of this PR is tested as part of a more "feature driven" test scenario.

> updated the name after Joe mention we should focus on a smaller scenario for this PR.

jbocce:
> @jbocce @sedghi I think that a playwright test specifically for that seems like a ad-hoc test scenario. I propose we add a `SegmentationDrawingToolsResizing` integration test. So the issue of this PR is tested as part of a more "feature driven" test scenario.
> 
> > updated the name after Joe mention we should focus on a smaller scenario for this PR.

I agree with having broader scoped tests. However that goes beyond the scope of this PR and it is something we would welcome in perhaps a different PR focusing on tests. Please ensure the tests added are centered around keyboard shortcuts changing the size of the brush(es).

extensions/cornerstone/src/commandsModule.ts
ViniciusResende: Can you check if 05ee871fbe29efbbc3a83329260db759a7935ffe matches your expectancy around that?

extensions/cornerstone/src/commandsModule.ts
ViniciusResende: Done at 831674bde538c6e955fcf3c2dd7cc482831871ee

extensions/cornerstone/src/commandsModule.ts
jbocce: Yes that looks like what I was expecting.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -2063,27 +2064,30 @@ function commandsModule({
         });
       }
     },
-    increaseBrushSize: () => {
+    _handleBrushSizeAction: (action: 'increase' | 'decrease') => {
```
jbocce: I think this should be moved to outside the `actions` object since it itself is not an action. Notice there are other utility functions above the `actions` object.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -2063,27 +2064,30 @@ function commandsModule({
...
       if (!toolGroupIds?.length) {
         return;
       }
 
       for (const toolGroupId of toolGroupIds) {
         const brushSize = segmentationUtils.getBrushSizeForToolGroup(toolGroupId);
-        segmentationUtils.setBrushSizeForToolGroup(toolGroupId, brushSize + 3);
+
+        const newBrushSize =
+          action === 'increase' ? Math.min(brushSize + 3, 99.5) : Math.max(brushSize - 3, 0.5);
```
jbocce: One thing that bothers me is that this capping between >= 0.5 and <= 99.5 could differ what is specified in the button options. Which is fine in a sense because this logic could and should exist outside of there being a button/control. However, the button evaluator should also call `segmentationUtils.setBrushSizeForToolGroup` for the case where the capping is different.

extensions/cornerstone-dicom-seg/src/getToolbarModule.ts
```diff
@@ -118,5 +120,22 @@ export function getToolbarModule({ servicesManager }: withAppTypes) {
...
+        const toolGroupIds = toolGroupService.getToolGroupIds();
+        if (!toolGroupIds?.length) {
+          return;
+        }
+
+        for (const toolGroupId of toolGroupIds) {
+          const brushSize = cstUtils.segmentation.getBrushSizeForToolGroup(toolGroupId);
+
+          if (brushSize) {
+            toolbarService.setButtonValue(button.id, radiusOptionId, brushSize);
```
jbocce: See my comments in `commandsModule`, but maybe have `setButtonValue` return the new value and call `segmentationUtils.setBrushSizeForToolGroup` whenever the `brushSize` is capped such that its value changes. I think we should also rename `setButtonValue` to `setButtonOptionValue` so it better conforms to the code and docs you added for that function.

ViniciusResende: I could do that but I think I should keep the check on the command too. Because if I don't, I experience a jump on the brush size, and then it shrinks back.

Your idea is to keep the check in both places or to rely just on this evaluator function?

jbocce: Keep it in both places. The command capping is needed for the case when there is no other capping.

jbocce: I wish we could simply centralize this, but I cannot think of a way. It is really too bad that the command caps AND the options cap - one hardcodes the cap values, the other is configured so to speak.

extensions/cornerstone-dicom-seg/src/getToolbarModule.ts
```diff
@@ -118,5 +120,22 @@ export function getToolbarModule({ servicesManager }: withAppTypes) {
...
+        const toolGroupIds = toolGroupService.getToolGroupIds();
+        if (!toolGroupIds?.length) {
+          return;
+        }
+
+        for (const toolGroupId of toolGroupIds) {
+          const brushSize = cstUtils.segmentation.getBrushSizeForToolGroup(toolGroupId);
+
+          if (brushSize) {
+            button.props.options.find(option => option.id === radiusOptionId).value = brushSize;
```
jbocce: I think we should consider adding a `getOptionById` method to the ToolbarService to do the find here. Likely a useful utility.

modes/segmentation/src/toolbarButtons.ts
```diff
@@ -667,6 +671,7 @@ const toolbarButtons: Button[] = [
           name: 'Radius (mm)',
           id: 'brush-radius',
           type: 'range',
+          explicitRunOnly: true,
           min: 0.5,
```
jbocce: Is it possible to remove the min and max from here and read it from the tool config?

modes/segmentation/src/toolbarButtons.ts
```diff
@@ -667,6 +671,7 @@ const toolbarButtons: Button[] = [
           name: 'Radius (mm)',
           id: 'brush-radius',
           type: 'range',
+          explicitRunOnly: true,
```
jbocce, line 676: What does this mean?

ViniciusResende: It is something that is already available at the `ToolbarService`, it prevents the execution of the command on the button initialization.

This was added to prevent a bug as stated at the commit:
> There was a bug where you'd change the brush size for a value different
from the default one, change to the eraser or threshold tool, change
it's value to a new value, when you use the keybinding for increasing or
decreasing the radius, it would use the initial set value as reference,
and not the latest one.

modes/segmentation/src/toolbarButtons.ts
ViniciusResende: Don't see a way to do this without breaking code orthogonality. 
If I just remove the min and max the UI will allow the user to select values outside of the 0.5 to 99.5 range. (Undesired)
If I make the `ToolbarService` to use the values from the Tool configuration, this would require a behavior change, as wells as a proper way to identify which option correlates with which configuration. 
I'll create a common `constant.ts` file where both files from segmentation mode can import from.

modes/segmentation/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const MIN_SEGMENTATION_DRAWING_RADIUS = 0.5;
```
jbocce, line 1: Consider adding BRUSH to both constant names.

extensions/cornerstone-dicom-seg/src/getToolbarModule.ts
ViniciusResende: Done at ccac4f5c2370ca5eadd8d96af58e56ba279c9ec9

modes/segmentation/src/constants.ts
jbocce: Or replace DRAWING with BRUSH

modes/segmentation/src/toolbarButtons.ts
ViniciusResende: Done at 8457f3ebc598092120e3d1bd3f975f1a13d03c31

modes/segmentation/src/constants.ts
ViniciusResende: Thought that the ERASER is not necessarily a BRUSH, so this is why I choose DRAWING as a more generic term. Even with that you think MIN_SEGMENTATION_BRUSH_RADIUS is more appropriate?

jbocce: Good point. Let's leave it.

tests/SEGDrawingToolsResizing.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  await page.waitForTimeout(500);
+}
+
+test('should resize segmentation brush tool', async ({ page }) => {
+  await page.getByTestId('Brush-btn').click();
+
+  await performDrawingToolInteraction(page, 'brush');
+
+  await checkForScreenshot(page, page, screenShotPaths.segDrawingToolsResizing.brushTool);
```
jbocce: As we discussed, please add an assertion for the cursor/brush radius by extracting the circle from the SVG DOM and asserting its radius. Also consider removing the screenshots as the SVG circle radius assertion covers what the screenshots are trying to accomplish.

ViniciusResende: Done at bce1c4e094439e01352ceaf27be03e99c12f551b, kept screenshots focused only on the viewport content itself. I think they're valuable enough, since they actually assert segmentation drawing interactions that aren't covered by the static assertions itself.

What do you think?

tests/SEGDrawingToolsResizing.spec.ts
jbocce, line 1: Should we test sphere brush as well?

ViniciusResende: Seems a bit redundant for me, since in the OHIF level circle brush and sphere brush behave the same in terms of UI. Do you think it is necessary? If so, sphere erasers and threshold as well?

jbocce: Ok this is fine for now. Thanks.

tests/SEGDrawingToolsResizing.spec.ts
jbocce: I think it is fine. Thank you.

jbocce (approved):
Great effort on this. Thanks.

PR #5545, opened 2025-11-05
Fix contour annotations RTSS saving

wayfarer3130:
### Context

RTSS saving wasn't working for contours.
Segmentation wasn't working on non-multiframe creatable images
RTSS and Segmentation would always save to a new series

### Changes & Results

Use CS3D contour saving
Apply series description from dialog correctly
Get the next series instance uid
Use the latest RTSS/segmentation label for saving
Increment series numbers on saving
Default to the previously saved series for at least SEG and RTSTRUCT

### Testing

Create a contour segmentation on each of:
  1. Single frame image (eg a CR or the like)
  2. Volume image (eg can use 'Juno' from the test dataset)
  3. Multiframe image (eg can use 1.2.276.0.7230010.3.1.2.862086195.1.1762719262.949510)
  4. 8 bit multiframe image (same image as above)

Save a labelmap as a segmentation on the above images, and then a contour segmentation as an RTSS

Load the segmentation/RTSTRUCT object
There are issues on load with changed appearance, but those are NOT part of this PR
The contour or labelmap should be on the same images it was before saving and should be the same general area.

Fixes: 
https://github.com/cornerstonejs/cornerstone3D/issues/2225

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


package.json
```diff
@@ -135,10 +135,14 @@
     ]
   },
   "resolutions": {
+    "@cornerstonejs/core": "link:@cornerstonejs/core",
```
sedghi: reminder for remove later

platform/ui/src/components/Button/Button.tsx
```diff
@@ -128,7 +128,11 @@ Button.propTypes = {
   /** Callback to be called when the button is clicked  */
   onClick: PropTypes.func.isRequired,
   /** Button size  */
-  size: PropTypes.oneOf([ButtonEnums.size.medium, ButtonEnums.size.small]),
+  size: PropTypes.oneOf([
```
sedghi: we don't have @ohif/ui anymore, this file should not be touched as it is legacy, use the ui-next button

platform/core/src/utils/downloadBlob.ts
```diff
@@ -0,0 +1,20 @@
...
+  const url = URL.createObjectURL(content);
+  downloadUrl(url, options);
+  URL.revokeObjectURL(url);
+}
+
+/**
+ * Downloads a URL
+ */
+export function downloadUrl(url, options?) {
+  const link = document.createElement('a');
```
sedghi, line 24: i like this can we move everytone to this then

sedghi: CleanShot 2025-11-14 at 17 52 17@2x

modes/segmentation/src/index.tsx
```diff
@@ -2,42 +2,10 @@ import { id } from './id';
...
- * Just two dependencies to be able to render a viewport with panels in order
- * to make sure that the mode is working.
- */
-const extensionDependencies = {
-  '@ohif/extension-default': '^3.0.0',
-  '@ohif/extension-cornerstone': '^3.0.0',
-  '@ohif/extension-cornerstone-dicom-seg': '^3.0.0',
-  '@ohif/extension-cornerstone-dicom-rt': '^3.0.0',
-};
+console.warn('cornerstone=', cornerstone);
```
sedghi: reminder for remove after

extensions/cornerstone/src/utils/segmentationExportUtils.ts
```diff
@@ -59,7 +59,7 @@ export const hasExportableLabelMapData = (
     SeriesInstanceUID
   );
 
-  return displaySet?.isReconstructable;
+  return !!displaySet;
```
sedghi, line 62: wait, do you mean we are solving the issue that we can't export segmentations on non reconstructibles? like dx, mg, us? etc

platform/core/src/types/DisplaySet.ts
```diff
@@ -49,6 +49,12 @@ export type DisplaySet = {
   SeriesTime?: string;
   instance?: InstanceMetadata;
 
+  /**
+   * The predecessor image id is the image id to use when replacing this instance.
+   * It will only be defined for replacable instances such as SR or SEG
+   */
+  predecessorImageId?: string;
```
sedghi, line 71: i know what you are doing, is this a good name or `sourceImageId` is better? i let you choose

platform/core/src/classes/MetadataProvider.ts
```diff
@@ -141,15 +141,16 @@ class MetadataProvider {
...
           seriesDate,
           seriesTime,
         };
         break;
       case WADO_IMAGE_LOADER_TAGS.PATIENT_STUDY_MODULE:
         metadata = {
-          patientAge: toNumber(instance.PatientAge),
-          patientSize: toNumber(instance.PatientSize),
-          patientWeight: toNumber(instance.PatientWeight),
+          patientAge: instance.PatientAge,
```
sedghi, line 138: not sure why we are removing the toNumber, i know lots of issues arise from not having number or not having string etc etc

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -366,6 +371,7 @@ class SegmentationService extends PubSubService {
       label?: string;
     }
   ): Promise<string> {
+    console.warn('******************** create contour for display set');
```
sedghi: reminder for remove at the end

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
```diff
@@ -17,19 +17,26 @@ function _getDisplaySetsFromSeries(
   servicesManager: AppTypes.ServicesManager,
   extensionManager
 ) {
-  const instance = instances[0];
+  utils.sortStudyInstances(instances);
```
sedghi, line 20: please add a migration guide note, in the rendering it would be helpful for people to know we know sort and grab last one

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
```diff
@@ -39,7 +46,8 @@ function _getDisplaySetsFromSeries(
     displaySetInstanceUID: utils.guid(),
     SeriesDescription,
     SeriesNumber,
-    SeriesDate,
+    SeriesDate: StructureSetDate || SeriesDate,
```
sedghi, line 53: is this a known thing that we should prefer the StructureSetDate over SeriesDate? same for time

extensions/default/src/Panels/createReportDialogPrompt.tsx
```diff
@@ -2,11 +2,16 @@ import PROMPT_RESPONSES from '../utils/_shared/PROMPT_RESPONSES';
 
 export default function CreateReportDialogPrompt({
   title = 'Create Report',
+  modality = 'SR',
+  minSeriesNumber,
```
sedghi: please add documentation above this function , and also migration guide about what is changed here (which is if not defined series number is picked as 4000)

extensions/cornerstone/src/utils/segmentationExportUtils.ts
wayfarer3130: Yes, that is exactly what the RTSS PR ended up doing.

platform/core/src/types/DisplaySet.ts
wayfarer3130: I like having it have the same name as the predecessor sequence, so I think I'd like to leave it if that is ok.

platform/core/src/classes/MetadataProvider.ts
wayfarer3130: Mostly it gives NaN as a response.  If we want a formatted version, then we should create secondary computations that operate on the base/default data.  There are a whole bunch of cases we return garbage data from the metadata provider, and if we find particular formatting issues, then I will create
patientFormatted
as a module name that calls patientModule and does the formatting.

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
wayfarer3130: Yes, it is a known thing - a bit ugly because the display set is really just the last instance.

platform/core/src/utils/downloadBlob.ts
wayfarer3130: Did all of them.

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
wayfarer3130: Added docs to this.

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
wayfarer3130: Added it to the base documentation as it isn't really migration but information for future usage.

platform/core/src/types/DisplaySet.ts
sedghi: Sounds good

extensions/cornerstone/src/initMeasurementService.ts
```diff
@@ -461,6 +461,7 @@ const connectMeasurementServiceToTools = ({
       const annotationManager = annotation.state.getAnnotationManager();
       const newAnnotation = {
         annotationUID: measurement.uid,
+        predecessorImageId: measurement?.predecessorImageId,
```
sedghi, line 466: We don't have this in Cornerstone, and you're adding it as if it exists. It might be just for handling state, but please add a comment above it.

extensions/default/src/getDisplaySetsFromUnsupportedSeries.js
```diff
@@ -32,7 +32,8 @@ export default function getDisplaySetsFromUnsupportedSeries(instances) {
     SOPClassUID: instance.SOPClassUID,
     SeriesDescription: instance.SeriesDescription || '',
     Modality: instance.Modality,
-    numImageFrames: instances.length,
+    instances,
+    instance: instances[instance.length - 1],
```
sedghi, line 36: people might be relying on this, either bring it back or document a migration

sedghi: I mean removing the numImageFrames

platform/core/src/classes/MetadataProvider.ts
```diff
@@ -346,16 +334,13 @@ class MetadataProvider {
...
-        const { PatientName } = instance;
-
-        let patientName;
-        if (PatientName) {
-          patientName = formatPN(PatientName);
-        }
-
         metadata = {
-          patientName,
+          patientName: instance.PatientName,
```
sedghi, line 338: if we are changing the formatting of the PatientName (not doing formatPN) we should document why in migrations

platform/core/src/services/DicomMetadataStore/DicomMetadataStore.ts
```diff
@@ -289,6 +296,7 @@ const BaseImplementation = {
   getInstance: _getInstance,
   getInstanceByImageId: _getInstanceByImageId,
   updateMetadataForSeries: _updateMetadataForSeries,
+  setMetaData: _setMetaData,
```
sedghi: `setMetadataProvider`

platform/core/src/services/DicomMetadataStore/DicomMetadataStore.ts
```diff
@@ -102,16 +108,16 @@ function _getInstance(StudyInstanceUID, SeriesInstanceUID, SOPInstanceUID) {
...
-  for (const study of _model.studies) {
-    for (const series of study.series) {
-      for (const instance of series.instances) {
-        if (instance.imageId === imageId) {
-          return instance;
-        }
-      }
-    }
-  }
+  const { studyInstanceUID, seriesInstanceUID, sopInstanceUID } = metaDataProvider.get(
```
sedghi: it should have a fallback if metadata provider does not have it to the for loop that we had

platform/core/src/services/DisplaySetService/DisplaySetService.ts
```diff
@@ -115,16 +115,52 @@ export default class DisplaySetService extends PubSubService {
     return this.activeDisplaySets;
   }
 
+  /**
+   * Gets the set of series with this series instance UID
```
sedghi: wrong documentation, get displaysets that include series instanceuid (i think ...)

platform/core/src/services/DisplaySetService/DisplaySetService.ts
```diff
@@ -115,16 +115,52 @@ export default class DisplaySetService extends PubSubService {
...
       displaySet => displaySet.SeriesInstanceUID === seriesInstanceUID
     );
   };
 
+  /**
+   * Given a reference to a series/sop, returns the set of display sets
+   * containing an instance from the references.
+   */
+  public getDisplaySetsForReference = (
+    reference: ReferencedSeriesSequence | ReferencedSeriesSequence[]
```
sedghi: strange naming since i see reference can be a list. i recommend references and always make it a list (even with one etnry) much more clear for human and ai

platform/core/src/types/DisplaySet.ts
```diff
@@ -49,8 +61,19 @@ export type DisplaySet = {
   SeriesTime?: string;
   instance?: InstanceMetadata;
 
+  /**
+   * The predecessor image id is the image id to use when replacing this instance.
```
sedghi: incorrect documentation, it is the imageid to the ... WHICH has a use case for when replacing ...

platform/core/src/types/DisplaySet.ts
```diff
@@ -49,8 +61,19 @@ export type DisplaySet = {
...
+   */
+  predecessorImageId?: string;
+
   isHydrated?: boolean;
   isRehydratable?: boolean;
+
+  measurements?: Array<any>;
+  referencedImages?: Array<string>;
+
+  isLoaded?: boolean;
```
sedghi: I wasn't expecting these. Could you please add documentation explaining what they are? I'm not a big fan of these additions, and I'd like to explore alternative solutions.

extensions/default/src/getDisplaySetsFromUnsupportedSeries.js
wayfarer3130: Actually it is a bug fix - unsupported series don't have image frames, and other display sets are already missing it when there are no image frames and all work correctly.  I can add it back in, but it causes problems on the study thumbnail browser.

platform/core/src/services/DisplaySetService/DisplaySetService.ts
wayfarer3130: Leaving the reference to be either reference or references since it comes from dcmjs and might occasionally be a singleton and I don't want the caller to have to check.
Renaming the method as requested.

PR #5557, opened 2025-11-11
tests(Playwright): Introduction of Page Object Models (POMs) to Playwright tests

ViniciusResende:
### Context
This PR introduces a new approach for dealing with playwright tests, with the use of Page Objects and fixtures we're now able to write tests with much more reusability, being capable of define simple and complex locators inside POMs that can be used across all the tests.

This is the first step of a more complex enhancement of the architecture of our Playwright tests.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] Node version: 20.9.0

### DIFF
--- a/tests/3DFourUp.spec.ts
+++ b/tests/3DFourUp.spec.ts
@@ -15,7 +15,7 @@ test.beforeEach(async ({ page }) => {
 
 test.describe('3D four up Test', async () => {
   test('should render 3D four up correctly.', async ({ page, mainToolbarPageObject }) => {
-    await mainToolbarPageObject.layoutSection.threeDFourUp.click();
+    await mainToolbarPageObject.layoutSelection.threeDFourUp.click();
 
     await attemptAction(() => reduce3DViewportSize(page), 10, 100);
 
--- a/tests/3DMain.spec.ts
+++ b/tests/3DMain.spec.ts
@@ -15,7 +15,7 @@ test.beforeEach(async ({ page }) => {
 
 test.describe('3D main Test', async () => {
   test('should render 3D main correctly.', async ({ page, mainToolbarPageObject }) => {
-    await mainToolbarPageObject.layoutSection.threeDMain.click();
+    await mainToolbarPageObject.layoutSelection.threeDMain.click();
     await attemptAction(() => reduce3DViewportSize(page), 10, 100);
     await page.waitForTimeout(5000);
     await checkForScreenshot(page, page, screenShotPaths.threeDMain.threeDMainDisplayedCorrectly);
--- a/tests/Crosshairs.spec.ts
+++ b/tests/Crosshairs.spec.ts
@@ -48,7 +48,7 @@ test.beforeEach(async ({ page }) => {
 
 test.describe('Crosshairs Test', async () => {
   test('should render the crosshairs correctly.', async ({ page, mainToolbarPageObject }) => {
-    await mainToolbarPageObject.layoutSection.MPR.click();
+    await mainToolbarPageObject.layoutSelection.MPR.click();
     await page.getByTestId('Crosshairs').click();
 
     await checkForScreenshot(page, page, screenShotPaths.crosshairs.crosshairsRendered);
@@ -58,7 +58,7 @@ test.describe('Crosshairs Test', async () => {
     page,
     mainToolbarPageObject,
   }) => {
-    await mainToolbarPageObject.layoutSection.MPR.click();
+    await mainToolbarPageObject.layoutSelection.MPR.click();
     await page.getByTestId('Crosshairs').click();
 
     await rotateCrosshairs(page, '#svg-layer-mpr-axial', 3);
@@ -72,7 +72,7 @@ test.describe('Crosshairs Test', async () => {
     page,
     mainToolbarPageObject,
   }) => {
-    await mainToolbarPageObject.layoutSection.MPR.click();
+    await mainToolbarPageObject.layoutSelection.MPR.click();
     await page.getByTestId('Crosshairs').click();
 
     await increaseSlabThickness(page, '#svg-layer-mpr-axial', 0, 'x');
@@ -86,7 +86,7 @@ test.describe('Crosshairs Test', async () => {
     page,
     mainToolbarPageObject,
   }) => {
-    await mainToolbarPageObject.layoutSection.MPR.click();
+    await mainToolbarPageObject.layoutSelection.MPR.click();
     await page.getByTestId('Crosshairs').click();
 
     await rotateCrosshairs(page, '#svg-layer-mpr-axial', 3);
@@ -102,7 +102,7 @@ test.describe('Crosshairs Test', async () => {
     page,
     mainToolbarPageObject,
   }) => {
-    await mainToolbarPageObject.layoutSection.MPR.click();
+    await mainToolbarPageObject.layoutSelection.MPR.click();
     await page.getByTestId('Crosshairs').click();
 
     await rotateCrosshairs(page, '#svg-layer-mpr-axial', 0);
--- a/tests/pages/MainToolbarPageObject.ts
+++ b/tests/pages/MainToolbarPageObject.ts
@@ -7,11 +7,11 @@ export class MainToolbarPageObject {
     this.page = page;
   }
 
-  get layoutSection() {
+  get layoutSelection() {
     const page = this.page;
 
     const button = page.getByTestId('Layout');
-    const layoutSection = {
+    const layoutSelection = {
       button,
       async click() {
         await button.click();
@@ -19,13 +19,13 @@ export class MainToolbarPageObject {
     };
 
     return {
-      ...layoutSection,
+      ...layoutSelection,
       get axialPrimary() {
         const button = page.getByTestId('Axial Primary');
         return {
           button,
           async click() {
-            await layoutSection.click();
+            await layoutSelection.click();
             await button.click();
           },
         };
@@ -35,7 +35,7 @@ export class MainToolbarPageObject {
         return {
           button,
           async click() {
-            await layoutSection.click();
+            await layoutSelection.click();
             await button.click();
           },
         };
@@ -45,7 +45,7 @@ export class MainToolbarPageObject {
         return {
           button,
           async click() {
-            await layoutSection.click();
+            await layoutSelection.click();
             await button.click();
           },
         };
@@ -55,7 +55,7 @@ export class MainToolbarPageObject {
         return {
           button,
           async click() {
-            await layoutSection.click();
+            await layoutSelection.click();
             await button.click();
           },
         };
@@ -65,7 +65,7 @@ export class MainToolbarPageObject {
         return {
           button,
           async click() {
-            await layoutSection.click();
+            await layoutSelection.click();
             await button.click();
           },
         };
@@ -75,7 +75,7 @@ export class MainToolbarPageObject {
         return {
           button,
           async click() {
-            await layoutSection.click();
+            await layoutSelection.click();
             await button.click();
           },
         };
--- a/tests/3DOnly.spec.ts
+++ b/tests/3DOnly.spec.ts
@@ -15,7 +15,7 @@ test.beforeEach(async ({ page }) => {
 
 test.describe('3D only Test', async () => {
   test('should render 3D only correctly.', async ({ page, mainToolbarPageObject }) => {
-    await mainToolbarPageObject.layoutSection.threeDOnly.click();
+    await mainToolbarPageObject.layoutSelection.threeDOnly.click();
     await attemptAction(() => reduce3DViewportSize(page), 10, 100);
     await page.waitForTimeout(5000);
     // Use a 4 percent diff pixel ratio to account for slight color differences in the 3D viewport
--- a/tests/3DPrimary.spec.ts
+++ b/tests/3DPrimary.spec.ts
@@ -15,7 +15,7 @@ test.beforeEach(async ({ page }) => {
 
 test.describe('3D primary Test', async () => {
   test('should render 3D primary correctly.', async ({ page, mainToolbarPageObject }) => {
-    await mainToolbarPageObject.layoutSection.threeDPrimary.click();
+    await mainToolbarPageObject.layoutSelection.threeDPrimary.click();
 
     await attemptAction(() => reduce3DViewportSize(page), 10, 100);
     await page.waitForTimeout(5000);
--- a/tests/AxialPrimary.spec.ts
+++ b/tests/AxialPrimary.spec.ts
@@ -8,7 +8,7 @@ test.beforeEach(async ({ page }) => {
 
 test.describe('Axial Primary Test', async () => {
   test('should render Axial Primary correctly.', async ({ page, mainToolbarPageObject }) => {
-    await mainToolbarPageObject.layoutSection.axialPrimary.click();
+    await mainToolbarPageObject.layoutSelection.axialPrimary.click();
     await checkForScreenshot(
       page,
       page,
--- a/tests/JumpToMeasurementMPR.spec.ts
+++ b/tests/JumpToMeasurementMPR.spec.ts
@@ -84,7 +84,7 @@ test('should hydrate in MPR correctly', async ({
 
   await checkForScreenshot(page, page, screenShotPaths.jumpToMeasurementMPR.jumpToMeasurementStack);
 
-  await mainToolbarPageObject.layoutSection.MPR.click();
+  await mainToolbarPageObject.layoutSelection.MPR.click();
 
   // wait 5 seconds
   await page.waitForTimeout(5000);
--- a/tests/MPR.spec.ts
+++ b/tests/MPR.spec.ts
@@ -8,7 +8,7 @@ test.beforeEach(async ({ page }) => {
 
 test.describe('MPR Test', async () => {
   test('should render MPR correctly.', async ({ page, mainToolbarPageObject }) => {
-    await mainToolbarPageObject.layoutSection.MPR.click();
+    await mainToolbarPageObject.layoutSelection.MPR.click();
     await checkForScreenshot(page, page, screenShotPaths.mpr.mprDisplayedCorrectly);
   });
 });
--- a/tests/MPRThenRTOverlayNoHydration.spec.ts
+++ b/tests/MPRThenRTOverlayNoHydration.spec.ts
@@ -12,7 +12,7 @@ test('should launch MPR with unhydrated RTSTRUCT chosen from the data overlay me
   mainToolbarPageObject,
   viewportPageObject,
 }) => {
-  await mainToolbarPageObject.layoutSection.MPR.click();
+  await mainToolbarPageObject.layoutSelection.MPR.click();
 
   // Wait 5 seconds for MPR to load. This is necessary in particular when screen shots are added or replaced.
   await page.waitForTimeout(10000);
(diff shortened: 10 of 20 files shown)


tests/pages/MainToolbarPageObject.ts
jbocce, line 1: I would rather have the suffix for ALL page objects be `PageObject` instead of just `Page`.

tests/pages/MainToolbarPage.ts
```diff
@@ -0,0 +1,261 @@
+import { Page } from '@playwright/test';
+
+export class MainToolbarPage {
+  readonly page: Page;
+
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  get layout() {
```
jbocce: Please rename this `layoutSelector`. `layout` has way too many different meanings.

tests/pages/MainToolbarPageObject.ts
ViniciusResende: I agree with that one, indeed I think it will make it clearer. Good advice.

tests/pages/MainToolbarPage.ts
ViniciusResende: I don't know if `layoutSelector` truly represents what it is. Although it can work as a selector for the layout button, it also wraps everything inside of it. 
This will always be used with the `MainToolbarPageObject.layout`, so at least for me the `MainToolbar` prefix tells me that I'm selecting a tool.

When I perform:
```js
await mainToolbarPage.layout.MPR.click();
```

I don't think there is any ambiguity on it, do you think otherwise?

ViniciusResende: Actually, I think that you mention `layoutSelector` not as the query selector for the layout element, but as the layout selection section haha.

Maybe call it `layoutSelection` is a good option too?

tests/pages/MainToolbarPage.ts
```diff
@@ -0,0 +1,261 @@
...
+    };
+
+    return {
+      ...layout,
+      get axialPrimary() {
+        return {
+          button: page.getByTestId('Axial Primary'),
+          async click() {
+            await layout.click();
+            await page.getByTestId('Axial Primary').click();
```
jbocce: Can we do `this.button.click();` instead?

tests/pages/MainToolbarPage.ts
jbocce: Hahaha. Yes that is what I meant. `layoutButton` or `layoutSelection` I suppose are both fine.

tests/pages/MainToolbarPage.ts
```diff
@@ -0,0 +1,261 @@
...
+    this.page = page;
+  }
+
+  get layout() {
+    const page = this.page;
+
+    const layout = {
+      button: page.getByTestId('Layout'),
+      async click() {
+        await page.getByTestId('Layout').click();
```
jbocce: Can we do `this.button.click()` instead?

tests/pages/MainToolbarPage.ts
ViniciusResende: So, it is possible. But I personally don't know if I like it, because we're inside a class. And in that moment we'll be using the keyword `this` not to reference the class but the object in which the functions in being defined. So we have the same keyword being used for different purposes. Do you think this is fine? 

Another option is to do like that:

```js
get axialPrimary() {
  const axialPrimaryButton = page.getByTestId('Axial Primary');
  return {
    button: axialPrimaryButton,
    async click() {
      await layout.click();
      await axialPrimaryButton.click();
    },
  };
},
```

What do you think?

jbocce: I do like your suggestion, but insead of using the constant name `axialPrimaryButton`, can we simply use `button` instead?

tests/pages/ViewportGridPageObject.ts
jbocce, line 1: So this PageObject is a little bit different than the MainToolbarPageObject. The primary difference I see is that each sub-section (or as I like to call them each sub page object) is simply a locator. For the MainToolbarPageObject, each sub page object was indeed a page object retured by various getters. I like what you did in MainToolbarPageObject and I think we should do the same here.

So in this case for example, `activeViewport` should return an object with the various getters for it. For example, I could see a getter for the SVG elements for a viewport some day. For now there may only be one memeber of the object - the locator. In fact, I think we should also have a `ViewportPageObject` that is returned by some of these getters.

I am also going to go a step further and suggest maybe that each PageObject should have a special `locator` getter that returns the locator for it. You used `button` for this special getter in the MainToolbarPageObject and here you named it `viewportGrid`, but if we come up with a convention then that will make a developer's life easier.

tests/utils/fixture.ts
```diff
@@ -0,0 +1,18 @@
+import { test as base } from 'playwright-test-coverage';
+import { MainToolbarPage, ViewportGridPage } from '../pages';
+
+type PageObjects = {
+  mainToolbarPage: MainToolbarPage;
+  viewportGridPage: ViewportGridPage;
+};
+
+export const test = base.extend<PageObjects>({
+  mainToolbarPage: async ({ page }, use) => {
```
jbocce: We will need a way to set these top-level page objecs to `null` or `undefined` when they are not available (e.g. if we put side-panels at this level, then some tests might not have certain side panels). This will allow the tests that don't use a specific top-level page object not to fail for no good reason.

tests/utils/fixture.ts
```diff
@@ -0,0 +1,18 @@
...
+    await use(new MainToolbarPage(page));
+  },
+  viewportGridPage: async ({ page }, use) => {
+    await use(new ViewportGridPage(page));
+  },
+});
+
+export { expect } from 'playwright-test-coverage';
```
jbocce, line 18: Why do we have to do this? Can't individual test files simply import this directly from `playwright-test-coverage`? I'm sure there are other similar things to import from playwright that I hope we don't have to do this for.

tests/utils/locators.ts
```diff
@@ -1,15 +1,5 @@
...
-  page,
-  viewportId,
-}: {
-  page: Page;
-  viewportId: string;
-}): Locator => {
-  return page.locator(`css=div[data-viewportid="${viewportId}"]`);
-};
-
 export const viewportSVGLocator = ({
```
jbocce: These will go nicely into the `ViewportPageObject`. 😊 That said maybe even there is an SVGPageObject too!

tests/3DFourUp.spec.ts
```diff
@@ -14,13 +14,8 @@ test.beforeEach(async ({ page }) => {
...
 test.describe('3D four up Test', async () => {
-  test('should render 3D four up correctly.', async ({ page }) => {
-    await page.getByTestId('Layout').click();
-    await page
-      .locator('div')
-      .filter({ hasText: /^3D four up$/ })
-      .first()
-      .click();
+  test('should render 3D four up correctly.', async ({ page, mainToolbarPage }) => {
+    await mainToolbarPage.layout.threeDFourUp.click();
```
jbocce: ❤️

tests/Crosshairs.spec.ts
```diff
@@ -1,17 +1,18 @@
...
   getMousePosition,
+  initializeMousePositionTracker,
+  screenShotPaths,
+  test,
+  visitStudy,
 } from './utils/index.js';
 
 const rotateCrosshairs = async (page: Page, id: string, lineNumber: number) => {
-  const locator = await page.locator(id).locator('line').nth(lineNumber);
+  const locator = page.locator(id).locator('line').nth(lineNumber);
```
jbocce, line 12: why do we no longer need an await here? I see this in a few other places too.

tests/TMTVAlignment.spec.ts
```diff
@@ -1,54 +1,55 @@
...
+import { expect, scrollVolumeViewport, test, visitStudy } from './utils';
 
-test.skip('PT should show slice closest to CT', async ({ page }) => {
+test.skip('PT should show slice closest to CT', async ({ page, viewportGridPage }) => {
   const studyInstanceUID = '1.2.840.113619.2.290.3.3767434740.226.1600859119.501';
   const mode = 'tmtv';
   await visitStudy(page, studyInstanceUID, mode);
 
-  const vp = page.getByTestId('viewport-pane');
+  const getNthVP = (index: number) => viewportGridPage.getNthViewportPane(index);
```
jbocce: Rename this getNthViepwortPane.

tests/pages/ViewportGridPage.ts
```diff
@@ -0,0 +1,29 @@
+import { Page, Locator } from '@playwright/test';
+
+export class ViewportGridPage {
+  readonly page: Page;
+
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  get activeViewport() {
```
jbocce: After thinking about this for a bit. I think each of these getters other than `viewportGrid` return, in essence, a viewport so adding the word `pane` will only lead to confusion. Let's get rid of the `pane`! As I get older, I love getting rid of the pain! 🤣

tests/pages/MainToolbarPageObject.ts
ViniciusResende: Done at da7b88e187c52b91d85bd3462a423a752869e77a

tests/pages/MainToolbarPage.ts
ViniciusResende: Done at ef3102875e9fca85767d9cbfba3031069f69a3d4

tests/Crosshairs.spec.ts
ViniciusResende: The act of getting the locator itself is not asynchronous, only the interaction with it that is. So, I removed since they were just adding complexity to te code unnecessarily.

tests/TMTVAlignment.spec.ts
ViniciusResende: Done at 51b20a00022e49848599275f3c052dee7e685d13

tests/utils/locators.ts
ViniciusResende: Yeah, this is why I didn't move these guys. I don't know if they'll fit nicely inside the `ViewportGridPageObject`, or if they deserve their own Page Object. I think they might fit on something called `ViewportOverlayPageObject`, that would cover all overlay buttons, text and SVG on top of the viewports.

What do you think? This should be handled as an additional Page Object or as subsections inside `ViewportGridPageObject`?

tests/utils/fixture.ts
ViniciusResende: It is not necessary, is just so that when we import, we can import both `test` and `expect` from `./utils`. I saw that is a common practice from the Playwright community to avoid importing from multiple sources.

Also, we could extend with fixtures the `expect` too in a given moment.

Even with that, do you think is best to remove it from here and import the expect separately?

tests/pages/MainToolbarPage.ts
ViniciusResende: Done at c78cc23325112b09c89842a301707437488b4244

tests/pages/MainToolbarPage.ts
ViniciusResende: Done at c78cc23325112b09c89842a301707437488b4244

tests/utils/fixture.ts
ViniciusResende: Hmmm I see what you mean, but instantiating the Page Object classes by themselves is not enough for a test to fail. And I don't see as a problem when I try to access something from a `LeftPanelPageObject`, and it is not visible, then the test fails. I see that as the desired behavior.

You suggest that we do some verification to instantiate the class only if the element is visible? In the way fixtures work I don't know if this is feasible.

tests/pages/ViewportGridPage.ts
ViniciusResende: Done at f197aec760ec2b36f1adf26102a18bbf95f575e9

tests/pages/ViewportGridPageObject.ts
ViniciusResende: I see what you mean, but I did that particularly for the `MainToolbarPageObject` due to the nature that, for performing a click on an element, I might need to open some sub-menu first.

But I agree that is good to have a similar experience across all Page Objects so one could be consistent using them. May we sync to decide about it?

tests/utils/fixture.ts
jbocce: Let's leave it out until we extend it in the future.

tests/utils/fixture.ts
jbocce: Upon further discussion and analysis, the way you currently have it is a great way to proceed since there is no dependency on the existence of the underlying elements of the page objects in the page object constructors. Nice work!

tests/3DMain.spec.ts
```diff
@@ -14,13 +14,8 @@ test.beforeEach(async ({ page }) => {
...
 test.describe('3D main Test', async () => {
-  test('should render 3D main correctly.', async ({ page }) => {
-    await page.getByTestId('Layout').click();
-    await page
-      .locator('div')
-      .filter({ hasText: /^3D main$/ })
-      .first()
-      .click();
+  test('should render 3D main correctly.', async ({ page, mainToolbarPageObject }) => {
+    await mainToolbarPageObject.layoutSection.threeDMain.click();
```
sedghi: are these typed? meaning do we have autocomplete when we do mainToolbarPageObject.

ViniciusResende: Yes, we do have auto complete.

sedghi: Noice

tests/pages/ViewportPageObject.ts
```diff
@@ -0,0 +1,192 @@
...
+        };
+      },
+    };
+  }
+
+  private getSvg(viewport: Locator, innerElement?: SvgInnerElement) {
+    return viewport.locator(`svg.svg-layer${innerElement ? ` ${innerElement}` : ''}`);
+  }
+
+  public viewportPageObjectFactory(viewport: Locator): IViewportPageObject {
```
jbocce: I don't think this should be public.

tests/pages/ViewportPageObject.ts
```diff
@@ -0,0 +1,192 @@
...
+    return {
+      doubleClickOn: async (point: { x: number; y: number }) => {
```
jbocce: doubleClickAt might be a better name.

tests/pages/ViewportPageObject.ts
```diff
@@ -0,0 +1,192 @@
...
+        await simulateDoubleClickOnElement({
+          locator: viewport,
+          point,
+        });
+      },
+      clicksOn: async (
```
jbocce: clickAt

tests/pages/ViewportPageObject.ts
```diff
@@ -0,0 +1,192 @@
...
+        points: { x: number; y: number }[],
+        button?: 'left' | 'right' | 'middle'
+      ) => {
+        await simulateClicksOnElement({
+          locator: viewport,
+          points,
+          button,
+        });
+      },
+      normalizedClicksOn: async (
```
jbocce: normalizedClickAt... In general use `click` and `at`... no need for plurals.

tests/Bidirectional.spec.ts
```diff
@@ -1,30 +1,21 @@
...
+  page,
+  mainToolbarPageObject,
+  viewportPageObject,
+}) => {
+  await mainToolbarPageObject.measurementTools.bidirectional.click();
+  await viewportPageObject.active.clicksOn([
+    { x: 405, y: 277 },
+    { x: 515, y: 339 },
+  ]);
   await page.getByTestId('prompt-begin-tracking-yes-btn').click();
```
jbocce, line 19: Here is another candidate for a PageObject? Only question is what page object should it be fetched from? Since it appears in a viewport, that would be my first choice.

tests/pages/ViewportPageObject.ts
ViniciusResende: I thought on you maybe using your custom locator to get a viewport element, and them you'd be able to get all the features of the `ViewportPageObject` on top of this custom locator. But given the YAGNI rule I'll move it to private for now.

tests/Bidirectional.spec.ts
ViniciusResende: I don't agree with that logic. I think that, even though it appears in the top of the viewport, it is not part of the viewport itself. 
For example, the Data Overlay menu, it is opened through the viewport, so I create an API for that. But the interaction with the menu itself must be held by another Page Object. I don't know if it can be instantiated inside the `ViewportPageObject` or it could be in the top level as the other ones.

But if we put everything inside the viewport page object it will eventually become a massive class. I think that, more than thinking only on the UI location of things, we could be thinking on the responsibility separation of these units.

tests/pages/ViewportPageObject.ts
ViniciusResende: Agree, I was following the convention of the utilities that were previously available.

Fixed at 44163a097039f6da0130dd3ea78bb6c4085475c0.

tests/pages/ViewportPageObject.ts
ViniciusResende: Fixed at 44163a097039f6da0130dd3ea78bb6c4085475c0.

tests/pages/ViewportPageObject.ts
ViniciusResende: Fixed at 44163a097039f6da0130dd3ea78bb6c4085475c0.

tests/pages/ViewportPageObject.ts
ViniciusResende: Fixed at 6ec2ab8aef0fe245ff3757ce32c2257513857b02

tests/Bidirectional.spec.ts
jbocce: To be clear. I meant that the tracking banner (or whatever we want to call it) should be its own PageObject with it containing the intereaction for itself. I suppose where you get it from does not matter too much. However if there were say one per viewport, then instantiating it or getting it from the viewport makes most sense to me.

tests/pages/MainToolbarPageObject.ts
```diff
@@ -0,0 +1,287 @@
+import { Page } from '@playwright/test';
+
+export class MainToolbarPageObject {
+  readonly page: Page;
+
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  get layoutSection() {
```
jbocce: I really don't like "section" part of the name here. Maybe `layoutSelection` is better?

tests/SEGDataOverlayForUnreferencedDisplaySetNoHydration.spec.ts
```diff
@@ -11,10 +10,11 @@ test.beforeEach(async ({ page }) => {
 
 test('should overlay an unhydrated SEG over a display set that the SEG does NOT reference', async ({
   page,
+  viewportPageObject,
 }) => {
   await page.getByTestId('study-browser-thumbnail').nth(2).dblclick();
 
-  await page.getByTestId('dataOverlayMenu-default-btn').click();
+  await viewportPageObject.getById('default').overlayMenu.dataOverlay.click();
   await page.getByTestId('AddSegmentationDataOverlay-default').click();
```
jbocce, line 18: I think one of the next items we should do in the next PR (along with other stuff too) is add any remaining page objects. Something to keep in mind.

tests/pages/MainToolbarPageObject.ts
ViniciusResende: Done at 67651971754f58e056225dae2b78813366c330e8

tests/Bidirectional.spec.ts
ViniciusResende: Yes, I do agree with that. Good thing to the next iterations of this effort.

tests/SEGDataOverlayForUnreferencedDisplaySetNoHydration.spec.ts
ViniciusResende: Indeed, same thing as we've mentioned on this thread https://github.com/OHIF/Viewers/pull/5557#discussion_r2543449952

PR #5595, opened 2025-11-27
fix: sculptor tool fixes

wayfarer3130:
### Context

Just updates the @cornerstonejs build to fix the sculptor tool

### Testing

Open a study and create a contour segmentation on an image
Draw with the freehand segmentation draw tool
Modify the freehand segmentation with the sculptor tool and it should create internal lines.

Note there are a few known bugs still in the sculptor tool:

- Creation of new empty regions inside existing ones doesn't work
- Splitting a region by drawing through it creates odd overlaps
- Regions can be sized to size zero

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


.github/workflows/playwright.yml
```diff
@@ -23,6 +23,8 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
+      - name: Install Yarn
```
jbocce, line 26: So did we do this in CS3D as well? I thought we just used bun instead there. We should be consistent in any case.

jbocce (approved):

PR #5608, opened 2025-12-03
tests(Playwright): Full implementation of Page Object Models (POMs) to Playwright tests

ViniciusResende:
### Context
This PR finalizes the effort that was introduced at #5557, of the introduction of POMs to our tests, in a way that we'll benefit in terms of readability and reusability, allowing our test structure to grow without prejudicing maintainability. 

This concludes the first step of the architecture enhancement of Playwright tests.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] Node version: 20.9.0

### DIFF
--- a/tests/utils/assertions.ts
+++ b/tests/utils/assertions.ts
@@ -13,8 +13,8 @@ export async function assertNumberOfModalityLoadBadges({
   page: Page;
   expectedCount: number;
 }) {
-  const DOMOverlayPageObject = new DOMOverlayPageObject(page);
-  const count = await DOMOverlayPageObject.viewport.getModalityLoadBadgeCount();
+  const domOverlayPageObject = new DOMOverlayPageObject(page);
+  const count = await domOverlayPageObject.viewport.getModalityLoadBadgeCount();
 
   expect(count).toBe(expectedCount);
 }
--- a/tests/pages/ViewportPageObject.ts
+++ b/tests/pages/ViewportPageObject.ts
@@ -73,7 +73,7 @@ export class ViewportPageObject {
 
   private getAnnotation(viewport: Locator, nth: number) {
     const page = this.page;
-    const DOMOverlayPageObject = new DOMOverlayPageObject(page);
+    const domOverlayPageObject = new DOMOverlayPageObject(page);
     const annotation = viewport.locator('g[data-annotation-uid]').nth(nth);
 
     return {
@@ -83,7 +83,7 @@ export class ViewportPageObject {
       },
       contextMenu: {
         open: async () => {
-          await DOMOverlayPageObject.viewport.annotationContextMenu.open(annotation);
+          await domOverlayPageObject.viewport.annotationContextMenu.open(annotation);
         },
       },
     };


sedghi:
@jbocce what is the status of this

jbocce:
> @jbocce what is the status of this

@sedghi it just got created. I will likely look at it next week.

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/untrackSeriesModal.tsx
```diff
@@ -12,8 +12,14 @@ export function UntrackSeriesModal({ hide, onConfirm, message }) {
       </div>
       <FooterAction className="mt-4">
         <FooterAction.Right>
-          <FooterAction.Secondary onClick={hide}>Cancel</FooterAction.Secondary>
+          <FooterAction.Secondary
+            dataCY="cancel-button"
```
jbocce: Consider making this more specific, like 'untracked-series-modal-cancel-button'

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/untrackSeriesModal.tsx
```diff
@@ -12,8 +12,14 @@ export function UntrackSeriesModal({ hide, onConfirm, message }) {
...
+            onClick={hide}
+          >
+            Cancel
+          </FooterAction.Secondary>
           <FooterAction.Primary
+            dataCY="confirm-button"
```
jbocce: This one too.

platform/ui-next/src/components/DataRow/DataRow.tsx
```diff
@@ -313,6 +313,7 @@ const DataRowComponent = React.forwardRef<HTMLDivElement, DataRowProps>(
                 isSelected || !isVisible ? 'opacity-100' : 'opacity-0 group-hover:opacity-100'
               }`}
               aria-label={isVisible ? 'Hide' : 'Show'}
+              dataCY="visibilityToggle"
```
jbocce: More specific.

platform/ui-next/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -88,7 +88,12 @@ const Thumbnail = ({
                   loadingProgress && loadingProgress < 1 && 'bg-primary/25'
                 )}
               ></div>
-              <div className="text-[11px] font-semibold text-white">{modality}</div>
+              <div
+                className="text-[11px] font-semibold text-white"
+                data-cy="series-modality-label"
```
jbocce, line 93: If this is for a particular series consider adding something to distinguish it from the other series thumbnail, like series instance uid or something. If the locator though is straightforward, then maybe this is not so big a deal, but something to consider. This comment applies to the other data-cy values in this file too.

tests/pages/DOMOverlayPageObject.ts
```diff
@@ -0,0 +1,137 @@
...
+          save: async () => {
+            await saveButton.click();
+          },
+        };
+      },
+      title: page.locator('[role="dialog"] h2'),
+    };
+  }
+
+  get viewport() {
```
jbocce, line 53: I am a little surprised to see viewport here. Do we need this level of indirection? And why are things like the overlay menu not here? It is a bit confusing. And why not the context menu that is above too? I think I would expect to see everything that appears in a viewport in the viewport PageObject regardless of where it is in the DOM. Something to consider anyway.

tests/pages/RightPanelPageObject.ts
```diff
@@ -0,0 +1,267 @@
...
+        const button = page.getByTestId('addSegmentation');
+        return {
+          button,
+          click: async () => {
+            await button.click();
+          },
+        };
+      },
+      get contourMenuButton() {
+        const button = page.getByTestId('panelSegmentationWithToolsContour-btn');
```
jbocce: This button and the labelMapMenuButton below are NOT part of the segmentation panel, but instead are the tabs/buttons to change what is displayed in the right panel. So please move these accessors directly into the RightPanelPageObject. Or maybe just rename them panelTabButton to make it clear these are the buttons to switch to that panel.

jbocce: I think we should have two segmentation panel page objects: one for contour and one for label map. They have different tools and are actually two separate panels.

tests/pages/DataOverlayPageObject.ts
```diff
@@ -0,0 +1,33 @@
+import { Page } from '@playwright/test';
+
+export class DataOverlayPageObject {
+  readonly page: Page;
+
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  get button() {
```
jbocce: This is weird to have button always get the first dataOverlayMenu. Is it being used? If so rename to make it clear that it is a specific (i.e. the first) instance. If not please remove.

tests/pages/DataOverlayPageObject.ts
```diff
@@ -0,0 +1,33 @@
...
+    await this.page.getByText('SELECT A SEGMENTATION').click();
+    await this.page.getByTestId(segmentation).click();
+  }
+
+  async changeSegmentation(from: string, to: string) {
+    await this.page.getByTestId(`overlay-ds-select-value-${from.toUpperCase()}`).click();
+    await this.page.getByTestId(`${to}-SEG`).click();
+  }
+
+  async remove() {
```
jbocce: This currently only ever removes the segmentation named (literally) SEGMENTATION. The name of the segmentation overlay should be passed in and templated into the selectors/locators.

tests/utils/assertions.ts
```diff
@@ -11,5 +13,8 @@ export async function assertNumberOfModalityLoadBadges({
   page: Page;
   expectedCount: number;
 }) {
-  await expect(page.locator('css=div[data-cy^="ModalityLoadBadge-"]')).toHaveCount(expectedCount);
+  const overlayPageObject = new OverlayPageObject(page);
```
jbocce: Shouldn't this page object be one of those that get created in fixture.ts?

tests/SegmentationPanel.spec.ts
```diff
@@ -7,95 +7,72 @@ test.beforeEach(async ({ page }) => {
...
-  await expect(renameDialog).toBeVisible();
-
-  const renameInput = page.getByRole('textbox', { name: 'Enter new label' });
-  await renameInput.fill('Segment One');
-
-  await page.getByTestId('input-dialog-save-button').click();
-
-  await expect(page.getByTestId('data-row')).toContainText('Segment One');
-  await expect(page.getByTestId('data-row')).not.toContainText('Segment 1');
+  await expect(segment1.locator).toContainText('Segment One');
```
jbocce, line 29: I like how this was all simplified. Good work!

jbocce (changes requested):
Great changes. A few comments to consider. Thanks.

platform/ui-next/src/components/Thumbnail/Thumbnail.tsx
ViniciusResende: Not actually, the idea is to be generic, and them from the selectors I can filter the one I want based on the Modality or Series Name, this is the pattern I'm more used to see on Cypress/Selenium tests at least.

tests/pages/DOMOverlayPageObject.ts
ViniciusResende: > I am a little surprised to see viewport here. Do we need this level of indirection?

So, the idea of the OverlayPageObject is to hold the API for interaction with any floating elements, being it dialogs, notifications, context menu etc.

> I think I would expect to see everything that appears in a viewport in the viewport PageObject regardless of where it is in the DOM.

So, the viewport there refers to where the overlay is shown, I thought it would be more readable that way, but I could remove this wrapping layer, what do you think?

I think that if everything, including the overlays that appear on top of the viewport , is contained inside the viewport page object, it would be a bit crowded. But I could instantiate the OverlayPageObject there, so it would be acessible through the ViewportPageObject too. This is what I did with the DataOverlayPageObject, the idea was to wrap it into a different PageObject so the Viewport one wouldn't grow much in complexity.

> And why not the context menu that is above too?

The context menu API was done in a way that you can open it in a given Locator element, not necessarily only in the viewport.

> And why are things like the overlay menu not here?

You mean the Data Overlay? The one to render the Segmentations?

tests/pages/RightPanelPageObject.ts
ViniciusResende: I see that making sense, I'll adapt accordingly.

tests/utils/assertions.ts
ViniciusResende: Yes, but the utility does not have access to it directly. And I preferred instantiating the class by using the page object it already received, than receiving more parameters in the function. I think that for the user of the utility it looks cleaner. What do you think?

tests/pages/DataOverlayPageObject.ts
ViniciusResende: Fixed at c34812656144de104c4e303291d58e4a5e41c8cb

tests/pages/DataOverlayPageObject.ts
ViniciusResende: Fixed at 34214db680af3c232af908c071bd78bb993bd22b

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/untrackSeriesModal.tsx
ViniciusResende: Done at 3b9a049218879b47b00f25ca1e809b7722a86717

extensions/measurement-tracking/src/panels/PanelStudyBrowserTracking/untrackSeriesModal.tsx
ViniciusResende: Done at 3b9a049218879b47b00f25ca1e809b7722a86717

platform/ui-next/src/components/DataRow/DataRow.tsx
ViniciusResende: Done at 1d7d875b92b7d0772f3c6b434bd251c8b4977048

tests/pages/RightPanelPageObject.ts
ViniciusResende: Done at 6644c97e6578b33bfaa4e1aaec10245338b1192c

tests/pages/DOMOverlayPageObject.ts
jbocce: > > I am a little surprised to see viewport here. Do we need this level of indirection?
> 
> So, the idea of the OverlayPageObject is to hold the API for interaction with any floating elements, being it dialogs, notifications, context menu etc.

I see ok. I think we need a better name so that it is clear what this OverlayPageObject applies to. Overlay is used to mean way too many things. I can't thing of a good name. Maybe DOMBodyPageObject?

> 
> > I think I would expect to see everything that appears in a viewport in the viewport PageObject regardless of where it is in the DOM.
> 
> So, the viewport there refers to where the overlay is shown, I thought it would be more readable that way, but I could remove this wrapping layer, what do you think?

If you change the name of the OverlayPageObject, then you can leave as is.

> 
> I think that if everything, including the overlays that appear on top of the viewport , is contained inside the viewport page object, it would be a bit crowded. But I could instantiate the OverlayPageObject there, so it would be acessible through the ViewportPageObject too. This is what I did with the DataOverlayPageObject, the idea was to wrap it into a different PageObject so the Viewport one wouldn't grow much in complexity.
> 
> > And why not the context menu that is above too?
> 
> The context menu API was done in a way that you can open it in a given Locator element, not necessarily only in the viewport.

I see what you mean. However, the page object for the context menu now is very specific to the viewport wouldn't you say? What would you plan to do if there is a context menu for the thumnails? Would that context menu page object have ALL possible options from EVERY context menu? That would be a bit strange.
> 
> > And why are things like the overlay menu not here?
> 
> You mean the Data Overlay? The one to render the Segmentations?

tests/utils/assertions.ts
jbocce: So why are some page objects created upfront and others are not? This might be confusing to some. For example, I think the main toolbar is created in fixture AND in utilities. It is a bit confusing.

ViniciusResende: The idea is that the Page Objects are classes but not necessarily singletons, so they can have multiple instances. I've chosen not to use the Singleton pattern because the class objects are basically stateless. So, it seemed as a good approach to me, so I don't get blocked by the need to use the same object from the class if they don't have any state dependency. 

So, the MainToolbarPageObject is the class, and I instantiate different objects of this class, since they're all stateless (the state is the DOM itself) they're the same, independently on where they are instanciated/created.

tests/pages/DOMOverlayPageObject.ts
ViniciusResende: > I see ok. I think we need a better name so that it is clear what this OverlayPageObject applies to. Overlay is used to mean way too many things. I can't thing of a good name. Maybe DOMBodyPageObject?

DOMBodyPageObject sound like a huge god class to me haha. The idea of the OverlayPageObject is that it would held floating elements that weren't necessarily attached to anything. But I get what you mean with the DOMBody, because they all are held by it. `FloatingPageObject`? `FloatingElementsPageObject`? `AbsolutePageObject`? 

> I see what you mean. However, the page object for the context menu now is very specific to the viewport wouldn't you say? What would you plan to do if there is a context menu for the thumnails? Would that context menu page object have ALL possible options from EVERY context menu? That would be a bit strange.

Yeah, I think you're right with that. Makes sense to move it inside the viewport sub section. Does it sound right?

jbocce: > > I see ok. I think we need a better name so that it is clear what this OverlayPageObject applies to. Overlay is used to mean way too many things. I can't thing of a good name. Maybe DOMBodyPageObject?
> 
> DOMBodyPageObject sound like a huge god class to me haha. The idea of the OverlayPageObject is that it would held floating elements that weren't necessarily attached to anything. But I get what you mean with the DOMBody, because they all are held by it. `FloatingPageObject`? `FloatingElementsPageObject`? `AbsolutePageObject`?
> 

FloatingElementsPageObject sounds ok.

> > I see what you mean. However, the page object for the context menu now is very specific to the viewport wouldn't you say? What would you plan to do if there is a context menu for the thumnails? Would that context menu page object have ALL possible options from EVERY context menu? That would be a bit strange.
> 
> Yeah, I think you're right with that. Makes sense to move it inside the viewport sub section. Does it sound right?

Yes. I think so.

ViniciusResende: Fixed at 15c7997172f86fe867a5d39cacee26c4f9d9d190 and 7f6a72473d08ef07573fafdf4ed8597d01afcbf5

tests/pages/FloatingElementsPageObject.ts
```diff
@@ -0,0 +1,134 @@
+import { Locator, Page } from '@playwright/test';
+
+export class FloatingElementsPageObject {
```
jbocce: As we discussed let's rename this to DOMOverlayPageObject.

ViniciusResende: Done at 3d68438b108ca2e3cf75a06d6eb8d824ab52c43b

jbocce (approved):
Fantastic stuff. Thank you so so much. Muito obrigado.

sedghi:
@ViniciusResenGde This is great contribution thank you so much for spending time to fixing our testing.

PR #5611, opened 2025-12-04
fix/RTStructInlineBinaryContourData

dxlin:
### Context
Certain contours in an RTStruct may present a very large array that does not get converted into an array and placed into ContourData field in an ROIContour object retrieved from PACS.

Current behaviour can be observed:
https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.12201.1091.126683095609223531686845324113579088978

Attempting to load FOR1 RTStruct on default image (CT FoR1) will result in error.

This fix checks for these situations and does the conversion to array. Result is above RTStruct will load. This issue could potentially also be resolved up stream in cornerstonejs.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

### DIFF
--- a/extensions/cornerstone-dicom-rt/src/loadRTStruct.js
+++ b/extensions/cornerstone-dicom-rt/src/loadRTStruct.js
@@ -80,7 +80,7 @@ async function checkAndLoadContourData({
 
         // Ensure strictly that we have a full set of 3 coordinates
         if (rawValues.length % 3 !== 0) {
-          return Promise.reject('ContourData raw values not divisibe by 3');
+          return Promise.reject('ContourData raw values not divisible by 3');
         }
 
         for (let i = 0; i < rawValues.length; i += 3) {


extensions/cornerstone-dicom-rt/src/loadRTStruct.js
```diff
@@ -69,6 +69,39 @@ async function checkAndLoadContourData({
...
+          typeof atob === 'function'
+            ? atob(base64String)
+            : Buffer.from(base64String, 'base64').toString('utf-8');
+
+        const rawValues = decodedText.split('\\');
+
+        const result = [];
+        for (let i = 0; i < rawValues.length; i += 3) {
+          // Ensure strictly that we have a full set of 3 coordinates
+          if (i + 2 < rawValues.length) {
```
jbocce, line 87: Perhaps log a warning or error if this is not true?

extensions/cornerstone-dicom-rt/src/loadRTStruct.js
```diff
@@ -69,6 +69,39 @@ async function checkAndLoadContourData({
...
+            const y = parseFloat(rawValues[i + 1]);
+            const z = parseFloat(rawValues[i + 2]);
+
+            // Only push if all three are valid numbers (filters out trailing empty splits)
+            if (!isNaN(x) && !isNaN(y) && !isNaN(z)) {
+              result.push(x);
+              result.push(y);
+              result.push(z);
+            } else {
+              // Ignore for now
```
jbocce: I really think we should log something here for sure.

extensions/cornerstone-dicom-rt/src/loadRTStruct.js
```diff
@@ -69,6 +69,39 @@ async function checkAndLoadContourData({
         promisesMap.has(referencedROINumber)
           ? promisesMap.get(referencedROINumber).push(bulkDataPromise)
           : promisesMap.set(referencedROINumber, [bulkDataPromise]);
+      } else if (contourData && contourData.InlineBinary) {
+        // Contour data is still in binary format, conversion needed
+        const base64String = contourData.InlineBinary;
+        const decodedText =
+          typeof atob === 'function'
```
jbocce: I don't think there is any need for this check. Is there a particular reason you added it that I am missing?

jbocce (changes requested):
Thanks for this. It is very much appreciated considering it is not something we are currently handling.

extensions/cornerstone-dicom-rt/src/loadRTStruct.js
jbocce: I think `atob` is generally available.

dxlin: good point, i've removed the check and went straight to atob.

extensions/cornerstone-dicom-rt/src/loadRTStruct.js
dxlin: just doing a simple console log, let me know if something more specific is desired.

extensions/cornerstone-dicom-rt/src/loadRTStruct.js
```diff
@@ -69,6 +69,36 @@ async function checkAndLoadContourData({
...
+            const z = parseFloat(rawValues[i + 2]);
+
+            // Only push if all three are valid numbers (filters out trailing empty splits)
+            if (!isNaN(x) && !isNaN(y) && !isNaN(z)) {
+              result.push(x);
+              result.push(y);
+              result.push(z);
+            } else {
+              // Ignore for now, log error to console
+              console.log('Error parsing contourData from InlineBinary format');
```
jbocce: I think this should be `log.warn`

extensions/cornerstone-dicom-rt/src/loadRTStruct.js
jbocce: Please add an else for this that does something like...

`console.warn('InlineBinary for RTStruct is not divisible by three');`

dxlin:
modified to return a Promise.reject with the error note (either set divisible by 3 or NaN parsed).

Issue #5597, opened 2025-11-28
[Bug] Annotation incorrectly appears in the active viewport after switching series with different Frame of Reference UID


Devu-trenser:
### Describe the Bug

When working with multiple series that have different Frame of Reference UIDs, annotations drawn in one series are incorrectly re-applied to unrelated viewports after a new series is loaded.
After drawing an annotation in axial orientation of Series-1 and then dragging/dropping Series-2 (different FOR UID) into the viewports, clicking the stored annotation from the measurement panel causes the following incorrect behavior:

- The annotation to be added both to the intended axial viewport (correct) and to the currently active viewport (incorrect).
- The active viewport’s orientation to be replaced by an axial orientation, resulting in MPR becoming 2 axial + 1 axial, instead of axial + coronal + sagittal.

This behavior worked correctly in OHIF 3.10, but is broken in more recent versions (e.g., 3.11+)

Test data: https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.2.840.113619.2.290.3.3767434740.232.1619607454.466

3.11 behavior: 

https://github.com/user-attachments/assets/0fb70137-c0d5-466b-a4f4-c38a1c21b7e4

3.10 behavior:

https://github.com/user-attachments/assets/1cd3bb6c-60f1-463f-b4e8-de93bf50397f

### Steps to Reproduce

- Load a series in MPR layout (Axial / Sagittal / Coronal). https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.2.840.113619.2.290.3.3767434740.232.1619607454.466
- In the axial viewport, draw an annotation (e.g., length tool) on Series-1.
- Confirm the measurement appears in the measurement panel.
- Drag and drop Series-2 (different Frame of Reference UID) into all viewports.
- Click on a non-axial viewport (coronal or sagittal).
- Click the annotation created on Series-1 from the measurement panel.

### The current behavior

The annotation from Series-1 is added to:

- The axial viewport that originally contained it (correct), and

- The currently active viewport (coronal/sagittal) showing annotation and changed orientation to axial as annotation is drawn in axial  (incorrect).

The active viewport’s orientation is forced to axial, causing the MPR layout to become:

- 2 axial + 1 coronal (if active viewport is sagittal)

- The original coronal/sagittal orientation is lost.

### The expected behavior

- When clicking the previously drawn annotation the annotation should be displayed in its correct orientation only.
- The MPR layout should retain its original orientation configuration:
[...]

PR #5630, opened 2025-12-12
fix: prevent annotation from appearing in active viewport when switching series with different Frame of Reference UID

Belbin-GK:
### Context

Annotations drawn in one series are incorrectly re-applied to unrelated viewports after loading a new series with a different Frame of Reference UID

Fixes : https://github.com/OHIF/Viewers/issues/5597

### Changes & Results

* Fallback to a viewport with matching orientation when no corresponding displaySet is found.
* Annotations now display only for the correct series and no longer appear in the active viewport when switching between studies or series with different Frame of Reference UIDs, ensuring accurate and consistent annotation behavior

Before:

https://github.com/user-attachments/assets/df5e2fe3-ef88-46c9-b721-b66280739179

After:

https://github.com/user-attachments/assets/b3057765-1c95-40cb-9b9f-709252ab4b71

### Testing

 Sample url : https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.2.840.113619.2.290.3.3767434740.232.1619607454.466

* Load the URL.
* Switch to MPR mode.
* Draw a measurement on a viewport and on a different slice.
* Load a series with a different Frame of Reference UID and select a different viewport.
* Click on the previously drawn measurement from the measurement panel.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS:  Windows 11
- [x] Node version: 22.12.0
- [x] Browser: Chrome 142.0.7444.176
 


extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -613,6 +614,9 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
    * case where the user is in MPR and a viewport other than active should be
    * the one to change to display the iamge.
    *
+   * Third choice is to use a viewport whose orientation best matches the
```
wayfarer3130: Can you extract the three types into separate methods, each with their own comments?  That reduces the total complexity of the top level method and easier to understand what is going on.

Belbin-GK: I have refactored the code.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -599,67 +600,112 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
+    // 5. Fallback: Use the active viewport
+    return {
+      viewportId: activeViewportId,
+      displaySetInstanceUID,
+      viewportOptions: { viewportType },
+    };
+  }
+
+  /**
+   * Determines if the viewport should be a STACK or VOLUME
```
wayfarer3130: Determines if the viewport should be STACK, VOLUME or from displaySet.viewportType

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -599,67 +600,112 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
+   */
+  private _determineTargetViewportType(displaySet, metadata): string {
```
wayfarer3130: Consider making these public methods without the _ on them.  They may well be useful elsewhere.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -668,13 +714,38 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
-      displaySetInstanceUID,
-      viewportOptions: { viewportType },
-    };
+  /**
+   * Use a viewport with matching orientation (IOP) when no direct displaySet match is found.
+   */
+  private _findViewportWithMatchingOrientation(displaySetInstanceUID, viewportType) {
+    const { displaySetService } = this.servicesManager.services;
+
+    const closestOrientation = getClosestOrientationFromIOP(
```
wayfarer3130: There is another PR coming up that sorts the viewports by closest orientation - can we wait on your PR until that PR merges, and then re-use the same logic (in this class even) to get the closest one possible rather than trying for an exact match?

wayfarer3130: https://github.com/OHIF/Viewers/pull/5090 is the PR I'd like to share code with.

wayfarer3130 (approved):

Issue #5651, opened 2025-12-17
[Bug] Segments missing when changing to 3D layout


aimee-ferreira:
### Describe the Bug

Segments are no longer available in the side panel when switching from a common to a 3D layout. The 3D viewport has the jump errors which if used produce an error 

### Steps to Reproduce

1. Display https://viewer-dev.ohif.org/segmentation?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5
2. Add segments on viewport 2
3. Switch to 3D main 

https://github.com/user-attachments/assets/0257516d-b1bb-4871-a7b0-815c156b0e6d

### The current behavior

- Segments are missing from the side panel 
- Jump arrows are available but produce an error if used

### The expected behavior

Segments are available in the side panel
Jump errors move to each segment with no errors 

### System Information

  System:
    OS: Windows 11 10.0.26100
    CPU: (14) x64 Intel(R) Core(TM) Ultra 7 265U
    Memory: 6.41 GB / 31.46 GB
  Browsers:
    Chrome: 143.0.7499.110
    Edge: Chromium (140.0.3485.54)

wayfarer3130:
Resolved in PR: https://github.com/OHIF/Viewers/pull/5669
Note there are separate issues around converting between various formats for contours/labelmaps

Issue #5643, opened 2025-12-16
[Bug] Hydrating RTSTRUCT when displaying 3D volume throws an exception and RSTRUCT is not hydrated


jbocce:
### Describe the Bug

When viewing a study using a hanging protocol with a 3D volume viewport, attempting to hydrate an RSTRUCT is not successful and an exception is thrown. The problem does NOT appear to be study specific, but a general problem.

Maybe https://github.com/OHIF/Viewers/issues/5266 is related, but launching a 3D volume HP when the RT is already hydrated appears to work ok. The only thing might be that the contours are not visible in the 3D volume viewport.

### Steps to Reproduce

See the video...

https://github.com/user-attachments/assets/790d47f9-92f9-4517-b0d3-72f7a55b55c9

### The current behavior

The following exception is thrown and the RTSTRUCT does NOT hydrate as expected - some viewports show the contours, others not so much.

```
TypeError: Cannot read properties of undefined (reading 'points')
    at https://viewer-dev.ohif.org/7431.bundle.4713909b8c30f5f97d1f.js:2:648748
    at Array.map (<anonymous>)
    at c (7431.bundle.4713909b8c30f5f97d1f.js:2:648442)
    at Module.B (7431.bundle.4713909b8c30f5f97d1f.js:2:650284)
    at async Object.render (7412.bundle.de73c323fff02c4238dc.js:1:503563)
```

### The expected behavior

No exception. Hydration happens and contours appears in each viewport - if not immediately then after some time once any calculations have been performed.

### System Information

System:
OS: Windows 10 10.0.26200
CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
Memory: 6.84 GB / 31.68 GB
Binaries:
Node: 23.9.0 - ~\AppData\Local\fnm_multishells\52784_1765830696412\node.EXE
Yarn: 1.22.22 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
npm: 10.9.2 - ~\AppData\Local\fnm_multishells\52784_1765830696412\npm.CMD
Browsers:
Edge: Spartan (20.9.0), Chromium (20.9.0), ChromiumDev (20.9.0)
Internet Explorer: 11.0.26100.7309
Chrome: 143.0.7499.110

wayfarer3130:
Resolved as far as the hydration/exceptions go in: https://github.com/OHIF/Viewers/pull/5669

PR #5669, opened 2025-12-19
fix: Segmentations disappearing

wayfarer3130:
### Context

Going into 3d main view after adding a segmentation causes the segmentation to disappear from all windows.

There is a separate issue that the segmentation isn't listed in the right hand panel, but I'd like to get the basic exceptions fixed and then worry about the UI fix separately.

### Changes & Results

List @icr/polyseg-wasm in the pluginConfig so that the library is available.

### Testing

See the issue: https://github.com/OHIF/Viewers/issues/5651 for details.
Also, https://github.com/OHIF/Viewers/issues/5643 is addressed as far as the exceptions/ability to hydrate.  

The issues with various viewports not getting correctly synced is a separate issue.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


platform/app/pluginConfig.json
```diff
@@ -105,6 +105,11 @@
     {
       "directory": "./platform/public"
     },
+    {
+      "packageName": "@icr/polyseg-wasm",
```
sedghi, line 109: this is very strange. it was working before, and we didn't have this here at the time it was working. So i'm wondering what has changed that we need this here

jbocce (changes requested):
See @sedghi's comment. Also, I think @aimee-ferreira should be listed as a reviewer to test this.

platform/app/pluginConfig.json
wayfarer3130: It wasn't failing before because it wasn't actually invoking the convert to labelmap code, it was just displaying the contour as is.  If you know a specific date/branch that it was both loading successfully AND auto-converting to labelmap, please let me know when that got added/changed.

wayfarer3130: Checked mid-april build - definitely was just loading rtstruct as contours at that point.

wayfarer3130: It doesn't work as of October 3rd either - so I'm wondering when you think this could have been working.

wayfarer3130: This is only needed for dev:fast running - somehow the full run is ok generally, whether I build it locally or it is the production build.

platform/core/src/services/MeasurementService/MeasurementService.ts
wayfarer3130, line 1: @jbocce - this fixes duplicated logging when the source type mapping is missing.  It was creatings thousands of lines of logs making it hard to debug.

Issue #5722, opened 2026-01-16
[Bug] Closing the measurement in microscopy clears all the measurements


aimee-ferreira:
### Describe the Bug

See the title.

### Steps to Reproduce

1. Display https://viewer.ohif.org/microscopy?StudyInstanceUIDs=2.25.269859997690759739055099378767846712697
2. Add some measurements 
3. Close and open the measurement side panel

https://github.com/user-attachments/assets/d02bb0b9-b376-4998-a9df-8259a10a0d8a

### The current behavior

All measurements are gone 

### The expected behavior

Measurements are retained

### System Information

  System:
    OS: Windows 11 10.0.26100
    CPU: (14) x64 Intel(R) Core(TM) Ultra 7 265U
    Memory: 12.28 GB / 31.46 GB
  Browsers:
    Chrome: 144.0.7559.59

jbocce:
I furthermore noticed is that whenever the browser window is resized ALL the measurements also disappear/clear.

PR #5725, opened 2026-01-19
fix(microscopy): Fix measurement visibility toggle

### LINKED ISSUE #5659: [Bug] In microscopy mode measurements cannot be hidden
### Describe the Bug

After adding various markups in microscopy mode, the measurements cannot be hidden and made visible in the right side measurements panel using the eye icon.

### Steps to Reproduce

1. View a study in microscopy mode.
2. Add a few different markups.
3. In the right-side measurements panel, click the eye/visibility icon for a measurement

### The current behavior

The measurement stays visible.

### The expected behavior

The measurement should hide

### System Information

  System:
    OS: Windows 11 10.0.26200
    Memory: 14.28 GB / 31.68 GB
  Binaries:
    Node: 23.9.0 - C:\Users\joebo\AppData\Local\fnm_multishells\9788_1765979860957\node.EXE
    npm: 10.9.2 - C:\Users\joebo\AppData\Local\fnm_multishells\9788_1765979860957\npm.CMD
  Browsers:
    Chrome: 143.0.7499.110
    Firefox: 144.0 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.7309

GhadeerAlbattarni:
### Context

Fixes #5659 

This PR fixes the inability to hide/show individual measurements in microscopy mode using the eye icon in the measurements panel. 
It Implements visibility toggle functionality in the dicom-microscopy extension

### Changes 

 - **RoiAnnotation.js**: Added `isVisible` property and `setVisibility()` method to track annotation visibility state
  - **MicroscopyService.ts**: Added `toggleROIVisibility()` method to coordinate visibility changes across all managed
  viewers
  - **viewerManager.js**: Added `toggleROIVisibility()` method to add/remove ROI from the dicom-microscopy-viewer
  - **getCommandsModule.ts**: Added `toggleMeasurementVisibility` command
  - **MicroscopyPanel.tsx**: Wired up the visibility toggle handler to the DataRow component

### Results

**Before**  
The measurement stays visible when the user clicks on the eye icon.

**After**  
The measurement can be hidden and shown when the user clicks on the eye icon.

https://github.com/user-attachments/assets/09f8035a-572a-4b05-a29c-5c3a0c6e2bb7

### Testing

1. Open a study in microscopy mode
2. Add few different markups 
3. In the right-side measurements panel, click the eye icon for a measurement
4.  Verify the measurement hides on the viewport
5. Click again on the eye icon 
6. Verify the measurement shows again

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.12.0
  


jbocce:
Please update your branch with master and push the result. Thanks.

jbocce:
I did some testing and each of the various measurements can be hidden and displayed.

I did notice https://github.com/OHIF/Viewers/issues/5722 which is a separate issue but continues to exist. It will/should be handled in a separate PR.

extensions/dicom-microscopy/src/services/MicroscopyService.ts
```diff
@@ -410,6 +410,25 @@ export default class MicroscopyService extends PubSubService {
     this.managedViewers.forEach(mv => mv.showROIs());
   }
 
+  /**
+   * Toggle visibility of a specific ROI
+   * @param uid The UID of the ROI to toggle
+   */
+  toggleROIVisibility(uid) {
```
jbocce, line 417: We are clobbering toggleROIVisibility that is on line 396. If we want to replace the old one, then this code should replace above. However I don't think that is what we want to do because the one on line 396 is toggling ALL the visibility of ALL ROIs. So this one should be renamed. Please come up with an appropriate name and/or consider asking AI for a name - it is good with that. 😊

extensions/dicom-microscopy/src/tools/viewerManager.js
```diff
@@ -396,6 +396,21 @@ class ViewerManager extends PubSubService {
...
 
+  /**
+   * Toggles the visibility of a specific ROI.
+   *
+   * @param {string} uid - Unique identifier of the ROI
+   * @param {Object} roiGraphic - ROI graphic object to re-add if making visible
+   * @param {boolean} isVisible - Whether the ROI should be visible or hidden
+   */
+  toggleROIVisibility(uid, roiGraphic, isVisible) {
+    if (isVisible) {
```
jbocce: In my opinion, adding and removing the ROI is NOT the same as hiding and showing it - eventhough it could be argued that is one way to implement visibility. Furthermore, if we look at `hideROIs` and `showROIs` they are NOT doing that to hide and show the ROIs. I looked further and unfortunately the underlying DICOM microscopy viewer does NOT appear to have an API for showing and hiding individual annotations. We may need to look into this further.

extensions/dicom-microscopy/src/services/MicroscopyService.ts
jbocce: Apologies. I see the name difference now. But they are still too similar in my opinion.

extensions/dicom-microscopy/src/getCommandsModule.ts
```diff
@@ -125,6 +125,9 @@ export default function getCommandsModule({
     toggleAnnotations: () => {
       microscopyService.toggleROIsVisibility();
     },
+    toggleMeasurementVisibility: ({ uid }) => {
```
jbocce, line 128: I think this name should be consistent with using the annotations instead of measurements.

extensions/dicom-microscopy/src/services/MicroscopyService.ts
jbocce, line 1: I know this was existing prior to your changes, but line 397 has an error. It should be `this.showROIs()` (i.e. an actual method invocation). We should update this.

jbocce (changes requested):
Thanks for this PR. You have a lot of the fundamentals needed for adding the visibility toggle. 👍 Your proposal for toggling the visibility using add and remove needs to be revisited. It is likely best to add methods in the DICOM micrsoscopy viewer API (https://github.com/ImagingDataCommons/dicom-microscopy-viewer/blob/master/src/viewer.js)  to accomplish this.

PR #5726, opened 2026-01-19
fix: the null value check issue in image spacing calculation

gz281620758:
### Context

image

An instance under the sequence is missing the ImagePositionPatient tag, which will throw an exception when loading the image, causing the program to interrupt

In these two files: isDisplaySetReconstructable.js、areAllImagePositionsEqual.ts
image
image
judgment has been added


gz281620758:
image
here miss the judgement

Issue #5728, opened 2026-01-20
[Bug] Overlay UI missing segmentation set if added to and then deleted from segmentation panel


dxlin:
### Describe the Bug

Overlay UI for segmentation is missing when its deleted from the side panel after it is added as an overlay.

### Steps to Reproduce

1. https://viewer-dev.ohif.org/segmentation?StudyInstanceUIDs=1.3.6.1.4.1.12201.1091.126683095609223531686845324113579088978
2. Add "SEGMENTATION FOR1" as segmentation via overlay UI to first viewport

<img width="285" height="203" alt="Image" src="https://github.com/user-attachments/assets/1c1ccc59-e561-4ac6-87be-da81f7eee4d7" />

3. Delete "Segmentation FOR1" from the segmentation side panel
4. "SEGMENTATION FOR1" is no longer selectable from the overlay UI

<img width="297" height="198" alt="Image" src="https://github.com/user-attachments/assets/ef0510d9-dddb-4863-a9c1-3ce3699bff1a" />

### The current behavior

The segmentation is missing after the delete.

### The expected behavior

Segmentation after delete should still be present in list if its is from PACS.

### System Information

  System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
    Memory: 2.85 GB / 15.72 GB
  Binaries:
    Node: 22.16.0
    Yarn: 1.22.22
    npm: 10.9.2
  Browsers:
    Chrome: 144.0.7559.59
    Edge: Chromium (143.0.3650.66)

Issue #5498, opened 2025-10-15
[Bug] RTStruct not visible or loaded as annotations in certain cases


dxlin:
### Describe the Bug

1. When rtstruct is loaded on referenced display set after loading on an unreferenced display set, segmentations load but are not visible.

<img width="952" height="936" alt="Image" src="https://github.com/user-attachments/assets/606b4227-14c0-4ff3-9bc3-c76158c55e28" />

2. When rtstruct is loaded with duplicate display sets (i.e. same display set but on multiple view ports), the segments show up on the duplicate view ports as annotations/measurements. Potentially intended behaviour.

<img width="945" height="920" alt="Image" src="https://github.com/user-attachments/assets/1fc38ef4-64f0-41e4-bee9-55439f9f5d4b" />

### Steps to Reproduce

Invisible Segments:
1. https://viewer-dev.ohif.org/segmentation?StudyInstanceUIDs=1.2.840.113619.2.290.3.3767434740.226.1600859119.501
2. Viewport 1 -> use overlay menu to load rtstruct (loads and visible)
3. Viewport 3 -> use overlay menu to load rtstruct (loads but not visible)

Extra Annotations:
1. https://viewer-dev.ohif.org/segmentation?StudyInstanceUIDs=1.2.840.113619.2.290.3.3767434740.226.1600859119.501
2. Set viewport 1 (optionally 2 and 4) to AC192
3. Viewport 3 -> use overlay menu to load rtstruct (loads and visible, but also visible as annotations on other viewports)

Test for above in playwright (but screenshots not valid):
https://github.com/dxlin/Viewers/commit/0e2b9d29923bd37560b7685e2cf116cfbcfd7fa7

### The current behavior

Invisible Segments:
rtstruct is invisible in referenced display set

Extra annotations:
loaded rtstructs show up on duplicate viewports as annotation/measurements

### The expected behavior

Invisible Segments:
rtstruct should be visible in referenced display set

Extra annotations:
loaded rtstructs should probably show up as segments or not at all on duplicate viewports

### System Information

  System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
    Memory: 5.20 GB / 15.72 GB
  Binaries:
    Node: 22.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Users\dx\AppData\Roaming\npm\yarn.CMD
[...]

PR #5729, opened 2026-01-20
fix(segmentation overlay): update viewport ds list upon seg delete - OHIF-2425

dxlin:
### Context
A potential solution for
https://github.com/OHIF/Viewers/issues/5728
(Also a test noted in https://github.com/OHIF/Viewers/issues/5498)

Overlay UI is missing segmentations when they are added then removed from the segmentation panel as it is still contained in the viewport's display set list.

### Changes & Results
Segmentation displayset layer is removed as part of the segmentation delete routine. Additionally, for segmentations created within viewer and not from an image source, it is cleaned up from the overall display set list (so as not to clutter the UI).

### Testing

User is now able to add segmentation via overlay UI, delete (whether by accident or on purpose), and still add it back to same viewport via overlay UI without refreshing page.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment
  System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
    Memory: 2.85 GB / 15.72 GB
  Binaries:
    Node: 22.16.0
    Yarn: 1.22.22
    npm: 10.9.2
  Browsers:
    Chrome: 144.0.7559.59
    Edge: Chromium (143.0.3650.66)
- [] OS: 
- [] Node version: 
- [] Browser:
  


dxlin:
Quick note that this only resolves for segmentations, RTStructs are still under investigation (is not able to be loaded a second time after the first delete).

jbocce:
Fixes #5728 
Closes <a href="https://linear.app/ohif/issue/OHIF-2425">OHIF-2425</a>

dxlin:
Hi @jbocce,

Latest commit for PR now correctly works for both segmentations and RTStructs. removeDisplaySetLayer handles segmentation representation so was not needed to be removed again.

As the commands for delete segmentation and removeDisplaySetLayer are pretty close, would it make sense to incorporate the displaySetService.deleteDisplaySet(segmentationid) snippet to removeDisplaySetLayer (this should also fix nothing happening when a segmentation created within the ohif is removed as a layer from overlay menu unless that is desired behaviour)?

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -764,7 +764,15 @@ class SegmentationService extends PubSubService {
       this.updateSegmentationInSource(segmentationId, data as Partial<cstTypes.Segmentation>);
     } else {
       // Add a new segmentation
-      this.addSegmentationToSource(data as cstTypes.SegmentationPublicInput);
+      const spInput = data as cstTypes.SegmentationPublicInput;
+      if (!spInput.representation?.type) {
+        // Safety check to prevent missing representation field error
+        console.warn(
```
jbocce: When do we get into this situation? Do we actually have examples where the representation type is not defined? I tried the example for this PR, put a break point here and never hit it.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -1727,8 +1727,28 @@ function commandsModule({
...
      */
     deleteSegmentationCommand: ({ segmentationId }) => {
-      const { segmentationService } = servicesManager.services;
-      segmentationService.remove(segmentationId);
+      const { segmentationService, viewportGridService, displaySetService } =
+        servicesManager.services;
+
+      // Cleanup segmentation from display set if not from image source
+      const seg = segmentationService.getSegmentation(segmentationId);
+      if (seg && seg.predecessorImageId === undefined) {
```
jbocce: A couple of things here...

- We still want to do a `segmentationService.remove(segmentationId);` always - even when seg.predecessorImageId is defined. I tried the PR deployment, hydrated a SEG via double click but then could not delete it. Likely this `if` was the reason
- The display set for the "manually" added display sets are created and added in `setUpSegmentationEventHandlers`. Likewise they should be deleted there as well - probably as a result of `SEGMENTATION_REMOVED` event.
- I think it is best to use the `displaySet.madeInClient` flag to check if the display set should be deleted or not

So I think it is best to replace this whole if with `segmentationService.remove(segmentationId);` and do the display set deletion in `setUpSegmentationEventHandlers`

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -1727,8 +1727,28 @@ function commandsModule({
...
+          // display set layer
+          segmentationService.remove(segmentationId);
+          displaySetService.deleteDisplaySet(segmentationId);
+        } catch (e) {
+          console.warn(e);
+        }
+      }
+
+      // Cleanup segmentation from display set layer
+      const viewportId = viewportGridService.getActiveViewportId();
```
jbocce: This code removes the layer display set only from the active viewport. But what happens if the segmentation were overlaying several viewports? I am also wondering if this is the correct place to do this or not, but I think it will do for now.

dxlin:
Added relevant code within removeSegmentationFromViewport as well; however, after pulling most recent commits (likely Jan. 27 ones) I'm running into a number of new issues with segmentations, will try to work through and apply suggestions noted in comments as well.

An example of new issue is I'm sometimes finding segmentations previously removed added back in now (additionally, adding segmentations to one viewport will end up adding it to other viewports with images that share a FoR).

extensions/cornerstone/src/commandsModule.ts
dxlin: segmentationService.remove(segmentationId)
-The reason for the if statement was to resolve an issue with adding RTStructs back in after a deletion (issue resolved after keeping the remove inside if)
   -> there may be difference with how SEG/RTstructs are handled/loaded via overlay and double clicking and this if statement may just be hiding some other issue

setUpSegmentationEventHandlers
-I see that the SEGMENTATION_REMOVED event is actually handled by SEGMENTATION_MODIFIED (ultimately by setupSegmentationModifiedHandler under utils\segmentationHandlers.ts), is the recommendation for a check to happen here and then delete the display set? or should the REMOVED event be pulled out and handled differently?

madeInClient
-I had previously attempted to try using this flag but it didn't match my expectations, admittedly I did not spend too much time trying to use it. I can try to look into this a bit more

extensions/cornerstone/src/commandsModule.ts
dxlin: The actual removal of displaysetlayer was fine for the activeViewport in my testing; however, I can confirm the deleteDisplaySet will cause issues if the segmentation is loaded in another viewport.

segmentations from PACS/Image source should be fine as we should be trying to not delete the display set; however, those madeInClient will run into issues (I've found I needed to add some extra displayset null checks, but this may just be hiding a problem that I'm not handling the delete cleanly).

extensions/cornerstone/src/commandsModule.ts
jbocce: Yes you are correct about the SEGEMENTATION_REMOVED being handled by SEGMENTATION_MODIFIED but I think that is a mistake on our part. They should be separate. So once they are separated we can peform logic similar to what you have in this `if` block in `setUpSegmentationEventHandlers`. Thanks for pointing out the lack of a SEGMENTATION_REMOVED event being fired.

One of the primary reasons for having a central handler for segmentation removal is that a segmentation could get removed from other sources and NOT just via this command.

We should be able to rely on `madeInClient` for sure.

extensions/cornerstone/src/commandsModule.ts
jbocce: I think we need to loop through every viewport and remove the display set from those that contain it.

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -245,7 +245,7 @@ class SegmentationService extends PubSubService {
 
     eventTarget.removeEventListener(
       csToolsEnums.Events.SEGMENTATION_REMOVED,
-      this._onSegmentationModifiedFromSource
+      this._onSegmentationRemovedFromSource
```
jbocce, line 250: I think this makes more sense than what was there before. Now we actually fire the SEGMENTATION_REMOVED event!

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -265,7 +265,7 @@ class SegmentationService extends PubSubService {
 
     eventTarget.removeEventListener(
       csToolsEnums.Events.SEGMENTATION_REPRESENTATION_REMOVED,
-      this._onSegmentationRepresentationModifiedFromSource
+      this._onSegmentationRepresentationRemovedFromSource
```
jbocce, line 270: Same goes for this one.

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
```diff
@@ -143,7 +144,8 @@ function _load(
   if (
     (rtDisplaySet.loading || rtDisplaySet.isLoaded) &&
     loadPromises[SOPInstanceUID] &&
-    cachedRTStructsSEG.has(rtDisplaySet.displaySetInstanceUID)
+    cachedRTStructsSEG.has(rtDisplaySet.displaySetInstanceUID) &&
+    _segmentationExists(rtDisplaySet)
```
jbocce, line 145: When an RTSTRUCT contour segmentation was deleted completed and then reloaded it wouldn't. So this check forces the reload when it no longer exists.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -1737,10 +1737,16 @@ function commandsModule({
...
-      segmentationService.removeSegmentationRepresentations(
-        viewportGridService.getActiveViewportId(),
-        { segmentationId }
-      );
+      const viewportId = viewportGridService.getActiveViewportId();
+      segmentationService.removeSegmentationRepresentations(viewportId, { segmentationId });
+
+      const displaySet = displaySetService.getDisplaySetByUID(segmentationId);
+      if (displaySet) {
+        commandsManager.runCommand('removeDisplaySetLayer', {
```
sedghi: what is removeDisplaySetLayer? i can't find it in our codebase

jbocce: It is a [command from the default extension](https://github.com/OHIF/Viewers/blob/4dfd80f1fde1872b23f379be2c2a2725a023313b/extensions/default/src/commandsModule.ts#L135). It does all the work for removing the layer from a viewport. Hmmm, would you suggest for this command to just call it or replace it? 🤔

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
```diff
@@ -219,6 +221,10 @@ function _deriveReferencedSeriesSequenceFromFrameOfReferenceSequence(
   return ReferencedSeriesSequence;
 }
 
+function _segmentationExists(segDisplaySet) {
+  return cstSegmentation.state.getSegmentation(segDisplaySet.displaySetInstanceUID);
```
wayfarer3130: Use !! in front of it to return a boolean

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
```diff
@@ -143,7 +144,8 @@ function _load(
   if (
     (rtDisplaySet.loading || rtDisplaySet.isLoaded) &&
     loadPromises[SOPInstanceUID] &&
-    cachedRTStructsSEG.has(rtDisplaySet.displaySetInstanceUID)
+    cachedRTStructsSEG.has(rtDisplaySet.displaySetInstanceUID) &&
```
wayfarer3130: Please remove cachedRTStructsSEG as a set and just use the new function _segmentationExists.  There just isn't any point in using both as they track the same thing.   You can then throw away the code that does the load tracking too, since the actual state information is definitive for that.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -1737,10 +1737,16 @@ function commandsModule({
      */
     removeSegmentationFromViewportCommand: ({ segmentationId }) => {
       const { segmentationService, viewportGridService } = servicesManager.services;
-      segmentationService.removeSegmentationRepresentations(
-        viewportGridService.getActiveViewportId(),
-        { segmentationId }
-      );
+      const viewportId = viewportGridService.getActiveViewportId();
+      segmentationService.removeSegmentationRepresentations(viewportId, { segmentationId });
```
wayfarer3130: This line is already in removeDisplaySetLayer:
const isSegmentation = DERIVED_OVERLAY_MODALITIES.includes(displaySet.Modality);
      if (isSegmentation) {
        segmentationService.removeSegmentationRepresentations(viewportId, {
          segmentationId: displaySetInstanceUID,
        });
      }

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -1737,10 +1737,16 @@ function commandsModule({
      */
     removeSegmentationFromViewportCommand: ({ segmentationId }) => {
```
wayfarer3130: Really, we don't need this command at all - it is exactly removeDisplaySetLayer - I'd suggest just making it a tiny layer on top of that command, just renaming segmentationid to displaySetId, or even just calling the other command directly and add handling in it to get viewportId as the current active viewport id, and assigning display set uid from segmentation id in that method, and assuming it is segmentation id if passed OR the value matches.

extensions/cornerstone/src/commandsModule.ts
wayfarer3130: Please rename this operation to: `removeRepresentationsFromViewport`
The existing name: `removeSegmentationRepresentations`  is wrong since we are NOT removing segmentation representations, we are removing the representations from the viewport.  The first one would be getting rid of the representations entirely, which is a required operation, but isn't what we are doing here.

extensions/cornerstone/src/commandsModule.ts
jbocce: Since this is an existing command, I will leave it and just call `removeDisplaySetLayer` doing some parameter massaging.

extensions/cornerstone/src/commandsModule.ts
jbocce: Sure thing. Since we are changing the public API for `SegmentationService`, I will also start a 3.12 -> 3.13 migration guide. Does this make sense @wayfarer3130 and @sedghi ?

extensions/cornerstone/src/utils/setUpSegmentationEventHandlers.test.ts
```diff
@@ -15,6 +15,7 @@ describe('setUpSegmentationEventHandlers', () => {
   const mockSegmentationService = {
     EVENTS: {
       SEGMENTATION_ADDED: 'SEGMENTATION_ADDED',
```
wayfarer3130, line 17: These are actually SEGMENTATION_ADDED_VIEWPORT and
SEGMENTATION_REMOVED_VIEWPORT to match the actual functionality - if you update the name in SegmentationService and then leave an @deprecated SEGMENTATION_ADDED with a comment that this is now .... to describe what it is actually doing then it will make it backwards compatible for some length of time, and will work to describe things going forward.

jbocce (approved):

PR #5741, opened 2026-01-23
studyExpandRow now supports selecting single or multiple sequences for review

gz281620758:
### Context

When only a specific series needs to be retrieved from the study, I can use a link similar to http://localhost:3000/viewer?StudyInstanceUIDs=xxx&SeriesInstanceUIDs=xxx. This presupposes that you know the corresponding studyUid and SeriesInstanceUIDs. Now we can use a more convenient interaction like this:
viewer

### Changes & Results

Here, the caching has been temporarily commented out to prevent reusing Promises in repeated requests, as this can lead to data inconsistency. If this issue can be resolved, the caching can be uncommented
image


sedghi:
we have other plans for this page, @dan-rukas can share some screenshots or links

PR #5752, opened 2026-01-28
fix(app): display study not found message for all modes

### LINKED ISSUE #5705: [Bug] Study not found message not displayed when launching from some modes
### Describe the Bug

When the study is not found/doesn't exist the following modes are not displaying the study not found message to return to study list:

- tmtv
- microscopy
- dynamic-volume

### Steps to Reproduce

Modify the studyUID when launching from tmtv, microscopy and dynamic-volume 
e.g. 
https://viewer-dev.ohif.org/tmtv?StudyInstanceUIDs=1.2.840.113619.2.290.3.3767434740.838.152646863
https://viewer-dev.ohif.org/microscopy?StudyInstanceUIDs=2.25.2698599976907597390550993787678467126)
https://viewer-dev.ohif.org/dynamic-volume?StudyInstanceUIDs=2.25.2327044207364477103179090041594928407

### The current behavior

A loading screen displays but nothing will load

### The expected behavior

Page is displayed like the modes indicating that the study is not found and a link to the study list

### System Information

  System:
    OS: Windows 11 10.0.26100
    Memory: 6.82 GB / 31.46 GB
  Browsers:
    Chrome: 143.0.7499.170

GhadeerAlbattarni:
### Context

Fixes #5705 

This PR fixes an issue where the study-not-found message was not displayed for `tmtv`, `microscopy`, and `dynamic-volume` modes.

### Root Cause
The study validation logic lives in `PanelStudyBrowser.tsx`, which is not rendered when `leftPanelClosed` is true (`tmtv`, `microscopy`) or when a different panel is used (`dynamic-volume`). As a result, the validation logic does not run in these cases, causing users to see an infinite loading screen instead of the error message.

### Changes & Results
This PR adds an early validation step in `defaultRouteInit.ts` that:
- Queries each study before retrieving series metadata
- Navigates to `/notfoundstudy` if any study is not found

This ensures validation runs for all modes during route initialization, before layout rendering.

Screenshot 2026-01-28 at 10 58 11 AM

### Testing
- Open a study from one of these modes `tmtv`, `microscopy`, and `dynamic-volume` then modify the `StudyInstanceUIDs`
- Verify the study not found message displayed and redirect to `notfoundstudy` page 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: macOS 10.15.4
- [] Node version: v22.12.0
- [] Browser: Chrome 83.0.4103.116
  

### DIFF
--- a/tests/StudyValidation.spec.ts
+++ b/tests/StudyValidation.spec.ts
@@ -62,7 +62,7 @@ test.describe('Study Validation', () => {
       page,
       notFoundStudyPageObject,
     }) => {
-      await page.goto(`/${mode}/ohif?StudyInstanceUIDs=${invalidStudyUID},${validStudyUID}}`);
+      await page.goto(`/${mode}/ohif?StudyInstanceUIDs=${invalidStudyUID},${validStudyUID}`);
 
       await page.waitForURL('**/notfoundstudy', { timeout: 15000 });
 


platform/app/src/routes/Mode/defaultRouteInit.ts
```diff
@@ -25,6 +26,20 @@ export async function defaultRouteInit(
 ) {
   const { displaySetService, hangingProtocolService, uiNotificationService, customizationService } =
     servicesManager.services;
+
+  // Validate that requested studies exist
```
jbocce: I think it might be better to put this logic into the ModeRoute component (platform\app\src\routes\Mode\Mode.tsx) right before the route is initialized. That way we can use `navigate` from the React router. Also be sure to remove the logic from `PanelStudyBrowser`.

jbocce (changes requested):
Thank you. See my comment. Also please add a Playwright test for this, but instead of using screen shots, let's use assertion/expect statements by querying the DOM. Also we should think about whether we want some type of error page object in the tests for this or not.

GhadeerAlbattarni:
I'll work next on writing the Playwright test

platform/app/src/routes/Mode/defaultRouteInit.ts
```diff
@@ -1,6 +1,7 @@
 import { DicomMetadataStore, log, utils, Enums } from '@ohif/core';
 import getStudies from './studiesList';
 import isSeriesFilterUsed from '../../utils/isSeriesFilterUsed';
+import { history } from '../../utils/history';
```
jbocce: Please remove these. Probably best to do `git checkout origin/master -- platform\app\src\routes\Mode\defaultRouteInit.ts`

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -130,6 +133,34 @@ export default function ModeRoute({
     };
   }, [location, ExtensionDependenciesLoaded]);
 
+  useEffect(() => {
```
jbocce, line 140: Add comment describing what this `useEffect` is doing. Maybe also add where it used to be and why we moved it.

tests/pages/NotFoundStudyPageObject.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export class NotFoundStudyPageObject {
+  readonly page: Page;
+
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  get errorMessage(): Locator {
+    return this.page.getByText(
```
jbocce: I think it would be best to add a `data-cy` id to the message in the `NotFoundStudy` component and use it here to fetch the message. Also I think we should also add yet another getter here for "return to study list message". Likewise give it a `data-cy` attribute so that it can be easily fetched as well.

tests/StudyValidation.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { expect, test } from './utils';
+
+test.describe('Study Validation', () => {
+  const invalidStudyUID = '9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9';
+
+  const modes = ['viewer', 'microscopy', 'tmtv'];
```
jbocce: Please add segmentation here too.

tests/StudyValidation.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { expect, test } from './utils';
+
+test.describe('Study Validation', () => {
+  const invalidStudyUID = '9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9.9';
+
+  const modes = ['viewer', 'segmentation', 'microscopy', 'tmtv'];
+
+  modes.forEach(mode => {
+    test(`should display error page when study UID is not found in ${mode} mode`, async ({
```
jbocce, line 11: Let's also add a series of tests where one study is valid and the other is not valid. Funny that I added this comment yesterday, but may github ate it. 🤣🤣

jbocce: Just to be clear the `goto` URL should have two studies: one valid and one invalid.

platform/app/src/routes/index.tsx
```diff
@@ -38,12 +38,22 @@ const NotFoundStudy = () => {
...
           One or more of the requested studies are not available at this time.
         </h4>
         {showStudyList && (
-          <p className="mt-2">
-            Return to the <Link className="text-primary-light" to="/">study list</Link> to select a different study to view.
+          <p
+            className="mt-2"
+            data-cy="return-to-study-list-message"
+          >
+            Return to the{' '}
```
jbocce, line 49: Nice change here. Thanks.

jbocce (changes requested):
Just one set of tests and I think you are good to go.

GhadeerAlbattarni:
Thanks Joe for your review!

aimee-ferreira (approved):
verification has passed. The correct 'study not found message' is now displaying in tmtv, microscopy and dynamic-volume modes. No new defects were logged.

PR #5775, opened 2026-02-03
fix(HistoryMemo): Segmentation delete wasn't remembered

wayfarer3130:
### Context

The delete of segment index wasn't being remembered for the history. 
This requires https://github.com/cornerstonejs/cornerstone3D/pull/2593 to be linked to work.

Note: Undo of a delete restores the segment index, and redo of a delete removes it.

### Changes & Results

Adds a history memo for delete/restore of the segment index.

Does not add support for remembering other changes on segment index like rename.

### Testing

Test for segment index for labelmap and contour indexes

Create segmentation (not undo create segmentation is NOT supported)
Draw a segment
Create a new segment index
Draw a segment
Create a new segment index
Draw a segment

Delete segment #1, then #2 then #3

Undo 6 times to remove all segment indices on screen
Segment indexes should appear as the 3 deletes are undone, AND the segment data should reappear.
Segments drawn should disappear as remaining 3 deletes are done.
NOTE: Segment index add is not recorded, so undoing it does nothing

Redo 6 times
Segments drawn should reappear in first 3 redo
Segments drawn should disappear as redo of next 3 deletes.  This should remove the segment index from the table.

Test the whole kittencaboodle with a contours after doing labelmaps.
Maybe test undo/redo of multi frames of segmentations - this should delete all the single segment index on all frames at once.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  


jbocce:
Testing Notes:

- I was not able to undo the creation of a spline contour; that is once created clicking undo does not (eventually) remove the spline contour. Furthermore after several clicks of undo, several clicks of undo adds somewhat of an identical spline, sort of duplicating what is there


https://github.com/user-attachments/assets/973db8e4-70c3-4433-a69a-6b34289ac446

- undo of Duplicate Contour does not work or is not implemented yet

wayfarer3130:
@aimee-ferreira - the test build for this has the remaining issues for OHIF 3.12 RC2

[OHIF-2461: [Bug] Toast message on redo of contours](https://linear.app/ohif/issue/OHIF-2461)
[OHIF-2460: [Bug] Redo action brings back labelmap on image but not in panel](https://linear.app/ohif/issue/OHIF-2460)
[OHIF-2466: [Bug] OHIF 3.12 release candidate is creating invalid DICOM seg objects](https://linear.app/ohif/issue/OHIF-2466)
[OHIF-2456: [Bug] Saving rectangles with HU as pixel units saves unit as 'mm'](https://linear.app/ohif/issue/OHIF-2456)

I have tested all 4 issues, plus Joe's navigation issue in combination with these.  The only remaining bugs I've found are:
- https://github.com/OHIF/Viewers/issues/5780
- https://github.com/OHIF/Viewers/issues/5781

bun.lock
```diff
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
```
jbocce, line 3: What version of bun are you running? Mine does not add this in to the lock file.

extensions/cornerstone/src/utils/updateSegmentationStats.ts
```diff
@@ -59,6 +59,12 @@ export async function updateSegmentationStats({
   Object.entries(stats).forEach(([segmentIndex, segmentStats]) => {
     const index = parseInt(segmentIndex);
 
+    if (!updatedSegmentation.segments[index]) {
```
jbocce, line 62: Perhaps safest to do `if (!updatedSegmentation?.segments?.[index]) {`?

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
```diff
@@ -964,8 +1030,21 @@ class SegmentationService extends PubSubService {
    * 3. If the removed segment was the active segment, it updates the active segment index.
    *
    */
-  public removeSegment(segmentationId: string, segmentIndex: number): void {
-    cstSegmentation.removeSegment(segmentationId, segmentIndex);
+  public removeSegment(
```
jbocce, line 1033: So why doesn't `SegmentationService.addSegment` memo/record its actions? Is it worth mentioning somewhere in the code comments?

bun.lock
wayfarer3130: 1.3.8 is what I run.  I don't think it is a big deal to have this added unless it causes problems in earlier versions.  The bun.lock is not hte official lockfile, but is for faster dev testing.

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
wayfarer3130: Because I only implemented the delete operation as I was trying to minimize how many changes we have at this point.  If we want, we can go ahead and fix all the remaining issues we've seen, but I specifically created a secondary issue to track things like the addSegment not adding the memo.

wayfarer3130:
Added https://github.com/cornerstonejs/cornerstone3D/issues/2599 to describe Joe's issue.

extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts
jbocce: Ok. Then maybe add a ToDo or something?

PR #2611, opened 2021-11-16
Add ability to build ExpandableToolMenu ToolbarRows recursively.

andrewwebbdeveloper:
- Add ability to build ExpandableToolMenu ToolbarRows recursively.
- Allows any number of sub-menus ( configured from the toolbarModule ).

### PR Checklist

- [ ] Brief description of changes
- [ ] Links to any relevant issues
- [ ] Required status checks are passing
- [ ] User cases if changes impact the user's experience
- [ ] `@mention` a maintainer to request a review

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #5806, opened 2026-02-10
feat: Add support for labelmap seg images in any supported tsuid (also for compressed bitmap)

wayfarer3130:
### Context

Integrates OHIF with the Cornerstone3D adapter rework (cs3d_ref: `fix/use-imageLoader-for-seg`, cornerstonejs/cornerstone3D#2611) that loads DICOM SEG through the **image loader** (one loadable imageId per frame) instead of fetching the whole binary SEG object and decoding it in one shot. The result is significantly faster SEG load/save, support for compressed SEG frames, and much smaller stored segmentations (RLE-compressed labelmaps).

### Changes & Results

**SEG loading (image-loader path)**

- `cornerstone-dicom-seg` SOP class handler loads SEG via the adapter's per-frame `createFromDicomSegImageId` and passes an explicit frame fetch/decode **concurrency** (16) so the many small per-frame requests overlap.

**Full-instance Part 10 prefetch (performance for static / per-frame back ends)**

- `DicomWebDataSource` gains `retrieve.prefetchInstanceFrames`: just before the SEG loader runs it can fetch the whole instance once (`retrieveInstance`, `multipart/related` unwrapped) and register it into the Cornerstone3D frame registry, so per-frame loads are served locally instead of one HTTP request per frame. Best-effort and non-fatal — on any failure it falls back to per-frame fetches.
- Gated by `loadMultiframeAsPart10RaceTimeMs` (per data source; `0`/unset = off). **Enabled for the static WADO back ends** in `config/default.js` and `config/e2e.js`.

**SEG saving / storage**

- Store SEG as an RLE-compressed labelmap (`utils/dicomWriter.ts`), with report dialog store options (`reportDialogCustomization`).
- Local SEG support via `registerNaturalizedDatasetForLocalWadouri` so generated SEGs load through the same local wadouri path as other instances.

**Supporting changes**

- `SegmentationService` / `CornerstoneViewportService` updates for stack labelmap representation / rendering; `MetadataProvider` / `combineFrameInstance` / `imageIdToURI` / `appendFrameQueryToImageId` helpers for per-frame imageIds.
- New **Behaviours** docs section documenting the multiframe Part 10 prefetch behaviour (`platform/docs/docs/behaviours/`).
- Dependency bumps to the matching cs3d packages.

### Testing

- [ ] Load a multiframe DICOM SEG and confirm it renders; verify it loads via per-frame imageIds (and, on a static back end, that the single-instance prefetch is used).
- [ ] Save → restore: create/edit a segmentation, save, reload, confirm it restores correctly.
- [ ] Download a saved segmentation and confirm the output is an RLE-compressed labelmap.
- [ ] Backward compat: existing/older segmentations still load.
- [ ] With `loadMultiframeAsPart10RaceTimeMs` unset, behaviour is unchanged (per-frame loads).


jbocce:
@wayfarer3130 please add details for testing this. Thank you.

jbocce:
@claude review

fedorov (commented):
## AI-Assisted Code Review

> **How this review was generated:** This review was produced using [Claude Code](https://claude.ai/code) (model: `claude-sonnet-4-6`, Anthropic, May 2026). The review process had two passes:
> 1. **Diff review** — the full PR diff (2,621 lines, 39 files) was analyzed for correctness, style, and known anti-patterns.
> 2. **Codebase analysis** — the branch was checked out locally at `/Users/af61/github/Viewers` and ~160 targeted file reads and greps were performed across the full monorepo to assess cross-cutting impact on unchanged code.
>
> All findings were verified against the actual source. Sections marked "Cleared" below were flagged in the diff pass but found safe after full-codebase inspection.

---

## Overview

The PR replaces the ArrayBuffer-based SEG loading path with a per-frame imageId approach that delegates pixel data loading to the Cornerstone image loader. Key changes:

- New imageId-based API for `adaptersSEG.Cornerstone3D.Segmentation.createFromDICOMSegBuffer`
- Support for the labelmap SEG SOP class (`1.2.840.10008.5.1.4.1.1.66.7`) in addition to bitmap
- Configurable store encoding (labelmap vs bitmap, transfer syntax) via `customizationService`
- OHIF runtime properties (`imageId`, `wadoRoot`, `frameNumber`, etc.) made non-enumerable to prevent dcmjs from writing them into stored DICOM
- Centralized DICOM serialization in a new `dicomWriter.ts` utility
- `dcmjs` bumped from 0.49.4 → 0.51.1 across the monorepo

The architectural direction is sound and several pre-existing bugs are fixed. However there is one critical sequencing issue that would likely prevent the primary feature from working.

---

## Critical Issues

### 1. `SEGMENTATION_LOADING_COMPLETE` fires before segmentation is in state — null-guard silently drops overlay

This is the most significant issue, found via full codebase analysis. In `SegmentationService.ts`, `SEGMENTATION_LOADING_COMPLETE` is broadcast (line ~587) **before** `addOrUpdateSegmentation` is called (line ~611), which is what actually puts the segmentation into state.

The new subscriber in `CornerstoneViewportService.ts` fires `applyRepresentation()` immediately on the event. That calls `addSegmentationRepresentation` → `getSegmentation(segmentationId)`. Because the segmentation isn't in state yet at that point, the null-guard introduced by this same PR at line 302 triggers — logs a warning and **silently returns** without adding the overlay representation. The SEG labelmap is parsed and loaded, but never applied to the viewport.

**Fix:** Move the `SEGMENTATION_LOADING_COMPLETE` broadcast to after `addOrUpdateSegmentation` completes, or restructure the subscriber to wait until the segmentation is in state before calling `addSegmentationRepresentation`.

### 2. Promise in `addOverlayRepresentationForDisplaySet` never resolves on SEG load failure

If `_loadSegments` throws (no imageId, adapter error, missing referenced display set), `SEGMENTATION_LOADING_COMPLETE` is never fired. The `Promise` returned by `addOverlayRepresentationForDisplaySet` never resolves or rejects, so `_addOverlayRepresentations` stalls indefinitely. In practice this blocks the `VIEWPORT_DATA_CHANGED` broadcast but not viewport rendering (since `viewport.setStack` completes before overlay setup is awaited). Still, any downstream update dependent on that broadcast is silently dropped with no way to recover short of a page reload. No error event subscription, timeout, or rejection path is present.

---

## Issues Flagged in Diff Pass, Cleared by Codebase Analysis

**`frameIndex` → `frameNumber` rename:** Not a breaking change. `MetadataProvider.getUIDsFromImageID` always overrides the stored field with the URL-parsed frame value; the stored field name was always ignored.

**`baseImageURIForMetadata` coverage:** Correct and consistent. The new function strips both `?frame=` and `&frame=`, fixing an inconsistency in the old code that only stripped `&frame=`. Keys remain consistent within a session.

**`getImageIdsForInstance` behavior change:** Correct fix. The old code always used `&frame=` even for URLs with no existing query params; `appendFrameQueryToImageId` correctly chooses `?` vs `&`. No callers depended on the old behavior.

**`displaySets={[referencedDisplaySet, segDisplaySet]}`:** Correct fix. `OHIFCornerstoneViewport` uses only `viewportData.data[0].imageIds` for the stack; SEG is picked up by `_processExtraDisplaySetsForViewport` as an overlay. No double rendering.

**Non-enumerable `frameNumber`:** Safe. All access patterns use direct property access (destructuring, dot notation) — no code uses `Object.keys`, `JSON.stringify`, or spread on combined frame instances expecting `frameNumber`.

---

## Medium Issues

### 3. `tmtv` RT export not updated

`tmtv/src/utils/dicomRTAnnotationExport/RTStructureSet/dicomRTAnnotationExport.js` still calls `dcmjs.data.datasetToBlob` directly without `makeExistingPropertiesNonEnumerable`. Any OHIF runtime properties that remain enumerable on RT export instances will be serialized as DICOM tags, potentially corrupting RT output. This is pre-existing, but the PR introduces the fix pattern everywhere else and misses this file.

### 4. Blob URL memory leak in `registerNaturalizedDatasetForLocalWadouri`

`dicomImageLoader.wadouri.fileManager.add(blob)` registers a blob URL that is never revoked. Each report store/download/clipboard operation accumulates a blob in memory for the session lifetime. No `fileManager.remove()` call exists anywhere in the codebase. For long sessions with many reports this could be meaningful.

### 5. Duplicate `getDatasetTransferSyntaxUID` function

Defined privately in both `dicomWriter.ts` and `DicomWebDataSource/index.ts` with slightly different fallback logic (the latter falls back to `EXPLICIT_VR_LITTLE_ENDIAN`; the former returns `undefined`). The `DicomWebDataSource` version should import from `dicomWriter.ts` to avoid future divergence.

### 6. Dual `transferSyntaxUID` / `transferSyntaxUid` keys in `segmentationConfig.ts`

```typescript
options.transferSyntaxUID = transferSyntaxUID;
options.transferSyntaxUid = transferSyntaxUID;  // camelCase inconsistency
```

This is clearly a workaround for an API inconsistency in `@cornerstonejs/adapters`. Without a comment referencing the upstream issue or PR, future readers will remove one key as an obvious duplicate and silently break one code path.

### 7. `LABELMAP_SEG_SOP_CLASS_UID` defined twice

Defined at the top of `getSopClassHandlerModule.ts` and again in `segmentationConfig.ts`. The local definition in `getSopClassHandlerModule.ts` is dead code since `getSegmentationParserType` is imported from `segmentationConfig.ts`.

### 8. `registerNaturalizedDatasetForLocalWadouri.js` formatting

Every line in this new file is separated by an extra blank line — the entire file is double-spaced. Likely an editor/diff artifact. Makes the file hard to read and should be cleaned up before merge.

---

## Minor / Nits

- **Verbose `console.info` in production paths** — `_logSegImageIds` dumps full imageId arrays on every SEG load; `registerNaturalizedDatasetForLocalWadouri` logs blob byte size and all frame imageIds. Useful for debugging, noisy in production.
- **PR checklist not filled out** — all checkboxes use `[]`; tested OS/Node/Browser fields are blank.
- **`@cornerstonejs/adapters 4.22.8` still depends on dcmjs 0.49.4** — per `bun.lock`, adapters may bundle their own copy alongside OHIF's 0.51.1. Worth verifying the adapters' behavior is compatible with the API surface used in `dicomWriter.ts`.

---

## What's Done Well

- **Event listener cleanup** — `try/finally` around `SEGMENTATION_LOAD_PROGRESS` correctly fixes a pre-existing leak where N SEG loads accumulated N persistent listeners.
- **Non-enumerable runtime properties** — clean, surgical fix preventing dcmjs from writing OHIF bookkeeping fields into stored DICOM output.
- **`dicomWriter.ts` centralization** — single place for dcmjs serialization with consistent options and transfer syntax preservation; a clear improvement over scattered ad-hoc `datasetToBlob` calls.
- **`_getSegDataSource` fallback logic** — correctly prefers `dicomlocal` for local-scheme URLs, falls back to the active data source.
- **`appendFrameQueryToImageId`** — fixes a real bug where `&frame=` was always used regardless of whether the base URL already had query params.
- **Null-guards in `addOrUpdateSegmentation` and `determineViewportAndSegmentationType`** — reasonable defensive programming; the warn log in each case makes skips visible rather than truly silent.
- **SEG viewport stack fix** — passing `[referencedDisplaySet, segDisplaySet]` instead of just `[segDisplaySet]` is a meaningful and correct fix; the old code left the viewport with only derived labelmap imageIds as the primary stack, which are not displayable as grayscale images.

---

## Summary

The critical issue is the event ordering in `SegmentationService.ts`: `SEGMENTATION_LOADING_COMPLETE` is broadcast before the segmentation is in state, so the new `CornerstoneViewportService` subscriber calls `addSegmentationRepresentation` too early, hits the null-guard introduced by this same PR, and silently exits — meaning the labelmap overlay is never applied. This would be caught only by end-to-end testing; the null-guard makes it invisible as a code-level error.

wayfarer3130:
# PR #5806 — Review Remediation Plan (OHIF side)

**PR:** feat: Add support for labelmap seg images in any supported tsuid (also for compressed bitmap)
**Branch:** `feat/load-seg-images` → `master` — https://github.com/OHIF/Viewers/pull/5806
**Companion:** cornerstonejs/cornerstone3D#2611 (`fix/use-imageLoader-for-seg`, see
`libs/@cornerstonejs/PR-2611-review-plan.md`)

Every item below was verified against the current checkout (top commit `22c7c60c24`,
"Bump @cornerstonejs/* pins 5.1.3 -> 5.4.10") with 5 parallel verification passes on
2026-07-09. Line numbers are the **actual current** locations. The review was written
against an **older head** (pre-bump), so one claim is partially stale — noted inline.

Work through items one at a time. Check the box and fill in `Status:` as you go.

Legend: `[ ]` todo · `[~]` in progress · `[x]` done · `[-]` won't-fix / documented instead

---

## Notes on scope (corrections to the review, confirmed during verification)

- **Bug 3 (version pins) is NOT a plan item — it's the designed CS3D_REF workflow.**
  Referencing an unreleased cs3d is the entire point of the repo's CS3D integration
  mechanism: CI tests against a cs3d branch (cloned into `libs/@cornerstonejs`, built,
  linked) when the PR carries the `ohif-integration` label + a `CS3D_REF:` line, and the
  **CS3D Branch Merge Guard** hard-fails merge while the ref is a branch — so the
  "registry lacks the APIs" state is expected and CI-enforced, not a bug. The review's
  factual halves were still checked: the "no bump in the diff" half is stale (commit
  `22c7c60c24` bumps 10 package.json files + lockfile to `5.4.10` consistently), and
  published `5.4.10` indeed lacks `createFromDicomSegImageId` / `prefetchPart10Instance`
  (tarballs unpacked and grepped) — which is exactly the state the merge guard exists
  for. What remains is CI wiring on the PR itself (see 1c).
- **Bug 1's blast radius is narrower than "every SEG/SR saved":** the remote STOW upload
  path (`DicomWebDataSource/index.ts:474`) builds its own hex-keyed dict via
  `denaturalizeDataset` + `writeDicomDictToPart10Buffer` and never runs
  `applyTransferSyntaxToFileMeta` — **server copies are clean**. Affected: download,
  copy-to-clipboard, the local-wadouri blob, and `DicomLocalDataSource.store.dicom`.
- **Bug 4 reaches the wire** (STOW bytes, download, clipboard) but NOT the local wadouri
  cached copy — the local blob is built *before* the mutation, so stored copy ≠ local
  copy (an internal inconsistency on top of the conformance issue).
- **Blob-leak risk:** the review said "blob URLs are never released" — slightly imprecise;
  no `URL.createObjectURL` is involved (the `dicomfile:N` scheme indexes an array), it's
  Blob *references* that accumulate. Same memory effect. Also: `fileManager.remove()` /
  `purge()` **already exist** in the loader — OHIF just never calls them.
- **settle-before-apply risk:** the promise **does** already wait for SEG pixel-data load
  (SEGMENTATION_LOADING_COMPLETE fires after `addOrUpdateSegmentation`); early-resolve
  only skips the *representation attachment* phase. Trivial case = a microtask; the real
  window is the async **stack→volume viewport conversion** inside `applyRepresentation`.
- **dcmjs engines risk is effectively refuted for this repo:** the repo's own
  `engines.node >= 24` (root package.json, `.node-version` 24.15.0, all CI at 24.15.0)
  is stricter than dcmjs's `>= 22.13`, and pnpm only *warns* on dependency engines
  (no `engine-strict` anywhere). No install can fail because of dcmjs here.
- **frameNumber non-enumerable risk:** mechanism confirmed, but a repo-wide consumer
  audit found **zero currently-broken readers** — one risky copy site exists
  (`SegmentationService.ts:511-514`) whose copies nothing reads `.frameNumber` from
  today. Latent footgun, not an active bug. A **related, un-reviewed hazard** was found
  in the same commit (see 3a).

---

## Phase 1 — Confirmed bugs (block merge)

### [x] 1a. Garbage (0000,0000) element written into every saved file's meta group
- **Where:** `extensions/default/src/utils/dicomWriter.ts:90`
  (`dicomDict.meta.TransferSyntaxUID = entry;` — the hex assignment at `:91` is correct)
- **Verified:** CONFIRMED with a runtime repro against installed dcmjs 0.52.0.
  `DicomMessage.write` iterates ALL meta keys; `Tag.fromString('TransferSyntaxUID')`
  parses to `(0000,0000)` via `parseInt(..., 16) → NaN → 0`. The re-read file meta
  contains an extra `'00000000'` UI element, it is **counted in
  FileMetaInformationGroupLength**, and (since `'T'` sorts after hex digits) it is
  written **after** the 0002 elements — doubly non-conformant (garbage tag + out-of-order
  element in group 2). dcmjs tolerates re-reading it; dciodvfy/strict archives may not.
- **Affected flows** (all via `datasetToDicomBlob` → `datasetToDicomPart10Buffer`):
  - `commandsModule.ts:793` — Download DICOM
  - `commandsModule.ts:806` — Copy to clipboard
  - `registerNaturalizedDatasetForLocalWadouri.js:26` — local wadouri blob (invoked from
    all three `createStoreFunction` branches, incl. the STOW path at `:824` — the local
    blob dicomParser must then re-parse carries the garbage element)
  - `DicomLocalDataSource/index.js:183` — local data source store
  - **NOT affected:** remote STOW upload (`DicomWebDataSource/index.ts:474`, separate
    writer path).
- **Fix:** Delete line 90 exactly as the review says. Keep line 91 — it is NOT redundant
  with `datasetToDict`: dcmjs only honors the object-form `_meta.TransferSyntaxUID`,
  while OHIF's `getDatasetTransferSyntaxUID` also handles plain-string `_meta` and
  top-level `dataset.TransferSyntaxUID` fallbacks that dcmjs would silently default to
  Explicit VR LE.
- **Pairs with:** 4a (regression test — the existing `dicomWriter.test.ts` never calls
  `datasetToDicomPart10Buffer`/`applyTransferSyntaxToFileMeta`; it only tests
  `writeDicomDictToPart10Buffer` fragmentation).
- **Status:** FIXED 2026-07-09 — deleted the naturalized-key assignment, kept the hex
  assignment with a comment explaining why hex-only. Regression test added (4a);
  mutation-checked (re-introducing the line fails both new meta tests).

### [x] 1b. rsbuild config crashes for any fresh clone (`source-map-loader` not a dependency)
- **Where:** `rsbuild.config.ts:82-87` — unconditional
  `require.resolve('source-map-loader')` at `:85`, evaluated at config-load time; no
  try/catch, no existsSync guard.
- **Verified:** CONFIRMED, and *worse* than reviewed: `source-map-loader` appears in
  **zero** package.json files in the working tree (not even an uncommitted local change —
  `git status` is clean), is absent from the committed `pnpm-lock.yaml`, and exists only
  as a stray manually-placed directory in local `node_modules` (v4.0.2, not a pnpm
  symlink). The `include` regex targets `libs/@cornerstonejs/packages/*/dist/esm`, which
  is **gitignored local cs3d-linking tooling** (`.gitignore:65,71`; `git ls-files libs`
  → 0 files) — harmless by itself (never matches), the crash is purely the
  `require.resolve`. Added in commit `f81c8b4333` ("Debug fixes"), in the PR range.
- **Fix:** Drop the hunk from the PR (it's personal debug tooling). If it should be
  shared: wrap the resolve in try/catch (or gate on
  `fs.existsSync(path.join(__dirname, 'libs/@cornerstonejs'))`) AND add
  `source-map-loader` to root devDependencies + lockfile.
- **Status:** FIXED 2026-07-09 — chose the guard (not the drop): `require.resolve`
  wrapped in try/catch at module top (`SOURCE_MAP_LOADER`), rule conditionally spread
  only when resolvable. Deliberately did NOT add it to devDependencies — that needs a
  `pnpm install`, which clobbers cs3d:link symlinks; decide separately if it should be
  a real dep. Config parse-checked (rsbuild loads configs transpile-only).

### [X] 1c. Wire up the CS3D_REF integration on PR #5806 (review's "unreleased APIs" item, reframed)
- **Not a code bug.** Depending on unreleased cs3d is by design — the CS3D_REF workflow
  exists precisely for this (README `:182-217`,
  `platform/docs/docs/development/cs3d-integration.md`): CI clones the cs3d branch into
  `libs/@cornerstonejs`, builds and links it, runs Playwright against it, and the
  **CS3D Branch Merge Guard** (`.scripts/ci/cs3d-branch-merge-guard.sh`) fails while the
  ref is a branch, preventing merge until re-run against a published version.
- **The actual gap (verified against the live PR):** the integration isn't currently
  triggered for #5806:
  1. The **`ohif-integration` label is not on the PR** (label list is empty) — without
     it `cs3d-check-integration.sh` disables the integration path entirely.
  2. The PR body mentions the ref as inline lowercase text
     (``cs3d_ref: `fix/use-imageLoader-for-seg` `` inside a sentence), but the parser
     (`cs3d-check-integration.sh:17-18`) only matches an uppercase `CS3D_REF:` at the
     start of a line — so even with the label, CI would fall back to the default
     (`4.19+`), which lacks the new APIs and would fail misleadingly.
- **Note on `.CS3D_REF` file:** verified there is **no file-based mechanism** in either
  repo (grepped OHIF + cs3d trees, CI scripts, workflows, docs) — the ref is carried
  only by the PR body line (`CS3D_REF:` here; symmetric `OHIF_REF:` on the cs3d side).
  A `.CS3D_REF` file would be read by nothing. If a checked-in file is preferred over
  the PR body (it does survive body edits and is visible in the diff), that's a small
  CI change: teach `cs3d-check-integration.sh` to prefer a repo-root `.CS3D_REF` file
  over the body line — optional, separate from this PR.
- **Fix:**
  1. Add the `ohif-integration` label to #5806.
  2. Add a properly formatted line to the PR body (own line, uppercase):
     `CS3D_REF: fix/use-imageLoader-for-seg`
     (if the cs3d branch lives on a fork, use the `owner:branch` form).
  3. At merge time (once cs3d #2611 is released): update the body line to the release
     version, re-run so the merge guard goes green, and apply the pins via
     `node .scripts/cs3d-set-version.mjs <version>` + commit — matching step B of the
     two-step rollout in the PR-2611 plan.
- **Supporting facts kept from verification:** published `5.4.10` tarballs lack both
  APIs; `createFromDicomSegImageId` call site is unguarded
  (`getSopClassHandlerModule.ts:515`, try/finally with no catch) while
  `prefetchPart10Instance` degrades gracefully (`DicomWebDataSource/index.ts:377-388`).
  No code change needed — the merge guard is the mechanism that prevents a registry
  install of this state from ever shipping.
- **Status:**

### [x] 1d. Bogus `NumberOfFrames` stamped onto SR/RTSTRUCT before store/download/clipboard
- **Where:** `extensions/default/src/utils/registerNaturalizedDatasetForLocalWadouri.js:44`
  — `dataset.NumberOfFrames = numberOfFrames` (min 1), plain enumerable assignment,
  unconditional. The `makeExistingPropertiesNonEnumerable` call at `:66` does NOT cover
  it (`NumberOfFrames` is not in `RUNTIME_INSTANCE_PROPERTY_KEYS`, `dicomWriter.ts:18-29`).
- **Verified:** CONFIRMED. All 3 callers are in `createStoreFunction`
  (`commandsModule.ts:788` download, `:801` clipboard, `:824` STOW), and each runs the
  mutation **before** serialization — `store.dicom` denaturalizes **the same mutated
  object** (`DicomWebDataSource/index.ts:469`) and STOWs it (`:481`); `dicomDict` is
  never supplied on the SEG/RTSS path, so no bypass. Affected IODs: **SR** (measurement
  reports) and **RTSTRUCT** gain a bogus `(0028,0008) = 1` in the wire/download/clipboard
  bytes; SEG is harmless (legitimately multiframe). Nuance: the **local** wadouri blob is
  built at `:26`, *before* the mutation, so local copy ≠ stored copy.
- **Fix:** Keep the computed count in a local variable (the per-frame loop at `:57-63` is
  the only consumer) and only write `dataset.NumberOfFrames` when the dataset is actually
  multiframe (`Array.isArray(PerFrameFunctionalGroupsSequence)` — the `:35-37` branch
  already does this) or has PixelData. If the provider truly needs it on the dataset,
  use `setNonEnumerableInstanceProperty` instead. Local-variable option is cleanest.
- **Pairs with:** 4b (test asserting SR datasets don't gain the tag).
- **Status:** FIXED 2026-07-09 — took the local-variable option: removed the
  unconditional `dataset.NumberOfFrames =` write; the `:35-37` multiframe branch
  remains the only dataset stamp. Regression test added (4b); mutation-checked.

---

## Phase 2 — Design concerns (confirmed; decide + fix)

### [x] 2a. Every report save permanently retains a full Part 10 Blob (and more)
- **Where:** `registerNaturalizedDatasetForLocalWadouri.js:26-27` →
  `dicomImageLoader.wadouri.fileManager.add(blob)`; all three `createStoreFunction`
  branches call it (SR, SEG, RTSS; download/clipboard/STOW alike).
- **Verified:** CONFIRMED. What accumulates per save, forever:
  1. The full Part 10 Blob in the fileManager's module-level `files: Blob[]` (SEG: many
     MB; SR/RTSS: KB–MB). `remove(index)` / `purge()` **exist**
     (`libs/@cornerstonejs/packages/dicomImageLoader/src/imageLoader/wadouri/fileManager.ts:13-19`)
     but OHIF only ever calls `add` (3 sites; zero `remove`/`purge` calls anywhere,
     including `onModeExit`).
  2. If the `dicomfile:` id is later loaded: a second full byte copy in
     `dataSetCacheManager.loadedDataSets` (`unload`/`purge` exist, also never called).
  3. Per-frame `imageURI → UIDs` entries in `MetadataProvider.imageURIToUIDs` (Map, no
     removal API; one entry per frame per save for multiframe SEG).
  (Review's "blob URLs" wording imprecise — no `URL.createObjectURL`; Blob references
  via the `dicomfile:N` index. Same memory effect.)
- **Fix:** Two complementary moves:
  1. **Narrow registration** to datasets that need local pixel loading (SEG / has
     PixelData) — SR/RTSS never need a wadouri image-loader entry (review's suggestion,
     confirmed correct).
  2. **Wire up release:** record the returned imageId index and call
     `fileManager.remove(index)` + `dataSetCacheManager.unload(uri)` on display-set/
     study close or `onModeExit` (the APIs exist; symmetric with the cs3d-side 3a fix —
     `clearCacheData()` in `onModeExit` — already applied for the metadata registry).
- **Status:** FIXED 2026-07-09 — both moves: (1) registration skips datasets with no
  PixelData and no PerFrameFunctionalGroupsSequence (SR/RTSS — verified no reader of
  `.url` on those IODs); (2) registered imageIds are tracked and the new
  `releaseLocalWadouriRegistrations()` (fileManager.remove + dataSetCacheManager
  unload-to-zero) runs from extensions/default `onModeExit`. Tests updated (skip /
  register / release / idempotence). `imageURIToUIDs` growth left as-is (no removal
  API; now SEG-only, bounded per session).

### [x] 2b. Viewport-ready promise resolves before the SEG representation is applied
- **Where:** `extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts`
  `addOverlayRepresentationForDisplaySet` (`:1285-1388`): in the
  SEGMENTATION_LOADING_COMPLETE handler, `settle()` at `:1345` runs before
  `await applyRepresentation()` at `:1348`; apply failures downgraded to `console.warn`
  (`:1350-1353`). Timeout (`SEG_LOADING_TIMEOUT_MS = 120000`, `:1325`) and
  loading-failure paths intentionally settle without applying.
- **Verified:** CONFIRMED, with an important nuance: the promise **already waits for the
  SEG pixel-data load**; early-resolve skips only representation attachment. Trivial
  case = one microtask. The real window: `applyRepresentation` can trigger an async
  **stack→volume conversion** (`handleStackViewportCase` → `convertStackToVolumeViewport`
  → wait for GRID_STATE_CHANGED), during which the awaiters proceed:
  `_setStackViewport:943` / `_setVolumeViewport:1219` continue into
  rotation/flip/`render()`/`setPresentations`, and `setViewportData`'s `.then` broadcasts
  **VIEWPORT_DATA_CHANGED** (`:543-548`) — consumed by the hanging-protocol
  `onViewportDataInitialized` counter (`commandsModule.ts:1469-1484`), toolbar refresh,
  overlays, e2e readiness — all before the labelmap actor exists. A readiness-signal lie:
  screenshot/e2e flake + protocol callbacks acting on an overlay-less viewport; plus
  swallowed apply failures are invisible to callers. Not data corruption.
- **Fix (with caveats — do not naively move settle):** keep `settled = true` +
  unsubscribe where they are (dedupe), move only `resolve()` into a `finally` after the
  await, keep resolve-not-reject (best-effort semantics), and do **not** clear the 120 s
  timeout until final resolve so a hung `applyRepresentation` stays bounded. Known
  re-entrancy hazard to test: `convertStackToVolumeViewport` →
  `setDisplaySetsForViewport` → a **new `setViewportData` for the same viewport** while
  the old `_setStackViewport` is suspended at `:943` — resolving after apply means the
  stale stack-setup continuation runs against a rebuilt volume viewport and the old
  VIEWPORT_DATA_CHANGED fires after the replacement setup began. No hard deadlock
  (GRID_STATE_CHANGED is emitted independently), but verify event ordering in the
  conversion scenario before landing. Timeout/failure paths need no change.
- **Status:** FIXED 2026-07-09 exactly per the caveats — `settled` + unsubscribe stay
  at handler entry (dedupe), `resolve()` moved to a `finally` after
  `await applyRepresentation()`, timeout NOT cleared until final resolve (a hung
  apply resolves at the 120 s bound with a distinct warn; double-resolve is a no-op).
  Load-failure / timeout-before-event paths unchanged (`settleWithoutApply`).
  Stack→volume conversion re-entrancy still needs a runtime check (4c open).

### [x] 2c. `loadMultiframeAsPart10RaceTimeMs`: configs ship 3000 ms, doc recommends 250–750 ms
- **Where:** `platform/app/public/config/default.js:91` and `config/e2e.js:118,159,187,214,244,276,309`
  (all 3000); doc `platform/docs/docs/behaviours/segmentation-multiframe-part10-prefetch.md:312-313`
  recommends 250–750 ms.
- **Verified:** CONFIRMED, and the cost is real, not cosmetic:
  `getSopClassHandlerModule.ts:509` does `await Promise.race([prefetch.done, raceTimer])`
  **before** `createFromDicomSegImageId(:515)` — frame decode does NOT start in parallel,
  and registration is not progressive (whole-instance fetch+parse before any frame is
  registered; progressive registration is doc'd as future "rollout step 4"). So on any
  origin where the full-instance fetch takes > 3 s, time-to-first-SEG-render is delayed
  by a flat 3 s with zero offsetting benefit at decode start. Fast origins (e2e
  static-wado) are unaffected — the fetch wins before the timer matters.
- **Fix:** Either lower the shipped configs into the doc's 250–750 range, or keep 3000
  and add a justifying comment in both configs (e.g. demo servers: bulk fetch usually
  wins and avoids hundreds of per-frame requests) — and reconcile the doc's
  recommendation to match whichever is chosen. Pure latency/bandwidth tuning; no
  correctness risk either way.
- **Status:** Lets just have the back end datasource configuration indicate that this is the way to fetch the data, and if so, wait until ti is done, OR it fails, and don't bother with a timeout at all here.
  IMPLEMENTED 2026-07-09 per the directive above — the knob is now the boolean
  `loadMultiframeAsPart10` (data source config, then
  `cornerstone.segmentation.loadMultiframeAsPart10` customization, default **true**):
  when on, the SEG load `await`s `prefetch.done` (completes or fails — no timer at
  all); `false` is the explicit per-frame exception. Race-timer code, the
  `loadMultiframeAsPart10RaceTimeMs` param, and all config occurrences removed;
  behaviours doc + README rewritten to match. Also corrected the bogus "static WADO
  has no efficient per-frame endpoint" comments — the endpoint is efficient, the win
  is that SEG frames are tiny and numerous, so one bulk fetch beats hundreds of tiny
  requests (probed the dev study: the SEG instance is ~424 MB, which is also why any
  finite race cap would always lose and storm per-frame).

---

## Phase 3 — Latent / informational (document or micro-fix; don't block)

### [-] 3a. `frameNumber` made non-enumerable — latent footgun, zero active breakage
- **Where:** `platform/core/src/utils/combineFrameInstance.ts:105-110` (multiframe) and
  `:150-155` (RTDOSE) — `enumerable: true → false` on this branch (commit "feat: load
  DICOM SEG images via imageLoader"); motive is dcmjs serialization cleanliness
  (`dicomWriter.ts:17-29` `RUNTIME_INSTANCE_PROPERTY_KEYS` + `makeExistingPropertiesNonEnumerable`).
- **Verified:** repo-wide audit of every `.frameNumber` consumer (measurement mappings,
  `getSOPInstanceAttributes`, MetadataProvider, SR hydration, SEG handlers, data sources,
  sortStudy, plus cs3d's `MeasurementReport.ts:651` direct-read path): **all read
  directly off originals or plain objects — none break.** Exactly ONE risky copy site:
  `extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts:511-514`
  spreads `metaData.get('instance', image.referencedImageId)` — for frame-qualified
  multiframe ids (exactly what this branch enables) the copy silently drops
  `frameNumber`; but no consumer of `displaySet.images` reads `.frameNumber` today.
- **Related hazard found (not in the review):** the branch also makes
  `instance.imageId/url/wadoRoot/wadoUri` non-enumerable
  (`setNonEnumerableInstanceProperty`, e.g. `DicomWebDataSource/index.ts:634-635,661`),
  and `combineFrameInstance.ts:85,118` builds `_parentInstance` via `{ ...instance }` —
  so combined frame instances now lose those props from the parent copy. No current
  reader found either; same class of latent break.
- **Fix (cheap insurance):** in `SegmentationService.ts:511-514`, don't spread the
  instance — keep a reference, or re-attach explicitly
  (`Object.assign({}, image, inst, { frameNumber: inst?.frameNumber })`). Optionally add
  a `copyInstance()` helper that re-applies `RUNTIME_INSTANCE_PROPERTY_KEYS` after any
  copy, and a one-line comment on `combineFrameInstance` that instances from
  `metaData.get('instance', …)` must not be spread. Acceptable to defer with a note in
  the PR description.
- **Status:** Document, don't change this.  The change was intentional since the frameNumber should NOT be copied.
  DOCUMENTED 2026-07-09 per the directive above — no behavior change. Added a note at
  the `SegmentationService` spread site stating the spread deliberately does NOT copy
  the non-enumerable runtime props (frameNumber in particular must not be carried
  onto derived objects; read it off the original instance), and a matching note in
  the `combineFrameInstance` JSDoc.

### [-] 3b. dcmjs 0.52.0 engines `node >= 22.13`
- **Verified:** premise true (installed dcmjs 0.52.0 declares it; the 0.49.4 → 0.52.0
  bump + lockfile override are on this branch), but the risk **cannot bite this repo**:
  root `engines.node >= 24` is stricter and is what pnpm actually hard-enforces;
  `.node-version` and every CI/Netlify pin are 24.15.0; no `engine-strict` anywhere, so
  dependency engines only warn. (Note: repo is pnpm 11 — the CLAUDE.md "yarn" text is
  stale for this fork.)
- **Action:** none. Optionally note in the PR description that dcmjs 0.52.0 raises the
  floor to Node 22.13 for downstream consumers of the published OHIF packages.
- **Status:** WON'T-FIX (documented here).

---

## Phase 4 — Tests (land alongside their fixes)

### [x] 4a. Regression test for `datasetToDicomPart10Buffer` file meta (pairs with 1a)
- `dicomWriter.test.ts` currently only exercises `writeDicomDictToPart10Buffer`
  fragmentation with a hand-built hex-keyed dict — `applyTransferSyntaxToFileMeta` /
  `datasetToDicomPart10Buffer` / `datasetToDicomBlob` have zero coverage, which is why
  1a shipped. Add: build a naturalized dataset (string-form `_meta.TransferSyntaxUID`
  too, to pin the fallback that makes line 91 necessary), run
  `datasetToDicomPart10Buffer`, re-read with `DicomMessage.readFile`, assert meta keys
  are exactly the 0002 group (no `'00000000'`), `'00020010'` matches, and
  FileMetaInformationGroupLength is consistent. (A working repro script exists from
  verification — trivially convertible to a jest test.)
- **Status:** DONE 2026-07-09 — two tests in `dicomWriter.test.ts`: (1) byte-level meta
  walk (all tags group 0002, ascending, group length spans exactly) + dcmjs re-read
  (no `'00000000'`, `'00020010'` matches); (2) string-form `_meta.TransferSyntaxUID`
  with Implicit VR LE pins the fallback that makes the hex assignment necessary.

### [x] 4b. SR datasets not mutated by local-wadouri registration (pairs with 1d)
- Feed a minimal SR (no PixelData, no PerFrameFunctionalGroupsSequence) through
  `registerNaturalizedDatasetForLocalWadouri` and assert no enumerable `NumberOfFrames`
  appears (and that a multiframe SEG dataset still gets frame imageIds registered).
- **Status:** DONE 2026-07-09 — `registerNaturalizedDatasetForLocalWadouri.test.js`
  (mocks dicom-image-loader + @ohif/core): SR gains no `NumberOfFrames` own-property;
  multiframe SEG gets `NumberOfFrames = 3` (enumerable) + 3 `?frame=` mappings;
  `dataset.url` stays non-enumerable.

### [ ] 4c. (Optional, pairs with 2b) event-ordering test for the settle fix
- If 2b lands, pin the stack→volume conversion scenario: VIEWPORT_DATA_CHANGED must not
  fire before the labelmap actor exists, and a same-viewport re-entrant
  `setViewportData` during conversion must not interleave stale continuations. Likely an
  integration/e2e-level check; scope to what's practical.
- **Status:**

---

## Release / dependency gating (ties into the PR-2611 plan)

- This PR is **step B** of the two-step rollout in
  `libs/@cornerstonejs/PR-2611-review-plan.md`: it hard-depends on a cs3d release
  containing #2611 (`createFromDicomSegImageId`, `prefetchPart10Instance`,
  `@cornerstonejs/metadata` `clearCacheData`). That gating is handled by the CS3D_REF
  workflow + merge guard (see 1c) — merge order is: wire up CS3D_REF on the PR → CI
  green against the branch → release cs3d with #2611 → switch CS3D_REF to the release
  version → `cs3d-set-version.mjs` + commit → merge #5806.
- The local `libs/` symlink setup masks 1b on dev machines — it only reproduces on a
  fresh clone / CI installing from the registry. The CS3D_REF **version-path** run
  (installs from npm, no cs3d:link) doubles as exactly that link-free install check.

---

## Suggested order

1. **1a, 1d** — small, self-contained conformance fixes; land with 4a/4b.
2. **1b** — one-hunk removal (or guard+commit); unblocks fresh clones immediately.
3. **2a, 2c** — mechanical once the policy call is made (narrow registration + wire
   release; pick a race-time value).
4. **2b** — needs the re-entrancy testing described; do after the mechanical items.
5. **3a** — cheap insurance fix or defer with a PR note.
6. **1c** — five-minute PR housekeeping (label + `CS3D_REF:` body line), do anytime;
   the version switch + `cs3d-set-version.mjs` commit happens at merge time.

sedghi (approved):
Great PR, merge after green CI

PR #5816, opened 2026-02-12
fix(dicom-pdf): enhance PDF loading with authentication support

sedghi:
This pull request enhances the handling of authenticated PDF loading in the DICOM PDF extension. The main improvement is that the viewport now supports fetching PDFs with authorization headers when required, ensuring secure access to protected resources. Additionally, the code has been refactored to simplify how dependencies are passed and managed.

PDF authentication and loading improvements:

* The `OHIFCornerstonePdfViewport` component now checks for an authorization header from `userAuthenticationService` and fetches the PDF using authenticated requests when needed. If authentication fails or is not required, it gracefully falls back to using the direct URL. Temporary object URLs are properly revoked on cleanup to prevent memory leaks.
* The component now uses the `servicesManager` from the system context to access authentication services, supporting more robust integration with the application's service architecture.

Code simplification:

* The `getViewportModule` function in `dicomPDFExtension` no longer requires `servicesManager` or `extensionManager` as parameters, simplifying its interface and usage.

<h3>Greptile Summary</h3>

This PR adds authentication support to the DICOM PDF viewport by fetching the PDF blob with authorization headers from `userAuthenticationService` and serving it as an object URL. The refactoring to use `useSystem()` inside the component is clean and follows OHIF conventions — `index.tsx` correctly simplifies `getViewportModule` by removing the now-unused `servicesManager`/`extensionManager` parameters.

Key changes:
- `OHIFCornerstonePdfViewport` now uses `useSystem()` to obtain `servicesManager` and `userAuthenticationService`
- When an `Authorization` header is present, the PDF is fetched with auth headers and exposed as a `blob:` object URL
- Object URLs are revoked in the cleanup function to prevent memory leaks
- `isCancelled` flag guards against state updates on unmounted components
- There is a race-condition memory leak: if the effect cleanup fires while `fetch()` or `response.blob()` is still in-flight, `objectUrl` is `undefined` at cleanup time, so `URL.revokeObjectURL` is skipped; when the async continuation later creates the object URL and sees `isCancelled === true`, it skips `setUrl` but never revokes the newly created object URL

Safe to merge after fixing the object URL memory leak in the authenticated fetch path

One P1 memory leak exists: objectUrl can be created after the cleanup function has already run (finding objectUrl still undefined), and is therefore never revoked. The fix is a two-line change. The P2 AbortController issue is a best-practice improvement but not blocking.

extensions/dicom-pdf/src/viewports/OHIFCornerstonePdfViewport.tsx — specifically the objectUrl creation/revocation logic at lines 70-74

| Filename | Overview |
|----------|----------|
| extensions/dicom-pdf/src/viewports/OHIFCornerstonePdfViewport.tsx | Adds authenticated PDF fetching via userAuthenticationService with object URL management; has a memory leak when cleanup fires mid-fetch before objectUrl is assigned |
| extensions/dicom-pdf/src/index.tsx | Simplifies getViewportModule by removing unused servicesManager/extensionManager params now that the component uses useSystem() directly — no issues |

</details>

</details>

`````markdown
Path: extensions/dicom-pdf/src/viewports/OHIFCornerstonePdfViewport.tsx
Line: 70-74

Comment:
**Memory leak when cleanup fires during in-flight fetch**

If the effect cleanup runs while `fetch()` or `response.blob()` is still awaiting (e.g. on unmount or `renderedUrl` change), `objectUrl` is still `undefined` at that moment. The cleanup's `URL.revokeObjectURL(objectUrl)` call on lines 87–89 is therefore a no-op. When the async continuation later resumes and reaches line 70, it creates the object URL, checks `isCancelled === true`, and correctly skips `setUrl` — but it **never revokes the newly created object URL**, permanently leaking that blob.

Fix: revoke `objectUrl` immediately if `isCancelled` is already `true` when it is created:

```suggestion
        objectUrl = URL.createObjectURL(blob);

        if (isCancelled) {
          URL.revokeObjectURL(objectUrl);
          return;
        }
        setUrl(objectUrl);
```

---

Path: extensions/dicom-pdf/src/viewports/OHIFCornerstonePdfViewport.tsx
Line: 60-63

Comment:
**In-flight fetch not aborted on cleanup**

The `fetch` request is never aborted when the effect cleanup runs (on unmount or dependency change). The network request continues to consume bandwidth and a connection slot after the component no longer needs the data. Consider using an `AbortController`:

```js
const controller = new AbortController();
const response = await fetch(resolvedUrl, {
  headers: authHeaders,
  signal: controller.signal,
});
// ...
return () => {
  isCancelled = true;
  controller.abort();
  if (objectUrl) {
    URL.revokeObjectURL(objectUrl);
  }
};
```

Note: with `AbortController`, the `catch` block should distinguish `AbortError` (ignore/return) from real errors (fall back to direct URL) to avoid incorrectly setting the URL after an intentional abort.

`````

</details>

<sub>**[Greploops](https://www.greptile.com/docs/mcp-v2/skills#greploop)** — Automatically fix all review issues by running `/greploops` in [Claude Code](https://claude.ai/download). It iterates: fix, push, re-review, repeat until 5/5 confidence.<br>Use the **[Greptile plugin for Claude Code](https://www.greptile.com/docs/integrations/claude-code#claude-code-plugin)** to query reviews, search comments, and manage custom context directly from your terminal.</sub>

<sub>Reviews (2): Last reviewed commit: ["Merge branch &#39;master&#39; into fix/rendered-..."](https://github.com/ohif/viewers/commit/847a1a9ac5ee2c91097d03d583a2b1f8a3108ba2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=22143663)</sub>

## Summary by CodeRabbit

* **New Features**
  * PDF and video viewports now use an async rendered-media path (`getRenderedUrl` / `getVideoUrl`) when available, enabling secure playback when credentials exist.
  * Rendered media URL retrieval is now supported end-to-end via the DICOMweb `renderedURL` endpoint, with fallback behavior where needed.
* **Bug Fixes**
  * Improved cancellation/cleanup for rendered/video loading, including safer resource revocation.
  * Added stronger handling for authorization failures and untrusted URLs to avoid unnecessary fetching.
* **Documentation**
  * Updated migration guides for removal of the standalone DICOM video viewport module.
* **Tests**
  * Added Jest coverage for rendered-media URL retrieval behavior.


extensions/dicom-pdf/src/viewports/OHIFCornerstonePdfViewport.tsx
wayfarer3130, line 1: Can you put the fetch stuff into the data source rather than using it here?  There is already at least one example like that for fetching the thumbnail in the data source.  That way it can be used for the video as well by just calling the same method with the right arguments.  Also, set to null if there isn't a response so if you update hte pdf and it fails to load it gets cleared.

extensions/dicom-video/src/viewports/OHIFCornerstoneVideoViewport.tsx
wayfarer3130, line 1: Can we just delete the dicom-video extension now, since CS3D directly supports video?

extensions/dicom-video/src/getSopClassHandlerModule.js
wayfarer3130, line 1: We might still need to keep this sop class handler module, but we shouldn't need the custom viewport any longer, nor do we want to because it does weird things with annotaitons etc.

wayfarer3130 (approved):
Consider deleting the OHIFCornerstoneVideoViewport, but otherwise I think this looks good.

PR #5832, opened 2026-02-22
fix(mode-route): validate studies after route init to support custom token

Sofien-Sellami:
### Context

This pull request corrects the order of two operations in `platform/app/src/routes/Mode/Mode.tsx`. The mode route validated the studies of the URL before the mode route completed the route setup.

Some modes set a custom authentication token. Such a mode sets the token in `onModeInit`, in `onModeEnter` or in `route.init`. The route setup calls all three of these hooks. The validation ran before the route setup, so the validation query did not have the token. The query failed, and the mode crashed.

The mode route now runs the validation after `setupRouteInit()` resolves. The validation query then has the token.

#### The order of the validation

- The mode route calls the validation after `setupRouteInit()` resolves.
- The mode route calls `mode.onSetupRouteComplete` in the same order as OHIF 3.13. The validation does not block `mode.onSetupRouteComplete`.

#### The new `validateModeEntry` hook

The mode route validated the studies of the URL for every mode, and a mode could not replace that check. OHIF added this check in #5752. A mode that does not load its data by StudyInstanceUID always failed the check.

The mode route now reads an optional `validateModeEntry` function from the mode object:

```js
const validateModeEntry = mode?.validateModeEntry ?? validateStudies;
```

- The mode route uses the study check as the default. A mode that does not declare the function keeps the behaviour of OHIF 3.13.
- The hook can be an async function.
- The hook can navigate away on its own. The mode route gives `navigate` to the hook.
- A mode that does not load its data by StudyInstanceUID declares the hook. Such a mode runs its own check, or such a mode gives an empty function to skip the check.

The mode route gives these properties to the hook: `studyInstanceUIDs`, `dataSource`, `navigate`, `servicesManager`, `extensionManager`, `commandsManager`, `appConfig` and `query`. `platform/app/src/routes/Mode/Mode.tsx` exports the type `ValidateModeEntryProps`.

#### Other corrections

- `validateStudies` moved out of the component, and the function takes its inputs as parameters. The effect that calls the function no longer needs the function in the dependency array. The lint scope of the React Compiler holds `platform/app/src/**`, and the rule `react-hooks/exhaustive-deps` reports an error for such a function.
- The promise chain has one `.then` again. The previous chain had two `.then` calls, but the first `.then` did not return the promise of `validateStudies()`. The second `.then` never waited for the validation.
- The mode route handles a rejection of the hook. The mode route also handles an error that the hook throws before the hook returns its promise.
- The mode route guards `navigate` with the `isMounted` ref. The validation starts later than before, so a navigation can happen after an unmount.

#### The base of the branch

This branch now holds a merge of `origin/master` at `3.14.0-beta.30`. The previous base was `3.13.0-beta.20`, and the previous base was 302 commits old. The checks of this pull request need a new run.

#### Documentation

- `platform/docs/docs/platform/modes/lifecycle.md` documents `validateModeEntry`. The page holds the table of the properties and an example.
- `platform/docs/docs/platform/modes/validity.md` separates `validateModeEntry` from `isValidMode`. `isValidMode` runs in the work list, and `validateModeEntry` runs when the user enters the mode.

### Testing

1. Open a study in a mode. Confirm that the viewer shows the study.
2. Open a URL that holds a StudyInstanceUID that does not exist. Confirm that the viewer navigates to `/notfoundstudy`.
3. Add a `validateModeEntry` function to a mode. Confirm that the mode route calls that function, and confirm that the mode route does not run the study check.
4. Use a mode that sets a custom authentication token. Confirm that the mode starts, and confirm that the validation query uses the token.

These checks ran on the branch:

- `npx tsc --noEmit -p tsconfig.json` reports no error for `platform/app/src/routes/Mode/Mode.tsx`.
- `npx prettier --check platform/app/src/routes/Mode/Mode.tsx` passes.

This check did not run: `pnpm lint:compiler`. The package `eslint-plugin-react-hooks` was absent from the local install.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: 
- [x] Node version: 
- [x] Browser:
  

## Summary by CodeRabbit

- **New Features**
  - Added validation when entering a mode to verify requested study data.
  - Missing studies or failed validation checks redirect to the not-found study page.
  - Modes can provide custom validation for data loaded through URL parameters.
  - Validation runs without delaying metadata retrieval and safely ignores navigation after leaving the mode.

- **Documentation**
  - Documented mode-entry validation, its timing, behavior, and relationship to existing mode validity checks.

### DIFF
--- a/platform/app/src/routes/Mode/Mode.tsx
+++ b/platform/app/src/routes/Mode/Mode.tsx
@@ -313,7 +313,7 @@ export default function ModeRoute({
 
         validateStudies()
       })
-      .then(unsubs => {
+      .then(() => {
 
         mode?.onSetupRouteComplete?.({
           servicesManager,
--- a/platform/app/src/routes/Mode/Mode.tsx
+++ b/platform/app/src/routes/Mode/Mode.tsx
@@ -137,33 +137,25 @@ export default function ModeRoute({
    * Validates study existence before loading the viewer.
    * Moved from PanelStudyBrowser.tsx to ensure validation runs in all modes
    */
-  useEffect(() => {
-    if (!ExtensionDependenciesLoaded || !studyInstanceUIDs?.length || !dataSource) {
-      return;
-    }
+  const validateStudies = async () => {
+    for (const studyInstanceUID of studyInstanceUIDs) {
+      try {
+        const qidoForStudyUID = await dataSource.query.studies.search({
+          studyInstanceUid: studyInstanceUID,
+        });
 
-    const validateStudies = async () => {
-      for (const studyInstanceUID of studyInstanceUIDs) {
-        try {
-          const qidoForStudyUID = await dataSource.query.studies.search({
-            studyInstanceUid: studyInstanceUID,
-          });
-
-          if (!qidoForStudyUID?.length) {
-            console.warn('Study not found:', studyInstanceUID);
-            navigate('/notfoundstudy');
-            return;
-          }
-        } catch (error) {
-          console.error('Error validating study:', studyInstanceUID, error);
+        if (!qidoForStudyUID?.length) {
+          console.warn('Study not found:', studyInstanceUID);
           navigate('/notfoundstudy');
           return;
         }
+      } catch (error) {
+        console.error('Error validating study:', studyInstanceUID, error);
+        navigate('/notfoundstudy');
+        return;
       }
-    };
-
-    validateStudies();
-  }, [studyInstanceUIDs, ExtensionDependenciesLoaded, dataSource, navigate]);
+    }
+  };
 
   useEffect(() => {
     if (!ExtensionDependenciesLoaded || !studyInstanceUIDs?.length) {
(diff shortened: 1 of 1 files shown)


salimkanoun:
@sedghi or @wayfarer3130 , can you give us our opinion about this PR ?
We have a problem here, the new validation process is made before Mode startup. On our side we need to make an custom auth process we used onRouteInit until now but that's happen after the new validation (so the request fails because we can generate the token we need to reach dicom web api)

One other way would be to use onModeInit but it is not async for now so we can call our custom API to generate the access token. 

I think in all cases we need some policy to say if you need a custom authentication you need to use this callback that will be fired before any dicom web request

Thanks for your help

sedghi:
@jbocce  can you review and merge when you can

platform/app/src/routes/Mode/Mode.tsx
```diff
@@ -137,33 +137,29 @@ export default function ModeRoute({
    * Validates study existence before loading the viewer.
    * Moved from PanelStudyBrowser.tsx to ensure validation runs in all modes
    */
-  useEffect(() => {
+  const validateStudies = async () => {
```
wayfarer3130: This can't go here - the problem is that a mode might not even be dealing with qido or studies - it might be doing something patient centric or use some other validation mechanism.  Even the studyInstanceUIDs?.length is probably wrong.
What should be done is that there should be a mode function 'validateModeEntry', which you can default to the studyInstanceUIDs validate check here if it is undefined, but that way there is a standard validation that can be replaced/over-written if needed.

wayfarer3130 (commented):
It looks like the validate studies is pre-existing, just in a different spot.  Given that, I will finish the review and maybe make a small change but generally get this included shortly.

wayfarer3130 (approved):

PR #5834, opened 2026-02-23
test(ContourSegNavigation): Add e2e tests for contour segmentation navigation

diattamo:
### Context

This pull request introduces end-to-end tests for contour segmentation navigation and adds new test IDs to the viewport action arrows for improved testability. The most important changes are summarized below.

### Changes & Results

* Added a new test suite `ContourSegNavigation.spec.ts` that verifies navigation through segmentation contours, both by clicking segments in the right panel and using the viewport action arrows. The tests check for correct SVG path updates and wraparound behavior.

* Extended the `DOMOverlayPageObject` with a `segmentationArrowNavigation` getter, providing convenient access to the left and right navigation buttons for the new tests.

* Added `data-cy` test IDs (`viewport-action-arrows`, `viewport-action-arrows-left`, and `viewport-action-arrows-right`) to the `ViewportActionArrows` component to facilitate reliable element selection in end-to-end tests.

### Testing

1. Run the ContourSegNavigation e2e tests:
   `bun test:e2e` or more specifically
   `bun test:e2e:ui -- tests/ContourSegNavigation.spec.ts`

#### PR

#### Code

  etc.)

The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [x] OS:  macOS 26.1
- [x] Node version:  v25.4.0
- [x] Browser:  Chrome 141.0.7390.37 (arm64)

<h3>Greptile Summary</h3>

This PR adds end-to-end tests for contour segmentation navigation in the OHIF Viewer. It introduces a new test suite (`ContourSegNavigation.spec.ts`) that validates two navigation paths: clicking segments in the right panel and using viewport action arrow buttons. Supporting changes include `data-cy` test ID attributes on `ViewportActionArrows` and a new `segmentationArrowNavigation` page object getter.

- **New test suite** covers right-panel segment clicking (with idempotency checks) and viewport arrow navigation (with forward/backward wraparound verification)
- **Test IDs added** to `ViewportActionArrows.tsx` (`viewport-action-arrows`, `viewport-action-arrows-left`, `viewport-action-arrows-right`) using the project's `data-cy` convention confirmed by `playwright.config.ts`
- **Page object extended** with `segmentationArrowNavigation` getter in `DOMOverlayPageObject`, following the established pattern of other getters like `segmentationHydration`
- **Minor style issues**: inconsistent indentation in the test file at lines 26-30 and 67

- This PR is safe to merge — it only adds test infrastructure and non-functional `data-cy` attributes with no production logic changes.
- The changes are limited to adding e2e tests and data-cy attributes. No production logic is modified. The test patterns follow existing conventions in the codebase. Only minor style inconsistencies were found.
- No files require special attention. `tests/ContourSegNavigation.spec.ts` has minor formatting inconsistencies that don't affect functionality.

| Filename | Overview |
|----------|----------|
| platform/ui-next/src/components/Viewport/ViewportActionArrows.tsx | Added `data-cy` test ID attributes to the arrow container and individual left/right arrow wrapper divs for e2e test targeting. No functional changes. |
| tests/ContourSegNavigation.spec.ts | New e2e test suite for contour segmentation navigation (right panel clicks and viewport arrow buttons). Minor indentation inconsistencies found on lines 26-30 and 67. |
| tests/pages/DOMOverlayPageObject.ts | Added `segmentationArrowNavigation` getter following the same pattern as existing getters (e.g., `segmentationHydration`). Clean implementation. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Test as E2E Test
    participant LP as Left Panel
    participant DOM as DOM Overlay
    participant RP as Right Panel
    participant VP as Viewport

    Note over Test: beforeEach setup
    Test->>LP: loadSeriesByModality('RTSTRUCT')
    Test->>DOM: segmentationHydration.yes.click()
    Test->>RP: nthSegment(0).click()
    Test->>VP: hover on default pane

    Note over Test: Test 1: Right Panel Navigation
    loop For each segment (0-3)
        Test->>RP: nthSegment(i).click()
        Test->>VP: getSvgPath() → verify non-null
    end
    loop Revisit segments (0-3)
        Test->>RP: nthSegment(i).click()
        Test->>VP: getSvgPath() → verify matches original
    end

    Note over Test: Test 2: Arrow Navigation
    loop Navigate forward (next × 4)
        Test->>DOM: segmentationArrowNavigation.next.click()
        Test->>VP: getSvgPath() → verify path changed
    end
    Note over Test: Wraparound: next returns to first
    Test->>DOM: segmentationArrowNavigation.next.click()
    Test->>VP: getSvgPath() → verify matches initial

    Note over Test: Wraparound: prev returns to last
    Test->>DOM: segmentationArrowNavigation.prev.click()
    Test->>VP: getSvgPath() → verify matches fourth

    loop Navigate backward (prev × 3)
        Test->>DOM: segmentationArrowNavigation.prev.click()
        Test->>VP: getSvgPath() → verify matches expected
    end
```

<sub>Last reviewed commit: 054ec78</sub>

<sub>(2/5) Greptile learns from your feedback when you react with thumbs up/down!</sub>

### DIFF
--- a/tests/utils/assertions.ts
+++ b/tests/utils/assertions.ts
@@ -63,5 +63,5 @@ export async function assertBoundingBoxIsContainedWithin({
 }
 
 export async function expectRowSelected(rowObject) {
-    await expect(rowObject.locator).toContainClass('bg-popover');
+  await expect(rowObject.locator).toContainClass('bg-popover');
 }


platform/ui-next/src/components/Viewport/ViewportActionArrows.tsx
```diff
@@ -12,11 +12,20 @@ const arrowClasses =
  */
 function ViewportActionArrows({ onArrowsClick, className }) {
   return (
-    <div className={classNames(className, 'flex')}>
-      <div className={arrowClasses}>
+    <div
+      data-cy="viewport-action-arrows"
```
jbocce, line 16: There is a chance that these could be visible/present in the DOM for more than one viewport. In that case these would not be unique. We should look into whether we can add the viewport id or something to make them unique.

tests/pages/DOMOverlayPageObject.ts
```diff
@@ -132,6 +132,26 @@ export class DOMOverlayPageObject {
           },
         };
       },
+      get segmentationArrowNavigation() {
```
jbocce: I know it is not straightforward, but the navigation arrows are actually part of the viewport and NOT overlaying it. So the arrows belong in `ViewportPageObject` (see ViewportPageObject.ts). In particular see the `getOverlayMenu` - it should now return something with a `get arrowsNavigation()` or something.

jbocce: Note that the arrows are NOT particular to segmentation, so let's leave segmentation out of the name. Also, having this in the viewport page object might mean we do not have to worry about a unique data-cy value for the arrows because maybe you can search/locate the arrow elements from the viewport locator.

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+  const mode = 'segmentation';
+  await visitStudy(page, studyInstanceUID, mode, 2000);
+  await leftPanelPageObject.loadSeriesByModality('RTSTRUCT');
+  await page.waitForTimeout(5000);
+  await DOMOverlayPageObject.viewport.segmentationHydration.yes.click();
+  await page.waitForTimeout(5000);
+  // Click segment 0 in the right panel to establish a known starting position
+  await rightPanelPageObject.contourSegmentationPanel.panel.nthSegment(0).click();
+  await page.waitForTimeout(500);
+  await viewportPageObject.getById('default').pane.hover();
```
jbocce: Is this necessary? If so, why and maybe it needs to be baked into something else?

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+const getSvgPath = async (viewportPageObject) => {
+  const viewportId = 'default';
+  return viewportPageObject
+    .getById(viewportId)
+    .svg('path')
+    .first()
+    .getAttribute('d');
+}
+
+test('should navigate through the contours when clicking through segments in the right panel', async ({
```
jbocce: Too many uses of the word "through". The first one can simply be deleted. The second one can be replace with "each".

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  await rightPanelPageObject.contourSegmentationPanel.panel.nthSegment(2).click();
+  const seg2 = await getSvgPath(viewportPageObject);
+  expect(seg2, 'Segment at index 2: expected a non-null SVG path').not.toBeNull();
+
+  await rightPanelPageObject.contourSegmentationPanel.panel.nthSegment(3).click();
+  const seg3 = await getSvgPath(viewportPageObject);
+  expect(seg3, 'Segment at index 3: expected a non-null SVG path').not.toBeNull();
+
+  // Clicking segments again should return to the original paths
```
jbocce: I love how you checked the navigation here. How about we add a little twist and instead of clicking the segments in the same order as what appears above, we change the order? The expectations should not change but it will show that order does not matter.

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+  await rightPanelPageObject.contourSegmentationPanel.panel.nthSegment(0).click();
+  const seg0Again = await getSvgPath(viewportPageObject);
+  expect(seg0Again, 'Segment 0 again: expected to match the original segment 0 path').toBe(seg0);
```
jbocce: I think we should also assert/expect each segment to be highlighted after being clicked.

Image

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+  await rightPanelPageObject.contourSegmentationPanel.panel.nthSegment(2).click();
+  const seg2Again = await getSvgPath(viewportPageObject);
+  expect(seg2Again, 'Segment 2 again: expected to match the original segment 2 path').toBe(seg2);
+
+  await rightPanelPageObject.contourSegmentationPanel.panel.nthSegment(3).click();
+  const seg3Again = await getSvgPath(viewportPageObject);
+  expect(seg3Again, 'Segment 3 again: expected to match the original segment 3 path').toBe(seg3);
+});
+
+test('should navigate through the segmentations using the Viewport action buttons', async ({
```
jbocce: Again remove the through and instead of "action buttons" it should be "arrow buttons"

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+  page,
+  DOMOverlayPageObject,
+  rightPanelPageObject,
+  viewportPageObject,
+}) => {
+  const navigateAndRefocus = async (direction: 'next' | 'prev') => {
+    await DOMOverlayPageObject.viewport.segmentationArrowNavigation[direction].click();
+    // Refocus the viewport after clicking the navigation buttons to ensure cursor remains on the default viewport
+    await viewportPageObject.getById('default').pane.hover();
```
jbocce: I'm curious what this hover is actually permitting for this test. The only thing I can see that it might affect for these tests is whether the SVG path can be extracted or not, but that seems weird. Please let me know why we need this. Thanks.

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  // Wraparound test — prev from first should return to last segment
+  await navigateAndRefocus('prev');
+  const svgPathWraparoundWithPrev = await getSvgPath(viewportPageObject);
+  expect(
+    svgPathWraparoundWithPrev,
+    'Expected svg path to match the fourth svg path after wrapping around with prev navigation'
+  ).toBe(fourthSvgPath);
+
+  await navigateAndRefocus('prev');
```
jbocce: Once again, nice job on this test. We should also added expectations for the selected segment in the side panel as well (i.e. like I mentioned in the test above).

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+  await page.waitForTimeout(5000);
+  await DOMOverlayPageObject.viewport.segmentationHydration.yes.click();
+  await page.waitForTimeout(5000);
+  // Click segment 0 in the right panel to establish a known starting position
+  await rightPanelPageObject.contourSegmentationPanel.panel.nthSegment(0).click();
+  await page.waitForTimeout(500);
+  await viewportPageObject.getById('default').pane.hover();
+});
+
+const getSvgPath = async (viewportPageObject) => {
```
jbocce: This might be a good candidate to add to our utils directory.

jbocce:
Be sure to update your branch as new tests and page objects are being added by others.

tests/pages/DOMOverlayPageObject.ts
diattamo: Good call! Moving it to the ViewportPageObject made it a lot easier and takes care of making sure we are hitting the slected viewport's buttons

tests/ContourSegNavigation.spec.ts
diattamo: I noticed after click events on the Cypress, the mouse pointer does not stay on the viewport of interest. The default position it goes back to happens to make it that it hovers over the next viewport hence triggering the issue you called out about that the arrow buttons are visible/present in the DOM for another viewport.
This was to keep the mouse pointer on the default viewport. 
It is not needed anymore with moving the code to the ViewportPageObject.

tests/ContourSegNavigation.spec.ts
diattamo: Fixed

tests/ContourSegNavigation.spec.ts
diattamo: I changed the order around here.

tests/ContourSegNavigation.spec.ts
diattamo: Added the asserts on both tests to verify this.

tests/ContourSegNavigation.spec.ts
diattamo: Fixed

tests/ContourSegNavigation.spec.ts
diattamo: I removed it per the previous comment. It is solved by accessing the buttons with the ViewportPageObject.

tests/ContourSegNavigation.spec.ts
diattamo: refactored to make it a util function

tests/pages/ViewportPageObject.ts
```diff
@@ -135,6 +146,28 @@ export class ViewportPageObject {
     return viewport.locator(`svg.svg-layer${innerElement ? ` ${innerElement}` : ''}`);
   }
 
+  private getNavigationArrows() {
```
jbocce: I think we want to pass the `viewport` as a parameter here and use it instead of `page` in the calls below.

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  const getSegment = (index: number) =>
+    rightPanelPageObject.contourSegmentationPanel.panel.nthSegment(index);
+
+  const seg0 = await getSvgPath(viewportPageObject);
+  expect(seg0, 'Segment at index 0: expected a non-null SVG path').not.toBeNull();
+
+  await getSegment(3).click();
+  const seg3 = await getSvgPath(viewportPageObject);
+  expect(seg3, 'Segment at index 3: expected a non-null SVG path').not.toBeNull();
+  await expect(getSegment(3).locator).toHaveClass(/bg-popover/);
```
jbocce: I think we should extend the panel row page object (i.e. that is the actual segment page object I think) (see getPanelRowDataObject) to have maybe an `expectActive` or `expectSelected` method instead. That is where the `toHaveClass` can be used.

tests/ContourSegNavigation.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+});
+
+test('should navigate the segmentations using the Viewport arrow buttons', async ({
+  rightPanelPageObject,
+  viewportPageObject,
+}) => {
+  const getSegment = (index: number) =>
+    rightPanelPageObject.contourSegmentationPanel.panel.nthSegment(index);
+
+  const navigateWithArrow = async (direction: 'next' | 'prev') => {
```
jbocce: This looks like another good utility. Please move to the utils dir.

diattamo: Done

tests/pages/ViewportPageObject.ts
diattamo: Change the method to use viewport as a parameter

tests/ContourSegNavigation.spec.ts
diattamo: Done. I went with checking that the highlight is visible on the row object.

tests/pages/RightPanelPageObject.ts
```diff
@@ -53,6 +53,9 @@ export class RightPanelPageObject {
       toggleVisibility: async () => {
         await row.getByTestId('data-row-visibility-toggle').click();
       },
+      get selectedIndicator() {
+        return row.and(row.page().locator('.bg-popover'));
```
jbocce: Can't or shouldn't this just be `return row.locator('.bg.popover')` or am I missing something? I find what is here now not easy to read.

platform/ui-next/src/components/DataRow/DataRow.tsx
```diff
@@ -244,6 +244,7 @@ const DataRowComponent = React.forwardRef<HTMLDivElement, DataRowProps>(
           } group relative cursor-pointer`}
           onClick={onSelect}
           data-cy="data-row"
+          data-selected={isSelected}
```
jbocce: I really don't want to add another attribute for this. Can we not use the `.bg-popover` class or similar from before?

jbocce (changes requested):
See my last comment.

tests/pages/RightPanelPageObject.ts
diattamo: That was my initial approach but itdoes not work. Found out that Playwright's `row.locator('.bg-popover')` searches for the class on the children of row, but bg-popover is a class on the row element itself. The approach I had checked for the class on the page and then verifies that the locator is the same one as the row locator (hence the .and...)
I changed the approach to add an attribute to the DataRow and check directly with it.

jbocce: I understand, but I think what we ultimately want is something like...

`await expect(getSegment(3).locator).toHaveClass('bg-popover');`

in our tests. I think it is best to add an `expectRowSelected` in the `assertions.ts` that the tests call that does something like the above give a row page object or something.

PR #5861, opened 2026-02-27
fix(segmentation): segment bidirectional tool error and show toast notification when no segment is drawn

### LINKED ISSUE #5593: [Bug] Toast message displayed when clicking on segment bidirectional tool without any segments
### Describe the Bug

If not segments are added to an image yet and the segment bidirectional tool is clicked a 'something went wrong' message is displayed 

### Steps to Reproduce

1. Display a study in segmentation 
2. In label map click add segmentation and click the segment bidirectional tool 

### The current behavior

A something went wrong message is displayed 

Image

### The expected behavior

Tool shouldn't be available if cannot be used

### System Information

  System:
    OS: Windows 11 10.0.26100
    Memory: 4.34 GB / 31.46 GB
  Browsers:
    Chrome: 142.0.7444.176

GhadeerAlbattarni:
### Context

Fixes #5593

When the user clicks on  the **Segment Bidirectional**  button without having drawn any segment in the viewport, it shows an error message. 
Now it shows a warning message that asks user to draw a segment in order to compute the Bidirectional data. 

### Changes & Results
Added two guards in `runSegmentBidirectional` (commandsModule.ts):

- **Pre-call guard** : use `getUniqueSegmentIndices` to check whether the active segment actually has painted voxels in the labelmap before calling `getSegmentLargestBidirectional`. If no voxels exist, a warning notification will be shown. 

- **Post-call guard**: if `getSegmentLargestBidirectional` returns an empty array for any reason, return early instead of attempting to access `.annotationUID` on an undefined value.

### Testing
1. open a study in a segmentation mode 
2. Click on add segmentation
3. Click on Segment Bidirectional button on the right side panel
4. Verify you will get a warning message. 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.12.0

<h3>Greptile Summary</h3>

Adds validation guards to prevent crashes when using Segment Bidirectional tool without drawn segments.

**Key changes:**
- Pre-call guard checks if segment has voxels using `getUniqueSegmentIndices` and shows warning notification if not
- Post-call guard attempts to handle empty `bidirectionalData` array

**Critical issues:**
- The post-call guard on lines 428-430 is placed **inside** the `forEach` loop, meaning it will never execute when `bidirectionalData` is empty (the exact scenario it's meant to protect against)
- Missing null check before accessing `activeBidirectional.annotationUID` on line 467, which will crash if `find()` returns undefined
- Minor: inconsistent i18n key formatting (space after colon doesn't match codebase pattern)

The pre-call guard works correctly, but the post-call safety mechanism is broken due to incorrect placement.

<h3>Confidence Score: 2/5</h3>

- This PR has critical logic errors that prevent the safety guards from working correctly
- While the intent is good (add guards to prevent crashes), the implementation has a critical bug where the post-call guard is placed inside the forEach loop, making it ineffective when the array is empty. Additionally, there's a missing null check that could still cause crashes. The pre-call guard works correctly, but the post-call safety mechanism is broken.
- extensions/cornerstone/src/commandsModule.ts requires immediate attention to fix the incorrectly placed post-call guard and add missing null check

| Filename | Overview |
|----------|----------|
| extensions/cornerstone/src/commandsModule.ts | Added guards for segment bidirectional tool, but post-call guard is incorrectly placed inside forEach loop (never executes when array is empty) and missing null check for activeBidirectional |

</details>

</details>

<sub>Last reviewed commit: 1c10f2d</sub>


jbocce:
@hongxiao-gif thanks for testing this. However, from the look of the screen captures, it appears you are running viewer-dev, but you should be running the deploy preview for this PR. The deploy preview is found [here](https://github.com/OHIF/Viewers/pull/5861#issuecomment-3974194373), but I will leave it here as a convenience... https://deploy-preview-5861--ohif-dev.netlify.app/

Also, the message you should expect should read... "Draw a segment before using bidirectional measurement"

FYI @GhadeerAlbattarni . Please add any other information for Hong that I might have missed, but I think I covered everything.

GhadeerAlbattarni:
Thanks @hongxiao-gif for testing this PR. Also, thanks @jbocce for your answer. 
A small note, with this fix the success message "Computing Largest Bidirectional completed successfully" will not appear if the user didn't draw any segment on the viewport.

extensions/cornerstone/src/commandsModule.ts
```diff
@@ -398,6 +398,21 @@ function commandsModule({
...
 
+      // Check if the segment has voxels before computing bidirectional measurement
+      const uniqueSegmentIndices = cstUtils.segmentation.getUniqueSegmentIndices(targetId);
+      const hasVoxels = uniqueSegmentIndices.includes(targetIndex);
+
+      if (!hasVoxels) {
+        uiNotificationService.show({
+          title: i18n.t('SegmentationPanel:Segment Bidirectional'),
+          message: i18n.t(
+            'SegmentationPanel: Draw a segment before using bidirectional measurement'
```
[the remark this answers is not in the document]
jbocce: This is a good catch by the greptile. @GhadeerAlbattarni please update this if you haven't already. Also, these strings need to be put in JSON files for translation purposes. Please add the strings you added to platform\i18n\src\locales\en-US\SegmentationPanel.json and also platform\i18n\src\locales\test-LNG\SegmentationPanel.json and any other languages you might know 🤣. Just kidding.

tests/Bidirectional.spec.ts
jbocce, line 1: This file name should better reflect the tool name. So maybe SegmentBidirectionalTool.spec.ts?

tests/Bidirectional.spec.ts
```diff
@@ -24,3 +31,24 @@ test('should display the bidirectional tool', async ({
...
+    await visitStudy(page, studyInstanceUID, mode, 2000);
+  });
+
+  test('should not show an error when Segment Bidirectional is clicked without a segment being drawn', async ({
+    page,
+    rightPanelPageObject,
+  }) => {
+    await rightPanelPageObject.labelMapSegmentationPanel.addSegmentationButton.click();
+
+    const errorMessage = waitForError(page, 3000);
```
jbocce: So there are a couple of things here:

- I think we want to wait for the warning message you added instead of waiting for no error message
- I think we need a page object for these popup messages in general - probably defined in DOMOverlayPageObject; That page object might do things asynchronously like grab the type of popup (e.g. warning, error, info, etc.), title and message so that they can be asserted later.

extensions/cornerstone/src/commandsModule.ts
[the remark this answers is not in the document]
GhadeerAlbattarni: Good morning Joe! I added before greptile feedback and I just added the strings to these files. Thanks for mentioning them

tests/Bidirectional.spec.ts
GhadeerAlbattarni: The notification messages does not appear in the test environment, so I couldn't detect them and that's why I went for the error message because it's possible to get through `ErrorBoundary. We can talk more about it today :)

jbocce: Ah good catch. Well if there are no popups in the test environment, then we shouldn't even have a test for this. I think that means most of the testing code can be removed. I would however leave the bidirectional button page object you added in RightPanelPageObject. So sorry about this, but thank you for looking into this and the effort you put into trying to add a test.

GhadeerAlbattarni: Sounds good, thank you :)

PR #5866, opened 2026-03-02
fix(microscopy): rename measurement in microscopy mode

### LINKED ISSUE #5797: [Bug] Rename measurement in microscopy doesn't work
### Describe the Bug

see title.

### Steps to Reproduce

1. Display a study in microscopy .e.g https://viewer-dev.ohif.org/microscopy?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.296485376.1.1665793212.499772
2. Add a measurement 
3. On the measurement panel select Rename. Enter in text and save 

https://github.com/user-attachments/assets/b63cd000-8f0c-4e18-8aab-0488e59f00c0

### The current behavior

Measurement still says empty

### The expected behavior

Measurement is renamed successfully or rename is not available if it's not supported in this mode.

### System Information

  System:
    OS: Windows 11 10.0.26100
    Memory: 7.62 GB / 31.46 GB
  Browsers:
    Chrome: 144.0.7559.133

GhadeerAlbattarni:
### Context

Fixes #5797

When the user adds a measurement in microscopy mode, then clicks on rename, it's not working. This PR fixes this issue. 

### Changes & Results
Updated the `setLabel` in `getCommandsModule.ts` to: 
- Use `async/await` with the current promise-based `callInputDialog` API
- Pre-fill the dialog with the annotation's existing label 

https://github.com/user-attachments/assets/fb13cd21-d14c-4b87-a3f6-dc700adfcb37

### Testing
1. Open a study in microscopy mode
2. Draw any annotation 
3. In the right panel, click on `...` then click on rename 
4. Enter a new label and click Save
5. Verify the label updates in the panel
6. Reopen the dialog and verify the current label is pre-filled
7. Click Cancel and verify the label does not change

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.12.0

<h3>Greptile Summary</h3>

Fixed the rename measurement functionality in microscopy mode by converting the `setLabel` command from callback-based to async/await and pre-filling the input dialog with the annotation's existing label.

**Key changes:**
- Made `setLabel` function async and uses `await` with `callInputDialog`
- Changed `defaultValue` from empty string to `roiAnnotation.label ?? ''` to pre-fill existing label
- Added proper null/undefined check for the returned value before applying the label

**Issue found:**
- Missing null check for `roiAnnotation` before accessing its properties could cause runtime error if annotation doesn't exist

- Safe to merge with one defensive check recommended
- The fix properly addresses the reported issue by pre-filling the dialog and using async/await. However, missing a null check for `roiAnnotation` could cause a runtime error in edge cases.
- The `getCommandsModule.ts` file needs a defensive null check for `roiAnnotation` before accessing its properties

| Filename | Overview |
|----------|----------|
| extensions/dicom-microscopy/src/getCommandsModule.ts | Updated `setLabel` to use async/await and pre-fill dialog with existing label |

</details>

</details>

<sub>Last reviewed commit: 5fccd9d</sub>

### DIFF
--- a/tests/MicroscopyPanel.spec.ts
+++ b/tests/MicroscopyPanel.spec.ts
@@ -23,11 +23,11 @@ test('should rename a microscopy measurement label', async ({
   const measurementRow = rightPanelPageObject.microscopyPanel.nthMeasurement(0);
   await expect(measurementRow.locator).toBeVisible();
 
-  await expect(measurementRow.locator).toContainText('(empty)');
+  await expect(measurementRow.title).toHaveText('(empty)');
 
   await measurementRow.actions.rename(newLabel);
 
   await expect(DOMOverlayPageObject.dialog.input.locator).toBeHidden();
 
-  await expect(measurementRow.locator).toContainText(newLabel);
+  await expect(measurementRow.title).toHaveText(newLabel);
 });
--- a/platform/ui-next/src/components/DataRow/DataRow.tsx
+++ b/platform/ui-next/src/components/DataRow/DataRow.tsx
@@ -280,6 +280,7 @@ const DataRowComponent = React.forwardRef<HTMLDivElement, DataRowProps>(
               <Tooltip>
                 <TooltipTrigger asChild>
                   <span
+                    data-cy="data-row-title"
                     className={`cursor-default text-base ${
                       isSelected ? 'text-highlight' : 'text-muted-foreground'
                     } [overflow:hidden] [display:-webkit-box] [-webkit-line-clamp:2] [-webkit-box-orient:vertical]`}
@@ -296,6 +297,7 @@ const DataRowComponent = React.forwardRef<HTMLDivElement, DataRowProps>(
               </Tooltip>
             ) : (
               <span
+                data-cy="data-row-title"
                 className={`text-base ${
                   isSelected ? 'text-highlight' : 'text-muted-foreground'
                 } [overflow:hidden] [display:-webkit-box] [-webkit-line-clamp:2] [-webkit-box-orient:vertical]`}
--- a/tests/pages/RightPanelPageObject.ts
+++ b/tests/pages/RightPanelPageObject.ts
@@ -46,6 +46,9 @@ export class RightPanelPageObject {
       get actions() {
         return getActionsMenu(row);
       },
+      get title() {
+        return row.getByTestId('data-row-title');
+      },
       click: async () => {
         await row.click();
       },


extensions/dicom-microscopy/src/getCommandsModule.ts
```diff
@@ -24,16 +24,20 @@ export default function getCommandsModule({
...
         uiDialogService,
-        defaultValue: '',
-        onSave: (value: string) => {
-          roiAnnotation.setLabel(value);
-          microscopyService.triggerRelabel(roiAnnotation);
-        },
+        defaultValue: roiAnnotation.label ?? '',
       });
+
+      if (value !== undefined && value !== null) {
```
jbocce: Do we do something similar for other, non-microscopy annotations if the value is null or undefined? Let's make sure and be consistent.

Also a more concise way to do the if is...
`if(value != null)`

However, what about is value is the empty string? In any case, let's first see what we do for other annotation renames.

tests/pages/MainToolbarPageObject.ts
```diff
@@ -205,6 +204,17 @@ export class MainToolbarPageObject {
           },
         };
       },
+      get line() {
```
jbocce, line 212: Why is this so different than for the other tools? Why can't we do the following?

```
      get line() {
        const button = page.getByTestId('line');
        return {
          button,
          async click() {
            await measurementTools.click();
            await button.click();
          },
        };
      },
```

tests/MicroscopyPanel.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  const measurementRow = rightPanelPageObject.microscopyPanel.nthMeasurement(0);
+  await expect(measurementRow.locator).toBeVisible();
+
+  await expect(measurementRow.locator).toContainText('(empty)');
+
+  await measurementRow.actions.rename(newLabel);
+
+  await expect(DOMOverlayPageObject.dialog.input.locator).toBeHidden();
+
+  await expect(measurementRow.locator).toContainText(newLabel);
```
jbocce: Maybe we should use toHaveText instead since we want it to be precisely the label.

jbocce:
@hongxiao-gif here is another for you to test. Thank you so much ahead of time!

hongxiao-gif (commented):
**Passed**
Testing steps:
1. Open a study in microscopy mode
2. Draw any annotation
2. In the right panel, click on ... then click on rename
3. Enter a new label and click Save
4. Verify the label updates in the panel
5. Reopen the dialog and verify the current label is pre-filled
6. Click Cancel and verify the label does not change

Testing environment:
 System:
    OS: Windows 11 10.0.26200
    Memory: 4.34 GB / 15.67 GB
  Browsers:
    Chrome: 145.0.7632.117
    Edge: Chromium (144.0.3719.104)

https://github.com/user-attachments/assets/46664257-9f53-4f21-a499-349e43cf8099

extensions/dicom-microscopy/src/getCommandsModule.ts
GhadeerAlbattarni: - Yes, in  `extensions/cornerstone/src/commandsModule.ts`, `handleMeasurementLabelDialog` uses same pattern. However looking at it again now, the current implementation can be even improved by adding `title` and `placeholder` similar to annotation. 

- When the value is an empty string, the value will set to default `(empty)` and it's same for other modes.

tests/pages/MainToolbarPageObject.ts
GhadeerAlbattarni: `line` tool is the default active in `microscopy` mode, so when I tried to do same pattern with other tools, I got an error, because it rendered in two places and Playwright finds both elements, so it throws an error.

However the other tools in the `microscopy` could follow same pattern for other tools and I tested that.

Alternatively, I could use `selectedTool` in `measurementTools` instead and it would work as it's just clicking the default, but I thought it might be useful to explicit show what tool has been used.

PR #5875, opened 2026-03-04
fix(seg hydration): auto-hydrate RT struct on second load with disableConfirmationPrompts

### LINKED ISSUE #5844: [Bug] Viewport goes blank when deleting an RT Struct twice with disableConfirmationPrompts.
### Describe the Bug

When the `disableConfirmationPrompts: true` setting is active, deleting a RT Struct for the second time causes the viewer resulting in a blank screen.

https://github.com/user-attachments/assets/e5ac0f8d-b688-412b-a0e1-fbd2181b833e

### Steps to Reproduce

**Pre Condition:** `disableConfirmationPrompts: true` flag should be enabled.

1. Launch the viewer with RT struct Study
2. Drag and drop the RT struct
3. Delete the RT Struct
4. Again drag and drop the RT struct

### The current behavior

Observe that

1. On loading the RT struct for the second time it is not hydrated properly
2. On deleting the RT struct also removes the image from the viewport

### The expected behavior

RT struct should be hydrated.
On deleting the RT struct the image should not be removed from the viewport

### System Information

    OS: Linux 6.6 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 593.00 MB / 7.63 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash

GhadeerAlbattarni:
### Context

Fixes #5844

This PR fixes a bug where RTSTRUCT segmentations would fail to hydrate (showing a "Load" badge instead of rendering) on the second load when the `disableConfirmationPrompts` flag is set to `true`. Also, it fixes the viewport goes blank after the user deletes the segmentation on the second load. 

**The issue:**
When `disableConfirmationPrompts` is enabled, the hydration logic executes synchronously as soon as the RT data is "Loaded".
- On the first load, the data loading delay provides enough time for React to mount the viewport and for its `useEffect`s to run.
- On the second load (when data is cached/returned instantly), the hydration command was firing immediately—before the `OHIFCornerstoneViewport` component had finished its render cycle and registered its ID with the `CornerstoneViewportService`. This resulted in viewport being undefined in the `hydrateSecondaryDisplaySet` in `extensions/cornerstone/src/commandsModule.ts` , causing the hydration to fail and leading to a black screen if the user subsequently tried to delete the RT.

### Changes & Results
In `promptHydrationDialog.ts`,  wrapped the `RTSTRUCT` hydration callback in `window.setTimeout(..., 0)` to ensure the viewport is fully initialized and registered in the `CornerstoneViewportService` before hydration happens. This aligns with the existing pattern for `SEG` hydration

**Before**

https://github.com/user-attachments/assets/6aef2257-0a8d-4d6d-a515-8494498a7013

### After

https://github.com/user-attachments/assets/51dfd23f-abd3-43c7-8856-376663c3b4e8

### Testing
1. Set `disableConfirmationPrompts: true` in the app configuration.
2. Open a study with `RT STRUCT` in a `viewer` mode (e.g., `StudyInstanceUIDs=1.2.840.113619.2.290.3.3767434740.226.1600859119.501)`
3. Load a study with an `RT STRUCT` (it should hydrate right away on the first load)
4. Open right side panel, then delete the segmentations. 
5. Verify the viewport does not go blank 

6. Load the same study again
7. Verify that the RT Struct hydrates automatically, without showing the "Load" badge at the top of the viewport 
8. Delete the segmentations from the right side panel 
9.  Verify the viewport does NOT go blank, and still has the `displayset` 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.12.0

<h3>Greptile Summary</h3>

This PR fixes a race condition in `extensions/cornerstone/src/utils/promptHydrationDialog.ts` where RTSTRUCT segmentations would fail to hydrate on the second load when `disableConfirmationPrompts` is `true`. The root cause was that with the flag enabled, the hydration callback fired synchronously before `OHIFCornerstoneViewport` had completed its render cycle and registered its ID with `CornerstoneViewportService`, leaving `getViewport` returning `undefined` and causing a blank viewport on subsequent delete operations.

Key changes:
- Wraps the RTSTRUCT `hydrateCallback` invocation in `window.setTimeout(..., 0)`, deferring execution to the next event loop tick — identical to the pattern already used for SEG hydration in the same function.
- The fix is minimal, targeted, and consistent with the established codebase pattern.
- **Note**: The `async` callback inside `window.setTimeout` for both RTSTRUCT (new) and SEG (pre-existing) lacks a `try/catch`, meaning a rejected `hydrateCallback` will silently swallow the error and leave the outer `Promise` permanently unresolved. Adding `try/catch` with a `reject(error)` call in both branches would improve robustness.

- This PR is safe to merge; it applies a well-understood one-liner fix that mirrors the existing SEG hydration pattern.
- The change is a single, minimal, targeted fix that correctly addresses the described race condition. It is consistent with the established SEG hydration pattern in the same file. The only outstanding concern is the absence of `try/catch` error handling inside the `setTimeout` async callback (shared with the pre-existing SEG branch), which could leave the outer Promise unresolved if `hydrateCallback` throws — but this does not represent a regression introduced by this PR.
- No files require special attention beyond the minor error-handling suggestion in `extensions/cornerstone/src/utils/promptHydrationDialog.ts`.

| Filename | Overview |
|----------|----------|
| extensions/cornerstone/src/utils/promptHydrationDialog.ts | Wraps RTSTRUCT hydration in `window.setTimeout(..., 0)` to match the existing SEG pattern and fix the race condition on second load with `disableConfirmationPrompts`; the `async` callback inside the timeout lacks a `try/catch`, meaning a rejected `hydrateCallback` will leave the outer Promise unresolved. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant App
    participant promptHydrationDialog
    participant CornerstoneViewportService
    participant OHIFCornerstoneViewport
    participant hydrateCallback

    App->>promptHydrationDialog: call (disableConfirmationPrompts=true)
    Note over promptHydrationDialog: shouldPrompt = false → RESPONSE.HYDRATE immediately

    rect rgb(255, 220, 220)
        Note over promptHydrationDialog: BEFORE fix (2nd load): hydration fires synchronously
        promptHydrationDialog->>hydrateCallback: await hydrateCallback() [RTSTRUCT]
        hydrateCallback->>CornerstoneViewportService: getViewport(viewportId)
        CornerstoneViewportService-->>hydrateCallback: undefined (viewport not yet registered!)
        hydrateCallback-->>promptHydrationDialog: hydration fails / black screen on delete
    end

    rect rgb(200, 255, 200)
        Note over promptHydrationDialog: AFTER fix: deferred via setTimeout(0)
        promptHydrationDialog->>promptHydrationDialog: window.setTimeout(..., 0)
        OHIFCornerstoneViewport->>CornerstoneViewportService: register viewportId (useEffect runs)
        promptHydrationDialog->>hydrateCallback: await hydrateCallback() [RTSTRUCT]
        hydrateCallback->>CornerstoneViewportService: getViewport(viewportId)
        CornerstoneViewportService-->>hydrateCallback: viewport (now registered ✓)
        hydrateCallback-->>promptHydrationDialog: isHydrated = true
        promptHydrationDialog-->>App: resolve(true)
    end
```

<sub>Last reviewed commit: d0db6d1</sub>


extensions/cornerstone/src/utils/promptHydrationDialog.ts
```diff
@@ -103,13 +103,15 @@ function promptHydrationDialog({
...
+        window.setTimeout(async () => {
+          const isHydrated = await hydrateCallback({
+            rtDisplaySet: displaySet,
+            viewportId,
+            servicesManager,
+          });
 
-        resolve(isHydrated);
+          resolve(isHydrated);
+        }, 0);
```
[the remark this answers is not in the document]
jbocce: Yes I think this is a good suggestion. Let's change it here and above.

platform/ui-next/src/components/SegmentationTable/SegmentationCollapsed.tsx
```diff
@@ -38,6 +38,7 @@ const SegmentationCollapsedDropdownMenu = ({ children }: { children: React.React
         <Button
           variant="ghost"
           size="icon"
+          data-cy="segmentation-collapsed-more-btn"
```
jbocce: I think we should add the segmentation type if possible.

tests/pages/RightPanelPageObject.ts
```diff
@@ -118,6 +118,10 @@ export class RightPanelPageObject {
     const getSegmentByText = (text: string) => this.getPanelRowByText(text);
 
     return {
+      deleteAll: async () => {
```
jbocce: I think just delete to match the label in the UI is best.

jbocce: I wonder if it makes sense to refactor out a "collapsed more menu" page object. I think it does make sense. 😊

tests/RTHydrationDisableConfirmation.spec.ts
```diff
@@ -0,0 +1,81 @@
+import { checkForScreenshot, expect, screenShotPaths, test, visitStudy } from './utils';
+
+test.beforeEach(async ({ page }) => {
```
jbocce, line 11: This is awesome. Great stuff! Let's kick it up a notch and move this to the utilities - say OHIFConfiguration.ts. I suppose the utility function would take a `test` object and the (partial) configuration object to be added to the base config.

tests/screenshots/chromium/RTHydrationDisableConfirmation.spec.ts/firstLoadPostHydration.png
jbocce, line 1: Because the blue square is so subtle, I think it would be best to navigate to another image with more contours and use it as the screenshot comparison. Use something similar to the following to navigate to a different image...

` await press({ page, key: 'ArrowDown', nTimes: 12 });`

GhadeerAlbattarni:
Thanks for your comments, I'll work on the requested changes.

tests/pages/RightPanelPageObject.ts
```diff
@@ -112,12 +136,14 @@ export class RightPanelPageObject {
     };
   }
 
-  private get segmentationPanel() {
+  private getSegmentationPanel(typeSuffix?: string) {
```
jbocce, line 139: I love how you did this. Some day (not now) we should refactor the opacity and border tool page objects to do something similar.

jbocce (approved):
Great stuff. Thanks for taking the time to update the PR the way you did!

PR #5900, opened 2026-03-16
fix(segmentation): restrict overlay segmentation menu to same frame of reference as viewport background display set 

### LINKED ISSUE #5890: [Bug] Segmentation menu allows selecting segmentations from a different frame of reference
### Describe the Bug

The segmentation overlay menu should not allow the user to select a segmentation that has a different Frame of Reference than the underlying display set.

https://github.com/user-attachments/assets/1d399d1d-9ff7-4664-9299-caafe61895ee

### Steps to Reproduce

1. Open this study in a viewer or segmentation mode. 
https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.12201.1091.126683095609223531686845324113579088978
2. Open the Data Overlay menu 
3. Click on add segmentation 
4. Notice that the segmentation menu has mix of FOR1 and FOR2 

### The current behavior

- The segmentation overlay menu lists segmentations from different frames of reference, including segmentations that do not match the viewport’s underlying display set.
- When the user selects a segmentation with a different FOR than the underlying display set, it shows an error message:
"Error: Volumes being added to viewport default do not share the same FrameOfReferenceUID. This is not yet supported"

### The expected behavior

The segmentation overlay menu should only list segmentations that share the same Frame of Reference as the viewport’s underlying display set.

### System Information

OS: macOS 10.15.4
Node version: v22.13.0
Browser: Chrome 83.0.4103.116

GhadeerAlbattarni:
### Context

Fixes #5890 

This PR fixes a bug where the segmentation overlay menu allowed users to select segmentations that did not share the same Frame of Reference (FOR) as the underlying viewport display set. 

Also, in segmentation mode, the hydrate segmentation synchronizer added segmentations to all viewports regardless of their Frame of Reference, causing segmentations to appear immediately in viewports with a different FOR.

### Root Cause

- `SEG` and `RTSTRUCT` display sets had no `FrameOfReferenceUID` so the FOR filter in `getEnhancedDisplaySets` silently passed every SEG/RT and mark them as `isOverlayable: true` since the the check never gets true, the check always  (`undefined && ...)` in `extensions/cornerstone/src/components/ViewportDataOverlaySettingMenu/utils.ts`
 
- The hydrate segmentation synchronizer did not check  whether the target viewport matched the source viewport's FOR. So the segmentations were added to all viewports with different FOR.

### Changes & Results
- Added `FrameOfReferenceUID` to the `DisplaySet` and ensured `SEG` and `RTSTRUCT` Sop class handlers set it (from instance and/or referenced display set).
In `getEnhancedDisplaySets`, overlay display sets (SEG/RTSTRUCT) whose `FrameOfReferenceUID` does not match the viewport background’s are marked non-overlayable, so they no longer appear as selectable in the overlay segmentation menu.
- The overlay menu’s optimistic state (optimisticOverlayDisplaySets) resets when the user changes the background display set, so the list stays in sync with the current viewport.
- Added FOR matching guard to the hydrate segmentation synchronizer to prevent the hydration synchronizer from blindly mirroring segmentations from a source viewport to a target viewport if their primary Frames of Reference do not align.

**After**

https://github.com/user-attachments/assets/2bf0d61b-da65-480b-88f7-cc6cd6ef4797

### Testing
1.  Open a study with multiple frames of reference in a segmentation mode
(e.g., StudyInstanceUIDs=1.3.6.1.4.1.12201.1091.126683095609223531686845324113579088978)
2. Open the overlay menu on a viewport 
3. Click on add segmentation  
4. Verify only segmentations matching that viewport’s background FOR should be listed.

5. Change the viewport background display set
6. Verify the segmentation display set will reset
7. Select a segmentation

8. Open another Viewport overlay menu
9. Verify that the selected segmentation (from previous viewport) was add only to the matching viewport display set FOR

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.13.0

<h3>Greptile Summary</h3>

This PR fixes a bug where the segmentation overlay menu allowed selecting segmentations with a different Frame of Reference (FOR) than the viewport's background display set, and where the hydrate segmentation synchronizer propagated segmentations to viewports with mismatched FORs.

- **SEG/RTSTRUCT `FrameOfReferenceUID` propagation**: Both the SEG and RTSTRUCT SOP class handlers now populate `FrameOfReferenceUID` on their display sets — from the instance metadata directly and from the referenced display set as a fallback. This enables the existing FOR check in `getEnhancedDisplaySets` (`utils.ts`) to properly filter non-matching overlays.
- **Optimistic state reset on background change**: A `useEffect` in `ViewportDataOverlayMenu` resets the optimistic overlay display set list when the background display set changes, keeping the UI in sync. However, `overlayDisplaySets` is missing from the effect's dependency array.
- **Hydration synchronizer FOR guard**: The `createHydrateSegmentationSynchronizer` now compares source and target viewport FORs before propagating segmentation representations, preventing cross-FOR hydration.
- **Type update**: `FrameOfReferenceUID` added as an optional property to the core `DisplaySet` type.

- This PR is safe to merge with one minor issue to address in the useEffect dependency array.
- The core fix is well-reasoned and addresses a real bug across multiple layers (SOP class handlers, overlay menu, synchronizer). The changes are consistent and minimal. The one issue found is a missing React hook dependency that could lead to stale state in an edge case.
- Pay close attention to `extensions/cornerstone/src/components/ViewportDataOverlaySettingMenu/ViewportDataOverlayMenu.tsx` — the `useEffect` dependency array is incomplete.

| Filename | Overview |
|----------|----------|
| extensions/cornerstone/src/components/ViewportDataOverlaySettingMenu/ViewportDataOverlayMenu.tsx | Adds useEffect to reset optimistic overlay state on background change, but the `overlayDisplaySets` dependency is missing from the effect's dependency array, which could cause stale state. |
| extensions/cornerstone-dicom-seg/src/getSopClassHandlerModule.ts | Correctly adds FrameOfReferenceUID from the instance to SEG display sets, with proper fallback to the referenced display set's FOR when available. |
| extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts | Adds FrameOfReferenceUID to RTSTRUCT display sets from ReferencedFrameOfReferenceSequence and referenced display set, consistent with the SEG handler pattern. |
| extensions/cornerstone/src/services/SyncGroupService/createHydrateSegmentationSynchronizer.ts | Adds a FOR matching guard to prevent hydration synchronizer from propagating segmentations across viewports with different frames of reference. The logic is correct and uses safe optional chaining for the source viewport. |
| platform/core/src/types/DisplaySet.ts | Adds the optional FrameOfReferenceUID property to the DisplaySet type definition, enabling type-safe FOR filtering across the codebase. |

</details>

</details>

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[SEG/RTSTRUCT Instance Loaded] --> B[SOP Class Handler]
    B --> C[Set FrameOfReferenceUID on DisplaySet]
    C --> D{Referenced DisplaySet Available?}
    D -->|Yes| E[Copy FOR from Referenced DisplaySet]
    D -->|No| F[Subscribe for DISPLAY_SETS_ADDED event]
    F --> G[On event: Copy FOR from added DisplaySet]
    
    E --> H[getEnhancedDisplaySets]
    G --> H
    
    H --> I{displaySet.FrameOfReferenceUID matches background?}
    I -->|Yes or undefined| J[isOverlayable: true]
    I -->|No| K[isOverlayable: false]
    
    J --> L[Shown in Overlay Menu]
    K --> M[Hidden from Overlay Menu]
    
    N[Segmentation Hydration Sync] --> O{Shared DisplaySet exists?}
    O -->|Yes| P[Add segmentation to target viewport]
    O -->|No| Q{Source FOR == Target FOR?}
    Q -->|Yes| P
    Q -->|No| R[Skip - different Frame of Reference]
```

<sub>Last reviewed commit: 9e5ae01</sub>

### DIFF
--- a/extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
+++ b/extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
@@ -56,7 +56,7 @@ function _getDisplaySetsFromSeries(
     StudyInstanceUID,
     SOPClassHandlerId,
     SOPClassUID,
-    FrameOfReferenceUID: null,
+    frameOfReferenceUID: null,
     referencedImages: null,
     referencedSeriesInstanceUID: null,
     referencedDisplaySetInstanceUID: null,
@@ -96,7 +96,7 @@ function _getDisplaySetsFromSeries(
   displaySet.referencedImages = instance.ReferencedSeriesSequence.ReferencedInstanceSequence;
   displaySet.referencedSeriesInstanceUID = referencedSeries.SeriesInstanceUID;
 
-  displaySet.FrameOfReferenceUID =
+  displaySet.frameOfReferenceUID =
     instance.ReferencedFrameOfReferenceSequence?.[0]?.FrameOfReferenceUID;
 
   const { displaySetService } = servicesManager.services;
@@ -117,7 +117,7 @@ function _getDisplaySetsFromSeries(
         if (addedDisplaySet.SeriesInstanceUID === displaySet.referencedSeriesInstanceUID) {
           displaySet.referencedDisplaySetInstanceUID = addedDisplaySet.displaySetInstanceUID;
           displaySet.isReconstructable = addedDisplaySet.isReconstructable;
-          displaySet.FrameOfReferenceUID = addedDisplaySet.FrameOfReferenceUID;
+          displaySet.frameOfReferenceUID = addedDisplaySet.frameOfReferenceUID;
           unsubscribe();
         }
       }
@@ -126,7 +126,7 @@ function _getDisplaySetsFromSeries(
     const [referencedDisplaySet] = referencedDisplaySets;
     displaySet.referencedDisplaySetInstanceUID = referencedDisplaySet.displaySetInstanceUID;
     displaySet.isReconstructable = referencedDisplaySet.isReconstructable;
-    displaySet.FrameOfReferenceUID = referencedDisplaySet.FrameOfReferenceUID;
+    displaySet.frameOfReferenceUID = referencedDisplaySet.frameOfReferenceUID;
   }
 
   displaySet.load = ({ headers, createSegmentation = true }) =>
--- a/extensions/cornerstone/src/components/ViewportDataOverlaySettingMenu/utils.ts
+++ b/extensions/cornerstone/src/components/ViewportDataOverlaySettingMenu/utils.ts
@@ -80,8 +80,8 @@ export function getEnhancedDisplaySets({ viewportId, services }) {
 
     // Check if Frame of Reference matches
     if (
-      displaySet.FrameOfReferenceUID &&
-      displaySet.FrameOfReferenceUID !== backgroundDisplaySet.FrameOfReferenceUID
+      displaySet.frameOfReferenceUID &&
+      displaySet.frameOfReferenceUID !== backgroundDisplaySet.frameOfReferenceUID
     ) {
       return {
         ...displaySet,
--- a/platform/core/src/types/DisplaySet.ts
+++ b/platform/core/src/types/DisplaySet.ts
@@ -57,7 +57,11 @@ export type DisplaySet = {
    */
   referencedDisplaySetInstanceUID?: string;
 
-  FrameOfReferenceUID?: string;
+  /**
+   * The FrameOfReferenceUID shared by every frame within this display set.
+   * It will be undefined if the frames do not all share the same Frame of Reference.
+   */
+  frameOfReferenceUID?: string;
 
   SeriesDate?: string;
   SeriesTime?: string;
--- a/extensions/cornerstone-dicom-seg/src/getSopClassHandlerModule.ts
+++ b/extensions/cornerstone-dicom-seg/src/getSopClassHandlerModule.ts
@@ -50,7 +50,7 @@ function _getDisplaySetsFromSeries(
     StudyInstanceUID,
     SOPClassHandlerId,
     SOPClassUID,
-    FrameOfReferenceUID,
+    frameOfReferenceUID: FrameOfReferenceUID,
     referencedImages: null,
     referencedSeriesInstanceUID: null,
     referencedDisplaySetInstanceUID: null,
@@ -106,7 +106,7 @@ function _getDisplaySetsFromSeries(
         if (addedDisplaySet.SeriesInstanceUID === displaySet.referencedSeriesInstanceUID) {
           displaySet.referencedDisplaySetInstanceUID = addedDisplaySet.displaySetInstanceUID;
           displaySet.isReconstructable = addedDisplaySet.isReconstructable;
-          displaySet.FrameOfReferenceUID = addedDisplaySet.FrameOfReferenceUID;
+          displaySet.frameOfReferenceUID = addedDisplaySet.frameOfReferenceUID;
           unsubscribe();
         }
       }
@@ -114,7 +114,7 @@ function _getDisplaySetsFromSeries(
   } else {
     displaySet.referencedDisplaySetInstanceUID = referencedDisplaySet.displaySetInstanceUID;
     displaySet.isReconstructable = referencedDisplaySet.isReconstructable;
-    displaySet.FrameOfReferenceUID = referencedDisplaySet.FrameOfReferenceUID;
+    displaySet.frameOfReferenceUID = referencedDisplaySet.frameOfReferenceUID;
   }
 
   displaySet.load = async ({ headers }) =>
--- a/extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
+++ b/extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
@@ -230,7 +230,7 @@ async function _load(
 
 function _measurementBelongsToDisplaySet({ measurement, displaySet }) {
   return (
-    measurement.coords[0].ReferencedFrameOfReferenceSequence === displaySet.FrameOfReferenceUID
+    measurement.coords[0].ReferencedFrameOfReferenceSequence === displaySet.frameOfReferenceUID
   );
 }
 
--- a/extensions/cornerstone-dicom-sr/src/utils/hydrateStructuredReport.ts
+++ b/extensions/cornerstone-dicom-sr/src/utils/hydrateStructuredReport.ts
@@ -145,7 +145,7 @@ export default function hydrateStructuredReport(
 
   for (const FrameOfReferenceUID of frameOfReferenceUIDs) {
     const displaySetsFOR = displaySetService.getDisplaySetsBy(
-      ds => ds.FrameOfReferenceUID === FrameOfReferenceUID && !ds.isDerivedDisplaySet
+      ds => ds.frameOfReferenceUID === FrameOfReferenceUID && !ds.isDerivedDisplaySet
     );
     const ds = chooseDisplaySet(displaySetsFOR, FrameOfReferenceUID);
     if (!ds) {
@@ -335,7 +335,7 @@ function getReferenceData3D(toolData, servicesManager: Types.ServicesManager) {
   const { points } = toolData.annotation.data.handles;
   const { displaySetService } = servicesManager.services;
   const displaySetsFOR = displaySetService.getDisplaySetsBy(
-    ds => ds.FrameOfReferenceUID === FrameOfReferenceUID
+    ds => ds.frameOfReferenceUID === FrameOfReferenceUID
   );
   if (!displaySetsFOR.length || !points?.length) {
     return {
--- a/extensions/cornerstone/src/getSopClassHandlerModule.js
+++ b/extensions/cornerstone/src/getSopClassHandlerModule.js
@@ -82,7 +82,7 @@ function _getDisplaySetsFromSeries(instances, servicesManager, extensionManager)
     SOPInstanceUID,
     SeriesInstanceUID,
     StudyInstanceUID,
-    FrameOfReferenceUID,
+    frameOfReferenceUID: FrameOfReferenceUID,
     SOPClassHandlerId,
     SOPClassUID,
     SeriesDescription: SeriesDescription || 'Microscopy Data',
--- a/extensions/default/src/getSopClassHandlerModule.js
+++ b/extensions/default/src/getSopClassHandlerModule.js
@@ -113,7 +113,7 @@ const makeDisplaySet = (instances, index) => {
     label:
       instance.SeriesDescription ||
       `${i18n.t('Series')} ${instance.SeriesNumber} - ${i18n.t(instance.Modality)}`,
-    FrameOfReferenceUID: instance.FrameOfReferenceUID,
+    frameOfReferenceUID: instance.FrameOfReferenceUID,
   });
 
   const imageIds = dataSource.getImageIdsForDisplaySet(imageSet);
--- a/extensions/dicom-microscopy/src/services/MicroscopyService.ts
+++ b/extensions/dicom-microscopy/src/services/MicroscopyService.ts
@@ -329,16 +329,16 @@ export default class MicroscopyService extends PubSubService {
     const studyMetadata = DicomMetadataStore.getStudy(StudyInstanceUID);
     const smDisplaySet = displaySets.find(ds => ds.Modality === 'SM');
 
-    const { FrameOfReferenceUID, othersFrameOfReferenceUID } = smDisplaySet;
+    const { frameOfReferenceUID, othersFrameOfReferenceUID } = smDisplaySet;
 
     if (!studyMetadata) {
       return;
     }
 
-    let derivedDisplaySets = FrameOfReferenceUID
+    let derivedDisplaySets = frameOfReferenceUID
       ? displaySets.filter(
           ds =>
-            ds.ReferencedFrameOfReferenceUID === FrameOfReferenceUID ||
+            ds.ReferencedFrameOfReferenceUID === frameOfReferenceUID ||
             // sometimes each depth instance has the different FrameOfReferenceID
             othersFrameOfReferenceUID.includes(ds.ReferencedFrameOfReferenceUID)
         )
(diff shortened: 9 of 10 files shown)


extensions/cornerstone/src/components/ViewportDataOverlaySettingMenu/ViewportDataOverlayMenu.tsx
```diff
@@ -43,6 +43,10 @@ function ViewportDataOverlayMenu({ viewportId }: withAppTypes<{ viewportId: stri
 
   const [thresholdOpacityEnabled, setThresholdOpacityEnabled] = useState(false);
 
+  useEffect(() => {
+    setOptimisticOverlayDisplaySets(overlayDisplaySets);
+  }, [backgroundDisplaySet?.displaySetInstanceUID]);
```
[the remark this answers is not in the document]
jbocce: @greptileai is correct here. Please add the missing dependency.

tests/pages/DataOverlayPageObject.ts
```diff
@@ -26,4 +29,25 @@ export class DataOverlayPageObject {
...
+    await this.page.getByText('SELECT A SEGMENTATION').click();
+  }
+
+  async getDropdownOptionLabels(): Promise<string[]> {
+    const options = this.page.getByRole('option');
+    const labels = (await options.allTextContents()).map(text => text.trim()).filter(Boolean);
+    return labels;
+  }
+
+  getActiveSegmentationLocator(label: string) {
```
jbocce: I am not crazy about the word `Locator` in this method name. Maybe ask AI for a better name?

GhadeerAlbattarni: I changed the name here :)

tests/DataOverlayMenuFrameOfReference.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  const studyInstanceUID = '1.3.6.1.4.1.12201.1091.126683095609223531686845324113579088978';
+  const mode = 'segmentation';
+  await visitStudy(page, studyInstanceUID, mode, 2000);
+});
+
+test('should only show segmentations matching the background Frame of Reference', async ({
+  page,
+  viewportPageObject,
+}) => {
+  const viewportId = await viewportPageObject.getIdOfNth(0);
```
jbocce: I am not crazy about this `getIdOfNth` method. Could we not simply use `await viewportPageObject.getNth(0)` and then get the overlay object for it?

tests/pages/ViewportPageObject.ts
```diff
@@ -321,6 +321,12 @@ export class ViewportPageObject {
     return this.viewportPageObjectFactory(viewport);
   }
 
+  async getIdOfNth(index: number): Promise<string> {
```
jbocce: See my comment in the spec about possibly removing this metody. In general Playwright strongly suggests to NOT assert and/or get attributes, text or data of any kind... https://playwright.dev/docs/api/class-locator#locator-get-attribute

tests/DataOverlayMenuFrameOfReference.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  const segmentationLabelsFOR2 = ['Segmentation FOR2SEG', 'FoR 2 RTstructRTSTRUCT'];
+
+  const dataOverlay = viewportPageObject.getById(viewportId).overlayMenu.dataOverlay;
+
+  await dataOverlay.toggle(viewportId);
+  await dataOverlay.openAddSegmentationDropdown(viewportId);
+  await page.waitForTimeout(2000);
+
+  const segmentationLabels = await dataOverlay.getDropdownOptionLabels();
+  expect(segmentationLabels.sort()).toEqual([...segmentationLabelsFOR1].sort());
```
jbocce, line 22: Are we sorting these because at present there is a bug whereby the labels might be in different order?

tests/pages/DataOverlayPageObject.ts
```diff
@@ -26,4 +29,25 @@ export class DataOverlayPageObject {
     await this.page.getByTestId(`overlay-ds-more-button-${segmentation}`).click();
     await this.page.getByTestId(`overlay-ds-remove-button-${segmentation}`).click();
   }
+
+  async openAddSegmentationDropdown(viewportId: string = 'default') {
+    await this.page.getByTestId(`AddSegmentationDataOverlay-${viewportId}`).click();
+    await this.page.getByText('SELECT A SEGMENTATION').click();
+  }
+
+  async getDropdownOptionLabels(): Promise<string[]> {
```
jbocce, line 44: There are two concerns I have about this method:
1. It goes against Playwright's principle to not get data and assert it. So instead this should be replaced by a utility that asserts the labels
2. It assumes that the drop down menu is open. Instead whether we keep this method or replace it with a utility, this function/method itself should open and close the menu to get and/or assert its data.

tests/DataOverlayMenuFrameOfReference.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+}) => {
+  const viewportId = await viewportPageObject.getIdOfNth(0);
+  const segmentationLabelsFOR1 = ['Segmentation FOR1SEG', 'FoR 1 RTstructRTSTRUCT'];
+  const segmentationLabelsFOR2 = ['Segmentation FOR2SEG', 'FoR 2 RTstructRTSTRUCT'];
+
+  const dataOverlay = viewportPageObject.getById(viewportId).overlayMenu.dataOverlay;
+
+  await dataOverlay.toggle(viewportId);
+  await dataOverlay.openAddSegmentationDropdown(viewportId);
+  await page.waitForTimeout(2000);
```
jbocce: No need to wait, especially if we replace getDropdownOptionLabels to be a utility that does an expect which will do the waiting for us.

tests/DataOverlayMenuFrameOfReference.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const segmentationLabels = await dataOverlay.getDropdownOptionLabels();
+  expect(segmentationLabels.sort()).toEqual([...segmentationLabelsFOR1].sort());
+  await dataOverlay.closeMenu(viewportId);
+
+  await dataOverlay.toggle(viewportId);
+  await dataOverlay.changeBackgroundDisplaySet('CT Std (FoR 2)-CT', viewportId);
+  await dataOverlay.openAddSegmentationDropdown(viewportId);
+  await page.waitForTimeout(2000);
```
jbocce: See comment above about waiting.

tests/DataOverlayMenuFrameOfReference.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const dataOverlay = viewportPageObject.getById(FOR1ViewportIdA).overlayMenu.dataOverlay;
+
+  // In FOR1 viewport A: add a segmentation FOR1 overlay
+  await viewportPageObject.getById(FOR1ViewportIdA).pane.hover();
+  await dataOverlay.toggle(FOR1ViewportIdA);
+  await dataOverlay.addSegmentation(segmentationFOR1Label, FOR1ViewportIdA);
+  await expect(dataOverlay.getOverlaySegmentationRow(segmentationFOR1Label)).toBeVisible();
+  await dataOverlay.closeMenu(FOR1ViewportIdA);
+  await page.waitForTimeout(2000);
```
jbocce: No wait.

tests/pages/DataOverlayPageObject.ts
```diff
@@ -11,9 +11,12 @@ export class DataOverlayPageObject {
     await this.page.getByTestId(`dataOverlayMenu-${viewportId}-btn`).click();
   }
 
+  async closeMenu(viewportId: string = 'default') {
+    await this.toggle(viewportId);
```
jbocce: I think it is also best to wait for the menu to no longer be visible before returning. Again these are the principles of GUI testing.

extensions/cornerstone/src/services/SyncGroupService/createHydrateSegmentationSynchronizer.ts
```diff
@@ -70,6 +70,14 @@ const segmentationRepresentationModifiedCallback = async (
     return;
   }
```
jbocce: @wayfarer3130 I think these too make sense, but I just wanted another pair of eyes on these changes.

extensions/cornerstone/src/services/SyncGroupService/createHydrateSegmentationSynchronizer.ts
```diff
@@ -70,6 +70,14 @@ const segmentationRepresentationModifiedCallback = async (
...
 
+  const targetFrameOfReferenceUID = viewport.getFrameOfReferenceUID();
```
wayfarer3130, line 69: This line should be above the if condition up above so you get the FOR once.

extensions/cornerstone/src/services/SyncGroupService/createHydrateSegmentationSynchronizer.ts
```diff
@@ -70,6 +70,14 @@ const segmentationRepresentationModifiedCallback = async (
...
+  const sourceFrameOfReferenceUID =
+    getEnabledElementByViewportId(sourceViewportId)?.viewport?.getFrameOfReferenceUID();
+
+  if (!sharedDisplaySetExists && targetFrameOfReferenceUID !== sourceFrameOfReferenceUID) {
```
wayfarer3130, line 77: I think I would add a comment inside
// Display sets can be overlayed if they apply to the same base display set, eg for instance frame references in a seg to a display set,
// OR they can be overlayed if they share a common FOR (eg different image display and segmentation applies to multiple FORs.)

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -870,16 +870,16 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
       initialImageIndexToUse = this._getInitialImageIndexForViewport(viewportInfo, imageIds) || 0;
     }
 
-    return viewport.setStack(imageIds, initialImageIndexToUse).then(() => {
+    return viewport.setStack(imageIds, initialImageIndexToUse).then(async () => {
```
wayfarer3130: This is already an async function, would prefer:
await viewport.setStack(....)
and then just leave the internal processing to handle afterwards rather than mixing async functions and .then usage.
I know it was previously related, but lets make it consistent now.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -1122,11 +1122,11 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
     await viewport.setVolumes(volumeInputArray);
 
     if (overlayProcessingResults?.length) {
```
wayfarer3130: Please add the overlay processing results to a single method since it looks like it gets repeated.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -1203,18 +1203,23 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
+          blendMode:
+            viewport?.getBlendMode?.() === 1
+              ? BlendModes.LABELMAP_EDGE_PROJECTION_BLEND
+              : undefined,
+        },
+      }
+    );
     // store the segmentation presentation id in the viewport info
     this.storePresentation({ viewportId: viewport.id });
+    return segmentationRepresentationPromise;
```
wayfarer3130, line 1208: Please update the return type for this method to be Promise<unknown> or void or something that indicates it returns a promise.

platform/core/src/types/DisplaySet.ts
```diff
@@ -57,6 +57,8 @@ export type DisplaySet = {
    */
   referencedDisplaySetInstanceUID?: string;
 
+  FrameOfReferenceUID?: string;
```
wayfarer3130: Since this is a computed field at least some of the time, not a display field, I'd prefer to try to move towards lowerCamelCase.  
In fact I'd like to do that for SeriesDate/SeriesTime etc as well, and might work with that for OHIF in the next release.

wayfarer3130: Also, you need a comment that this is the FOR that applies for every frame within this display set, and that it will be undefined otherwise.

extensions/cornerstone/src/services/SyncGroupService/createHydrateSegmentationSynchronizer.ts
GhadeerAlbattarni: Thank you @wayfarer3130  for your review. I noticed that at the top of this method, it comments what this is doing and it reflects your comment here. So I didn't add it inside again, however if you think I should, I would do that.

GhadeerAlbattarni:
Thank you so much @wayfarer3130 and @jbocce for reviewing this PR. I made the requested changes

extensions/cornerstone/src/services/SyncGroupService/createHydrateSegmentationSynchronizer.ts
jbocce: Yes there is a comment at the top of the method, so I think it is fine as is.

extensions/cornerstone-dicom-rt/src/getSopClassHandlerModule.ts
```diff
@@ -56,6 +56,7 @@ function _getDisplaySetsFromSeries(
     StudyInstanceUID,
     SOPClassHandlerId,
     SOPClassUID,
+    frameOfReferenceUID: null,
```
sedghi, line 59: why are we using frameOfReferenceUID instead of FrameOfReferenceUID which is what is in dcmjs and matches the rest of SeriesDate,SeriesTime, etc. Anything dicom should follow same notation

@jbocce @wayfarer3130

extensions/cornerstone/src/components/ViewportDataOverlaySettingMenu/utils.ts
```diff
@@ -80,8 +80,8 @@ export function getEnhancedDisplaySets({ viewportId, services }) {
 
     // Check if Frame of Reference matches
     if (
-      displaySet.FrameOfReferenceUID &&
-      displaySet.FrameOfReferenceUID !== backgroundDisplaySet.FrameOfReferenceUID
+      displaySet.frameOfReferenceUID &&
```
sedghi, line 83: seems like this is breaking change and is not documented

Issue #5617, opened 2025-12-05
[Bug] Freehand/Spline/Livewire tools cannot start a new contour after Backspace removes all points


masatoh-array:
### Describe the Bug

When using annotation tools such as Freehand ROI, Spline ROI, or Livewire, deleting points during drawing using Backspace leaves the tool in an inconsistent internal state.
After all intermediate points are removed, the tool does not exit the drawing mode.
As a result, clicking on another location does not start a new contour.
The tool only recovers after performing a double-click, which forcibly finalizes the previous drawing state.

This behavior appears to be unintended and differs from expected tool state transitions.

### Steps to Reproduce

1. Activate Freehand ROI, Spline ROI, or Livewire tool.
2. Start drawing a contour (add 1 or more points).
3. Press Backspace repeatedly until all points are removed.
4. Click in a different location to begin a new contour.

### The current behavior

Nothing happens. The tool remains stuck in a “drawing” mode even though the contour has no points.

### The expected behavior

When all points are removed with Backspace, the tool should:
- Exit the current drawing session
- Return to a neutral/tool-ready state
- Allow a new contour to begin on the next click

### System Information

OHIF Viewer:
- OHIF Viewer: viewer.ohif.org demo v3.11.1
- Commit hash: 9d94e26d5d15ccd19f12020dbc5438364d88d253

System:
- OS: Windows 11

Browser: 
- Microsoft Edge 142.0.3595.94 （Official Build）（64 bit）

PR #5905, opened 2026-03-17
fix(measurement): Restore viewport interactivity when deleting in-progress Spline or Livewire measurement

### LINKED ISSUE #5673: [Bug] Viewport interactivity is lost after deleting Spline or Livewire tool with right click menu while drawing it
### Describe the Bug

If user right click and select Delete Measurement from context menu while drawing annotation with either the Spline or the Livewire tool, they will no longer be able to complete the annotation and use any other tools including the active annotation. <br>User have to refresh the viewer to regain the interactivity. Even dropping a new series to viewport or switching to a new layout will not fix this behavior.

### Steps to Reproduce

1. Start drawing a Spline or Livewire tool
2. Right click and select Delete Measurement before completing the measurement
3. Try to draw again or switch to any other annotation tool or zoom/pan/windowing etc. tool.

### The current behavior

The user interactivity is lost. Users cannot do anything in viewport with the tools.

### The expected behavior

After deleting, the user should be able to start drawing a new annotation with selected tool. Or they should be able to switch to any other available tool to continue working with the viewer.

### System Information

**OHIF Version:**<br>Data: [https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125113417.1](<https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125113417.1>) <br>OHIF Viewer: 3.12.0 beta.112<br>Commit Hash: cd253f20bde68f33a01ac1fe5e1cf6b987d9d5e2<br>Current Browser & OS: Chrome 143.0.0, Linux x86_64

**System:**<br>    OS: Linux 6.11 Ubuntu 24.10 24.10 (Oracular Oriole)<br>    CPU: (12) x64 13th Gen Intel(R) Core(TM) i5-1335U<br>    Memory: 8.66 GB / 15.31 GB<br>    Container: Yes<br>    Shell: 5.2.32 - /bin/bash<br>  **Binaries:**<br>    Node: 24.6.0 - /home/username/.nvm/versions/node/v24.6.0/bin/node<br>    Yarn: 1.22.22 - /home/username/.nvm/versions/node/v24.6.0/bin/yarn<br>    npm: 11.5.1 - /home/username/.nvm/versions/node/v24.6.0/bin/npm<br>    bun: 1.3.2 - /home/username/.nvm/versions/node/v24.6.0/bin/bun<br>  **Browsers:**<br>    Chrome: 143.0.7499.109<br>    Firefox: 146.0.1<br>    Firefox Developer Edition: 146.0.1

GhadeerAlbattarni:
### Context

Fixes #5673, #5617

This PR fixes loss of viewport interactivity when a user deletes an in-progress Spline or Livewire measurement from the context menu.

When the user starts drawing with Spline or Livewire tool, then attempt to delete the measurement before it's completed (whether through backspace or right click), the viewport becomes non-interactive, the user cannot draw again, switch tools, until they double-click on the image or refresh the viewer.

### Root cause
Deleting the measurement only removed the annotation from annotation state. The Spline/Livewire tool was never told to cancel, so it kept its in-progress state (`isDrawing`). The tool stayed in “active manipulation” mode and continued to block other interaction with the viewport.
 

### Changes & Results
- `initMeasurementService.ts`
In the `MEASUREMENT_REMOVED` subscription, get the active viewport element (same pattern as `MEASUREMENT_UPDATED`) and call `cancelActiveManipulations(element)` before `removeAnnotation`.

**Before**
After deleting a non completed measurement, the viewport becomes interactive and can't use any tools. 

 **After**
 The viewport stay active after deleting a measurement
 
 

https://github.com/user-attachments/assets/9c32a2f0-7e13-4cc1-916d-e0049c4aeb30

 

### Testing
1. Open a study in the viewer mode.
2. Select Spline or Livewire tool.
2. Start drawing (place one or more points).
3. Right-click on the viewport and choose 'Delete measurement' or backspace
4. Confirm the measurement is removed and the viewport stays interactive: you can immediately draw a new measurement, switch tools, etc. 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.13.0

<h3>Greptile Summary</h3>

This PR fixes a viewport interactivity bug where deleting an in-progress Spline or Livewire annotation (via the context menu or backspace) left the tool in its `isDrawing` state, blocking all further interaction until the user double-clicked or refreshed. The root cause was that `MEASUREMENT_REMOVED` only removed the annotation from state without notifying the active tool to cancel its manipulation.

**Key changes:**
- `initMeasurementService.ts`: Imports `cancelActiveManipulations` from `@cornerstonejs/tools` and calls it on the active viewport element at the start of the `MEASUREMENT_REMOVED` handler — following the same `getActiveViewportEnabledElement` pattern already used in the `MEASUREMENT_UPDATED` handler. The fix also uses `?.viewport?.element` (with extra optional chaining on `viewport`) which is slightly safer than the existing `?.viewport.element` usage on line 408.
- `tests/Livewire.spec.ts` & `tests/Spline.spec.ts`: Two new E2E tests each verify that after deleting an in-progress annotation via the context menu, the viewport remains interactive and a new annotation can be successfully drawn and confirmed.

**Minor concerns:**
- `cancelActiveManipulations` is called for **every** `MEASUREMENT_REMOVED` event, not only when the annotation is in-progress. This could interrupt an active drawing session if a separate completed annotation is deleted concurrently (e.g., from the measurements side-panel).
- The `Livewire.spec.ts` test includes an unused `rightPanelPageObject` parameter that should be removed.

- Safe to merge with minor cleanup; the core fix is correct and well-targeted.
- The fix correctly resolves the reported bug using an established pattern already present in the file, and is backed by two new E2E tests. A small logic concern exists around calling `cancelActiveManipulations` unconditionally for all removals (not just in-progress ones), and there is an unused parameter in the Livewire test. Neither issue is blocking, but the unconditional cancellation could introduce subtle regressions in edge cases.
- `extensions/cornerstone/src/initMeasurementService.ts` — review the unconditional `cancelActiveManipulations` call to ensure it doesn't affect scenarios where a completed annotation is deleted while drawing is active.

| Filename | Overview |
|----------|----------|
| extensions/cornerstone/src/initMeasurementService.ts | Adds `cancelActiveManipulations` call in the `MEASUREMENT_REMOVED` handler to restore viewport interactivity after deleting an in-progress annotation. The fix is correct for the reported bug but is applied unconditionally to all removals, which could inadvertently cancel an active drawing if a different annotation is removed concurrently. |
| tests/Livewire.spec.ts | Adds a new E2E test covering the delete-in-progress-annotation scenario for Livewire. The test logic is correct and mirrors the Spline test, but the function signature includes an unused `rightPanelPageObject` parameter that should be removed. |
| tests/Spline.spec.ts | Adds a new E2E test covering the delete-in-progress-annotation scenario for Spline. Clean implementation with no issues found. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Viewport
    participant ContextMenu
    participant MeasurementService
    participant initMeasurementService
    participant CornerstoneTools

    User->>Viewport: Start drawing (Spline/Livewire)
    Viewport->>CornerstoneTools: Tool enters isDrawing=true state
    User->>ContextMenu: Right-click → Delete measurement
    ContextMenu->>MeasurementService: remove(measurementId)
    MeasurementService->>initMeasurementService: MEASUREMENT_REMOVED event
    Note over initMeasurementService: NEW: cancelActiveManipulations(element)
    initMeasurementService->>CornerstoneTools: cancelActiveManipulations(element)
    CornerstoneTools->>Viewport: Reset isDrawing=false, restore interactivity
    initMeasurementService->>CornerstoneTools: removeAnnotation(measurementId)
    Viewport-->>User: Viewport is interactive again
```

<sub>Last reviewed commit: c076066</sub>

> Greptile also left **2 inline comments** on this PR.

### DIFF
--- a/tests/Livewire.spec.ts
+++ b/tests/Livewire.spec.ts
@@ -75,6 +75,7 @@ test('should restore viewport interactivity after deleting an in-progress Livewi
     { x: 523, y: 232 },
   ]);
 
+  // Ensure the three points clicked above are rendered in the DOM before pressing Backspace
   await expect(activeViewport.svg('circle')).toHaveCount(3);
   await press({ page, key: 'Backspace' });
 
--- a/tests/Spline.spec.ts
+++ b/tests/Spline.spec.ts
@@ -77,6 +77,7 @@ test('should restore viewport interactivity after deleting an in-progress Spline
     { x: 523, y: 232 },
   ]);
 
+  // Ensure the three points clicked above are rendered in the DOM before pressing Backspace
   await expect(activeViewport.svg('circle')).toHaveCount(3);
   await press({ page, key: 'Backspace' });
 


extensions/cornerstone/src/initMeasurementService.ts
```diff
@@ -504,6 +504,12 @@ const connectMeasurementServiceToTools = ({
       if (source?.name && source.name !== CORNERSTONE_3D_TOOLS_SOURCE_NAME) {
         return;
       }
+      // Cancel any active tool manipulation (e.g., Spline/Livewire) to avoid leaving the tool
+      // in a drawing state after deleting a not completed measurement, which can block viewport interactivity.
+      const element = getActiveViewportEnabledElement(viewportGridService)?.viewport?.element;
+      if (element) {
+        cancelActiveManipulations(element);
+      }
```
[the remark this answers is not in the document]
GhadeerAlbattarni: Hello @jbocce. I think greptile has a point here. However, I waited to hear your opinion whenever you have time :)

tests/Livewire.spec.ts
```diff
@@ -24,3 +24,38 @@ test('should display the livewire tool', async ({
...
+}) => {
+  await mainToolbarPageObject.measurementTools.livewireContour.click();
+  await viewportPageObject.active.clickAt([
+    { x: 380, y: 299 },
+    { x: 420, y: 236 },
+    { x: 523, y: 232 },
+  ]);
+
+  await viewportPageObject.active.clickAt([{ x: 550, y: 312 }], 'right');
+  await page.waitForTimeout(500);
```
jbocce: I don't think we need a wait here because the beauty of playwright is that a lot of waits are baked in. So in this case line 44 has implicit waits for the context menu and the delete menu item to be visible. Furthermore, it means that we do not need the wait on line 45 either because the click itself does a wait too I believe. Give it a try! 😃

tests/Livewire.spec.ts
jbocce, line 1: Do you think we can manage a test with the backspace key?

tests/Spline.spec.ts
```diff
@@ -24,3 +24,39 @@ test('should display the spline tool', async ({
...
+  await mainToolbarPageObject.measurementTools.splineROI.click();
+
+  await viewportPageObject.active.clickAt([
+    { x: 380, y: 299 },
+    { x: 420, y: 236 },
+    { x: 523, y: 232 },
+  ]);
+
+  await viewportPageObject.active.clickAt([{ x: 550, y: 312 }], 'right');
+  await page.waitForTimeout(500);
```
jbocce: See my comment in the livewire spec.

extensions/cornerstone/src/initMeasurementService.ts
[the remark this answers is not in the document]
jbocce: I suppose greptile has a point, but its suggestion is too specific and goes beyond what is currently exposed in CS3D's public API. And I would say it is unlikely that someone would delete a completely different measurement while adding/editing another. And if they do, one could argue that the other manipulation should be cancelled. Just my two cents.

tests/Livewire.spec.ts
```diff
@@ -59,3 +58,34 @@ test('should restore viewport interactivity after deleting an in-progress Livewi
...
+  viewportPageObject,
+}) => {
+  await mainToolbarPageObject.measurementTools.livewireContour.click();
+  await viewportPageObject.active.clickAt([
+    { x: 380, y: 299 },
+    { x: 420, y: 236 },
+    { x: 523, y: 232 },
+  ]);
+
+  await page.waitForTimeout(500);
```
GhadeerAlbattarni: Actually, we do need to wait here, otherwise the test will fail.

tests/Spline.spec.ts
```diff
@@ -60,3 +59,35 @@ test('should restore viewport interactivity after deleting an in-progress Spline
...
+}) => {
+  await mainToolbarPageObject.measurementTools.splineROI.click();
+
+  await viewportPageObject.active.clickAt([
+    { x: 380, y: 299 },
+    { x: 420, y: 236 },
+    { x: 523, y: 232 },
+  ]);
+
+  await page.waitForTimeout(500);
```
GhadeerAlbattarni: Similar here, we need to wait otherwise the test will fail

tests/Livewire.spec.ts
GhadeerAlbattarni: You're right no need to wait here, the test will pass

jbocce:
Please update your branch here.

tests/Livewire.spec.ts
```diff
@@ -25,3 +26,69 @@ test('should display the livewire tool', async ({
...
+}) => {
+  await mainToolbarPageObject.measurementTools.livewireContour.click();
+  const activeViewport = await viewportPageObject.active;
+  await activeViewport.clickAt([
+    { x: 380, y: 299 },
+    { x: 420, y: 236 },
+    { x: 523, y: 232 },
+  ]);
+
+  await expect(activeViewport.svg('circle')).toHaveCount(3);
```
jbocce, line 79: Probably worth putting a comment above this line stating that this waits for the three points clicked above to actually be in the DOM.

tests/Spline.spec.ts
```diff
@@ -25,3 +26,71 @@ test('should display the spline tool', async ({
...
+  await mainToolbarPageObject.measurementTools.splineROI.click();
+
+  const activeViewport = await viewportPageObject.active;
+  await activeViewport.clickAt([
+    { x: 380, y: 299 },
+    { x: 420, y: 236 },
+    { x: 523, y: 232 },
+  ]);
+
+  await expect(activeViewport.svg('circle')).toHaveCount(3);
```
jbocce, line 81: Comment above here too.

GhadeerAlbattarni:
Thanks @jbocce for reviewing this PR. I updated the branch and added a comment. 

It seems that Playwrite tests got canceled here too.

PR #5919, opened 2026-03-25
fix: prevent black viewport when navigating series with client-created segmentation

### LINKED ISSUE #5888: [Bug] Viewport crashes when using Page Down hotkey to navigate between series in multi-series studies with a new segmentation
### Describe the Bug

When navigating between series using the Page Down hotkey in a study containing multiple series, the viewport crashes if a new segmentation has been created.

https://github.com/user-attachments/assets/ca1c07a1-88e1-4ad0-bbd4-1e17a413ef36

### Steps to Reproduce

- Launch a study containing multiple image series.
- Open any series in the viewport.
- Create a new segmentation.
- After segmentation is added, press Page Down to navigate to the next series.
- Observe the viewport behavior.

### The current behavior

The viewport crashes when Page Down is pressed after creating a new segmentation.

### The expected behavior

The viewport should smoothly navigate to the previous series using the Page Down hotkey, even after a new segmentation is created.

### System Information

System:
OS: Linux 6.6 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
Memory: 4.29 GB / 7.63 GB
Container: Yes
Shell: 5.1.16 - /bin/bash
Binaries:
Node: 22.19.0 - ~/.nvm/versions/node/v22.19.0/bin/node
Yarn: 1.22.22 - ~/.nvm/versions/node/v22.19.0/bin/yarn
npm: 11.6.0 - ~/.nvm/versions/node/v22.19.0/bin/npm

GhadeerAlbattarni:
### Context

Fixes #5888 

When a user opens a study with multiple series, adds a segmentation, and then keeps using `Page Down` to navigate between series, the viewport goes black once reaches the added segmentation displayset.

**Root cause**
When a segmentation is added via "Add Segmentation", a client-created display set is pushed into `activeDisplaySets`.

The Page Down hotkey triggers `updateViewportDisplaySet`, which iterates through `activeDisplaySets` to determine the next display set to render.

Since these client-created segmentation display sets are not filtered out, the command attempts to render them as standalone viewport display sets, causing the viewport to go black.

### Changes & Results
Skip display sets marked with `madeInClient` when selecting the next display set to show in `updateViewportDisplaySet` during Page Down/Up navigation 

**Before**
After adding a segmentation, repeated `Page Down` navigation could leave the viewport blank 

**After**
`Page Down` now skips client-created segmentation display sets and keeps the viewport on a valid series.

### Testing
1. open a study with multiple series in a `segmentation` mode (e.g., StudyInstanceUIDs=1.2.840.113619.2.290.3.3767434740.226.1600859119.501 )
2. click `Add segmentation`
3. Double click on the default viewport
4. Press `Page Down` repeatedly to navigate across all display sets 
5. Verify the viewport continues to show valid series instead of turning black once it reaches the last display set in the left panel. 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.13.0

<h3>Greptile Summary</h3>

This PR fixes a black viewport issue that occurred when pressing Page Down to navigate series after adding a client-created segmentation in OHIF. The root cause was that `updateViewportDisplaySet` iterated over `activeDisplaySets` without filtering out display sets marked `madeInClient = true`, so the viewport tried to render a segmentation-only display set as a standalone image, resulting in a blank screen.

**Key changes:**
- `extensions/default/src/commandsModule.ts`: Inside the `updateViewportDisplaySet` navigation loop, a `continue` guard is added before the existing modality filter. If `nextDisplaySet.madeInClient` is `true`, the loop immediately moves to the next candidate instead of breaking on it. This is consistent with how `madeInClient` is guarded elsewhere (e.g., `setUpSegmentationEventHandlers.ts` line 84).
- `tests/SegmentationSeriesNavigation.spec.ts`: New Playwright E2E test that adds a segmentation to a known study, double-clicks into the viewport, presses PageDown 6 times, and asserts the instance-number overlay remains visible after each press — verifying the viewport never goes black.

- Safe to merge; the logic change is minimal, targeted, and consistent with existing `madeInClient` guards across the codebase.
- The one-line `continue` guard is a small, well-scoped change with no regression risk for normal series navigation. The only non-blocking concern is the hardcoded `totalPageDownPresses = 6` in the test, which may not definitively exercise the boundary condition depending on the study's series count.
- No files require special attention.

| Filename | Overview |
|----------|----------|
| extensions/default/src/commandsModule.ts | Adds a `madeInClient` guard in the `updateViewportDisplaySet` loop to skip client-created segmentation display sets during Page Down/Up navigation, preventing black viewports. |
| tests/SegmentationSeriesNavigation.spec.ts | New E2E test that adds a segmentation and verifies the viewport overlay remains visible across 6 PageDown presses; the hardcoded count of 6 may not reliably trigger the boundary condition on all study configurations. |

</details>

</details>

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[User presses Page Down] --> B[updateViewportDisplaySet called]
    B --> C[Get activeDisplaySets from DisplaySetService]
    C --> D[Find activeDisplaySetIndex of current viewport]
    D --> E[Loop: next index = activeDisplaySetIndex + direction]
    E --> F{nextDisplaySet.madeInClient?}
    F -- Yes --> G[continue — skip this display set]
    G --> E
    F -- No --> H{excludeNonImageModalities AND modality is non-image?}
    H -- Yes: skip --> E
    H -- No: use it --> I[break — found valid display set]
    E --> J{Index out of bounds?}
    J -- Yes --> K[return — no navigation]
    I --> L[getViewportsRequireUpdate]
    L --> M[setDisplaySetsForViewports]
    M --> N[Viewport renders valid display set]
```

<sub>Reviews (1): Last reviewed commit: ["test(e2e): add a test for Page Down seri..."](https://github.com/ohif/viewers/commit/cb0fafc9eee0121057e28a6dbb5512139f9e59df) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26308269)</sub>

### DIFF
--- a/tests/SegmentationSeriesNavigation.spec.ts
+++ b/tests/SegmentationSeriesNavigation.spec.ts
@@ -15,7 +15,8 @@ test('should keep the viewport rendered when navigating series with Page Down af
 }) => {
   // Study has 5 series; pressing 8 times verifies navigation remains stable even after
   // reaching the client-created SEG display set appended by "Add Segmentation".
-  const totalPageDownPresses = 8;
+  const minimumSeriesExpected = 5;
+  const totalPageDownPresses = minimumSeriesExpected + 3;
 
   await rightPanelPageObject.labelMapSegmentationPanel.addSegmentationButton.click();
 
@@ -24,7 +25,7 @@ test('should keep the viewport rendered when navigating series with Page Down af
     .poll(async () => {
       return await thumbnailsLocator.count();
     })
-    .toBeGreaterThanOrEqual(5);
+    .toBeGreaterThanOrEqual(minimumSeriesExpected);
 
   const activeViewport = await viewportPageObject.active;
   await activeViewport.pane.click();


GhadeerAlbattarni:
Hello @jbocce. For the test I didn't use the screenshots approach as I noticed in the test environments, the series in the left panel render in a different order. So when navigating through them, we might not always be on the same first display set which makes the test fails. 

I went with asserting the viewport image/instance number. 

I also removed the fix, and the run the test. The test failed as expected when reached the black screen.

extensions/default/src/commandsModule.ts
```diff
@@ -734,10 +734,11 @@ const commandsModule = ({
         displaySetIndexToShow > -1 && displaySetIndexToShow < currentDisplaySets.length;
         displaySetIndexToShow += direction
       ) {
-        if (
-          !excludeNonImageModalities ||
-          !nonImageModalities.includes(currentDisplaySets[displaySetIndexToShow].Modality)
-        ) {
+        const nextDisplaySet = currentDisplaySets[displaySetIndexToShow];
```
jbocce, line 737: This looks good, but we should test this in the scenario right after saving a segmentation to say a local orthanc instance. In that scenario, an actual display set is created, so the page down/page up should work and navigate to that new display set. Be sure not to refresh the browser after saving the segmentation.

tests/SegmentationSeriesNavigation.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  const totalPageDownPresses = 6;
+
+  await rightPanelPageObject.labelMapSegmentationPanel.addSegmentationButton.click();
+
+  await viewportPageObject.active.pane.click();
+  await viewportPageObject.active.pane.dblclick();
+
+  for (let i = 0; i < totalPageDownPresses; i++) {
+    await press({ page, key: 'PageDown' });
+
```
jbocce: I hate to say it, but I think we need a wait timeout here. 🤣

tests/SegmentationSeriesNavigation.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  const mode = 'segmentation';
+  await visitStudy(page, studyInstanceUID, mode, 2000);
+});
+
+test('should keep the viewport rendered when navigating series with Page Down after adding a segmentation', async ({
+  page,
+  rightPanelPageObject,
+  viewportPageObject,
+}) => {
+  const totalPageDownPresses = 6;
```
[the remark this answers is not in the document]
jbocce: This is a good point. One possibility might be to add an expect the number of thumbnails in the side panel to be at least 6. To stick with the Playwright principle of using waits for assertions, we should do something like this...

```
await expect.poll(async () => {
    return await thumbnailsLocator.count();
  }).toBeGreaterThanOrEqual(6);
```

https://playwright.dev/docs/test-assertions#expectpoll

tests/SegmentationSeriesNavigation.spec.ts
jbocce: Actually, no, I am wrong. With what you have the expect does the wait! Good stuff!

tests/SegmentationSeriesNavigation.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+test('should keep the viewport rendered when navigating series with Page Down after adding a segmentation', async ({
+  page,
+  leftPanelPageObject,
+  rightPanelPageObject,
+  viewportPageObject,
+}) => {
+  // Study has 5 series; pressing 8 times verifies navigation remains stable even after
+  // reaching the client-created SEG display set appended by "Add Segmentation".
+  const totalPageDownPresses = 8;
```
jbocce: To keep the number of series tied to this consider something like this...
```
const minimumSeriesExpected = 5;
const totalPageDownPresses = minimumSeriesExpected + 3;
```
And then on line 27 you can use the `minimumSeriesExpected` variable too.

jbocce (changes requested):
One minor comment left...

tests/SegmentationSeriesNavigation.spec.ts
GhadeerAlbattarni: Thanks for pointing this :)

PR #5965, opened 2026-04-16
fix(MetadataProvider): Correctly assign imageId for multiframe images and remove unused frame information retrieval method

sedghi:
### Context

add a console log after 

const { PixelSpacing, type } = getPixelSpacingInformation(instance) || {};

in metadataprovider 

see imageId in multiframe case is wrong even if you scroll always show the first one 

try this case 

https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.2723277605.10168.1676054414.178

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

<h3>Greptile Summary</h3>

This PR fixes a bug where all frames of a multiframe DICOM image always reported the first frame's `imageId`. The root cause is that `DicomMetadataStore` keeps only one instance per SOPInstanceUID; `combineFrameInstance` returns a new per-frame object but that object inherits `imageId` from the parent, which always pointed to frame 1. The fix explicitly stamps `result.imageId = imageId` after `combineFrameInstance` returns. It also removes the `getFrameInformationFromURL` helper (no external callers remain) and replaces URL-scraping with reading `uids.frameNumber` stored at ingestion time, while adding an explicit null-guard that was missing in the previous code.

Safe to merge — the fix is narrowly scoped, no regressions found across all existing data source call sites

No P0/P1 issues found. The imageId stamp is correct: for multiframe images combineFrameInstance returns a new per-frame Object.create() object so only that frame-specific cached instance is mutated, never the shared DicomMetadataStore instance. The new uids.frameNumber fallback is equivalent to the old URL-scraping for every current data source. The removed getFrameInformationFromURL has zero callers remaining in the repo.

No files require special attention

| Filename | Overview |
|----------|----------|
| platform/core/src/classes/MetadataProvider.ts | Two targeted fixes: reassigns imageId on the combined frame instance to unblock per-frame metadata look-ups, and hardens getUIDsFromImageID with an explicit null guard while replacing URL-scraping with a stored uids.frameNumber read |

</details>

</details>

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `platform/core/src/services/DicomMetadataStore/DicomMetadataStore.ts`, line 323 ([link](https://github.com/ohif/viewers/blob/0de4c55d6a9df8588bd2d91b0e312f0df106cfcf/platform/core/src/services/DicomMetadataStore/DicomMetadataStore.ts#L323)) 

   <a href="#">P0</a> <a href="#">security</a> **Accidental debug code exposes internal state globally**

   `window._model=_model` is leftover debug code that leaks the internal `_model` object (which holds the full DICOM studies/series/instances tree) onto the global `window` object. This bypasses encapsulation, is a TypeScript type error (no semicolon, implicit `any` on `window`), and must not ship.
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(MetadataProvider): Add check for und..."](https://github.com/ohif/viewers/commit/811ac6020b37f055ab5e9e1cf4a80412926c5f94) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28688207)</sub>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed multiframe metadata so combined results preserve the originally requested image identifier.
  * Improved frame-aware UID resolution by correctly interpreting `frame`/`frameNumber` for query-based URIs and preferring exact frame-specific matches.
  * Corrected dynamic-volume display so timepoint and thumbnail selection use the proper frame-specific image identifiers.
* **Tests**
  * Added Jest coverage for frame-aware UID resolution (including exact frame matches) and for dynamic-volume handling with frame-specific image IDs.


platform/core/src/classes/MetadataProvider.ts
```diff
@@ -9,6 +9,24 @@ import combineFrameInstance from '../utils/combineFrameInstance';
...
 const { calibratedPixelSpacingMetadataProvider, getPixelSpacingInformation } = utilities;
 
+const FRAME_QUERY_PARAM = /[?&]frame=([^&#]*)/;
+const FRAME_QUERY_PARAM_WITH_SEPARATOR = /([?&])frame=[^&#]*(&)?/;
+
+function getFrameNumberFromImageURI(imageURI: string): string | undefined {
+  return imageURI.match(FRAME_QUERY_PARAM)?.[1];
+}
+
+function removeFrameNumberFromImageURI(imageURI: string): string {
```
wayfarer3130: There are existing remove frame number versions already - lets re-use an existing one so that we can have all the removal stuff in one place.

platform/core/src/classes/MetadataProvider.ts
```diff
@@ -9,6 +9,24 @@ import combineFrameInstance from '../utils/combineFrameInstance';
 
 const { calibratedPixelSpacingMetadataProvider, getPixelSpacingInformation } = utilities;
 
+const FRAME_QUERY_PARAM = /[?&]frame=([^&#]*)/;
+const FRAME_QUERY_PARAM_WITH_SEPARATOR = /([?&])frame=[^&#]*(&)?/;
+
+function getFrameNumberFromImageURI(imageURI: string): string | undefined {
```
wayfarer3130: Ditto - there are already calls for this in cs3d metadata.

platform/core/src/classes/MetadataProvider.ts
```diff
@@ -60,7 +78,15 @@ class MetadataProvider {
...
     }
 
-    return (frameNumber && combineFrameInstance(frameNumber, instance)) || instance;
+    const result = (frameNumber && combineFrameInstance(frameNumber, instance)) || instance;
+
+    // We reassign the imageId on the instance because multiframe images processed
+    // through combineFrameInstance will mistakenly get the first imageId.
+    // This happens because the DICOM web data store only keeps the first instance.
+    if (result) {
+      result.imageId = imageId;
```
wayfarer3130: This has to be a non-iterable getter.:
Object.defineProperty(newInstance, 'imageId', {
      value: imageId,
         });
Otherwise if you spread this object into another one, it will preserve the imageId, which is the wrong behaviour for things like segmentation where the object changes.

platform/core/src/classes/MetadataProvider.ts
wayfarer3130: getUriModule(imageId) (line 17) — this is your extractor and remover in one. It returns:

baseImageId → the image id with the frame portion removed (imageId.substring(0, framesMatch.index))
frameNumber → the extracted frame as a number
framesString → the raw frame string
remaining → anything after the frame segment (query params, etc.)

wayfarer3130 (approved):
Just use the shared utility from cs3d for getting the frame number/extracting the base uri:
getUriModule(imageId) (line 17) — this is your extractor and remover in one. It returns:

baseImageId → the image id with the frame portion removed (imageId.substring(0, framesMatch.index))
frameNumber → the extracted frame as a number
framesString → the raw frame string
remaining → anything after the frame segment (query params, etc.)

wayfarer3130 (approved):

PR #5907, opened 2026-03-19
[REPLACED] fix: Misc ohif-downstream integration test failures

wayfarer3130:
Being replaced by individual changes for some of this.

Miscellaneous ohif-downstream integration test failures

No functional changes

### Changes & Results

Update to grid comparison for 3D tests

What are the effects of this change?
- Before vs After
- Screenshots / GIFs / Videos
-->

#### PR

- [] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

<h3>Greptile Summary</h3>

This PR fixes flaky downstream integration test failures in the 3D viewport tests by scoping the screenshot comparison from the full page to only the `viewport-grid` element (via `viewportPageObject.grid`). Updated baseline screenshots are included to match the new comparison area.

- **`tests/3DFourUp.spec.ts` & `tests/3DMain.spec.ts`**: `viewportPageObject` fixture added to the test destructuring; the second argument to `checkForScreenshot` changed from `page` (full-page screenshot) to `viewportPageObject.grid` (`data-testid="viewport-grid"` locator). Test names updated with a "grid compare" suffix to reflect the change.
- **Baseline PNGs updated**: Both `threeDFourUpDisplayedCorrectly.png` and `threeDMainDisplayedCorrectly.png` regenerated to match the narrowed comparison region.
- The PR description states "No functional changes", but this change does meaningfully narrow the screenshot comparison area — it now ignores toolbar/sidebar chrome that may render differently across environments, which is intentional and desirable.
- The `ViewportPageObject.grid` getter correctly returns `this.page.getByTestId('viewport-grid')`, which is a valid `Locator | Page` input for `checkForScreenshot`, so there are no type or runtime concerns.

<h3>Confidence Score: 5/5</h3>

- This PR is safe to merge — all changes are confined to test files and baseline screenshots with no production code impact.
[...]

PR #5967, opened 2026-04-18
fix(seg): prevent segmentations from spreading to all viewports before hydration confirmation in 3D four-up

### LINKED ISSUE #5953: [Bug] MPR viewports don't load correctly in 3d four up spec test
### Describe the Bug

Somehow the screenshots in the 3d four up compared against a single-line image for hte lower two viewports

### Steps to Reproduce

Run SEGHydrationFrom3DFourUp.spec.ts

### The current behavior

Note the garbage display at the bottom
Image

### The expected behavior

Renders the image - REALLY the correct behaviour is to NOT render anything except the SEG load panel since the premise of the load viewports is that you can see something loaded without applying it.  Thus, every window except that load should stay unchanged

### System Information

NA

GhadeerAlbattarni:
### Acknowledgement

This PR is based on the implementation from PR #5907 by @wayfarer3130, so thanks to all his amazing work! 

Also, thanks to @jbocce for all the efforts he made to improve the tests!

### Context

Fixes #5953 

When a segmentation (SEG) series was loaded into one of the MPR viewports in the 3D four-up layout, all viewports rendered the segmentation before user confirm hydration. 

This PR updates the behavior so that:
- SEG loads only in the target viewport initially
- Other viewports remain unchanged
- Segmentations render in all viewports only after user confirmation

### Changes & Results

- `HangingProtocolService.ts`: Return only the active viewport for overlay display sets (SEG/RTSTRUCT) — prevents all viewports being updated with the SEG before user confirmation 
- `CornerstoneViewportService.ts`:  Skip `setVolumes()` when the same base volume is already loaded; add `render()` after `setPresentations()` so segmentation is visible immediately; guard `jumpToSlice` so it doesn't reset MPR position when volume was unchanged 
- `useSegmentationPresentationStore.ts`: Filter overlay UIDs from the segmentation presentation ID to match the key used by `updateStoredSegmentationPresentation` 
- `commandsModule.ts`: Guard `loadSegmentationDisplaySetsForViewport` against an empty `updatedViewports` list and
set viewports needsRender before hydration
- `hydrationUtils.ts`: Add `mergeVolumeSharingViewports` as a safety-net for non-HP layouts: finds all grid viewports that share the same referenced volume and includes them in the post-hydration update 
- `hydrationUtils.test.ts`:  Unit tests for the new `mergeVolumeSharingViewports` behaviour 
- `waitForViewportsRendered.ts`: New test utility with `waitForViewportsRendered` and `waitForAnyViewportNeedsRender` 
- `checkForScreenshot.ts`: Clean up intermediate screenshot artifacts (actual/diff/expected) between retry attempts; increase default retry delay from 500ms to 1250ms
- Update affected tests accordingly

Screenshot 2026-04-17 at 9 49 40 PM

### Testing

1. Open a study that has segmentation in viewer mode (e.g., StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.1706.8374.643249677828306008300337414785)
2. Click on 3D four up
3. Load segmentation display set in one viewport
4. Verify no segmentations show/load into other viewports 
5. Click yes to confirm hydration 
6. Verify segmentations load to the other viewports 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.13.0

<h3>Greptile Summary</h3>

This PR prevents SEG/RTSTRUCT display sets from spreading to all viewports before hydration confirmation in MPR and 3D four-up layouts. The fix spans the HP service (early return for overlays), `CornerstoneViewportService` (skip `setVolumes` when volume unchanged, second `render()` after presentations), the segmentation presentation store (filter overlay UIDs from key), and a new `mergeVolumeSharingViewports` utility that ensures all volume-sharing viewports are updated post-hydration. Tests replace fixed `page.waitForTimeout` calls with event-driven render-cycle helpers.

Safe to merge; one P2 test-reliability concern around waitForViewportRenderCycle used for layout changes.

All production-code changes are well-reasoned and covered by existing + new unit/E2E tests. The only finding is a P2 in the test utility usage that could cause intermittent CI failures on the MPR layout-change step but does not affect production behaviour.

tests/SEGHydrationFromMPR.spec.ts (waitForViewportRenderCycle on layout change), extensions/cornerstone/src/utils/hydrationUtils.ts (pre-existing HP-entry overwrite risk in mergeVolumeSharingViewports).

| Filename | Overview |
|----------|----------|
| platform/core/src/services/HangingProtocolService/HangingProtocolService.ts | Early-returns for overlay display sets (SEG/RTSTRUCT) before protocol matching, preventing pre-hydration spread to all viewports — core fix for the reported issue. |
| extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts | Adds skip logic for setVolumes() when base volume is already loaded, adds a second render() after setPresentations(), and guards jumpToSlice when volume was unchanged. |
| extensions/cornerstone/src/utils/hydrationUtils.ts | Adds mergeVolumeSharingViewports to spread post-hydration updates to all grid viewports sharing the same referenced volume; the grid-scan add() overwrites HP-computed UID lists (flagged in a prior review thread). |
| extensions/cornerstone/src/commandsModule.ts | Guards loadSegmentationDisplaySetsForViewport against empty updatedViewports; calls setNeedsRender synchronously before setDisplaySetsForViewports to enable reliable Playwright polling. |
| extensions/cornerstone/src/stores/useSegmentationPresentationStore.ts | Filters overlay UIDs from the segmentation presentation ID key so it matches the key written by updateStoredSegmentationPresentation, fixing hydrated segmentations silently not applying. |
| tests/utils/waitForViewportsRendered.ts | New test utility with waitForViewportsRendered and waitForViewportRenderCycle; the doc comment explicitly warns that layout changes don't trigger needsRender, but waitForViewportRenderCycle is used in that scenario in SEGHydrationFromMPR. |
| tests/SEGHydrationFromMPR.spec.ts | Replaces fixed timeouts with render-cycle helpers; uses waitForViewportRenderCycle for the post-layout-change screenshot, which may be incompatible with layout changes that don't synchronously set needsRender. |
| tests/utils/checkForScreenshot.ts | Adds between-retry cleanup of Playwright screenshot artifacts and raises default retry delay from 500ms to 1250ms for more stable comparisons. |
| tests/SEGHydrationFrom3DFourUp.spec.ts | Replaces fixed timeouts with render-cycle utilities; sets renderedTimeout to 180 000 ms for the 4-viewport 3D layout, which is appropriate given render complexity. |
| extensions/cornerstone/src/utils/hydrationUtils.test.ts | Adds a well-structured unit test for mergeVolumeSharingViewports covering the 3D four-up scenario. |
| platform/core/src/types/HangingProtocol.ts | Adds the ViewportUpdate interface needed to type the hydrationUtils merge function; viewportId is correctly optional to match existing HP service return shapes. |

</details>

</details>

`````markdown
Path: tests/SEGHydrationFromMPR.spec.ts
Line: 59-63

Comment:
**`waitForViewportRenderCycle` may time out on layout change**

`waitForViewportRenderCycle` first calls `waitForAnyViewportNeedsRender` (5 s default timeout) before it will proceed to wait for 'rendered'. The JSDoc on `waitForViewportsRendered` explicitly warns: _"Examples such as changing the hanging protocol currently don't set such a state and thus can't be rendered without a delay."_ A layout switch to `axialPrimary` recreates viewports through the HP path, so newly-created viewports may never be observed in `needsRender` within that window, causing a 5 s timeout failure in CI.

Consider using `waitForViewportsRendered` directly here, or set a generous `needsRenderTimeout`:

```suggestion
  const viewportRenderAfterLayoutChange = waitForViewportRenderCycle(page, {
    needsRenderTimeout: 15000,
  });
```

`````

</details>

<sub>Reviews (6): Last reviewed commit: ["apply PR feedback 2"](https://github.com/ohif/viewers/commit/d82a9d31e754de342b23b63bef11d91c7f0ac92e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28845229)</sub>

### CODE REVIEW

File: extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts [hunk 1]
```diff
@@ -49,6 +49,45 @@ const EVENTS = {
 const MIN_STACK_VIEWPORTS_TO_ENQUEUE_RESIZE = 12;
 const MIN_VOLUME_VIEWPORTS_TO_ENQUEUE_RESIZE = 6;
 
+function getOrderedVolumeActorReferencedIds(viewport: Types.IVolumeViewport): string[] {
```
jbocce: The name of this function implies ordering yet I see no ordering. Thoughts @wayfarer3130?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -49,6 +49,45 @@ const EVENTS = {
...
+  const actors = viewport.getActors?.() ?? [];
+  return actors
+    .filter(ae => ae.actor?.getClassName?.() === 'vtkVolume')
+    .map(ae => ae.referencedId)
+    .filter(Boolean) as string[];
+}
+
+function volumeIdPrefixesMatch(a: string[], bPrefixLen: number, b: string[]): boolean {
```
jbocce: Let's give better names to a, bPrefixLen and b. Thanks.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -49,6 +49,45 @@ const EVENTS = {
...
+  for (let i = 0; i < bPrefixLen; i++) {
+    if (a[i] !== b[i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/**
+ * Only treat base volumes as identical (skip setVolumes) when the enabled viewport
```
jbocce: I suspect this comment is better suited where this function is called???

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -49,6 +49,45 @@ const EVENTS = {
...
+ * already matches the desired OHIF viewport type. Otherwise stack → MPR (same volumeId)
+ * can incorrectly skip rebuilding the volume viewport.
+ */
+function viewportMatchesDesiredVolumePresentation(
+  viewport: Types.IViewport,
+  viewportInfo: ViewportInfo
```
jbocce: Maybe rename this desiredViewportInfo?

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -1112,7 +1151,42 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
+    const existingVolumeIds = getOrderedVolumeActorReferencedIds(viewport);
+
+    let skippedIdenticalBaseVolumes = false;
+
+    if (baseVolumeInputs.length) {
+      const singleBaseViewport = nextBaseVolumeIds.length === 1;
+
+      if (
+        singleBaseViewport &&
+        nextBaseVolumeIds.length &&
```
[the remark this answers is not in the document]
jbocce: @wayfarer3130 any comment on this? `nextBaseVolumeIds.length` is only checked in this condition so it is a bit odd IMHO.

jbocce:
Please have a look at the failing Playwright tests. Thanks.

extensions/cornerstone/src/utils/hydrationUtils.ts
```diff
@@ -1,11 +1,80 @@
...
+
+  const add = (viewportId: string, uids: string[]) => {
+    if (!viewportId) {
+      return;
+    }
+    byId.set(viewportId, { viewportId, displaySetInstanceUIDs: uids });
+  };
+
+  if (Array.isArray(hangingProtocolUpdates)) {
+    for (const u of hangingProtocolUpdates) {
```
jbocce: rename u to hpUpdate?

tests/utils/waitForViewportsRendered.ts
jbocce, line 1: This is perhaps one of the greatest additions for our automated tests. Thank you for this.

I was thinking that perhaps a third function should also be added that combines the two we have here. Call it say `waitForViewportRenderCycle`. The code might be something like the following. In general this new utility would be the most used and the other two just in case someone wants to do something more specific for whatever reason. Thoughts?

```
type WaitForRenderCycleToCompleteOptions = {
  /**
   * Timeout for waiting until at least one viewport reaches `needsRender`.
   */
  needsRenderTimeout?: number;
  /**
   * Timeout for waiting until all viewports are `rendered`
   * (and optionally volume-loaded).
   */
  renderedTimeout?: number;
  /**
   * If true (default), also waits for any volume actors referenced by the
   * viewports to report loaded during the rendered phase.
   */
  waitVolumeLoad?: boolean;
};

/**
 * Waits for a full render cycle:
 * 1) any viewport requests render (`needsRender`)
 * 2) all viewports finish rendering (`rendered`)
 */
const waitForViewportRenderCycle = async (
  page: Page,
  options: WaitForRenderCycleToCompleteOptions = {}
) => {
  const {
    needsRenderTimeout = 5000,
    renderedTimeout = 15000,
    waitVolumeLoad = true,
  } = options;
  await waitForAnyViewportNeedsRender(page, { timeout: needsRenderTimeout });
  await waitForViewportsRendered(page, { timeout: renderedTimeout, waitVolumeLoad });
};

```

jbocce (changes requested):
Thanks for the changes. Please see my comments.

tests/utils/waitForViewportsRendered.ts
GhadeerAlbattarni: @jbocce  This is a smart idea, I like it. I test it and it works great.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -1112,7 +1153,45 @@ class CornerstoneViewportService extends PubSubService implements IViewportServi
...
+        nextBaseVolumeIds.length > existingVolumeIds.length &&
+        volumeIdPrefixesMatch(existingVolumeIds, existingVolumeIds.length, nextBaseVolumeIds) &&
+        typeof viewport.addVolumes === 'function'
+      ) {
+        const toAdd = baseVolumeInputs.slice(existingVolumeIds.length);
+        if (toAdd.length) {
+          await viewport.addVolumes(toAdd);
+        }
+      } else {
+        await viewport.setVolumes(baseVolumeInputs);
```
GhadeerAlbattarni, line 1188: @jbocce  Regarding the test (SEGDataOverlayForUnreferencedDisplaySetNoHydration.spec.ts) we were discussing this morning, this is what caused the change.

Before, setVolumes was called with `volumeInputArray`, which had both the background display set (Apparent Diffusion Coefficient) and the SEG which caused later the to create its images and displayed in the viewport as we saw. 
`await viewport.setVolumes(volumeInputArray);`

Now `setVolumes` is called with `baseVolumeInputs` and this comes from `filteredVolumeInputArray` which filters out the SEG so base volume has only (Apparent Diffusion Coefficient)

I do think that what we currently have (after the change) is the correct behaviour, just to display the selected background  and the SEGs masks on the top of it. 

What do you think :) ?

jbocce: @GhadeerAlbattarni, yes I agree. Sorry for the delay to respond.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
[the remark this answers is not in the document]
GhadeerAlbattarni: @jbocce you're right about this

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
GhadeerAlbattarni: @jbocce you have a point here

jbocce (approved):
Looks great to me. We are now headed for decreased testing time. @wayfarer3130 please have a look, I believe I tagged you on a few items.

extensions/cornerstone/src/services/ViewportService/CornerstoneViewportService.ts
```diff
@@ -49,6 +49,47 @@ const EVENTS = {
 const MIN_STACK_VIEWPORTS_TO_ENQUEUE_RESIZE = 12;
 const MIN_VOLUME_VIEWPORTS_TO_ENQUEUE_RESIZE = 6;
 
+function getVolumeActorReferencedIds(viewport: Types.IVolumeViewport): string[] {
+  const actors = viewport.getActors?.() ?? [];
+  return actors
+    .filter(ae => ae.actor?.getClassName?.() === 'vtkVolume')
```
wayfarer3130: Could we use ac instead of ae?  ae is typically used to mean application entity in the DICOM sense, so it took me a moment to realize it wasn't that.

extensions/cornerstone/src/utils/hydrationUtils.ts
```diff
@@ -1,11 +1,80 @@
...
+  const add = (viewportId: string, uids: string[]) => {
+    if (!viewportId) {
+      return;
+    }
+    byId.set(viewportId, { viewportId, displaySetInstanceUIDs: uids });
+  };
+
+  if (Array.isArray(hangingProtocolUpdates)) {
+    for (const hpUpdate of hangingProtocolUpdates) {
+      const entry = hpUpdate as {
```
wayfarer3130: Could we get the hpUpdates list type declared upstream and passed into this function properly declared?  I'd rather add a type of the hanging protocol code that can get applied here and pass it around properly.  I'm not sure if that is in grid viewport or hp updates.  If you use an interface, it is easy enough to add fields later if we need.

extensions/default/src/utils/layerConfigurationUtils.ts
```diff
@@ -131,7 +137,8 @@ export function configureViewportForLayerRemoval(params: {
     viewport.viewportOptions = {};
   }
 
-  viewport.viewportOptions.viewportType = 'volume';
+  const csViewport = cornerstoneViewportService.getCornerstoneViewport(viewportId);
```
wayfarer3130: Could you write this
csViewportOrthographic = ...?.type === 'orthographic'
It just makes the assignment to viewportType a bit more obvious.

wayfarer3130 (approved):
Just a couple of quite minor changes for improved typing/naming in one spot.

jbocce (approved):

PR #5986, opened 2026-04-28
feat(customization): remove or replace the header's undo/redo via ohif.headerRightSide

salimkanoun:
Lets deployments remove the Undo/Redo buttons from the viewer header through configuration — they are noise in workflows without segmentation.

That intent is unchanged from the original version of this PR (see the comment below for the original description). What changed is the mechanism: instead of an `appConfig.hideUndoRedo` boolean that only switches the built-in buttons off, the header now asks the customization service what belongs in that area. A boolean can only say "no"; a customization can also say "this instead", which is the more common request once one site asks to remove something from the header.

### The seam

The right side of the header's menu bar is supplied by the `ohif.headerRightSide` customization. `ViewerHeader` renders it as a component, and `null` renders nothing:

```tsx
const RightSide = customizationService.getCustomization('ohif.headerRightSide');
// ...
<Header RightSide={RightSide ? <RightSide /> : null} />
```

The key is named for **where** it sits rather than **what** it holds, so it still reads correctly if the default contents change. The shipped default (`extensions/default/src/customizations/headerRightSideCustomization.tsx`) is the Undo/Redo buttons, moved out of `ViewerHeader` unchanged. Because it is rendered as a component rather than called as a plain function, a replacement is an ordinary component and may use hooks — the default uses `useSystem()` to reach the commands manager.

### Removing the buttons

`extension-default` ships the `null` override as a named customization module, so it stays a one-line config change:

```js
window.config = {
  customizationService: ['@ohif/extension-default.customizationModule.hideHeaderRightSide'],
};
```

### Putting something else there

```js
import MyHeaderActions from './MyHeaderActions';

window.config = {
  customizationService: [{ 'ohif.headerRightSide': { $set: MyHeaderActions } }],
};
```

### Core change this required

`CustomizationService` resolved scopes with `global ?? mode ?? default`, so a `null` override fell straight through to the default and the override above would have been a silent no-op. Scope resolution now falls through only on `undefined`; an explicit `null` is a value and wins, which is what makes "remove this slot" expressible anywhere in the customization system. Three unit tests cover `null`, `$set: null`, and the unchanged `undefined` fall-through.

### Also in here

- `Header`'s `UndoRedo` prop is renamed `RightSide`, for the same naming reason. **Breaking** for anyone composing `@ohif/ui-next`'s `Header` directly, though in-tree only `ViewerHeader` uses it.
- The separator after that slot is now guarded, so it no longer renders when the slot is empty. This was Greptile's outstanding P1 on the earlier iteration.
- `appConfig.hideUndoRedo` is gone: removed from `AppTypes.Config`, `App.tsx` defaults and propTypes, all 22 `public/config/*.js` files, and `configurationFiles.md`. It never shipped in a release, so there is nothing to deprecate.
- `ohif.headerRightSide` is declared in `AppTypes.Customizations` (typed as `ComponentType | null`), so it gets autocomplete and value checking.
- Documented under `platform/docs/.../customization-service/specificCustomizations.md`.

`master` is merged in; 82 `CustomizationService` tests pass.

## Summary by CodeRabbit

- **New Features**
  - Added an ordered, customizable set of components to the right side of the viewer header.
  - Added header undo/redo controls with an opt-in customization to hide them.
  - Header items can now be reordered, extended, or removed through customization settings.
  - Header patient information continues to respect visibility settings.

- **Documentation**
  - Updated customization guidance and examples for configuring header items.


wayfarer3130:
Reworking the description now that the implementation has changed shape. The goal of the PR is unchanged — let people remove the Undo/Redo buttons through configuration — so preserving the original description here for the record:

> Hi there,
>
> Here is a PR which add a setting to hide the Undo/Redo button (not used in scenario without segmentation), from the header

Thanks for this @salimkanoun — the need is real, we just wanted the seam to be a customization rather than a dedicated config flag so that the whole area can be replaced and not only switched off. Details in the updated description.

salimkanoun:
Thank you very much @wayfarer3130  ! 
I'm really confused, I should have finished my pull request, I had a complicated summer and forgot it.
Thank you for taking care of it.

wayfarer3130 (approved):

PR #5992, opened 2026-05-04
feat: Add customization URL parameter

wayfarer3130:
### Context

In order to allow custom versions of OHIF to be defined/added without having to rebuild OHIF, it is necessary to have a customization framework that can load dynamic modules.  This has been added as a customization= parameter.

### Changes & Results

Added a customization handler for the customization= parameter
Added a requires= export in the loaded global customizations to allow customizations to depend on other ones, eg veterinary depends on veterinaryOverlay
Add an example customization to test with, the start of a veterinary example.

### URL customization examples (`?customization=`)

This PR ships a set of runtime-loadable customization modules under `platform/app/public/customizations/` to demonstrate the `?customization=` framework. Each is a plain ES module whose `global` payload is applied as a global customization at bootstrap (using immutability-helper commands, so it extends the built-in defaults rather than replacing them). Append `&customization=<name>` to any viewer URL to try one — e.g. `/viewer?StudyInstanceUIDs=<uid>&customization=smoothRotate`.

| `?customization=` | What it demonstrates |
|---|---|
| `smoothRotate` | Adding a toolbar button at runtime |
| `ctPresets` | Overriding window/level presets |
| `measurementLabels` | Driving measurement labelling |
| `veterinaryOverlay` | Replacing the viewport overlay layout |
| `veterinary` | A `requires` dependency chain (pulls in `veterinaryOverlay`) |

#### `smoothRotate` — add a toolbar button

Adds a Smooth Rotate button to the More Tools menu that activates the cornerstone `PlanarRotate` tool (drag to rotate the image to any angle, unlike the fixed-90° Rotate Right).

This is enabled by a supporting change: the basic/longitudinal viewers no longer hard-code their toolbar — the cornerstone extension now registers it as customizations (`cornerstone.toolbarButtons` = button defs, `cornerstone.toolbarSections` = the section→ids layout), and the modes reference those by name in `onModeEnter`. The module `$push`es a button definition and its id into the `MoreTools` section. `PlanarRotate` is registered as an available tool by default. Because the defaults register at *default* scope and the URL module applies at *global* scope, the push extends the built-in toolbar instead of clobbering it.

#### `ctPresets` — site-specific window/level presets

`$merge`s a custom `CT` preset list into `cornerstone.windowLevelPresets`. The new presets appear in the window-level menu for CT; presets for other modalities (PT, etc.) are preserved because only the `CT` key is replaced.

#### `measurementLabels` — predefined measurement labels

Sets `measurementLabels` with `labelOnMeasure: true` and an `exclusive` item list, so completing a measurement prompts the user to pick a label (Head/Shoulder/Knee/Toe) instead of free-typing.

#### `veterinaryOverlay` — replace the viewport overlay

Replaces `viewportOverlay.topLeft` / `viewportOverlay.topRight` with a veterinary-style demographics layout (species/breed), demonstrating overlay customization.

#### `veterinary` — dependency chains

A near-empty module that declares `requires: ['veterinaryOverlay']`, showing how one module can pull in another so layered packs (base → shared → site-specific) load in order.

### Security notes

- Loading is allowlisted: values must normalize to `/prefix/name`, full URLs and path traversal (including percent-encoded forms like `%2e%2e`) are rejected, and the resolved URL is verified to stay within the configured prefix directory.
- Modules are applied once at bootstrap and deduped for the page session; they aren't re-imported on SPA navigation.
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

## Release Notes

* **New Features**
  * Added URL-driven customization module support, including a veterinary overlay pack (`?customization=veterinaryOverlay`).

* **Improvements**
  * Overlay item values are now properly formatted, and empty/null values are hidden.
  * Query preservation now keeps customization-related keys and preserves repeated values.
  * Viewer header query preservation now includes customization context.

* **Documentation**
  * Expanded customization service docs with security/behavior details.

* **Tests**
  * Added Jest coverage for customization URL loading/validation and Playwright coverage for veterinary overlay rendering.

* **Refactor**
  * Adjusted customization initialization during app startup.


sedghi (commented):
I’m not fully convinced the added value of this PR justifies the new security surface yet. This introduces URL-driven runtime JavaScript loading via `?customization=`, which means a shared link can change viewer behavior and, depending on deployment config, potentially load executable code into the same browser context as OHIF. The validation does block obvious arbitrary URLs and path traversal, so this is not an immediate “any URL can execute code” issue. But the security boundary becomes the configured customization prefix and whoever can publish files there. If that directory/CDN is writable by the wrong party, this could become XSS-equivalent: token/session access, DICOM metadata exposure, UI manipulation, report tampering, or authenticated API abuse from the victim’s browser.

My current view is that this level of runtime configurability may be better kept in downstream forks or deployment-specific builds, where the deploying team can own the threat model and hosting controls explicitly. I’m not sure it should become a default upstream capability.

wayfarer3130:
> I’m not fully convinced the added value of this PR justifies the new security surface yet. This introduces URL-driven runtime JavaScript loading via `?customization=`, which means a shared link can change viewer behavior and, depending on deployment config, potentially load executable code into the same browser context as OHIF. The validation does block obvious arbitrary URLs and path traversal, so this is not an immediate “any URL can execute code” issue. But the security boundary becomes the configured customization prefix and whoever can publish files there. If that directory/CDN is writable by the wrong party, this could become XSS-equivalent: token/session access, DICOM metadata exposure, UI manipulation, report tampering, or authenticated API abuse from the victim’s browser.
> 
> My current view is that this level of runtime configurability may be better kept in downstream forks or deployment-specific builds, where the deploying team can own the threat model and hosting controls explicitly. I’m not sure it should become a default upstream capability.

If the CDN is writable by the wrong party, it doesn't matter what you do, they can replace the entire OHIF source control.  At that point you are completely open.    

What about adding a user configuration option to specifically and manually add customization prefixes rather than allowing it to be done via customization?  That way we can default to one customization deploy somewhere that we control for the demonstration deployments, making note that is intended for demo purposes only, and same-host http /customization/ prefix path options so that we can deploy with a fixed deployment?

It is clear the advisory board wants SOMETHING that allows dynamic loading.  I agree it needs to be controlled, but it also has to be external to the build process of OHIF, otherwise we will never meet the goals of allowing OHIF to be customized by non-developers.

Some other things we could consider:
1. Allow JSON customizations from controlled locations for full setup
2. Allow JS customizations to be loaded from given path names, but require a SHA sum to match.  The list of valid SHA sums could come from a fixed list of valid/verified items
3. Allow users to "add" new locations for SHA sum validators

That value of this is clearly extremely high given how many people on the meeting wanted something better.  The only question becomes how to make it reasonably safe.  It isn't fully safe, but neither is OHIF in the current configuration.

wayfarer3130:
@sedghi 
Thanks Alireza — I agree with this concern, and I think your framing is right. The issue is not just whether the current validation blocks obvious bad paths or arbitrary URLs. The larger issue is that the current PR creates a new runtime code-loading surface where a URL-controlled value can affect what JavaScript gets loaded into the same browser context as OHIF.

I think we should change the direction of this PR to avoid that for the first version.

What I would propose instead is a narrower model:

1. Executable code must still be trusted at build time.
2. `pluginConfig.json` remains the source of truth for which executable modes/plugins/extensions/customizations are allowed.
3. Any JavaScript module that can be loaded by a customization must be referenced from `pluginConfig.json`, resolved during the build, and included in the build output.
4. `?customization=<name>` may select a customization by name, but it must not be able to introduce a new JavaScript file, package, module specifier, URL, registry source, or version at runtime.
5. Customizations in `public/customization/` should be data-only for now, for example JSON/JSONC/JSON5, and should only be able to reference capabilities that were already approved in the build.

One nuance I want to preserve is that a customization should still be able to add existing build-approved modes/plugins/extensions to the active load list, even if they were not part of the original default load list. In other words, the customization can choose from a build-time-approved set, but cannot expand that set at runtime.

The lifecycle would be something like:

* plugins: if referenced by the customization and present in the build-approved set, they can be loaded immediately;
* modes: if referenced by the customization and present in the build-approved set, they can be added during startup;
* extensions: if referenced by a selected mode/customization and present in the build-approved set, they can be loaded when that mode is loaded.

So the runtime customization can change which already-bundled capabilities become active, but it cannot cause the viewer to fetch or execute new code that was not already included by the build.

That does make this less flexible than the original PR, but I think it gives us a much cleaner security boundary:

* shared links can select among known customizations;
* shared links cannot introduce new executable code;
* the trust boundary stays with the build/deployment process;
* deployments do not need to solve runtime code signing, hashing, CDN trust, or remote plugin provenance yet;
* if a customization references something not included in the build-time allowlist, the viewer should fail closed with a clear error.

I also agree that a full remote plugin ecosystem is a separate design problem. Eventually we may want a trusted plugin catalog, immutable versions, hashes/signatures, provenance, and tighter policy controls. But I do not think this PR needs to solve that yet. For this PR, I think the safer target is: runtime customization can activate/configure build-approved capabilities, but cannot load arbitrary runtime JavaScript.

Would that address your main concern if I rework the PR in that direction?

platform/app/src/appInit.js
```diff
@@ -93,6 +93,11 @@ async function appInit(appConfigOrFunc, defaultExtensions, defaultModes) {
   const loadedExtensions = await loadModules([...defaultExtensions, ...appConfig.extensions]);
   await extensionManager.registerExtensions(loadedExtensions, appConfig.dataSources);
 
+  const { customizationService } = servicesManager.services;
+  // Merge extension default/global modules first; then URL ?customization= globals layer on top.
+  customizationService.init(extensionManager);
+  await customizationService.applyWindowUrlCustomizations();
```
sedghi: shouldn't this be defualt off as we disccussed? i expected a flag/config check here

platform/core/src/services/CustomizationService/CustomizationService.ts
```diff
@@ -159,6 +305,184 @@ export default class CustomizationService extends PubSubService {
     this.modeCustomizations.clear();
   }
 
+  private _urlDefaultImport(url: string): Promise<any> {
```
sedghi: let's move the private methods at the bottom

platform/core/src/services/MultiMonitorService.ts
```diff
@@ -191,7 +191,7 @@ export class MultiMonitorService {
    * Try moving the screen to the correct location - this will only work with
    * screens opened with openWindow containing no more than 1 tab.
    */
-  public async onModeEnter() {
+  public onModeEnter() {
```
sedghi, line 194: add migration guide pleaes

platform/docs/docs/platform/services/customization-service/specificCustomizations.md
```diff
@@ -0,0 +1,131 @@
...
+
+- **Purpose**: Controls how values in the `?customization=` URL parameter are resolved and loaded.
+- **Main field**: `prefixes` maps logical prefixes to base URLs.
+- **Default prefix**: `default` maps to `./customizations/`.
+- **Not the same as preserve keys**: This setting does not decide which query keys are preserved. It only controls how customization modules are loaded from `?customization=...`.
+
+Example:
+
+```js
+window.config = {
```
sedghi, line 50: this is not clear what it will end up as. Please add documentation what is default, what is prefix, what is remote

wayfarer3130:
@sedghi - as requested, I've added additional examples, described in the testing block above, as well as in the documentation internally.

platform/core/src/services/MultiMonitorService.ts
wayfarer3130: This was done incorrectly before - the method is already documented as NOT being async, so this is just fixing an incorrect declaration.

platform/app/src/appInit.js
wayfarer3130: Changing to be only when configured.

wayfarer3130: Did this via app config, plus added the dev.js app config and the netlify.js version for those two deployments.

platform/docs/docs/platform/services/customization-service/specificCustomizations.md
wayfarer3130: I've updated this documentation to specifically add the security information and describe where this is located, and that it is off by default.  I believe that addresses your concerns.

platform/app/src/appInit.js
```diff
@@ -93,6 +93,15 @@ async function appInit(appConfigOrFunc, defaultExtensions, defaultModes) {
...
   await extensionManager.registerExtensions(loadedExtensions, appConfig.dataSources);
 
+  const { customizationService } = servicesManager.services;
+  // Merge extension default/global modules first, then layer on any customizations
+  // requested via the `?customization=` URL parameter. Those are only loaded when
+  // `appConfig.customizationUrlPrefixes` allows their prefix; the feature is off by
+  // default, and a value with an unconfigured prefix throws here (aborting startup)
+  // rather than being silently ignored.
+  customizationService.init(extensionManager);
+  await customizationService.applyWindowUrlCustomizations();
```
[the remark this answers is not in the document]
wayfarer3130: This is intended behaviour.  Illegal customizations stop the app.  Missing customizations log but continue.

platform/docs/docs/migration-guide/3p12-to-3p13/customization-url.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+The same shape is accepted by the app config. The legacy array / object form is
+still supported and is applied to the Global scope during `init()` exactly as
+before:
+
+```js
+window.config = {
+  customizationUrlPrefixes: { default: './customizations/' },
```
sedghi, line 107: can we add a throw that customizationUrlPrefixes itslef never ever gets to be configured by customization itself

platform/docs/docs/migration-guide/3p12-to-3p13/customization-url.md
```diff
@@ -0,0 +1,192 @@
...
+A customization module — whether loaded from a `?customization=` data file or
+declared inline in `appConfig.customizationService` — can tag its payload with
+the lifecycle phase it should be applied in. This makes ordering deterministic
+regardless of when extensions and modes load, and lets a single source target
+the pre-extension, global, and mode-entry time frames at once.
+
+```jsonc
+{
+  // Other URL customization data files to resolve FIRST (depth-first).
+  "requires": ["base"],
```
sedghi: "base" is a bit unclear, what does this mean to require base?

sedghi: is base a customization itself?

platform/docs/docs/migration-guide/3p12-to-3p13/customization-url.md
```diff
@@ -0,0 +1,192 @@
...
+
+  // Applied (Global scope) BEFORE extensions register — in place while they init.
+  "preExtension": { "someId": { "$set": "value" } },
```
sedghi: I'm not a big fan of preExtension, how about bootstrap? or earlyGlobal?

sedghi (approved):
Thanks for pushing this forward and taking the previous feedback into account. Please take a look at my latest comments.

platform/docs/docs/migration-guide/3p12-to-3p13/customization-url.md
sedghi: no one can have this key in their customization

wayfarer3130: That is already the point of having it in the window.config is that it ISN'T in the customization, so no matter what one does, it doesn't get updated via customizations.

PR #5994, opened 2026-05-05
fix(measurement-tracking): restore tracked state on undo after Delete all

### LINKED ISSUE #5367: [Bug] Undo after Delete All:  Measurements Reappeared as Untracked
### Describe the Bug

When a user performs an undo after deleting all the measurements using the Delete all button, the measurments reappeared as untracked. if the measurement tracking mode is set as "measurementTrackingMode: 'simplified'" expecting the measurments should be reappeared as tracked.

https://github.com/user-attachments/assets/65f4dcde-b227-4723-ba63-3bcc6534a489

### Steps to Reproduce

1. Add the config of measurement tracking mode in default.js config file (measurementTrackingMode: 'simplified')
2. Launch a Dicom study
3. Draw few measurements
4. Delete all the measurements using the closed button
5. Perform the undo operation and observe the viewport and measurement panel

### The current behavior

The measurements reappeared on the viewport as untrackedu

### The expected behavior

All measurements should reappear on the viewport in single undo operation and should be tracked.

### OS

windows 11

### Node version

22

### Browser

139.0.7258.139

GhadeerAlbattarni:
### Context

Fixes #5367 

When a user draws measurements, tracks them, then clicks **Delete All → Untrack**,  pressing **Undo** brings the measurements back as untracked (comes as dashed line)

Another bug found, when user clicks on `delete` after undo, `delete confirmation prompt` skipped and doesn't shown. 

## Root Cause
Two issues:

**1. XState assign batching**
clearContext ran before clearAllMeasurements, losing tracking state
**2. Undo system missing tracking state**
Cornerstone's `DefaultHistoryMemo` restored annotations, but OHIF tracking context was not restored

**3. For delete confirmation issue**
After undo, restored measurements have isDirty = false, so the delete confirmation was never shown on the second delete attempt. 

### Changes 

| File | Change |
|------|--------|
| `platform/core/src/services/MeasurementService/MeasurementService.ts` | Add optional `metadata` param to `clearMeasurements()` — backward compatible, all existing callers unaffected |
| `extensions/cornerstone/src/initMeasurementService.ts` | In `MEASUREMENTS_CLEARED` subscriber, push a tracking-state memo into the undo group when `trackingContext` is present |
| `extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/TrackedMeasurementsContext.tsx` | Read tracking context from `evt` in `clearAllMeasurements`; register `restoreTrackedSeries` command |
| `extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx` | Pass `trackedStudy` and `trackedSeries` with the `UNTRACK_ALL` event;  replace `isDirty` check with presence check so the delete confirmation always appears when measurements exist |

### Results
- Measurements comes back as tracked when user clicks undo after delete all.
- Prompt deletion dialog shows when user clicks on delete after undo. 

https://github.com/user-attachments/assets/6a09cab5-d1a3-4952-94e4-ce75ff038d9a

### Testing
- Open a study in viewer mode
-  Draw one or more measurements in the viewport
- Click Yes to track measurements
- Verify measurements are tracked (solid lines, visible in panel)
- Open the right measurement panel
- Click Delete → Untrack
-- Verify measurements are removed and the panel is empty
- Click Undo
-- Verify measurements are restored with solid lines (not dashed)
-- Verify measurements appear in the panel
- Click Delete → Untrack again
-- Verify the confirmation prompt is shown again

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.13.0
  

<h3>Greptile Summary</h3>

This PR fixes two bugs in the measurement-tracking flow: (1) undo after "Delete All → Untrack" restored annotations as untracked (dashed lines) because XState's `clearContext` ran before `clearAllMeasurements`, discarding the series context needed for the undo memo; (2) after undo, the delete confirmation prompt was skipped because restored measurements had `isDirty = false`.

- Captures `trackedStudy`/`trackedSeries` from the event payload (rather than context) in `clearAllMeasurements`, then pushes a `DefaultHistoryMemo` into the Cornerstone undo group that restores or clears XState tracking context on undo/redo respectively.
- Adds `CLEAR_TRACKING_CONTEXT` to both `idle` (no-op for strict-mode safety) and `tracking` (redo path) states, and replaces the `isDirty` delete-guard with a measurement-presence check so the confirmation dialog always appears when measurements exist.
- Includes three new E2E tests covering the restored-tracked-state and reappearing-prompt scenarios in both standard and simplified tracking modes.

The change is well-scoped and no data-loss or silent-failure paths were found in the changed code.

The root causes are correctly identified and addressed. The tracking memo is atomically grouped with annotation memos. The new CLEAR_TRACKING_CONTEXT event is guarded in both machine states. The isDirty → presence-check replacement is correct. All findings are style or robustness suggestions with no impact on runtime correctness in the expected usage paths.

No files require special attention. The loosest point is the Record<string, unknown> metadata typing in MeasurementService.clearMeasurements, but it carries no runtime risk given the single current caller.

| Filename | Overview |
|----------|----------|
| extensions/cornerstone/src/initMeasurementService.ts | Pushes a `DefaultHistoryMemo` into the Cornerstone undo group when `trackingContext` is present, enabling undo to call `restoreTrackedSeries` and redo to call `clearTrackedSeries` — keeping XState in sync with Cornerstone's annotation history. |
| extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/TrackedMeasurementsContext.tsx | Reads `trackedStudy`/`trackedSeries` from the event payload in `clearAllMeasurements` (before XState clears context), registers `restoreTrackedSeries` and `clearTrackedSeries` commands, and adds an optional-chain guard on `evt.data?.isBackupSave`. |
| extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/measurementTrackingMachine.js | Adds `CLEAR_TRACKING_CONTEXT` as a no-op in `idle` state (strict-mode safety) and as a context-clearing self-transition in `tracking` state (redo path). Minor concern: `setTrackedStudyAndMultipleSeries` appends series instead of replacing, safe in current flow but fragile. |
| extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx | Passes `trackedStudy`/`trackedSeries` with `UNTRACK_ALL` event so `clearAllMeasurements` captures them before context is wiped; replaces `isDirty` check with measurement-presence check so the delete confirmation always appears after undo. |
| platform/core/src/services/MeasurementService/MeasurementService.ts | Adds optional `metadata` param to `clearMeasurements()`, spread into the broadcast event payload — backward compatible; `metadata` type is `Record<string, unknown>`, which allows key collisions with `measurements` at the call site. |
| tests/MeasurementPanel.spec.ts | Adds three new E2E tests covering undo-after-delete-all (standard and simplified modes) and delete confirmation reappearance; `.nth(2)` selector for the second measurement line is positionally fragile. |
| extensions/default/src/ViewerLayout/ViewerHeader.tsx | Adds `data-cy="undo-btn"` and `data-cy="redo-btn"` test attributes to the undo/redo buttons for Playwright targeting. |
| tests/pages/MainToolbarPageObject.ts | Exposes `undo` and `redo` getters on the page object, using the new `data-cy` test IDs. |
| tests/pages/RightPanelPageObject.ts | Adds a `rows` locator for `data-row` elements alongside the existing `getMeasurementCount` helper. |
| tests/pages/ViewportPageObject.ts | Extends `SvgInnerElement` union type to include `'line'` so tests can query SVG line elements for dashed-stroke assertion. |

</details>

</details>

`````markdown
Fix the following 3 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 3
platform/core/src/services/MeasurementService/MeasurementService.ts:719-725
**Overly-permissive `metadata` type allows silent key collision**

`Record<string, unknown>` permits any key, including `measurements`. If a future caller accidentally passes `{ measurements: [] }` in `metadata`, the spread (`{ measurements, ...metadata }`) will silently replace the actual measurements array in the broadcast event, and all subscribers downstream (including the new tracking memo logic) will receive an empty list without any error. Narrowing the type to a concrete interface would make this impossible at compile time.

### Issue 2 of 3
extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/measurementTrackingMachine.js:422-435
**`setTrackedStudyAndMultipleSeries` appends rather than replaces series**

The `SET_TRACKED_SERIES` handler in the `tracking` state calls `setTrackedStudyAndMultipleSeries`, which does `[...ctx.trackedSeries, ...seriesInstanceUIDs]`. In the current undo flow this is safe because `clearContext` always empties `trackedSeries` before the tracking memo fires `SET_TRACKED_SERIES`. However, the append-not-replace semantics mean any code path that sends `SET_TRACKED_SERIES` while `trackedSeries` is already non-empty would silently accumulate duplicate series, causing incorrect filtering in the measurements panel. Consider using `trackedSeries: seriesInstanceUIDs` (replace) when the event carries an explicit `SeriesInstanceUIDs` array, or introduce a dedicated `restoreTrackedSeries` action that always replaces.

### Issue 3 of 3
tests/MeasurementPanel.spec.ts:272-277
**`.nth(2)` skips the element at index 1 for the second measurement line**

Both the standard and simplified-mode tests assert the second measurement via `activeViewport.svg('line').nth(2)`, which skips index 1. If a `LengthTool` annotation renders two SVG `<line>` elements per annotation (main segment + tick/handle), index 2 would be the second annotation's line — intentional. However, if CS Tools changes the element count per annotation, this assertion will silently target the wrong line. A locator tied to a data attribute specific to the second annotation would be more resilient.

`````

</details>

<sub>Reviews (7): Last reviewed commit: ["refactor: use if/else instead of ternary..."](https://github.com/ohif/viewers/commit/44c9f4ec57f1364d96681453e55f3801e98f2417) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30926244)</sub>

### CODE REVIEW

File: extensions/cornerstone/src/initMeasurementService.ts [hunk 1]
```diff
@@ -373,6 +375,20 @@ const connectMeasurementServiceToTools = ({
...
+        operationType: 'trackingState',
+        restoreMemo(undo?: boolean) {
+          if (undo === true) {
+            commandsManager.run('restoreTrackedSeries', trackingContext);
+          }
+        },
+      });
+    }
+
     commandsManager.run('endRecordingForAnnotationGroup');
```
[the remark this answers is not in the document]
GhadeerAlbattarni: @greptile-apps Yes, it is. `DefaultHistoryMemo.push()` itself checks `this.isRecordingGrouped` and calls `pushGrouped()` when inside a recording session (see `HistoryMemo.push` in `@cornerstonejs/core`). The tracking-state memo is captured in the same group as the annotation deletions. So, one undo press restores everything.

tests/MeasurementPanel.spec.ts
```diff
@@ -213,3 +221,132 @@ test('checks if measurement item can be deleted through the context menu on the
...
+  await mainToolbarPageObject.undo.click();
+
+  await waitForViewportsRendered(page);
+
+  await expect(rightPanelPageObject.measurementsPanel.panel.rows).toHaveCount(2);
+
+  const activeViewport = await viewportPageObject.active;
+
+  const firstMeasurementLine = activeViewport.svg('line').first();
+  await expect(firstMeasurementLine).not.toHaveAttribute('stroke-dasharray', '4,4');
```
jbocce: Would it not be safer to expect the dashing to be solid (i.e. not dashed)? I worry that if we ever change the dashinig pattern for untracked measurements the test might pass but it should fail.

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
```diff
@@ -23,19 +22,15 @@ function PanelMeasurementTableTracking(props) {
...
-    sendTrackedMeasurementsEvent('UNTRACK_ALL', {});
+    sendTrackedMeasurementsEvent('UNTRACK_ALL', { trackedStudy, trackedSeries });
   };
 
   const onDelete = () => {
-    const hasDirtyMeasurements = measurementService
-      .getMeasurements()
-      .some(measurement => measurement.isDirty);
-    hasDirtyMeasurements
+    const hasMeasurements = measurementService.getMeasurements().length > 0;
```
jbocce, line 32: I think this is somewhat of an unrelated change, but I think it is the right thing to do! Good stuff!

tests/MeasurementPanel.spec.ts
jbocce: Maybe use expect(firstMeasurementLine).not.toHaveAttribute('stroke-dasharray') instead?

jbocce (approved):
It looks good to me.

jbocce (changes requested):
Just one small change.

jbocce:
@claude review

jbocce:
Please update your branch.

tests/MeasurementPanel.spec.ts
GhadeerAlbattarni: This looks better

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
GhadeerAlbattarni: You're right, it's not related. Thanks!

extensions/cornerstone/src/initMeasurementService.ts
```diff
@@ -373,6 +375,29 @@ const connectMeasurementServiceToTools = ({
...
+        restoreMemo(undo?: boolean) {
+          if (undo === true) {
+            commandsManager.run('restoreTrackedSeries', trackingContext);
+          } else if (undo === false) {
+            commandsManager.run('clearTrackedSeries');
+          }
+        },
+      });
+    }
+
```
jbocce, line 400: Please consider adding a similar comment in the measurementTrackingMachine for CLEAR_TRACKING_CONTEXT.

extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/TrackedMeasurementsContext.tsx
```diff
@@ -376,6 +382,22 @@ function TrackedMeasurementsContextProvider(
...
+          SeriesInstanceUIDs,
+        }),
+    });
+  }, [commandsManager, sendTrackedMeasurementsEvent]);
+
+  useEffect(() => {
+    commandsManager.registerCommand('DEFAULT', 'clearTrackedSeries', {
+      commandFn: () => sendTrackedMeasurementsEvent('CLEAR_TRACKING_CONTEXT'),
+    });
+  }, [commandsManager, sendTrackedMeasurementsEvent]);
```
jbocce, line 394: Please combine these three useEffect into one - see below. I also have a suggested, updated comment to use.

```
    useEffect(() => {
      // These commands are bound to the context's sendTrackedMeasurementsEvent, so they have
      // to be registered from inside this (React) component. Re-registration on each
      // render is safe — registerCommand overwrites by key.
      commandsManager.registerCommand('DEFAULT', 'loadTrackedSRMeasurements', {
        commandFn: props => sendTrackedMeasurementsEvent('HYDRATE_SR', props),
      });
      commandsManager.registerCommand('DEFAULT', 'restoreTrackedSeries', {
        commandFn: ({ StudyInstanceUID, SeriesInstanceUIDs }) =>
          sendTrackedMeasurementsEvent('SET_TRACKED_SERIES', {
            StudyInstanceUID,
            SeriesInstanceUIDs,
          }),
      });
      commandsManager.registerCommand('DEFAULT', 'clearTrackedSeries', {
        commandFn: () => sendTrackedMeasurementsEvent('CLEAR_TRACKING_CONTEXT'),
      });
    }, [commandsManager, sendTrackedMeasurementsEvent]);
```

jbocce (changes requested):
Some small changes.

extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/TrackedMeasurementsContext.tsx
GhadeerAlbattarni: Thanks for this :)

extensions/measurement-tracking/src/panels/PanelMeasurementTableTracking.tsx
```diff
@@ -23,19 +22,15 @@ function PanelMeasurementTableTracking(props) {
...
+    hasMeasurements
```
wayfarer3130: Since you are tweaking this line anyways, can you modify it to an if statement - ? operator is intended for deciding between values, not for control flow, and doing a show is definitely control flow.

extensions/measurement-tracking/src/contexts/TrackedMeasurementsContext/TrackedMeasurementsContext.tsx
```diff
@@ -369,11 +375,22 @@ function TrackedMeasurementsContextProvider(
   ]);
 
   useEffect(() => {
-    // The command needs to be bound to the context's sendTrackedMeasurementsEvent
-    // so the command has to be registered in a React component.
+    // These commands are bound to the context's sendTrackedMeasurementsEvent, so they have
+    // to be registered from inside this (React) component. Re-registration on each
+    // render is safe — registerCommand overwrites by key.
     commandsManager.registerCommand('DEFAULT', 'loadTrackedSRMeasurements', {
```
wayfarer3130, line 381: In general I dislike this type of registration - I'd rather see the context use zustand, but I understand why it is necessary right now.  Can you create an issue suggesting the use of zustand for this, and at the same time make it multi-study aware, using display sets instead of series for consistency?  It is too big a change for this PR, but it would be nice to track associated with this issue.

wayfarer3130 (approved):
Tested with and without tracking enabled, seems to work, and given the current state of the tracking design, I think it is worth including this.  One small code change for consistency was requested, and a separate issue to be created.

jbocce (approved):
Looks great. Thanks for this.

PR #6002, opened 2026-05-07
feat(tests): Add tests for duplicating contour segments

diattamo:
### Context

Adds end-to-end coverage for the contour-segment duplicate action exposed in the
right-panel segmentation list. Until now there were no tests verifying that
duplicating a contour segment:

- adds a new row to the segmentation panel,
- can be repeated against the same source segment,
- and renders an identical contour on the viewport as the source.

### Changes & Results

- Add `tests/ContourSegmentDuplicate.spec.ts` with three Playwright tests against the
  segmentation mode and the bundled RTSTRUCT study:
  - `should duplicate a contour segment and add a new row to the panel` — asserts the
    panel goes from 4 → 5 rows, the new row's default name is `Segment 5`, and the
    original titles (`Threshold`, `Big Sphere`) are unchanged.
  - `should duplicate the same segment multiple times` — duplicates segment 0 twice
    and verifies both new rows (`Segment 5`, `Segment 6`) appear in order.
  - `should render the duplicated contour on the viewport` — hides all segments,
    isolates the source via `toggleVisibility`, captures its SVG path with
    `getSvgPath`, duplicates it, isolates the duplicate, and asserts the SVG path
    `d` attribute matches the source.

### Testing
`bun test:e2e:ui -- tests/ContourSegmentDuplicate.spec.ts`
3. All three tests should pass.

#### PR

#### Code

etc.)

additions or removals. (N/A — test-only change)

#### Tested Environment

- [x] OS: macOS 15 (Darwin 25.3.0)
- [x] Node version: v25.4.0
- [x] Browser: Chrome 146.0.7680.180 (Official Build, arm64) / Chromium (Playwright)

<h3>Greptile Summary</h3>

This PR adds end-to-end tests for the contour-segment duplicate action and refactors the existing `getSvgPath` utility into a more general `getSvgAttribute` helper that is adopted across all three affected spec files.

- **New spec** (`ContourSegmentDuplicate.spec.ts`): three Playwright tests covering panel row count after duplication, repeated duplication of the same segment, and SVG-path equality between source and duplicate on the viewport.
- **Utility refactor**: `getSvgPath` is deleted and replaced by `getSvgAttribute`, which accepts element type, attribute name, viewport ID, and `nth` index; `SvgInnerElement` is now exported from `ViewportPageObject.ts` and the invalid `'d'` member (an attribute, not a tag) is removed from the union.
- **Migration**: `ContourSegNavigation.spec.ts` and `ContourSegmentToggleVisibility.spec.ts` are updated to use the new helper with no functional changes.

Merge after fixing the two unawaited `toHaveCount` assertions in the third test; the rest of the change is correct.

Two `toHaveCount` assertions in `ContourSegmentDuplicate.spec.ts` (lines 90 and 103) are missing `await`. These guards confirm that exactly one SVG path is visible before capturing and comparing paths — without `await`, a wrong count is silently swallowed and the final `toBe` assertion can compare unintended paths while still reporting success.

tests/ContourSegmentDuplicate.spec.ts — the two unawaited `toHaveCount` assertions need fixing before the third test can be trusted.

| Filename | Overview |
|----------|----------|
| tests/ContourSegmentDuplicate.spec.ts | New test file with three duplicate-contour specs; two `toHaveCount` assertions are missing `await`, making the isolation guards in the SVG-path comparison test silently non-functional; `page` is also destructured but unused in tests 1 and 2. |
| tests/utils/getSvgAttribute.ts | New generalised utility replacing `getSvgPath`; correctly parameterises element type, attribute name, viewport ID, and nth index. |
| tests/pages/ViewportPageObject.ts | Exports `SvgInnerElement` type and correctly removes the invalid `'d'` member (an attribute name, not an element tag). |
| tests/ContourSegNavigation.spec.ts | Migrated all `getSvgPath` calls to the new `getSvgAttribute` helper; logic and assertions unchanged. |
| tests/ContourSegmentToggleVisibility.spec.ts | Migrated `getSvgPath` calls to `getSvgAttribute`; no functional changes. |
| tests/utils/index.ts | Barrel updated to re-export `getSvgAttribute` in place of the deleted `getSvgPath`. |

</details>

</details>

`````markdown
Fix the following 4 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 4
tests/ContourSegmentDuplicate.spec.ts:89-90
Missing `await` on `toHaveCount` assertions — these are async Playwright matchers that return `Promise<void>`. Without `await`, if the count is wrong (e.g. the global hide didn't fully isolate the segment), the rejected Promise is silently swallowed and the test passes anyway, making the isolation guards ineffective before the SVG-path comparison. The same pattern applies to the second occurrence at line 103.

```suggestion
  const sourceSvgPaths = (await viewportPageObject.getById('default')).svg('path');
  await expect(sourceSvgPaths, 'Expected only one SVG path element for the original segment').toHaveCount(1);
```

### Issue 2 of 4
tests/ContourSegmentDuplicate.spec.ts:102-103
Same missing `await` issue as line 90 — `toHaveCount` returns a `Promise<void>` that must be awaited. Without it, a wrong count at this isolation checkpoint silently passes.

```suggestion
  const duplicatedSvgPaths = (await viewportPageObject.getById('default')).svg('path');
  await expect(duplicatedSvgPaths, 'Expected only one SVG path element for the duplicated segment').toHaveCount(1);
```

### Issue 3 of 4
tests/ContourSegmentDuplicate.spec.ts:1-7
`getSvgAttribute` is already re-exported from the `./utils` barrel (`utils/index.ts` line 60), so importing it directly from `./utils/getSvgAttribute` is inconsistent with how every sibling spec uses it. Importing from the barrel keeps all utility imports in one place.

```suggestion
import {
  expect,
  test,
  visitStudy,
  waitForViewportsRendered,
  getSvgAttribute,
} from './utils';
```

### Issue 4 of 4
tests/ContourSegmentDuplicate.spec.ts:28-31
`page` is destructured in both the first and second tests but never referenced in either test body. Removing unused fixture parameters avoids TypeScript no-unused-vars warnings and keeps the fixture list accurate.

```suggestion
test('should duplicate a contour segment and add a new row to the panel', async ({
  rightPanelPageObject,
}) => {
```

`````

</details>

<sub>Reviews (8): Last reviewed commit: ["fix accidental deletion"](https://github.com/ohif/viewers/commit/0511bf84754e6e302def56b47dd379b20876d133) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31234350)</sub>

### CODE REVIEW

File: tests/ContourSegmentDuplicate.spec.ts [hunk 1]
```diff
@@ -0,0 +1,104 @@
...
+  // Hide again to show duplicate only
+  await segment0.toggleVisibility();
+  await duplicatedSegment.click();
+
+  const duplicatedSvgPath = await getSvgPath(viewportPageObject);
+
+  expect(
+    duplicatedSvgPath,
+    'Expected the duplicated segment to have the same SVG path as the source'
+  ).toBe(sourceSvgPath);
```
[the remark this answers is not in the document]
diattamo: added the null checks

tests/ContourSegmentDuplicate.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  await expect(segment0.title).toHaveText(defaultSegment0Name);
+
+  await segment0.actions.duplicate();
+
+  const countAfterDuplicate = await panel.getSegmentCount();
+  expect(countAfterDuplicate, 'Expected one additional segment row after duplicating').toBe(5);
+
+  //New segment's default name is formatted as "Segment {segmentCount}"
+  const newSegmentTitle = await panel.nthSegment(initialCount).title.textContent();
+  expect(newSegmentTitle, 'Expected correct title for duplicated segment').toBe(`Segment 5`);
```
jbocce: I think whenever possible (I know it is not always convenient) we want to follow the Playwright best practices. So shouldn't the above two lines be rolled into the one that follows?
```
await expect(panel.nthSegment(initialCount).title, 'Expected correct title for duplicated segment').toHaveText(`Segment 5`);
```

jbocce: Any place where we get textContent and then assert/expect it should be changed.

tests/ContourSegmentDuplicate.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  // Hide again to show duplicate only
+  await segment0.toggleVisibility();
+  await duplicatedSegment.click();
+
+  const duplicatedSvgPath = await getSvgPath(viewportPageObject);
+  expect(duplicatedSvgPath, 'Expected a visible SVG path for the duplicated segment').not.toBeNull();
+
+  expect(
+    duplicatedSvgPath,
```
jbocce, line 106: I love what you did here. Do you think it might be worth also adding a `getSvgPaths` utility that gets ALL the paths for a viewport? That way we could also assert the number of paths after dupliation?

jbocce (changes requested):
Great tests. A few suggestions.

tests/ContourSegmentDuplicate.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+test('should duplicate the same segment multiple times', async ({
+  page,
+  rightPanelPageObject,
+}) => {
+  const panel = rightPanelPageObject.contourSegmentationPanel.panel;
+
+  const segment0 = panel.nthSegment(0);
+
+  await segment0.actions.duplicate();
+  await expect(page.getByTestId('data-row')).toHaveCount(5);
```
jbocce: Here we went to the page.getByTestId, yet in other places we used getSegmentCount. Using page object is always preferred. I guess this is one thing we should point out to AI as well somehow (perhaps you already have), that instead of page.*, it should use page objects or add to them as is necessary.

tests/ContourSegmentDuplicate.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+}) => {
+  const mode = 'segmentation';
+  await visitStudy(page, studyInstanceUID, mode, 2000);
+
+  await leftPanelPageObject.loadSeriesByModality('RTSTRUCT');
+  await waitForViewportsRendered(page);
+  await expect(DOMOverlayPageObject.viewport.segmentationHydration.locator).toBeVisible();
+
+  await DOMOverlayPageObject.viewport.segmentationHydration.yes.click();
+  await expect(page.getByTestId('data-row')).toHaveCount(4);
```
jbocce: Please use panel.getSegmentCount. See my note below.

tests/ContourSegmentDuplicate.spec.ts
diattamo: Updated to use toHaveText()

tests/ContourSegmentDuplicate.spec.ts
diattamo: Updated to use the getSegmentCount()

tests/ContourSegmentDuplicate.spec.ts
diattamo: Just to clarify. 
Will the `getSvgPaths` returns the total count of paths to then check count before and after 1 duplication for example, to assert it went up by 1?

jbocce: Yes I think so.

jbocce (approved):
@diattamo  Approved. However, please let me know if you will be adding that getSvgPaths or not. Thanks.

tests/ContourSegmentDuplicate.spec.ts
diattamo: I implemented the check for the count as discussed. 
It required doing some refactoring on ther test files fyi

tests/utils/getSvgAttribute.ts
```diff
@@ -0,0 +1,17 @@
+import { SvgInnerElement, ViewportPageObject } from '../pages/ViewportPageObject';
+
+const getSvgAttribute = async (
+  viewportPageObject: ViewportPageObject,
```
jbocce: Instead of using positional parameters, let's refactor to an object. So something like the following which allows more easily for excluding say viewportId and specifying nth. With positional parameters, we would always have to pass a `viewportId` of `undefined` or `default` if we wanted the default viewport but an `nth` other than `0`.

```
const getSvgAttribute = async ({
  viewportPageObject,
  svgInnerElement,
  attributeName,
  viewportId = 'default',
  nth = 0,
}: {
  viewportPageObject: ViewportPageObject;
  svgInnerElement: SvgInnerElement;
  attributeName: string;
  viewportId?: string;
  nth?: number;
}) => {
  return (await viewportPageObject.getById(viewportId))
    .svg(svgInnerElement)
    .nth(nth)
    .getAttribute(attributeName);
};
```

tests/ContourSegmentDuplicate.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  // Hide everything, to be able to grab only the SVG path of the segment to duplicate
+  await rightPanelPageObject.contourSegmentationPanel.segmentsVisibilityToggle.click();
+  const segment0 = panel.nthSegment(0);
+  await segment0.toggleVisibility();
+  await segment0.click();
+
+  const sourceSvgPath = await getSvgAttribute(viewportPageObject, 'path', 'd');
+  expect(sourceSvgPath, 'Expected a visible SVG path for the source segment').not.toBeNull();
+  const sourceSvgPathsCount = await (await viewportPageObject.getById('default')).svg('path').count();
+  expect(sourceSvgPathsCount, 'Expected only one SVG path element for the original segment').toBe(1);
```
jbocce: I think the playwright way to expect/assert an count is...

expect(viewportPageObject.getById('default')).svg('path')).toHaveCount(1);

tests/ContourSegmentDuplicate.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  const duplicatedSegment = panel.nthSegment(4);
+
+  // Hide again to show duplicate only
+  await segment0.toggleVisibility();
+  await duplicatedSegment.click();
+
+  const duplicatedSvgPath = await getSvgAttribute(viewportPageObject, 'path', 'd');
+  expect(duplicatedSvgPath, 'Expected a visible SVG path for the duplicated segment').not.toBeNull();
+  const duplicatedSvgPathsCount = await (await viewportPageObject.getById('default')).svg('path').count();
+  expect(duplicatedSvgPathsCount, 'Expected only one SVG path element for the duplicated segment').toBe(1);
```
jbocce: See previous commment about asserting counts. Thanks.

jbocce (changes requested):
Couple of small things. Thanks.

tests/utils/getSvgAttribute.ts
diattamo: Updated + refactor other tests

tests/ContourSegmentDuplicate.spec.ts
diattamo: Gotcha.

jbocce (approved):
Your code looks great. However there is a failing playwright test not caused by you that I am investigating as well.

PR #5660, opened 2025-12-17
components(Calendar): Updated date picker components for improved UX

dan-rukas:
### Context

Migrated the Calendar component to the latest shadcn/ui implementation using react-day-picker v9 API. Improves UX consistency with arrows for paging between months and better dropdown month/year navigation.

### Changes & Results

Screenshot 2025-12-17 at 12 11 49 PM

- New design replaces a previous design before added additional components
- New component UX includes arrows for paging between months
- Calendar Component: Replaced with updated shadcn Calendar (react-day-picker v9)
- Month/Year Navigation: Added dropdown selectors for quick date navigation (captionLayout="dropdown")
- i18n Support: Preserved locale support for 12 languages via react-day-picker/locale
- DateRange Component: Updated to use new Calendar API (initialFocus → autoFocus)
- Styling: Matches OHIF Design, Aligned with shadcn design patterns, fixed popover sizing issues

### Testing

Open Study List and filter by Study Date.

#### PR

- [x] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [x] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [x] OS: macOS Sequoia 15.7.1 (24G231)(
- [x] Node version: v20.19.1
- [x] Browser: Google Chrome 143.0.7499.110 (Official Build) (arm64)
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

## Summary by CodeRabbit

* **New Features**
  * Added new Calendar day button component for enhanced date selection.
  * Introduced locale support to Calendar with automatic language detection.
  * Added customizable caption layouts and button variants to Calendar.

* **Improvements**
  * Enhanced date range picker with dropdown month navigation and configurable year constraints.

* **Chores**
  * Updated package dependencies including date-fns and react-day-picker.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OHIF/Viewers/pull/5660?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<h3>Greptile Summary</h3>

[...]

PR #6005, opened 2026-05-08
feat(WorkListUINext): New Study List (WorkListUINext)

jbocce:
## Context
This PR replaces the legacy `WorkList` study-list route with a new `WorkListUINext` route, built on a fresh `ui-next` study-list system (`StudyList`, `DataTable`, `PreviewContainer`, `PreviewPatientSummary`, etc.). The legacy `WorkList` remains importable for fallback, but `/` now resolves to `WorkListUINext` by default. The change set spans 138 commits and ~5,600 lines added.

Key drivers:

- New design for the study list (filters, pagination, sorting, preview panel with patient summary + thumbnails/list toggle, settings popover, workflow buttons, white-labelled logo, OIDC logout, investigational-use dialog).
- Table state (sorting, pagination, filters) synced to URL and `sessionStorage` via the new `useStudyListStateSync` hook.
- Refetch behavior in `DataSourceWrapper` simplified to client-side pagination — one server query per filter change instead of per page change.
- DICOMWeb data source extended with a `thumbnailRequestStrategy` (`bulkDataRetrieve` default, or `fetch`) and a series-level QIDO fallback for thumbnail fetching.
- New `workListUINext.previewSeriesView` customization (`'all' | 'thumbnails' | 'list'`) with documentation.

### Basics
image

### Workflows/modes
image

### Preview side panel
image

### 1. Routing (`platform/app/src/routes/index.tsx`)

- The default `/` route now renders `WorkListUINext` via `DataSourceWrapper`.
- The legacy `WorkList` import is retained and commented as a swap-in alternative.

### 2. `DataSourceWrapper` (`platform/app/src/routes/DataSourceWrapper.tsx`)

- Refetch logic is simplified to client-side pagination — server queries fire once per filter change (with a single bulk query up to `appConfig.queryLimit ?? 101`) instead of refetching on every page/offset change.
- Pagination, sorting, and offsets are handled in-memory by `WorkListUINext` via `useStudyListStateSync`.
- A new `hasFetchedOnce` flag is plumbed through to the layout so the worklist can suppress the empty state until the first query resolves; it resets on data-source change and on `onRefresh`.
- The legacy `pageNumber`/`resultsPerPage`/`offset`/`location` shape is dropped from internal state, along with the `areLocationsTheSame` heuristic, replaced by a `queryFilterValues` deep-equality check (`areQueryFiltersEqual`) that handles array filters like `modalitiesInStudy`.
- `defaultDataSourceName` now reads from `useAppConfig()` rather than `window.config` directly.

### 3. WorkListUINext route (`platform/app/src/routes/WorkListUINext/`)

- New `WorkListUINext.tsx` and `useWorkListToolbarActions.tsx`.
- A loading indicator is shown until the first fetch completes and during filter changes.
- The `datasources` query param is preserved across navigation.
- The investigational-use dialog and white-label logo are integrated here.

### 4. Study list state sync (`platform/app/src/hooks/useStudyListStateSync.ts`)

- URL is the source of truth.
- `sessionStorage` is the fallback.
- Component state drives reactivity.
- Restores correctly on filter/sort/pagination combinations after a known regression in earlier commits.

### 5. ui-next StudyList system (`platform/ui-next/src/components/StudyList/`)

- New components: `Layout`, `Table`, `PreviewContainer`, `PreviewContent`, `PreviewPatientSummary`, `PreviewSeriesList`, `PreviewToggleButton`, `SettingsPopover`, `WorkflowsProvider`, `WorkflowMenu`.
- New hook: `useDefaultWorkflow`.
- New utilities and column definitions: `defaultColumns`, `formatStudyDate`, `tokenizeModalities`.

### 6. ui-next primitives

- New `DataTable` compound component (`ColumnHeader`, `FilterRow`, `Pagination`, `Toolbar`, `ViewOptions`, `ActionOverlayCell`).
- New `InputMultiSelect`, `Badge`, and generic `Table` components.
- `DateRange` is expanded with placeholder editing and is sized to match other filters.
- `Thumbnail` is refined with blob-URL revoke on error and a viewport query param for quality.
- Several new `Icons` are added: `OHIFLogoHorizontal`, `PanelRight`, `PatientStudyList`, `SeriesPlaceholder`, `SettingsStudyList`, `SortingNew`/`Ascending`/`Descending`, `CloudSettings`.

### 7. DICOMWeb data source (`extensions/default/src/DicomWebDataSource/index.ts`)

- Adds `thumbnailRequestStrategy` (`bulkDataRetrieve` default, or `fetch`).
- `getGetThumbnailSrc` now accepts a `ThumbnailContext` and falls back to a single QIDO `instances?limit=1` query when a series-level fetch fails.
- Documented in `dicom-web.md`.

### 8. Customization

- `workListUINext.previewSeriesView` (`'all' | 'thumbnails' | 'list'`) is registered via `getCustomizationModule.tsx`.
- The preview is forced to `'list'` when the data source uses `wadors`/`thumbnailDirect` rendering or `bulkDataRetrieve` retrieval.
- Documented in `WorkListUINext.md`.

### 9. Misc

- The data source configuration UI is rebuilt (`DataSourceConfigurationComponent.tsx`).
- An OIDC logout menu item is added.
- The investigational-use dialog is wired into the new worklist.
- i18n entries for `DataTable` are added.

### 10. Removed

- The legacy Cypress suite `OHIFStudyList.spec.js` is removed.
- The obsolete `tests/Worklist.spec.ts` is removed (no longer applicable to the new study list).

### 1. Smoke test

- Run the deploy created for this PR  and confirm `/` loads the new `WorkListUINext` against the default data source.

### 2. State sync

- Apply a sort, page, and a couple of filters (text + modality multi-select + date range).
- Reload the page; the URL state should be restored.
- Open a new tab to `/` (no params); sessionStorage state should restore until tab close.

### 3. `DataSourceWrapper` refetch behavior

- With the network panel open, page forward/back through results and confirm no new query fires.
- Change a filter (modality, date range, MRN) and confirm exactly one new query fires.
- Switch data sources and confirm the wrapper resets and re-queries.
- Set `appConfig.queryLimit` to a small value (e.g. 25) and confirm only that many studies come back.

### 4. Preview panel

- Toggle the panel open/closed.
- Switch between thumbnails and list views.
- Verify the `workListUINext.previewSeriesView` customization values (`'all'`, `'thumbnails'`, `'list'`) hide/lock the toggle as documented.

### 5. Thumbnail strategies

- Smoke-test against a `wadors` source (e.g. `local_orthanc`, `local_dcm4chee`) and confirm the preview is forced to list view.
- Against a `thumbnail`/`rendered` source verify both `bulkDataRetrieve` and `fetch` strategies render thumbnails.
- Confirm that a deliberately-broken series falls back to the QIDO `limit=1` lookup without crashing.

### 6. Workflow buttons & default mode

- Select a study, change default workflow via the settings popover, double-click a row — confirm it launches the active workflow.
- Verify long workflow labels truncate with ellipsis.

### 7. Toolbar/extras

- White-label logo renders when configured.
- OIDC logout menu item appears under an OIDC config.
- Investigational-use dialog appears on first load and dismisses correctly.

### 8. Pagination persistence

- Change rows-per-page, page forward, then sort and filter — the persistence restore should not double-restore or reset to page 1 (the regression fixed in `a292d59f7a`).

### 9. Legacy fallback

- Temporarily swap the commented block in `routes/index.tsx` back to `WorkList` and confirm the legacy view still renders.

### PR

### Code

  etc.)

  additions or removals.

####  System:
    OS: Windows 11 10.0.26200
    Memory: 5.27 GB / 31.68 GB
#### Binaries:
    Node: 20.9.0 - C:\Users\joebo\AppData\Local\fnm_multishells\53780_1778175901087\node.EXE
    npm: 10.1.0 - C:\Users\joebo\AppData\Local\fnm_multishells\53780_1778175901087\npm.CMD
#### Browsers:
    Chrome: 147.0.7727.138
    Edge: Chromium (146.0.3856.84)
    Internet Explorer: 11.0.26100.8115

<h3>Greptile Summary</h3>

This PR replaces the legacy study list with a new `WorkListUINext` system built on `ui-next` components, covering the full worklist surface: table state synced to URL and `sessionStorage`, client-side pagination (single server query per filter change), a preview side panel with bounded-concurrency thumbnail fetching, and a new `thumbnailRequestStrategy` for the DICOMweb data source.

- **`DataSourceWrapper`** is simplified to one query per filter change with a `hasFetchedOnce` gate, proper `finally`-block loading-state reset, and a retry button on failure; `DEFAULT_DATA` is now module-scoped.
- **`useStudyListStateSync`** and `studyListFilterContract` provide a typed, drift-proof URL contract for sorting, pagination, and all filter keys \u2014 including the previously flagged `patientName` mismatch, which is now resolved.
- **`SidePanelPreview`** manages thumbnail fetching with an `AbortController`, a capped worker pool, and explicit `URL.revokeObjectURL` tracking via `ownedBlobUrlsRef`, addressing prior blob-leak concerns.

Safe to merge after the rows-per-page dropdown close bug is fixed; the rest of the change is well-structured with no data-loss paths.

The rows-per-page dropdown in Pagination.tsx calls e.preventDefault() inside onSelect, which per Radix UI docs prevents the menu from auto-closing after every page-size selection. All other findings are non-blocking style or edge-case correctness issues.

platform/ui-next/src/components/DataTable/Pagination.tsx — the rows-per-page dropdown; platform/app/src/routes/DataSourceWrapper.tsx — the pre-existing stale in-flight query on data-source switch (open from prior review round).

| Filename | Overview |
|----------|----------|
| platform/ui-next/src/components/DataTable/Pagination.tsx | New pagination component; e.preventDefault() in DropdownMenuItem.onSelect prevents the rows-per-page dropdown from auto-closing after selection |
| platform/app/src/routes/DataSourceWrapper.tsx | Refactored to client-side pagination with filter-change-only refetch, proper error recovery (finally block, retry button), and hasFetchedOnce gate; stale in-flight query on data-source switch remains a known open issue |
| platform/app/src/routes/WorkList/SidePanelPreview.tsx | New component with bounded-concurrency thumbnail pool, AbortController cleanup, and blob URL revocation via ownedBlobUrlsRef; correctly addresses previously flagged memory-leak concerns |
| platform/app/src/hooks/useStudyListStateSync.ts | New hook syncing sort/page/filter state across URL, sessionStorage, and component state with 200ms debounce; uses centralized studyListFilterContract to avoid key drift |
| platform/app/src/utils/studyListFilterContract.ts | New single source of truth for URL query keys; resolves the previously flagged patientName key mismatch between the URL writer and DataSourceWrapper reader |
| extensions/default/src/DicomWebDataSource/retrieveThumbnail.ts | New thumbnail strategy module with fetch+QIDO-fallback path; thumbnail and rendered bulkDataRetrieve branches are duplicated code |
| platform/app/src/utils/shallowEqualIgnoringArrayOrder.ts | New utility; set-based array comparison incorrectly equates arrays with duplicate elements; not a practical concern for modalities but violates the documented contract |
| platform/app/src/routes/WorkList/WorkList.tsx | Replaced with thin ui-next StudyList wrapper; delegates sorting/pagination/filters to useStudyListStateSync and hands preview management to SidePanelPreview |
| platform/app/src/routes/index.tsx | Default / route now selects WorkList (ui-next) or LegacyWorkList based on workList.variant customization; clean migration path preserved |

</details>

</details>

<details open><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `platform/app/src/routes/DataSourceWrapper.tsx`, line 141-155 ([link](https://github.com/ohif/viewers/blob/89f1aa9bdfd7e5011cc09c9ab9a7fb905b3d72b4/platform/app/src/routes/DataSourceWrapper.tsx#L141-L155)) 

   <a href="#">P1</a> **Loading state stuck after query failure**

   `getData()` sets `isLoading = true` on entry but only calls `setIsLoading(false)` on the happy-path completion. When `dataSource.query.studies.search()` throws, the `.catch()` handler logs the error or shows a modal but never resets `isLoading`. Because `isDataInvalid` is guarded by `!isLoading`, every subsequent location/filter change will be ignored — no further queries fire until the page is reloaded.

   The new `!hasFetchedOnce` condition in `showStudyListLoading` makes this worse: after a failed first fetch both the loading spinner and the empty-state are suppressed indefinitely, leaving users with a blank, unresponsive worklist.

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   Path: platform/app/src/routes/DataSourceWrapper.tsx
   Line: 141-155

   Comment:
   **Loading state stuck after query failure**

   `getData()` sets `isLoading = true` on entry but only calls `setIsLoading(false)` on the happy-path completion. When `dataSource.query.studies.search()` throws, the `.catch()` handler logs the error or shows a modal but never resets `isLoading`. Because `isDataInvalid` is guarded by `!isLoading`, every subsequent location/filter change will be ignored — no further queries fire until the page is reloaded.

   The new `!hasFetchedOnce` condition in `showStudyListLoading` makes this worse: after a failed first fetch both the loading spinner and the empty-state are suppressed indefinitely, leaving users with a blank, unresponsive worklist.

   `````
   </details>

2. `platform/app/src/routes/DataSourceWrapper.tsx`, line 141-156 ([link](https://github.com/ohif/viewers/blob/baab339d6dac695915648dfd9fdcc6b9c03046eb/platform/app/src/routes/DataSourceWrapper.tsx#L141-L156)) 

   <a href="#">P1</a> `getData()` calls `setIsLoading(true)` on entry but only calls `setIsLoading(false)` on the happy path. If `dataSource.query.studies.search()` throws, the `.catch()` error handler shows a modal but never resets the flag. Because `isDataInvalid = !isLoading && filtersChanged`, the wrapper refuses to fire any further query once `isLoading` is stuck at `true`. Combined with `!hasFetchedOnce` remaining `false`, both the loading spinner and the empty state are suppressed indefinitely after a failed first fetch, leaving users with a blank, unresponsive worklist until they reload the page.

   

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   Path: platform/app/src/routes/DataSourceWrapper.tsx
   Line: 141-156

   Comment:
   `getData()` calls `setIsLoading(true)` on entry but only calls `setIsLoading(false)` on the happy path. If `dataSource.query.studies.search()` throws, the `.catch()` error handler shows a modal but never resets the flag. Because `isDataInvalid = !isLoading && filtersChanged`, the wrapper refuses to fire any further query once `isLoading` is stuck at `true`. Combined with `!hasFetchedOnce` remaining `false`, both the loading spinner and the empty state are suppressed indefinitely after a failed first fetch, leaving users with a blank, unresponsive worklist until they reload the page.

   

   `````
   </details>

3. `platform/app/src/routes/DataSourceWrapper.tsx`, line 128-144 ([link](https://github.com/ohif/viewers/blob/1b9d3002d6265aa1c1c84373e0329d839d3b7e88/platform/app/src/routes/DataSourceWrapper.tsx#L128-L144)) 

   <a href="#">P1</a> **Stale in-flight query overwrites new data source results after a switch**

   `dataSourceChangedCallback` calls `refresh()` (which sets `data.queryFilterValues = null`) to ensure a new query fires, but any in-flight `getData()` from the old data source is not cancelled. If the old query completes *after* the new data source has already fetched and populated `data.queryFilterValues`, the old query's `setData({ ..., queryFilterValues: oldQFV })` overwrites the new results. The effect then re-evaluates `filtersChanged = !shallowEqualIgnoringArrayOrder(oldQFV, newQFV)` — since both derive from the same URL, they are equal → `filtersChanged = false` → no corrective refetch fires. The user sees the old data source's studies with no indication that anything is wrong, and no further query is issued.

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   Path: platform/app/src/routes/DataSourceWrapper.tsx
   Line: 128-144

   Comment:
   **Stale in-flight query overwrites new data source results after a switch**

   `dataSourceChangedCallback` calls `refresh()` (which sets `data.queryFilterValues = null`) to ensure a new query fires, but any in-flight `getData()` from the old data source is not cancelled. If the old query completes *after* the new data source has already fetched and populated `data.queryFilterValues`, the old query's `setData({ ..., queryFilterValues: oldQFV })` overwrites the new results. The effect then re-evaluates `filtersChanged = !shallowEqualIgnoringArrayOrder(oldQFV, newQFV)` — since both derive from the same URL, they are equal → `filtersChanged = false` → no corrective refetch fires. The user sees the old data source's studies with no indication that anything is wrong, and no further query is issued.

   `````
   </details>
</details>

`````markdown
Fix the following 3 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 3
platform/ui-next/src/components/DataTable/Pagination.tsx:45-48
Rows-per-page dropdown never auto-closes after selection. Radix UI's documented behavior for `DropdownMenuItem.onSelect` is: calling `event.preventDefault()` prevents the dropdown from auto-closing when the item is selected. Here `e.preventDefault()` is called before `table.setPageSize(size)`, so after clicking a page size the menu stays open indefinitely — users must press Escape or click outside to dismiss it. Remove `e.preventDefault()` to restore the expected close-on-select behavior.

```suggestion
              onSelect={() => {
                table.setPageSize(size);
              }}
```

### Issue 2 of 3
platform/app/src/utils/shallowEqualIgnoringArrayOrder.ts:21-42
Set-based comparison silently mishandles arrays with duplicate values. For `['CT','CT','MR']` vs `['CT','MR','MR']`: both have length 3, and `new Set(...)` on each yields `{CT, MR}` with size 2 — so the function returns `true` even though the arrays differ. In the current modality use-case duplicates are unlikely, but this violates the documented "unordered set" contract for any repeated-element array. Sorting-then-comparing fixes it correctly.

```suggestion
  for (const key of allKeys) {
    const val1 = a[key];
    const val2 = b[key];

    if (Array.isArray(val1) && Array.isArray(val2)) {
      if (val1.length !== val2.length) {
        return false;
      }
      const sorted1 = [...val1].sort();
      const sorted2 = [...val2].sort();
      if (sorted1.some((v, i) => v !== sorted2[i])) {
        return false;
      }
    } else if (val1 !== val2) {
      return false;
    }
  }
```

### Issue 3 of 3
extensions/default/src/DicomWebDataSource/retrieveThumbnail.ts:110-151
The `thumbnail` and `rendered` bulkDataRetrieve branches are character-for-character identical — both build a `bulkDataURI` from the endpoint path and wrap the result in a `Blob`. The only difference is the `thumbnailRendering` string itself, which is already captured by `buildThumbnailEndpointPath`. Consider extracting a shared `buildBulkDataThumbnailSrc` helper to reduce the maintenance surface and eliminate the risk that a future fix in one branch is missed in the other.


`````

</details>

<sub>Reviews (21): Last reviewed commit: ["Refactor the thumbnail retrieval to a se..."](https://github.com/ohif/viewers/commit/c8c480cafab773d032443285f11cdef7602bf399) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31261196)</sub>

## Summary by CodeRabbit

* **New Features**
  * Modern WorkList: Study List table with side-panel preview plus legacy variant option
  * Improved study previews and thumbnail loading with selectable retrieval strategies and better error handling

* **UI Improvements**
  * New DataTable/StudyList UI, responsive columns, badge & multi-select controls, new icons and settings/preview controls

* **Configuration Changes**
  * New thumbnail request strategy option; recommended concurrent thumbnail requests reduced to ~5

* **Documentation**
  * Updated migration and configuration docs for WorkList, paging, and thumbnail behavior

### CODE REVIEW

File: extensions/default/src/DicomWebDataSource/index.ts [hunk 1]
```diff
@@ -47,6 +47,15 @@ export type DicomWebConfig = {
    thumbnail - render using the thumbnail endpoint on wadors using bulkDataURI, passing authentication params  to the url.
     rendered - should use the rendered endpoint instead of the thumbnail endpoint
 */
+  thumbnailRequestStrategy?: 'bulkDataRetrieve' | 'fetch';
```
jbocce, line 51: I introduced this setting for the (potentially) odd case where someone was using the existing `bulkDataRetrieve` reliably with `thumbnailRendering=thumbnail` or `thumbnailRendering=rendered`. However, my investigations have shown that there are the following problems when `bulkDataRetrieve` is used...

orthanc
- does not like the extra Accept headers added by dicomweb-client for the `/rendered` endpoint

google cloud healthcare
- does not like the added `?accept=image/jpeg` 

dcm4chee
- in general works "ok", however with whatever dicomweb-client is doing requires `?accept=image/jpeg` which then conflicts with google
- thumbnails are grainy without the `?viewport=` addition to the url

platform/app/cypress/integration/study-list/OHIFStudyList.spec.js
wayfarer3130, line 1: Can you create an equivalent?  Even better, leave the OHIF study list, and add some sort of customization to allow switching back to the old study list for some length of time - it should get deprecated, not removed.

platform/app/src/routes/DataSourceWrapper.tsx
```diff
@@ -242,6 +221,43 @@ DataSourceWrapper.propTypes = {
 
 export default DataSourceWrapper;
 
+/**
```
wayfarer3130: This should be externalized to a utility, since keeping it here just makes the code longer and harder to follow.

wayfarer3130:
@jbocce - it seems like the new worklist is missing pagination.  You would have to test that with orthanc or dcm4chee with lots of studies, but it appears from the code to be missing.

jbocce:
> @jbocce - it seems like the new worklist is missing pagination. You would have to test that with orthanc or dcm4chee with lots of studies, but it appears from the code to be missing.

@wayfarer3130 there is plenty of code dedicated to pagination. In the UI you can find it in the screenshot below...
image

Here is an AI summary of the pagination code...
- **WorkListUINext (route)**  
  - Calls `useStudyListStateSync()` for table state.  
  - Passes `pagination` and `onPaginationChange` into `StudyList.Table`.  
  - Does not implement paging math itself; it only wires state.

- **`useStudyListStateSync` (app hook)**  
  - Holds TanStack-style pagination: `pageIndex`, `pageSize`.  
  - Initializes from session storage or URL (`pagenumber`, `resultsperpage`).  
  - Debounces updates back into the URL; mirrors full state to session storage.  
  - Serializes pagination into query params (e.g. `pageNumber`, `resultsPerPage` when non-default).

- **`StudyList.Table` (ui-next)**  
  - Thin wrapper around the shared `DataTable`.  
  - Forwards the same pagination props unchanged.

- **`DataTable` (TanStack Table)**  
  - Uses controlled `pagination` plus `onPaginationChange`.  
  - Enables `getPaginationRowModel()` → **client-side** paging over the `data` array.  
  - Body rows come from `getPaginationRowModel().rows` when pagination is active.  
  - On filter or sort change, resets `pageIndex` to `0` unless pagination changed in the same render.

- **`DataTable/Pagination` (UI)**  
  - Shows range text and prev/next; page-size menu (e.g. 25 / 50 / 100).  
  - Total count is `getFilteredRowModel().rows.length` (filtered in-memory set).

- **Data loading vs. table paging**  
  - `DataSourceWrapper` uses `offset: 0` and a single `limit` (batch fetch), not per-page server pagination.  
  - Table pages slice whatever studies were already returned, not a new network request per page.

platform/app/src/routes/WorkListUINext/WorkListUINext.tsx
wayfarer3130, line 1: I would rather call this one WorkListUI and have the old one left in place as WorkListUIDeprecated
I know that is some migration support required, but long term WorkListUINext isn't a good name.

platform/app/src/routes/WorkListUINext/WorkListUINext.tsx
```diff
@@ -0,0 +1,374 @@
...
+            continue;
+          }
+          if (row.thumbnailStatus?.status === PreviewThumbnailStatusState.NotApplicable) {
+            continue;
+          }
+
+          void (async () => {
+            let src: string | null = null;
+            try {
+              const getThumbnailSrc = dataSource?.retrieve?.getGetThumbnailSrc?.(
```
wayfarer3130: This allows an unbounded number of thumbnail fetches.  Some types of studies have a very large number of series, and the thumbnail fetches can take a significant amount of time.   CS3D already has a retrieve size control for thumbnails that will wrap async requests so that only so many of them get initialized at a time.
The real issue with this is if things are still in progress, then going to the view pages can look like it hangs as it waits for thumbnails to load before being able to load metadata and even layout the study.

platform/app/src/routes/WorkListUINext/WorkListUINext.tsx
```diff
@@ -0,0 +1,374 @@
...
+import { thumbnailNoImageModalities } from '@ohif/core/src/utils/thumbnailNoImageModalities';
+import {
+  PreviewThumbnailStatusState,
+  type PreviewThumbnailStatus,
+  type StudyRow,
+} from '@ohif/ui-next';
+
+type Props = withAppTypes & {
+  data: any[];
+  dataTotal: number;
```
wayfarer3130: Unused

platform/app/src/routes/WorkListUINext/WorkListUINext.tsx
wayfarer3130, line 1: In general, this class combines too many different things.  The way to allow this to work more generally is to use component composition so that things like the fetch logic for thumbnails can be provided in the outer wrapper, and that a new instance of the wrapper next could be used with different rendering content.

platform/app/src/routes/WorkListUINext/WorkListUINext.tsx
```diff
@@ -0,0 +1,374 @@
...
+              toolbarRightComponent={
+                !isPreviewOpen ? (
+                  <div className="relative -top-px mt-1 ml-2 flex items-center gap-1">
+                    <StudyListSettingsPopover />
+                    <StudyList.OpenPreviewButton />
+                  </div>
+                ) : undefined
+              }
+            />
+            <StudyList.Preview>
```
wayfarer3130: Hard coding this directly in WorkListUINext with the logic around it means that someone can't replace the side panel preview effectively, and that is true for other components too.

wayfarer3130:
I'd really like to see a few customization examples for simple things like adding columns/re-ordering them, replacing the about box, but also thinking about how to use the worklistuinext with a different layout, but the same fetch/retrieve/thumbnail type logic.

hongxiao-gif (commented):
I have two defects created:
https://github.com/OHIF/Viewers/issues/6029
https://github.com/OHIF/Viewers/issues/6028

I’m not sure if this is expected behavior or by design.
1. When all columns are unselected from the column multiselect list, the study list page displays as empty. A default column set or validation should prevent the page from rendering without any visible columns.
2. The column widths are not resizable.
3. When a row is selected, closing the default workflow mode in the preview panel also disables the default mode in the settings.

Testing environment:
npx envinfo --system --browsers

  System:
    OS: Windows 11 10.0.26200
    Memory: 3.90 GB / 15.67 GB
  Browsers:
    Chrome: 148.0.7778.168
    Edge: Spartan (24.13.1), Chromium (24.13.1), ChromiumDev (24.13.1)

npm notice
npm notice New minor version of npm available! 11.8.0 -> 11.14.1
npm notice Changelog: https://github.com/npm/cli/releases/tag/v11.14.1
npm notice To update run: npm install -g npm@11.14.1
npm notice

https://deploy-preview-6005--ohif-dev.netlify.app/

platform/app/cypress/integration/study-list/OHIFStudyList.spec.js
jbocce: We really are no longer creating cypress tests, so there is no point maintaining this test. Furthermore, I am not sure how easy it is in the cypress environment to set or switch a config variable on the fly.

platform/app/src/routes/WorkListUINext/WorkListUINext.tsx
jbocce: I am not crazy about renaming the old Worklist for various reasons - the most important of which is for anyone that might be referencing it somewhere. So perhaps maybe we can simply rename WorklistUINext to something you like better?

platform/app/src/routes/WorkListUINext/WorkListUINext.tsx
jbocce: I just committed and pushed some changes to address this.

platform/app/src/routes/WorkListUINext/WorkListUINext.tsx
jbocce: Ok after thinking it over and considering the other good suggestions you have made in this PR, I think I will rename the old one but use the name `LegacyWorklist` - `legacy` is something we've used often in the past and I think there are still some components with the legacy prefix. This works well with the customization to revert back to the old one (i.e. `worklist.variant:'legacy' vs `worklist.variant:'default')

jbocce:
> I'd really like to see a few customization examples for simple things like adding columns/re-ordering them, replacing the about box, but also thinking about how to use the worklistuinext with a different layout, but the same fetch/retrieve/thumbnail type logic.

@wayfarer3130 I believe I have addressed most, if not all, of the customization improvement you have suggested. Note that when we first set out to replace the previous WorkList we agreed to minimize the customization changes. So I think what we have here thus far is more than sufficient. Let me know what you think.

platform/app/src/routes/DataSourceWrapper.tsx
jbocce: I refactored this. Please have a look and let me know if it was what you were looking for.

platform/app/src/hooks/useStudyListStateSync.ts
```diff
@@ -0,0 +1,217 @@
...
+  state.filters.forEach(filter => {
+    if (filter.id === COLUMN_IDS.MODALITIES && Array.isArray(filter.value)) {
+      query.modalities = filter.value.join(',');
+    } else if (filter.id === COLUMN_IDS.STUDY_DATE_TIME) {
+      const dateRange = filter.value as StudyDateRangeFilter | undefined;
+      if (dateRange?.startDate) {
+        query.startDate = dateRange.startDate;
+      }
+      if (dateRange?.endDate) {
+        query.endDate = dateRange.endDate;
```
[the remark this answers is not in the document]
jbocce: @greptileai I believe this is now addressed.

platform/app/src/routes/DataSourceWrapper.tsx
```diff
@@ -29,32 +18,31 @@ const areLocationsTheSame = (location0, location1) => {
...
-  const navigate = useNavigate();
   const { children: LayoutTemplate, ...rest } = props;
   const params = useParams();
   const location = useLocation();
   const lowerCaseSearchParams = useSearchParams({ lowerCaseKeys: true });
   const query = useSearchParams();
+  const [appConfig] = useAppConfig();
+
   // Route props --> studies.mapParams
   // mapParams --> studies.search
```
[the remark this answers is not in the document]
jbocce: @greptileai, I believe this has been addressed now.

jbocce:
@greptileai I believe even your comments outside diff have been addressed.

dan-rukas (approved):

extensions/default/src/DicomWebDataSource/index.ts
```diff
@@ -261,22 +282,40 @@ function createDicomWebApi(dicomWebConfig: DicomWebConfig, servicesManager) {
        *    or is already retrieved, or a promise to a URL for such use if a BulkDataURI
        */
 
-      getGetThumbnailSrc: function (instance, imageId) {
-        if (dicomWebConfig.thumbnailRendering === 'wadors') {
+      getGetThumbnailSrc: function (thumbnailContext: ThumbnailContext, imageId) {
+        const thumbnailRendering = dicomWebConfig.thumbnailRendering;
```
sedghi: can we refactor this out, it is adding noise to data source

extensions/default/src/DicomWebDataSource/index.ts
```diff
@@ -741,4 +823,164 @@ function retrieveBulkData(value, options = {}) {
   });
 }
 
+function buildThumbnailEndpointPath(
```
sedghi: same

platform/app/public/config/default.js
```diff
@@ -20,7 +20,7 @@ window.config = {
   allowMultiSelectExport: false,
   maxNumRequests: {
     interaction: 100,
-    thumbnail: 75,
+    thumbnail: 5,
```
sedghi, line 23: this should get communicated as breaking change in the docs, same as below

jbocce: I would say these are NOT breaking changes. Those values should never have been that high for thumbnail requests.

platform/app/src/routes/WorkList/SidePanelPreview.tsx
```diff
@@ -0,0 +1,232 @@
...
+    'workList.previewSeriesView'
+  );
+  const configuredSeriesView: PreviewSeriesView = ALLOWED_PREVIEW_SERIES_VIEWS.includes(
+    customizationSeriesView as PreviewSeriesView
+  )
+    ? (customizationSeriesView as PreviewSeriesView)
+    : 'all';
+  const seriesView: PreviewSeriesView = forceListView ? 'list' : configuredSeriesView;
+
+  useEffect(() => {
```
sedghi: this is massive useEffect, it should be its own useSeriesFetch hook

platform/app/src/routes/DataSourceWrapper.tsx
sedghi: same here this useEffect better be refactored as a hook

platform/app/src/utils/shallowEqualIgnoringArrayOrder.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Limitation: this is a *shallow* comparison. Nested objects are compared
+ * by reference; this function does not recurse. It is intended for flat
+ * records whose values are primitives or arrays of primitives.
+ *
+ * @param {object} a - First object
+ * @param {object} b - Second object
+ * @returns {boolean} True if the two are equal under the rules above.
+ */
+export function shallowEqualIgnoringArrayOrder(a, b): boolean {
```
sedghi, line 14: can we add tests for this

platform/ui-next/src/components/DateRange/DateRange.tsx
```diff
@@ -22,6 +31,11 @@ export function DatePickerWithRange({
   startDate,
   endDate,
   onChange,
+  inputClassName,
```
sedghi: lots of props mean we are not following shadcn composition correctly, can we rethink this with composition instead of props

platform/ui-next/src/components/StudyList/utils/formatStudyDate.ts
```diff
@@ -0,0 +1,46 @@
+import moment from 'moment';
+
+/**
+ * Formats a DICOM study date and time for display.
+ *
+ * @param date - Raw date string (YYYYMMDD or YYYY.MM.DD format)
+ * @param time - Raw time string (HH, HHmm, HHmmss, or HHmmss.SSS format)
+ * @returns Formatted date string (DD-MMM-YYYY HH:mm) or empty string if invalid
+ */
+export function formatStudyDate(date?: string, time?: string): string {
```
sedghi: this feels like it should have been implemented elsewhere already, please make sure we are not adding a new one

platform/ui-next/package.json
```diff
@@ -53,6 +53,7 @@
     "cmdk": "1.1.1",
     "date-fns": "3.6.0",
     "framer-motion": "6.2.4",
+    "moment": "2.30.1",
```
sedghi: i really don't want to add a new time-related dependency, can we use the new Temporal api that is widely supported 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal 

see also other blog posts on how to migrate

https://www.smashingmagazine.com/2026/03/moving-from-moment-to-temporal-api/

sedghi (changes requested):
please see my comments

platform/app/public/config/default.js
sedghi: maybe don't label as breaking change but communicate we have changed these

extensions/default/src/DicomWebDataSource/index.ts
wayfarer3130: What is your suggestion for what part remains in datasource?  The problem is that different datasources end up with different methods/capabilities here.  I'm ok with it moving out, but would like to know what pieces go where and how the choices are made.

platform/app/src/routes/WorkList/SidePanelPreview.tsx
wayfarer3130: That will also make it effectively useable by other components.

platform/docs/docs/platform/services/customization-service/sampleCustomizations.tsx
```diff
@@ -356,6 +356,205 @@ window.config = {
...
+      </>
+    ),
+    default: '(defaults) => defaults',
+    configuration: `
+window.config = {
+  // rest of window config
+  customizationService: [
+    {
+      'workList.columns': {
+        // Hide the MRN column and move Instances to the front.
```
wayfarer3130: Can you include an example of adding a column?

platform/docs/docs/platform/services/customization-service/WorkList.md
```diff
@@ -0,0 +1,105 @@
...
+
+:::tip
+
+The customizations below — `workList.columns`, `workList.renderPreviewContent`, and `workList.settingsMenuItems` — accept functions whose return values frequently need access to React hooks, the services manager, or the commands manager (e.g. to open modals, navigate, run commands, or build translated labels). They can be set from `window.config`, but they're generally easier to author in a custom extension's `getCustomizationModule`, where the services manager is in scope and components can use hooks normally. Plain config is best suited to simple tweaks like reordering, removing, or inserting items that only need static handlers.
+
+:::
+
+## `workList.columns`
+
+Builds the column set for the WorkList table. The customization is a function with the same shape as `StudyList.defaultColumns` — it receives the default `ColumnDef[]` and must return a `ColumnDef[]`. Use it to reorder, hide, or insert columns without rewriting the defaults.
```
wayfarer3130: Can this be done as a fixed list rather than a function returning a new column def?  That could then be modified directly by the invariant $ functions using straight properties, at least as long as there was a basic column type that could display a column.

platform/ui-next/src/components/StudyList/columns/defaultColumns.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+// Text filter column IDs (columns that use text-based filtering)
+export const TEXT_FILTER_COLUMN_IDS = [
+  COLUMN_IDS.PATIENT,
+  COLUMN_IDS.MRN,
+  COLUMN_IDS.ACCESSION,
+  COLUMN_IDS.DESCRIPTION,
+] as const;
+
+export function defaultColumns(): ColumnDef<StudyRow, unknown>[] {
```
wayfarer3130: Is there a reason this has to be a function?  It looks more or less like a constant which could be re-used/re-created.
Ideally the shadcn component composition could be used to create a completely different layout by including alternate header functions/render body functions, or the default columns could be applied.

extensions/default/src/DicomWebDataSource/index.ts
jbocce: I refactored all of the thumbnail retrieval logic/code to a file `retrieveThumbnail.ts` that lives next to the `index.ts` file. I hope you both find this adequate.

extensions/default/src/DicomWebDataSource/index.ts
jbocce: see https://github.com/OHIF/Viewers/pull/6005#discussion_r3306649006

platform/app/public/config/default.js
jbocce: done

platform/ui-next/package.json
jbocce: moment is already a dependency in core, ui and devDependencies. I am simply adding it also to ui-next. So it is not new at all.

platform/ui-next/src/components/StudyList/utils/formatStudyDate.ts
jbocce: Yes thanks for pointing this out. This might have been something leftover from the initial prototype that AI built.

As much as this was a good catch, it was also painful to investigate 🤣 as it seems there are way too many places already formatting and parsing DICOM dates - including lots of utilities directly in the LegacyWorklist file. So let me try to best describe the approach I did to resolve your comment.

The date formatting already lived in `platform\ui-next\src\utils\formatDICOMDate.ts`, and the time/sort logic was duplicated between the old formatStudyDate.ts and LegacyWorkList's inline moment block.  Commit a43a4c1cb90249bf16617b0e2f41da8ca35b96a5 consolidates all of it:

- Deleted `formatStudyDate` — its date+time composition is now inline in the study-date column's accessorFn, built from the shared utils formatDICOMDate and formatDICOMTime.
- Date now goes through the existing formatDICOMDate (ui-next) in both the new and legacy worklists.
Time uses a new sibling formatDICOMTime (there was no shared DICOM-time formatter — platform\core\src\utils\formatDate.js's core counterpart (formatTime) only parses HH:mm:ss, not raw DICOM time).
- Sort — moved parseStudyDateTimestamp next to the other format utils and exported it; LegacyWorkList now uses it instead of its private _sortStringDates, which let me drop the file's last moment usage entirely.
- While here, both formatters were switched to a (value, options) signature (with a migration note for `formatDICOMDate`), and I added unit tests for all three utils.
- Net result: one shared formatting/sort path for both worklists, (same-date ties just sort stably now instead of arbitrarily).

platform/app/src/utils/shallowEqualIgnoringArrayOrder.ts
jbocce: done

platform/ui-next/src/components/DateRange/DateRange.tsx
jbocce: Noted. I think this is something @dan-rukas and I will look at.

dan-rukas: I had a [PR that updated to the latest shad/cn date picker components](https://github.com/OHIF/Viewers/pull/5660) (previously for the Dec release). 

That was meant to replace these small changes. Instead, I integrated the full changes from the PR with the latest components and tested them in both the new and old study lists. So, the custom props have also been removed. Short overview below.

**What changed:**
- **Calendar.tsx** — Full rewrite for react-day-picker v9. Uses `DayButton` composition, `getDefaultClassNames()`, and the `data-slot` transparency pattern for popover nesting.
- **DateRange.tsx** — Updated to use the new Calendar API (`autoFocus`, `captionLayout`, `startMonth`/`endMonth`). Compact defaults baked in (`h-7`, shorter "Start"/"End" placeholders, `text-primary` icon). No custom props needed.
- **Popover.tsx** — Added `data-slot="popover-content"` to match current shadcn convention, so Calendar goes transparent inside popovers instead of painting `bg-background` over `bg-popover`.
- **package.json** — `react-day-picker` 8 → 9, `date-fns` 3 → 4.
- **Table.tsx** — Removed the old prop overrides from the DatePickerWithRange call site.
- **InputDateRange.tsx** (legacy ui) — Uses Tailwind descendant selectors to match legacy input sizing without touching ui-next components.

platform/ui-next/src/components/StudyList/columns/defaultColumns.tsx
jbocce: @wayfarer3130  regarding your three comments regarding `defaultColumn` and column customization, I have pushed a set of changes that address all of them (commit hash 743caed7757f5c5a070edd3014236155749302cc). Please have a look in particular to the changes in:
- `platform\core\src\services\CustomizationService\CustomizationService.ts` to allow for React elements that carry a `$$typeof` brand to be udpated/processed
- `extensions\default\src\DicomWebDataSource\qido.js` - added referring physician to the results so that if someone tries the sample in the docs for adding a column it will actually work
- all the doc changes; hopefully it is clear enough how to customize the columns

sedghi (approved):

PR #6056, opened 2026-06-03
Feat/clean remove panel

theoc0702:
### Context

Currently, there is no reliable, native way to remove a single panel by its ID from the `PanelService` and guarantee a proper React UI re-render. Developers trying to dynamically hide a panel during a stage change or layout update had to manually fetch the panels, filter them, and attempt to force a reset. This often led to race conditions or silent failures where the panel's state was updated in memory but remained visible on screen.

### Changes & Results

- **Added `removePanel(panelId: string, options = {})`** to `platform/core/src/services/PanelService/PanelService.tsx`.

**What are the effects of this change?**
- **Before:** Removing a panel required complex workarounds outside the service, often failing to trigger the UI layout redraw.
- **After:** The new method retrieves panels across all positions (`Left`, `Right`, `Bottom`), filters out the target `panelId`, and passes the clean lists directly to `this.setPanels()`. Because `setPanels` natively handles `this.reset()` under the hood, the UI is instantly and cleanly updated.

### Testing

1. Open a study in the viewer.
2. Ensure a panel (e.g., a custom extension panel or a default one) is active and visible on the right or left.
3. Call `panelService.removePanel('your-panel-id')` via a custom command (e.g., during an `onLayoutChange` event in a Hanging Protocol).
4. Verify the targeted panel is immediately unmounted and removed from the UI, while the rest of the layout remains stable.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 / macOS 14 
- [x] Node version: 18.16.1 
- [x] Browser: Chrome 125.0 

<h3>Greptile Summary</h3>

This PR adds a `removePanel` method to `PanelService` so callers can remove a single panel by ID without manually reconstructing the panel list. However, the PR also includes a large set of private `@pixilib` org-specific configurations across build tooling and default app config that are unrelated to the stated goal and would break builds for any public contributor.

- **`PanelService.removePanel`**: The new method has two functional bugs — a `position` parameter that is declared but silently ignored, and a `Bottom` panel wipe-out caused by routing removal through `setPanels`/`reset()` without restoring the Bottom position.
- **Org-specific config leak**: `pluginConfig.json`, both webpack configs, `default.js`, and `tailwind.config.js` all reference private `@pixilib` packages that are not published and will cause immediate build failures for anyone without access to that private registry.
- **Production source map**: `webpack.base.js` disables production source maps with a commented-out original value, indicating a local debug tweak that was committed by mistake.

<h3>Confidence Score: 1/5</h3>

Not safe to merge — private org packages embedded in shared build and default config files will break builds for all public contributors, and the core feature itself has two logic errors.

The build configuration files (`pluginConfig.json`, webpack configs, `default.js`) reference unpublished private `@pixilib` packages. Anyone cloning the public OHIF repo and attempting a build would hit immediate resolution failures. The `removePanel` implementation also has a dead `position` parameter and silently drops `Bottom` panels on every call. The production source-map disable in `webpack.base.js` appears to be an accidental local commit.

`platform/app/pluginConfig.json`, `platform/app/public/config/default.js`, `.webpack/webpack.base.js`, and `platform/app/.webpack/webpack.pwa.js` all need the `@pixilib` entries removed before this can land. `PanelService.tsx` needs the `position` parameter usage and `Bottom` panel handling fixed.

| Filename | Overview |
|----------|----------|
| platform/core/src/services/PanelService/PanelService.tsx | Adds `removePanel` method — the `position` parameter is declared but never used, and calling the method silently wipes all `Bottom` panels because `setPanels`/`reset()` is used without restoring the Bottom position. |
| platform/app/pluginConfig.json | Registers eight `@pixilib` extensions and two `@pixilib` modes — private org packages that are unpublished and will break builds for any downstream OHIF developer. |
| .webpack/webpack.base.js | Disables production source maps (`devtool: false`) and adds a private `pixilib` node_modules resolver path — both appear to be local development leftovers. |
| platform/app/.webpack/webpack.pwa.js | Adds eight private `pixilib` module resolver paths that don't exist in the public repository and will cause resolution failures for other contributors. |
| platform/app/public/config/default.js | Changes `customizationService` to reference private `@pixilib` modules and adds `disableConfirmationPrompts: true` — org-specific overrides committed to the shared default config. |
| platform/app/tailwind.config.js | Adds a Tailwind content glob for `@pixilib` packages and removes the `@type` JSDoc comment — minor org-specific addition that shouldn't be in upstream. |

</details>

</details>

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `platform/app/pluginConfig.json`, line 68-141 ([link](https://github.com/ohif/viewers/blob/5731401ced700e8a873c664a74e723f32519ae20/platform/app/pluginConfig.json#L68-L141)) 

   <a href="#">P0</a> **Private `@pixilib` packages committed to the upstream OHIF config**

   This file is the canonical plugin registry for the public OHIF Viewers project. Adding eight `@pixilib/*` extensions and two `@pixilib/*` modes (all at `"version": "0.0.1"`) will cause every downstream developer who tries to install or build OHIF to fail, because these packages are not published to any public registry. The same concern applies to the corresponding paths added in `.webpack/webpack.base.js`, `platform/app/.webpack/webpack.pwa.js`, `platform/app/tailwind.config.js`, and `platform/app/public/config/default.js`. These org-internal entries should be kept out of the upstream repository.

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   Path: platform/app/pluginConfig.json
   Line: 68-141

   Comment:
   **Private `@pixilib` packages committed to the upstream OHIF config**

   This file is the canonical plugin registry for the public OHIF Viewers project. Adding eight `@pixilib/*` extensions and two `@pixilib/*` modes (all at `"version": "0.0.1"`) will cause every downstream developer who tries to install or build OHIF to fail, because these packages are not published to any public registry. The same concern applies to the corresponding paths added in `.webpack/webpack.base.js`, `platform/app/.webpack/webpack.pwa.js`, `platform/app/tailwind.config.js`, and `platform/app/public/config/default.js`. These org-internal entries should be kept out of the upstream repository.

   `````
   </details>

</details>

`````markdown
Fix the following 4 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 4
platform/core/src/services/PanelService/PanelService.tsx:146-158
**`position` parameter is declared but never used**

The method signature declares a `position: PanelPosition` argument, but the body ignores it entirely — it always reads and filters both `Left` and `Right` regardless. Any caller who passes a specific position expecting targeted removal will be confused by this silent discard. Either remove the parameter (matching the PR description's `removePanel(panelId, options = {})` signature) or use it to scope the removal to the specified position.

### Issue 2 of 4
platform/core/src/services/PanelService/PanelService.tsx:146-158
**Bottom panels are silently wiped on every `removePanel` call**

`setPanels` internally calls `this.reset()`, which clears all positions — including `Bottom`. The `removePanel` implementation only reconstructs `Left` and `Right` panels before passing them back to `setPanels`; `Bottom` panels are never read and therefore never restored. Any panel currently registered at `PanelPosition.Bottom` will be irreversibly dropped as a side-effect of removing an unrelated Left or Right panel. The fix is to include Bottom in the filter-and-restore logic (or avoid `setPanels`/`reset()` altogether and instead mutate only the target position directly).

### Issue 3 of 4
platform/app/pluginConfig.json:68-141
**Private `@pixilib` packages committed to the upstream OHIF config**

This file is the canonical plugin registry for the public OHIF Viewers project. Adding eight `@pixilib/*` extensions and two `@pixilib/*` modes (all at `"version": "0.0.1"`) will cause every downstream developer who tries to install or build OHIF to fail, because these packages are not published to any public registry. The same concern applies to the corresponding paths added in `.webpack/webpack.base.js`, `platform/app/.webpack/webpack.pwa.js`, `platform/app/tailwind.config.js`, and `platform/app/public/config/default.js`. These org-internal entries should be kept out of the upstream repository.

### Issue 4 of 4
.webpack/webpack.base.js:68
Disabling source maps for production builds (`false`) removes the ability to debug minified production errors. The commented-out value suggests this was a temporary local tweak that was accidentally committed. Restore the original value.

```suggestion
    devtool: isProdBuild ? 'source-map' : 'cheap-module-source-map',
```


`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(core): add removePanel method to Pa..."](https://github.com/ohif/viewers/commit/5731401ced700e8a873c664a74e723f32519ae20) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35342247)</sub>

> Greptile also left **3 inline comments** on this PR.

## Summary by CodeRabbit

* **New Features**
  * Added support for removing panels from the viewer layout.
* **Bug Fixes**
  * Improved DICOMweb thumbnail rendering by switching to the thumbnail pipeline and using a fetch-based request strategy.
  * Enabled wildcard matching for DICOMweb request handling.
  * Adjusted default thumbnail request throttling to better balance thumbnail fetching with other requests.

### CODE REVIEW

File: platform/app/.webpack/webpack.pwa.js [hunk 1]
```diff
@@ -85,6 +80,14 @@ module.exports = (env, argv) => {
         // Hoisted Yarn Workspace Modules
         path.resolve(__dirname, '../../../node_modules'),
         SRC_DIR,
+        path.resolve(__dirname, '../pixilib/modes/gaelo-mode/node_modules'),
```
wayfarer3130: This is a custom change for your paths/setup - please don't push changes like this not for merge into master.
Also, the right way to add dependencies like this is to add them to pluginConfig.json as references so that you don't need to update webpack.pwa.js

platform/core/src/services/PanelService/PanelService.tsx
```diff
@@ -143,6 +143,20 @@ export default class PanelService extends PubSubService {
     this._broadcastEvent(EVENTS.PANELS_CHANGED, { position, options });
   }
 
+  public removePanel(panelId: string, position: PanelPosition, options): void {
```
wayfarer3130: Address the greptile comment about panelPosition - suggest just removing it.

platform/core/src/services/PanelService/PanelService.tsx
```diff
@@ -143,6 +143,20 @@ export default class PanelService extends PubSubService {
...
+
+    const leftPanels = this._panelsGroups.get(PanelPosition.Left)?.map((p) => p.id) ?? [];
+    const rightPanels = this._panelsGroups.get(PanelPosition.Right)?.map((p) => p.id) ?? [];
+
+    const newLeftPanels = leftPanels.filter((id) => id !== panelId);
+    const newRightPanels = rightPanels.filter((id) => id !== panelId);
+
+    this.setPanels({
```
wayfarer3130, line 154: Probably need to fix setPanels so that it only updates panels actually provided.  That will need an update to the reset to work.

PR #6120, opened 2026-07-05
fix(dicom-json): correct frame indexing and clarify metadata requirements

driavysinus:
### Context

Related to #3093.

The required metadata FAQ already documents the broad set of fields OHIF needs, but DICOM JSON users still have to decide which metadata to keep when reducing JSON payload size. This update clarifies the practical baseline metadata and modality-specific fields needed for display set creation, rendering, sorting, cine, ultrasound, PT SUV correction, SEG, RTSTRUCT, SR, PDF, and video handling.

### Changes & Results

- Fixed the DICOM JSON generator to emit zero-based frame query parameters for multi-frame objects.
- Reworked the required metadata FAQ into baseline image metadata, rendering metadata, modality/feature metadata, and DICOM JSON specific notes.
- Clarified `PlanarConfiguration` for color/RGB images, `NumberOfFrames` for multi-frame objects, and zero-based `frame` query parameters.
- Expanded PT SUV correction notes to include the required radiopharmaceutical sequence fields used by OHIF's PT metadata helper.
- Added a short DICOM JSON guide note about metadata pruning and multi-frame frame URLs.
- Updated both current docs and the `version-3.11` docs copy.

### Testing

- `git diff --check`
- `corepack pnpm dlx prettier@3.6.2 --check platform/docs/docs/faq/technical.md platform/docs/docs/configuration/dataSources/dicom-json.md platform/docs/versioned_docs/version-3.11/faq/technical.md platform/docs/versioned_docs/version-3.11/configuration/dataSources/dicom-json.md`
- Verified the current and versioned FAQ copies match after the update.

Full docs build was not run because the monorepo dependencies were not installed locally.

#### PR

- [x] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [x] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [x] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: v24.14.0
- [x] Browser: Not run; docs-only change

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

## Documentation

* Updated DICOM JSON guidance on required metadata for display set creation, rendering, sorting, and modality-specific tools.
[...]

PR #6121, opened 2026-07-05
fix(dicom-json): handle multiframe image ids

### LINKED ISSUE #5674: [Bug] Multiframe DICOM do not work with DICOMJSON datasource
### Describe the Bug

There is a clear bug in `extensions/default/src/getSopClassHandlerModule.js`  the `imageSet.setAttributes` call  must be called before `dataSource.getImageIdsForDisplaySet` because it expects `SeriesInstanceUID` set at `imageSet` level! This change is in a commit from 7 months ago (lines 91-100).

After fixing this at least the code moved ahead. However it got stuck again subsequently. At which point we gave up and have to move back to 3.9 for now as too much time has been spent debugging what was a just a simple startup and load sequence.

Basically multi-frame logic **seems** to be broken completely in v3.110 and partially even in v3.9.0. 

In v3.11.0 If we remove `NumberOfFrames` from the json it shows a single frame successfully. However as soon as we have a count it fails for the same json and dicom file. 

In 3.9.0 we were able to remove `NumberOfFrames` but provide equal number of instances in the instances array which worked (though cine player crashes). 

In v3.11.0 neither approach works viz having `NumberOfFrames` with 1 instance, or with N instances or not having `NumberOfFrames` but having N instances. None of this works.

### Steps to Reproduce

the app-config.js (default.js) has standard extensions and a single datasource of type dicomjson.

Here is the partial relevant code. The remaining is just white labeling.

`window.config = {
  // 1. Core Deployment Settings
  routerBasename: '/dv',
  showStudyList: false, // We launch directly via URL, so no study list needed.

  // 2. Extensions (CRITICAL for v3.9.0)
  // These provide the underlying functionality (DICOM parsing, Tools, etc.)
  extensions: [
    '@ohif/extension-default',
    '@ohif/extension-cornerstone',
    '@ohif/extension-measurement-tracking',
    '@ohif/extension-cornerstone-dicom-sr',
    '@ohif/extension-dicom-pdf',
    '@ohif/extension-dicom-video'
  ],

  // // 3. Modes (CRITICAL for v3.9.0)
  // // These define the actual Viewer Layout (Radiology, PET/CT, etc.)
  modes: ['@ohif/mode-longitudinal'],

  // FIX: Leave these empty. The build system fills them automatically.
  // extensions: [],
  // modes: [],

  // 4. Data Source: JSON File (Serverless)
  dataSources: [
    {
      namespace: '@ohif/extension-default.dataSourcesModule.dicomjson',
      sourceName: 'dicomjson',
      configuration: {
        friendlyName: 'JSON File Source',
        name: 'json',
      },
    },
  ],

  // This tells OHIF to use the 'dicomjson' source by default
  defaultDataSourceName: 'dicomjson',`

Create a DICOMJSON file.

We have tried many versions of this file, but reproduction purposes we are just adding a sample where we have reduced the frames to just 2

[series_3.json](https://github.com/user-attachments/files/24285175/series_3.json)

Note that even though our XA file is multiframe, and we would expect OHIF to handle it from a single "instance" in v3.9.0 it only worked when we had an instance for each frame with a &frame=N suffix, and we are okay with that, we generated the file accordingly with minimum metatdata per instance.

Since that did not work with v3.11.0 at all, we added much more metadata at instance leve.

### The current behavior

An exception is thrown and OHIF crashes. 

### The expected behavior

All series should load.
Cine player should play them. (in v3.9.0 in spite of loading all the series in the study with correct frame counts, cine player crashes)
All frames should get displayed when scrolling.

### System Information

$ node --version
v20.19.5

$ yarn -v
1.22.22

Windows 11 24H2
16GB RAM
512 GB SSD

driavysinus:
### Context

Fixes #5674.

DICOM JSON multiframe handling currently conflates two different JSON shapes:

- a single DICOM instance with `NumberOfFrames > 1`
- multiple DICOM JSON entries that point to individual frames of the same `SOPInstanceUID`

The issue sample uses the second shape: the series has `NumberOfFrames: 2`, but each item in `series.instances` already points at a specific frame URL. During ingestion, the series-level `NumberOfFrames` was copied onto each instance, so each per-frame entry could be treated as its own multiframe object. Later, `getImageIdsForDisplaySet` grouped instances by `SOPInstanceUID` and could reuse the first URL for non-multiframe per-frame entries.

### Changes & Results

- Added DICOM JSON-specific helper functions for instance metadata creation and display set image ID expansion.
- Avoid copying a series-level `NumberOfFrames` onto every instance when the JSON already contains multiple frame entries.
- Preserve `NumberOfFrames` for the single-instance multiframe case.
- Keep the current per-frame URL for per-frame DICOM JSON entries that share a `SOPInstanceUID`.
- Append `&frame=N` only when a single multiframe instance URL needs expansion and the URL does not already specify a frame.
- Added unit coverage for the issue sample shape and the single-instance multiframe shape.

### Testing

- `corepack pnpm install --frozen-lockfile`
- `corepack pnpm exec prettier --check extensions/default/src/DicomJSONDataSource/index.js extensions/default/src/DicomJSONDataSource/index.test.js`
- `git diff --check`
- `corepack pnpm exec jest extensions/default/src/DicomJSONDataSource/index.test.js --runInBand --no-coverage`

The targeted Jest suite passes with 6 tests. Manual viewer validation against a real reachable XA DICOM object was not performed; the issue attachment uses sample/redacted URLs that are not directly retrievable.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: v24.14.0
- [x] Browser: Not run; datasource unit test only

## Summary by CodeRabbit

* **New Features**
  * Improved DICOM JSON image handling with accurate per-frame image IDs for multi-frame instances.
  * Added consistent image ID generation for display sets and individual instances.
  * Added display-set UID extraction and enhanced instance metadata normalization.
* **Bug Fixes**
  * Corrected frame query handling, including root URL configurations.
  * Improved fallback behavior for missing URLs and frame metadata.
* **Tests**
  * Expanded coverage for image-ID generation, display-set handling, and metadata behavior.


driavysinus:
Update: PR_CHECKS, Netlify, and CodeRabbit are green on the latest commit, with no actionable comments in the latest CodeRabbit pass.

@wayfarer3130, when you have a chance, could you take a human review? This is the runtime multiframe image-ID half; #6120 contains the related generator and documentation correction.

PR #6129, opened 2026-07-08
test(measurements): add DOM assertions for measurement annotation text

GhadeerAlbattarni:
### Context
Adds DOM-based annotation text assertions for measurement tools, following the pattern established in `ArrowAnnotate.spec.ts`.

### Changes & Results

- Add `tests/utils/expectAnnotationText.ts` with `expectAnnotationStatsText` and `measurementTextFormatters` to assert measurement text from `cachedStats` in the panel and viewport SVG
- Update measurement specs: Length, Bidirectional, Angle, CobbAngle, Circle, Ellipse, Rectangle, Probe
- Refactor `ArrowAnnotate.spec.ts`
- Export new helpers from `tests/utils/index.ts`

**Implementation notes**
- Panel and SVG formatting can differ per tool (e.g. area panel line vs `Area:` SVG prefix; CobbAngle panel uses `roundNumber`, SVG uses `toFixed(2)`).
- `assertStats` in each spec adds numeric/unit sanity checks on raw `cachedStats`.
- `roundNumber` is copied rather than imported because the root Playwright E2E harness does not resolve `@ohif/core` workspace imports in Node.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.22.0

## Summary by CodeRabbit

* **Tests**
  * Expanded measurement tool coverage across angle, bidirectional, circle, Cobb angle, ellipse, length, probe, rectangle, and arrow annotations.
  * Added viewport-specific screenshot validation and stronger checks for displayed values, units, rounding, and numeric formatting in the measurements panel and viewport.
  * Introduced shared validation utilities for consistently checking annotation labels and measurement statistics.


tests/Ellipse.spec.ts
```diff
@@ -25,4 +34,35 @@ test('should display the ellipse tool', async ({
...
+    formatSvgLines: [
+      measurementTextFormatters.areaSvgLine,
+      measurementTextFormatters.meanSvgLine,
+      measurementTextFormatters.maxLine,
+      measurementTextFormatters.minSvgLine,
+      measurementTextFormatters.stdDevSvgLine,
+    ],
+    assertStats: stats => {
+      expect(stats.areaUnit).toBe('mm²');
+      expect(Math.round(stats.area as number)).toBe(16778);
```
GhadeerAlbattarni: @jbocce I found the value for this area is different from the value that is in the current screenshot. If this is the right one, then we might need to update the screenshot as well.

tests/Bidirectional.spec.ts
```diff
@@ -25,4 +34,26 @@ test('should display the bidirectional tool', async ({
...
+      measurementTextFormatters.bidirectionalWidthLine,
+    ],
+    formatSvgLines: [
+      measurementTextFormatters.bidirectionalLengthLine,
+      measurementTextFormatters.bidirectionalWidthLine,
+    ],
+    assertStats: stats => {
+      expect(stats.unit).toBe('mm');
+      expect(Math.round(stats.length as number)).toBe(195);
+      expect(Math.round(stats.width as number)).toBe(130);
```
GhadeerAlbattarni: @jbocce I do think It's good to pass hardcoded numbers to assert the stats as well. If you agree with that I can update the rest of the tests in this PR to do the same.

jbocce: Yes that is perfect as along as we are confident the values are to be expected.

tests/Ellipse.spec.ts
jbocce: Well the screenshot will eventually no longer contain the text, so yes we need a new screenshot but that will be done whenever we remove the text from the screenshots.

tests/utils/expectAnnotationText.ts
```diff
@@ -0,0 +1,277 @@
...
+ *   await rightPanelPageObject.measurementsPanel.select();
+ */
+export async function expectAnnotationStatsText({
+  page,
+  activeViewport,
+  rightPanelPageObject,
+  toolName,
+  measurementIndex = 0,
+  annotationUID,
+  formatPanelPrimaryLines,
```
jbocce: These are all prefixed with format. Let's chat about changing this prefix to assert when we meet tomorrow.

tests/utils/expectAnnotationText.ts
```diff
@@ -0,0 +1,277 @@
...
+
+/**
+ * Vendored copy of `roundNumber` from `platform/core/src/utils/roundNumber.js`
+ * (identical to `@cornerstonejs/core`’s implementation).
+ *
+ * The Playwright E2E harness runs in Node and does not resolve `@ohif/core`
+ * workspace imports, so we cannot import this helper directly. Keep this copy
+ * in sync with core if the algorithm changes.
+ */
+function roundNumber(value, precision = 2) {
```
jbocce: This is my least favourite part of this code review. I understand why it is here, but now if something changes in the core code it has to change here too and that is not cool. Also this logic is a bit much for a testing environment.

I am wondering, if each of the measurementTextFormatters below took an expected value (and other needed stuff like unit) instead of stats, then there would be no need for rounding right? So something like the test assertion for an length tool would be something like the following...

```
await expectAnnotationStatsText({
    page,
    activeViewport,
    rightPanelPageObject,
    toolName: 'Length',
    formatPanelPrimaryLines: [measurementTextFormatters.lengthLine(278)],
    formatSvgLines: [measurementTextFormatters.lengthLine(278)],
    assertStats: stats => {
      expect(stats.unit).toBe('mm');
      expect(Math.round(stats.length as number)).toBe(278);
    },
  });
```

And the `lengthLine` would be something like the following...

```
lengthLine: (expectedValue: number, expectedUnit: string) =>
    `${expectedValue} ${expectedUnit}`,
```

Let's discuss tomorrow.

jbocce (changes requested):
A good start. See my comments.

tests/Rectangle.spec.ts
```diff
@@ -25,4 +34,32 @@ test('should display the rectangle tool', async ({
...
+    formatSvgLines: [
+      measurementTextFormatters.areaSvgLine,
+      measurementTextFormatters.meanSvgLine,
+      measurementTextFormatters.maxLine,
+      measurementTextFormatters.minSvgLine,
+      measurementTextFormatters.stdDevSvgLine,
+    ],
+    assertStats: stats => {
+      expect(stats.areaUnit).toBe('mm²');
+      expect(stats.area as number).toBeGreaterThan(0);
```
jbocce: We should assert a specific value here. We can determine if we have the "correct" value or not by leveraging the length tool and maybe even looking at the raw dicom to determine that the calibration is correct.

tests/Rectangle.spec.ts
```diff
@@ -25,4 +34,35 @@ test('should display the rectangle tool', async ({
...
+    expectedSvgLines: [
+      measurementTextFormatters.areaSvgLine('15959'),
+      measurementTextFormatters.meanSvgLine('80.4'),
+      measurementTextFormatters.maxLine('295'),
+      measurementTextFormatters.minSvgLine('-77.0'),
+      measurementTextFormatters.stdDevSvgLine('38.2'),
+    ],
+    assertStats: stats => {
+      expect(stats.areaUnit).toBe('mm²');
+      expect(Math.round(stats.area as number)).toBe(15959);
```
GhadeerAlbattarni: @jbocce The value for this area is different from the value that is in the current screenshot.

jbocce: Let's check it via the length tool. That said, I think there was a problem with the rectangle area that was fixed probably since the screenshot was taken, so I am not surprised by this at all. Shameful... I know... 🙁

tests/Angle.spec.ts
```diff
@@ -26,4 +35,18 @@ test('should display the angle tool', async ({
...
   );
+
+  await rightPanelPageObject.measurementsPanel.select();
+
+  await expectAnnotationStatsText({
+    page,
+    activeViewport,
+    rightPanelPageObject,
+    toolName: 'Angle',
+    expectedPanelPrimaryLines: [measurementTextFormatters.angleLine('53.1')],
```
jbocce: Maybe use a constant for 53.1 and then use interpolation where a string is needed?

tests/Bidirectional.spec.ts
```diff
@@ -25,4 +34,26 @@ test('should display the bidirectional tool', async ({
...
+
+  await rightPanelPageObject.measurementsPanel.select();
+
+  await expectAnnotationStatsText({
+    page,
+    activeViewport,
+    rightPanelPageObject,
+    toolName: 'Bidirectional',
+    expectedPanelPrimaryLines: [
+      measurementTextFormatters.bidirectionalLengthLine('195'),
```
jbocce: use constants

jbocce (changes requested):
I think we just need some constants and we will be good to go.

jbocce (approved):

PR #6130, opened 2026-07-09
fix: Several new worklist issues

wayfarer3130:
### Context

The previous worklist preferred prefix-matching inputs for the modalities over other matches.  That is almost always what a user wants so that if they type "PT" and hit enter, they DONT get "CTPT" as the value selected.
As well, added double-click behaviour so that the row doesn't disappear on a double click, and instead launches the default instance.  I believe that had been agreed on but didn't get implemented.  Made that customizeable, although I'd prefer to allow a cusotmization that specifies a command instead of a function.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] OS: 
- [] Node version: 
- [] Browser:
  

## Summary by CodeRabbit

- **New Features**
  - Added `workList.onStudyDoubleClick` customization to control what happens when double-clicking a study row in the default Work List.
  - Double-click can now launch the configured command/workflow, with smart fallback to an applicable default/first workflow when needed.
  - Improved study-row selection and double-click flow so the row is selected before actions run.
  - Added a built-in workflow launcher command (`launchDefaultMode`) for double-click behavior.
- **Documentation**
  - Updated customization samples and descriptions with the new `workList.onStudyDoubleClick` option and default configuration.

### DIFF
--- a/platform/app/src/routes/WorkList/WorkList.tsx
+++ b/platform/app/src/routes/WorkList/WorkList.tsx
@@ -9,7 +9,7 @@ import {
   Icons,
   InvestigationalUseDialog,
   type StudyRow,
-  type Workflow,
+  type OnStudyDoubleClick,
 } from '@ohif/ui-next';
 import { StudyListSettingsPopover } from './StudyListSettingsPopover';
 import { SidePanelPreview } from './SidePanelPreview';
@@ -57,9 +57,7 @@ export default function WorkList({
   // (launch the default workflow, falling back to the first applicable one).
   const customOnStudyDoubleClick = customizationService.getCustomization(
     'workList.onStudyDoubleClick'
-  ) as
-    | ((study: StudyRow, context: { defaultWorkflow?: Workflow; workflows: Workflow[] }) => void)
-    | undefined;
+  ) as OnStudyDoubleClick | undefined;
 
   const columns = useMemo(() => {
     // `workList.columns` is registered as a value (StudyList.defaultColumns) and
--- a/platform/app/src/routes/WorkList/WorkList.tsx
+++ b/platform/app/src/routes/WorkList/WorkList.tsx
@@ -4,7 +4,13 @@ import { useAppConfig } from '@state';
 import { preserveQueryParameters } from '../../utils/preserveQueryParameters';
 import { useStudyListStateSync, useWorkListToolbarActions } from '../../hooks';
 
-import { StudyList, Icons, InvestigationalUseDialog, type StudyRow } from '@ohif/ui-next';
+import {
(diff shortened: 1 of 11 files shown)


jbocce:
You mentioned that you prefer a command instead of a function, why did you choose not to implement a custom command for double click?

jbocce:
From claude: 

- Type duplication. The context shape { defaultWorkflow?: Workflow; workflows: Workflow[] } and the full handler signature are written out in three places
  (Table.tsx TableProps, the onStudyDoubleClick cast in WorkList.tsx). TableContent already reuses TableProps['onStudyDoubleClick'] — good. Consider
  exporting a single OnStudyDoubleClick type from the StudyList barrel and importing it in WorkList.tsx so the customization cast can't drift from the prop.

jbocce (changes requested):
Sorry for perhaps dropping some of these in the implementation although I am not sure they were explicitly mentioned in the requirements.

See my comments.

wayfarer3130:
> You mentioned that you prefer a command instead of a function, why did you choose not to implement a custom command for double click?

Because the modes don't know how to register command handlers, so the timing is wrong for anything to be available.

wayfarer3130:
@jbocce - I updated the PR to address your concerns and also added the command registration in modes so that it becomes possible to configure a customization as a command name instead of a function.  That makes it cleaner overall.

platform/app/src/routes/WorkList/WorkList.tsx
```diff
@@ -53,11 +55,21 @@ export default function WorkList({
...
+  // The study and its applicable workflows are merged into the command options
+  // at call time, so an override only needs to name a command and any static
+  // options (e.g. a specific `workflowId`).
+  const studyDoubleClickCommand = customizationService.getCustomization(
     'workList.onStudyDoubleClick'
-  ) as OnStudyDoubleClick | undefined;
+  ) as RunInput;
+  const onStudyDoubleClick = useCallback<OnStudyDoubleClick>(
+    (study, { defaultWorkflow, workflows }) => {
+      commandsManager.run(studyDoubleClickCommand, { study, defaultWorkflow, workflows });
```
jbocce, line 69: Should we handle the case if someone "undefines" workList.onStudyDoubleClick?

jbocce (approved):
One minor comment to consider, but approved!

platform/app/src/routes/WorkList/WorkList.tsx
wayfarer3130: That is already handled - if you pass in undefined it is a no-op internally in commandsManager

PR #6133, opened 2026-07-10
feat(customization): typed customization registry via AppTypes.Customizations

sedghi:
### Context

The CustomizationService exposes ~90 customization ids, but `getCustomization(id)` returns the loose `Customization` union for all of them. There is no discoverability of which ids exist or what shape each expects, and consumers compensate with ~20 `as unknown as X` / `as any` casts across the repo. `setCustomizations` payloads (including `$set` / `$push` command specs used by modes and app config) are not checked at all.

### Changes

This PR adds the typing infrastructure (no per-key population yet, and no runtime changes):

- **`AppTypes.Customizations` registry** — an empty, declaration-merged global interface in `platform/core/src/types/AppTypes.ts`, following the exact pattern already used by `AppTypes.Services` and `PresentationIds`. Each extension (in-tree or third-party) declares the ids it owns:

  ```ts
  declare global {
    namespace AppTypes {
      interface Customizations {
        'viewportOverlay.topLeft': OverlayItem[];
        'panelSegmentation.disableEditing': boolean;
      }
    }
  }
  ```

- **Typed overloads** on `getCustomization`, `getValue`, and `hasCustomization`: registered ids get autocomplete and their declared return type; any other string (dynamic keys, undeclared third-party keys) falls back to the existing `Customization | undefined` signature, so nothing breaks.

- **Checked `setCustomizations` payloads** via the new `CustomizationEntries` type: registered ids accept their declared value directly or as an immutability-helper spec (`{ $set: ... }`, `{ $push: [...] }`, the custom `$filter`, ...). `CustomizationPhaseInput` reuses it, so phase-tagged `appConfig.customizationService` blocks written in TypeScript are checked the same way.

- **Two small call-site fixes**: `usAnnotation` and `basic-test-mode` passed the raw string `'mode'` where the `CustomizationScope` enum is expected (Mode is already the default scope, so the argument is dropped; runtime behavior identical).

- **`CUSTOMIZATION_TYPING_PLAN.md`** documents the design, the conventions for declaring keys, and the follow-up phases (populating the registry per extension, then optionally generating the docs table and a JSON Schema for JSONC customization files from the registry).

### Testing

- All 68 CustomizationService unit tests pass.
- Full-repo `tsc` diffed against a pre-change baseline: three pre-existing false positives fixed (valid customization objects rejected by the old `Record<string, Customization>` input type), zero new errors.
- A standalone compile-time smoke test exercised the mechanism end to end (registry augmentation from an extension, typed reads, wrong-type reads rejected, `$push` of wrong element type rejected, `$filter` accepted, unknown keys and string module refs still compiling) under both the repo compiler settings and `--strict`.

## Summary by CodeRabbit

* **New Features**
  * Added typed and discoverable customization IDs with improved autocomplete and strongly typed lookups.
  * Enabled safer customization updates via a typed entries format, including support for additional update commands.
* **Bug Fixes**
  * Improved typing for sorting-related customization reads without changing behavior.
* **Documentation**
  * Added a rollout plan and expanded guidance on typing customizations, including how declared types relate to resolved values and how to handle edge cases.

### DIFF
--- a/CUSTOMIZATION_TYPING_PLAN.md
+++ b/CUSTOMIZATION_TYPING_PLAN.md
@@ -1,0 +1,130 @@
+# Typed Customizations: Design and Rollout Plan
+
+This document accompanies the PR that introduces the `AppTypes.Customizations`
+registry and describes the follow-up work needed to make every customization id
+in OHIF fully typed and discoverable.
+
+## Problem
+
+`customizationService.getCustomization(id)` accepts any string and returns the
+`Customization` union, which is broad enough to be effectively untyped. As a
+result:
+
+- There is no way to discover which customization ids exist from the code; the
+  docs table (`platform/docs/docs/platform/services/customization-service/sampleCustomizations.tsx`)
+  is hand-maintained and already drifts from the registered keys.
+- Consumers cast at the call site (`as unknown as ColorbarCustomization`,
+  `as string`, `as any`, ...) — roughly 20 such casts exist across the repo.
+- `setCustomizations` payloads (including `$set` / `$push` / `$merge` command
+  specs used by modes and app config) are not checked at all.
+
+## Design (implemented in this PR)
+
+A declaration-merged registry, following the same pattern already used for
+`AppTypes.Services` and `PresentationIds`:
+
+1. `platform/core/src/types/AppTypes.ts` seeds an empty global interface:
+
+   ```ts
+   declare global {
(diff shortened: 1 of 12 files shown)


wayfarer3130 (approved):
Thank you for doing this - I've been meaning to add this fix in for a while now and haven't gotten to it.

PR #6138, opened 2026-07-13
fix(core): report single-location multi-frame series as non-reconstructable

TFRadicalImaging:
### Context

A multi-frame cine / multi-time series (e.g. a cardiac cine) is many frames acquired at a **single spatial location** — every frame shares one `ImagePositionPatient`, so the series has no through-plane extent (span `0mm`).

`processMultiframe` never checks the spread of the frame positions — that is the intent of the existing `// TODO - check spacing consistency` — so such a series is reported as `isReconstructable: true`. Any volume-based layout (MPR / 3D) then builds a **degenerate volume**: on a real 30-frame cardiac cine we measured `dims [208, 174, 30]` with `spacing [1.92, 1.92, 0]` — a zero Z spacing. Every MPR plane renders **black**, and because the layout switch replaces the hanging protocol there is no way back short of a page reload.

The same hazard exists on the single-frame path: many single-frame instances acquired at one location fall through to the spacing calculation and produce the same degenerate volume.

### Changes & Results

`platform/core/src/utils/isDisplaySetReconstructable.js`:

- Add `zeroExtentTolerance` (`0.01mm`). Real slice spacing is on the order of millimetres, so frames whose positions all collapse within this distance occupy a single location.
- Add `hasSingleSpatialLocation(positions)` — true when **two or more** positions are known and all collapse to one location. It **fails open**: with fewer than two valid positions it returns `false`, so a genuine volume is never wrongly blocked when the geometry can't be determined confidently.
- Add `getPerFramePositions(multiFrameInstance)` — reads each frame's `ImagePositionPatient` from `PerFrameFunctionalGroupsSequence`, falling back to `SharedFunctionalGroupsSequence` broadcast to `NumberOfFrames`. This fallback matters: a single-location cine has one *constant* position, which the more common DICOM encoding stores in the **shared** group rather than repeating it per frame. Without it the per-frame read returns `[]` and the check never fires on the exact degenerate case it targets.
- Guard both `processMultiframe` (per-frame positions) and `processSingleframe` (instance positions).
- Export the two helpers.

**Before:** single-location cine reports `isReconstructable: true` → MPR/3D builds a zero-Z-spacing volume → all viewports black, unrecoverable without a reload.
**After:** it reports `isReconstructable: false` → volume layouts correctly treat it as non-volumetric (they are gated on `isReconstructable`), so the series stays in its stack/cine view and nothing goes black. A genuine volume series is unaffected and still reconstructs.

### Testing

New unit tests: `platform/core/src/utils/isDisplaySetReconstructable.test.js` (13 cases, none existed before) — helper-level (collapse within tolerance, real extent, fewer than two positions, malformed positions, per-frame read, shared-group fallback, absent sequences) and end-to-end through `isDisplaySetReconstructable` for both multi-frame and single-frame series.

```
cd platform/core && jest src/utils/isDisplaySetReconstructable.test.js
```

Manually verified against **real acquisition data** in a downstream deployment, with two series from the same study:

- `4 CV Cine` — 30 frames, all at one `ImagePositionPatient`, span **0mm** → now `isReconstructable: false`; selecting MPR is a safe no-op instead of turning every viewport black.
- `AXIAL TRUFI CHEST` — 20 frames, span **190mm** → still `isReconstructable: true`; MPR builds a real three-plane volume that renders normally.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 15 (Darwin 25.5.0)
- [x] Node version: 22.17.1
- [x] Browser: Chrome 138

---

This fix is contributed by the **University of Calgary**, who are donating it back to OHIF.

It was originally implemented as a layout guard in a downstream fork; @wayfarer3130 pointed out during review that the correct fix belongs in core — reporting the series as non-reconstructable, so that every surface derives the right behaviour from the data rather than from a UI-level guard — which is what this PR does.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reconstruction validation for multi-frame and single-frame imaging datasets.
  * Prevented reconstruction when frames/images collapse to a single spatial location with no measurable through-plane extent.
  * Improved handling of image position data by prioritizing per-frame positions with a shared fallback when needed.
* **Tests**
  * Added Jest coverage for single-location detection, per-frame position extraction, and reconstruction outcomes across per-frame and shared metadata scenarios.


TFRadicalImaging:
@wayfarer3130 this is the core fix you asked for during the downstream review — reporting a single-location multi-frame series as non-reconstructable, rather than guarding the layout switch in the fork. Could you take a look when you get a chance? cc @jbocce

TFRadicalImaging:
> `getPerFramePositions` trusts per-frame data on a single valid entry, bypassing the shared-group fallback.

Applied. The per-frame groups are now preferred only once they carry two valid positions, which is the same bar `hasSingleSpatialLocation` needs to decide anything — below it, a lone per-frame position was shadowing a shared position that would have shown the frames are co-located. Added a test for that case.

> consider adding the exact 2-frame/2-instance boundary case

Added. The spacing check only runs for more than two instances, so a co-located pair rests entirely on the single-location guard; the test fails without it (returns `true`).

wayfarer3130 (approved):

PR #6139, opened 2026-07-13
fix(cornerstone-dicom-sr): handle content-less imaging measurement reports

TFRadicalImaging:
### Context

An Imaging Measurement Report can be stored **without its report body** — the `ConceptNameCodeSequence` identifies it as an Imaging Measurement Report, but `ContentSequence` (0040,A730) is absent. We hit this with real reports produced by a third-party system.

OHIF classifies such a report as an Imaging Measurement Report on the concept name alone, then takes the measurement path on content that isn't there:

- `_load` → `_getReferencedImagesList(ContentSequence)` and `_getMeasurements(ContentSequence)` with `ContentSequence === undefined`.
- `_getMeasurements` immediately calls `.find` on it, which throws.

The exception escapes and takes down the viewer with the generic **"Something went wrong"** screen — the user can't open the study at all, not just that one report.

### Changes & Results

`extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts`:

- Split the concept-name test (`conceptIsImagingMeasurementReport`) from the resulting classification. A report whose concept says Imaging Measurement Report but which carries no `ContentSequence` is now flagged `isContentlessImagingMeasurementReport` and **classified as a plain SR** (`isImagingMeasurementReport: false`).
- Because both the loader (`_load`, which gates the measurement parsing on `isImagingMeasurementReport`) and the viewport (`OHIFCornerstoneSRViewport`, which routes on the same flag) already branch on that flag, downgrading it is enough to keep both off the measurement path that assumes content exists. No changes to the parsing code itself.

`extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRTextViewport.tsx`:

- The downgraded report now renders in the plain-SR text viewport, which would otherwise show an empty panel with no explanation. Show a warning notification when such a report is opened.
- The notification fires from the viewport (i.e. when the user actually opens the report) rather than at display-set creation, so a study containing one of these reports doesn't throw a notification at load time for a report the user may never open.

**Before:** opening a study containing a content-less Imaging Measurement Report crashes the viewer with "Something went wrong".
**After:** the report opens as an empty plain SR with a "This report has no readable content and cannot be displayed." warning; the rest of the study is unaffected. Healthy SRs — any report that has a `ContentSequence` — are untouched and still take the full measurement path.

### Testing

To reproduce, you need an SR whose `ConceptNameCodeSequence.CodeValue` is `126000` (Imaging Measurement Report) and which has **no** `ContentSequence` (0040,A730).

1. Load a study containing such a report.
2. Open the SR from the study panel.
3. **Before this change:** the viewer crashes to the "Something went wrong" screen.
4. **After this change:** the report opens as an empty SR with a warning notification, and the rest of the study remains usable.

Regression check — a healthy Imaging Measurement Report (one with a `ContentSequence`) still loads its measurements and renders in the measurement viewport, unchanged.

This fix has been running in a downstream production deployment against the real reports that triggered the crash.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 15 (Darwin 25.5.0)
- [x] Node version: 22.17.1
- [x] Browser: Chrome 138

---

This fix is contributed by the **University of Calgary**, who are donating it back to OHIF. It was co-authored with Ron Leisti.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of imaging measurement reports that lack report content.
  * Prevented attempts to parse incomplete reports, avoiding display errors.
  * Added a visible message indicating when a report has no content to display.
  * Display sets now surface this issue consistently so it can be identified in the interface.


TFRadicalImaging:
@wayfarer3130 @jbocce could one of you take a look? This is a small classification fix in the SR SOP class handler: an Imaging Measurement Report stored without its ContentSequence currently crashes the viewer outright ("Something went wrong"), because the measurement path calls `.find` on the missing content. Downgrading it to a plain SR keeps both the loader and the viewport off that path.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRTextViewport.tsx
```diff
@@ -1,13 +1,29 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useEffect } from 'react';
 import { ExtensionManager } from '@ohif/core';
+import i18n from '@ohif/i18n';
 import { OHIFCornerstoneSRContainer } from './OHIFCornerstoneSRContainer';
 
 function OHIFCornerstoneSRTextViewport(props: withAppTypes) {
-  const { displaySets } = props;
+  const { displaySets, servicesManager } = props;
```
wayfarer3130: Prefer useServices() instead of props passing.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRTextViewport.tsx
```diff
@@ -1,13 +1,29 @@
...
   const displaySet = displaySets[0];
   const instance = displaySet.instances[0];
 
+  // A report flagged as an Imaging Measurement Report but stored without its
+  // ContentSequence is downgraded to a plain SR (see getSopClassHandlerModule)
+  // and lands here with nothing to show. Warn the user that the report is empty
+  // when they open it.
+  useEffect(() => {
+    if (!displaySet.isContentlessImagingMeasurementReport) {
```
wayfarer3130: This adds specific behaviour for only this type of display set, rather than building on the general error/warning framework used for volumes already.  That means that other/new types of warnings can't be built.  Can you use the same sort of fields as the generic volume behaviour ones?   You might need a new "type" field or other flag to indicate this should be shown to the user on load.

wayfarer3130 (changes requested):
Can you make this use the standard display set error handling - that will then get reported in the display set tray, but that is ok.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRTextViewport.tsx
TFRadicalImaging: The warning moved out of this component entirely, so it no longer reads the services manager and this file is back to its previous form — no props passing and no useServices() needed here.

extensions/cornerstone-dicom-sr/src/components/OHIFCornerstoneSRTextViewport.tsx
TFRadicalImaging: Reworked onto the generic mechanism. The handler now adds a `MISSING_REPORT_CONTENT` code to the display set's `DisplaySetMessageList` instead of setting a flag that only this viewport read, so it surfaces in the display set tray through `_mapDisplaySets` like any other display set problem, and further SR warnings can be added as codes on the same list.

I also widened the condition: it now covers any SR without a ContentSequence rather than only one whose concept name claims to be an Imaging Measurement Report, since neither has anything to render.

On the "shown to the user on load" part — I left that out, taking your note that the tray is acceptable. Adding it would mean a severity/type field on `DisplaySetMessage` that every consumer would need to handle, which felt like a separate change rather than something to fold in here. Happy to do it if you would rather have it in this PR.

TFRadicalImaging:
Switched to the standard display set error handling as asked — the empty report is now reported through `DisplaySetMessageList` and shows in the display set tray, and the bespoke notification and its display set flag are gone.

Verified `platform/core` unit tests (39 suites / 356 tests) after the change. Two things I could not run in this checkout: eslint fails at startup on an ajv/eslintrc incompatibility, and it does so on files this PR does not touch, so it looks environmental rather than related — CI will be the real check on both.

@wayfarer3130 re-requesting review.

wayfarer3130 (approved):

PR #6145, opened 2026-07-15
fix(Layout): apply the selected preset instead of restoring a stale grid

TFRadicalImaging:
### Context

Two independent stale-grid bugs make a layout change land on the wrong grid. Both were found and fixed on a private fork and are donated back here; they reproduce on current `master`.

**1. An advanced layout preset restores a previously cached custom grid instead of the preset's own layout.**
`setHangingProtocol` computes `restoreProtocol = !reset && viewportGridState[storedHanging]` and, when set, calls `viewportGridService.set(viewportGridState[storedHanging])` — replaying whatever grid was cached for that protocol/stage. That cache is the right behaviour when *navigating* (returning to a study should restore how you left it), but when the user *explicitly picks a preset* from the toolbar layout selector it means a grid left behind by an earlier, unrelated layout carries over. The result is a grid that doesn't match the chosen preset — panes dropped, remaining cells stale or blank. A clear trigger is any preset whose selector doesn't match the active series (e.g. a frame-based layout on a single-frame stack): it keeps the previous grid, that grid gets cached, and the next preset (MPR, 3D, …) then inherits it.

**2. Double-clicking a 1×1 preset restores a stale one-up grid.**
`toggleOneUp` decides "am I in one-up?" purely from `layout.numRows === 1 && numCols === 1`, then restores `useToggleOneUpViewportGridStore`. That store is written only when toggling *into* one-up (double-click) and cleared only on mode exit. But advanced presets ("3D only", "Axial Primary", …) reach a 1×1 grid through `setHangingProtocol`, not `setViewportGridLayout`. So after using one-up once and later selecting a 1×1 preset, a double-click sees "1×1" and replays the stale grid from the earlier one-up instead of doing nothing.

### Changes & Results

**Bug 1 — `setHangingProtocol` / `ToolbarLayoutSelector`:**
- Add an optional `restoreCachedLayout` param to `setHangingProtocol` (default `true`, so restore-on-navigation is unchanged) and gate the restore on it: `restoreProtocol = !reset && restoreCachedLayout && viewportGridState[storedHanging]`.
- The toolbar layout selector passes `restoreCachedLayout: false` for advanced presets, so an explicit pick applies the protocol's own stage layout while still preserving the active series via `displaySetSelectorMap`.

*Result:* each preset applies its own layout from a clean state; cross-layout transitions no longer drop or blank viewports.

**Bug 2 — `useToggleOneUpViewportGridStore` lifecycle:**
- Clear the one-up store on any explicit layout change that actually proceeds: at the success end of `setHangingProtocol` (not on the `catch`), and in `setViewportGridLayout` after the `onLayoutChange` veto guard (so a rejected change keeps the toggle-back state). Also clear it after the one-up is toggled back.

*Result:* the store now means strictly "a one-up entered via double-click, pending toggle-back". Double-clicking a preset-reached 1×1 is a no-op instead of a stale-grid jump; the normal one-up round-trip is unchanged.

### Testing

Unit test added: `extensions/default/src/__tests__/commandsModuleToggleOneUp.test.ts` (3 cases) — asserts `setViewportGridLayout` and `setHangingProtocol` clear a pending one-up when the change proceeds, and that a vetoed `onLayoutChange` preserves it. Reverting the source makes the two "clears" cases fail; the veto case passes either way.

```
cd extensions/default
npx jest --ci --runInBand --coverage=false src/__tests__/commandsModuleToggleOneUp.test.ts
# 3 passed
```

Manual (both fixes were live-validated in the fork):
- Bug 1: from a layout whose selector doesn't match the active series (e.g. a frame layout on a single-frame CT), pick MPR / a 3D preset → the selected preset's layout is applied with no dropped/blank panes (previously it came back constrained to the prior grid). Advanced presets require the active viewport's series to be reconstructable.
- Bug 2: enter one-up via double-click, then select a 1×1 preset ("3D only") → double-click stays 1×1 (no stale grid). The normal 2×2 → one-up → toggle-back round-trip still restores the 2×2.

### Attribution

Donated by the University of Calgary. The original fork changes were authored and reviewed by Ron Leisti (@rleisti).

#### PR

#### Code

#### Tested Environment

- [x] OS: macOS 15
- [x] Node version: 24.3.0
- [x] Browser: Chrome

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved layout switching for advanced and preset configurations by preventing unintended restoration of previously cached layouts.
  * One-up viewport states are now cleared consistently after confirmed grid changes, applied presets, or returning to the previous layout.
  * Pending one-up states are preserved when a layout change is rejected by the active protocol.


TFRadicalImaging:
cc @wayfarer3130 @jbocce — would appreciate a review when you have a moment. Both fixes are in the generic `extensions/default` command module; happy to adjust the approach.

wayfarer3130 (approved):

PR #6183, opened 2026-07-26
fix(worklist): make staticWado study filtering case-insensitive

### LINKED ISSUE #6029: [Bug] MRN, Description, and Accession Number filtering does not work properly for lowercase input.
### Describe the Bug

MRN, Description, and Accession Number filtering works as expected for uppercase input, but filtering does not work properly for lowercase input.

### Steps to Reproduce

1. https://deploy-preview-6005--ohif-dev.netlify.app/
2. Enter an uppercase value in the MRN, Description, or Accession Number filter field and observe that the filtering works correctly.
3. Click Reset to clear the filter
4. Enter the same value using lowercase letters.
5. Observe that the filtering does not return the expected results.

https://github.com/user-attachments/assets/784b5ea0-b849-45ee-88dc-17d8bf2644c9

### The current behavior

MRN, Description, and Accession Number filtering does not work properly for lowercase input.

### The expected behavior

MRN, Description, and Accession Number filtering should work properly for both uppercase and lowercase input.

### System Information

 System:
    OS: Windows 11 10.0.26200
    Memory: 3.90 GB / 15.67 GB
  Browsers:
    Chrome: 148.0.7778.168
    Edge: Spartan (24.13.1), Chromium (24.13.1), ChromiumDev (24.13.1)

awatson1978:
### Context

Fixes #6029

If you open the study list on a staticWado-backed data source (the default AWS demo data, deploy previews, etc.) and type a filter value in lowercase — say `chest` in the Description field — you get "No studies available", even though plenty of studies have "CHEST" in their description. Type the same thing in uppercase and it works. That's surprising behavior for a worklist search box, and it affects MRN, Description, and Accession Number alike.

### Changes & Results

**Why this happens:** staticWado sources can't run real server-side queries, so `StaticWadoClient` fetches the full study list and filters it client-side in `compareValues()`. That function has two paths: a fuzzy-matching path that normalizes case, and a plain path that compares strings exactly as typed. The fuzzy path only ever applies to patient-name fields, and it was recently disabled on the default sources anyway (#6163) — so MRN, Description, and Accession Number always go through the case-sensitive path.

**The fix:** lowercase both the typed filter value and the stored attribute right before the plain string comparison, so matching is case-insensitive in both directions (lowercase input matches "CHEST", uppercase input matches "Chest CT Routine"). Wildcard handling (`*`) keeps working on the normalized strings, and date-range comparisons are unaffected since those are digit strings.

```ts
// Worklist filtering is expected to be case-insensitive regardless of
// fuzzy matching support (e.g. MRN, description, accession number).
if (typeof actual === 'string' && typeof desired === 'string') {
  actual = actual.toLowerCase();
  desired = desired.toLowerCase();
}
```

One thing worth flagging for reviewers: strict QIDO-RS matching is case-sensitive for non-PN attributes. Since this filter is a client-side worklist convenience rather than a real QIDO implementation, case-insensitive feels like the right default — but happy to gate it behind a config flag if you'd prefer.

**Before** (lowercase `chest` → no results) / **After** (same input → all CHEST studies match):

### Testing

1. `pnpm dev` with the default (staticWado AWS) data source
2. In the Study List, type `chest` (lowercase) into the Description filter
3. Before: no studies returned. After: all studies with CHEST/Chest/chest in the description are listed
4. Repeat with uppercase `CHEST` — mixed-case descriptions like "Chest CT Routine" still match

##### 6029-before-lowercase-chest-no-results
6029-before-lowercase-chest-no-results

##### 6029-after-lowercase-chest-matches
6029-after-lowercase-chest-matches

#### PR

#### Code

      etc.)

- [x] The documentation page has been updated as necessary. (no doc changes needed)

#### Tested Environment

- [x] "OS: macOS 25.5.0
- [x] "Node version: 24.x
- [x] "Browser: Chromium (Playwright)

## Summary by CodeRabbit

* **Bug Fixes**
  * Worklist filtering is now case-insensitive when comparing text values.
  * Wildcard and fuzzy matching provide more consistent results regardless of letter casing.

### DIFF
--- a/extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
+++ b/extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
@@ -190,6 +190,13 @@ export default class StaticWadoClient extends api.DICOMwebClient {
       );
     }
 
+    // Worklist filtering is expected to be case-insensitive regardless of
+    // fuzzy matching support (e.g. MRN, description, accession number).
+    if (typeof actual === 'string' && typeof desired === 'string') {
+      actual = actual.toLowerCase();
+      desired = desired.toLowerCase();
+    }
+
     if (typeof actual == 'string') {
       if (actual.length === 0) {
         return true;


wayfarer3130:
Some behaviour/explanation documentation should get added to explain differences between search types.  I'm not sure we don't also need some UI explanation so that users searching on different back ends have things explained as to why they get different behaviours.

extensions/default/src/DicomWebDataSource/utils/StaticWadoClient.ts
```diff
@@ -190,6 +190,13 @@ export default class StaticWadoClient extends api.DICOMwebClient {
       );
     }
 
+    // Worklist filtering is expected to be case-insensitive regardless of
```
wayfarer3130, line 193: Worklist filtering is not expected to ever be case insensitive according to the DICOM standard.  This will never work for any other qido server so I'm not sure it should be supported here when the stnadard specifically says it is case sensitive.

wayfarer3130 (changes requested):
@sedghi - I'm recommending we reject this PR because it makes the search for static dicomweb back ends different from everything else in that it is case insensitive.  The DICOM standard specifically indicates searches are case sensitive (except VR PN)

PR #6188, opened 2026-07-26
fix(Toolbar): honor hideWhenDisabled on array evaluators

### LINKED ISSUE #5587: [Bug] `hideWhenDisabled` evaluate property flag not working
### Describe the Bug

There is at least one instance in which the `hideWhenDisabled` flag is not working as expected. This particular case is for the `'evaluate.modality.supported'` evaluator when applied to the `UltrasoundDirectionalTool`. The problem may be more general, but this is the only instance/example tried.

### Steps to Reproduce

1. In the `toolbarButtons.ts` file for basic mode, add the `hideWhenDisabled: true` flag as an evaluator property like this...
```
  {
    id: 'UltrasoundDirectionalTool',
    uiType: 'ohif.toolButton',
    props: {
      icon: 'icon-tool-ultrasound-bidirectional',
      label: i18n.t('Buttons:Ultrasound Directional'),
      tooltip: i18n.t('Buttons:Ultrasound Directional'),
      commands: setToolActiveToolbar,
      evaluate: [
        'evaluate.cornerstoneTool',
        {
          name: 'evaluate.modality.supported',
          supportedModalities: ['US'],
          hideWhenDisabled: true,
        },
      ],
    },
  },
```
2. Run OHIF and display a non US study.

### The current behavior

The UltrasoundDirectionalTool is displayed in the more tools menu.

### The expected behavior

The UltrasoundDirectionalTool is NOT displayed in the more tools menu.

### System Information

  System:
    OS: Windows 11 10.0.26200
    Memory: 8.30 GB / 31.68 GB
  Binaries:
    Node: 23.9.0 - C:\Users\joebo\AppData\Local\fnm_multishells\34000_1764094687187\node.EXE
    npm: 10.9.2 - C:\Users\joebo\AppData\Local\fnm_multishells\34000_1764094687187\npm.CMD
  Browsers:
    Chrome: 142.0.7444.176
    Firefox: 143.0.4 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.1882

RubenHaisma:
### Context

Fixes #5587.

`refreshToolbarState` reads the flag off `props.evaluateProps`:

```ts
const evaluateProps = props.evaluateProps;
const hideWhenDisabled = evaluateProps?.hideWhenDisabled || props.hideWhenDisabled;
```

But in `handleEvaluate`, the **array** branch assigned the array itself:

```ts
const evaluateProps = props.evaluate;   // <- this is the array
props.evaluate = args => { /* ... */ };
props.evaluateProps = evaluateProps;
```

An array has no `hideWhenDisabled` property, so that lookup was always `undefined` and the flag was silently ignored — for **every** button whose `evaluate` is an array, not just `UltrasoundDirectionalTool`. The reporter suspected the problem was more general, and it is.

The single-object branch a few lines below assigns an object:

```ts
const { name, ...options } = evaluate;
const evaluateProps = props.evaluate;   // <- an object, so the lookup works
```

which is why that form works and the bug looked tool-specific.

### Changes & Results

The array branch now merges the option objects into a flat object. `name` is dropped: it selects which evaluator to run rather than configuring it, and has no single value once several evaluators are merged.

`props.evaluateProps` has exactly one consumer in the codebase (the `hideWhenDisabled` lookup above), so reshaping it from array to object affects nothing else.

**Before / after** for the exact config in the issue:

```ts
evaluate: [
  'evaluate.cornerstoneTool',
  { name: 'evaluate.modality.supported', supportedModalities: ['US'], hideWhenDisabled: true },
]
```

| | `evaluateProps` | `hideWhenDisabled` | tool on a non-US study |
|---|---|---|---|
| before | the array | `undefined` | visible in More Tools |
| after | `{ supportedModalities: ['US'], hideWhenDisabled: true }` | `true` | hidden |

### Testing

Added `platform/core/src/services/ToolBarService/__tests__/ToolbarService.evaluateProps.test.ts` (5 cases).

```
cd platform/core && npx jest src/services/ToolBarService
# Test Suites: 1 passed, Tests: 5 passed
```

Reverting only the source change and re-running gives **3 failed, 2 passed** — the three array-form assertions fail, while the single-object case and the evaluator-merging behaviour still pass. So the test pins the actual regression rather than restating the implementation.

Full `platform/core` suite: **357 passed, 39/40 suites**. The one failing suite is `src/utils/absoluteUrl.test.js` with `SecurityError: Cannot initialize local storage without a --localstorage-file path` — reproduced on an unmodified checkout, so it is pre-existing and unrelated (a Node 25 vs `jest-environment-node` artifact; this repo declares `node >=24`).

To verify in the app, apply the `hideWhenDisabled: true` snippet from the issue to `UltrasoundDirectionalTool` in a mode's `toolbarButtons.ts`, load a non-US study, and confirm the tool no longer appears in the More Tools menu.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 15 (Darwin 25.4.0, Apple Silicon)
- [x] Node version: 25.2.0, pnpm 11.5.2
- [x] Browser: not required — unit-tested at the service layer

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed toolbar buttons with multiple evaluators so evaluator configuration options (e.g., `hideWhenDisabled`) are correctly carried through.
  - Prevented evaluator names from being treated as persisted configuration properties.
  - Improved disabled-state handling so a disabled result from any evaluator updates the toolbar state as expected.
  - Maintained existing behavior for single-evaluator configurations.

- **Tests**
  - Added coverage for multi-evaluator and single-evaluator toolbar behavior, including visibility changes driven by evaluation results.

### DIFF
--- a/platform/core/src/services/ToolBarService/ToolbarService.ts
+++ b/platform/core/src/services/ToolBarService/ToolbarService.ts
@@ -718,7 +718,20 @@ export default class ToolbarService extends PubSubService {
         return evaluateFunction;
       });
 
-      const evaluateProps = props.evaluate;
+      // Object entries in the array carry per-evaluator options such as
+      // `hideWhenDisabled`. Merge those options into a flat object: assigning the
+      // array itself left `evaluateProps.hideWhenDisabled` permanently undefined,
+      // so the flag was silently ignored for every multi-evaluator button.
+      const evaluateProps = evaluate.reduce((acc, evaluator) => {
+        if (typeof evaluator !== 'object' || evaluator === null) {
+          return acc;
+        }
+        // `name` selects which evaluator to run; it is not an option, and it has no
+        // single value once several evaluators are merged.
+        const { name: _name, ...options } = evaluator;
+        return { ...acc, ...options };
+      }, {});
+
       props.evaluate = args => {
         const results = evaluators.map(evaluator => evaluator(args)).filter(Boolean);
 
--- a/platform/core/src/services/ToolBarService/__tests__/ToolbarService.evaluateProps.test.ts
+++ b/platform/core/src/services/ToolBarService/__tests__/ToolbarService.evaluateProps.test.ts
@@ -88,3 +88,89 @@ describe('ToolbarService.handleEvaluate — evaluateProps', () => {
     expect(props.evaluateProps.hideWhenDisabled).toBe(true);
   });
 });
+
+/**
+ * The suite above pins how `evaluateProps` is derived. This one covers the effect the
+ * issue actually reports: a button with `hideWhenDisabled` on an array evaluator must
+ * disappear from the toolbar once it evaluates as disabled.
+ */
+describe('ToolbarService.refreshToolbarState — hideWhenDisabled visibility', () => {
+  const US_ONLY_EVALUATORS = {
+    'evaluate.cornerstoneTool': () => ({ disabled: false }),
+    // Stands in for evaluate.modality.supported: disabled unless the study is US.
+    'evaluate.modality.supported': ({ modality, supportedModalities }) =>
+      supportedModalities?.includes(modality)
+        ? undefined
+        : { disabled: true, disabledText: 'Not supported for this modality' },
+  };
+
+  const usDirectionalButton = (extra = {}) => ({
+    id: 'UltrasoundDirectionalTool',
+    props: {
+      evaluate: [
+        'evaluate.cornerstoneTool',
+        {
+          name: 'evaluate.modality.supported',
+          supportedModalities: ['US'],
+          hideWhenDisabled: true,
+        },
+      ],
+      ...extra,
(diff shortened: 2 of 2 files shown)


RubenHaisma:
Thanks — good catch, and fair: the original cases pinned how `evaluateProps` is derived, which is the mechanism rather than the symptom.

Pushed four cases that drive `refreshToolbarState` and assert on the resulting button props:

- a button with `hideWhenDisabled` on an array evaluator is `visible: false` on a non-US study
- the same button stays visible on a US study
- without the flag, a disabled button stays visible
- the pre-existing props-level flag still works

The first of those fails against `master`, so the suite now fails for the reported symptom and not only for the internals. Full run: 9 passed.

PR #6189, opened 2026-07-26
fix(tmtv): stop Rectangle ROI Threshold erasing other segments

### LINKED ISSUE #5476: [Bug] In TMTV mode, all segments created by other segmentation tools are removed from the viewport when the user runs the Rectangle ROI Threshold tool.
https://github.com/user-attachments/assets/033e1ba6-bc32-45b6-a5f8-18ca10cfa573

### Describe the Bug

In TMTV mode, all segments created by other segmentation tools are removed from the viewport when the user runs the Rectangle ROI Threshold tool.
**Note:** 
The issue exists in OSS 3.10 and 3.11

### Steps to Reproduce

1. Launch the study in TMTV mode
2. Add New segmentation
3. Draw segments with segmentation tools like brush or threshold, or shape
4. Select the Rectangle ROI threshold tool and run the tool

### The current behavior

In TMTV mode, all segments created by other segmentation tools should display in the viewport when the user runs the Rectangle ROI Threshold tool.

### The expected behavior

In TMTV mode, all segments created by other segmentation tools are removed from the viewport when the user runs the Rectangle ROI Threshold tool.

### System Information

Windows 11
Node version 20.9.0
Browser 141.0.7390.54

RubenHaisma:
### Context

Fixes #5476. @sedghi diagnosed this in the thread and pointed at [`commandsModule.ts#L222`](https://github.com/OHIF/Viewers/blob/master/extensions/tmtv/src/commandsModule.ts#L222) — this implements it. @bijil-trenser confirmed in February that the issue still reproduces.

`thresholdSegmentationByRectangleROITool` passed `overwrite: true`. Cornerstone implements that by zeroing the **entire labelmap** — the whole volume, not the ROI bounds — before writing anything:

```js
// @cornerstonejs/tools · utilities/segmentation/thresholdVolumeByRange.js
if (overwrite) {
  for (let i = 0; i < scalarDataLength; i++) {
    segVoxelManager.setAtIndex(i, 0);
  }
}
```

So running the tool deleted every segment in the segmentation, including ones drawn with the brush, shape or other threshold tools.

### Changes & Results

Clear only the segment being recomputed, then write additively:

```ts
const targetSegmentIndex = segmentIndex ?? 1;
clearSegmentFromLabelmap(labelmapVolume.voxelManager, targetSegmentIndex);

return csTools.utilities.segmentation.rectangleROIThresholdVolumeByRange(
  annotationUIDs,
  labelmapVolume,
  [/* ... */],
  { overwrite: false, segmentIndex: targetSegmentIndex, segmentationId }
);
```

**Why not just `overwrite: false` on its own?** Without any clearing, the segment could only ever grow — voxels from a previous run would linger, so shrinking the ROI or raising the threshold would never remove anything. Scoping the clear to the target segment keeps re-runs idempotent, which is the property `overwrite: true` was presumably there for, without touching anyone else's work.

| | before | after |
|---|---|---|
| other segments in the segmentation | erased | preserved |
| re-running on the same segment | replaces | replaces |
| shrinking the ROI and re-running | correct | correct |

`clearSegmentFromLabelmap` is a small pure helper over the voxel-manager interface, so it is testable without a volume or a viewport.

No extra event is needed — `thresholdVolumeByRange` still calls `triggerSegmentationDataModified` once at the end, and the clear happens before it, so there is exactly one update as before.

### Testing

Six unit tests in `extensions/tmtv/src/utils/clearSegmentFromLabelmap.test.ts`: only the target segment is cleared, absent segment is a no-op, segment 0 leaves the labelmap untouched, the operation is idempotent, empty labelmap, and a labelmap holding only the target segment.

```
cd extensions/tmtv && npx jest
# Test Suites: 1 passed, Tests: 6 passed
```

**This PR also adds `extensions/tmtv/jest.config.js`.** The package already declares `test:unit` and `test:unit:ci` scripts but had no config, so nothing in it was ever collected. The root `jest.config.js` already globs `<rootDir>/extensions/*/jest.config.js`, so adding the file wires tmtv into the root run — confirmed with `npx jest --listTests`. Happy to split that into its own PR if you would rather keep this one to the fix.

To verify in the app: TMTV mode → add a segmentation → draw with the brush → select Rectangle ROI Threshold and run it. The brush segment should still be there.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 15 (Darwin 25.4.0, Apple Silicon)
- [x] Node version: 25.2.0, pnpm 11.5.2
- [x] Browser: not required — the changed logic is unit-tested; app-level repro steps above

## Summary by CodeRabbit

- **Bug Fixes**
  - Rectangle-based segmentation thresholding now preserves existing labelmap segments instead of overwriting the entire labelmap.
  - Re-running the tool updates only the selected segment, and safely rolls back if recomputation fails.

- **Tests**
  - Added Jest coverage for clearing/restoring individual segments, including empty, repeated, background, and single-segment labelmaps.
  - Added command-level coverage to verify correct behavior and safe error handling for the Rectangle ROI threshold flow.

### DIFF
--- a/extensions/tmtv/jest.config.js
+++ b/extensions/tmtv/jest.config.js
@@ -1,0 +1,9 @@
+const base = require('../../jest.config.base.js');
+
+module.exports = {
+  ...base,
+  moduleNameMapper: {
+    ...base.moduleNameMapper,
+    '@ohif/(.*)': '<rootDir>/../../platform/$1/src',
+  },
+};
--- a/extensions/tmtv/jest.config.js
+++ b/extensions/tmtv/jest.config.js
@@ -4,6 +4,9 @@ module.exports = {
   ...base,
   moduleNameMapper: {
     ...base.moduleNameMapper,
+    // Extension packages live under extensions/, so this must be matched before the
+    // generic @ohif/* rule below, which resolves into platform/.
+    '^@ohif/extension-(.*)$': '<rootDir>/../../extensions/$1/src',
     '@ohif/(.*)': '<rootDir>/../../platform/$1/src',
   },
 };
--- a/extensions/tmtv/src/commandsModule.test.ts
+++ b/extensions/tmtv/src/commandsModule.test.ts
@@ -1,0 +1,214 @@
+/**
+ * Command-level cover for #5476.
+ *
+ * The helper tests alone would still pass if this command cleared the wrong volume or
+ * kept sending `overwrite: true`, so assert the wiring: which segment gets cleared,
+ * what reaches cornerstone, and that a throw does not leave the segment erased.
+ */
+
+jest.mock('@ohif/core', () => ({
+  __esModule: true,
+  default: {},
+  classes: { MetadataProvider: { get: jest.fn() } },
+  utils: { formatPN: (v: string) => v },
+}));
+
+jest.mock('@ohif/i18n', () => ({ __esModule: true, default: { t: (k: string) => k } }));
+
+jest.mock('@ohif/extension-cornerstone', () => ({
+  getViewportFocalPoint: jest.fn(),
+}));
+
+jest.mock('@cornerstonejs/core', () => ({
+  cache: {
+    getVolume: jest.fn(),
+    getVolumeContainingImageId: jest.fn(),
+  },
+  getEnabledElement: jest.fn(),
+}));
+
+jest.mock('@cornerstonejs/tools', () => ({
+  Enums: { SegmentationRepresentations: { Labelmap: 'Labelmap' } },
+  segmentation: { state: { getSegmentation: jest.fn() } },
+  annotation: { selection: { getAnnotationsSelectedByToolName: jest.fn() } },
+  utilities: { segmentation: { rectangleROIThresholdVolumeByRange: jest.fn() } },
+}));
+
+jest.mock('./utils/getThresholdValue', () => ({
+  __esModule: true,
+  default: () => ({ ptLower: 1, ptUpper: 2, ctLower: 3, ctUpper: 4 }),
+}));
+
+jest.mock('./utils/createAndDownloadTMTVReport', () => ({ __esModule: true, default: jest.fn() }));
+jest.mock('./utils/dicomRTAnnotationExport/RTStructureSet', () => ({
+  __esModule: true,
+  default: jest.fn(),
+}));
+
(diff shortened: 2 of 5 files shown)


RubenHaisma:
Three findings here, all worth acting on — pushed in 175f35f.

**Invalid `targetSegmentIndex` (major).** Correct, and the reason is slightly different from the comment: cornerstone writes `options.segmentIndex || 1`, so passing `0` did not write to segment 0. It made us clear segment 0 (background, a no-op) while cornerstone appended to **segment 1 without clearing it first**. `normalizeSegmentIndex` now resolves the index exactly the way cornerstone does, so the segment we clear is always the segment it writes to. Covered for `0`, negatives, non-integers, `undefined`, `null`, `NaN` and strings.

**Atomicity (major).** This one is more serious than it first looks and I had it wrong initially, so thank you. Cornerstone validates annotation tool names *inside* `rectangleROIThresholdVolumeByRange`, **before** its own zeroing — so the previous `overwrite: true` never lost data when that validation threw. Moving the clear ahead of the call introduced a data-loss path that did not exist before, and it is reachable: `ROI_THRESHOLD_MANUAL_TOOL_IDS` includes `CircleROIStartEndThreshold`, which cornerstone rejects.

I went lighter than the suggested temporary labelmap, though. `clearSegmentFromLabelmap` now returns the indices it cleared, and the call is wrapped so `restoreSegmentToLabelmap` puts them back before rethrowing. Same guarantee for the target segment, but memory scales with the segment rather than duplicating a full PET/CT volume — which matters at these volume sizes. Happy to switch to a temp volume if you would rather have rollback cover writes the failed call made to *other* segments too; cornerstone only writes the one index, so I judged that unnecessary.

**Command-level test.** Agreed, and this was the real gap — the helper tests would pass with the wiring wrong. Added `extensions/tmtv/src/commandsModule.test.ts`: which segment is cleared, that `overwrite: false` and the resolved `segmentIndex` reach cornerstone, the segment-0 case, and that a throw leaves the labelmap untouched. **Four of its six cases fail against `master`.** Full tmtv run: 23 passed.

One config note: `jest.config.js` needed `@ohif/extension-* -> extensions/*` mapping, since the generic `@ohif/* -> platform/*` rule cannot resolve extension packages.

PR #6206, opened 2026-08-05
test(contour): Livewire contour draw coverage

diattamo:
### Context

`LivewireContourSegmentation.spec.ts` currently covers only one path — cancelling an
in-progress contour with `Escape`. Nothing verified that a Livewire contour can actually be
*completed*, or how completed contours behave once they land in a segment.

This PR expands the spec to cover the completed-draw behaviors: persistence across slice and
segment navigation, tool enablement, segment ownership, and the three boolean outcomes of
drawing a second contour into the same segment (disjoint, overlapping, and hole-cutting with
`Shift`).

### Changes & Results

Test-only change — no production code touched.

- **`tests/LivewireContourSegmentation.spec.ts`** — 6 new tests, plus the existing `Escape`
  test retuned to a shared 4-point `contourClicks` polygon (large enough to leave room for a
  hole inside it, and clear of neighbouring outlines):
  - **persists after navigating away and back** — draws a contour, closes it by
    double-clicking the last vertex, captures the SVG `path` `d` attribute, navigates to the
    last slice, then reselects Segment 1 and asserts the re-rendered geometry is byte-identical.
  - **tool enablement** — Livewire is disabled with no contour segmentation and becomes
    enabled once one is created.
  - **draws into the active segment** — contour drawn while Segment 2 is active survives
    hiding Segment 1 and disappears/reappears with Segment 2's visibility toggle.
  - **disjoint contours stay separate** — two non-touching contours in one segment render as
    two paths.
  - **overlapping contours merge** — a second contour crossing the first collapses to a
    single path.
  - **`Shift` cuts a hole** — an inner contour drawn with `Shift` held is subtracted into the
    same path rather than added as a new one.
- **`tests/pages/RightPanelPageObject.ts`** — expose `livewireContour.livewireToolButton`.
  The `data-cy` element is a span wrapper (so its tooltip still fires while the tool is
  unavailable); the inner `-btn` element is what carries the disabled state, so enablement is
  asserted against that.
- **`tests/utils/keyboardUtils.ts`** — add `keyDown`/`keyUp` so a modifier can be held across
  mouse interactions (needed for the `Shift` hole-cutting flow).
- **`tests/utils/screenShotPaths.ts`** + 3 baselines — the boolean outcomes (merge, hole) are
  canvas-only, so those three tests back the SVG path-count assertion with a
  viewport-scoped screenshot compare.
- **`.agents/skills/ohif-test-agent/SKILL.md`** — the "never substitute a state read for a
  render assertion" rule now names cornerstone state (`window.cornerstone`,
  `window.cornerstoneTools`) alongside `window.services`, and points at a viewport-scoped
  screenshot as the default when no DOM/SVG signal captures the result (contour geometry,
  merges, holes).

### Testing

```bash
cross-env TEST_ENV=true pnpm exec playwright test tests/LivewireContourSegmentation.spec.ts
```

All 7 tests pass locally (Chromium).

#### PR

#### Code

  etc.)

  additions or removals. 

#### Tested Environment

- [x] OS: macOS 26.5.1
- [x] Node version: 25.4.0
- [x] Browser: Chromium (Playwright)

## Summary by CodeRabbit

* **Tests**
  * Expanded Livewire contour segmentation coverage for cancellation, redraw, navigation persistence, visibility, segment assignment, and tool prerequisites.
  * Added tests for disjoint contours, same-segment overlap merging, cross-segment separation, and Shift-based hole creation.
  * Added visual verification for overlapping contours, holes, and separate segment contours.
  * Improved keyboard interaction testing, including held modifier keys.
  * Added coverage for disabled and enabled tool states.
  * Updated tool interaction checks for more reliable targeting.


tests/pages/RightPanelPageObject.ts
```diff
@@ -380,6 +380,10 @@ export class RightPanelPageObject {
           const button = page.getByTestId('LivewireContourSegmentationTool');
           return {
             button,
+            // The data-cy element above is a span wrapper, so its tooltip still
+            // fires while the tool is unavailable. The inner button is what
+            // carries the disabled state, so assert enablement against this.
+            toolButton: page.getByTestId('LivewireContourSegmentationTool-btn'),
```
jbocce, line 386: This is a bit confusing to have "two buttons" for the same page object. Can we consolidate to one? I suspect `'LivewireContourSegmentationTool-btn'` could be used for clicking too, but I might be wrong. Thanks.

tests/LivewireContourSegmentation.spec.ts
```diff
@@ -14,33 +32,339 @@ test('should cancel an in-progress Livewire contour segmentation via Escape', as
...
+test('should enable the Livewire Contour tool only once a contour segmentation exists', async ({
+  rightPanelPageObject,
+}) => {
+  const contourPanel = rightPanelPageObject.contourSegmentationPanel;
+
+  // Open the contour toolbox without creating a segmentation.
+  await contourPanel.select();
+
+  await expect(
+    contourPanel.tools.livewireContour.toolButton,
```
jbocce, line 132: With regards to my earlier comment regarding "two buttons". I suppose we have to watch for things like this too (i.e. checking for disabled). However we could get around this by adding a page object method like `isEnab led` or whatever.

tests/screenshots/chromium/LivewireContourSegmentation.spec.ts/contourWithHole.png
jbocce, line 1: This and the other screenshots have text in them. This is now a no-no. Do we  need to update the agent skills to reflect this?

tests/LivewireContourSegmentation.spec.ts
```diff
@@ -14,33 +32,339 @@ test('should cancel an in-progress Livewire contour segmentation via Escape', as
...
+  await expect(vertexHandles, 'Expected the second contour vertices to render').toHaveCount(3);
+  await activeViewport.doubleClickAt(overlappingContourClicks[2]);
+
+  // The handles clearing is the completion signal, so this settles the draw before the compare.
+  await expect(
+    vertexHandles,
+    'Expected the second contour vertex handles to clear on completion'
+  ).toHaveCount(0);
+
+  await checkForScreenshot({
```
jbocce, line 262: I think we want to check for viewport screenshot here, but shouldn't we also check for a single path too (i.e. path count is exactly one)?

tests/LivewireContourSegmentation.spec.ts
```diff
@@ -14,33 +32,339 @@ test('should cancel an in-progress Livewire contour segmentation via Escape', as
...
+  });
+});
+
+test('should keep overlapping Livewire contours in different segments separate', async ({
+  page,
+  rightPanelPageObject,
+  viewportPageObject,
+}) => {
+  // Same overlap geometry as the same-segment merge test; across segments it must not merge.
+  const overlappingContourClicks = [
```
jbocce, line 275: I think this is declared twice. Is it worth refactoring to a constant?

tests/LivewireContourSegmentation.spec.ts
```diff
@@ -14,33 +32,339 @@ test('should cancel an in-progress Livewire contour segmentation via Escape', as
...
+  await activeViewport.doubleClickAt(overlappingContourClicks[2]);
+
+  // The handles clearing is the completion signal, so this settles the draw before the compare.
+  await expect(
+    vertexHandles,
+    'Expected the second contour vertex handles to clear on completion'
+  ).toHaveCount(0);
+
+  // The overlap that merges within one segment must stay two outlines across segments.
+  await checkForViewportScreenshot({
```
jbocce, line 315: Shouldn't we also check for two paths?

tests/LivewireContourSegmentation.spec.ts
```diff
@@ -14,33 +32,339 @@ test('should cancel an in-progress Livewire contour segmentation via Escape', as
...
+    },
+  });
+
+  // The handles clearing is the completion signal, so this settles the draw before the compare.
+  await expect(
+    vertexHandles,
+    'Expected the hole vertex handles to clear on completion'
+  ).toHaveCount(0);
+
+  await checkForScreenshot({
```
jbocce, line 365: Maybe also check for one path?

tests/utils/keyboardUtils.ts
```diff
@@ -24,3 +24,31 @@ export const press = async (params: PressParams) => {
...
+  key: string;
+  action: () => Promise<void>;
+};
+
+/**
+ * Runs an action with a key held down, releasing it afterwards even if the action
+ * throws. A held modifier (e.g. 'Shift') applies to the mouse events the action
+ * performs, which is how modifier-dependent tools are driven from a test.
+ */
+export const withKeyHeld = async (params: WithKeyHeldParams) => {
```
jbocce, line 45: Please do a quick check if this could be useful elsewhere (i.e. is there some place else we are doing something similar in the tests already)?

jbocce (changes requested):
See my suggestions and comments. Thanks so much.

tests/pages/RightPanelPageObject.ts
diattamo: Got it. Yes I updated it to just use the one button.

tests/utils/keyboardUtils.ts
diattamo: I did not see anywhere where this was being performed. Other place is with the other contour tools (freehand/spline, etc)

diattamo:
@jbocce ready to be looked at again

jbocce (approved):

PR #6208, opened 2026-08-07
fix(ToolbarService): stop toolbar option wrapping from leaking closures

pedrokohler:
### Problem

`_mapButtonToDisplay` writes enhanced options back onto the button's props (`item.options = createEnhancedOptions(...)`). Consumers such as `useActiveToolOptions` read options from there, so the write-back has to stay — but it means the next refresh wraps the *previous wrapper*. Each new `onChange` closes over the old one, and the newest wrapper stays reachable from `state.buttons`, so the chain is pinned and grows by one closure layer per refresh.

`refreshToolbarState` runs on every `SEGMENTATION_MODIFIED`, so brush painting grows it continuously — the toolbar leaks in proportion to how much the user paints.

### Fix

Resolve each option back to the one the button was defined with (`WeakMap` wrapper → original), so a refresh always yields exactly one closure layer. Write-back unchanged.

### Measured

120 brush strokes on a 1315-slice labelmap segmentation with 2 segmentations, heap sampled after a forced GC:

| | before | after |
|---|---|---|
| retained after loading the next volume | 490 MB | **75 MB** |
| heap objects | 2.34M → 23.24M (**9.9×**) | 2.32M → 5.61M (**2.4×**) |
| retained `onChange` closures | 2,830,072 | **0** |
| JS heap | 493 → 1474 MB | 492 → **562 MB** |

`_mapButtonToDisplay` went from 375 MB of 470 MB sampled allocation (80%) to off the list. Same voxels painted before and after — no behaviour change.

### Tests

6 unit tests in `platform/core/src/services/ToolBarService/__tests__/`, each verified to fail against the code it guards.

An earlier revision of this PR rebuilt the wrapper from the original option, which discarded whatever an evaluate function had written onto the button props in the meantime — `synchronizeDrawingRadius` sets the live brush size there, and `handleEvaluate` runs before the options are wrapped, so the radius input reverted to its default. That showed up as 3 `SEGDrawingToolsResizing` failures on the first CI run; it is covered by two of the tests above.

### Not addressed

Nothing unsubscribes `ToolbarService._serviceSubscriptions`, and `extensions/cornerstone` registers `SEGMENTATION_MODIFIED` from both `init.tsx` and `index.tsx`. I tried deduplicating that and reverted it — `SegmentationService` is a singleton whose `onModeExit` → `destroy()` → `reset()` clears its listeners, so suppressing the re-registration on the next mode enter silently stops the toolbar updating. Duplicate listeners still cost CPU per event, but with the chain fixed they no longer leak.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed toolbar refresh behavior to prevent repeated updates from accumulating redundant actions.
  * Ensured toolbar controls consistently reflect current values after refreshes, including nested controls.
  * Preserved command execution and toolbar state notifications when options change.

* **Tests**
  * Added coverage for repeated refreshes, independent toolbar options, live values, command execution, and state updates.

### DIFF
--- a/platform/core/src/services/ToolBarService/ToolbarService.ts
+++ b/platform/core/src/services/ToolBarService/ToolbarService.ts
@@ -111,6 +111,12 @@ export default class ToolbarService extends PubSubService {
   _servicesManager: AppTypes.ServicesManager;
   _evaluateFunction: Record<string, EvaluateFunction> = {};
   _serviceSubscriptions = [];
+  /**
+   * Enhanced option -> the pristine option it was built from, so repeated
+   * refreshes re-wrap the original rather than the previous wrapper. Weak so
+   * discarded wrappers stay collectable.
+   */
+  private _pristineOptions = new WeakMap<ButtonOptions, ButtonOptions>();
 
   constructor(
     commandsManager: CommandsManager,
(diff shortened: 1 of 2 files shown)


sedghi:
@jbocce

PR #6221, opened 2026-08-20
feat(ReportDialog): say whether the save creates a new series or extends one

### LINKED ISSUE #6106: Export Dialog for Segmentation Changes

Working with some cardiologists using the segmentation mode, they found the save dialog very confusing, even when they could find it. I’d like to suggest a small change for OHIF 3.13 that should be easy to implement and could improve this workflow based on their feedback.

The main issue is that the current dialog mixes two different actions:

- Saving to an existing segmentation series.
- Creating a new segmentation series.

Currently, Create new series appears as an option inside the same dropdown as the existing series. This makes it unclear whether the user is selecting an existing target series or changing the save behavior. The text field beside it is also ambiguous because it only applies when creating a new series, not when saving to an existing one.

Proposal: Improve Segmentation Save Dialog for OHIF 3.13
1. Move the save launch button into the Segmentation side panel
The action that opens this dialog should move out of the current Export menu beside the segmentation selector and become a direct action in the Segmentation side panel.

This should follow the same pattern as the Measurements panel save button.

Recommended behavior:

Add a direct Save button to the Segmentation side panel.

Place it near the segmentation list/header, similar to the Measurements save button.

Use this button to open the revised Save Segmentation dialog.

De-emphasize or remove the current Export menu entry for this workflow.

Keep Download available as a separate action in the dialog.

The user is editing segmentations in the Segmentation panel, so saving should be launched from that same panel.

2. Replace the current mixed dropdown with two explicit save destinations
The dialog should clearly separate the two destination choices:

Save to existing series

Create new series

These should be shown as separate visible sections, using radio buttons, checkboxes, or selectable cards.

The key point is that Create new series should no longer appear as an item in the existing-series dropdown.

Recommended Dialog Layout
Save Segmentation                                      [X]

Choose where to save this segmentation.

(•) Save to existing series
    Save changes to an existing segmentation series.

    Existing series
    [ RLE LAB F2S2 20260521 / undefined 3102 v ]

    The existing series name cannot be changed here.

( ) Create new series
    Create a separate segmentation series.

    New series name
    [ RLE LAB F2S2 20260521 copy ]

[Download]                                      [Save]
Interaction Rules
Dialog close/cancel behavior
The dialog does not need a separate Cancel button.

Closing the dialog with the X cancels the operation.

Default selection
Save to existing series should be selected by default when a suitable existing segmentation series is available.

Create new series should be selected by default when no suitable existing series is available.

Save to existing series
When Save to existing series is selected:

The existing-series dropdown is enabled.

The new-series name field remains visible but disabled.

The dialog should state that the existing series name cannot be changed here.

Saving should update or add to the selected existing segmentation series.

The existing series name should not be editable in this path because it is a series-level attribute.

Create new series
When Create new series is selected:

The existing-series dropdown may remain visible but inactive, or remain enabled only as a source for the default new name.

The new-series name field becomes enabled.

The new-series name should be initialized from the selected/current series name.

The user can edit the new-series name before saving.

Saving should create a separate segmentation series.

Auto-selection behavior
The Create new series option should be automatically selected when the user interacts with the new-series section.

Specifically:

Clicking the Create new series checkbox/radio button selects it.

Clicking the Create new series label selects it.

Clicking the explanatory text or section body selects it.

Clicking into the new-series name field selects it.

Starting to type in the new-series name field selects it.

The new-series name field should be visible even when disabled, so the user can see where naming happens, but it should only become editable after Create new series is selected.

Save and Download buttons
The dialog should include:

Download

Save

These actions should remain distinct.

Save stores the segmentation using the selected destination option.

Download downloads/exports the segmentation result as a file and should not imply that the result has been saved back to the study/server.

The Download button should not change the selected save destination.

Why this change is useful
This is a small UI change, but it removes an important ambiguity.

The current dialog makes Create new series look like one of the existing series choices. The proposed layout makes the user’s decision explicit:

Am I saving to an existing series?

Or am I creating a new series?

It also avoids implying that the series name can be edited when saving to an existing series. The new-series name field is only editable for the new-series workflow, where editing the name is valid.

This should make the segmentation save workflow easier to understand without requiring a major redesign of the underlying storage behavior.

Suggested OHIF 3.13 scope
Suggested 3.13 scope:

Add a direct Save button to the Segmentation side panel.

Rename the dialog to Save Segmentation if needed.

Split the dialog into Save to existing series and Create new series sections.

Move Create new series out of the existing-series dropdown.

Show the new-series name field as visible but disabled until Create new series is selected.

Auto-select Create new series when the user clicks or types in the new-series section.

Keep Download and Save as separate actions.

This should be implementable as a focused UX improvement without changing the larger segmentation architecture.

Confusing for users and it is easy to implement using AI

wayfarer3130:
### Context

Fixes #6106.

The save dialog mixed two different operations into a single series drop down -
creating a new series, and storing into a series that already exists. Which one
was about to happen was not visible, and the description field beside the drop
down silently applied to only one of them (it was disabled, without saying why,
when an existing series was selected).

### Changes & Results

The destination is an explicit choice of three, built to @dan-rukas's design in
[this comment](https://github.com/OHIF/Viewers/pull/6221#issuecomment-5542970808):
a two-column form whose `Series` row is a segmented control, with a line of help
under it and the same words repeated on the button that commits the save.

- **Save to current** adds a version to the series the data was loaded from. The
  default when there is such a series.
- **Save as new** creates a separate series, with no predecessor. Its series
  number (offered as one past the existing series of this modality) and
  description are editable. The default otherwise.
- **Replace existing** adds a version to another loaded series of this modality,
  chosen from a select of their descriptions. This is what the old drop down
  could do, now as its own destination rather than an entry mixed in among the
  series.

A series that already exists keeps its own number and description, so both are
read-only for `Save to current` and `Replace existing`, and the number follows
the series picked. The description and number rows are shared and only switch
between read-only text and an editable control.

All three store the **same object**: all of the current data, as selected from
the service holding it. The destination only decides which series that object
belongs to, and so which instance it supersedes. `Replace existing` does not
merge - it neither loads what the chosen series already holds to add to it, nor
leaves any of the current data out. The instance it supersedes is kept and stops
being the one loaded by default. Since storing into a series makes that series
the one the data was last stored as, the save after a `Replace existing` defaults
to `Save to current` on the series that was replaced.

Three things to check: a destination with no series to work with is **disabled
rather than hidden** (with a `title` saying why), so the control does not change
shape between studies; `Replace existing` **excludes** the series the data was
loaded from, since `Save to current` is that same operation on that series; and
the series it offers come from the whole display set cache, so with a prior study
loaded it can list a series of another study - as the old drop down could. Worth
deciding whether to scope that list to the study being saved.

The footer is now a `FooterAction` with `Download` on the left and `Cancel` and
the primary on the right, so the dialog has a Cancel button for the first time
rather than only the close box. The prompt also asks for `max-w-lg`: the
destination control is wider than the dialog service's default `max-w-md` and was
sticking out past the dialog's edge.

**Series descriptions are remembered**, per type of item being stored (`SEG`,
`RTSTRUCT`, `SR`), in local storage:

- the field is prefilled with the one last used, falling back to the name of the
  thing being saved (the segmentation name, `Contours` for a contour export,
  `Measurements` for a report);
- a pull down offers that provided name first, then the remembered ones, most
  recent first;
- typing narrows the list to what it can complete, **Tab** completes, and the
  arrow keys plus Enter pick one;
- how many to remember is a provided count, **5** by default, and **0** turns the
  whole feature off - nothing stored, nothing offered, no pull down.

`createReportDialogPrompt` gained `defaultSeriesDescription`, `itemType` and
`rememberedDescriptionCount` inputs (all optional), and a `seriesNumber` output
so that an edited series number reaches the encoder. `priorSeriesNumber` is now
`seriesNumber - 1`, so callers computing `1 + priorSeriesNumber` keep working and
pick up the edit.

Two behaviors changed as a side effect, both fixing existing oddities:

- a measurement report saved without typing a name was stored with the
  **dialog's title** as its series description (`Create Report`); it now uses
  `Measurements`.
- a `predecessorImageId` that does not belong to a loaded series used to select
  an existing series the dialog could not describe; it now falls back to
  creating a new series.

The dialogs are renamed for the `Save` action in them: **Save Segmentation**,
**Save Contours** and **Save Measurements**, from `Store Segmentation`, `Store
Contours` and `Create Report`. A mode that sets its own `defaultSaveTitle` is
unaffected. With the save already launched from the Segmentation side panel, that
completes #6106.

The menus that launch them say **Save** rather than `Export`, which reads as
taking a copy away rather than storing back into the study: the segmentation
panel's submenu holding `DICOM SEG` and `DICOM RTSS`, and the measurements
panel's `Create SR` button. The `Save` key is added to en-US and the test locale
for both namespaces; other languages fall back to it until translated.

**The object just saved becomes the predecessor of the next save.** Extending a
series relies on the data knowing which instance it was last stored as, and that
was only known for data loaded from a store - so the first save created a new
series, and so did every save after it. An instance stored from the viewer is now
identified the way a loaded one is (`registerStoredInstanceImageId` gives it the
imageId that loading it back would use, and maps that imageId to its UIDs, before
it reaches the metadata store), so the display set made from it has a
`predecessorImageId`. Each save type then records it where it belongs:

- segmentations and contours are reloaded from the series just written, so the
  reloaded segmentation takes it from that display set through the normal load
  path;
- measurements are not reloaded from their report, so the new
  `recordMeasurementsPredecessor` command (`CORNERSTONE`) records it on them
  directly - on the measurement and on the annotation it is derived from, so
  editing a measurement afterwards does not lose it.

Saving the same data twice therefore defaults to **Save to current** the second
time, pointing at the series the first save created.

**Segmentations were being stored into a new series however the dialog was
answered** - always one called `Research Derived series` numbered 99, with no
predecessor sequence. Those are dcmjs' `DerivedDataset` defaults, which is the
tell: the SEG adapter's `generateSegmentation` resolves the predecessor's series
data and then assigns it to the derivation object it returns rather than to the
`dataset` inside it, and the dataset is what gets stored
([5.8.2, both call sites](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts)).
The SR and RTSTRUCT paths assign to the dataset and were unaffected. This PR
applies the predecessor to the dataset after generating it, so the instance lands
in the chosen series with that series' number and description. It is the same data
the adapter resolves, so it stays correct once the adapter is fixed upstream -
**that upstream one-liner is still outstanding**, and 5.8.2 still has the bug.

Screenshots to follow.

### Dependencies

The cornerstone3D packages move from **5.6.12 to 5.8.2** (core, tools, adapters,
ai, dicom-image-loader, metadata, labelmap-interpolation,
polymorphic-segmentation), and the codecs move to the versions
`dicom-image-loader@5.8.2` itself depends on (charls 1.2.5,
libjpeg-turbo-8bit 1.2.4, openjpeg 1.3.2, openjph 2.4.9) rather than to their
newest published versions. The codecs were behind the loader's own dependencies,
so pnpm was installing the loader's private copies of three of them alongside the
ones the direct imports use; matching the loader leaves one copy of each. That
includes raising the `codec-openjpeg` override in `pnpm-workspace.yaml`, which
holds the whole tree to a single version.

### Testing

Unit tests: `pnpm run test:unit` - **1191 tests in 100 suites pass**, including 33
new ones: 23 on the dialog (each destination and its availability, the editable
series number and description, the remembered descriptions, download and cancel),
6 on identifying a stored instance, and 4 on the measurement save recording its
predecessor. `pnpm run build` compiles on 5.8.2 with no errors.

The store fix itself has no unit test - the `cornerstone-dicom-seg` extension has
no jest project, and a meaningful test would need the adapter, the store function
and the segmentation service all stubbed - so step 2 below is what confirms it.

In the viewer:

1. Open a study in the segmentation mode, create a segmentation, and save it as
   DICOM SEG. The dialog should start on **Save as new** with **Save to current**
   disabled, show the next series number as editable, and offer the segmentation
   name as the description. Save it.
2. Save the same segmentation again. It should now start on **Save to current**,
   showing that series' number and description read-only, and saving should add an
   instance to **that** series - not a second `Research Derived series` / 99 - with
   a `PredecessorDocumentsSequence` in the stored object.
3. Switch to **Replace existing**: the select should list the other loaded series
   of the modality (not the one just saved into), the number should follow the
   series picked, and the save button should be disabled until one is.
4. Back on **Save as new**, open the pull down beside the description: the
   segmentation name first, then the descriptions used before. Type a prefix of
   one of them and press Tab to complete it.
5. Repeat with `DICOM RTSS`, and with measurements: save a report, then save it
   again and check that it starts on **Save to current**. Each type should
   remember its own descriptions.

#### PR

#### Code

  etc.)

  additions or removals.

A `3.13 -> 3.14` migration guide is added, with a page covering the destination
change, the new inputs and outputs, and the remembered descriptions.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 24.2.0
- [ ] Browser: not yet exercised in a browser - unit tested only, so the steps
  above still need a run through the viewer.

## Summary by CodeRabbit

- **New Features**
  - Choose whether to save measurements, segmentations, and contours to the current series, a new series, or replace an existing series.
  - Edit series numbers and descriptions, with remembered descriptions available by item type.
  - Preserve relationships between measurements and saved reports.
  - Improve handling of stored DICOM instances across data sources.

- **Bug Fixes**
  - Apply modality-specific save defaults and improve reliability with incomplete image references.

- **Documentation**
  - Added migration guidance for the updated save workflows.


wayfarer3130:
@dan-rukas Can you review the appearance - it looks like:
image

and then:

image

wayfarer3130:
@dan-rukas - I'm also passing this by a clinical user to make sure it fixes the reported defect and is clear/usable.

dan-rukas:
Here's an updated and simpler design for the dialog. 

First is a large image of the three dialog states and layouts followed by some of the component and behavior details:

dialog-states

### Layout

The dialog keeps the standard title from the dialog service (`Save Measurements`, `Save Segmentation`, `Save Contours`). The body is a two-column form: a label on the left, the control on the right. Three rows, in this order:

1. **Series** – the destination choice
2. **Series Description**
3. **Series Number**

### Components

- **Dialog** – existing `Dialog` / `DialogHeader` / `DialogTitle` via `uiDialogService`. Nothing new here.
- **Series** – `Tabs` / `TabsList` / `TabsTrigger` used as a segmented control with three options: `Save to current`, `Save as new`, `Replace existing`. No `TabsContent`; the rows below are shared and only change between read-only text and an editable control. Directly under the tabs is a one-line helper paragraph in `text-muted-foreground` whose text changes with the selected tab (see below).
- **Series Description** – depends on the tab:
  - Save to current: read-only text showing the current series description.
  - Save as new: `Input` with placeholder `Series description`.
  - Replace existing: `Select` (`SelectTrigger` / `SelectContent` / `SelectItem`) listing the loaded series of this modality, placeholder `Series description`.
- **Series Number** – depends on the tab:
  - Save to current: read-only text (the current series number).
  - Save as new: `Input`, prefilled with the next series number.
  - Replace existing: read-only text, updates to the number of the series picked in the Select.
- **Footer** – `FooterAction` with `FooterAction.Left` holding `Download` as a `FooterAction.Secondary`, and `FooterAction.Right` holding `Cancel` (`FooterAction.Secondary`) and the primary action (`FooterAction.Primary`). Today the dialog uses `InputDialog.Actions`, which only gives a right-aligned cluster, so this is the one structural change in the footer: Download sits on the left, Cancel and the primary on the right.

### State changes

The helper text and the primary button label both follow the selected tab. The primary button always repeats the tab label so the user reads the same words twice before committing.

| Tab | Helper text | Description row | Number row | Primary button |
|---|---|---|---|---|
| Save to current | Adds a new version to this series as the new default. Earlier versions are kept. | read-only text | read-only text | **Save to current** |
| Save as new | Creates a separate series. | `Input` | `Input` | **Save as new** |
| Replace existing | Choose a series to replace. Current measurements become default and earlier versions are kept. | `Select` of loaded series | read-only, follows the Select | **Replace existing** |

`Download` and `Cancel` do not change between states.

### Mapping to the current PR

- `Save to current` is the PR's **Extend Existing**.
- `Save as new` is the PR's **New Series**.
- `Replace existing` is new relative to this branch. It brings back the ability the old dropdown had, choosing any loaded series of the modality as the destination, but as its own explicit mode with the same "new default, earlier versions kept" behavior as Save to current. The Select replaces the toggle button and tooltip that switch modes today.

wayfarer3130:
@dan-rukas - can you approve now?  I applied your most recent change.

dan-rukas (approved):

PR #6222, opened 2026-08-20
fix(sorting): order display sets by the creation date/time of their instance, and show it

wayfarer3130:
### Context

Derived series are listed after the images in reverse date/time order, so the most recent
report or segmentation is the one nearest the images. In a study holding several SEGs and
SRs saved over a session, they did not interleave: all the SEGs clustered above all the SRs
regardless of when each was created, and the SEGs' order among themselves was arbitrary.

Three separate causes.

**1. There was no single answer to "when was this display set created".** `compareSeriesDateTime`
joined `SeriesDate` and `SeriesTime` into one string, so every SEG compared as
`"20260817 undefined"` — and `"u"` sorts above any digit, making a SEG look newer than
every dated series of the same day. Two SEGs from the same day tied outright and fell
through to reverse `SeriesInstanceUID`.

Each handler had its own fallback chain instead, and they disagreed: SR passed
`SeriesDate`/`SeriesTime` straight through; RTSTRUCT preferred `StructureSetDate` and
`StructureSetTime`, resolved **independently**, so a structure set with a `SeriesDate` but
no `SeriesTime` took its time from the structure set — possibly a different day, giving a
timestamp that never existed; SEG, PMAP, video, PDF and chart passed no time at all.

None of them used the instance date/time, which is the only date/time that moves when an
object is saved into an existing series: every instance of a series carries that series'
`SeriesDate`/`SeriesTime`, so a second report saved into an existing SR series still reports
the date the series was first created.

**2. The same-series compare never decided anything.** `compareSameSeriesDisplaySet`
returned the registered comparison only when it was **zero**:

```js
if (!compareValue) {
  return compareValue;
}
```

so a real answer was discarded and `sortByInstanceNumber` decided instead. That makes
`addSameSeriesCompare` inert for ordering, and the `sortVector` that
`notes-requirements.md` documents as "display sets which match on series instance uid are
then compared using the sort vector" has no effect either.

`sortByInstanceNumber` has the same shape of bug in its guard clause:

```js
return (!a && !b && 0) || (!a && -1) || 1;
```

Its own `0` is falsy, so two missing instances fall through it to `-1` — meaning a pair of
display sets that both lack an instance compares as `-1` in **both** directions. An
inconsistent comparator makes `Array.prototype.sort` produce an order that depends on the
starting order.

The existing `sortStudy.test.js` passed only because of that: with every comparison
answering `-1`, `sort` reverses the input, and the fixture happened to be supplied in
reverse. Feeding the same five display sets in a different starting order returned
`ds4, ds3, ds2, ds1, ds5`.

**3. Newly saved objects were not identifiable as the most recent of their series.** The
adapters derive a stored report's instance number as `1 +` that of the one predecessor
instance it was given. The most recently created instance of a series is not necessarily
the one with the highest instance number, so that can collide with an instance that already
exists — and then the handlers' "last instance is the newest one" no longer holds. Nothing
set an instance level date/time on the way out either, so there was nothing to fall back to.

### Changes & Results

**One rule for the date/time — `platform/core/src/utils/seriesDateTime.ts`.**
`getSeriesDateTime` chooses a single date/time from every attribute an instance carries:
`InstanceCreationDate`/`Time`, `ContentDate`/`Time`, `AcquisitionDate`/`Time` (or the
combined `AcquisitionDateTime`), `StructureSetDate`/`Time`,
`PresentationCreationDate`/`Time` and `SeriesDate`/`Time`.

- The date is the latest date found in any of them.
- The time is the latest time found in an attribute carrying **that exact same date**, so a
  time is never combined with a date it did not arrive with and the result is always a
  date/time that really occurred.
- With no time for the winning date the time is empty, and the ordering is accurate to the
  day — as good as the data allows.
- `StudyDate`/`StudyTime` are excluded: every series in the study shares them, so they
  cannot tell one series from another.
- A value that is not a DICOM DA counts as no date at all. Some series level metadata carries
  a date already formatted for display, and `19-Jan-2026` would otherwise read as `192026` —
  ordering by day of month, and making two different months compare as equal.

**Sorting — `platform/core/utils/sortStudy`.**

- *Instances* sort by increasing instance number, as before. Only when the instance numbers
  do not decide — they tie, or neither has one — does the sort fall back to the creation
  date/time, and then to the sop instance uid. The last instance of a series is taken to be
  the most recently created one, so an instance number that fails to say which that is has
  to be replaced by something that does.
- Two frames of one instance are excluded from that fallback: they share the single date/time
  their instance has, so only the frame number orders them. They are recognised by their
  shared sop instance uid, which is also what keeps a large multi frame series cheap to sort —
  its frames all carry the same instance number, so every pair reaches the fallback.
- *Display sets* are ordered by the creation date/time of the instance they show, which they
  carry as `instance`.
- *Series*, and display sets whose instance says nothing about when it was created, have
  nothing but their own `SeriesDate`/`SeriesTime` and are ordered by those alone — so
  sorting a list of series rather than display sets stays the plain series date/time sort it
  always was.
- The registered same-series comparison is returned when it is non-zero, and two missing
  instances are treated as equal.
- `processSeriesResults` in the DICOMweb data source sorts on the raw `SeriesDate` and
  formats it for display afterwards. It was sorting series rows whose date had already been
  formatted, which no date rule can order.

**Handlers** — `cornerstone-dicom-seg`, `cornerstone-dicom-rt`, `cornerstone-dicom-sr`,
`cornerstone-dicom-pmap`, `dicom-video`, `dicom-pdf` and the chart SOP class handler in
`extensions/default` take their display set's `SeriesDate`/`SeriesTime` from
`getSeriesDateTime` on the instance they chose, replacing seven different fallback chains.
The RTSTRUCT handler's existing comment already asked for this — "the display set date …
should be the date of the instance being used" — and the date shown in the study panel is
now the same one the sort uses.

**Saving — `platform/core/utils/updateNewInstanceMetadata`**, called from `storeMeasurements`
and `storeSegmentation`. Every report, segmentation and structure set OHIF stores is stamped
with the current `InstanceCreationDate`/`Time` and `ContentDate`/`Time`, and with an
instance number one higher than **every** instance already in its series, read from
`DicomMetadataStore`. The adapters apply their predecessor metadata after the caller's
options, so the instance level stamp has to be corrected on the way out rather than passed in.

Both are read as wall clock values in the object's own timezone —
`TimezoneOffsetFromUTC` when it declares one, the local zone otherwise — since DICOM DA and
TM are displayed exactly as they are stored.

**The series level date/time is passed in at generation time instead.** It cannot be stamped
alongside the instance one: an object added to an existing series has to keep that series'
own date/time, and `updateNewInstanceMetadata` cannot tell the two cases apart. So the store
commands hand the generation the values it should use, from `getCurrentDicomDateTime` and in
the same zone: `SeriesDate`/`SeriesTime` when a new series is being created, and
`StructureSetDate`/`StructureSetTime` for every structure set, since that pair is the
creation stamp of the structure set itself whichever series it goes into. `ContentDate`/`Time`
go with them for a SEG, whose download path does not reach the stamp at all.

Both dcmjs (`DerivedDataset`) and the adapters (`createInstance`) honour these as options and
otherwise default them to UTC — which west of UTC is a day ahead around midnight, and then
wins the latest date `getSeriesDateTime` reads. That defeated the instance stamp for every
new series: a report saved at 20:00 on 19 Aug in New York got `SeriesDate` 20 Aug from dcmjs
against its own `ContentDate` of 19 Aug, and read as a day ahead in both the sort and the
thumbnail.

**`generalImageModule`** in `platform/core/classes/MetadataProvider` now carries
`sopClassUID`, `instanceCreationDate`/`Time` and `contentDate`/`Time`.
`@cornerstonejs/metadata` lists these among the module's attributes, and derived instance
creation and the predecessor reference both read them, but OHIF supplied none of them.

**Showing the date/time the order comes from — `studyBrowser.thumbnailDetails`.** With the
sort fixed, the study browser still gave no way to tell that it was right: the thumbnail
detail line shows the series number and the instance count only, so several reports or
segmentations saved on the same day read as identical and their order looks arbitrary — the
same complaint the sort itself started from. The date/time being sorted on has to be visible
for the ordering to be checkable at all, which is what this customization is for.

It declares what goes on that line the same way the viewport overlay items declare theirs: a
list of items, each with an `id`, an optional `condition` deciding whether to include it, and
a value taken from its own `contentF`, from a named `source`, or from an `attribute` of the
instance the display set shows. `label` prefixes the value, `title` is its tooltip, and
`iconName` puts an icon before it. An item with no value is left out.

`condition` and `iconName` may each be a function **or** a name — resolved against
`studyBrowser.thumbnailDetailTests` and `studyBrowser.thumbnailDetailSources` — so the whole
line can be declared as data, which is what a `?customization=` file is limited to: those are
fetched and parsed as JSONC and never executed, so a function-only design could not be driven
from the URL at all.

The default is the series number and the instance count, exactly as the thumbnails have
always shown, so nothing changes until someone overrides it. That default lives in the
thumbnail itself, so it is reachable only while nothing is resolved for it: with no items to
resolve `resolveThumbnailDetails` answers with nothing at all rather than with no items,
whereas an empty list is a customization asking for an empty line and is honoured as one.
Named sources ship as
`seriesNumber`, `numInstances`, `seriesDate` and `instanceDateTime`; the one named test is
`isDerivedDisplaySet`.

`platform/app/public/customizations/studyBrowser/derivedDateTime.jsonc` is the example,
written to match the veterinary demographics overlay — a `global` payload, `$push` so the
default items are kept rather than replaced, and names rather than functions:

```jsonc
{ "id": "InstanceDateTime", "source": "instanceDateTime",
  "condition": "isDerivedDisplaySet", "title": "Created" }
```

`instanceDateTime` reads `getSeriesDateTime` off the display set's `instance` — the same value
the sort uses — and formats it to `DD-MMM-YYYY HH:mm`. No seconds and nothing below them: the
second a report was written says nothing a reader can use, and is not reliably recorded
either.

The panel resolves the items and hands the thumbnail plain data, so `resolveThumbnailDetails`
is a pure function and `ui-next` stays presentational — only one file there imports
`@ohif/core` today, with a note about why that is fragile. Resolution sits in a wrapper around
the display set mapper in `PanelStudyBrowser`, which covers all three of its call sites and
the measurement tracking panel's own mapper.

| | study browser thumbnail |
|---|---|
| default | `SEG  Segmentation` / `S:42  ⧉ 1` |
| `?customization=studyBrowser/derivedDateTime` | `SEG  Segmentation` / `S:42  ⧉ 1  13-Sep-2022 16:35` |

Before / after for a study with an MR series, SRs at 09:00 and 14:00, and SEGs at 09:30 and
13:30:

```
before   MR | SEG 13:30 | SEG 09:30 | SR 14:00 | SR 09:00
after    MR | SR 14:00  | SEG 13:30 | SEG 09:30 | SR 09:00
```

Note the same-series fix changes ordering for anyone who registers an `addSameSeriesCompare`
function — that registration now takes effect, where before it was ignored.

### Testing

`jest` at the root — 105 suites / 1268 tests pass. `tests/ThumbnailDetails.spec.ts` passed
against the e2e server when it was added; the changes since then are to the sort rules and to
the metadata generated on save, neither of which it exercises.

- `platform/core/src/utils/seriesDateTime.test.js` covers the date/time choice: the latest
  date, the latest time of the winning date, a time borrowed from another attribute of the
  same date, never taking a time from a different date, an ignored study date, a date that is
  not a DICOM DA (and the dotted retired form, which is one), the combined
  `AcquisitionDateTime`, the lower camel case spelling of series metadata, and the sort key's
  padding and ordering.
- `platform/core/src/utils/sortStudy.test.js` covers display sets ordered by their
  instance's creation date/time, series with no instance ordered by the series date/time, the
  fallback when the instance has no date, `compare` symmetry, the instance number taking
  precedence over the creation date/time and the fallback when it ties, the frames of one
  instance ordering by frame number, sources with no sop instance uid keeping the order they
  came in, and derived series interleaving. Reverting `sortStudy.ts` alone fails its cases.
- `platform/core/src/utils/updateNewInstanceMetadata.test.js` covers the next instance number
  over a series whose highest number is not the last one, the first instance of a new series,
  and the stamped date/time being what `getSeriesDateTime` then reports.
- `extensions/default/src/SOPClassHandlers/chartSOPClassHandler.test.ts` asserts the date and
  time reach the display set, including the instance date winning over an older series date.
- `extensions/default/src/Panels/StudyBrowser/resolveThumbnailDetails.test.ts` covers the
  detail items: the default coming to the series number and the instance count, each of the
  three ways an item gets its value, a named and a function `condition`, an item named after a
  test that is not registered, no items resolving to nothing at all while an empty list
  resolves to an empty line, and the `instanceDateTime` source reporting the instance's
  creation date/time to the minute rather than the series one.
- `platform/ui-next/src/components/Thumbnail/Thumbnail.test.ts` renders both view presets and
  asserts the detail line still reads `S:<series number>` and the instance count when no
  `details` are supplied, and the supplied items in order when they are.
- `tests/ThumbnailDetails.spec.ts` opens a study with a SEG and checks the default item ids,
  then loads `?customization=studyBrowser/derivedDateTime` and checks that the date/time is
  appended and formatted to the minute.

Two small things were needed for the component test: `platform/ui-next/src/__mocks__/
fileMock.js`, since the shared jest config maps every static asset there and this package had
no copy, so any test reaching the Icons barrel failed to resolve; and the test is a `.test.ts`
using `createElement`, because the jest projects' `testMatch` does not include `.tsx`. Happy
to widen `testMatch` instead if that is preferred.

For the display set changes: open a study that has both a SEG and an SR saved at different
times of the same day (saving one of each from the viewer is enough) and check the study
panel — the derived series read newest-first as a single group rather than SEGs first. Then
save a second report into an existing series and confirm it moves to the front of the group.
Adding `&customization=studyBrowser/derivedDateTime` shows the date/time each of them is
ordered by.
Of the handlers touched, only the chart one lives in a package with a jest project, so it is
the one with a direct handler test; `cornerstone-dicom-seg`, `cornerstone-dicom-rt`,
`cornerstone-dicom-sr`, `cornerstone-dicom-pmap`, `dicom-video` and `dicom-pdf` have none
upstream, and the change in each is now the same two lines. Happy to add the project configs
if wanted. The same goes for the two store commands: the date/time they pass to the object
generation has no unit test, and is verified by reading the paths it flows through —
`DerivedDataset.assignFromOptions` covers `SeriesDate`/`SeriesTime`/`ContentDate`/`ContentTime`,
`createInstance` assigns `StructureSetDate`/`Time` because `RTSS_INSTANCE_DATA` already
carries those keys, nothing in `DerivedPixels`, `StructuredReport` or the SEG adapter copies a
date from the reference dataset, and the predecessor `Object.assign` that restores an existing
series' own date/time runs afterwards and carries no structure set stamp.

#### PR

- [x] My Pull Request title is descriptive, accurate and follows the semantic-release
  format and guidelines.

#### Code

- [x] The documentation page has been updated as necessary for any public API additions or
  removals. (`notes-requirements.md` gains a "Display Set Date and Time" section stating the
  instance, display set and series sort rules; `sampleCustomizations.tsx` gains entries for
  `studyBrowser.thumbnailDetails`, `studyBrowser.thumbnailDetailSources` and
  `studyBrowser.thumbnailDetailTests`.)

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 24.2.0
- [x] Browser: Chromium, via the e2e server for the study browser change; the sort changes are
  covered by unit tests (jsdom) and change no rendering

## Summary by CodeRabbit

- **New Features**
  - Study Browser thumbnails now support customizable detail lines, including series number, instance count, and derived-series date/time.
  - Newly saved segmentations and measurements receive current date/time and sequential instance metadata.

- **Bug Fixes**
  - Improved ordering of series and instances using reliable date/time and instance-number comparisons.
  - Display-set dates and times now consistently reflect the most relevant available DICOM attributes.

- **Documentation**
  - Added guidance for configuring thumbnail details and derived-series date/time display.


wayfarer3130:
@jbocce - this is just some series sorting issues for saving SEGs multiple times on the same day (or RTSTRUCT) - they ended up randomly ordered.

extensions/cornerstone-dicom-seg/src/getSopClassHandlerModule.ts
```diff
@@ -266,6 +268,12 @@ function _getDisplaySetsFromSeries(
     SeriesDescription,
     SeriesNumber,
     SeriesDate: SeriesDate || StructureSetDate || '',
+    // Derived modalities are sorted in reverse date/time order, comparing
+    // `SeriesDate SeriesTime` as a single string. Without a time here, every SEG
+    // compared as `<date> undefined`, which sorts above every dated series of the
+    // same day and ties SEGs with each other, so SEG and SR never interleaved by
+    // when they were created.
+    SeriesTime: SeriesTime || StructureSetTime || '',
```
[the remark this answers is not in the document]
jbocce: Not sure if this is pertinent or not. Please have look.

jbocce (dismissed):
Please check the code rabbit comment I flagged. Thanks.

extensions/cornerstone-dicom-seg/src/getSopClassHandlerModule.ts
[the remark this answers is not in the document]
wayfarer3130: Good catch, and thanks for flagging it @jbocce - it was pertinent, so it is fixed now, along with the more general problem behind it.

The mixing was real: `SeriesDate || StructureSetDate` and `SeriesTime || StructureSetTime` were resolved independently, so a SEG with `SeriesDate` but no `SeriesTime` took its time from the structure set, which may belong to a different day. CodeRabbit's example produced `20260818 090000`, a timestamp that never existed.

Rather than pick a fixed pair per modality, there is now one rule for all of them, in `platform/core/src/utils/seriesDateTime.ts`. `getSeriesDateTime` chooses a single date/time from every attribute the instance carries - `InstanceCreationDate`/`Time`, `ContentDate`/`Time`, `AcquisitionDate`/`Time` (or the combined `AcquisitionDateTime`), `StructureSetDate`/`Time`, `PresentationCreationDate`/`Time`, `SeriesDate`/`Time`:

- the date is the latest date in any of them;
- the time is the latest time in an attribute carrying **that exact same date**, so a time is never combined with a date it did not arrive with;
- with no time for the winning date the time is empty, and the ordering is accurate to the day - as good as the data allows;
- `StudyDate`/`StudyTime` are excluded, as every series in the study shares them.

So the reported case now yields `20260818` with **no** time instead of a synthetic timestamp. That is covered by `never takes a time from a date other than the winning one` in `platform/core/src/utils/seriesDateTime.test.js`, together with the neighbouring cases for the latest date, the latest time of the winning date, and a time borrowed from another attribute of the same date.

You are right that the RTSTRUCT handler selected date and time independently too - it did `StructureSetDate || SeriesDate` and `StructureSetTime || SeriesTime`. Both handlers now go through the same function, so neither can mix sources.

While in here, the wider problem: the date/time a display set is ordered by is the creation date/time of the instance it shows (`displaySet.instance`). Every instance of a series carries that series' `SeriesDate`/`SeriesTime`, so a report saved into an existing series has the date of the series as it was first created, and only the instance level date/time say when the report was made. Instances themselves still sort by increasing instance number, falling back to the creation date/time only when the instance numbers do not decide. Series with no instance keep the plain series date/time sort. Details in `platform/docs/docs/development/notes-requirements.md`.

wayfarer3130:
@dan-rukas one for you to weigh in on: should the date/time be on by default?

Context — with the sort fixed, the study browser gave no way to see that it *was* fixed. The
thumbnail detail line shows the series number and the instance count only, so several reports
or segmentations saved on the same day read as identical and their order still looks
arbitrary, which is the complaint this PR started from. So the last commit makes that line
customizable (`studyBrowser.thumbnailDetails`, declared like the viewport overlay items) and
ships an example that appends the creation date/time the sort actually uses:

| | study browser thumbnail |
|---|---|
| default | `SEG  Segmentation` / `S:42  ⧉ 1` |
| `?customization=studyBrowser/derivedDateTime` | `SEG  Segmentation` / `S:42  ⧉ 1  13-Sep-2022 16:35` |

I kept the default as it is today so this PR changes nothing visually, but I think there is a
good case for the date/time version being the default, at least for derived series:

- it is the field the list is ordered by, and an order whose key is invisible is one nobody
  can trust or report a bug against;
- it is the one thing that distinguishes two reports on the same series — the series number
  and description are identical for both;
- it costs one line of already-loaded metadata, and the detail line has room for it in both
  view presets.

Against: it is a visible change to every deployment's study browser, and for plain image
series it mostly repeats the series date already shown elsewhere — which is why the example
restricts it to derived series via `condition: 'isDerivedDisplaySet'`.

Flipping it is a one-line change to the default items in
`extensions/default/src/customizations/thumbnailDetailsCustomization.ts` — the same item the
example JSONC pushes. Happy to do it in this PR or leave it for a follow-up; your call on
whether it wants a design review first.

platform/core/src/utils/seriesDateTime.ts
```diff
@@ -0,0 +1,153 @@
...
+ * never compares above a fraction of the next one.
+ */
+const timeSortKey = (value): string => {
+  const [whole = '', fraction = ''] = `${value ?? ''}`.split('.');
+  const digits = whole.replace(/[^0-9]/g, '').slice(0, 6);
+  if (!digits) {
+    return '';
+  }
+  const fractionDigits = fraction.replace(/[^0-9]/g, '').slice(0, 6);
+  return `${digits.padEnd(6, '0')}.${fractionDigits.padEnd(6, '0')}`;
```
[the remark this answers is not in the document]
wayfarer3130: There just isnt' any good way of knowing the UTC value for anything else so this isn't obvious how to do - just dropping the utc is as good as we can do right now.

extensions/cornerstone-dicom-sr/src/getSopClassHandlerModule.ts
```diff
@@ -103,14 +103,17 @@ function _getDisplaySetsFromSeries(
...
     ConceptNameCodeSequence,
     SOPClassUID,
     imageId: predecessorImageId,
   } = instance;
   validateSameStudyUID(instance.StudyInstanceUID, instances);
 
+  // The date/time of the display set is that of the instance it shows.  A
+  // report saved into an existing series keeps the original SeriesDate, so only
+  // the instance level content date/time places it as the newest one.
+  const { SeriesDate, SeriesTime } = utils.getSeriesDateTime(instance);
```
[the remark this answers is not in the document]
wayfarer3130: Series date/time is invariant across all instances of hte series.  ONLY the per-instance date/time can be updates, and I think that one is retrieved from the instance anyways.

wayfarer3130:
@jbocce - I updated the PR, it was indeed getting the wrong behaviour and that actually brought up a couple of other issues so I added a customization to allow displaying the sorting information in the view so it is obvious what is happening.  We can decide later to include that as a default, or can leave it as is.

extensions/cornerstone-dicom-seg/src/getSopClassHandlerModule.ts
```diff
@@ -258,14 +256,22 @@ function _getDisplaySetsFromSeries(
     imageId: predecessorImageId,
   } = instance;
 
+  // The date/time of a display set is the date/time of the instance it shows,
+  // chosen from all the attributes that instance carries - for a SEG that is
+  // typically the content or structure set date/time rather than the series
+  // one, and for a SEG saved into an existing series only the instance level
+  // date/time reflects the save.
+  const { SeriesDate, SeriesTime } = utils.getSeriesDateTime(instance);
```
jbocce, line 264: **1. SEG date selection is safe, but worth a test.** The concern about pairing a `SeriesDate` with a `StructureSetTime` from a different day can't actually happen here: `getSeriesDateTime` only takes a time from an attribute carrying the same winning date (see `consider()` in `seriesDateTime.ts`). No code change needed. But there's no test for a SEG whose `SeriesDate` and `StructureSetDate` differ — could we add one asserting the later date wins with its own time (or none)?

extensions/default/src/Panels/StudyBrowser/PanelStudyBrowser.tsx
```diff
@@ -73,7 +74,39 @@ function PanelStudyBrowser({
     setViewPresets(newViewPresets);
   };
 
-  const mapDisplaySetsWithState = customMapDisplaySets || _mapDisplaySets;
+  const mapDisplaySets = customMapDisplaySets || _mapDisplaySets;
+
+  // The detail line of each thumbnail is customizable, and is added to whatever
+  // the mapping produced so that a panel supplying its own mapping - the
+  // measurement tracking one does - gets it too.
+  const mapDisplaySetsWithState = useCallback(
```
jbocce, line 82: **2. This callback isn't in its effects' dependency arrays.** `mapDisplaySetsWithState` is now a `useCallback` but is missing from both effects that call it (the dep arrays around lines 270 and 373). It works today only because the one input that changes its identity (`customMapDisplaySets`) is already listed, and the customization is read fresh at call time. Cleaner to list the callback itself so correctness doesn't rest on that coincidence.

platform/ui-next/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -69,6 +70,45 @@ const Thumbnail = ({
...
   const shouldRenderThumbnailImage = Boolean(imageSrc && !imageLoadFailed);
 
+  /**
+   * The detail line under the description. `details` comes from the
+   * `studyBrowser.thumbnailDetails` customization, resolved by the panel; the
+   * series number and instance count below are the same two items the default
+   * customization declares, kept here so the component still stands alone.
+   */
+  const renderDetails = (textClass: string, firstItemClass?: string) => {
+    const items = details ?? [
```
jbocce, line 84: **3. An empty details array blanks the line.** `details ?? [default]` falls back only on null/undefined, so `details: []` renders nothing instead of the series number and count. The shipped defaults never produce an empty array, but a customization that filters everything out would silently blank the line. Suggest `details?.length ? details : [default]`.

platform/core/src/utils/sortStudy.ts
```diff
@@ -106,20 +126,35 @@ export const seriesSortCriteria = {
...
+    return aInstance - bInstance;
   }
-  return compare(a.SOPInstanceUID, b.SOPInstanceUID) || compare(a.frameNumber, b.frameNumber);
+  // The instance numbers do not order these two - they are the same, or neither
+  // instance has one - so fall back to when each of them was created.  The last
+  // instance of a series is taken to be the most recently created one, so an
+  // instance number that fails to say which that is has to be replaced by
+  // something that does.
+  return (
+    compare(getSeriesDateTimeSortKey(a), getSeriesDateTimeSortKey(b)) ||
```
jbocce, line 180: **4. Instance sort now falls back to creation date/time in a hot path.** `sortByInstanceNumber` used to break ties by SOP UID; it now uses creation date/time first. Normal image series have unique instance numbers so this never fires for them, but this is a widely-used function — could we get a one-line confirmation in the PR that image ordering is unchanged? Low risk, just careful territory.

platform/core/src/utils/updateNewInstanceMetadata.ts
```diff
@@ -0,0 +1,59 @@
...
+  const priorInstanceNumber = instances.reduce(
+    (highest, instance) => Math.max(highest, Number(instance?.InstanceNumber) || 0),
+    0
+  );
+  dataset.InstanceNumber = priorInstanceNumber + 1;
+
+  const { date, time } = getCurrentDicomDateTime();
+  dataset.InstanceCreationDate = date;
+  dataset.InstanceCreationTime = time;
+  dataset.ContentDate = date;
```
jbocce: **5. Save stamps overwrite ContentDate/Time with "now" (local time).** Every save sets InstanceCreation and Content date/time to the current local time (called from the SEG and SR command modules). For a freshly authored report that's correct — just confirming no adapter sets a `ContentDate` we'd want preserved, and noting the value is local rather than UTC. Fine for same-machine sorting; worth stating.

platform/core/src/utils/sortStudy.ts
```diff
@@ -53,7 +54,7 @@ export const compareSameSeriesDisplaySet = (a, b) => {
       compareA === compareB
         ? compareA.compare(a, b)
         : compare(compareA.priority, compareB.priority);
-    if (!compareValue) {
+    if (compareValue) {
```
jbocce, line 57: **6. `addSameSeriesCompare` becomes active — needs a migration note.** Flipping this guard from `if (!compareValue)` to `if (compareValue)` means a registered same-series comparator now actually runs, where before it was silently ignored. Anyone relying (even unknowingly) on the old inert behavior could see series reorder. A short line in the changelog / the new "Display Set Date and Time" section of `notes-requirements.md` stating this is now active would cover it or maybe a migration doc intead? You decide.

tests/ThumbnailDetails.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { expect, test, visitStudyOptions } from './utils';
+
+// The only derived series in this study is a SEG, which is what the derived
+// series detail line matters for.
+const studyInstanceUID = '1.3.6.1.4.1.14519.5.2.1.256467663913010332776401703474716742458';
+
+const derivedThumbnail = page =>
+  page.locator('[data-cy="study-browser-thumbnail-no-image"]').first();
```
jbocce: **7. This test uses raw selectors instead of a page object.** It reaches into the DOM directly (`study-browser-thumbnail-no-image`, `thumbnail-detail-*`), but thumbnail access is meant to live in `LeftPanelPageObject`, which already owns a `thumbnails` getter and the load helpers. This is a good moment to add a `ThumbnailPageObject` (returned from `LeftPanelPageObject`) exposing a thumbnail's data — detail-item ids, series-number/count text, created date/time — so the spec asserts through that API instead of hand-rolled locators and `evaluateAll`. Keeps the `data-cy` selectors in one place and gives future thumbnail tests the same handle.

platform/ui-next/src/components/Thumbnail/Thumbnail.tsx
wayfarer3130: I actually don't see that as a problem - if someone configures it that way, maybe there is a good reason to.

platform/core/src/utils/updateNewInstanceMetadata.ts
wayfarer3130: If we want to preserve content date at some point, we would definitely need to update this logic, so I think it can goes there.

extensions/cornerstone-dicom-seg/src/getSopClassHandlerModule.ts
wayfarer3130: Added two tests to `seriesDateTime.test.js` covering the SEG shape directly — `prefers a later structure set date over the series date of a SEG` (later `StructureSetDate` wins with its own `StructureSetTime`) and `leaves the time empty when the later SEG date carries none` (the winning date takes no time rather than borrowing the series time from the day the images were acquired).

I put them at the util level rather than behind a new jest harness in `cornerstone-dicom-seg`: the extension has no jest setup today, and standing one up means mocking `@cornerstonejs/core`, `/tools` and `/adapters` for a single assertion about a value the handler reads straight out of `getSeriesDateTime`. Happy to add the harness if you would rather see it asserted through `_getDisplaySetsFromSeries`.

extensions/default/src/Panels/StudyBrowser/PanelStudyBrowser.tsx
wayfarer3130: Agreed — `mapDisplaySetsWithState` is now listed in both dep arrays (the one around line 265 and the subscription one around line 368). You are right that it only worked by coincidence; the callback is the honest dependency.

platform/ui-next/src/components/Thumbnail/Thumbnail.tsx
wayfarer3130: I made this change and then backed it out — the `??` is deliberate, and there is a test pinning it: `Thumbnail.test.ts` → `leaves the detail line empty when there are no details`.

The two states mean different things:

- `details` unset → no customization was resolved for this thumbnail, so the defaults stand in. `PanelStudyBrowser` relies on this: when it cannot find the display set it leaves `details` unset precisely so the series number and count still show.
- `details` empty → a customization ran and resolved to no items, and that empty line is honoured.

Switching to `details?.length ? …` collapses the two and removes any way for a customization to say "no detail line for this one" — the `thumbnailDetailTests` filtering you are describing is the mechanism for exactly that. So a customization that filters everything out blanking the line is the intended outcome, not a silent failure.

I have left the code as it was and added a comment spelling out the `??` vs `||` distinction so the next reader does not have to infer it. If you would still rather the defaults always win, say so and I will flip it and update the test.

platform/core/src/utils/sortStudy.ts
wayfarer3130: Confirmed: image ordering is unchanged.

The creation date/time only runs when `aInstance !== bInstance` is false — i.e. the instance numbers tie, or neither instance has one. Every instance of an image series carries a unique `InstanceNumber`, so the branch never executes for one and the order is the instance number order it has always been.

Added a test making that explicit rather than leaving it as a claim in the PR: `orders an image series by instance number alone` builds a series whose acquisition times run backwards relative to its instance numbers and asserts the sort still comes out `[1, 2, 3]`. The existing `the instance number decides before the creation date time` covers the pairwise case.

platform/core/src/utils/updateNewInstanceMetadata.ts
wayfarer3130: Good catch on the local time — it was a real problem, and I have changed it to UTC.

On the first half: no adapter sets a `ContentDate` worth preserving. dcmjs `DerivedDataset` (`src/derivations/DerivedDataset.js:22-23`) sets `ContentDate`/`ContentTime` to *now* by default, and neither the SEG nor the SR command module passes an explicit `options.ContentDate`. So the value being overwritten is itself a "now" stamp — nothing authored is lost.

On the second half, the local/UTC split was worse than a note. dcmjs derives its stamps from `toISOString()`, so `DerivedDataset` was writing `SeriesDate`/`SeriesTime` in **UTC** while `updateNewInstanceMetadata` wrote `ContentDate`/`InstanceCreationDate` in **local** time. Two clocks on one object, and since the display set date/time is the latest date any attribute carries, a local stamp in a zone behind UTC can be passed over in favour of the UTC series date it was meant to supersede — e.g. saving at 20:00 EDT stamps `ContentDate` 20260903 while dcmjs writes `SeriesDate` 20260904, and the SEG sorts by the series date instead of its own.

`getCurrentDicomDateTime` now reads the UTC components, so both levels are on one clock. Added `reads the date and time in UTC, not the local zone` to pin it (23:30 UTC is the previous day in every zone behind UTC, so a local reading cannot produce that result), and the existing formatting test now builds its date with `Date.UTC` so it does not depend on the runner timezone. Noted in the docs section too.

platform/core/src/utils/sortStudy.ts
wayfarer3130: Added a `Behaviour change: addSameSeriesCompare comparators now run` subsection to the new "Display Set Date and Time" section of `notes-requirements.md`, rather than a separate migration doc — there is no `3p13-to-3p14` guide yet, and the behaviour belongs next to the sorting it affects.

It states what the old guard actually did (returned the comparator answer only when that answer was `0`, and discarded it whenever it ordered the two sides, falling through to the instance compare), what it does now, and that a registered comparator taking effect may reorder display sets within a series that were previously ordered by instance number alone — with `addSameSeriesCompare(name, null, priority)` as the way back out.

tests/ThumbnailDetails.spec.ts
wayfarer3130: Done — added `tests/pages/ThumbnailPageObject.ts`, reached through `LeftPanelPageObject` the way `DicomTagBrowserPageObject` is reached through `DOMOverlayPageObject`.

`LeftPanelPageObject` gains `derivedThumbnails`, `thumbnail(locator)`, `thumbnailAt(nth)` and `derivedThumbnailAt(nth)`. The thumbnail object exposes `root`, `details`, `detail(id)`, `detailAt(nth)`, `detailIds()`, `seriesNumber`, `instanceCount`, `instanceDateTime`, `modality`, `description` and `load()`, all scoped to the thumbnail root rather than to `page`.

The spec now reads through that API — no `study-browser-thumbnail-no-image` or `thumbnail-detail-*` strings left in it, and both assertions wait web-first on `toHaveCount` before reading the id list, so nothing races a partly rendered line.

Both tests pass locally against the dev server (`npx playwright test ThumbnailDetails` — 2 passed).

extensions/cornerstone-dicom-seg/src/getSopClassHandlerModule.ts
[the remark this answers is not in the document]
wayfarer3130: This one is superseded by the newer thread on the same file — the SEG date selection question is answered there, with tests added for a SEG whose `SeriesDate` and `StructureSetDate` differ. `getSeriesDateTime` only takes a time from an attribute carrying the winning date, so a date/time pair from two different days cannot be produced.

platform/core/src/utils/updateNewInstanceMetadata.ts
wayfarer3130: Correcting my previous reply: I had this backwards, and the UTC switch is reverted.

DICOM DA and TM are wall clock values that carry no zone, and are displayed as they are stored — so a stamp read in UTC is simply the wrong date/time to show, and around midnight the wrong day. Matching dcmjs by storing UTC bought internal consistency at the cost of a value no viewer displays correctly.

`getCurrentDicomDateTime` now takes the dataset's `TimezoneOffsetFromUTC` (0008,0201) and reads the instant in that zone, falling back to local time when the object declares no offset or a malformed one. `updateNewInstanceMetadata` passes the offset of the dataset it is stamping. Tests cover the local reading, a declared offset in both directions, the day boundary in a declared zone, and the malformed-offset fallback.

The mismatch you would then expect with dcmjs is real but is dcmjs's bug, not ours: `DicomMetaDictionary.date()`/`.time()` derive from `toISOString()`, so `DerivedDataset` writes `SeriesDate`/`SeriesTime` in UTC. Filing that separately.

platform/core/src/utils/sortStudy.ts
wayfarer3130: Update: master has since landed a `3p13-to-3p14` migration guide (#6221), so this has moved there — `platform/docs/docs/migration-guide/3p13-to-3p14/display-set-ordering.md`, listed in that guide's index beside the save dialog changes. Your "maybe a migration doc instead?" was the right call; there just was not one to put it in at the time.

The page covers all four behaviour changes in this PR, not only the comparator:

- display sets ordering by their instance's creation date/time rather than the series date/time, and what still orders by series date/time;
- `addSameSeriesCompare` comparators now running, with `addSameSeriesCompare(name, null, priority)` as the way back out;
- `sortByInstanceNumber` tie-breaking by creation date/time before the SOP instance UID, noting image series are unaffected because their instance numbers are unique;
- the stamps `updateNewInstanceMetadata` writes on save, including that the instance number now comes from the highest in the series rather than from one predecessor.

`notes-requirements.md` keeps the description of how the date/time is chosen and links to the guide for what moves.

wayfarer3130:
Merged `origin/master` (through beta.21, including #6221).

One conflict, in `extensions/cornerstone-dicom-seg/src/commandsModule.ts`: #6221 added the predecessor series `Object.assign` at the same point this branch added the `updateNewInstanceMetadata` stamp. Both are kept, and **the order between them matters** — the assign runs first:

```js
if (series) {
  Object.assign(naturalizedReport, metaData.get(csEnums.MetadataModules.PREDECESSOR_SEQUENCE, series));
}
utils.updateNewInstanceMetadata(naturalizedReport);
```

`updateNewInstanceMetadata` looks the series up in `DicomMetadataStore` by `dataset.SeriesInstanceUID` to find the instances already in it. Until the assign has run, that UID is still the one dcmjs invented for a brand new series, so the lookup would find no prior instances and number the new instance `1` — colliding with what is already in the series being extended. Stamping second also means the creation date/time win over anything the predecessor data carries, which is what we want for a newly authored object.

Worth a look from whoever knows #6221 best, since it is the kind of dependency that survives a merge silently.

Also moved the display set ordering behaviour changes out of `notes-requirements.md` and into the `3p13-to-3p14` guide that #6221 introduced — see the reply on the `addSameSeriesCompare` thread.

PR #6223, opened 2026-08-21
fix(hotkeys): hand keystrokes back to the browser inside modal dialogs

TFRadicalImaging:
### Context

OHIF hotkeys stay live while a modal dialog is open. `HotkeysManager._bindHotkeys` (`platform/core/src/classes/HotkeysManager.ts`) binds every hotkey through Mousetrap and **always** `preventDefault`s before running the command:

```js
mouseTrapAPI.bind(combinedKeys, evt => {
  evt.preventDefault();
  ...
```

Mousetrap's default `stopCallback` opts a keystroke out only when the target is an `input`, `select`, `textarea` or `contenteditable` element. A focused modal dialog is none of those, so two things happen while a modal is open:

- Single-key viewport hotkeys (rotate, scroll, layout, …) still fire behind the modal, even though the modal has focus.
- A mode that binds a combo the browser already owns — `ctrl+c`, `ctrl+a`, `ctrl+v` — takes select-all, copy and paste away from every dialog too. Text shown inside a modal selects but never copies, because the keystroke was claimed and its default action cancelled.

### Changes & Results

A Mousetrap `stopCallback` plugin (`allowNativeKeysPlugin`), registered next to the existing `pausePlugin` and `recordPlugin`. `stopCallback` is Mousetrap's supported way to opt a keystroke out of the hotkey system and let its default action through — the same hook it already uses to ignore shortcuts typed into inputs and textareas.

The test is on **where the keystroke landed**, not on what is selected and not on which combo was pressed. No OHIF hotkey fires when the event target is inside:

- **`[role="dialog"]`** — every modal, with no markup changes. A `DialogContent` renders that role, carries `tabindex="-1"` and focuses itself on open, so a keydown's target is inside it from the moment the dialog appears. Blocking hotkeys under an open modal is correct on its own terms.
- **`.ohif-text-select`** — an opt-in marker for selectable text that is *not* in a dialog (a value in a side panel, say). Nothing uses it yet; it is there so the next such case is a class plus a `tabIndex` rather than another plugin. The plugin's doc comment spells out the two requirements (the class must sit on the focused element or an ancestor, and that element must be focusable).

It sits in `platform/core` beside the other Mousetrap plugins, and after `pausePlugin`, so the pause check still wins while paused.

Before: `ctrl+c` over selectable text in a modal fires the bound annotation command and copies nothing; `r`, arrows, etc. drive the viewport behind an open modal.
After: inside a dialog (or a marked region) the browser keeps `ctrl+a`/`ctrl+c`/`ctrl+x`/`ctrl+v` and OHIF hotkeys stay silent; everywhere else every hotkey is unchanged.

### Testing

Unit: `platform/core/src/utils/hotkeys/allowNativeKeysPlugin.test.js` — 7 tests over real DOM nodes (target inside a dialog, target *being* the dialog, target inside `.ohif-text-select`, target in neither, every argument forwarded incl. the `sequence`, the wrapped `stopCallback` still deciding the rest, a target with no `closest()`).

Manual, in a build that binds a clipboard combo to a command:
- Open a modal with selectable text, drag-select it, press `ctrl+c` → the text is copied and no bound command fires; check `defaultPrevented` is `false` on the `ctrl+c` keydown.
- With a modal open, press a single-key viewport hotkey (e.g. rotate) → nothing happens on the viewport behind it.
- Over a viewport, every hotkey behaves exactly as before.

### Notes

Contributed by University of Calgary. Thanks to @wayfarer3130 for directing the fix toward `platform/core` and asking for it upstream.

#### PR

#### Code

#### Tested Environment

- [x] OS: macOS 15
- [x] Node version: 24.3.0
- [x] Browser: Chrome

## Summary by CodeRabbit

* **Bug Fixes**
  * Native browser keyboard behavior is now preserved when typing or selecting text in dialogs and designated text-selection areas.
  * Application hotkeys continue to work normally outside these regions.
* **Tests**
  * Added coverage for dialog handling, text-selection areas, pass-through behavior, and compatibility with different event targets.

### DIFF
--- a/platform/core/src/utils/hotkeys/allowNativeKeysPlugin.js
+++ b/platform/core/src/utils/hotkeys/allowNativeKeysPlugin.js
@@ -1,0 +1,72 @@
+/**
+ * Hands a keystroke back to the browser — instead of firing a bound OHIF hotkey
+ * — whenever it lands inside a modal dialog, or inside an element explicitly
+ * marked as a native-keyboard region.
+ *
+ * `HotkeysManager` always `preventDefault`s a matched hotkey before running its
+ * command (`_bindHotkeys` in `platform/core/src/classes/HotkeysManager.ts`), so a
+ * bound keystroke is claimed **app-wide** — including while a modal dialog is
+ * open. Two things follow from that:
+ *
+ *   - Single-key hotkeys (rotate, scroll, layout, …) still fire on the viewport
+ *     behind an open modal, even though the modal has focus.
+ *   - A mode that binds a combo the browser already owns — `ctrl+c`, `ctrl+a`,
+ *     `ctrl+v` — takes select-all, copy and paste away from every dialog too, so
+ *     text shown in a modal selects but never copies.
+ *
+ * Mousetrap's default `stopCallback` already opts keystrokes out of the hotkey
+ * system when the target is an `input`, `select`, `textarea` or `contenteditable`
+ * element, letting their native action through. This plugin extends that same
+ * rule from "the target is editable" to "the target is somewhere the browser's
+ * own keyboard behaviour should win".
+ *
+ * The test is on *where the keystroke landed*, not on what happens to be
+ * selected, and not on which combo was pressed. That keeps it predictable: every
+ * hotkey is live everywhere except in the regions below, rather than flickering
+ * off whenever a stale selection exists somewhere in the document.
+ */
+
+/**
+ * `[role="dialog"]` covers every modal in the app with no markup changes at all.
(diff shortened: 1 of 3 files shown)


TFRadicalImaging:
@wayfarer3130 upstream port of the hotkeys-inside-dialogs fix, as requested. Just the generic `platform/core` Mousetrap plugin — the Manta-specific behaviour doc stays downstream. `@ohif/core` suite green (41 suites / 378 tests) on Node 24.

wayfarer3130 (approved):

PR #6228, opened 2026-08-26
fix(dicom-pdf): guarantee the type of encapsulated documents

wayfarer3130:
## Problem

`MIMETypeOfEncapsulatedDocument` is supplied by whoever produced the instance, and a `blob:` URL inherits the origin of the document that created it. A spoofed `text/html` payload therefore executed **same-origin with the viewer**. Measured in Chrome against a real study:

```
SCRIPT RAN. origin=http://localhost:3001 | parent.localStorage keys=3
```

The `type="application/pdf"` attribute on `<object>` never filtered anything — it is only a hint, and the browser keys off the resource's actual Content-Type (the blob's type, or the server's header).

The same hole existed on the raw-URL path, where a WADO-RS URL reached the element directly and the PACS chose the `Content-Type`.

## Fix

The payload is read the way every other bulkdata value is read — inline content when the metadata carries it, otherwise the data source's own `retrieveBulkData`, preferring the buffer it caches on `value.Value`. This mirrors `resolveBulkDataTags` and matches the path the video display sets take, so authentication, bulkdata URI resolution and caching all behave the same here.

Those bytes are then wrapped in a `Blob` whose type comes from an allowlist:

```ts
const blob = new Blob([payload], { type: documentType.mimeType });
```

After that line neither the instance nor the origin server has any say in how the browser interprets the bytes. PDF payloads additionally have to satisfy a `%PDF-` signature check.

The allowlist maps each canonical type to how it is embedded:

| Type | Element |
|---|---|
| `application/pdf` | `<object type="application/pdf">` |
| `text/html`, `application/xhtml+xml`, `text/xml`, `application/xml`, `text/plain` | `<iframe sandbox="">` |
| anything else | not rendered |

Non-standard spellings (`application/html`, `application/x-pdf`, `application/acrobat`, `text/pdf`, `application/xhtml`) fold onto their canonical entry. Accepting an alias is safe because the canonical entry — not the declared spelling — decides both the blob type and the signature.

The viewport's status and fallback strings live in a new `EncapsulatedDocument` i18n namespace, registered in the `en-US` and `test-LNG` locale indexes.

## Why PDF is not sandboxed

This split is forced by the browsers, not chosen. Chrome renders a PDF under no sandbox at all, and under **none** of the token combinations:

| Sandbox | PDF renders |
|---|---|
| *(no sandbox attribute)* | yes |
| `sandbox=""` | no |
| `sandbox="allow-scripts"` | no |
| `sandbox="allow-same-origin"` | no — *"This page has been blocked by Chromium"* |
| `sandbox="allow-scripts allow-same-origin"` | no |
| `+ allow-downloads` | no |

Even the fully permissive combination fails, and `allow-plugins` no longer exists, so there is no way to sandbox a PDF and still display it. PDFs rely on the type guarantee instead. Markup types render correctly inside a sandbox, so they get one with no tokens at all — scripts inert, opaque origin, no access to the viewer.

Only Chrome was measured. If Firefox or Safari behave differently the allowlist is the single place to adjust.

## Verification

Both strategies verified end to end in the viewer against a real study:

- `application/pdf` → renders in the native viewer, `<object type="application/pdf">` with a blob source.
- `text/html` → renders as `<iframe sandbox="" allow="" referrerpolicy="no-referrer">`, with an in-document probe reporting `PASS - scripts did not run` and `PASS - no access to the embedding page`.

27 new unit tests cover normalization, alias folding, allowlist rejection, signature matching, and the load path — each resolution branch (inline, already-decoded inline, cached `value.Value`, `retrieveBulkData`), the canonical media type on the retrieve call, refusing HTML that claims to be a PDF, and abort/revoke semantics.

The i18n wiring was checked with `&lng=test-LNG`, since under `en-US` the key and the value are identical and a missing bundle would be invisible. Forcing the failure branch renders `Test Unable to retrieve this document`, i.e. the value resolves from the namespace bundle rather than falling back to the key.

This extension had neither a jest nor a babel config, so the root `extensions/*/jest.config.js` project glob was never collecting its tests. Both are added.

## Notes for review

- Documents no longer stream or range-request; the payload is fully resolved so it can be checked and re-typed. Fine for reports, worth knowing for anything large.
- The display set now carries a `getDocument` loader instead of `renderedUrl`/`getRenderedUrl`. Those were consumed only by this viewport, and `directURL` eagerly created an object URL per DOC instance at metadata time that nothing revoked.
- A document whose value offers neither inline content nor `retrieveBulkData` renders nothing. That is deliberate — fail closed.
- `react-i18next` is declared as a peer dependency, matching the other extensions that use `useTranslation`. That required a lockfile update; alongside the three lines for this package, pnpm also corrected pre-existing drift of its own (`@babel/runtime` 7.29.2 → 7.29.7 in four unrelated snapshot entries, and a `deprecated:` note on `crypto-js`). Happy to split that out if you would rather it landed separately.

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying validated PDF, HTML, and plain-text documents from DICOM data.
  * Added MIME-type normalization and support for common MIME aliases.
  * Added document-specific rendering with secure sandboxed iframe support where appropriate.
  * Added support for inline, cached, and retrieved document data.

* **Bug Fixes**
  * Improved document validation using file signatures before rendering.
  * Added cleanup of temporary document URLs to prevent resource leaks.
  * Improved messaging for unsupported, unavailable, or invalid documents.

### DIFF
--- a/extensions/dicom-pdf/src/getSopClassHandlerModule.js
+++ b/extensions/dicom-pdf/src/getSopClassHandlerModule.js
@@ -1,6 +1,10 @@
 import { SOPClassHandlerId } from './id';
 import { utils, Types as OhifTypes } from '@ohif/core';
 import i18n from '@ohif/i18n';
+import {
+  getDisplayableDocumentType,
+  normalizeDocumentMimeType,
+} from './utils/displayableDocumentTypes';
 
 const SOP_CLASS_UIDS = {
   ENCAPSULATED_PDF: '1.2.840.10008.5.1.4.1.1.104.1',
@@ -14,10 +18,17 @@ const _getDisplaySetsFromSeries = (instances, servicesManager, extensionManager)
     const { Modality, SOPInstanceUID } = instance;
     const { SeriesDescription = 'PDF', MIMETypeOfEncapsulatedDocument } = instance;
     const { SeriesNumber, SeriesDate, SeriesInstanceUID, StudyInstanceUID, SOPClassUID } = instance;
+    // The declared type is carried through to the viewport, which resolves it
+    // against the displayable-type allowlist and re-wraps the payload in a Blob
+    // of the canonical type. Here it only decides what to ask the origin server
+    // for, so the canonical spelling is preferred when the type is one we know.
+    const mimeType = normalizeDocumentMimeType(MIMETypeOfEncapsulatedDocument) || 'application/pdf';
+    const documentType = getDisplayableDocumentType(mimeType);
+
     const renderedUrlParams = {
       instance,
       tag: 'EncapsulatedDocument',
-      defaultType: MIMETypeOfEncapsulatedDocument || 'application/pdf',
+      defaultType: documentType?.mimeType || mimeType,
       singlepart: 'pdf',
     };
     const renderedUrl = dataSource.retrieve.directURL(renderedUrlParams);
@@ -42,6 +53,7 @@ const _getDisplaySetsFromSeries = (instances, servicesManager, extensionManager)
       measurements: null,
       renderedUrl: renderedUrl,
       getRenderedUrl,
+      mimeType,
       instances: [instance],
       thumbnailSrc: null,
       isDerivedDisplaySet: true,
--- a/extensions/dicom-pdf/src/getSopClassHandlerModule.js
+++ b/extensions/dicom-pdf/src/getSopClassHandlerModule.js
@@ -1,10 +1,8 @@
 import { SOPClassHandlerId } from './id';
 import { utils, Types as OhifTypes } from '@ohif/core';
 import i18n from '@ohif/i18n';
-import {
-  getDisplayableDocumentType,
-  normalizeDocumentMimeType,
-} from './utils/displayableDocumentTypes';
+import { normalizeDocumentMimeType } from './utils/displayableDocumentTypes';
+import { loadDisplayableDocument } from './utils/loadDisplayableDocument';
 
 const SOP_CLASS_UIDS = {
   ENCAPSULATED_PDF: '1.2.840.10008.5.1.4.1.1.104.1',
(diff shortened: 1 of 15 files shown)


extensions/dicom-pdf/src/getSopClassHandlerModule.js
```diff
@@ -14,10 +18,17 @@ const _getDisplaySetsFromSeries = (instances, servicesManager, extensionManager)
     const { Modality, SOPInstanceUID } = instance;
     const { SeriesDescription = 'PDF', MIMETypeOfEncapsulatedDocument } = instance;
     const { SeriesNumber, SeriesDate, SeriesInstanceUID, StudyInstanceUID, SOPClassUID } = instance;
+    // The declared type is carried through to the viewport, which resolves it
+    // against the displayable-type allowlist and re-wraps the payload in a Blob
+    // of the canonical type. Here it only decides what to ask the origin server
+    // for, so the canonical spelling is preferred when the type is one we know.
+    const mimeType = normalizeDocumentMimeType(MIMETypeOfEncapsulatedDocument) || 'application/pdf';
+    const documentType = getDisplayableDocumentType(mimeType);
```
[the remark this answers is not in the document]
wayfarer3130: The sop class requires pdf, so I think this is a reasonable default.

wayfarer3130:
@jbocce - should we include this PR in the next 3.13 release?  Can you review as is now?

pnpm-workspace.yaml
```diff
@@ -79,7 +79,9 @@ overrides:
   protobufjs: 7.6.1
   tmp: 0.2.7
   shell-quote: 1.10.0
-  'fast-uri@>=3.0.0 <3.1.5': 3.1.5
+  # 3.1.6 patches GHSA-5jgf-p345-68v8, GHSA-f65p-4m7j-42xc, GHSA-fph4-wmhf-6fwf
```
jbocce, line 82: If you update with master you won't need these any longer.

extensions/dicom-pdf/src/utils/displayableDocumentTypes.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  if (!signature?.length) {
+    return true;
+  }
+
+  if (payload.byteLength < signature.length) {
+    return false;
+  }
+
+  const head = new Uint8Array(payload, 0, signature.length);
```
jbocce: There is one case that is not handled. Not sure how much of an edge case it is, but a real PDF carrying a leading UTF‑8 BOM, or leading whitespace/junk (which tolerant PDF readers accept), starts with EF BB BF/0x20 and would fail. I tried it myself and can provide you with both the dcm file that contains the pdf and the raw pdf itself (the pdf itself I can include here). Here is a recommended implementation. I suspect you might want to also flag which document types could possible have the UTF-8 BOM.

```
const UTF8_BOM = [0xef, 0xbb, 0xbf];

function bytesMatchAt(bytes: Uint8Array, expected: number[], offset: number): boolean {
  if (bytes.length < offset + expected.length) {
    return false;
  }
  return expected.every((byte, index) => bytes[offset + index] === byte);
}

export function matchesDocumentSignature(
  documentType: DisplayableDocumentType,
  payload: ArrayBuffer
): boolean {
  const { signature } = documentType;

  if (!signature?.length) {
    return true;
  }

  const bytes = new Uint8Array(payload);

  // A UTF-8 BOM (EF BB BF) before the header is common in real producers, and
  // every mainstream reader tolerates it (Acrobat scans the first 1024 bytes
  // for %PDF-). Allow the signature to sit at offset 0 or immediately after a
  // single BOM. The Blob is still typed from the allowlist, so this does not
  // change how the browser parses the bytes - only which valid files we accept.
  const offset = bytesMatchAt(bytes, UTF8_BOM, 0) ? UTF8_BOM.length : 0;

  return bytesMatchAt(bytes, signature, offset);
}
```

[real-bom.pdf](https://github.com/user-attachments/files/31805073/real-bom.pdf)

jbocce:
Overall this fix looks great. I tried four cases...

1. An Encapsulated PDF instance whose inner document is declared as HTML and actually contains an HTML page. It confirmed the viewer now shows this kind of content inside a locked-down frame, so the page displays but cannot run its own code or reach anything in the viewer.
2. An Encapsulated PDF instance that declares its inner document as a PDF but actually contains HTML. It confirmed the viewer's content check catches the mismatch and refuses to show it, displaying a "content does not match its declared type" message instead.
3. An Encapsulated PDF instance carrying a completely valid PDF that happens to begin with three extra invisible marker bytes (a UTF-8 BOM) before the usual %PDF- start. It showed that the viewer's strict start-of-file check wrongly rejects this legitimate file, even though standard PDF readers open it fine — a robustness gap worth fixing.
4. An Encapsulated PDF instance carrying a valid PDF that starts with %PDF- and passes the viewer's check, but has embedded JavaScript inside it. It showed that passing the check doesn't mean the PDF is free of active content, since the check only looks at the first few bytes and never inspects what's inside. We also confirmed that this embedded JavaScript runs in the browser's isolated PDF engine and cannot reach the embedding viewer page's DOM, URL, or storage.

3 is the only one to address.

wayfarer3130:
Thanks @jbocce - fixed 3 in 085a056bf3.

The check required `%PDF-` at offset 0. It now scans the first 1024 bytes for it, matching Adobe's implementation note that mainstream readers follow, so a BOM or any other leading bytes the renderer skips anyway no longer fail the check. Configured per type via `signatureSearchLimit`; types without one still require offset 0.

Your case 2 still fails as before - HTML declared as PDF has no `%PDF-` anywhere in it. Added tests for BOM, header at the last searched offset, and header past the window.

Also merged master.

jbocce (approved):

PR #6236, opened 2026-08-31
test(contour): add smooth edges and merge operation E2E tests

diattamo:
### Context

Adds E2E coverage for two contour utilities on the hydrated RTSTRUCT study, extending the contour test suite (#6117, #6206).

### Changes & Results

**Merge operation (`ContourCombineOperations.spec.ts`)**
- New test: merge Big Sphere with nested Small Sphere (no edge crossing) produces a single new segment shaped like Big Sphere.
- Since the union looks identical to Big Sphere, the test asserts exactly one contour path for the result on top of the panel row assertions — guarding against a silent no-op (empty segment) or a holed result — and pairs that with a viewport screenshot.

**Smooth Edges operation (`ContourSmoothOperation.spec.ts`, new)**
- Covers Smooth Edges (the `smoothContours` command) from the "Smooth Contours" utility popover, acting on the active Threshold segment.
- Captures the segment's SVG path as a baseline, smooths, then uses web-first retrying assertions: the path count stays 1, the `d` attribute no longer matches the baseline, and the contour stays closed (`/Z\s*$/`), followed by a viewport screenshot of the smoothed result.
- Page object: adds a `smoothContours` popover object (`open`/`close`/`smoothEdges`) to `RightPanelPageObject`; `close` dismisses the popover so it does not overlap the screenshot.

### Testing

```sh
pnpm run test:e2e tests/ContourCombineOperations.spec.ts
pnpm run test:e2e tests/ContourSmoothOperation.spec.ts
```

#### PR

- [x] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [x] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [x] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [x] OS: macOS 26.5.1 (arm64, Apple M5 Pro)
- [x] Node version: 25.4.0 (pnpm 11.5.2)
- [x] Browser: Playwright 1.56.1 bundled Chromium 141.0.7390.37

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for merging contours and smoothing contour edges.
  * Added checks for segment panel state and resulting contour geometry.
  * Added viewport screenshot validation for merge and smoothing workflows.
  * Improved test interactions with contour operation popovers to ensure consistent behavior.

PR #6117, opened 2026-06-29
test(contour): add Freehand and Spline contour drawing tool E2E tests

diattamo:
### Context

This PR adds E2E coverage for the contour drawing tools (Freehand, Spline) — verifying
that a drawn contour **persists unchanged when you navigate away from its slice/segment
and come back**.

Tests are organized one file per tool so each tool's file can grow with future
tool-specific behavior beyond persistence.

### Changes & Results

Test-only change — no production code touched. Reuses existing infrastructure already on
master: the `visitStudyAndHydrate` util, the `getSvgAttribute` util, and the
`RightPanelPageObject` contour tool handlers (`freehandContour`, `splineContour`).

- **Add `tests/FreehandContourSegmentation.spec.ts`** (2 tests)
- **Add `tests/SplineContourSegmentation.spec.ts`** (2 tests)

Each spec hydrates an RTSTRUCT contour segmentation and verifies, for both an
un-pre-selected segment and a freshly added segment:

1. drawing the contour adds it to the viewport (SVG path count 1 → 2);
2. navigating away (selecting another segment, which jumps slices) leaves slice 1;
3. navigating back re-renders the contour (path count returns to 2 on slice 1).

### Testing

```bash
cross-env TEST_ENV=true pnpm exec playwright test \
  tests/FreehandContourSegmentation.spec.ts \
  tests/SplineContourSegmentation.spec.ts
```

All 4 tests pass locally (Chromium).

#### PR

- [x] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [x] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [x] The documentation page has been updated as necessary for any public API
  additions or removals. 

#### Tested Environment

- [x] OS: macOS 26.5.1
- [x] Node version: 25.4.0
- [x] Browser: Chromium (Playwright)

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage ensuring freehand contour drawings persist when navigating slices and switching segmentation segments, including when no segment is pre-selected and after adding a new segment.
  * Added end-to-end coverage ensuring spline contour drawings persist across slice navigation and segment switching, including validation that the persisted spline contour output matches the originally drawn result after returning to slice 1.

PR #6260, opened 2026-09-08
test(contour): add fill, outline and opacity contour display coverage

diattamo:
### Context

The contour segmentation appearance config had no E2E coverage: nothing asserted that the
display mode tabs (fill / outline / fill & outline), the fill opacity and the outline width
actually change what is rendered. This adds that coverage and extends the contour test suite
(#6117, #6206, #6236).

### Changes & Results

**New spec (`ContourSegmentationFillOutlineOpacity.spec.ts`, 6 tests)**

- A panel-created contour renders with both fill and outline by default.
- Switching display modes toggles fill and outline on the rendered contour.
- A display mode applies to contours drawn afterwards *and* re-renders existing ones, across
  two segments.
- Typing an opacity value changes the fill alpha, leaves the outline untouched, and a zero
  alpha does not silently turn the fill display off.
- Typing a border value changes the outline width and leaves the fill untouched.
- Changing the opacity while in outline mode keeps the fill hidden, and re-enabling fill
  applies the alpha chosen while it was hidden.

Every assertion reads the rendered SVG (`fill-opacity`, `stroke-width`, `fill`) rather than the
state of the panel control, so the tests verify the effect rather than the widget. No
screenshots were added.

**Page object (`RightPanelPageObject`)**

- The contour and labelmap appearance config are unified into `getSegmentationConfig(typeSuffix)`
  and `getNumericConfig(control, typeSuffix)`, replacing two near-identical copies. The display
  tabs now expose the `button` locator, which carries the `data-state` reflecting the selected
  mode.
- Config actions expand the section themselves, so a caller no longer has to remember to open it
  first, matching the way `MainToolbarPageObject` combines opening a menu with clicking one of
  its items. The expand is guarded on the section already being visible, because the toggle
  collapses a section that is already expanded.
- The three call sites that used `config.toggle.click()` to *open* the section now call
  `config.open()`. This removes a latent race: the raw toggle never waited for the section to
  settle, so a following action could run against a section that had not finished expanding.
  `toggle` remains for deliberately collapsing.

**New utility**

- `drawFreehandContour({ segmentationPanel, viewport, path, activateTool })` draws a closed
  freehand contour along a normalized viewport path, with `activateTool: false` for the case
  where the tool is still armed from a previous contour.

### Testing

```sh
pnpm run test:e2e tests/ContourSegmentationFillOutlineOpacity.spec.ts tests/SegmentationPanel.spec.ts
```

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 26.5.1 (arm64)
- [x] Node version: 25.4.0 (pnpm 11.5.2)
- [x] Browser: Playwright 1.56.1 bundled Chromium 141.0.7390.37

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for contour segmentation display modes, including fill, outline, opacity, and border width.
  * Added validation for independent active and inactive segmentation rendering, opacity settings, visibility toggling, and switching between segmentations.
  * Improved coverage for drawing contours across multiple segmentations and applying display settings before and after drawing.
  * Strengthened validation of segmentation panel controls and configuration interactions.

### DIFF
--- a/tests/ContourSegmentationFillOutlineOpacity.spec.ts
+++ b/tests/ContourSegmentationFillOutlineOpacity.spec.ts
@@ -63,8 +63,9 @@ test.beforeEach(async ({ page, rightPanelPageObject }) => {
   await expect(contourPanel.panel.rows, 'Expected the default segment row').toHaveCount(1);
   await waitForViewportsRendered(page);
 
-  // Expand the appearance config section holding the display tabs and sliders.
-  await contourPanel.config.toggle.click();
+  // Expand the appearance config section holding the display tabs and sliders, so the tests
+  // that only assert on its controls find them. The config actions expand it on their own.
+  await contourPanel.config.open();
 });
 
 test('should render a panel-created contour with both fill and outline by default', async ({
--- a/tests/ContourSegmentationFillOutlineOpacity.spec.ts
+++ b/tests/ContourSegmentationFillOutlineOpacity.spec.ts
@@ -1,13 +1,19 @@
-import { expect, test, visitStudy, waitForViewportsRendered } from './utils';
+import { drawFreehandContour, expect, test, visitStudy, waitForViewportsRendered } from './utils';
+import type { RightPanelPageObject } from './pages';
+import type { IViewportPageObject } from './pages/ViewportPageObject';
 
 const studyInstanceUID = '1.3.12.2.1107.5.2.32.35162.30000015050317233592200000046';
 const mode = 'segmentation';
 
-// Default contour style for segmentations created from the panel: the fill is
-// enabled on creation and inherits the type-level fill alpha, the outline uses
-// the default outline width (see segmentationPanelCustomization / contourConfig).
+// Default contour style for segmentations created from the panel: the fill is enabled on
+// creation (segmentationPanelCustomization) and inherits the Contour type-level fill alpha
+// declared in the cornerstone extension init, while the outline uses the default width.
 const defaultFillOpacity = '0.5';
 const defaultOutlineWidth = '1';
+
+// First entry of the cornerstone color LUT, which a panel-created segmentation gets for its
+// first segment (SegmentationService seeds the LUT with the background color only).
+const defaultSegmentColor = 'rgb(221, 84, 84)';
 
 const dragShape = [
   { x: 0.4, y: 0.4 },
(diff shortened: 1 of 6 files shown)


tests/utils/drawFreehandContour.ts
```diff
@@ -0,0 +1,25 @@
...
+  path,
+  activateTool = true,
+}: {
+  segmentationPanel: RightPanelPageObject['contourSegmentationPanel'];
+  viewport: IViewportPageObject;
+  path: { x: number; y: number }[];
+  activateTool?: boolean;
+}) {
+  if (activateTool) {
+    await segmentationPanel.tools.freehandContour.click();
```
jbocce, line 21: Do we really need the guard? Doing a click on an active tool is harmless and anyone calling this util will want the tool enabled. If you feel it is necessary then I would also argue that the guard goes in the page object.

tests/pages/RightPanelPageObject.ts
```diff
@@ -21,6 +21,74 @@ export class RightPanelPageObject {
...
+   * driven through the number input.
+   */
+  private getNumericConfig(control: 'opacity' | 'border' | 'opacity-inactive', typeSuffix: string) {
+    const input = this.page
+      .getByTestId(`segmentation-config-${control}-${typeSuffix}`)
+      .locator('input[type="number"]');
+
+    return {
+      input,
+      fill: async (value: string) => {
```
jbocce, line 48: Is `opacity` or perhaps `fillOpacity` a better name for this? Or maybe even something else? `fill` can be misleading.

tests/pages/RightPanelPageObject.ts
```diff
@@ -21,6 +21,74 @@ export class RightPanelPageObject {
...
+        click: async () => {
+          await open();
+          await button.click();
+        },
+      };
+    };
+
+    return {
+      open,
+      toggle: {
```
jbocce, line 75: I am not crazy about there being an open and a toggle. Do we do something like this elsewhere? The thing that bothers me the most about it is that the "constant" for looking up the button to  open/toggle appears both on line 60 and 33. Consider having just a toggle that takes an open/close boolean or have both an open and close method that share the constant.

tests/ContourSegmentationFillOutlineOpacity.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+  contourPanel: RightPanelPageObject['contourSegmentationPanel'];
+  viewport: IViewportPageObject;
+}) {
+  const paths = viewport.svg('path');
+
+  await expect(paths, 'Expected the starting number of paths to be 0').toHaveCount(0);
+  await drawFreehandContour({ segmentationPanel: contourPanel, viewport, path: dragShape });
+  await expect(paths, 'Expected the freehand contour to be added').toHaveCount(1);
+
+  const contour = paths.nth(0);
```
jbocce, line 51: contourPath might be a better name.

tests/ContourSegmentationFillOutlineOpacity.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+  await expect(contour, 'Expected the fill to render fully transparent').toHaveAttribute(
+    'fill-opacity',
+    '0'
+  );
+  // A zero alpha is the opacity value hiding the fill, not the fill display being turned off.
+  await expect(
+    display.fillAndOutline.button,
+    'Expected the panel to stay in fill & outline mode'
+  ).toHaveAttribute('data-state', 'active');
+  await expect(contour, 'Expected the outline to be unaffected by opacity').toHaveAttribute(
```
jbocce, line 250: If we are checking that the panel shows the correct outline, should we also check the SVG stroke width attribute too?

jbocce (changes requested):
Just a few comments and one question here: are you planning to test the inactive opacity in another PR? I think that is perfectly fine, but I wanted to have the answer recorded here. 😃

diattamo:
> Just a few comments and one question here: are you planning to test the inactive opacity in another PR? I think that is perfectly fine, but I wanted to have the answer recorded here. 😃

I'd missed that portion. I figured it was easier to just add it to this PR. So there are test cases for it now.

jbocce (approved):

PR #6273, opened 2026-09-14
fix(ui-next): modify numeric and double slider inputs behavior to allow dynamic set of min and max and easier value modification (better UX)

### LINKED ISSUE #6269: [Bug] The segmentation Threshold range slider limits the values to -1000...1000, and the number boxes refuse an empty field
## Describe the Bug

The Threshold tool of the segmentation toolbox has a "Range" mode. That mode shows a
double-range slider with a number box on each side. The slider has two problems.

**Problem 1. The value limits are wrong for CT.**

The option declares `min: -1000` and `max: 1000` in
`extensions/cornerstone/src/customizations/segmentationToolbarCustomization.ts`:

```js
{
  name: 'ThresholdRange',
  type: 'double-range',
  id: 'threshold-range',
  min: -1000,
  max: 1000,
  step: 1,
  value: [50, 600],
  ...
}
```

A CT image holds values above 1000 HU and below -1000 HU. Dense bone, contrast media, and
metal go above 1000 HU. The padding value outside the reconstruction circle is often -2000 HU
or -3000 HU. The user cannot set a threshold outside -1000...1000 HU, so the user cannot
select these regions.

The same limits also make the slider hard to use. The slider bar covers 2000 HU in a small
number of pixels. One pixel of the bar is many HU. The negative half of the bar has little
value for most CT work, and that half takes 50% of the bar.

**Problem 2. The number boxes reject an empty field, so the user cannot edit a value.**

`platform/ui-next/src/components/DoubleSlider/DoubleSlider.tsx` holds both number boxes as
controlled inputs. The change handler ignores a string that `parseFloat` cannot read:

```js
const handleInputChange = React.useCallback(
  (index: 0 | 1, inputValue: string) => {
    const newValue = parseFloat(inputValue);
    if (!isNaN(newValue)) {
      const clampedValue: [number, number] = [...value];
      clampedValue[index] = roundToStep(Math.min(Math.max(newValue, min), max));
      ...
    }
  },
  [value, min, max, onValueChange, step]
);
```

`parseFloat('')` and `parseFloat('-')` both give `NaN`. The handler makes no state change, and
React writes the old value back into the box. Therefore the box can never become empty, and
the user can never delete the minus sign.

The handler also clamps the value on every keystroke. A partly typed number becomes a clamped
number before the user finishes the number.

## Steps to Reproduce

1. Open the OHIF viewer, and load a CT study.
2. Open the segmentation mode, and create a labelmap segmentation.
3. Select the **Threshold** tool in the toolbox.
4. Set the **Threshold** radio control to **Range**. The double-range slider shows the default
   values `[50, 600]`.
5. Drag the right thumb of the slider to the right end. The value stops at 1000.
6. Click in the left number box, and type `-1500`.
7. Set the left number box to `-1`. Put the text cursor at the end of the text. Press
   `Backspace` two times.

## The current behavior

- Step 5: the upper limit stops at 1000 HU. The user cannot select bone above 1000 HU, or
  metal, or contrast media.
- Step 6: the box clamps the value to -1000. The user cannot include the air regions and the
  padding regions of the CT, which hold values near -1024 HU or lower.
- Step 7: the first `Backspace` gives the text `-`, and the box shows `-1` again. The second
  `Backspace` gives the text `` (empty), and the box shows `-1` again. The user cannot delete
  the `1`, and the user cannot delete the `-`. To get the value 30, the user must select all of
  the text first. There is no way to edit the text one character at a time.
- The clamp on each keystroke also changes the text while the user types. Type `1500` in the
  right box: the box shows `1`, then `15`, then `150`, then `1000`.

## The expected behavior

- The slider bar covers 0...1000 by default. That range covers the usual soft tissue values and
  bone values of a CT, and it gives good precision on the bar.
- Each number box accepts a value outside the range of the bar, and the box moves the limit of
  the bar. Type `-1500` in the left box: the lower limit of the bar becomes -1500, and the
  threshold includes all of the black regions of the CT. Type `3000` in the right box: the
  upper limit of the bar becomes 3000.
- Each number box holds the text that the user types. The box accepts an empty field, a lone
  `-`, and a lone `.` as an intermediate state. The box commits the value on `Enter` and on
  blur, and not on each keystroke. The box does not clamp the text while the user types.

## Files

- `extensions/cornerstone/src/customizations/segmentationToolbarCustomization.ts` — the
  `threshold-range` option and the limits -1000...1000.
- `platform/ui-next/src/components/DoubleSlider/DoubleSlider.tsx` — the number boxes and
  `handleInputChange`.
- `platform/ui-next/src/components/Numeric/Numeric.tsx` — `Numeric.DoubleRange`, which passes
  `min`, `max`, and `step` to `DoubleSlider`.
- `platform/ui-next/src/components/OHIFToolSettings/RowDoubleRange.tsx` — the row that the
  toolbox draws for a `double-range` option.

`Numeric.SingleRange`, `Numeric.NumberInput`, and `Numeric.NumberStepper` in `Numeric.tsx` show
the same behaviour. Each one is a controlled input, and each one holds no text state.
Therefore each one reverts an empty field. The same fix applies to these three components.

## System Information

System:
  OS: Windows 11 10.0.26200
Browsers:
  Chrome: (latest)

OHIF: `master` (v3.11 line).

eloisalgado:
### Context
Fixes #6269 

This PR fixes issues related to numeric and double slider inputs and implements the expected results and acceptance criteria mentioned in the ticket. New tests were added to cover the new behavior and existing ones were updated accordingly.

### Changes & Results
- Modified `DoubleSlider.tsx` and `Numeric.tsx` components to add optional parameters (`hardMin` and `hardMax`) that allows dynamic set of minimum and maximum values. 
- New values are now set through `blur` or `Enter` key events, no longer with keydown. 
  - This provides better user experience when configuring new values without the need of full text selection and other hacks.
- Components that use such elements were also updated to pass through the new parameters as necessary.
- New tests added

### Testing
Please refer to the original ticket which includes steps to test these changes.

#### PR

- [X] My Pull Request title is descriptive, accurate and follows the

#### Code

- [X] My code has been well-documented (function documentation, inline comments,
  etc.)

  additions or removals.

#### Tested Environment

- [X] OS: MacOS 26.6.2 
- [X] Node version: v24.17.0
- [X] Browser: Chrome 152.0.7977.83 (Official Build) (arm64)

## Summary by CodeRabbit

* **New Features**
  * Numeric range controls now support typed values beyond their default slider bounds when expansion is enabled.
  * Typed expansion can be unrestricted or constrained to configured minimum and maximum limits.
  * Threshold ranges support entering negative and out-of-range values, with slider bounds expanding accordingly.

* **Bug Fixes**
  * Numeric inputs now reliably commit valid values, restore invalid drafts, respect controlled updates, and handle Enter during text composition correctly.
  * Empty range inputs reset to their configured limits.


extensions/cornerstone-dynamic-volume/src/panels/DynamicVolumeControls.tsx
```diff
@@ -179,6 +179,8 @@ const DynamicVolumeControls = ({
                 mode="doubleRange"
                 min={1}
                 max={numDimensionGroups || 1}
+                hardMin={1}
```
wayfarer3130: The default is already to copy the min/max, so this file change is unnecessary.

extensions/cornerstone/src/components/VOIManualControlMenu/VOIManualControlMenu.tsx
wayfarer3130, line 1: Ditto

extensions/cornerstone/src/customizations/segmentationToolbarCustomization.ts
```diff
@@ -726,7 +726,7 @@ const segmentationToolbarButtons: Button[] = [
           name: 'ThresholdRange',
           type: 'double-range',
           id: 'threshold-range',
-          min: -1000,
+          min: 0,
```
wayfarer3130, line 729: This is the one spot that a hardMin/hardMax SHOULD be defined as this one should allow a larger range.
Something like -64k to 64k would be a good range.
It would also be fine to just allow an open range here - any value plus or minus.

platform/docs/src/pages/components/numeric.tsx
```diff
@@ -29,8 +29,10 @@ function NumericPageContent() {
     { name: 'values', type: '[number, number]', default: '—', description: 'Controlled range values (doubleRange mode)' },
     { name: 'defaultValues', type: '[number, number]', default: '[30%, 70%]', description: 'Initial uncontrolled range values' },
     { name: 'onChange', type: '(val: number | [number, number]) => void', default: '—', description: 'Called when any value changes' },
-    { name: 'min', type: 'number', default: '0', description: 'Minimum allowed value' },
-    { name: 'max', type: 'number', default: '100', description: 'Maximum allowed value' },
+    { name: 'min', type: 'number', default: '0', description: 'Minimum value, or the initial slider minimum in doubleRange mode' },
+    { name: 'max', type: 'number', default: '100', description: 'Maximum value, or the initial slider maximum in doubleRange mode' },
+    { name: 'hardMin', type: 'number', default: '—', description: 'Optional hard minimum for typed doubleRange values' },
```
wayfarer3130: Suggest describing what adding a hardMin/hardMax does - it allows values down to or up to the given value to be typed into the text box.   Omitting it allows only down to min/up to max
Otherwise you end up changing the default behaviour of the tool, which isn't desirable - just add the new capability and only if activated should it be allowed.

platform/ui-next/src/components/DoubleSlider/DoubleSlider.test.ts
wayfarer3130, line 1: There should also be a test that one can delete the entire input field, and if one hits enter it resets it, or if one then enters a new value it is ok - that is the issue that the user can't manually back space to delete the -1 to replace it with 30 test.

platform/ui-next/src/components/DoubleSlider/DoubleSlider.tsx
```diff
@@ -28,10 +32,15 @@ const DoubleSlider = React.forwardRef<HTMLDivElement, DoubleSliderProps>(
     ref
   ) => {
     const [value, setValue] = React.useState<[number, number]>(defaultValue);
+    const [effectiveMin, setEffectiveMin] = React.useState(min);
```
wayfarer3130: How about sliderMin and sliderMax?   effectiveMin/max doesn't mean much to me.

platform/ui-next/src/components/DoubleSlider/DoubleSlider.tsx
```diff
@@ -41,10 +50,21 @@ const DoubleSlider = React.forwardRef<HTMLDivElement, DoubleSliderProps>(
...
       ) {
         setValue(defaultValue);
+        setInputValues([defaultValue[0].toString(), defaultValue[1].toString()]);
+        setEffectiveMin(currentMin => Math.min(currentMin, defaultValue[0]));
+        setEffectiveMax(currentMax => Math.max(currentMax, defaultValue[1]));
         prevDefaultValueRef.current = defaultValue;
       }
     }, [defaultValue]);
 
+    React.useEffect(() => {
```
wayfarer3130: @jbocce - is there an alternative to useEffect that is more forwards compatible?

platform/ui-next/src/components/DoubleSlider/DoubleSlider.tsx
```diff
@@ -53,36 +73,85 @@ const DoubleSlider = React.forwardRef<HTMLDivElement, DoubleSliderProps>(
...
+        }
+
+        const hardLimitedValue = Math.max(
+          hardMin ?? -Infinity,
+          Math.min(parsedValue, hardMax ?? Infinity)
+        );
+        const newValue: [number, number] = [...value];
+        const roundedValue = roundToStep(hardLimitedValue);
+        newValue[index] = Math.max(
+          hardMin ?? -Infinity,
```
wayfarer3130: You could set it to min/max if the user has deleted the entries - that would allow resetting the sliderMin/max values potentially

platform/ui-next/src/components/DoubleSlider/DoubleSlider.tsx
wayfarer3130, line 1: I'd suggest allowTypedExpansion=true to allow any expansion, and if you really want to support it, allowTypedExpansion=[hardMin, hardMax]
to set it to a hard range - but the basic allowTypedExpansion is sufficient - it isn't necesary to force a specific hard range.

tests/SegmentationPanel.spec.ts
```diff
@@ -174,12 +194,15 @@ test.describe('Segmentation panel config input validation for labelmap', () => {
       const { opacityInactive } = rightPanelPageObject.labelMapSegmentationPanel.config;
 
       await opacityInactive.fill('0');
```
wayfarer3130, line 196: @jbocce - we should move this bit of test code into a helper since it gets repeated.  However, I'm not quite sure if we should just have a local helper that we re-use just locally or if this belongs in the shared code - any thoughts?

wayfarer3130:
**Suggestion: replace the `useEffect` calls with an adjustment during the render**

This PR adds two `useEffect` calls in `platform/ui-next/src/components/DoubleSlider/DoubleSlider.tsx`, and it keeps three more in `platform/ui-next/src/components/Numeric/Numeric.tsx`. All five effects do the same job. Each effect adjusts state when a prop changes. The React documentation describes this job in [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes), and the documentation recommends a different pattern. In that pattern, the component keeps the previous prop in state, and the component compares the previous prop to the new prop during the render.

**Why the pattern helps the React upgrade**

@jbocce prepares a PR for a newer React version. Three points apply:

- React 18 and React 19 both recommend the pattern. The code then needs no second rewrite at the time of the upgrade.
- React 19 offers the React Compiler. The compiler supports a `setState` call during the render, because the React documentation defines that call as a correct pattern. The compiler also makes the seven `useCallback` calls in `DoubleSlider.tsx` unnecessary.
- StrictMode runs each effect two times in development. An adjustment during the render gives the same result every time, so StrictMode cannot change that result.

**The pattern also corrects two faults today**

1. `prevDefaultValueRef.current` starts at `null`, so the first effect always runs on mount. If `defaultValue` is outside `[min, max]`, the browser paints the narrow track first, and the browser paints the wide track after that. The user sees a flash. The VOI menu supplies such a `defaultValue`.
2. `min`, `max` and `defaultValue` can all change in one commit. The result then depends on the declaration order of the two effects, and no comment states that intent.

The three effects in `Numeric.tsx` cause a different fault. The number field shows the previous value for one frame after each move of the slider handle.

**The proposed code for `DoubleSlider.tsx`**

The two refs and the two effects go away:

```tsx
const [value, setValue] = React.useState<[number, number]>(defaultValue);
const [inputValues, setInputValues] = React.useState<[string, string]>([
  defaultValue[0].toString(),
  defaultValue[1].toString(),
]);
// The initial state holds the expansion for the mount. An effect for the mount is not necessary.
const [effectiveMin, setEffectiveMin] = React.useState(() => Math.min(min, defaultValue[0]));
const [effectiveMax, setEffectiveMax] = React.useState(() => Math.max(max, defaultValue[1]));

// The previous props stay in state, and not in a ref, so the comparison happens during the render.
// See https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
const [prevDefaultValue, setPrevDefaultValue] = React.useState(defaultValue);
const [prevLimits, setPrevLimits] = React.useState({ min, max });

let nextMin = effectiveMin;
let nextMax = effectiveMax;

if (prevLimits.min !== min || prevLimits.max !== max) {
  setPrevLimits({ min, max });
  nextMin = min; // a new series resets the domain
  nextMax = max;
}

if (prevDefaultValue[0] !== defaultValue[0] || prevDefaultValue[1] !== defaultValue[1]) {
  setPrevDefaultValue(defaultValue);
  setValue(defaultValue);
  setInputValues([defaultValue[0].toString(), defaultValue[1].toString()]);
  nextMin = Math.min(nextMin, defaultValue[0]); // the new value then widens the domain
  nextMax = Math.max(nextMax, defaultValue[1]);
}

if (nextMin !== effectiveMin) {
  setEffectiveMin(nextMin);
}
if (nextMax !== effectiveMax) {
  setEffectiveMax(nextMax);
}
```

The order is now explicit. A change of `min` or `max` resets the domain, and the new `defaultValue` then widens the domain. React discards this render, and React runs the component again with the new state. React does this work before React writes to the DOM, so the browser paints one time only.

Each of the three effects in `Numeric.tsx` becomes three lines:

```tsx
const [inputValue, setInputValue] = useState(() => formatDisplayValue(singleValue));
const [prevValue, setPrevValue] = useState(singleValue);
if (prevValue !== singleValue) {
  setPrevValue(singleValue);
  setInputValue(formatDisplayValue(singleValue));
}
```

**Two cautions**

A component can set the state of that same component during the render only. The guard condition is necessary, because the guard condition stops the loop. Please keep a comment with the link to the documentation. A later reader can otherwise change the code back to an effect.

One point belongs to the React upgrade PR, and not to this PR. React 19 makes `ref` a normal prop, and React 19 deprecates `forwardRef`. `DoubleSlider` uses `forwardRef`, and a codemod makes that change.

**A larger option for a separate PR**

`DoubleSlider` can become a fully controlled component, with a `value` prop and an `onValueChange` prop, and with no synchronization from `defaultValue`. That change deletes the `value` state and the whole `defaultValue` branch, because the parent already owns the value. The change breaks the API of the published `@ohif/ui-next` package, so the change needs its own PR.

wayfarer3130 (changes requested):
A number of inline comments need addressing - the most important of which is to be backwards compatible in behaviour without adding parameters.

eloisalgado:
Thank you @wayfarer3130 for your review and comments. 

I have implemented all PR feedback:
- Replaced hardMin/hardMax with allowTypedExpansion: false preserves min/max, true allows any finite value, and [min, max] supports bounded expansion.
- Enabled open typed expansion for segmentation thresholds.
- Renamed internal limits to sliderMin/sliderMax.
- Replaced all five prop-synchronization effects with guarded render-time state adjustments.
- Empty inputs now reset to min/max and collapse the corresponding expanded slider limit.
- Added coverage for deletion/re-entry, Enter reset, bounds, initial expansion, simultaneous prop changes, and controlled Numeric updates.
- Updated Numeric documentation and removed obsolete hard-bound usage.

Please review it again when you can. Thanks in advance!

wayfarer3130 (approved):

tests/SegmentationPanel.spec.ts
jbocce: I am not 100% sure what you are referring to, but if it is the repeated calling to `blur`, then I think the anwser is simple. It should simply move to the `fill` method of the page object. That makes most sense here and fits with the page object paradigm 100%. Or do you mean something else?

PR #6278, opened 2026-09-16
fix(sr): skip derived display sets when placing SR measurements

TFRadicalImaging:
### Context

Loading a structured report while a segmentation is hydrated in another viewport ends the session with `TypeError: Cannot destructure property 'SOPInstanceUID' of 'metadataProvider.getUIDsFromImageID(...)' as it is undefined`.

Hydrating a segmentation replaces its display set's `images` with the labelmap's derived images, so `getImageIdsForDisplaySet` on that display set returns `derived:<uuid>` ids. When an SR loads, `_checkIfCanAddMeasurementsToDisplaySet` walks every active display set and builds a SOPInstanceUID → imageId map from `metadataProvider.getUIDsFromImageID(imageId)`. A `derived:` id is neither WADO-RS nor WADO-URI and is never registered, so that call returns `undefined`. The destructure at `getSopClassHandlerModule.ts` then throws and the rejection reaches the ErrorBoundary.

No upstream issue was filed; the failure is described in full here.

### Changes & Results

- `_checkIfCanAddMeasurementsToDisplaySet` now passes over derived display sets (SEG, RTSTRUCT, SR), next to the `unsupported` ones it already skips.
- An SR measurement references source images, never the images of a derived display set — which is why `hydrateStructuredReport` already filters on `!isDerivedDisplaySet` in both places it picks a display set. This applies the same rule to the load. It also keeps a SCOORD3D measurement off a segmentation, which shares its source's `FrameOfReferenceUID` and would otherwise match.
- Adds a unit test for the SOP class handler (new `jest.config.js` for the extension, since it had none, plus a stand-in for `@ohif/extension-cornerstone`).

**Before:** loading an SR after a segmentation is opened throws and ends the session.
**After:** the SR's measurements land on the same source images as with no segmentation open; the segmentation stays drawn.

### Testing

1. Open a study that has a segmentation and a structured report referencing the same source images.
2. Hydrate the segmentation in one viewport.
3. Drag the SR into a second viewport.

Before this change that throws `Cannot destructure property 'SOPInstanceUID' …`; after it, the SR loads and its measurements appear on the source images.

Unit test: `cd extensions/cornerstone-dicom-sr && npx jest src/getSopClassHandlerModule.test.ts` — 2/2 pass. Negative check: removing the `isDerivedDisplaySet` condition makes the segmentation case fail with the same TypeError.

#### PR

#### Code

#### Tested Environment

- [x] OS: macOS 15
- [x] Node version: 24.3.0
- [x] Browser: Chrome

---

This fix is contributed as a donation by UCalgary. Thanks to @wayfarer3130 for directing it upstream to keep the extension identical to core OHIF.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved DICOM Structured Report measurement handling when reports reference derived segmentations.
  - Prevented errors when referenced image identifiers cannot be resolved.
  - Ensured measurements remain unloaded until a valid source image is available.
  - Preserved correct annotation loading and measurement associations for reports referencing source images.

- **Tests**
  - Added coverage for derived segmentation and source-image measurement scenarios.


wayfarer3130 (approved):
Just waiting for tests to run.

PR #6283, opened 2026-09-18
docs: add the project direction statement

wayfarer3130:
## What this pull request adds

This pull request adds `platform/docs/docs/direction.md`. The document states the direction of four
projects: OHIF, the OHIF Viewer, Cornerstone3D and dcmjs.

No document of this kind exists today. The repository has no `GOVERNANCE`, `CHARTER`, `ROADMAP` or
`MISSION` file.

## What the document contains

- **The four projects, and the kind of each one.** dcmjs is a library. Cornerstone3D is a family of
  libraries. OHIF is a **framework**. The OHIF Viewer is an **application**, and it is separate from
  the framework.
- **Seven goals.** Compliant, correct, fast, easy, customizable, private, compatible.
- **The order of the goals**, and the rule that settles a conflict: `correct > compliant > fast >
  compatible > easy > customizable`. The order is a matter of degree. Privacy is a constraint, and
  the projects do not trade privacy.
- **A table** that says what each goal means for each of the four projects.
- **A definition of "clinically significant"**, with five categories and two quantitative tests.
- **A definition of "fast"**, with five named operations.
- **A rule for data that is not conformant**: record every repair, tell the user only when the error
  is subtle, and let a deployment declare a known deviation of a back end.
- **A privacy statement** and **a regulatory statement**.

## Three corrections to existing text

The document corrects three statements that the repositories make today:

1. The OHIF documentation calls OHIF a **platform** and a **framework**. It never calls OHIF a
   library. The document uses **framework**, and it says that `@ohif/core` and the user-interface
   components are usable as plain libraries.
2. The Cornerstone3D `scope.md` says that Cornerstone3D does **not** load images or parse metadata.
   The document leads the Cornerstone3D statement with **rendering**, and it names
   `@cornerstonejs/dicom-image-loader` and `@cornerstonejs/adapters` as the packages that load and
   write.
3. dcmjs has **no network code**. A search of `src` for `fetch(`, `XMLHttpRequest`, `axios`,
   `node-fetch` and `dicomweb-client` finds nothing. The document therefore says that dcmjs parses,
   manipulates and generates. It does not say that dcmjs fetches.

## Statements that are not yet true

The document is in the present tense. Five statements carry the mark `[Not yet true]`, and the
document lists all five in one table:

- One single-instance DICOM parser exists, and dcmjs is that parser.
- The specifications live in `platform/docs/docs/development/specs/`.
- A verification log for a repair exists in the software.
- The documentation states what the software stores in the browser, and for how long.
- A document lists the public API surface of each project.

## Clauses that stay open

Three clauses carry the mark `[Open]`. Each one states a default, and that default has full force
until the steering committee changes it:

- The two API tiers, and the rule that an undocumented export is internal.
- The length of the deprecation window. The default is two minor releases.
- Accessibility and internationalization as a goal. The default is that it is not a goal.

The order of the goals is **not** open, because the order is the mechanism that applies every other
clause.

## What this pull request does not do

- It does not add the companion document in the Cornerstone3D documentation.
- It does not add `DIRECTION.md` to the dcmjs repository.
- It does not correct the claim "We publish our roadmap" in the Cornerstone3D `README.md`. The OHIF
  process documentation says the opposite: "we don't yet have a long-term or quarterly road map".

Those three changes belong to other repositories, and they need their own pull requests.

## Review

The steering committee is the first reader of this document. Contributors and maintainers are the
first users of it.

The document gives the steering committee the power to reject a pull request that disagrees with a
goal. That rejection is permitted, and it is not automatic. Please read the section **What this
document is** first, because that section states the force of the whole document.

## Summary by CodeRabbit

- **Documentation**
  - Added comprehensive project direction documentation covering goals, boundaries, ownership, governance, and responsibilities.
  - Documented priorities for correctness, compliance, performance, compatibility, usability, and customizability.
  - Clarified privacy requirements, regulatory disclaimers, compatibility expectations, and open decisions.


wayfarer3130:
Going to work on this as externally for a while and when ready add a read-only copy here.

